PR #4, opened 2013-04-23
Jrbl/feature/doc update

jrbl:
- Rename *.yaml -> \1.yml
  - It can be either way (ansible doesn't care) but it should be only one way, and the preponderance of existing configuration used .yml.
  - Update to README to reflect changed filenames.
- Make AWS region selectable on CLI
  - Make it convenient (albeit, mandatory) to set AWS region on the command line when calling ansible-playbook for cloudformation
- README updates
  - Wrap to 80 columns
  - improved markdown conformance and plaintext readability
  - note about setting PYTHONPATH inside squeaky clean virtualenvs
  - note about using WikiWords as your cloudformation stack names
  - note about how sometimes ansible says 'ok' when cloudformation says something broke


jrbl:
@sefk. I don't appear to have commit access. Also I'd like someone else to use the region environment key for further sanity checking.

jarv:
Hi Joe,
I just gave you commit access and merged the pull.
-John

jrbl:
cool, thanks!

PR #6, opened 2013-04-24
Jbau/bugfix/apt get update

jbau:
We were having some issues with ubuntu servers being brought up with cloud formation scripts not being able to 'apt-get install'.  adding an 'apt-get update' fixes this.

### DIFF
--- a/playbooks/files/edx-server-ubuntu-configuration.json
+++ b/playbooks/files/edx-server-ubuntu-configuration.json
@@ -149,7 +149,7 @@
             "  exit 1\n",
             "}\n",
 
-            "apt-get update",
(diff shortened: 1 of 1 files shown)


jarv:
Funny 30 minutes ago I ran across this and did the exact same thing,
One thing that helped me troubleshoot it was adding this to the top of the bash script:

```
            "#!/bin/bash\n",
            "exec >> /home/ubuntu/cflog.log\n",
            "exec 2>> /home/ubuntu/cflog.log\n",
```

I'm about to check in an autoscaling cloudformation config if you haven't worked that out yet, the cloudformation json format is inscrutable

PR #7, opened 2013-04-24
Adding pre-commit and post-checkout hooks

jrbl:
- Pre-commit hook lints JSON
- Post-checkout hook causes hooks directory to update with current
  project hooks
- Readme updated so installation section reminds you to install hooks

The first time you use this, you run util/sync_hooks.sh, and then after that your hooks should get updated when you checkout a branch.

### DIFF
--- a/util/sync_hooks.sh
+++ b/util/sync_hooks.sh
@@ -1,31 +1,28 @@
 #!/bin/bash
+# A small utility to symlink the files from git-hooks/ with filenames ending
+# like .in into the directory .git/hooks/
+#
+# It's intended this be run once near the start of a project by hand, and then
+# subsequently a hook that it installs keeps it running at project checkouts.
 
-# Are we running from the git root?
-GITHOOKSFOUND='false'
-if [ -d git-hooks ]; then
-    GITHOOKSFOUND='true';
-fi
-DOTGITHOOKSFOUND='false'
-if [ -d .git -a -d .git/hooks ]; then
-    DOTGITHOOKSFOUND='true';
-fi
+
+# Save current directory so we can come back; change to repo root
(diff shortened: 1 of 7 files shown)


util/sync_hooks.sh
```diff
@@ -0,0 +1,28 @@
...
+# It's intended this be run once near the start of a project by hand, and then
+# subsequently a hook that it installs keeps it running at project checkouts.
+
+
+# Save current directory so we can come back; change to repo root
+STARTED_FROM=`pwd`
+cd $(git rev-parse --show-toplevel)
+
+# Sync git-hooks directory entries into .git/hooks/
+for file in git-hooks/*.in; do
```
cpennington, line 14: If we remove a hook, this will leave a dangling link. That said, love the idea. Really tempted to set the same sort of thing up in mitx to run pep8 on files that have changed to make sure no new violations have been introduced.

jrbl, line 14: Yeah, I didn't think that the dangling link problem would be such a big deal, but I guess one could do something like `ls -1F .git/hooks/ | grep '[^~@$' | sed -e 's/\(.*\)@$/\1/'` to get the set of symlinks existing in .git/hooks and compare them to the set of .in files, and remove links with missing targets. But at that point, one wonders if it doesn't make more sense to write a Python script instead.  os.fstat() would probably make this easier.  

Kind of feels like a 'future feature' to me though.

I tried to hack this up in such a way that it could easily be picked up and dropped into other repos. I think auto pep8 checking (with whatever flags are necessary to allow 120 columns, natch) and pylinting and auto jshinting etc. are really helpful. And people can always do --no-verify on their commit, but it forces them to acknowledge that they're doing something icky. :)

PR #8, opened 2013-04-24
Jarv/multi instance

jarv:


jrbl:
Looking through the diffs, I didn't see that you have added the apt parts to any of the ansible bits yet, is that right?

PR #10, opened 2013-04-24
Feature/jarv/move vars to role

jarv:


playbooks/roles/common/vars/main.yml
```diff
@@ -1,10 +1,5 @@
-# Variables for all playbooks
-#
-# These variables should apply to all roles and environments. 
-# All definitions can be overrided in the 
-# the group files that are in this directory
-# 
-# If a new variable is added please document it!
+# Variables for the common role
```
jbau, line 1: i see that that the `common` role is in all edxapps.  so this is good

jbau:
I will run this branch once and merge if/when it succeeds

jbau:
my test deploy succeeded, so i think it's good to merge.

PR #11, opened 2013-04-24
Implements simple rolling deploy logic.

jarv:
Requires the latest ansible to use (you may need to remove and add it again if
you are in a virtualenv)

See `edxapp_rolling_example.yml`

`ansible-playbook -v --user=ubuntu  edxapp_rolling_example.yml -i ./ec2.py  --private-key=/path/to/key --module-path /path/to/repo/modules`

A new module is necessary for ELB information since that is not
available in EC2 metadata (which is what ec2_facts fetches).

ec2_facts grabs the instance ids and other meta information
ec2_elb_facts fetches ELB information
elb_reg.py takes an elb and instance id and will either register or
deregister. (called as pre and post).

`Serial` should be set to the number of instances that you want to
operate on simultaneously, currently set to 1.

### DIFF
--- a/playbooks/util/elb_reg.py
+++ b/playbooks/util/elb_reg.py
@@ -1,0 +1,76 @@
+#!/usr/bin/env python
+
+from argparse import ArgumentParser
+import time
+import boto
+
+
+def await_elb_instance_state(lb, instance_id, awaited_state):
+    """blocks until the ELB reports awaited_state
+    for instance_id.
+    lb = loadbalancer object
+    instance_id : instance_id (string)
+    awaited_state : state to poll for (string)"""
+
+    start_time = time.time()
+    while True:
+        state = lb.get_instance_health([instance_id])[0].state
+        if state == awaited_state:
+            print "Load Balancer {lb} is in awaited state " \
+                  "{awaited_state}, proceeding.".format(
+                  lb=lb.dns_name,
+                  awaited_state=awaited_state)
+            break
+        else:
+            print "Checking again in 2 seconds. Elapsed time: {0}".format(
+                time.time() - start_time)
+            time.sleep(2)
+
+
+def deregister():
+    """Deregister the instance from all ELBs and wait for the ELB
+    to report them out-of-service"""
+
+    for lb in active_lbs:
+        lb.deregister_instances([args.instance])
+        await_elb_instance_state(lb, args.instance, 'OutOfService')
+
+
+def register():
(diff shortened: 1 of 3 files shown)


jarv:
merging this since it does not affect anything that's already in place.

playbooks/util/elb_reg.py
```diff
@@ -0,0 +1,76 @@
...
+def deregister():
+    """Deregister the instance from all ELBs and wait for the ELB
+    to report them out-of-service"""
+
+    for lb in active_lbs:
+        lb.deregister_instances([args.instance])
+        await_elb_instance_state(lb, args.instance, 'OutOfService')
+
+
+def register():
```
jrbl, line 39: We were just looking at this stuff @jarv and got concerned. It looks like you're adding the host to every available load balancer. But shouldn't you build a list of load balancers it was part of when you deregister it?  

Not that I have a good suggestion about how to do that. But we (already) have lots of load balancers on one set of AWS creds, and don't want instances added to them willy-nilly.

jarv, line 39: Here is the sequence broken down, it's not intuitive at all unfortunately looking at the playbook (I should have added more comments to the example)

>  pre_tasks:
> - name: Gathering ec2 facts 
>    ec2_facts: 
> - name: Gathering ELB facts
>    local_action: ec2_elb_facts

These two modules "ec2_facts" and "ec2_elb_facts" are invoked in the pre_tasks and give us the following facts which are variables that can be used in the playbook:
- `$elbs` which is `{ '<instanceid-1>': [ <elb1>, <elb2>, .. ], <instanceid-2>... }` Where elb1, elb2 are the elbs attached to the instance.
- `$ansible_ec2_instance_id` which is the current instance that operations are being run on

The way the command is invoked is as follows:

> - local_action: command util/elb_reg.py -e {{ ",".join(elbs[ansible_ec2_instance_id]) }} -i {{ ansible_ec2_instance_id }} deregister

Where -e is the list of elbs that the current instances belong to and -i is the instance
"active_elbs" are the elbs that are passed in.

> - local_action: command util/elb_reg.py -e {{ ",".join(elbs[ansible_ec2_instance_id]) }} -i {{ ansible_ec2_instance_id }} register

Register will pass in the same elb list and the same instance id to add it back to the pool

jbau, line 39: Thanks for explaining.

PR #14, opened 2013-04-26
change module->library, update the example playbook

jarv:
By default ansible will load libs in library/ which means
you do not need to specify a module path if you are invoking
ansible from the playbooks/ dir.

Added some comments to the rolling example


jrbl:
Thanks @jarv, that makes it a lot clearer.

PR #15, opened 2013-04-26
Jrbl/feature/git checkout

jrbl:
- This is the bits needed for automatically checking out the mitx repository to /opt/wwc.

### DIFF
--- a/playbooks/roles/lms/tasks/main.yml
+++ b/playbooks/roles/lms/tasks/main.yml
@@ -3,11 +3,11 @@
 #  - nginx/tasks/main.yml
 ---
 - name: create lms application config
-  template: src=env.json.j2 dest=/opt/wwc/lms.env.json
+  template: src=env.json.j2 dest=$app_base_dir/lms.env.json
   sudo: True
 
 - name: create lms auth file
-  template: src=auth.json.j2 dest=/opt/wwc/lms.auth.json
+  template: src=auth.json.j2 dest=$app_base_dir/lms.auth.json
   sudo: True
 
 - include: ../../nginx/tasks/nginx_site.yml state=link site_name=lms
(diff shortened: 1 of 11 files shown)


playbooks/roles/lms/tasks/main.yml
```diff
@@ -5,8 +5,30 @@
...
+  copy: src=vars/secure/ssh_deploy_private dest=/home/ubuntu/.ssh/id_rsa force=yes owner=ubuntu group=ubuntu mode=600
+- name: install read-only ssh key for mitx repo (public)
+  copy: src=vars/secure/ssh_deploy_public dest=/home/ubuntu/.ssh/id_rsa.pub force=yes owner=ubuntu group=ubuntu mode=644
+- name: install read-only ssh key for mitx repo (host github known)
+  copy: src=vars/secure/ssh_deploy_known_hosts dest=/home/ubuntu/.ssh/known_hosts force=yes owner=ubuntu group=ubuntu mode=600
+
+# Check out mitx repo to /opt/wwc
+- name: set permissions on /opt/wwc sgid for edx
+  file: path=/opt/wwc owner=root group=edx mode=2775 state=directory
+  file: path=/opt/wwc/mitx owner=ubuntu group=edx mode=2775 state=directory
```
jarv: propose that we create a `$app_base_dir` var in group_vars/all in case we decide to change the location later.

jrbl: Ok, I just pushed an update that sets app_base_dir in group_vars/all and
then uses it everywhere. I believe the string /opt/wwc only appears in one
place now.  I also added some tags to the nginx tasks so they could be run
with -t nginx.

playbooks/roles/lms/tasks/main.yml
```diff
@@ -3,10 +3,32 @@
...
+# Check out mitx repo to $app_base_dir
+- name: set permissions on $app_base_dir sgid for edx
+  file: path=$app_base_dir owner=root group=edx mode=2775 state=directory
+  file: path=$app_base_dir owner=ubuntu group=edx mode=2775 state=directory
+  sudo: True
+- name: install git and its recommends
+  apt: pkg=git state=installed install_recommends=yes 
+  sudo: True
+- name: git checkout mitx repo into $app_base_dir
+  git: dest=$app_base_dir/mitx repo=git@github.com:MITx/mitx.git
```
jarv, line 34: We may want to put the repo in a var as well, MITx/mitx.git will become edX/mitx.git on May 1st and then edX/edx.git sometime on or before June 1st.

PR #17, opened 2013-04-30
Feature/jbau/env parameterizaton

jbau:
@jarv what do you think about this setup for instance parameterization / secrets?


jbau:
I moved the reference to the private vars/secure repo outside of this repo.

PR #21, opened 2013-05-01
Playbook that installs ruby.

jbau:
By default it installs 1.9.3-p374 in /opt/www/.rbenv and gems go
in /opt/www/.gem, which replicates the vagrant box

@jarv what do you think?

### DIFF
--- a/playbooks/group_vars/all
+++ b/playbooks/group_vars/all
@@ -1,5 +1,6 @@
 ---
 app_base_dir: /opt/wwc
+ruby_base: /opt/www
 
 #where are the secure files on the deploying machine?
 secure_file_dir: ../../edx-secret/ansible/vars/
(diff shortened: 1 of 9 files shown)


playbooks/group_vars/all
```diff
@@ -1,5 +1,6 @@
 ---
 app_base_dir: /opt/wwc
+ruby_base: /opt/www
```
jarv: I was thinking that group_vars/all would be limited to variables that are global to all roles, in this case I think the right place to put this is in roles/ruby/vars/main.yml

playbooks/roles/common/tasks/main.yml
```diff
@@ -3,6 +3,9 @@
 - name: Create application root
   sudo: True
   file: path=$app_base_dir state=directory owner=root group=root
+- name: Create ruby base
+  sudo: True
+  file: path=$ruby_base state=directory owner=www group=www
```
jarv: would this be better put in playbooks/roles/ruby/tasks/main.yml?

jbau:
@jarv made the changes you suggested.

PR #25, opened 2013-05-01
high-level overview of ssh config

e0d:


sefk:
@e0d -- I think I've found a nice feature that will make this ssh config much simpler.  I had dumped a bunch of this into hipchat, copy/pasting in here:

The VPC has a setting that creates internal DNS names within the VPC. So you get names like this:
`ip-10-0-11-97.us-west-1.compute.internal` the bastion host can resolve those names.  It's this nice setting on the VPC.  This will set up the DNS in the VPC so it can resolve these internal names now.  That means we won'tt have to maintain lots of host mappings in our ~/.ssh/config, just one that looks like this:

```
Host *.us-west-1.compute.internal
  ProxyCommand ssh -W %h:%p sef-edx-prodtues3
  ForwardAgent yes
  User ubuntu

Host sef-edx-prodtues3
  HostName 54.241.173.203
  ForwardAgent yes
  User ubuntu
```

You get the idea.  

Unfortunately the command to turn this on is selectible through the UI (under the main VPC configuration, the setting is "Enable DNS hostname support for instances launched in this VPC.") and via the command line tools, but not through CloudFormation as far as I can tell.  

So I'll submit a pull request with this as an additional manual command a la Elasticache.

sefk:
Pulling this for now since it's a helpful comment, I'll revise the README with my changes when I make them.

PR #26, opened 2013-05-01
Install debian and python dependencies for LMS/CMS

jrbl:
- Create a global variable lms_source_repo to make it easy to update
  where to pull mitx from when mitx moves
- Create global variable venv_base to store location of edx virtual
  environment, defaults to /opt/edx
- Create edx virtual environment to store python modules
- Add tags to LMS and CMS tasks to make selective running and testing
  easier
- Install a long, static list of debian packages, which has been
  manually curated from other edx configurations; this probably needs
  more attention and in the future should probably be divided better
  among various roles.
- Install the list of python packages distributed with the mitx
  repository
- Install some really useful software, namely mosh and tmux


jrbl:
oops, merge conflict. I'll resolve that now.

playbooks/group_vars/all
```diff
@@ -1,2 +1,4 @@
 ---
 app_base_dir: /opt/wwc
+venv_dir: /opt/edx
+lms_source_repo: git@github.com:MITx/mitx.git
```
jarv: I think we should put `lms_source_repo` in the lms role vars/main.yml

playbooks/roles/lms/tasks/main.yml
```diff
@@ -17,18 +22,165 @@
...
+  - puppetmaster
+  - puppetmaster-common
+  - puppet-lint
+  - vim-puppet
+  - python-nosexcover
+  - libwww-perl
+  - libswitch-perl
+  - libcrypt-ssleay-perl
+  - libnet-amazon-ec2-perl
+  - github-cli
```
jarv: Can we move the list to roles/lms/vars/main.yml?

playbooks/group_vars/all
jrbl: Ok, I'll move that now.

jrbl:
I resolved the merge conflict and rebased on current master. Should be a bit nicer with the apt list alphabetized. I'll make the suggested change now and re-push in a few minutes.

playbooks/group_vars/all
```diff
@@ -1,5 +1,6 @@
 ---
 app_base_dir: /opt/wwc
+venv_dir: /opt/edx
```
jarv, line 3: I'm thinking `app_base_dir` and `venv_dir` should move to the common role but we can do that later once we figure out the secure_dir stuff.

jrbl, line 3: Yeah, I think you should make a trello so we remember to do this, and then let's put it off.  I think some of the other things belong in common ultimately too.

PR #28, opened 2013-05-02
Refactor how we manage secure data, git-identity changes

jarv:
- I've moved the example secure directory to playbooks/secure_example,
  the directory structure in your secure repo should mirror that in
  secure_example/
- If you run the lms role without a valid git-identity file it will
  fail, there is an empty file checked into the secure_example dir
- Using GIT_SSH env for git checkout, we will also need to use that
  for pip installs that use git if there are private repos in
  requirements.


README.md
```diff
@@ -39,6 +39,46 @@ version instead of the official v1.1 release._
...
+
+* Usernames
+* Public keys (keys are ok to be public, but can be used to figure out usernames)
+* Hostnames
+* Passwords, api keys
+
+The folowing yml files and examples serve as templates that should be overridden with your own
+environment specific configuration:
+
+* vars in `vars/secure_default/`
```
jbau: i think you now mean `secure_example` here and in the next line

playbooks/edxapp_custom.yml
```diff
@@ -1,10 +1,10 @@
 - hosts: tag_Group_edxapp_custom
   vars_files:
   # using conditional loading to override defaults for site-specific installs
-    - ["{{ secure_file_dir }}/edxapp_stage_vars.yml", "vars/secure_default/edxapp_stage_vars.yml"]
```
jbau, line 4: by removing the list of paths to search, are you envisioning that  in operation people would change the `secure_dir` variable in the `group_vars/all` file?  I kind of understand that these putting search paths everywhere is ugly, but I wish we could use the repo as is in actual operation.  

Maybe the answer is to use --extra-vars on the command line to override?  Let me check if this works with variable precedence.  If it works I think we should probably add it to the readme.

README.md
jarv: yeah, fixed.

playbooks/edxapp_custom.yml
jarv, line 4: --extra-vars will work too though I was thinking that group_vars/tag_Group_<name> will override what is in group_vars/all

For example the Group tag on my test EC2 instances is  `another_multi` so `group_vars/tag_Group_anothermulti` has:

```
---
# this path is relative to the playbook dir
secure_dir: '../../configuration-secure/ansible'
```

Which is the relative path to my secure repo.

jbau, line 4: I see, so are those can persist by being untracked files... that could also be a good way

jarv, line 4: yeah, or checked in even for us.  Others outside of edx who use it they would create their own private files in group_vars/ that override the default secure_dir that's set in group_vars/all.

playbooks/roles/lms/tasks/main.yml
```diff
@@ -39,7 +40,6 @@
 # Check out mitx repo to $app_base_dir
 - name: set permissions on $app_base_dir sgid for edx
   file: path=$app_base_dir owner=root group=edx mode=2775 state=directory
-  file: path=$app_base_dir owner=ubuntu group=edx mode=2775 state=directory
```
jarv, line 42: This looked like a dupe so i removed it

playbooks/edxapp_custom.yml
```diff
@@ -1,10 +1,10 @@
...
-    - ["{{ secure_file_dir }}/edxapp_custom_vars.yml", "vars/secure_default/edxapp_custom_vars.yml"]
-    - ["{{ secure_file_dir }}/users.yml", "vars/secure_default/users.yml"]
-    - ["{{ secure_file_dir }}/edxapp_stage_users.yml", "vars/secure_default/edxapp_stage_users.yml"]
+    - "{{ secure_dir }}/edxapp_stage_vars.yml"
```
jbau: looks like we need all of these paths to be `{{ secure_dir }}/vars/...`, or else we should change how `secure_dir` is structured.

jbau:
works after testing. merging.

PR #29, opened 2013-05-02
cloudformation template changes for outbound SSH, internal DNS names

sefk:
Three changes in here
- **New SSH tunnel method** use private DNS names so we don't have to maintain individual host maps in ssh configs.  instead have ec2.py return the private name (only resolvable within the VPC) and then rely on that to be resolvable on the bastion host.  That way we can use a general route to the bastion host in ssh config.  Updated the readme accordingly.
- **Up limit on app servers** no reason to limit
- **Outbound port 22** so we can fetch over ssh from github.com.  

Please take a look @e0d.


playbooks/ec2.ini
```diff
@@ -1,6 +1,6 @@
 [ec2]
 regions=all
 destination_variable=public_dns_name
-vpc_destination_variable=ip_address
+vpc_destination_variable=private_dns_name
```
sefk, line 4: this way ec2.py returns values that look like `ip-10-0-31-115.us-west-1.compute.internal` which of course is bogus outside a VPC, but is resolvable within the VPC.  Nice!

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -3192,6 +3200,19 @@
...
       }
+    },
+    "EnableDNSHostnamesForVPCCommand":{
+      "Description":"MANUAL COMMAND #3: Change VPC setting to enable automatic DNS hostname resolution from within the VPC. Setting not supported in Cloudformation yet.",
+      "Value":{
+        "Fn::Join":[
+          "",
+          [
+            "ec2-modify-vpc-attribute ",
+            "-c ", { "Ref":"EdxVPC" }, " ",
```
sefk, line 3211: http://docs.aws.amazon.com/AmazonVPC/latest/UserGuide/vpc-dns.html explains how dns hostnames in VPC's work

PR #37, opened 2013-05-02
cloudformation: open tcp/9418 for git

sefk:
We need to use the git protocol to access read-only git repos.  It's a quirk that when these are referenced from pip files, they need to go over a source control protocol.  I wish it could go over http/s or ssh, but it can't, so port 9418 it is.

I had to open inbound and outbound 9418 to get this working, don't know why.  Leaving both open for now, will debug later so we can close down.

### DIFF
--- a/cloudformation_templates/edx-reference-architecture.json
+++ b/cloudformation_templates/edx-reference-architecture.json
@@ -1316,6 +1316,12 @@
             "CidrIp":{
               "Ref":"SSHLocation"
             }
+          },
+          {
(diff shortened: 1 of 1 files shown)


cloudformation_templates/edx-reference-architecture.json
```diff
@@ -3123,7 +3135,7 @@
                 "Endpoint.Address"
               ]
             },
-            "'\n",
+            "',\n",
```
sefk, line 3138: opportunistic fix while I was in here

PR #38, opened 2013-05-02
role for upstart tasks, parameterized in the style the nginx role

jbau:
where it can take a {{ service_variant }} param

### DIFF
--- a/playbooks/roles/gunicorn/templates/lms.conf.j2
+++ b/playbooks/roles/gunicorn/templates/lms.conf.j2
@@ -1,0 +1,37 @@
+# gunicorn
+
+description "gunicorn server"
+author "Calen Pennington <cpennington@mitx.mit.edu>"
+
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+respawn
+respawn limit 3 30
+
+env PID=/var/tmp/lms.pid
+#env NEW_RELIC_CONFIG_FILE=${app_base_dir}/opt/wwc/newrelic.ini
+#env NEWRELIC=${app_base_dir}/bin/newrelic-admin
+env WORKERS=1
+env PORT=8000
+env LANG=en_US.UTF-8
+env DJANGO_SETTINGS_MODULE=lms.envs.aws
+env SERVICE_VARIANT="lms"
+
+
+chdir ${app_base_dir}/mitx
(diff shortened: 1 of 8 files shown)


playbooks/roles/gunicorn/templates/lms.conf.j2
```diff
@@ -0,0 +1,37 @@
...
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+respawn
+respawn limit 3 30
+
+env PID=/var/tmp/lms.pid
+#env NEW_RELIC_CONFIG_FILE=${app_base_dir}/opt/wwc/newrelic.ini
+#env NEWRELIC=${app_base_dir}/bin/newrelic-admin
+env WORKERS=1
```
jarv: This should scale depending on the number of cores, maybe just make it a variable for now?

playbooks/roles/lms/tasks/main.yml
```diff
@@ -111,3 +114,6 @@
   tags:
   - lms
   - cms
+  
+- include: ../../gunicorn/tasks/main.yml
```
jarv: Why not just include the role in the top-level playbook?  Or maybe we should add these tasks to the lms role since they are closely related?

jbau: yup.  I'm fine with the top level for installing gunicorn.  I do think that the upstart script creation would be ok as an include in the manner of nginx

playbooks/roles/gunicorn/templates/lms.conf.j2
jbau: sounds good.

playbooks/roles/lms/tasks/main.yml
jarv: Actually if we are creating a role per service variant (lms, cms, lms-preview, etc) which will each have their own upstart job maybe the right thing to do is include it with the lms, cms, lms-preview role.

jbau: yeah, so i think we could leave it here then?  I've moved the pip gunicorn install to the top-level playbook

jbau:
how 'bout now?

jarv:
yeah looks good to me

PR #39, opened 2013-05-03
Partial CMS configuration

jrbl:
- Needs more work, but seems to have many of the basics.

This is at least part of a CMS setup. I'd love to have @jarv and @jbau look this over, or possibly pull it into a local branch and finish the job.

### DIFF
--- a/playbooks/roles/cms/templates/auth.json.j2
+++ b/playbooks/roles/cms/templates/auth.json.j2
@@ -1,0 +1,1 @@
+{{ auth_config | to_nice_json }}
--- a/playbooks/roles/cms/templates/auth.json.j2
+++ b/playbooks/roles/cms/templates/auth.json.j2
@@ -1,1 +1,2 @@
+{# {% do auth_config.update( { < override some hash keys > } )  %} #}
 {{ auth_config | to_nice_json }}
--- a/playbooks/roles/cms/vars/main.yml
+++ b/playbooks/roles/cms/vars/main.yml
@@ -1,0 +1,29 @@
+# variables common to the cms role, automatically loaded
+# when the role is included
+---
+auth_config:
+  'CONTENTSTORE':
+    'ENGINE':  'xmodule.contentstore.mongo.MongoContentStore'
+  'MODULESTORE':
+    'default':
+      'ENGINE':  'xmodule.modulestore.mongo.MongoModuleStore'
+ 
+env_config:
+  'CACHES':
+    'default':
+      'BACKEND':  'django.core.cache.backends.memcached.MemcachedCache'
+      'KEY_FUNCTION':  'util.memcache.safe_key'
+    'mongo_metadata_inheritance':
+      'BACKEND':  'django.core.cache.backends.memcached.MemcachedCache'
+      'KEY_FUNCTION':  'util.memcache.safe_key'
+      'TIMEOUT': 300
+    'staticfiles':
+      'BACKEND':  'django.core.cache.backends.memcached.MemcachedCache'
+      'KEY_FUNCTION':  'util.memcache.safe_key'
+  'LOG_DIR': '/mnt/logs/edx'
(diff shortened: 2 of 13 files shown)


playbooks/roles/cms/templates/auth.json.j2
```diff
@@ -0,0 +1 @@
+{{ auth_config | to_nice_json }}
```
jarv, line 2: So for the slight variations that will be in the auth and env json files for the cms I was thinking we could do something like this in the jinja template

```
 {% do auth_config.update( { < override some hash keys > } )  %}
 {{ auth_config | to_nice_json }}
```

I couldn't think of any cleaner way to do it.

playbooks/roles/cms/vars/main.yml
```diff
@@ -0,0 +1,29 @@
...
+    'staticfiles':
+      'BACKEND':  'django.core.cache.backends.memcached.MemcachedCache'
+      'KEY_FUNCTION':  'util.memcache.safe_key'
+  'LOG_DIR': '/mnt/logs/edx'
+  'LOGGING_ENV': 'lms-dev'
+  'SITE_NAME': 'studio.lms-dev.m.edx.org'
+  'SYSLOG_SERVER': 'syslog.a.m.i4x.org'
+  'LMS_BASE': 'lms-dev.m.edx.org'
+  'SESSION_COOKIE_DOMAIN': '.lms-dev.m.edx.org'
+
```
jarv, line 29: I'm actually not certain if roles have an isolated variable namespace unfortunately.
There was a recent thread on the ansible list that I commented on here -> https://groups.google.com/forum/?fromgroups=#!topic/ansible-project/rUPBdFP6SVc

We may just want to share the same variable for cms, lms, lms-preview and override specific keys in the jinja template.

playbooks/roles/lms/vars/main.yml
```diff
@@ -97,11 +97,11 @@ lms_debian_pkgs:
...
-  - puppet
-  - puppet-common
-  - puppet-lint
-  - puppetmaster
-  - puppetmaster-common
+  #- puppet
+  #- puppet-common
+  #- puppet-lint
+  #- puppetmaster
+  #- puppetmaster-common
```
jarv: yeah we can definitely remove these :)

playbooks/roles/nginx/templates/cms.j2
```diff
@@ -0,0 +1,97 @@
...
+  if ($http_x_forwarded_proto = "http") {
+    set $do_redirect "true";
+  }
+
+  # Run our actual redirect...
+  if ($do_redirect = "true") {
+    rewrite ^ https://$host$request_uri? permanent;
+  }
+
+}
```
jarv, line 97: Do you think you guys are going to use your own template with custom domain info or should we template this one out with variables?

playbooks/roles/cms/templates/auth.json.j2
jrbl, line 2: I think I see what you mean here, though at this point I still don't have any real idea what these values should be. :)

playbooks/roles/cms/vars/main.yml
jrbl, line 29: Yes, I think this sounds reasonable. Or maybe carrying around a dictionary or namespacing as you suggested. But between lms and cms, most of these values seem to be the same anyway.

playbooks/roles/lms/vars/main.yml
jrbl: This is probably trumped by @e0d's branch to skim cruft and move things around. woot.

playbooks/roles/nginx/templates/cms.j2
jrbl, line 97: We'll have our own, like we do for LMS.

PR #41, opened 2013-05-03
Jarv/better elb rolling

jarv:
This is a re-write of the ec2_elb_facts module so that the de-reg and register is done by the module.  The main advantage is that you can now see when action is taken (will report changed when an instance is registered and deregistered).  It also removed the extra call to a python script which I think is nice.


playbooks/library/ec2_elb
```diff
@@ -0,0 +1,198 @@
...
+# along with Ansible.  If not, see <http://www.gnu.org/licenses/>.
+
+DOCUMENTATION = """
+---
+module: ec2_elb
+short_description: De-registers or registers instances from EC2 ELB(s)
+description:
+  - This module de-registers or registers an AWS EC2 instance from the ELB(s)
+    that it belongs to.
+  - Returns fact "elb_ec2" which is a list of elbs attached the instance
```
jbau: i think you want "attached the instance" -> "that detached the instance" or something similar

playbooks/library/ec2_elb
```diff
@@ -0,0 +1,198 @@
...
+
+        while True:
+            state = lb.get_instance_health([self.instance_id])[0].state
+            if state == awaited_state:
+                break
+            else:
+                time.sleep(1)
+
+    def _get_instance_lbs(self, ec2_elbs=None):
+        """Returns a list of ELBs attached to self.instance_id"""
```
jbau: Maybe expand this docstring to explain what it does when the ec2_elbs kwarg is supplied, which doesn't quite fit the function name (i.e. returning the subset of ec2_elbs that are active?)

playbooks/library/ec2_elb
jbau: also, I don't see `elb_ec2` returned.  Isn't it `ec2_elbs` according to line 190?

jbau: or maybe you want to change L190 instead?

playbooks/library/ec2_elb
```diff
@@ -0,0 +1,198 @@
...
+    required: true
+
+  instance_id:
+    description:
+      - EC2 Instance ID
+    required: true
+
+  ec2_elbs:
+    description:
+      - List of ELB names, required for registration. The elb_ec2 fact should be used if there was a previous de-register.
```
jbau: same comment about `elb_ec2` versus `ec2_elbs`

playbooks/library/ec2_elb
```diff
@@ -0,0 +1,198 @@
...
+    instance_id = module.params['instance_id']
+    elb_man = ElbManager(module, instance_id, ec2_elbs, ec2_access_key,
+                         ec2_secret_key)
+
+    if module.params['state'] == 'register':
+        elb_man.register()
+    elif module.params['state'] == 'deregister':
+        elb_man.deregister()
+
+    ansible_facts = {'ec2_elbs': [lb.name for lb in elb_man.lbs]}
```
jbau, line 193: it may be that you want `elb_ec2` here instead of `ec2_elbs`

playbooks/edxapp_rolling_example.yml
```diff
@@ -8,21 +8,20 @@
   pre_tasks:
     - name: Gathering ec2 facts 
       ec2_facts:
```
jbau, line 10: I'm curious why we need to invoke `ec2_facts` before running the rest of the playbook.

playbooks/library/ec2_elb
```diff
@@ -0,0 +1,198 @@
...
+            ec2_secret_key={'default': None, 'aliases': ['EC2_SECRET_KEY']},
+            ec2_access_key={'default': None, 'aliases': ['EC2_ACCESS_KEY']}
+        )
+    )
+
+    ec2_secret_key = module.params['ec2_secret_key']
+    ec2_access_key = module.params['ec2_access_key']
+    ec2_elbs = module.params['ec2_elbs']
+
+    if module.params['state'] == 'register' and 'ec2_elbs' not in module.params:
```
jbau: i think you meant `present`

playbooks/library/ec2_elb
```diff
@@ -0,0 +1,198 @@
...
+    if not ec2_secret_key and 'EC2_SECRET_KEY' in os.environ:
+        ec2_secret_key = os.environ['EC2_SECRET_KEY']
+    if not ec2_access_key and 'EC2_ACCESS_KEY' in os.environ:
+        ec2_access_key = os.environ['EC2_ACCESS_KEY']
+
+    instance_id = module.params['instance_id']
+    elb_man = ElbManager(module, instance_id, ec2_elbs, ec2_access_key,
+                         ec2_secret_key)
+
+    if module.params['state'] == 'register':
```
jbau: same here and in L187:  you probably want `present` and `absent`

playbooks/edxapp_rolling_example.yml
jarv, line 10: I think that's going to be the best way to get the instance ID, otherwise we would need to figure it out from the dns name which we might be able to do but using instance ID is always going to work.

playbooks/library/ec2_elb
jarv: The module returns a json object and ec2_elbs is a fact, sorry for the naming mixup should be fixed now (had trouble settling on a naming convention)

playbooks/library/ec2_elb
jarv: done

jarv:
addressed comments

jbau:
ok lgtm

PR #46, opened 2013-05-08
Feature/jrbl/prod cleanups

jrbl:
an extension of the partial cms setup (not really much more done to set that up yet, I'm afraid), with a couple nickel and dime fixes needed on the vpc hosts to be (hopefully) prod worthy


playbooks/roles/cms/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+  - cms
+
+- include: ../../nginx/tasks/nginx_site.yml state=link site_name=cms
+  tags:
+  - cms
+  - cms-env
+- include: ../../nginx/tasks/nginx_site.yml state=link site_name=cms-backend
+  tags:
+  - cms
+  -cms-env
```
jarv: it looks like we have a couple conventions for spacing around lists, do you prefer aligning the "-"'s with the key name or indenting?

playbooks/roles/cms/vars/main.yml
```diff
@@ -0,0 +1,29 @@
+# variables common to the cms role, automatically loaded
+# when the role is included
+---
+auth_config:
```
jarv: I'm not sure how the precedence will work if auth_config and env_config is defined in each role (pretty sure it's undefined if the same var is defined in two roles/<name>/vars/main.yml files).

The way I was thinking we would do this is defining these two vars one time (right now they are defined in the lms role but maybe we should move them) and then override keys in the cms, lms, and lms-preview roles.

playbooks/roles/cms/vars/main.yml
```diff
@@ -0,0 +1,29 @@
...
+      'TIMEOUT': 300
+    'staticfiles':
+      'BACKEND':  'django.core.cache.backends.memcached.MemcachedCache'
+      'KEY_FUNCTION':  'util.memcache.safe_key'
+  'LOG_DIR': '/mnt/logs/edx'
+  'LOGGING_ENV': 'lms-dev'
+  'SITE_NAME': 'studio.lms-dev.m.edx.org'
+  'SYSLOG_SERVER': 'syslog.a.m.i4x.org'
+  'LMS_BASE': 'lms-dev.m.edx.org'
+  'SESSION_COOKIE_DOMAIN': '.lms-dev.m.edx.org'
```
e0d: these values are coming in from Vagrant, probably want the session cookie domain to be dynamic based on fqdn, lms base too.  Also the memcached references here have no config to point to the actual backend.  Should like something like

```
"default": {
      "BACKEND": "django.core.cache.backends.memcached.MemcachedCache",
      "LOCATION": [
        "deploycache-large.foo-bar.amazonaws.com:11211",
        "deploycache-large.bar-baz.amazonaws.com:11211"
      ],
      "KEY_FUNCTION": "util.memcache.safe_key",
      "KEY_PREFIX": "stage.edx.org"
    },
```

playbooks/roles/common/tasks/main.yml
```diff
@@ -1,11 +1,24 @@
...
+  file: path=/mnt/logs state=directory mode=2770 group=adm owner=root
   sudo: True
+  tags:
+  - pre_install
+- name: Create aliases to the log directory
+  file: state=link src=/mnt/logs path=$app_base_dir/log
+  tags:
+  - pre_install
+- name: Touch the edx log file into place 
+  command: touch -a /mnt/logs/edx.log
```
e0d, line 17: This logfile should be replaced by /mnt/logs/*/edx.log where \* is lms|lms-xml|cms|lms-preview

playbooks/roles/cms/tasks/main.yml
jrbl: oh, um. I prefer the style in the 'create cms auth file' play. I'll fix the other.

playbooks/roles/in_production/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
...
+- name: Make sure LMS is running
+  service: name=lms state=started
+  tags:
+  - production
+
+- name: Disable HTTP Basic Auth on site
+  file: path=/etc/nginx/sites-enabled/basic-auth state=absent
+  notify: restart nginx
+  tags:
+  - production
```
jarv, line 18: Another way to do this is to pass an argument (or define a variable) to the lms and nginx role which will cause the service to be started.

in the top playbook:

```
roles:
  - lms  lms_start=true
...
```

in the lms role, as either a handler or a task

```
- name: Start the LMS
  service: name=lms state=started
  when: lms_start is true

```

Also something similar for http auth.

playbooks/roles/nginx/templates/cms.j2
```diff
@@ -0,0 +1,97 @@
...
+  location /heartbeat {
+    try_files $uri @proxy_to_cms_app;
+  }
+
+  # Check security on this
+  location ~ /static/(?P<file>.*) {
+    auth_basic            "Restricted";
+    auth_basic_user_file  /etc/nginx/nginx.htpasswd;
+    root {{app_base_dir}};
+    try_files /staticfiles/$file /course_static/$file =404;
```
e0d: You won't need the course_static stuff here as this is XML LMS only.

playbooks/roles/nginx/templates/cms.j2
```diff
@@ -0,0 +1,97 @@
...
+  #
+  # Send error response when request host isn't under our control
+  # We will no longer respond to proxy attempts like this with 
+  # anything.
+  # curl -i -A '' -x http://www.edx.org:80 --proxy-negotiate -U u:p -u u:p http://chat.sdtz.com
+  #
+
+  set $reject 'no';
+
+  if ($host !~* (edx.org|edxonline.org)$ ) {
```
e0d, line 18: This needs to be re-worked.  Set of acceptable hostnames should be a variable.

jrbl:
I think that the current version of this branch addresses most of your concerns, @jarv and @e0d, at least at a first pass. I'm anxious to get this landed so I can use the non-CMS parts on prod standup and rely on this base of CMS config. Permission to merge requested. :)

jarv:
ok to merge

jrbl:
thanks, merged.

PR #48, opened 2013-05-08
creates /opt/wwc/data

jbau:


playbooks/roles/lms/tasks/main.yml
```diff
@@ -18,6 +18,11 @@
 - include: ../../nginx/tasks/nginx_site.yml state=link site_name=lms
 - include: ../../nginx/tasks/nginx_site.yml state=link site_name=lms-backend
 
+- name: Change permissions on datadir
+  file: path={{ app_base_dir }}/data state=directory owner=www-data group=www-data
```
jarv, line 22: Can we set the permissions to what we want in "Create data dir" and remove this task?

jrbl, line 22: I don't think so, because I don't think that the www-data user exists yet
when Common runs.

jarv:
good to merge

jrbl:
I rebased on current master but didn't change anything, so should still be good to merge.

Shipit.

jrbl:
I re-rebased and tested, and added the cms tag to the directory permission update. Merging.

PR #49, opened 2013-05-09
Permissions issues found during standup

jrbl:
@sefk or @jbau this is the things I had written down when I left.  Plz pull. :)


sefk:
Ship it! (you can do the pull yourself @jrbl )

PR #50, opened 2013-05-09
fixing subnet association with ELBs

e0d:
I should have split these commits out more.
- changing web server port
- changing volume device mapping
- formatting and mounting in user data
- updating the ELBs subnet specifications.


playbooks/edxapp_stage.yml
```diff
@@ -1,4 +1,5 @@
-- hosts: tag_Group_edxapp_stage
+<<<<<<< Updated upstream
```
jarv: confict markup in commit

jarv:
other than the one comment this is good to merge

PR #52, opened 2013-05-09
Jbau/cms

jbau:
upstart/gunicorn and nginx setup for CMS


jbau:
this worked when installing on our instances

jrbl:
Shipit

PR #53, opened 2013-05-09
numerous improvements

e0d:
I should have split these commits out more.
- changing web server port
- changing volume device mapping
- formatting and mounting in user data
- updating the ELBs subnet specifications.
- updating cache commands and security groups.


jarv:
good to merge

PR #55, opened 2013-05-10
Configure rsyslog to catch logs from edx

jrbl:
- Touches directories and files into existence expected during setup for
  LMS and CMS
- Configures rsyslog to catch local0 and local1 from edx, and route into
  the appropriately named files in the appropriately named
  subdirectories
- Permissions updates to ensure rsyslog can write everywhere it's
  supposed to
- Restarts rsyslog after reconfiguration is complete

Tested on my rolling hosts, @jbau

### DIFF
--- a/playbooks/group_vars/all
+++ b/playbooks/group_vars/all
@@ -1,5 +1,6 @@
 ---
 app_base_dir: /opt/wwc
+log_base_dir: /mnt/logs
 venv_dir: /opt/edx
 
 # these pathes are relative to the playbook dir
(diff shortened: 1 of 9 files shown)


jrbl:
I should say, I tested that all the files are created correctly and hav ethe right perms on my rolling hosts, and it doesn't crash syslog.

I haven't tested that lms logs to it on my rolling hosts, because my rolling hosts don't have lms installed. But I did test with manual deployment on one of the prod hosts, and that machine is logging lms correctly with this. So I'm like 80% sure this will work.

jbau:
ok. this lgtm.

jbau:
@jrbl actually clicked merge

PR #57, opened 2013-05-11
Configure logrotate to treat edx logs appropriately

jrbl:
- Bugfix to rsyslog configuration; configuration file had missing
  extension
- Logrotate recipes for edx log files
- Files reorganized, renamed

@jbau look reasonable?

### DIFF
--- a/playbooks/roles/nginx/tasks/main.yml
+++ b/playbooks/roles/nginx/tasks/main.yml
@@ -29,8 +29,22 @@
   service: name=nginx state=started
   tags:
   - nginx
+
 - name: Create nginx log file location (just in case)
   file: path={{log_base_dir}}/nginx state=directory owner=syslog group=adm mode=2770
   tags:
   - nginx
   - logging
+
+# Commented out until default config has nginx log to {{log_base_dir}}/nginx
+# and also until default logrotate task 'nginx' gets removed
+###
+#- name: Set up nginx access log rotation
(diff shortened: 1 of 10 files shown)


jrbl:
This is the branch that has the bugfix for earlier, Jason.

playbooks/roles/nginx/tasks/main.yml
```diff
@@ -29,8 +29,22 @@
...
   tags:
   - nginx
+
 - name: Create nginx log file location (just in case)
   file: path={{log_base_dir}}/nginx state=directory owner=syslog group=adm mode=2770
   tags:
   - nginx
   - logging
+
+# Commented out until default config has nginx log to {{log_base_dir}}/nginx
```
e0d, line 39: Can we stash stuff like this rather than leaving in commented code?

e0d:
Other than the one comment, I think this is good to merge.

playbooks/roles/nginx/tasks/main.yml
jrbl, line 39: Yeah in general that's what I would prefer also. In this specific instance
I decided to leave it visible since I plan to be uncommenting that code
sometime this weekend.  I most emphatically don't want to start a trend
with this. :-)

jrbl:
Merged; the nginx configs that are commented out will be uncommented as of my next branch, which is almost ready.

PR #58, opened 2013-05-11
Documentation improvements, separation of phases, moving toward not usin...

e0d:
...g Ansible for cf

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,108 @@
 # Configuration Management
+
+## Introduction
+
+**This project is currently in alpha**
+
+The goal of the edx/configuration project is to provide a simple, but
+flexible, way for anyone to stand up an instance of the edX platform
+that is fully configured and ready-to-go.
+
+Building the platform takes place to two phases:
+
+* Infrastruce provisioning
+* Service configuration
+
+As much as possible, we have tried to keep a clean distinction between
+provisioning and configuration.  You are not obliged to use our tools
+and are free to use one, but not the other.  The provisioing phase 
+stands-up the required resources and tags them with role identifiers
+so that the configuration tool can come in and complete the job.
+
+The reference platform is provisioned using an Amazon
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,10 +1,108 @@
...
+tunneled.
+
+Add something like the following to your `~/.ssh/config` file.
+
+```
+Host *.us-west-1.compute-internal
+  ProxyCommand ssh -W %h:%p vpc-00000000-jumpbox
+  IdentityFile /path/to/aws/key.pem
+  ForwardAgent yes
+  User ubuntu
```
sefk, line 70: My preference is to not set User or IdentityFile in the ssh config.  We want to do machine bringup as the ubuntu user, but for subsequent operations (installs) we want do do those as individual users.  So I've gone away from those two lines.

README.md
```diff
@@ -1,10 +1,108 @@
...
+
+* The [AWS console](https://console.aws.amazon.com/cloudformation/home)
+* The AWS [CloudFormation CLI](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-installing-cli.html)
+* Via Ansible
+
+If you don't have experience with CloudFormation, the web console is a
+good place to start because it will use a form wizard to gather
+configuration parameters, it will give you continuous feedback during
+the process of building the stack and useful error messages when
+problems occur.
```
sefk, line 52: :+1: for using the UI for driving cloud formation.  It's not something we do often.  Most people will just want to do it once, twice if they're setting up a staging environment.  And if you do via the UI you get nice prompts.

sefk:
Ship it.  Everything in here is ok.  We can continue to improve and we can make that change to the ~/.ssh/config I recommend later if you agree @e0d

README.md
e0d, line 70: Agree, but this was moved from elsewhere in the doc.  We should probably have an initial user config ansible run to set up local users and perform subsequent config with those identities.

PR #59, opened 2013-05-11
Nginx log reconfiguration

jrbl:
- sends logs to /mnt/logs/nginx instead of /var/log. Note that on fresh
  server startup, nginx will still log to /var/log until it reads its
  configuration.
- Installs logrotate configuration for freshly moved nginx logs
- Some directory and file permissions fixups
- Improved order of operations on nginx installation/bringup

There is a corresponding change in stanford's local repo in the nginx templates.

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -56,7 +56,7 @@
 
 - name: Set permissions on edx log file
   # This is done for the benefit of the rake commands, which expect it
-  file: path={{log_base_dir}}/edx.log owner=syslog group=adm mode=770
+  file: path={{log_base_dir}}/edx.log owner=syslog group=adm mode=640
   tags:
   - pre_install
   - logging
--- a/playbooks/roles/common/tasks/edx_logging_base.yml
+++ b/playbooks/roles/common/tasks/edx_logging_base.yml
@@ -16,7 +16,7 @@
   - logging
 
 - name: Set permissions on tracking file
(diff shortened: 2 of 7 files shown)


jrbl:
I already pulled the change in the other repo.

jrbl:
Oh, and I ran this on only host 238. I couldn't see anything wrong when I checked it, so I think it's safe to run on all our prod hosts.

playbooks/roles/common/tasks/main.yml
```diff
@@ -56,7 +56,7 @@
 
 - name: Set permissions on edx log file
   # This is done for the benefit of the rake commands, which expect it
-  file: path={{log_base_dir}}/edx.log owner=syslog group=adm mode=770
+  file: path={{log_base_dir}}/edx.log owner=syslog group=adm mode=640
   tags:
```
jarv, line 60: Is this a case where these two files {tracking,edx}.log were already in place with the wrong permissions?
We should be able to let rsyslog write these with the correct permissions when the application starts as long as `{{ log_base_dir }}` is writable by syslog.

jrbl, line 60: I'm not sure how the permissions get specified by rsyslog, but they were
wrong on the host I was testing with for tracking. So I'm not sure what's
going on there.

There's a play to touch edx.log into existence entirely because the mitx
rake target gather_assets fails if it's not. This particular location is
unused by everything in the world, including the rakefile, afaict.

jrbl:
Merging; I think everyone realizes the behavior of rake is unfortunate and should get fixed, and when it does no log files should get touched into place any more.

PR #62, opened 2013-05-14
Use the new mitx repo locale

jrbl:
This is right, right?


sefk:
LGTM @jrbl .  Ship it!

PR #64, opened 2013-05-14
Jbau upstart restart handler

jbau:
We were missing this service handler before.  It's not the prettiest, because it hardcodes service names "lms" "cms" and "lms-preview", but those are role names anyway.


jarv:
looks good

PR #65, opened 2013-05-14
Feature/jrbl/tagging update install

jrbl:
@sefk gets annoyed not knowing which things in the tag tree should be run and which shouldn't, so we created the update tag so he can simply run all the update tasks every time, safely and quickly.

@jarv should look this over and see if I screwed it up.


jrbl:
Now with no merge conflicts!

jbau:
I think this looks fine, only tags are changed.

jbau:
@jrbl recommending merge

PR #66, opened 2013-05-14
Jarv/cleanup lms and examples

jarv:
This is an attempt to cleanup the example playbooks and the example secure_dir, we may want to discuss it some before merging.

With Jason's last pull for hash merging I think it makes more sense to use empty hashes for the env and auth vars associated with roles and put the entire hash into the secure_example/vars/ so all of the merging in contained in one yaml file.


jarv:
@jbau can you look over this?

playbooks/roles/cms/vars/main.yml
```diff
@@ -1,35 +1,5 @@
 # variables common to the cms role, automatically loaded
 # when the role is included
 ---
-cms_auth_config:
```
jbau, line 4: I think do like moving these out.  Having vars here _seems_ useful for defaults, but ansible's variable precedence rules for variables defined in "playbook-like" environments (vars in playbooks files, var_files in playbook files, vars directories in roles, etc) are sufficiently murky that it becomes a bit confusing what becomes defaults and what are overrides.   As I read it, the third paragraph http://ansible.cc/docs/playbooks2.html#understanding-variable-precedence pretty much puts all of these in the same tier, with some not-fully-intuitive ordering imposed, so it sort of makes me a bit shy about hash merging between say a var_file and a vars/main.yml .

That's a long winded way of saying I think it's best to cut these example defaults out, though I wish we didn't have to.

playbooks/secure_example/vars/edxapp_ref_vars.yml
```diff
@@ -2,6 +2,11 @@
 #Use YAML references (& and *) and hash merge <<: to factor out shared settings
 #see http://atechie.net/2009/07/merging-hashes-in-yaml-conf-files/
 lms_auth_config:  &lms_auth
+  'CONTENTSTORE':
```
jbau: I think `CONTENTSTORE` and `MODULESTORE` were already in the secure example and got repeated

playbooks/secure_example/vars/edxapp_ref_vars.yml
```diff
@@ -42,6 +47,47 @@ lms_auth_config:  &lms_auth
     'url':  'hidden-prod'
  
 lms_env_config:  &lms_env
+  'CACHES':
```
jbau: ditto `CACHES`

playbooks/roles/cms/vars/main.yml
jbau, line 4: The duplications below are making me think a tiny bit:  We had two dimensions of default->override behavior.  One for non-secret -> secret and one for shared_env_btw_service_variant -> customized_env_for_service_variant.  The one use case for having these files would be for the non-secret settings for each service, which could then be overridden by the secrets specified via var_files (it works, happily, perhaps by chance).  But maybe keeping both dimensions is just too confusing ultimately?

playbooks/edxapp_ref.yml
```diff
@@ -1,9 +1,8 @@
-- hosts: tag_Group_edxapp_prod
```
jbau, line 1: I think it's fine to consolidate to edxapp_ref.yml, but we'll probably checkin our prod playbook again.
Mainly because we want to get to a state where we can use the unmodified checkout to admin our instances.
We found a way using tag assignments in the AWS console and boolean operations on hosts to make that possible.  I.e. something like

```
- hosts: tag_environment_prod:&tag_function_webserver
```

playbooks/roles/cms/vars/main.yml
jarv, line 4: Yeah I thought that hash merging would be a good way to avoid data duplication but the closer i look it seems like by far the majority of our config is environment specific (secrets, hostnames, etc) and the rest is boilerplate that will probably not change.  I think what it comes down to is being able to grok the entire lms config hash without opening up multiple yaml files and understanding the precedence or how the variable files are ordered in the playbooks.

playbooks/edxapp_ref.yml
jarv, line 1: yeah that's exactly what I'm doing now and I wanted to talk at some point about managing these playbooks. 
right now I'm keeping mine as untracked files, dumping them all in playbooks/ doesn't seem like the right thing to do either.

jbau, line 1: yeah maybe we can just suggest a tagging convention, or add them to the CF scripts.

jbau:
ok. I think this looks good to merge now that you've fixed the hashes

jarv:
I submitted an issue so we can figure out how to manage the playbooks/ dir
https://github.com/edx/configuration/issues/67

PR #70, opened 2013-05-15
setting auto-scaling min/max to 2

jarv:
Until we support autoscaling we should keep min/max
the same.  We might want to make it 1/1.


jbau:
yeah, we had to do this manually to 3/3, so I am in favor of this

PR #75, opened 2013-05-16
fill-in previously missing tasks that create /opt/wwc/staticfiles

jbau:
@jarv @jrbl take a look.  this was a gap in our ansible scripts which I discovered on the fabric run.

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -3,19 +3,25 @@
 
 - name: Create application root
   # In the future consider making group edx r/t adm
-  file: path=$app_base_dir state=directory owner=root group=adm mode=2775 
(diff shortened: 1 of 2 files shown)


jarv:
looks good

playbooks/roles/common/tasks/main.yml
```diff
@@ -20,6 +20,12 @@
   - pre_install
   - update
   
+- name: Create staticfiles dir
+  file: path={{ app_base_dir }}/staticfiles state=directory owner=root group=adm mode=2775
```
jrbl, line 24: If you're going to later chgrp to www-data, this can be just 0775; but if we want the sgid behavior, the below should be www-data:adm.

jrbl:
Nit about mismatched perms; after fix, shipit.

PR #76, opened 2013-05-16
Vik/ml api v1

VikParuchuri:
- Add in support for provisioning ml api on sandbox/stage/prod.
- Add in discern role.


playbooks/mlapi_prod.yml
```diff
@@ -0,0 +1,20 @@
+- hosts:
+    - tag_Group_mlapi_prod
+  vars_files:
+    - ["{{ secure_dir }}/vars/mlapi_prod_vars.yml"]
+    - ["{{ secure_dir }}/vars/users.yml"]
+    - ["{{ secure_dir }}/vars/mlapi_prod_users.yml"]
```
jarv: fyi - when a list is used for a var_files element it will pick the first one it finds in the list, in this case the list isn't necessary since there is only one item but it doesn't matter.

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,112 @@
...
+    - celery
+    - nginx
+    - discern
+
+#Upload config files for django (auth and env)
+- name: create discern application config env.json file
+  template: src=env.json.j2 dest=$app_base_dir/env.json owner={{remote_user}} group=edx mode=0640
+
+- name: create discern auth file auth.json
+  template: src=auth.json.j2 dest=$app_base_dir/auth.json owner={{remote_user}} group=edx mode=0640
```
jarv: nitpick: can use `{{app_base_dir}}` for consistency

VikParuchuri: Sure.  Just as a note, I had to use the $ notation sometimes because the {{ notation files on line starts (ie right after a - )

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,112 @@
...
+
+#Needed if using redis to prevent memory issues
+- name: change memory commit settings -- needed for redis
+  command: sysctl vm.overcommit_memory=1
+
+#Needed to pull repos from github, due to hanging on host add prompt otherwise
+- name: add github to the known hosts file
+  shell: ssh-keyscan -H github.com > ~/.ssh/known_hosts
```
jarv: I would prefer not rely on settings in ~/.ssh if possible, we have been copying the private key to the server and creating a git.sh script

```
#!/bin/sh
exec ssh -i "/path/to/git_identity" -o "StrictHostKeyChecking no" "$@"
```

Then when calling git add this to your task:

```
environment:
  GIT_SSH: /path/to/git_ssh.sh
```

If key forwarding does work for what you are doing you can just include the `-o "StrictHostKeyChecking no"` in your git.sh script and use the GIT_SSH env.

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,112 @@
...
+  pip: requirements="{{item}}/pre-requirements.txt" virtualenv="{{venv_dir}}" state=present
+  with_items:
+    - ${discern_dir}
+    - ${ease_dir}
+
+- name : install python requirements for discern and ease
+  pip: requirements="{{item}}/requirements.txt" virtualenv="{{venv_dir}}" state=present
+  with_items:
+    - ${discern_dir}
+    - ${ease_dir}
```
jarv, line 96: in ~~mitx~~ edx-platform we decided to organize the requirements this way:

```
requirements/<virtaulenv-name>/base.txt  <--- for everything includuing numpy
requirements/<virtualenv-name>/post.txt  <-- for scipy which needs to be installed after numpy
```

see: https://github.com/edx/edx-platform/tree/master/requirements

playbooks/roles/discern/vars/main.yml
```diff
@@ -0,0 +1,10 @@
+discern_source_repo: git@github.com:edx/discern.git
+ease_source_repo: git@github.com:edx/ease.git
+ease_dir: $app_base_dir/ease
+discern_dir: $app_base_dir/discern
+discern_settings: discern.aws
+nltk_data_dir: /usr/share/nltk_data
+remote_user: ubuntu
```
jarv: If all tasks are running as sudo why do we need remote_user?

I think in general we don't want any files owned by ubuntu (since he has full sudo) and operationally we don't want to depend on that user being around.  It's good to set permissions on our application files so that they are not owned by root, maybe a new system user like "machine-learning"?

playbooks/roles/discern/templates/discern.conf.j2
```diff
@@ -0,0 +1,21 @@
...
+author "Vik Paruchuri <vik@edx.org>"
+
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+respawn
+respawn limit 3 30
+
+env PID=/var/run/gunicorn/discern.pid
+env WORKERS=4
```
jarv: I think you can do this:

```
env WORKERS={{ ansible_processor_cores * 2 }}
```

to scale the number of workers by the number the cores.

playbooks/roles/discern/templates/discern.conf.j2
```diff
@@ -0,0 +1,21 @@
...
+env PORT=7999
```
jarv: can you add this to `discern/vars/main.yml`?

playbooks/roles/discern/templates/sudoers.j2
```diff
@@ -0,0 +1,30 @@
+#
```
jarv: If we have custom sudo config can it be added to `sudoers.d/`?

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,112 @@
...
+  template: src=celery.conf.j2 dest=/etc/init/celery.conf
+
+- name: render discern service from template
+  template: src=discern.conf.j2 dest=/etc/init/discern.conf
+
+- name: create nginx directory and set perms
+  file: path=/etc/nginx/sites-available owner=root group=edx mode=2775 state=directory
+
+- name: render nginx sites available
+  template: src=default.j2 dest=/etc/nginx/sites-available/default
```
jarv: Generally the convention is to copy a custom config into sites-available and then link to it from sites-enabled.
What I think you should do here is
- rm the symlink /etc/nginx/sites-enabled/default to remove the default config (assuming you don't want it)
- create `/etc/nginx/sites-available/machine-learning` or a config with a similar name that has what you want in it
- link `/etc/nginx/sites-enabled/machine-learning` to `/etc/nginx/sites-available/machine-learning` to activate it

playbooks/roles/discern/vars/main.yml
```diff
@@ -0,0 +1,10 @@
...
+ease_branch: master
+discern_branch: dev
+nginx_listen_port: 80
```
jarv: I think we should have a `venv_dir` variable defined here that is in the machine learning namespace like `ml_venv_dir`.

playbooks/roles/discern/templates/sudoers.j2
VikParuchuri: This config is to enable ssh key forwarding as root.  I'll look into sudoers.d .  Never used it before.

playbooks/roles/discern/vars/main.yml
VikParuchuri: Any reason not to use /opt/edx as the venv?  It keeps it compatible with common to do it this way.

playbooks/roles/discern/tasks/main.yml
VikParuchuri, line 96: Changing it to mirror this would entail other changes to the repos, docs, etc, so going to hold off on this for now.

VikParuchuri:
Have made fixes.  Testing now.

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
+---
+#Allows us to recover from a bad sudoers file
+- name: Install policykit
+  apt: pkg=policykit-1 install_recommends=yes state=present update_cache=yes
+
+#Discern user is admin
+- name: Create discern user
+  user: name={{ discern_user }} append=yes groups={{ "adm,edx" }} shell=/bin/bash
```
jarv: should this be `groups="adm,edx"`?

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+
+- name: upload sudoers template to /tmp/sudoers
+  copy: src=sudoers-discern dest=/tmp/{{site_name}} owner=root group=root mode=0440
+
+#Verify file and move to sudoers.d folder
+- name: move temp file to sudoers.d
+  shell: visudo -q -c -f /tmp/{{site_name}} && cp /tmp/{{site_name}} /etc/sudoers.d/{{site_name}}
+
```
jarv, line 25: good idea, we should do this with all of our sudo config changes.

playbooks/roles/discern/tasks/main.yml
```diff
@@ -0,0 +1,133 @@
...
+- name: Creating nginx config link {{ site_name }}
+  file: src=/etc/nginx/sites-available/{{ site_name }} dest=/etc/nginx/sites-enabled/{{ site_name }} state=link owner=root group=root
+
+- name: start all services
+  service: name={{item}} state=started
+  with_items:
+    - celery
+    - nginx
+    - discern
```
jarv: If the nginx configuration, upstart file, or discern files change you will probably want to restart the corresponding service.  With `state=started` once the the service is running it will remain running even if changes are made to the configs.

you could do a state=restarted to restart them every time the playbook is run but ideally what you would do is create a handler for each service and then notify the corresponding handler from the tasks that require service restarts.

example:

```

- name: Removing default nginx config
  file: path=/etc/nginx/sites-enabled/default state=absent
  notify: restart nginx

(add a notify for every action that affects nginx)

then in role/discern/handlers/main.yml:
- name: restart nginx
  action: service: name nginx state=restarted

```

playbooks/roles/discern/templates/nginx-discern.j2
```diff
@@ -0,0 +1,93 @@
...
+#       #}
+
+#       # deny access to .htaccess files, if Apache's document root
+#       # concurs with nginx's one
+#       #
+#       #location ~ /\.ht {
+#       #       deny all;
+#       #}
+# }
+
```
jarv: why is this section commented out, can it be removed?

jarv:
looks good to merge

VikParuchuri:
Tested, appears to work fine.

PR #77, opened 2013-05-16
set correct permissions for htpasswd

jbau:
@jrbl


jrbl:
shipit

jrbl:
@jbau said I could hit the button.

PR #81, opened 2013-05-16
readme: manual post-cloudformation commands

sefk:
I had taken a screenshot at some point with the intention to add to the doc, with nicely redacted private info.  Figured I'd better add before I deleted it by mistake.  Added here.


jbau:
I think you can just pull this, since it's just readme

sefk:
Yes, thanks @jbau

PR #85, opened 2013-05-19
subdir with symlinks for organizing playbooks

jarv:
This puts edx-east playbooks in 
`configuration/playbooks/edx-east`
with symlinks to roles, group_vars, files, ec2.py, secure_example


jrbl:
lgtm.

PR #86, opened 2013-05-19
consolidating lms/cms/lms-preview into edxapp role

jarv:
Creates edxapp upstart job and combines lms, lms-preview. cms
into the edxapp role.


jbau:
cool.  let me give this a look over the next day or so.

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -90,43 +73,6 @@
...
-  - install
-
-- name: store remote apt_packages list for ansible use
-  command: cat {{ apt_packages_file }}
-  register: apt_packages_list
-  tags:
-  - lms
-  - cms
-  - install
-
```
jarv, line 129: I was having trouble installing npm after enabling these repos. We don't have them enabled in our AWS environment so I don't think they are necessary (these are for local dev envs I assume).  Thoughts?

jbau, line 129: Perhaps @cpennington can weigh in, but I'd be fine removing them.

playbooks/roles/edxapp/vars/main.yml
```diff
@@ -1,14 +1,25 @@
...
-apt_sources_file:  "{{ platform_code_dir }}/requirements/system/{{os_name}}/apt-repos.txt"
-apt_packages_file: "{{ platform_code_dir }}/requirements/system/{{os_name}}/apt-packages.txt"
+# These vars are for creating the application json config
+# files.  There are two for each service that uses the
+# 'edx-platform' code.  Defining them will create the upstart
+# job and nginx configuration for the corresponding service.
+# It will also enable the corresponding section in the 
+# 'edxapp' upstart job.
+
+lms_auth_config: false
```
jbau: Here, we're relying on http://ansible.cc/docs/playbooks2.html#understanding-variable-precedence saying that playbook var_files settings have precedence over 'variables defined in vars sections', under which I'm guessing vars/main.yml falls.  Sort of wish the documentation was more explicit about variables in the vars directory of each role.

playbooks/roles/gunicorn/templates/lms-xml.conf.j2
```diff
@@ -0,0 +1,37 @@
...
+author "Calen Pennington <cpennington@mitx.mit.edu>"
+
+start on started edxapp
+stop on stopped edxapp
+
+respawn
+respawn limit 3 30
+
+env PID=/var/tmp/lms-xml.pid
+env NEW_RELIC_CONFIG_FILE=/opt/wwc/newrelic.ini
```
jbau: We had commented out the NEW RELIC stanzas

jbau: Also, we used the following vars to parameterize here, if you're interested:
`{{app_base_dir}}`
`{{venv_dir}}`
`{{ ansible_processor_cores }}`
`{{platform_code_dir}}`

playbooks/roles/nginx/templates/lms-xml.j2
```diff
@@ -0,0 +1,99 @@
+server {
+  # LMS-preview configuration file for nginx, templated by ansible
```
jbau, line 2: I think this is supposed to be lms-xml?  Not sure there's a huge difference, but pointing it out

playbooks/roles/edxapp/tasks/main.yml
jrbl, line 129: Well, we're doing these repos to get the ppa version of node, which may be the only package we're installing from this set of PPAs (I'm not sure). Now, if you're not running the ppa version of node in your prod, clearly it's not important that we do either. But it makes me skittish that something is documented as a 'requirement' that's actually not.

If you want to pull it out, go ahead - we can stick it back in on our side if we need it. I have a mild preference for seeing this turned into a positive change rather than a negative one. That is, if you're going to say these repos aren't necessary, let's actually have a play that asserts that they be absent rather than simply not adding them.  Then sometime in the future we can drop those plays.

jbau, line 129: also, just briefly, if we're removing these, npm itself should be removed from edxapp/vars/main.yml/lms_debian_pkgs, otherwise the install fails on a machine I tried.

jarv, line 129: I added npm to lms_debian_pkgs in this branch, the reason why I was seeing failures were due to these extra repos.
Since this won't remove them if they were added previously I had to do something like

```
rm -f /etc/apt/sources.d/*
apt get update
```

After that the npm system package installs fine.

playbooks/roles/edxapp/vars/main.yml
jarv: Right, I'm think now we should create a new set of variables 

```
lms_service_variant_enabled:  false
lms_xml_service_variant_enabled: false
cms_service_variant_enabled: false
lms_preview_service_variant_enabled: false
```

And change make env and auth hashes default to an empty hash again.
This way we can put examples in secure_example directory without enabling the services.

playbooks/roles/edxapp/tasks/main.yml
jbau, line 129: oic.  let me try that.

jbau, line 129: yup. the npm issue is fixed by removing the apt-sources

playbooks/roles/edxapp/vars/main.yml
jbau: ok. i'm taking a crack at this now.  let me know if i should stop

jbau:
Okay. I added the "toggle" variables actually as a member of a service_variants_enabled list.  Seems to run ok on my test env.

I'm satisfied with testing on the PR, so merge when ready, for me.

PR #89, opened 2013-05-22
Clustering config for rabbitmq

jbau:
@VikParuchuri these are the changes in configuration I did to enable a 2-node cluster.  See the `enable_ha` var

Writing out the configuration files works, but I think is restarting/starting rabbitmq is pretty brittle, so I'm actually not sure that ansible is the right tool for that.  Maybe @jarv can comment.

### DIFF
--- a/playbooks/ec2.ini
+++ b/playbooks/ec2.ini
@@ -1,6 +1,6 @@
 [ec2]
 regions=all
 destination_variable=public_dns_name
-vpc_destination_variable=private_ip_address
+vpc_destination_variable=private_dns_name
 cache_path=/tmp
 cache_max_age=300
--- a/playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
+++ b/playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
@@ -1,18 +1,20 @@
 % This file managed by ansible
 % keep erlang kernel communcations to a range
 [
+ {% if enable_ha %}
  {kernel,
   [{inet_dist_listen_min, 65500},
    {inet_dist_listen_max, 65510}
   ]
  },
+ {% endif %}
  {rabbit, [
     {% if enable_ha %}
        {cluster_nodes, {[
          {% for svr in groups[ha_group_tag] %}
            'rabbit@{{hostvars[svr]['ansible_hostname']}}'{% if loop.index != loop.length %},{% endif %}
          {% endfor %}
-       ], 'disk'}}
+       ], 'disc'}}
     {% endif %} 
  ]},
   % Configure the Stomp Plugin listening port
--- a/playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
+++ b/playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
@@ -1,0 +1,20 @@
+% This file managed by ansible
+% keep erlang kernel communcations to a range
+[
+ {kernel,
+  [{inet_dist_listen_min, 65500},
+   {inet_dist_listen_max, 65510}
+  ]
+ },
+ {rabbit, [
+    {% if enable_ha %}
+       {cluster_nodes, {[
+         {% for svr in groups[ha_group_tag] %}
+           'rabbit@{{hostvars[svr]['ansible_hostname']}}'{% if loop.index != loop.length %},{% endif %}
+         {% endfor %}
+       ], 'disk'}}
+    {% endif %} 
+ ]},
+  % Configure the Stomp Plugin listening port
+ {rabbitmq_stomp, [{tcp_listeners, [${rabbitmq_stomp_port}]} ]}
+].
--- a/playbooks/roles/rabbitmq/vars/main.yml
+++ b/playbooks/roles/rabbitmq/vars/main.yml
@@ -1,0 +1,5 @@
+ha_group_tag: tag_Group_mlapi-rabbitmq_prod
+enable_ha: true
+rabbitmq_node_port: 5672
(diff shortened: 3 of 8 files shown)


playbooks/ec2.ini
```diff
@@ -1,6 +1,6 @@
 [ec2]
 regions=all
 destination_variable=public_dns_name
-vpc_destination_variable=private_ip_address
+vpc_destination_variable=private_dns_name
```
jbau, line 4: I think we need the setting to `private_dns_name` for our .ssh/config to work within a VPC.  But I got the clustering to work this way.

What was the reason you changed it originally?

playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
```diff
@@ -0,0 +1,22 @@
+% This file managed by ansible
+% keep erlang kernel communcations to a range
+[
+ {% if enable_ha %}
+ {kernel,
+  [{inet_dist_listen_min, 65500},
```
jbau, line 6: need to make sure these ports are open in VPC for the security group.
also need 4369 to be open for Erlang Port Mapper Daemon, for cluster formation

playbooks/roles/rabbitmq/vars/main.yml
```diff
@@ -0,0 +1,5 @@
+ha_group_tag: tag_Group_mlapi-rabbitmq_prod
+enable_ha: true
```
jbau, line 2: toggles clustering

playbooks/roles/rabbitmq/vars/main.yml
```diff
@@ -0,0 +1,5 @@
+ha_group_tag: tag_Group_mlapi-rabbitmq_prod
```
jbau, line 1: this tag controls which hostnames are included in the cluster.  all hosts that match this tag get included.

playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
```diff
@@ -0,0 +1,22 @@
...
+   {inet_dist_listen_max, 65510}
+  ]
+ },
+ {% endif %}
+ {rabbit, [
+    {% if enable_ha %}
+       {cluster_nodes, {[
+         {% for svr in groups[ha_group_tag] %}
```
jbau, line 14: see significance of `ha_group_tag`

jbau:
@jarv do you want to take a look too?

playbooks/ec2.ini
VikParuchuri, line 4: I couldn't get it to work the other way, so I changed it to private ip, even after enabling vpc hostname resolution.  This was one of the things I wanted to cleanup before making a PR to master.

VikParuchuri:
@jbau Do you want to use set-policy to automatically define mirroring/syncing?  See http://www.rabbitmq.com/ha.html .

playbooks/roles/rabbitmq/vars/main.yml
VikParuchuri, line 1: Is it necessary to have this, or can we just assume that all hosts defined in the playbook should be a part of the HA group?

playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
```diff
@@ -0,0 +1,22 @@
...
+           'rabbit@{{hostvars[svr]['ansible_hostname']}}'{% if loop.index != loop.length %},{% endif %}
```
VikParuchuri, line 15: You would have to enable hostname resolution in the VPC for this to work, right?  

Is it preferable to assume that it will be enabled, tell people loudly to enable it, or to add to /etc/hosts to map the hostnames to ip addresses?  Or maybe a fourth option I am ignoring.

playbooks/roles/rabbitmq/vars/main.yml
jbau, line 1: I just didn't know the ansible magic to get all hosts in the playbook into a runtime var, though I looked around for it to start.  Be happy to switch it if found.

playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
jbau, line 15: I'm pretty sure our ssh proxy config requires name resolution to work.  Also, we do provide output for people do to this at then end of the CF run:  
https://github.com/edx/configuration/blob/master/cloudformation_templates/edx-reference-architecture.json#L3212
So I think we can assume it, or at least mention it in a readme.

playbooks/ec2.ini
jbau, line 4: Did the IPs work for you?  I had to open up ports to get it to work for hostnames.

e0d, line 4: Either should work just fine, but ports will need to be open in either case.  I don't much like using the internal names, but the IPs are probably worse in this case as they are subject to change on reboot.  I think we need to think about using either a DNS convention here or elastic ips.

playbooks/roles/rabbitmq/vars/main.yml
jarv, line 1: `groups['tag_Group_mlapi-rabbitmq_prod']` will give you the internal dns names of all hosts in the group. That will however match all hosts that have that tag which may be more than we want if there are multiple clusters in a single account.  I think we need a custom inventory script that limits the results to a single vpc to make this work.

jbau:
this is ancient and unused. closing

PR #90, opened 2013-05-22
sets --distribute flag in virtualenv creation

jbau:
and conditionalizes nginx site creation in edxapp role

@jrbl also @jarv might want to take a look at the conditionalization.

### DIFF
--- a/playbooks/roles/nginx/tasks/main.yml
+++ b/playbooks/roles/nginx/tasks/main.yml
@@ -60,4 +60,11 @@
   tags:
   - nginx
   - update
+  
+- name: Register the fact that nginx has run
+  command: echo True
+  register: nginx_role_run
+  tags:
+  - nginx
+  - update
 
--- a/playbooks/roles/common/tasks/create_venv.yml
+++ b/playbooks/roles/common/tasks/create_venv.yml
@@ -1,9 +1,12 @@
 ---
 # create the 'edx' virtual environment in /opt so that roles can populate it
-- name: easy_install pip
-  easy_install: name=pip
+- name: Install python and pip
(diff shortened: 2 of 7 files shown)


playbooks/roles/nginx/tasks/main.yml
```diff
@@ -60,4 +60,11 @@
   tags:
   - nginx
   - update
+  
+- name: Register the fact that nginx has run
+  command: echo True
+  register: nginx_role_run
+  tags:
+  - nginx
+  - update
```
jarv, line 69: Is the idea here that we might include the edxapp role without including the nginx role?

jbau, line 69: exactly.  I just had to do that to stand up "util" celery worker machines that will eventually send bulk emails.

I think they will need a different upstart script that starts celeryd rather than gunicorn and  they will certainly need the auth/env.json, which are part of the edxapp role.

jbau, line 69: one way to do this (probably more ansible friendly) is to factor things out, but this is more expedient…

jrbl:
lgtm. Shipit.

playbooks/roles/nginx/tasks/main.yml
jarv, line 69: hm, if this a common pattern it might be cleaner to move the `when_set: $nginx_role_run` to the tasks in `nginx_site.yml`, either way good to merge

jbau, line 69: you're right.  Oversight.  I'll move it over.

jbau:
moved the conditionals and tested again. seems to work so merging.

PR #96, opened 2013-06-03
removing the ec2_elb library, now in 1.2 core

jarv:


jtauber:
if @mpdehaan says to do it, sounds good to me :-)

PR #99, opened 2013-06-05
feanil/pre requirements

feanil:
I updated the edxapp playbook to add a task for the pre requirements file so that numpy actually gets installed before everything else.


e0d:
+1

PR #103, opened 2013-06-11
Feanil/ansible deploy

feanil:
Work to let ansible fully deploy the edxapp servers.


playbooks/edx-east/feanil_deploy.yml
```diff
@@ -0,0 +1,12 @@
+- hosts: tag_aws_cloudformation_stack-name_feanilpractice:&tag_group_edxapp
```
e0d, line 1: Do we want to commit stuff like this?

e0d:
Looks good to me.

playbooks/edx-east/feanil_deploy.yml
jarv, line 1: I'm not sure what playbooks we want here, this edx-east dir with symlinks was a hack so that we can organize things better but I'm not married to it.  Maybe do something similar in our secure repo and make that the playbook repository for edx?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -0,0 +1,165 @@
...
+  tags:
+  - lms-preview
+  - deploy
+
+- name: stop cms service
+  service: name=cms state=stopped
+  tags:
+  - cms
+  - deploy
+
```
jarv, line 10: should we use the edxapp service here instead? I'm not the behavior of the edxapp upstart job when you pull the rug under it by stopping all the child services.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -0,0 +1,165 @@
...
+# TODO: Move it here.
+
+# Do Post Checkout Tasks.
+
+# This checks needs to be run to see if rake can be used but its failure should not stop the run.
+- name: check if rake gather_assets is available
+  shell: executable=/bin/bash chdir={{ platform_code_dir }} rake -T | grep --quiet gather_assets
+  environment: "{{ deploy_environment }}"
+  register: grep_gather_assets 
+  ignore_errors: yes
```
jarv: If we omit the `--quiet` flag rake won't give us the broken pipe error.

playbooks/roles/edxapp/tasks/deploy.yml
feanil, line 10: I wasn't sure what would be the better option since I didn't want to have to restart all the applications just because something changed for one.  But now that I think about it more, since they are all based on the same repository, they will probably all need to be restarted together.  I'll pull out the 4 handlers and put in one for edxapp.

jbau, line 10: Yeah, edxapp is the way to go.  The separate child upstart scripts are a legacy from before.

playbooks/roles/edxapp/handlers/main.yml
```diff
@@ -0,0 +1,208 @@
...
+  tags:
+  - lms
+  - lms-xml
+  - lms-preview
+  - cms
+  - deploy
+
+# Ruby plays that need to be run after platform updates.
+- name: gem | gem install bundler
+  shell: RBENV_ROOT=${rbenv_root} GEM_HOME=${gem_home} ${rbenv_root}/shims/gem install bundle chdir=${platform_code_dir}
```
jarv: can we use jinja syntax here?

playbooks/roles/edxapp/handlers/main.yml
```diff
@@ -0,0 +1,208 @@
...
+  tags:
+  - ruby
+  - deploy
+  - install
+
+- name: bundle | bundle install
+  shell: RBENV_ROOT=${rbenv_root} GEM_HOME=${gem_home} ${gem_home}/bin/bundle install --binstubs chdir=${platform_code_dir}
```
jarv: same comment

jarv:
in git.py we are using _run_if_changed() to only run some tasks if the source files have changed. I don't think we need to do that now but we should plan to eventually to save deployment time.  Maybe add a TODO?

jarv:
looks good

PR #104, opened 2013-06-11
nginx logrotate fixup

jrbl:
- Set postrotate script to SIGHUP nginx after rotation
- Explicitly set permissions on freshly created logfiles


jarv:
looks good

jrbl:
I've already run this on our prod boxes.

PR #106, opened 2013-06-14
Jbau/new edx theme

jbau:
@jarv @feanil 

This does 3 things basically:
- creates edx-west playbook symlinks
- adds support for installing theming
- refactors some of the deployment handlers (since restart edxapp doesn't work)

### DIFF
--- a/playbooks/roles/edxapp/handlers/main.yml
+++ b/playbooks/roles/edxapp/handlers/main.yml
@@ -1,13 +1,33 @@
 ---
-# Restart Start all services.
-- name: restart edxapp
-  service: name=edxapp state=restarted
+- name: invoke all deploy handlers
+  shell: echo "running deploy handlers"
+  notify:
+  - gem | gem install bundler
+  - bundle | bundle install
+  - Install edx-platform npm dependencies
+  - install python pre-requirements
+  - install python base-requirements
+  - install python post-requirements
+  - install python post-post requirements
+  - check if rake gather_assets is available
+  - check if django can collect lms static data
+  - check if django can collect cms static data
+  - check if django can update cms templates
+  - gather lms static assets with rake
+  - gather lms static assets with django
+  - gather cms static assets with rake
+  - gather cms static assets with django
+  - update cms templates
+  - lms asset static failure check
+  - cms asset static failure check
+  - stop edxapp
+  - start edxapp
   tags:
   - lms
-  - lms-xml
-  - lms-preview
   - cms
+  - install
   - deploy
+
 
 # Ruby plays that need to be run after platform updates.
 - name: gem | gem install bundler
@@ -30,6 +50,7 @@
   tags:
   - npm
   - update
+  - deploy
 
 # Python plays that need to be run after platform updates.
 
@@ -40,6 +61,7 @@
   - lms
   - cms
   - install
+  - deploy
 
 # Install the python modules into {{ venv_dir }}
 - name : install python base-requirements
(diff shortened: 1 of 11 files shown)


playbooks/edx-west/ec2.ini
```diff
@@ -0,0 +1 @@
+../ec2.ini
```
jbau: all of these `../` things are symlinks

playbooks/roles/edxapp/handlers/main.yml
```diff
@@ -206,3 +229,23 @@
   tags:
   - cms
   - deploy
+
+- name: restart edxapp
```
jbau: we want to restart edxapp _after_ all of the other actions in this file run

jbau: also, we've noticed that `sudo restart edxapp` doesn't work, so split it into 2 actions, stop and start.

playbooks/roles/edxapp/vars/main.yml
```diff
@@ -24,10 +24,16 @@ cms_env_config: {}
 lms_xml_auth_config: {}
 lms_xml_env_config: {}
 
-lms_source_repo: git@github.com:edx/edx-platform.git
+#Theming
+edxapp_theme_name: ''
```
jbau: this switches off theming.  any non-empty string turns it on.

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -19,13 +19,13 @@
   - lms-env
   - update
 
-
-- name: Create lms log target directory
```
jbau, line 23: this doesn't belong here.  Moved it into lms task, and also tasks of other lms variants

playbooks/roles/edxapp/handlers/main.yml
```diff
@@ -206,3 +235,21 @@
   tags:
   - cms
   - deploy
+
+- name: stop edxapp
```
feanil, line 239: So what was the issue with restart edxapp?  What about it didn't work?

jbau, line 239: I'm not sure if you see this also, but if I do this on our app servers, the edxapp process does bounce up and down, but none of the children gunicorn processes come back up.

feanil, line 239: That's very strange, I'm not seeing that but usually, by the time the
restarted task is applied the service is already stopped because of the
stop directive in deploy.yml

jbau, line 239: That's true, but state=restarted doesn't work even in that case for us.  we had to use state=started.

playbooks/roles/edxapp/vars/main.yml
feanil: Can you add a comment here to indicate this?  Or have a flag var that is separate from the name.

jbau:
rebased so it's mergable again.

playbooks/roles/edxapp/vars/main.yml
jrbl: I'd like a 1-line comment specifying that you use the empty string to indicate using no theme.

jrbl:
I'd like to see this rebased and squashed, with a more composed commit message, and I'd like to see the comment I refer to above. Otherwise, lgtm.

We're using this in practice for our prod deploys, and both last night and this morning went about as well as could be expected, considering the state of everything else.

I want this merged.

jbau:
Squashed the commits and added the comment @jrbl and @feanil wanted.  Will leave this up till tomorrow for anyone to comment on blockers, but will merge then if nothing blocking.

jrbl:
:+1:

PR #107, opened 2013-06-14
Extend ec2.py to accept an ini file

jrbl:
- End the constant reversion of ec2.ini preferences by allowing ec2.py
  to accept an ini file on the command line, and checking in an ini for
  edx-west that suits their needs.

@jbau check it out.


jbau:
lgtm

jbau:
ship it

PR #110, opened 2013-06-18
Jarv/edx sandbox

jarv:


playbooks/roles/edxapp/handlers/main.yml
```diff
@@ -236,6 +237,12 @@
   - cms
   - deploy
 
+- name: syncdb and migrate
+  shell: sudo -u www-data SERVICE_VARIANT=lms /opt/edx/bin/django-admin.py syncdb --migrate --noinput --settings=lms.envs.aws --pythonpath=/opt/wwc/edx-platform
+  when: migrate_db is defined
+  tags:
+  - syncdb
+
```
jbau, line 250: might want to tag with `-deploy` here?  We typically run deploys with "--tags=deploy" and I wonder if an untagged handler runs or not.

playbooks/secure_example/vars/edxapp_sandbox.yml
```diff
@@ -0,0 +1,139 @@
...
+    'ENABLE_PEARSON_HACK_TEST': false
+    'SUBDOMAIN_BRANDING': false
+    'SUBDOMAIN_COURSE_LISTINGS': false
+  'WIKI_ENABLED': true
+  'SYSLOG_SERVER':  'syslog.a.m.i4x.org'
+  'SITE_NAME':  'example.com'
+  'LOG_DIR':  '/mnt/logs/edx'
+  'MEDIA_URL':  ''
+  'ANALYTICS_SERVER_URL':  ''
+  'EEDBACK_SUBMISSION_EMAIL': 'bugs@edx.org'
```
jbau: typo here

playbooks/secure_example/vars/edxapp_sandbox.yml
```diff
@@ -0,0 +1,139 @@
...
+    'celery':
+      'BACKEND':  'django.core.cache.backends.memcached.MemcachedCache'
+      'KEY_FUNCTION':  'util.memcache.safe_key'
+      'KEY_PREFIX':  'sandbox'
+      'LOCATION': [ 'localhost:11211' ]
+  'CELERY_BROKER_TRANSPORT': 'amqp'
+  'CELERY_BROKER_HOSTNAME': ''
+  'COMMENTS_SERVICE_URL':  ''
+  'LOGGING_ENV':  'sandbox'
+  'SESSION_COOKIE_DOMAIN':  'amazonaws.com'
```
jbau: maybe make a comment about making sure people don't leave this in if they go to production?  Actually, will leaving this as `''` work?

jbau:
looks pretty good.  I don't think this will conflict with what we're doing.  just the few comments,

playbooks/edx_sandbox.yml
```diff
@@ -0,0 +1,32 @@
...
+    instance_type: m1.large
+    security_group: sandbox
+    image: ami-d0f89fb9
+    region: us-east-1
+
+  roles:
+    - launch_instance
+
+- name: Configure instance(s)
+  hosts: ec2-54-226-85-68.compute-1.amazonaws.com
```
jbau: should this also be localhost?  are you picturing users will run this from their instances?  I actually think that makes sense, and if so you can make this point a bit more clearly in the readme.

playbooks/secure_example/vars/edxapp_sandbox.yml
jbau: or maybe simple `'SESSION_COOKIE_DOMAIN':`, which should give `null` and then possibly translate to `None`?

https://docs.djangoproject.com/en/dev/ref/settings/#session-cookie-domain

playbooks/roles/nginx/templates/cms.j2
```diff
@@ -1,33 +1,13 @@
 server {
   # CMS configuration file for nginx, templated by ansible
       
-  listen 80;
+  listen {{cms_nginx_port}};
+
+  server_name studio.*;
```
jbau, line 6: wouldn't this still require dns?  I noticed you took this out for lms

playbooks/roles/nginx/templates/lms-preview.j2
```diff
@@ -1,31 +1,10 @@
 server {
   # LMS-preview configuration file for nginx, templated by ansible
 
-  listen 80;
+  listen {{lms_preview_nginx_port}};
 
-  server_name trace-lms-preview.*
-              preview.lms-dev.m.edx.org;
+  server_name preview.*;
```
jbau, line 6: same comment as cms server_name

playbooks/secure_example/vars/edxapp_sandbox.yml
jarv: I made the SESSION_COOKIE_DOMAIN "!!null" which should work for defaulting it to None.  I also added some instructions at the top.

playbooks/roles/nginx/templates/cms.j2
jarv, line 6: By default you can just use ports to get at the different services, in edxapp/vars/main.yml:

```
lms_xml_nginx_port: 18030
lms_nginx_port: 80
lms_preview_nginx_port: 18020
cms_nginx_port: 18010
```

So to access the cms after running this playbook on vagrant or an aws instance, http://host:18010

If you override the cms port so that it's port 80 (which is what we are doing for studio.sandbox.edx.org) it will also do a dns match for studio.*.  Seemed generic enough but I'm open to changes.

PR #111, opened 2013-06-18
Create roles for apache and shibboleth that run the lms and

jbau:
support registration and login via apache + mod_wsgi + mod_shib

Note that we now thing that gunicorn and mod_wsgi should not live
on the same machine if they are running the exact same codebase
with the exact same edx logging directories.  This seems to
cause deadlock at random times


jbau:
@jrbl and @jarv take a look?

playbooks/roles/apache/tasks/apache_site.yml
```diff
@@ -0,0 +1,21 @@
+# Requires nginx package
+---
+- name: Copying apache config {{ site_name }}
+  template: src={{ item }} dest=/etc/apache2/sites-available/{{ site_name }}
```
jarv, line 4: what does `{{ item }}` resolve to in this context?

playbooks/roles/shibboleth/vars/main.yml
```diff
@@ -0,0 +1,9 @@
+shib:
+  sp_entity_id: 'https://class.stanford.edu/shibboleth'
+  idp_entity_id: 'https://idp.stanford.edu/'
+  idp_metadata_url: 'https://idp.stanford.edu/Stanford-metadata.xml'
```
jarv: might want to make these more generic.

jarv:
one question about a var, other than that it looks good to merge.

playbooks/roles/apache/tasks/apache_site.yml
jbau, line 4: oh. it's just the `first_available_file`.

PR #112, opened 2013-06-18
Feanil/edxapp gunicorn

feanil:
Moving the gunicorn install to common and the edxapp scripts to the edxapp role.


jarv:
looks good

PR #113, opened 2013-06-18
Feanil/ansible xserver

feanil:


playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -46,13 +46,6 @@
...
   - update
 
-# Check out edx-platform repo to $app_base_dir
-- name: install git and its recommends
-  apt: pkg=git state=present install_recommends=yes 
-  tags:
-  - lms
-  - cms
-
 - name: install a bunch of system packages on which LMS and CMS rely
```
jarv, line 31: why is this removed?

feanil, line 31: It's now installed from common.

jarv:
xserver will require an nginx configuration as well right?
I would like for the default configuration to use non-clashing port numbers for both gunicorn and nginx which will help for putting all the services on a single box.  How does this sound?
- gunicorn ports:

```
lms-xml: 8030
lms: 8000
lms-preview: 8020
cms: 8010

xqueue: 8040
xserver: 8050
```
- nginx ports:

```
lms-xml: 18030
lms: 80
lms-preview: 18020
cms: 18010

xqueue: 18040
xserver: 18050

``
```

jarv:
reversed nginx and gunicorn in my previous comment, corrected.

feanil:
Yes,  the Nginx configuration is missing, I'll add that in and set its default port to 18040.  It looks like the aparmor configurations in the xserver example don't match up with what is happening in puppet.  I'm going to update them to match puppet.

jarv:
It doesn't look like requirements.txt installation into the virtualenv is covered yet

PR #114, opened 2013-06-18
Initial set of files to bring up a staging edx

jrbl:
@jbau this is the last bit needed to work with staging with ansible. Not much here; most of the goodies are in the vars in the other repo.

The main thing of interest here is stage-ansible.cfg and stage-ssh-config, which taken together offer a way to have an environment with different access control policies and a different VPC jumpbox without disturbing the usual case of working with your VPC production hosts.

To run this, edit stage-ssh-config with a sensible jumpbox configuration and make sure the options in stage-ansible.cfg are the set of ansible options you expect to use, then invoke ansible with ANSIBLE_CONFIG like so:

```
ANSIBLE_CONFIG=stage-ansible.cfg ansible-playbook -v -i './ec2.py' edxapp_stage.yml
```

### DIFF
--- a/playbooks/edx-west/edxapp_stage.yml
+++ b/playbooks/edx-west/edxapp_stage.yml
@@ -1,0 +1,15 @@
+- hosts: tag_environment_stage:&tag_function_webserver
+  sudo: True
+  vars:
+    secure_dir: ../../../edx-secret/ansible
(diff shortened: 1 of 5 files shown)


jrbl:
I've rebased and things. I'm canceling this until I'm pretty sure I'm ready to merge.

PR #115, opened 2013-06-18
if facter_processorcount is set, use that to determine # lms/cms/etc workers

jbau:
@jarv  this uses facter rather than ansible facts for no. of cores


jarv:
looks good

jbau:
going to merge for now.  may consider changing to `len(ansible_processor)`

PR #117, opened 2013-06-19
Jarv/xqueue support

jarv:


playbooks/roles/nginx/templates/xqueue.j2
```diff
@@ -0,0 +1,24 @@
+upstream app_server {
+    # For a TCP configuration:
+    server 127.0.0.1:{{ xqueue_gunicorn_port }} fail_timeout=0;
+}
+
+server {
+  listen {{ xqueue_nginx_port }} default_server;
```
feanil: The port variables should be changed to match  the new xqueue map.

feanil:
looks good.

playbooks/roles/xqueue/files/git_ssh.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /usr/bin/ssh -o StrictHostKeyChecking=no -i /etc/git-identity "$@"
```
jbau: Since I think you're using the public repo for xqueue, do we need this git identity part?

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -0,0 +1,36 @@
...
+  - xqueue
+
+# Do A Checkout
+- name: git checkout xqueue repo into $app_base_dir
+  git: dest={{xqueue_code_dir}} repo={{xqueue.source_repo}} version={{xqueue.version}}
+  tags:
+  - xqueue
+
+# Do Post Checkout Tasks.
+- name: create xqueue code dir
```
jbau, line 16: Minor point, but you're just changing perms here, right?  The name could be better.

playbooks/roles/xqueue/handlers/main.yml
```diff
@@ -0,0 +1,46 @@
...
+  - deploy
+
+
+
+# Install the python pre requirements into {{ venv_dir }}
+- name : install python pre-requirements
+  pip: requirements="{{xqueue.pre_requirements_file}}" virtualenv="{{venv_dir}}" state=present
+  tags:
+  - xqueue
+
```
jbau, line 22: Should all the handler tasks here have the `deploy` tag?  Just for completeness and accurate labeling?

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -0,0 +1,36 @@
+# Stop xqueue service.
+- name: stop xqueue service
+  service: name=xqueue state=stopped
+  tags:
+  - xqueue
+
```
jbau, line 7: and I think all the tasks in this file should definitely have the `--deploy` tag, so that a deploy can be run by specifying just this tag (though the entire role doesn't take much longer)

jbau:
thanks for letting me know about the change from platform_code_dir to edx_platform_code_dir.  Will update some of our upstart scripts accordingly

playbooks/roles/xqueue/files/git_ssh.sh
jarv: Yeah all of this was just copied from edxapp, I'll fix it up here since it looks like we are removing it from edxapp as well.

playbooks/roles/xqueue/tasks/deploy.yml
jarv, line 16: Yeah this was copied from edxapp where we are doing something similar, I don't remember the original reason for this task:

```
# Do Post Checkout Tasks.
- name: create platform code dir
  file: path={{edx_platform_code_dir}} state=directory owner=www-data group=www-data mode=755
  tags:
  - lms
  - cms
  - install

```

feanil, line 16: Because of the git identity stuff it is hard to checkout code with the www-data user so we have to change the permissions on the folder after the fact.  With the repo being public, this can probably get cleaned up a bit.

PR #118, opened 2013-06-19
Update git source repos to https

jrbl:
- Now that edx-platform is fully open source, the right thing to do is
  have read-only http URLs be the default and get rid of all the
  nonsense ssh-oriented infrastructure for config/deployment.
- Also contains updates to edx-west/edxapp_prod.yml which make its
  behavior less mysterious. (but this isn't the important part)

@jarv does this look ok to you?


jarv:
I think this means we can get rid of the identity file too right?

jrbl:
How do you like this version, @jarv and @jbau? I've made the changes more extensively, including removing the git identity file stuff.

jarv:
looks good to me

jbau:
there is merge conflict, but otherwise lgtm

PR #119, opened 2013-06-20
First pass at a vagrantfile that uses ansible to set up all edx apps

cpennington:


jarv:
I don't have time to test this right now but I think we should go ahead and merge it.
The one thing that might be a problem is the relative path to secure_example/,  secure_dir: "secure_example" .
This seems to work fine in 1.2 but I'm seeing issues in 1.3.

jarv:
ok to merge

PR #121, opened 2013-06-21
This adds termination support to the ec2 module

jarv:


e0d:
The license should be reviewed before this is merged.

jarv:
james gave the goahead for merging in the gpl lib.
Eventually this will be in core as well.

PR #123, opened 2013-06-21
use edxapp instead of individual gunicorn processes

jbau:
just a bunch of little bugfixes:  mainly in_production now just makes sure that edxapp is running instead of individual upstart jobs for lms, cms, etc

also handles the case where an env user is assigned no new groups.

@jarv should be quick


jrbl:
shipit

PR #124, opened 2013-06-21
Add lynx to packages for converting html to plaintext

kluo:
@jrbl
- For multipart email


jrbl:
Done. You can delete your branch if you'd like.

PR #126, opened 2013-06-22
Jbau/fix/xqueue underscore

jbau:
@jarv 

I think the find-and-replace of "xqueue." with "xqueue_" was a bit too aggressive and broke thing in a bunch of places.  This mainly fixes that and also the XQUEUE entry in envs.json, which I'm guess is a single hash and not a list of hashes.


jarv:
thanks, looks good.

PR #128, opened 2013-06-24
Add an empty default xserver env.

feanil:


playbooks/roles/xserver/vars/main.yml
```diff
@@ -1,6 +1,7 @@
 # Variables for the xserver.
 ---
 
+xserver_env_config: {}
```
jarv, line 4: do we need auth too?

feanil, line 4: xserver didn't have an auth as far as I could tell.

jarv, line 4: looks good

PR #134, opened 2013-06-27
Feanil/ansible rabbitmq

feanil:


playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
```diff
@@ -0,0 +1,16 @@
+{# Get the list of hosts that are in the same stack as the current machine
+   and also a rabbitmq machine.
+#}
+{%- set hosts= [] -%}
+{%- for name in  group_names -%}
+  {%- if name.startswith("tag_aws_cloudformation_stack-name_") -%}
```
jarv: Will this be a problem if we have multiple vpcs in the same AWS account that have rabbit clusters?

playbooks/roles/rabbitmq/vars/main.yml
```diff
@@ -0,0 +1,16 @@
...
+rabbitmq_config_dir: "/etc/rabbitmq"
+rabbitmq_cookie_dir: "/var/lib/rabbitmq"
+rabbitmq_cookie_location: "{{rabbitmq_cookie_dir}}/.erlang.cookie"
+
+rabbitmq_mnesia_folder: "{{rabbitmq_cookie_dir}}/mnesia"
+
+rabbitmq_port: 5672
+
+rabbitmq_auth_config:
+  erlang_cookie: "CHANGE ME"
```
jarv, line 16: maybe put a dummy admin user in secure_example/vars?
We will want a way for xqueue to connect when both it and rabbit are installed on the same machine.

playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
feanil: No, group_names is specific to this machine so it would only be an issue if a machine belonged to multiple cloudformation stacks which I don't think is possible.

jarv: ok, makes sense

playbooks/roles/rabbitmq/vars/main.yml
feanil, line 16: done.

playbooks/roles/rabbitmq/templates/rabbitmq.config.j2
jbau: can we parameterize these group names?

feanil:
name parametrized and also added a section for an alternative way of providing the clustered hosts for people not using AWS.

jarv:
looks like it needs a rebase, as soon as that done looks good to merge

jarv:
also we should fix the tests before merging, looking into it

feanil:
Rebased and merged into master.

PR #135, opened 2013-06-28
Changing stop/start back to restart now that things work for stanford

jbau:
I was noticing that a full run of the playbook edx-west/edxapp_prod.yml left edxapp stopped.  So this should fix it and get rid of the distinct stop and start 

@jarv @feanil

### DIFF
--- a/playbooks/roles/common/tasks/create_venv.yml
+++ b/playbooks/roles/common/tasks/create_venv.yml
@@ -5,6 +5,7 @@
   with_items:
   - python2.7
   - python-pip
+  - python2.7-dev
   tags:
   - pre_install
   - install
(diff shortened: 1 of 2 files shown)


jarv:
looks good

jbau:
hmm. notice that http://jenkins.edx.org:8080/job/ansible-tests/56/console failed because edxapp upstart didn't come back up on restart.  I guess this possibly replicates what I was seeing on the Stanford servers.
Maybe it only succeeds in certain machine states, which would be scary.  I wonder if the upstart script itself needs some attention?

jbau:
I'll manually do another run to see.

jbau:
and http://jenkins.edx.org:8080/job/ansible-tests/57/console fails b/c it can't stop edxapp services.

jbau:
I'm going to merge b/c it has worked several times on our release.

PR #136, opened 2013-06-28
Jbau/fix/migration flag

jbau:
@jarv @feanil 

I put in playbook prompts for migrate_db and edx-platform version, and made the migration flag checking more prompt friendly.

When we do set migrate_db in the code it currently is always set to True, so the flag checking change shouldn't break that.


jarv:
looks good

PR #140, opened 2013-07-03
A couple of bug fixes.

jbau:
- should be *_debian_pkgs instead of *_debian_packages
- removed duplicate task

@jarv some bugfixes missed in the previous PR and review.


jarv:
looks good

PR #141, opened 2013-07-08
closing single quote

yawboakye:
Closing single quote before closing back-tick


feanil:
This change is out of date and was fixed when the secure dir portion of the example went away..

PR #142, opened 2013-07-08
Feanil/cloudformation updates

feanil:


jarv:
looks good

e0d:
Looks good.

I guess we should consider whether we want to put xqueue in the main public subnet or make it's subnet public.  We'll definitely allows access at the network level to, say, the bastion box, that we might rather control to xqueue.  This could be an enhancement though.

PR #147, opened 2013-07-11
Jarv/single instance fixes

jarv:
This pull removes the edx_sandbox.yml variable definitions in the secure_example and instead makes the defaults for each role assume a local instance install.  It also fixes some misc issues we came across while troubleshooting google compute engine.

Most of the changes are cherry-picked from the jarv/gce branch except those that are specific to gce.


playbooks/roles/common/tasks/main.yml
```diff
@@ -15,13 +15,13 @@
   - update
 
 - name: Create data dir
-  file: path={{ app_base_dir }}/data state=directory owner=root group=root
+  file: path={{ app_base_dir }}/data state=directory owner=www-data group=root
   tags:
```
jarv, line 26: Studio needs to be able to write to this directory, unfortunately.

playbooks/roles/edxapp/vars/main.yml
```diff
@@ -29,13 +29,13 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
       'user':  'mongo'
   'MODULESTORE':
     'default':
-      'ENGINE': 'xmodule.modulestore.mongo.MongoModuleStore'
+      'ENGINE': 'xmodule.modulestore.mongo.DraftMongoModuleStore'
       'OPTIONS':  &lms_modulestore_default_options
```
jarv, line 33: This was causing studio to throw an exception, I believe it has always been wrong.

feanil, line 33: Is there a pull for making the same fix in configuration-secure?

jarv, line 33: not yet

feanil:
changes look good and you should merge them but I would also like to see a PR for the modulestore change in the configuration-secure repo.

playbooks/roles/common/tasks/main.yml
jbau, line 26: cool, good to get that fix

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -29,8 +29,7 @@
   - update
 
 - name: install a bunch of system packages on which LMS and CMS rely
-  apt: pkg={{item}} state=present
-  with_items: lms_debian_pkgs
+  apt: pkg={{','.join(lms_debian_pkgs)}} state=present
```
jbau, line 32: Just curious why this change--does it give better printouts, I wonder?

feanil, line 32: If you run apt with the 'with_items' directive, it runs apt seperately for each item.  If you join the package list together, it will run apt once for all the packages which should be significantly faster.

playbooks/roles/ora/vars/main.yml
```diff
@@ -9,8 +9,54 @@ ease_venv_dir: "{{ venv_dir }}"
 ora_gunicorn_workers: 4
 ora_nginx_port: 18091
 ora_gunicorn_port: 8091
-ora_auth_config: {}
-ora_env_config: {}
+
+ora_env_config:
```
jbau, line 18: perhaps add a comment about using secure_dir to if you want to override?

playbooks/roles/common/tasks/main.yml
jbau, line 26: (one slight concern)  is www-data a user before nginx or apache are installed?

playbooks/roles/ora/vars/main.yml
```diff
@@ -9,8 +9,54 @@ ease_venv_dir: "{{ venv_dir }}"
...
+  "DEBUG": false
+  "SYSLOG_SERVER": "localhost"
+  "USE_S3_TO_STORE_MODELS": false
+  "S3_BUCKETNAME": "openended-prod"
+
+ora_auth_config:
+  "USERS":
+    "xqueue_pull": "password"
+    "lms": "password"
+  "XQUEUE_INTERFACE":
```
jbau, line 49: It occurred to me that lms and cms should have these settings as well, i'll check if they do, and if not, will add that to this PR.

playbooks/roles/edxapp/vars/main.yml
```diff
@@ -55,8 +55,8 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
       'PORT': '3306'
   'PEARSON_TEST_PASSWORD':  ''
   'OPEN_ENDED_GRADING_INTERFACE':
-    'url':  ''
-    'password':  ''
+    'url':  'http://localhost:18091'
```
jbau, line 58: I added this.  hardcoded the default port b/c not sure if ora role would be run with this

playbooks/roles/edxapp/vars/main.yml
```diff
@@ -15,10 +15,10 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
   'AWS_SECRET_ACCESS_KEY':  ''
   'SECRET_KEY':  ''
   'XQUEUE_INTERFACE':
-    'basic_auth': [ 'noauth',  'noauth']
+    'basic_auth': [ 'edx',  'edx']
```
jbau, line 18: I changed this to be like ora / xqueue defaults

playbooks/roles/edxapp/vars/main.yml
```diff
@@ -15,10 +15,10 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
...
     'django_auth': { 'password':  'password',
       'username':  'lms'}
-    'url':  'https://localhost'
+    'url':  'https://localhost:18040'
```
jbau, line 21: I added this. hardcoded the default port b/c not sure if xqueue role would be run with this

jbau:
@jarv you should probably look at the changes I made, which were all to roles/edxapp/vars/main.yml
also, the one semi-blocking question I have is whether www-data exists when we assign it ownership of the /opt/wwc/data dir.

playbooks/roles/common/tasks/main.yml
jarv, line 26: I ran through this from scratch and it works fine though I'm not sure what is creating it, in any case I added a task to create the user in case it doesn't exist which might be necessary for another distro

PR #149, opened 2013-07-12
Need an extra package if we're using debian for rabbit clustering.

feanil:


jarv:
A bunch of roles have debian package requirements, maybe we can standardize on a name like `<role>_debian_pkgs` as a variable and also a standard convention across the roles for checking the OS?
I would at least like to get the package into a var in vars/main.yml.

feanil:
@jarv updated as you suggested.

jarv:
looks good

PR #150, opened 2013-07-12
Get end to end grader testing working.

feanil:


playbooks/roles/nginx/tasks/main.yml
```diff
@@ -62,3 +62,8 @@
   - nginx
   - update
 
+# If tasks that notify restart nginx don't change the state of the remote system
+# their corresponding notifications don't get run.  If nginx has been stopped for
+# any reason, this will ensure that it is started up again.
+- name: make sure nginx has started
+  service: name=nginx state=started
```
jarv, line 69: cool, I noticed this same problem

playbooks/roles/xqueue/tasks/main.yml
```diff
@@ -35,6 +44,11 @@
   tags:
   - xqueue
 
+- name: create xqueue consumer upstart script
+  template: src=xqueue_consumer.conf.j2 dest=/etc/init/xqueue_consumer.conf mode=0640 owner=root group=adm
+  tags:
+  - xqueue
+
```
jarv, line 63: I think it makes sense to make a handler for restarting xqueue and notify it whenever something changes in main.yml that requires a restart.  Then again if we are always including deploy.yml it will restart anyway.  I guess we need to figure out how deploys and changes that require service restarts will work.  Maybe we should put all service impacting changes in deploy.yml instead of splitting them between main.yml and deploy.yml?

playbooks/roles/xqueue/templates/xqueue_consumer.conf.j2
```diff
@@ -0,0 +1,17 @@
...
+
+respawn
+respawn limit 3 30
+
+env LANG=en_US.UTF-8
+
+chdir /opt/wwc/xqueue
+setuid www-data
+
+exec /opt/edx/bin/django-admin.py run_consumer --pythonpath=/opt/wwc/xqueue --settings=xqueue.aws_settings 12
```
jarv: what does `12` refer to?

playbooks/roles/xserver/tasks/main.yml
```diff
@@ -12,6 +12,11 @@
 - name: create sandbox python directory
   file: path={{ xserver_sandbox_venv_dir }} owner=ubuntu group=adm mode=2775 state=directory
 
+- name: create sandbox sudoers file
+  template: src=99-sandbox.j2 dest=/etc/sudoers.d/99-sandbox owner=root group=root mode=0440
+  tags:
+  - testit
```
jarv: should this tag remain?

playbooks/roles/xqueue/tasks/main.yml
feanil, line 63: won't most things in main.yml affect service if they require change?  We may need a better way of thinking about this.

playbooks/roles/xqueue/templates/xqueue_consumer.conf.j2
feanil: it's the number of workers per queue, I'll pull this out into a variable.

playbooks/roles/xqueue/templates/xqueue_consumer.conf.j2
```diff
@@ -0,0 +1,17 @@
...
+
+respawn
+respawn limit 3 30
+
+env LANG=en_US.UTF-8
+env WORKERS_PER_QUEUE={{xqueue_env_config.XQUEUE_WORKERS_PER_QUEUE}}
+chdir {{xqueue_code_dir}}
+setuid www-data
+
+exec {{venv_dir}}/bin/django-admin.py run_consumer --pythonpath={{xqueue_code_dir}} --settings=xqueue.aws_settings $WORKERS_PER_QUEUE
```
feanil, line 17: Noticing that this is started using the django dev server, should this be using gunicorn instead?  Note that this is also what it is doing in puppet template currently.

jarv, line 17: Yeah we should definitely do something else but this is the way we are running for now, maybe add a backlog item for it?

feanil, line 17: added to the backlog.

jarv:
looks good to merge

PR #151, opened 2013-07-12
E0d/forums

e0d:


e0d:
Tested with this playbook on Ubuntu and Debian

```
- name: Configure instance(s)
  hosts: edx_gce
  sudo: True
  gather_facts: True
  vars:
    migrate_db: True
    mysql5_workaround: True
  roles:
    - common
    - mongo
    - oraclejdk
    - elasticsearch
    - forum
```

playbooks/roles/rbenv/vars/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+
+#forum_user: forum
+
+# Ruby Specific Vars
+#ruby_base: /opt/forum
+#rbenv_root: "{{ ruby_base }}/.rbenv"
```
jarv, line 7: unfortunately these names will conflict with ruby vars in edxapp/vars/main.yml. 
Will need to clean up the edxapp role var names.

jarv:
As we discussed the rvm role can take paramters, something like

```
roles:
  - { role: 'rbenv', rbenv_user: 'forum' , rbenv_ruby_version: '1.9.3...' }

```

And then a sanity check in the role:

```
- fail: rbenv_user required for role
  when: rbenv_user is not defined
```

e0d:
Closing in favor of a version against master that has diverged substantially from this PR.

PR #158, opened 2013-07-15
Fixes bug where WORKERS was sometimes being set to "NAN"

jarv:
For the case where  "ansible_processor_cores": "NA"  and
facter_processorcount wasn't the WORKERS was set to the string "NA" *
the multiplier (like batman).
This will try to convert both to numbers and if the conversion fails it
will fall through and just set the WORKERS to the multiplier.

### DIFF
--- a/playbooks/roles/edxapp/templates/cms.conf.j2
+++ b/playbooks/roles/edxapp/templates/cms.conf.j2
@@ -13,10 +13,12 @@ respawn limit 3 30
 env PID=/var/tmp/cms.pid
 #env NEW_RELIC_CONFIG_FILE={{app_base_dir}}/newrelic.ini
 #env NEWRELIC={{venv_dir}}/bin/newrelic-admin
-{% if facter_processorcount %}
+{% if facter_processorcount|int(default="NaN") is number  %}
 env WORKERS={{ worker_core_mult.cms * facter_processorcount|int }}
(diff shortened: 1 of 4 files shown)


playbooks/roles/edxapp/templates/cms.conf.j2
```diff
@@ -13,10 +13,12 @@ respawn limit 3 30
 env PID=/var/tmp/cms.pid
 #env NEW_RELIC_CONFIG_FILE={{app_base_dir}}/newrelic.ini
 #env NEWRELIC={{venv_dir}}/bin/newrelic-admin
-{% if facter_processorcount %}
+{% if facter_processorcount|int(default="NaN") is number  %}
 env WORKERS={{ worker_core_mult.cms * facter_processorcount|int }}
+{% elif ansible_processor_cores|int(default="NaN") is number %}
+env WORKERS={{ worker_core_mult.cms * ansible_processor_cores|int }}
```
jbau, line 19: I think ansible_processory_cores is already an int, but I guess converting an int to an int will work. Is this path covered by the jenkins run?

jarv, line 19: yeah, I just wanted to keep the conversions consistent

jbau:
discussed with @jarv over hipchat. lgtm

PR #162, opened 2013-07-17
Use Relative pathes for ora and xqueue auth.json and env.json

jbau:
for when ora and xqueue are not installed right under app_base_dir

@jarv @feanil

### DIFF
--- a/playbooks/edx-west/edxapp_prod.yml
+++ b/playbooks/edx-west/edxapp_prod.yml
@@ -1,7 +1,7 @@
 # this gets all running prod webservers
 #- hosts: tag_environment_prod:&tag_function_webserver
 # or we can get subsets of them by name
-- hosts: ~tag_Name_app(4)_prod
+- hosts: ~tag_Name_app(12|22)_prod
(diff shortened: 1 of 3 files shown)


jarv:
seems ok to me, can we get a little more detail on why you want the change? I think I would rather have another variable defined like `{{ ora_base_dir }}`

jbau:
Basically, I was configuring a dev machine where ora and xqueue are both installed.  They both look for auth.json and env.json in their parent directory, so actually they can't be installed with a common app_base_dir (otherwise the later one clobbers the earlier).  Since we configure ora_code_dir and xqueue_code_dir, it's easier to use those as base

I think that also argues against another variable, since the settings files of both ora and xqueue expect the parent directory of the repo to contain the json files, so we might as well reflect that convention.

BTW, does your single-instance install case have this clobbering problem also?

jarv:
Yeah we probably have the same issue, I think I would prefer to change the name of the json file so that it's something like <role>.{env,auth}.json for it as well as xqueue, certificates and xserver.  No need to do that now though.

PR #165, opened 2013-07-17
DO NOT MERGE: initial commit of forum related roles

e0d:
works from scratch on gce and better generalization

missed file

changing user for checkout task

reverting path change

fix typo

fully working on debian

working upstart

moving env from bashrc

removing unused template

adding template for ruby env setting

removing unused template

removing forum and ruby env from bashrc, general cleanup

---

Run with a playbook like follows:

```
    - common
    - mongo
    - oraclejdk
    - elasticsearch
    - { role: 'rbenv', rbenv_user: 'forum', rbenv_user_home: '/opt/forum', rbenv_ruby_version: '1.9.3-p448' }
    - forum
```


playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -0,0 +1,18 @@
+- name: check for elasticsearch
+  command: test -f /var/tmp/elasticsearch-${elastic_search_version}.deb
+  ignore_errors: true
+  register: es_present
+
+- name: download elasticsearch
+  get_url: url={{ elastic_search_url }} dest=/var/tmp/{{ elastic_search_file }}
```
jarv: If you set `force=no` on `get_url` (which I think is the default but I'm not certain) it won't download the file if it already exists, eliminating the need for the check.

playbooks/roles/forum/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
+---
```
jarv, line 1: I think most of these tasks will need to notify the restart handler

playbooks/roles/forum/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
...
+
+- name: forum | copy cs_comments_service upstart script
+  template: src=cs_comments_service.conf.j2 dest=/etc/init/cs_comments_service.conf owner=root group=root mode=750
+  tags:
+  - forum
+  - install
+  only_if: "'${ansible_distribution}' == 'Ubuntu'"
+
+- include: deploy.yml
+- include: test.yml
```
jarv, line 56: Should including test.yml (in general) be the default or should we only do it when a `run_tests` var is defined.  I was thinking the latter but could go either way (or maybe a `skip_tests` var instead)

playbooks/roles/forum/tasks/test.yml
```diff
@@ -0,0 +1,8 @@
+---
+
+- name: forum | test that the required service are listening
```
jarv, line 3: This might be cleaner:

```
- name: wait for service to become available
  wait_for: port={{ item.port }} timeout=5
  with_items: "{{ form_services }}"
```

We could also put this right after the service restart handler if it's something we want to run every time.  If tests are optional then I would leave it here.

playbooks/roles/forum/templates/forum.sh.j2
```diff
@@ -0,0 +1,7 @@
+export MONGOHQ_USER=cs_comments_service
+export MONGOHQ_PASS=password 
+export RACK_ENV=development
+export SINATRA_ENV=development
+export API_KEY=tX5Dslg1y3OsCpFzRq
```
jarv: should these be variables?

playbooks/roles/forum/templates/forum_env.j2
```diff
@@ -0,0 +1,14 @@
...
+## This file is sourced in .bashrc.
+## The environment variables are used by the cs_comments_service
+## to configure itself at start-time.
+##
+export MONGOHQ_USER=cs_comments_service
+export MONGOHQ_PASS=password 
+export RACK_ENV=development
+export SINATRA_ENV=development
+export API_KEY=tX5Dslg1y3OsCpFzRq
+export SEARCH_SERVER=http://localhost:9200
```
jarv: should these be variables?

playbooks/roles/elasticsearch/tasks/main.yml
e0d: The check is also needed to short circuit installation.

playbooks/roles/forum/tasks/main.yml
feanil, line 56: I think I would rather have it run tests by default and have a skip_tests var instead if you wanted to skip them.  Especially as we are building the tests, it would be good to get them run often in a variety of environments.

playbooks/roles/forum/tasks/test.yml
e0d, line 3: Hmm, that may be more portable.  I think that I think that tests shouldn't be optional, but would be run independently.  We could pass in the test tag and validate all of our services across the fleet.

playbooks/roles/oraclejdk/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
...
+- name: oraclejdk | install Oracle Java
+  shell: exectuable=/bin/bash mkdir -p /usr/lib/jvm && tar -C /usr/lib/jvm -zxvf /var/tmp/{{ java_file }}
+  sudo: true
+  when_failed: $java_present
+  tags:
+    - forum 
+    - install
+
+#
+# TODO: relocate, maybe?
```
feanil: Is this todo needed anymore?  Are you still thinking of moving java instead of symlinking to it?

playbooks/roles/forum/templates/forum_env.j2
```diff
@@ -0,0 +1,14 @@
+##
+## Managed by Ansible, do not edit.
```
jarv: We should start using `{{ ansible_managed }}` at the top of our template files.

> Six additional variables can be used in templates: ansible_managed (configurable via the defaults section of ansible.cfg) contains a string which can be used to describe the template name, host, modification time of the template file and the owner uid, template_host contains the node name of the template’s machine, template_uid the owner, template_path the relative path of the template, template_fullpath is the absolute path of the template, and template_run_date is the date that the template was rendered.

playbooks/roles/rbenv/tasks/main.yml
```diff
@@ -0,0 +1,163 @@
...
+
+#
+# It's unfortunate that this needs to happen here,
+# but the rbenv needs to exist when the service
+# that relies on it is provisioned and the service
+# user and it's home needs to exist for rbenv to 
+# install into.
+#
+- name: rbenv | create rbenv owner $forum_user
+  user: name=$forum_user state=present shell=/bin/bash home=$forum_home
```
feanil, line 23: Is there a reason you're using this variable syntax instead of the jinja syntax?  In general I think we should stick with the jinja double curly braces for variable referencing.

playbooks/roles/elasticsearch/tasks/main.yml
jarv: Hm, the problem here is that if the file downloads but the package install fails it will skip subsequent runs.  Unfortunately it doesn't look like the apt module supports installing from .deb,

A better way might be to use creates where it's something like:

```
shell:  ... creates=/path/to/file/inpackage
```

Issue #159, opened 2013-07-16
Playbook fails on "'facter_processorcount' is undefined"


yuvadm:
Running the edx-platform ansible playbook on a single server scenario (Ubuntu 12.04 on AWS) halts on the following error:

``` bash
TASK: [writing lms upstart script to /etc/init] *******************************
fatal: [localhost] => {'msg': "'facter_processorcount' is undefined", 'failed': True}
```

yuvadm:
Upon playbook retry, the task completes successfully. Some sort of race condition here, possibly?

elimence:
I have tried using this before and i'm interested in knowing if the
commands are executed locally or on an ec2 instance?
Thanks

jarv:
elimence -

The commands are executed on the machine you are provisioning.
Ansible can either be invoked from a remote machine (in which case the
commands are run over ssh) or on the local machine.
The current instructions are for running Ansible from the machine you are
provisioning for simplicity.

Yuval - I haven't seen this error yet, there was a bug related to these
variables that was fixed recently -
https://github.com/edx/configuration/commit/fd917444a5a9b38f6147ca0a0c7a85b865c4e8aa
When you ran it do you know if it was before or after this fix?
-John

yuvadm:
@jarv run time was just a few minutes before I opened this issue, so definitely after `fd917444` was merged.

antoviaque:
@jarv Attempted a fix in #166 - not sure if it's the best way (I'm new to Ansible), but it allowed me to go through this.

feanil:
@antoviaque  https://github.com/edx/configuration/pull/168 should more cleanly deal with the issue.  The problem was being caused by not having facter installed before the first run of ansible.  This is why the first run would fail to have facter facts.  Subsequent runs would see that facter was installed(as part of the failing first run) and use it, which is why the issue happened only once.

PR #166, opened 2013-07-18
Fixes #159 - Playbook fails on "'facter_processorcount' is undefined"

antoviaque:
Fixes #159, related to #158

When facter isn't installed, the `facter_processorcount` variable makes
the integer conversion fail by being undefined.


feanil:
https://github.com/edx/configuration/pull/168 fixes this #159 by no longer relying on facter.

PR #167, opened 2013-07-18
Feanil/vagrant updates

feanil:
Updates to make the vagrant image work.


Vagrantfile
```diff
@@ -5,7 +5,7 @@ Vagrant.configure("2") do |config|
   config.vm.box     = "precise64"
   config.vm.box_url = "http://files.vagrantup.com/precise64.box"
 
-  config.vm.network :private_network, ip: "192.168.111.222"
+  config.vm.network :private_network, ip: "192.168.33.10"
```
jarv, line 8: any particular reason for "33.10"

jarv:
looks good

Vagrantfile
feanil, line 8: That is what ed was previously using, so just to be consistent with the vagrant boxes we were previously handing around.

feanil:
@jbau This pull makes changes to fix a bug with the migrate_db flag.  Because the flag was being set to false in group_vars/all it was overrding anything set in the playbook 'vars' section.  Now, migrations will run whenever the variable is defined at all, no matter its value.

PR #169, opened 2013-07-19
Make the condition on migrate_db a bit safer for using in var_prompt

jbau:
@feanil  @jarv 

Since we use vars_prompt for migrate_db, I wanted to reduce the possibility of accidentally running migrations.  So I've make the condition on it a bit stronger.


jarv:
looks good

feanil:
Looks good.

jarv:
I think we should eventually do a db migration dry run and then prompt if one is necessary but that can come later.

jbau:
@jarv yeah, I think that feature would be nice. (future PR =) )

PR #172, opened 2013-07-20
for edxapp, install python sandbox requirements into regular venv

jbau:
temporary fix to allow some things that depend on this (like
drag-and-drop questions) to not break.  Also, this configuration
might be what trial installs prefer, anyway.

@jarv @feanil

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -128,6 +128,24 @@
   - install
   - deploy
 
+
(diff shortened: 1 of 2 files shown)


feanil:
What problem is this trying to solve?  Why can't there be two seperate environments like the xserver has?

jbau:
we don't necessarily want to run the python sandbox without some testing on it first, so this is the temporary solution that has been endorsed by @nedbat , that we'd like to run with for a week or two.

feanil:
looks good.

PR #173, opened 2013-07-23
The number of workers need a multiplier, which I missed in code review

jbau:
@jrbl  b/c this caused downtime for us

@feanil  for FYI


jrbl:
shipit.

PR #176, opened 2013-07-23
Feanil/multiple vagrants

feanil:


playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -8,7 +8,6 @@
 
 - name: install python-software-properties if debian
   apt: pkg={{",".join(rabbitmq_debian_pkgs)}} state=present
-  when: ansible_distribution == 'Debian'
```
feanil, line 11: Removed this because, precise on vagrant also needs this package, it just that aws usually has it installed.  I think it makes sense to ensure it's always there for rabbit.

playbooks/roles/xqueue/tasks/main.yml
```diff
@@ -4,7 +4,7 @@
 #  - nginx/tasks/main.yml
 ---
 - name: Change permissions on datadir
-  file: path={{xqueue_code_dir}}/../data state=directory owner=www-data group=www-data
```
feanil, line 7: The relative paths were causing some issues in the vagrant ansible provisioning, it would error out saying the parent directory already existed.

jarv, line 7: cool, much better than using the relative paths regardless.

playbooks/vagrant-fullstack.yml
```diff
@@ -0,0 +1,19 @@
+- name: Configure instance(s)
+  hosts: vagrant
+  sudo: True
+  gather_facts: True
+  vars:
+    migrate_db: "yes"
+    openid_workaround: True
+  vars_files:
+    - "group_vars/all"
+    - "{{ secure_dir }}/vars/edxapp_ref_users.yml"
```
jarv: I think we can remove the reference to `edxapp_ref_users.yml`

playbooks/vagrant-fullstack.yml
```diff
@@ -0,0 +1,19 @@
...
+  roles:
+    - common
+    - nginx
+    - edxlocal
+    - edxapp
+    - rabbitmq
+    - xqueue
+    - xserver
+    - { role: 'edxapp', celery_worker: True }
```
jarv, line 18: would like to add ora/ease to this too if not now eventually.

jarv:
looks good

PR #180, opened 2013-07-25
forum role cleanup

jarv:
- consistent jinja variable syntax
- comment blocks on the tasks/main.yml files
- some syntax errors, variable name issues
- indenting, breaking up long lines
- various bug fixes

Also adds the playbook for sandbox.edx.org and studio.sandbox.edx.org


e0d:
+1

PR #193, opened 2013-07-30
Feanil/nat failover

feanil:


cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1533,7 +1695,142 @@
...
+          },
+          {
+            "Key":"environment",
+            "Value":{
+              "Ref":"EnvironmentTag"
+            },
+            "PropagateAtLaunch":true
+          }
+        ],
+        "UserData": { "Fn::Base64" : { "Fn::Join" : ["", [
```
jarv, line 1712: nitpick - this would look a little nicer if you did something like 

```
{ "Fn::Join" : ["\n", [<lines..>]] ....
```

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1533,7 +1695,142 @@
...
+          "#!/bin/bash -v\n",
+          "mkdir -p /opt/edx/bin\n",
+          "cd /opt\n",
+          "apt-get update\n",
+          "apt-get install openjdk-6-jre-headless unzip -y\n",
+          "wget http://s3.amazonaws.com/ec2-downloads/ec2-api-tools.zip\n",
```
jarv, line 1718: leaving the zip file in /opt is a little messy, can we clean up or download it to a different location?

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1559,6 +1856,24 @@
...
+            "IpProtocol":"tcp",
+            "FromPort":"80",
+            "ToPort":"80",
+            "CidrIp":"0.0.0.0/0"
+          },
+          {
+            "IpProtocol":"tcp",
+            "FromPort":"443",
+            "ToPort":"443",
+            "CidrIp":"0.0.0.0/0"
```
jarv, line 1874: what's driving the change to open up port 80 and 443 to the world?

feanil, line 1874: This is for egress, because we need to reach the outside world to do apt-updates to install the JRE and to get the ec2 tools.

feanil:
@jbau @jrbl Would love your input on this as well.

jbau:
top level question:  what testing has this undergone?

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1533,7 +1695,143 @@
...
+          "BACKUP_NAT_IP=`/opt/ec2-api-tools/bin/ec2-describe-instances $BACKUP_NAT_ID -U $EC2_URL | grep PRIVATEIPADDRESS -m 1 | awk '{print $2;}'`\n",
+          "\n",
+          "echo `date` \"-- Starting NAT monitor\"\n",
+          "\n",
+          "while [ . ]; do\n",
+          "  # Check the health of both instances.\n",
+          "  primary_pingresult=`ping -c $Num_Pings -W $Ping_Timeout $PRIMARY_NAT_IP| grep time= | wc -l`\n",
+          "\n",
+          "  if [ \"$primary_pingresult\" == \"0\" ]; then\n",
+          "    backup_pingresult=`ping -c $Num_Pings -W $Ping_Timeout $BACKUP_NAT_IP| grep time= | wc -l`\n",
```
jbau, line 1781: what happens when the backup nat fails but the primary nat does not?  should there be notification of that case?  Otherwise when the primary fails after the backup fails both immediately fail.

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1533,7 +1695,143 @@
...
+          "        NAT_STATE=`/opt/ec2-api-tools/bin/ec2-describe-instances $PRIMARY_NAT_ID -U $EC2_URL | grep INSTANCE | awk '{print $5;}'`\n",
+          "        if [ \"$NAT_STATE\" == \"stopped\" ]; then\n",
+          "          echo `date` \"-- NAT($PRIMARY_NAT_ID) instance stopped, starting it back up\"\n",
+          "          /opt/ec2-api-tools/bin/ec2-start-instances $PRIMARY_NAT_ID -U $EC2_URL\n",
+          "          NAT_HEALTHY=1\n",
+          "          sleep $Wait_for_Instance_Start\n",
+          "        else\n",
+          "          if [ \"$STOPPING_NAT\" == \"0\" ]; then\n",
+          "            echo `date` \"-- NAT($PRIMARY_NAT_ID) instance $NAT_STATE, attempting to stop for reboot\"\n",
+          "            /opt/ec2-api-tools/bin/ec2-stop-instances $PRIMARY_NAT_ID -U $EC2_URL\n",
```
jbau, line 1806: Also, just FYI, when our NAT wedged the first time, we could not stop it via the console. So we could not "restart" it to get it back into operation.  We literally had to bring up a new NAT instance.

I think some sort of manual intervention would basically be required if either the primary or secondary NAT goes down, so I think some form of notification would be essential.

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1533,7 +1695,143 @@
...
+          "            STOPPING_NAT=1\n",
+          "          fi\n",
+          "          sleep $Wait_for_Instance_Stop\n",
+          "        fi\n",
+          "      done\n",
+          "\n",
+          "      # Backup nat was healthy so we switched to it.  It is now the primary.\n",
+          "      if [ \"$NAT_HEALTHY\" == \"1\" ]; then\n",
```
jbau: I feel like `$NAT_HEALTHY` is not the right conditional for this flip?  Isn't `$ROUTE_HEALTHY` what's actually set after the routing tables are changed to use the backup NAT?

cloudformation_templates/edx-reference-architecture.json
```diff
@@ -1533,7 +1695,143 @@
...
+          "          echo `date` \"-- NAT($PRIMARY_NAT_ID) heartbeat failed, using $BACKUP_NAT_ID for $NAT_RT_ID default route\"\n",
+          "          /opt/ec2-api-tools/bin/ec2-replace-route $NAT_RT_ID -r 0.0.0.0/0 -i $BACKUP_NAT_ID -U $EC2_URL\n",
+          "          ROUTE_HEALTHY=1\n",
+          "        fi\n",
+          "        # Check NAT state to see if we should stop it or start it again\n",
+          "        NAT_STATE=`/opt/ec2-api-tools/bin/ec2-describe-instances $PRIMARY_NAT_ID -U $EC2_URL | grep INSTANCE | awk '{print $5;}'`\n",
+          "        if [ \"$NAT_STATE\" == \"stopped\" ]; then\n",
+          "          echo `date` \"-- NAT($PRIMARY_NAT_ID) instance stopped, starting it back up\"\n",
+          "          /opt/ec2-api-tools/bin/ec2-start-instances $PRIMARY_NAT_ID -U $EC2_URL\n",
+          "          NAT_HEALTHY=1\n",
```
jbau: also, I wish that `NAT_HEALTHY` would not be set until we pinged that the primary actually came up.

cloudformation_templates/edx-reference-architecture.json
feanil, line 1806: In the case of this script,  when the primary is unreachable, the backup takes over the routes as soon as possible.  Once that is done the script will attempt to stop and restart the primary NAT instance.  One thing that I will add in here is to continue to monitor the backup NAT while we are trying to restore the primary one.

I disagree that manual intervention is necessary when only one of the two NATs is down.  As for notifications, I was thinking that for the production systems we would have external notifications/alerts in datadog. If we start using VPCs fro sandboxes, I don't necessarily want a notification for a switch like this in every sandbox but I don't feel to strongly about that so I'm sure you could convince me otherwise if you think it's important.

jbau, line 1806: Whether manual _intervention_ is required or not, wouldn't you want _notification_ immediately either NAT box goes down, leaving a SPOF?  If you got notified that one of the production NAT machines (even the backup) went down, wouldn't you want to at least take a look manually to see if the script successfully reset the redundant configuration?  I still maintain that _notification_ is essential when either NAT goes down, which actually doesn't seem contradictory to your comment.  I don't think we want the first notification to arrive when internet connectivity has been lost to the app servers.

Also, we, for one, don't use datadog.  I'm not sure I understand how you intend to use it as a notification channel.  Do you picture it on the NAT instances themselves, on the bastion host, or something else?

feanil:
Answer with regards to testing:

I tested this monitor script by creating a new vpc which comes up with two nat devices, one of which is active.  The script logs at /var/log/nat_monitor.log so I tailed the file while I did the following.
1.  In the AWS Web Console: Stop the primary NAT box.
2.  Observe in the log to ensure that the monitor notices that the primary NAT is un-reachable(can't ping it).
3.  Observe in the log that the route has been switched over to the healthy NAT box.
4.  Verify that the route was actually change to the healthy NAT in the AWS VPC Console.
5.  Observe in the log that the nat monitor is trying to stop and start the NAT that is down.
6.  Verify that the NAT instance is being started up in the AWS EC2 Console.
7.  Observe in the log that the previously unreachable NAT has been started.
8.  Notice that the monitor swapped the primary and the backup NATs and is now monitoring the NAT  box that has all the routes going through it.

I then repeated the above process a couple of times to make sure that the script can make the switch in both directions.

To address some of @jbau 's other concerns.  Yes, we should have some sort of notification, but I'm not sure what the best place for that is.  My thought was that the health of the NAT boxes can be observed easily from Cloudwatch or Datadog(which connects to cloudwatch).  Alarms of this sort could be put in CloudWatch for the instances that we actually care about monitoring(We may be able to do this from the cloudformation template as well but that seemed like a related but separate task.)  We had been thinking about setting up instances on demand for developers here and in that case I didn't want notifications from those sandboxes to be going to the operational alerting system by default.

I don't have a lot of experience with the cloudwatch alarm system which is why I initially suggested datadog(their interface seems a bit easier to me at first glance).  I bet given some time we could add some alarms to the cloudformation template that would e-mail a configurable e-mail address when the alarms trigger but I think that could be done as a separate task.

jarv:
these changes look good to me but we should wait for @jbau 's approval before merging

feanil:
@jbau any more thoughts on this?

jbau:
sorry for being incommunicado, @feanil . got busy with some other stuff.
I think what you have merged looks good.  Truth be told, we are not really using the cloudformation templates, but thanks for the reference implementation for when we install our failover.  It'd be nice, when you have the cycles, to research/add the notification bit.

Thanks.

PR #194, opened 2013-07-31
E0d/notifier

e0d:
- Initial notifier role
- Simple, parameterized role for virtualenvs
- Github user creation task in common
- Style cleanup of files touched along the way.

I'm going to do a clean-room run after these updates to test for tabula rasa bugs after re-factoring.  That's not yet complete.

### DIFF
--- a/playbooks/roles/common/tasks/create_github_users.yml
+++ b/playbooks/roles/common/tasks/create_github_users.yml
@@ -1,0 +1,45 @@
+---
+
+# Overview:
+# 
+# Creates OS accounts for users based on their github credential.
+# Expects to find a list in scope named github_users with
+# the following structure:
+#
+# github_users:
+#   - user: me_at_github
+#     groups:
+#     - adm
+#   - user: otheruser
+#     groups: 
+#     - users
+#
+
+- name: common | create local user for github user
+  user: 
+    name={{ item.user }} 
+    {% if item.groups %}groups={{ ",".join(item.groups) }}{% endif %}
+    shell=/bin/bash
+  with_items: github_users
+  tags:
+    - users
+    - update
+
+- name: common | create .ssh directory
+  file: 
+    path=/home/{{ item.user }}/.ssh state=directory mode=0600 
+    owner={{ item.user }} group={{ item.user }}
+  with_items: github_users
+  tags:
+    - users
+    - update
+
+- name: common | copy github key[s] to .ssh/authorized_keys
+  get_url: 
+    url=https://github.com/{{ item.user }}.keys 
+    dest=/home/{{ item.user }}/.ssh/authorized_keys mode=0600 
+    owner={{ item.user }} group={{ item.user }}
+  with_items: github_users
+  tags:
+    - users
+    - update
--- a/playbooks/roles/notifier/tasks/deploy.yml
+++ b/playbooks/roles/notifier/tasks/deploy.yml
@@ -1,0 +1,77 @@
+#
+# TODO: Needed while this repo is private
+#
+- name: notifier | upload ssh script
+  copy: 
+    src=git_ssh.sh dest=/tmp/git_ssh.sh 
+    force=yes owner=root group=adm mode=750
+  tags:
+  - notifier
+  - install
+  - update
+
+#
+# TODO: Needed while this repo is private
+#
+- name: notifier | install read-only ssh key required for checkout
+  copy: 
+    src={{ notifier_git_identity_path }} dest=/etc/git-identity 
+    force=yes owner=ubuntu group=adm mode=60
+  tags:
+  - deploy
+  - install
+  - update
+
+- name: notifier | checkout code
+  git: 
+    dest={{ notifier_code_dir }} repo={{ notifier_source_repo }} 
+    version={{ notifier_version }}
+  environment:
+    GIT_SSH: /tmp/git_ssh.sh
+  tags:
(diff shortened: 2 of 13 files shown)


playbooks/roles/common/tasks/create_github_users.yml
```diff
@@ -0,0 +1,45 @@
...
+#     - adm
+#   - user: otheruser
+#     groups: 
+#     - users
+#
+
+- name: common | create local user for github user
+  user: 
+    name={{ item.user }} 
+    {% if item.groups %}groups={{ ",".join(item.groups) }}{% endif %}
```
jarv, line 21: neat, didn't know you could do this

playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -0,0 +1,77 @@
+#
```
jarv, line 1: should there be a stop and start at the beginning and end of this play?

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
...
+    home={{ notifier_home }} createhome=yes
+  tags:
+  - notifier
+  - install
+  - update
+
+- name: notifier | setup the notifier env
+  template: 
+    src=notifier_env.j2 dest="{{ notifier_home }}/notifier_env" 
+    owner="{{ notifier_user }}" group="{{ notifier_user }}"
```
jarv, line 75: nitpick - i don't think the double quotes are necessary here.

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -0,0 +1,74 @@
+---
```
jarv, line 1: I think we will want most of these tasks to notify the notifier handler

playbooks/roles/notifier/templates/notifier_env.j2
```diff
@@ -0,0 +1,24 @@
...
+# Comments Service Endpoint, for digest pulls
+export CS_URL_BASE="{{ notifier_comment_service_base }}"
+export CS_API_KEY="{{ notifier_comment_service_api_key }}"
+
+# User Service Endpoint, for notification prefs
+export US_URL_BASE="{{ notifier_user_service_base }}"
+export US_API_KEY="{{ notifier_user_service_api_key }}"
+
+# celery
+export BROKER_URL="{{ notifier_celery_broker_url }}"
```
jarv: For the env file config what do you think about generating the configuration from a hash? Something like:

```
notifier_cfg:
  NOTIFIER_ENV: ...
  EMAIL_HOST: ...
...
```

playbooks/roles/notifier/tasks/deploy.yml
e0d, line 1: Not yet.  For now this will run as a cron job and the API portion will be available in future release.

playbooks/roles/notifier/tasks/main.yml
e0d, line 1: As above, when the service is ready we'll need to.

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+
+- name: notifier | drop a bash_profile
+  copy: >
+    src=../../common/files/bash_profile 
+    dest={{ notifier_home }}/.bash_profile 
+    owner={{ notifier_user }} 
+    group={{ notifier_user }}
+
+- name: notifier | ensure .bashrc exists
+  shell: touch {{ notifier_home }}/.bashrc
```
feanil, line 90: why not use the file ansible module for this?

http://www.ansibleworks.com/docs/modules.html#file

playbooks/roles/notifier/templates/notifier.conf.j2
```diff
@@ -0,0 +1,10 @@
+description "Notifier Service"
+
+start on runlevel [2345]
+stop on runlevel [!2345]
+
+env PID=/var/tmp/notifier_service.pid
+
+chdir {{ notifier_code_dir }}
+
+exec {{ notifier_home }}/src/manage.py runserver
```
feanil, line 11: this should be run as the www-data user right?  Is that being enforced somewhere else or just missing here?

playbooks/roles/virtualenv/tasks/main.yml
```diff
@@ -0,0 +1,116 @@
...
+    path={{ virtualenv_user_home }} 
+    owner={{ virtualenv_user }}
+    group={{ virtualenv_user }}
+    state=directory
+  tags:
+    - install
+    - update
+
+- name: virtualenv | install pre-requisities
+  apt: pkg={{item}} install_recommends=yes state=present update_cache=yes
```
feanil: doing a join will be faster than with_item, since with_item here will call apt-get install individually for each item in the list.
eg from common/tasks/main.yml

```
pkg={{','.join(common_debian_pkgs)}} install_recommends=yes state=present update_cache=yes
```

playbooks/roles/notifier/templates/notifier.conf.j2
e0d, line 11: Not picking up the change, but I've added that to the upstarts script.  This will be replaced soon with supervisord scripts and the service won't be running until some significant changes to the service land.

PR #203, opened 2013-08-03
Basic role for consistently creating roles.

e0d:
I've been creating a lot of new roles and thought something like this would be useful.  It creates a role and stubs out some files.  I'd like it to include a header file in all the templates, but it's not working currently.  I think this would help us enforce consistency as we create more and more roles.

Runs something like so.

```
ansible-playbook ansible_role.yml -i local.ini --connection=local --extra-vars "role_name=haproxy"
```

### DIFF
--- a/playbooks/roles/ansible-role/templates/header.j2
+++ b/playbooks/roles/ansible-role/templates/header.j2
@@ -1,0 +1,9 @@
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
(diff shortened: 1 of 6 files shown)


playbooks/roles/ansible-role/templates/header.j2
```diff
@@ -0,0 +1,3 @@
+#
+# This file was created by the edX ansible-role ansible role.
+#
```
feanil: Where does this come in to play?  Doesn't seem like this file gets called/put in the new role.

e0d:
Includes now working.

feanil:
cool, looks good, did you need to make it a '.txt' to get it to work?

jarv:
looks good to me, the only addition we might consider is adding a deploy.yml though i guess that won't apply to every single role.

PR #206, opened 2013-08-07
remote servers need python-mysql to run ansible mysql module

jbau:
@jrbl this fixes the xqueue install problems we were seeing.

@feanil and @jarv FYI:  the mysql_db module requires apt package python-mysqldb to be installed on the remote machine, and xqueue didn't have it

### DIFF
--- a/playbooks/roles/xqueue/tasks/main.yml
+++ b/playbooks/roles/xqueue/tasks/main.yml
@@ -8,15 +8,6 @@
   tags:
   - xqueue
 
-- name: xqueue | create xqueue db
-  mysql_db: >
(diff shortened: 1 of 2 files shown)


playbooks/roles/xqueue/tasks/main.yml
```diff
@@ -31,6 +22,15 @@
   tags:
   - xqueue
 
+- name: xqueue | create xqueue db
```
jbau, line 25: moving mysql_db play after the apt-get install play, so python-mysqldb gets installed before it's required.

jrbl:
LGTM. :+1:

jrbl:
Merged. Check and delete your branch at your convenience.

PR #208, opened 2013-08-09
E0d/notifier updates

e0d:


playbooks/roles/notifier/files/opt/notifier/bin/forums_digest.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+. $HOME/.bashrc
+
+minutes=$1
+digest_date=`date --utc '+%Y-%m-%dT%H:%MZ'`
+
+cd /opt/wwc/notifier/src && /opt/wwc/notifier/virtualenvs/notifier/bin/python /opt/wwc/notifier/src/manage.py forums_digest --to_datetime=${digest_date} --minutes=${minutes}
```
feanil, line 8: can we make this a template and use vars for the location of the virtualenv and the notifier source directory?

e0d, line 8: This will go away entirely once we figure out what the issue with Celery beat is, but if I need to touch this again, I'll make it a template.

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -115,4 +126,71 @@
...
+  environment:
+    DD_API_KEY: "{{ notifier_dd_api_key }}"
+  when: dd_present|failed
+  tags:
+  - notifier
+  - install
+  - update
+  - ubuntu
+
+- name: common | create notifier/bin directory
```
jarv: typo

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -115,4 +126,71 @@
...
+  tags:
+  - notifier
+  - install
+  - update
+
+- name: notifier | install datadog agent
+  shell: bash -c "$(wget -qO- http://dtdg.co/agent-install-ubuntu)"
+  environment:
+    DD_API_KEY: "{{ notifier_dd_api_key }}"
+  when: dd_present|failed
```
jarv: can we use the "creates" option to the shell module instead here?

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -115,4 +126,71 @@
...
+    dest={{ notifier_home }}/bin/forums_digest.sh 
+    owner={{ notifier_user }} 
+    group={{ notifier_user }}
+  tags:
+  - notifier
+  - install
+  - update
+
+- name: notifier | make the script executable
+  file: path={{ notifier_home }}/bin/forums_digest.sh state=file mode=2755
```
jarv: the copy module also has an owner, group and mode option so this task can be combined with the previous one

playbooks/roles/notifier/templates/etc/supervisor/conf.d/notifier-celery-beat.conf.j2
```diff
@@ -0,0 +1,26 @@
...
+exitcodes=0,2
+stopsignal=TERM
+stopwaitsecs=10
+user=notifier
+redirect_stderr=false
+stdout_logfile=/tmp/notifier-celery-beat-stdout.log
+stdout_logfile_maxbytes=1MB
+stdout_logfile_backups=10
+stdout_capture_maxbytes=1MB
+stderr_logfile=/tmp/notifier-celery-beat-stderr.log
```
jarv: I'm not sure where to put the supervisord stdout and stderr logs but I think we should put them somewhere else, maybe a new directory under /mnt?

playbooks/roles/notifier/vars/main.yml
```diff
@@ -7,8 +7,11 @@ notifier_venv_dir: "{{ notifier_home }}/virtualenvs/notifier"
 notifier_source_repo: "git@github.com:edx/notifier.git"
 notifier_code_dir: "{{ notifier_home }}/src"
 notifier_version: "rc/digests"
+#notifier_version: "d98ad359c1e61cabc7ded3c81b03c5dc01dddd71"
```
jarv: should this be removed?

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -115,4 +126,71 @@
...
+  sudo_user: "{{ notifier_user }}"
+  tags:
+  - notifier
+  - install
+  - update
+
+- name: notifier | supervisord config for celery workers
+  template: 
+    src=etc/supervisor/conf.d/notifier-celery-workers.conf.j2 dest=/etc/supervisor/conf.d/notifier-celery-workers.conf 
+  notify: notifier | restart supervisor
```
jarv: restarting supervisor will have the effect in this case but as we add more supervisor services we will probably want to restart the individual service instead of supervisord

playbooks/roles/notifier/tasks/main.yml
e0d: This is also a temporary shim that will be replace when the celery beat issues have been resolved.

playbooks/roles/notifier/templates/etc/supervisor/conf.d/notifier-celery-beat.conf.j2
feanil: @e0d I agree with @jarv this should be moved out of tmp unless there is a good reason for them to be here?

playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -5,8 +5,11 @@
   copy:
```
jarv, line 5: Is it ok that we are doing the deploy steps while the service is running?

e0d, line 5: In this case I think so, but I've added code to stop before updating any way, because this will expose a RESTful interface before too long.

PR #209, opened 2013-08-09
Script for creating new jenkins server

jzoldak:
@wedaly @lapentab @e0d


playbooks/roles/edxlocal/tasks/main.yml
```diff
@@ -72,3 +72,69 @@
...
+
+- name: Install gcli
+  shell: rbenv exec gem install github_cli
+
+- name: install libxslt-dev
+  apt: pkg=libxslt-dev
+
+- name: Install rake
+  apt: pkg=rake
+
```
jarv: instead of going through each individually let's create a `edxlocal_debian_pkgs` var in `vars/main.yml` for this role and loop over it, with something like:

```
- name: edxlocal | install edxlocal specific system packages
  apt: pkg={{','.join(edxlocal_debian_pkgs)}} state=present
```

playbooks/roles/edxlocal/tasks/main.yml
```diff
@@ -72,3 +72,69 @@
...
+- name: TEMPORARILY DISABLE gcli
+  shell: echo "#!/bin/bash\necho gcli" > /usr/local/bin/gcli
```
jarv: please add a comment for the reason why this is necessary

playbooks/roles/jenkins/tasks/browsers.yml
```diff
@@ -0,0 +1,61 @@
...
+- name: Untar PhantomJS
+  command: tar -xjf /var/tmp/{{ phantomjs_archive }} -C /var/tmp/
+           creates=/var/tmp/{{ phantomjs_folder }}
+
+- name: Install PhantomJS
+  command: mv /var/tmp/{{ phantomjs_folder }} /usr/local/bin/phantomjs
+           creates=/usr/local/bin/phantomjs
+
+- name: Set PhantomJS permissions
+  command: chmod go+rx /usr/local/bin/phantomjs
```
jarv: you can use the file module here to prevent it from reporting "changed" after the permissions are set properly

playbooks/roles/jenkins/tasks/browsers.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Install chrome dependencies 4
+  apt: pkg=libcurl3
+
+- name: Install chrome dependencies 5
+  apt: pkg=xdg-utils
+
+- name: Install unzip O_o
+  apt: pkg=unzip
+
```
jarv, line 25: same comment as above about using a list of packages

playbooks/roles/jenkins/tasks/browsers.yml
```diff
@@ -0,0 +1,61 @@
...
+  command: wget https://dl.google.com/linux/direct/google-chrome-stable_current_amd64.deb
+           chdir=/var/tmp
+
+- name: Install Google Chrome 2
+  command: dpkg -i google-chrome-stable_current_amd64.deb
+           chdir=/var/tmp
+  ignore_errors: True
+
+- name: Install Google Chrome 3
+  command: apt-get -f -y install
```
jarv: missing package name?

playbooks/roles/jenkins/tasks/browsers.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Install ChromeDriver
+  shell: wget --quiet https://chromedriver.googlecode.com/files/chromedriver_linux64_2.1.zip
```
jarv: instead of wget here and for the fetch above can you use the get_url module?

playbooks/roles/jenkins/tasks/browsers.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Install ChromeDriver 2
+  shell: unzip chromedriver_linux64_2.1.zip
+
+- name: Install ChromeDriver 3
+  shell: mv chromedriver /usr/local/bin/chromedriver
+
+- name: Install Chromedriver 4
+  shell: chmod go+rx /usr/local/bin/chromedriver
+
```
jarv, line 46: should use the "creates" option to the shell module so that these commands are idempotent

playbooks/roles/jenkins/tasks/jenkins.yml
```diff
@@ -0,0 +1,68 @@
...
+    - shiningpanda.hpi
+    - ssh-credentials.hpi
+    - ssh-slaves.hpi
+    - subversion.hpi
+    - thinBackup.hpi
+    # - timestamper.hpi
+    - tmpcleaner.hpi
+    - token-macro.hpi
+    - translation.hpi
+    - violations.hpi
```
jarv: let's move this list into vars/main.yml for the jenkins role and use with_items for the plugin install, also remove commented lines unless there is a reason why we need to keep them.

playbooks/roles/jenkins/tasks/python-pkgs.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Symbolic link site packages
+  shell: ln -s -f /usr/local/lib/python2.7/site-packages/ /usr/local/lib/python2.7/dist-packages/
```
jarv: use the file module so this is idempotent

playbooks/roles/jenkins/tasks/python-pkgs.yml
```diff
@@ -0,0 +1,9 @@
...
+
+- name: install numpy
+  shell: pip install numpy==1.6.2
+
+- name: install scipy
+  shell: pip install scipy==0.11.0
```
jarv, line 10: why are these installed at the system level?

playbooks/secure_example/vars/edxapp_sandbox.yml
```diff
@@ -0,0 +1,226 @@
...
+cms_env_config:
+  <<: *edxapp_generic_env
+lms_preview_auth_config:
+  <<: *edxapp_generic_auth
+lms_preview_env_config:
+  <<: *edxapp_generic_env
+rabbitmq_auth_config:
+  admin:
+  - name: 'admin'
+    password: 'the example admin password'
```
jarv: this may have been committed in error, we moved our defaults out of the secure_example directory and into the individual roles.

playbooks/roles/jenkins/tasks/python-pkgs.yml
jzoldak, line 10: The problem we are trying to solve is to be able to wipe out the workspace and create a new virtualenv for each build so we know it's clean. But numpy and scipy take too long. Since we rarely change versions for these packages, we can mitigate this by installing them once at the system level.

jarv, line 10: I see, unfortunately unless you use the `--system-site-packages` option when you create the virtualenv it won't be able to see any of the system libraries.  Is that the current plan?

jzoldak, line 10: Yeah, we merged that change into jenkins/test.sh already. https://github.com/edx/edx-platform/blob/master/jenkins/test.sh#L58

playbooks/roles/jenkins/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: start xvfb
```
feanil, line 2: Didn't xvfb get moved into a jenkins plugin?  Is this still necessary?

playbooks/vars/jenkins.yml
```diff
@@ -0,0 +1,2 @@
+---
```
feanil: Is this file necessary?

playbooks/roles/edxlocal/tasks/main.yml
```diff
@@ -70,5 +70,25 @@
     password=password
     state=present
 
-- name: edxlocal | install memcached
-  apt: pkg=memcached state=present
+- name: edxlocal | install edxlocal specific system packages
+  apt: pkg={{','.join(edxlocal_debian_pkgs)}} state=present
+
+- name: Git clone ruby-build
```
feanil: Is there a reason this is used instead of using the rbenv role?  If there is, I'm happy to leave this here but I would externalize the ruby version to a var in vars/main.yml.

playbooks/roles/jenkins/handlers/main.yml
lapentab, line 2: It is not moved to a jenkins plugin to my knowledge.

playbooks/roles/edxlocal/tasks/main.yml
lapentab: rbenv was having issues and this seemed to solve it. I have parameterized the ruby version

feanil: The rbenv role was having issues?  Can you elaborate?

playbooks/roles/jenkins/handlers/main.yml
jzoldak, line 2: @feanil this first pass is to achieve parity with the current jenkins setup first. We do intend to look into the xvfb plugin and if it meets our needs will swap out in favor of that, but haven't gotten a chance to do that yet.

playbooks/roles/jenkins/vars/main.yml
```diff
@@ -0,0 +1,47 @@
+---
+jenkins_home: /var/lib/jenkins
+phantomjs_url: https://phantomjs.googlecode.com/files/phantomjs-1.9.1-linux-x86_64.tar.bz2
+phantomjs_archive: phantomjs-1.9.1-linux-x86_64.tar.bz2
+phantomjs_folder: phantomjs-1.9.1-linux-x86_64
+edxlocal_debian_pkgs: ["memcached", "ruby-bundler","rubygems","rbenv","npm","libgraphviz-dev","gfortran","libopenblas-dev","liblapack-dev","libxml2-dev","libgeos-dev","python-dev","libmysqlclient-dev","build-essential","pkg-config","libxslt1-dev","rake"]
```
feanil: change this to list view for easier editing and consumption.

playbooks/jenkins.yml
```diff
@@ -0,0 +1,12 @@
+- name: Configure instance(s)
+  hosts: jenkins
+  sudo: True
+  gather_facts: True
+  roles:
+    - role: rbenv
+      rbenv_user: jenkins
```
feanil: These should be variables set in the jenkins role in the same way that we currently do this for the forums role:

https://github.com/edx/configuration/blob/master/playbooks/roles/forum/vars/main.yml#L3-L5

playbooks/jenkins.yml
```diff
@@ -0,0 +1,12 @@
+- name: Configure instance(s)
+  hosts: jenkins
+  sudo: True
+  gather_facts: True
+  roles:
+    - role: rbenv
```
feanil: Is there a reason this gets run before common?

playbooks/roles/jenkins/vars/main.yml
```diff
@@ -0,0 +1,67 @@
+---
+jenkins_home: /var/lib/jenkins
+phantomjs_url: https://phantomjs.googlecode.com/files/phantomjs-1.9.1-linux-x86_64.tar.bz2
+phantomjs_archive: phantomjs-1.9.1-linux-x86_64.tar.bz2
+phantomjs_folder: phantomjs-1.9.1-linux-x86_64
+jenkins_user: "jenkins"
+jenkins_user_home: /home/jenkins
+jenkins_ruby_version: "1.9.3-p374"
+edxlocal_debian_pkgs:
+    ["memcached",
```
feanil: You shouldn't need the bracket and comas, just follow the same pattern as the jenkins_plugins list.

playbooks/roles/jenkins/vars/main.yml
```diff
@@ -0,0 +1,67 @@
+---
+jenkins_home: /var/lib/jenkins
+phantomjs_url: https://phantomjs.googlecode.com/files/phantomjs-1.9.1-linux-x86_64.tar.bz2
+phantomjs_archive: phantomjs-1.9.1-linux-x86_64.tar.bz2
+phantomjs_folder: phantomjs-1.9.1-linux-x86_64
+jenkins_user: "jenkins"
+jenkins_user_home: /home/jenkins
+jenkins_ruby_version: "1.9.3-p374"
+edxlocal_debian_pkgs:
```
feanil: This should be renamed to jenkins_debian_pkgs

playbooks/roles/jenkins/vars/main.yml
```diff
@@ -0,0 +1,67 @@
...
+    "liblapack-dev",
+    "libxml2-dev",
+    "libgeos-dev",
+    "python-dev",
+    "libmysqlclient-dev",
+    "build-essential",
+    "pkg-config",
+    "libxslt1-dev",
+    "rake"]
+ruby_version: 1.9.3-p374
```
feanil: This variable should no longer be necessary, sinces jenkins_ruby_version exists.  That should be used instead of this wherever this is referenced.

playbooks/roles/jenkins/vars/main.yml
```diff
@@ -0,0 +1,67 @@
+---
+jenkins_home: /var/lib/jenkins
+phantomjs_url: https://phantomjs.googlecode.com/files/phantomjs-1.9.1-linux-x86_64.tar.bz2
```
feanil: the phantom vars should be prefixed with the role name eg. 'jenkins_phantomjs_url'

feanil:
looks good. merge it when the final testing is done.

jzoldak:
PLAY RECAP *******************************************************************\* 
192.168.33.10              : ok=106  changed=92   unreachable=0    failed=0   

Looks good!
:thumbsup:

PR #212, opened 2013-08-14
fix some weird bugs in lms_auth hash merge conflict

jbau:
@jarv @feanil @jrbl 

Ansible's variable precedence strikes again.

When the hash merged happened to combine the lms_auth_config in this file and the one in our secrets file, the merge ended up taking data from _both_ sources, in a way I do not understand well.

Let me describe what was being merged.  
The customized settings in our secrets repo had:

```
lms_auth_config:  &lms_auth
  'MODULESTORE':   
    'default':   
      'ENGINE': 'xmodule.modulestore.mongo.MongoModuleStore'
      'OPTIONS':  &lms_modulestore_default_options
          'password':  'OUR REAL PASSWORD'
```

`roles/edxapp/vars/main.yml` used to have

```
edxapp_generic_auth_config:  &edxapp_generic_auth
  'MODULESTORE':
    'default':
      'ENGINE': 'xmodule.modulestore.mongo.DraftMongoModuleStore'
      'OPTIONS':  &generic_modulestore_default_options
         'password':  'password'

lms_auth_config:
  <<: *edxapp_generic_auth
```

When these two hashes merged, we ended up with

```
lms_auth_config:
  'MODULESTORE':
    'default':
      'ENGINE': 'xmodule.modulestore.mongo.DraftMongoModuleStore'
      'OPTIONS':  
         'password':  'OUR REAL PASSWORD'
```

which is just bizzare.

This PR fixes the problem for now, though perhaps at this point we should make explicit override hashes like `lms_auth_override_config` and use jinja2 syntax to do clearer preference...this kind of bug is sort of insidious.

### DIFF
--- a/playbooks/roles/edxapp/vars/main.yml
+++ b/playbooks/roles/edxapp/vars/main.yml
@@ -35,7 +35,7 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
   'MODULESTORE':
     'default':
       'ENGINE': 'xmodule.modulestore.mongo.DraftMongoModuleStore'
-      'OPTIONS':  &lms_modulestore_default_options
+      'OPTIONS':  &generic_modulestore_default_options
         'collection':  'modulestore'
         'db':  'edxapp'
         'default_class':  'xmodule.hidden_module.HiddenDescriptor'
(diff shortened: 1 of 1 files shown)


jarv:
Strange, I wonder if the hash merging precedence is different than the variable precedence? I'll spend some time today to see if I can reproduce the behavior. Looks good to merge.

PR #213, opened 2013-08-14
Making plugin installation more flexible, installing the management CLI.

e0d:


feanil:
I tested this in a local vagrant fullstack.  It seems to be working as expected.  I think it's good to merge.

PR #215, opened 2013-08-15
Staging setup for stanford

jrbl:
These are things I tweaked in the process of getting our staging infrastructure up and running. It's mostly all in edx-west.  @jarv can you confirm this doesn't break your world?

And @jbau please.

### DIFF
--- a/playbooks/roles/edxapp/vars/main.yml
+++ b/playbooks/roles/edxapp/vars/main.yml
@@ -91,7 +91,7 @@ generic_env_config:  &edxapp_generic_env
   'WIKI_ENABLED': true
   'SYSLOG_SERVER':  'syslog.a.m.i4x.org'
   'SITE_NAME':  'example.com'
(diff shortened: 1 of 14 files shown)


playbooks/roles/edxapp/vars/main.yml
```diff
@@ -91,7 +91,7 @@ generic_env_config:  &edxapp_generic_env
   'WIKI_ENABLED': true
   'SYSLOG_SERVER':  'syslog.a.m.i4x.org'
   'SITE_NAME':  'example.com'
-  'LOG_DIR':  '/mnt/logs/edx'
+  'LOG_DIR':  '/logs/edx'
```
jarv: what's this change for?

jrbl: Oh, sorry, that's a mistake. We use /logs on our hosts, but that shouldn't
get checked in for general use. I just pushed a commit that sets it right.
Nice catch!

PR #224, opened 2013-08-23
Include item.host for service checking

pchiu33:
Useful when database host isn't localhost (default value)


jarv:
dupe of https://github.com/edx/configuration/pull/235 closing

PR #230, opened 2013-08-27
Edx-West staging setup for workers, certificates

jrbl:
- First pass at a configuration that sets up certificate generation on
  West's staging instance
- Installs gpg components in place (from secrets), installs verification
  nginx recipe.

@jbau please.

This isn't anywhere near complete, but it's sufficient for us to put the verify.\* subdomain in place for today's install. This shouldn't affect non-stanford hosts.

### DIFF
--- a/playbooks/edx-west/edxapp_prod.yml
+++ b/playbooks/edx-west/edxapp_prod.yml
@@ -18,6 +18,7 @@
     # this indicates the path to site-specific (with precedence)
     # things like nginx template files
     local_dir:  '../../../configuration-secure/ansible/local'
+    certs_s3_bucket: 'http://cert-bucket.s3-website-us-west-1.amazonaws.com'
   vars_files:
     - "{{ secure_dir }}/vars/edxapp_prod_vars.yml"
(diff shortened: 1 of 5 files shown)


playbooks/edx-west/stage-app.yml
```diff
@@ -10,6 +10,7 @@
     not_prod: true
     secure_dir: ../../../edx-secret/ansible
     local_dir: ../../../edx-secret/ansible/local
+    certs_s3_bucket: 'http://cert-bucket.s3-website-us-west-1.amazonaws.com'
```
jbau: might we want different buckets for stage to prevent pollution?

playbooks/edx-west/edxapp_prod.yml
```diff
@@ -27,6 +28,7 @@
     - common
     - nginx
     - edxapp
+    - { role: 'certhost', cert_worker: False }
```
jbau, line 31: do we really want this on the app server?

jbau:
closed for now.

PR #207, opened 2013-08-08
ora-virtualenv: Use different virtual envs for ORA and EASE to avoid conflicts

antoviaque:
On a single instance, some libraries versions would conflict - like requests==0.14.2 (LMS) vs requests==1.1.0 (ORA)

The symptomatic error would then be:

```
(edx)www-data@edx-sandbox:/opt/edx$  DJANGO_SETTINGS_MODULE=lms.envs.aws SERVICE_VARIANT="lms-xml" /opt/edx/bin/gunicorn --preload -b 127.0.0.1:8030 -w 2 --timeout=300 --pythonpath=/opt/wwc/edx-platform  lms.wsgi
[...]
2013-08-08 13:08:31,185 ERROR 5392 [gunicorn.error] glogging.py:190 - Error handling request
Traceback (most recent call last):
  File "/opt/edx/local/lib/python2.7/site-packages/gunicorn/workers/sync.py", line 103, in handle_request
    respiter = self.wsgi(environ, resp.start_response)
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/handlers/wsgi.py", line 219, in __call__
    self.load_middleware()
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 45, in load_middleware
    mod = import_module(mw_module)
  File "/opt/edx/local/lib/python2.7/site-packages/django/utils/importlib.py", line 35, in import_module
    __import__(name)
  File "/opt/wwc/edx-platform/lms/djangoapps/course_wiki/course_nav.py", line 9, in <module>
    from courseware.courses import get_course_with_access
  File "/opt/wwc/edx-platform/lms/djangoapps/courseware/courses.py", line 9, in <module>
    from .module_render import get_module
  File "/opt/wwc/edx-platform/lms/djangoapps/courseware/module_render.py", line 54, in <module>
    requests_auth,
  File "/opt/wwc/edx-platform/common/lib/capa/capa/xqueue_interface.py", line 67, in __init__
    self.session = requests.session(auth=requests_auth)
TypeError: session() takes no arguments (1 given)
```

PR #234, opened 2013-09-04
sandbox-ora: Include xqueue & ORA on the `edx_sandbox` playbook

antoviaque:
Includes modifications to allow to run on the xqueue & ORA services on the same instance as the LMS & CMS, to fit them all into the sandbox VM.
- Switches to use the `virtualenv` role as suggested by @feanil in https://github.com/edx/configuration/pull/207
- Uses its own user to prevent the `virtualenv` role from attempting to modify the user `www-data`, which fails
- Adds handling of the `SERVICE_VARIANT` variable by xqueue and ORA (cf pull request https://github.com/edx/xqueue/pull/65 and https://github.com/edx/edx-ora/pull/136 ), which works the same way as for the LMS/CMS services, and allows to use a non-default filename for the env/auth configuration JSON files. Support for service variants is more than what is strictly needed here, but I thought it would be useful to maintain the configuration file naming mechanisms consistent across services.
- Allowed to configure the binding IP of rabbitmq, which by default binds on the main IP, while the default configuration of the other services expects localhost
- Added creation of edx/edx user in the default configuration of rabbitmq, to allow Xqueue to connect
- Fixed various issues with the default configuration values for ORA & Xqueue: missing '/' in config, 'NaN' printed in configuration files when ansible can't guess the number of cores, HTTP instead of HTTPS, etc.

You will need to set the following ansible variables to use the modified versions of ora and xqueue:

```
xqueue_source_repo: https://github.com/antoviaque/xqueue.git
ora_source_repo: https://github.com/antoviaque/edx-ora.git
```

### DIFF
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -35,3 +35,12 @@
       rbenv_user_home: "{{ forum_home }}"
       rbenv_ruby_version: "{{ forum_ruby_version }}"
     - forum
+    - { role: "xqueue", update_users: True }
+    - role: virtualenv
+      virtualenv_user: "edx-ora"
+      virtualenv_user_home: "/opt/edx-ora"
+      virtualenv_name: "edx-ora"
+    - role: ora
+      ora_user: "edx-ora"
+      ora_venv_dir: "/opt/edx-ora/virtualenvs/edx-ora"
+      ease_venv_dir: "/opt/edx-ora/virtualenvs/edx-ora"
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -35,6 +35,10 @@
       rbenv_user_home: "{{ forum_home }}"
       rbenv_ruby_version: "{{ forum_ruby_version }}"
     - forum
+    - role: virtualenv
+      virtualenv_user: "{{ xqueue_user }}"
+      virtualenv_user_home: "{{ xqueue_user_home }}"
+      virtualenv_name: "{{ xqueue_user }}"
     - { role: "xqueue", update_users: True }
     - role: virtualenv
       virtualenv_user: "{{ ora_user }}"
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -37,10 +37,7 @@
     - forum
     - { role: "xqueue", update_users: True }
     - role: virtualenv
-      virtualenv_user: "edx-ora"
-      virtualenv_user_home: "/opt/edx-ora"
-      virtualenv_name: "edx-ora"
+      virtualenv_user: "{{ ora_user }}"
+      virtualenv_user_home: "{{ ora_user_home }}"
+      virtualenv_name: "{{ ora_user }}"
     - role: ora
-      ora_user: "edx-ora"
-      ora_venv_dir: "/opt/edx-ora/virtualenvs/edx-ora"
-      ease_venv_dir: "/opt/edx-ora/virtualenvs/edx-ora"
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -26,7 +26,7 @@
     - nginx
     - edxlocal
     - edxapp
-    - rabbitmq
+    - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
     - { role: 'edxapp', celery_worker: True }
     - oraclejdk
     - elasticsearch
--- a/playbooks/roles/edxapp/vars/main.yml
+++ b/playbooks/roles/edxapp/vars/main.yml
@@ -23,7 +23,7 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
     'basic_auth': [ 'edx',  'edx']
     'django_auth': { 'password':  'password',
       'username':  'lms'}
-    'url':  'https://localhost:18040'
+    'url':  'http://localhost:18040'
   'CONTENTSTORE':
     'ENGINE':  'xmodule.contentstore.mongo.MongoContentStore'
     'OPTIONS':
--- a/playbooks/roles/edxapp/vars/main.yml
+++ b/playbooks/roles/edxapp/vars/main.yml
@@ -60,7 +60,7 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
       'PORT': '3306'
   'PEARSON_TEST_PASSWORD':  ''
   'OPEN_ENDED_GRADING_INTERFACE':
-    'url':  'http://localhost:18091'
+    'url':  'http://localhost:18091/'
     'password':  'password'
     'peer_grading':  'peer_grading'
     'staff_grading':  'staff_grading'
--- a/playbooks/roles/ora/tasks/deploy.yml
+++ b/playbooks/roles/ora/tasks/deploy.yml
@@ -85,7 +85,7 @@
   - deploy
 
 - name: ora | syncdb and migrate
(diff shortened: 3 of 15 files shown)


playbooks/edx_sandbox.yml
```diff
@@ -35,3 +35,12 @@
       rbenv_user_home: "{{ forum_home }}"
       rbenv_ruby_version: "{{ forum_ruby_version }}"
     - forum
+    - { role: "xqueue", update_users: True }
+    - role: virtualenv
+      virtualenv_user: "edx-ora"
```
feanil: You should set the ora related user/home/dirs in the ora/vars/main.yml.  Vars are read in before the play is run.  Take a look at https://github.com/edx/configuration/blob/master/playbooks/jenkins.yml for an example.

feanil:
@jbau can you also look this over?

jbau:
Yup will do in a couple of hrs.

playbooks/edx_sandbox.yml
antoviaque: Thanks for the example - and ok, I'll do the change. Do you also want me to move the ORA user dir to `/home`, like for `/home/jenkins`?

antoviaque:
@jbau Thanks! Don't hesitate to ask me to do things differently - I'm still learning ansible, and I'm new to the `configuration` repo, so advices will be appreciated : )

feanil:
No I think the location is fine, just externalize the data to the vars
file.

On Wed, Sep 4, 2013 at 12:25 PM, Xavier Antoviaque <notifications@github.com

jbau:
@antoviaque I'm going to run this install on a fresh instance as test

jbau:
@antoviaque hmm. I wonder if xqueue should also have its own virtualenv.  I think the dependencies are tracked in different repos.

jbau:
getting the following stack dump:

```
TASK: [xqueue | syncdb and migrate] ******************************************* 
failed: [localhost] => {"changed": true, "cmd": "sudo -u www-data SERVICE_VARIANT=xqueue /opt/edx/bin/django-admin.py syncdb --migrate --noinput --settings=xqueue.aws_settings --pythonpath=/opt/wwc/xqueue ", "delta": "0:00:00.200853", "end": "2013-09-05 06:29:42.259699", "item": "", "rc": 1, "start": "2013-09-05 06:29:42.058846"}
stderr: Traceback (most recent call last):
  File "/opt/edx/bin/django-admin.py", line 5, in <module>
    management.execute_from_command_line()
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 443, in execute_from_command_line
    utility.execute()
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 382, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 261, in fetch_command
    klass = load_command_class(app_name, subcommand)
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 69, in load_command_class
    module = import_module('%s.management.commands.%s' % (app_name, name))
  File "/opt/edx/local/lib/python2.7/site-packages/django/utils/importlib.py", line 35, in import_module
    __import__(name)
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/management/commands/syncdb.py", line 8, in <module>
    from django.core.management.sql import custom_sql_for_model, emit_post_sync_signal
  File "/opt/edx/local/lib/python2.7/site-packages/django/core/management/sql.py", line 6, in <module>
    from django.db import models
  File "/opt/edx/local/lib/python2.7/site-packages/django/db/__init__.py", line 11, in <module>
    if DEFAULT_DB_ALIAS not in settings.DATABASES:
  File "/opt/edx/local/lib/python2.7/site-packages/django/utils/functional.py", line 184, in inner
    self._setup()
  File "/opt/edx/local/lib/python2.7/site-packages/django/conf/__init__.py", line 42, in _setup
    self._wrapped = Settings(settings_module)
  File "/opt/edx/local/lib/python2.7/site-packages/django/conf/__init__.py", line 93, in __init__
    mod = importlib.import_module(self.SETTINGS_MODULE)
  File "/opt/edx/local/lib/python2.7/site-packages/django/utils/importlib.py", line 35, in import_module
    __import__(name)
  File "/opt/wwc/xqueue/xqueue/aws_settings.py", line 5, in <module>
    with open(ENV_ROOT / "env.json") as env_file:
IOError: [Errno 2] No such file or directory: path('/opt/wwc/env.json')

```

Basically it looks like xqueue actually reads its env.json (and auth.json, i presume) without the prefix.  so the SERVICE_VARIANT thing in the upstart scripts isn't doing much, I don't think.  I believe this should be fixed either by not installing it at /opt/wwc/xqueue (but rather /opt/wwc/xqueue/xqueue to allow the auth.json and env.json files to retain those names), or else a change has to be done in the xqueue repo changing the expected name of these files (this would also mean reinstalling Stanford and EdX's prod systems to pick up this change).  So maybe installing it not at /opt/wwc/xqueue is the path of least pain.

BTW, I believe /opt/wwc/ora will encounter the same issue

antoviaque:
@jbau Thanks for the review!

For adding a virtualenv for xqueue -- sounds good, I'll add one. Currently it works without it, but it might not always be the case, so it's a good precaution.

For the error with xqueue, I did provide a patch for that, to get it to take into account the `SERVICE_VARIANT` environment variable - cf the description of the current pull request: 

> Adds handling of the SERVICE_VARIANT variable by xqueue (cf pull request edx/xqueue#65), which works the same way as for the LMS/CMS services, and allows to use a non-default filename for the env/auth configuration JSON files. Support for service variants is more than what is strictly needed here, but I thought it would be useful to maintain the configuration file naming mechanisms consistent across services.

I should have made this more explicit, but until/if this is merged, to take it into account you need:

```
xqueue_source_repo: https://github.com/antoviaque/xqueue.git
```

I had opted for this path to maintain consistency between services, as a similar approach is already used for the LMS, and it keeps the config files all at the same place - do you want me to revert this?

antoviaque:
Ah, and for ORA, I can also add a similar mechanism - although since ORA remains the only service on the `edx_sandbox` playbook to use /opt/wwc/[auth|env].json, it doesn't currently conflict with the other services. That might however be useful to add other services to the same instance, later on.

Let me know how you prefer that I handle this, and I'll do it.

jbau:
ah I see.  My bad on the xqueue stuff.  @jtauber  can you help https://github.com/edx/xqueue/issues/65 land sometime soon (dunno who to ask).  @antoviaque I'm assuming in that PR if SERVICE_VARIANT isn't set it'll just default to looking for (auth|env).json, is that right?

Also, for ORA, let's handle it the same way as xqueue, to maintain consistency.

antoviaque:
@jbau Sounds good! I'll add the config prefix mechanism for ORA, the virtualenv for xqueue, and the change on the variables asked by feanil.

Thanks for the review.

antoviaque:
@jbau For your question about `SERVICE_VARIANT` - yes, that's correct. By default the prefix is empty, and only set if `SERVICE_VARIANT` is not empty.

antoviaque:
@jbau @feanil Done -- let me know if that's what you had in mind, and if you find any other issue.

I also created a pull request for ora, similar to xqueue: https://github.com/edx/edx-ora/pull/136

You will need to add the following to use the modified versions of ora and xqueue:

```
xqueue_source_repo: https://github.com/antoviaque/xqueue.git
ora_source_repo: https://github.com/antoviaque/edx-ora.git
```

feanil:
thanks @antoviaque

antoviaque:
@feanil Thanks for the review & merge! (and thanks @jbau too)

PR #237, opened 2013-09-05
Handle cases where we need to refresh ruby-build due to new ruby version

jbau:
@feanil 
Ran into a case where the ruby version requested was not in the already installed ruby-build. The solution here is to detect and reinstall when appropriate.

reference: https://github.com/sstephenson/ruby-build


jbau:
adding @e0d also since he's been near this code recently

e0d:
Looks reasonable to me.

PR #238, opened 2013-09-05
Added GEM_HOME environment to rbenv gem install

wedaly:
Add parameter to configure GEM_HOME for rbenv.

This resolves an issue I'm seeing with the Jenkins scripts in which the second time you run `gem install bundler` it tries to install to `/var/lib/gems` and gets a permission denied.  Explicitly specifying `GEM_HOME` ensures that gems are installed where you expect.

I tested this with the Jenkins script and Vagrant shortstack, and both scripts were able to install bundler successfully.

@e0d


wedaly:
Realized that the issues I was seeing were caused by not setting up bashrc correctly.  I need to rethink how I solve this for the Jenkins scripts.

PR #239, opened 2013-09-05
initial working version, tested on ubuntu.

e0d:


feanil:
looks good.

playbooks/roles/datadog/vars/main.yml
```diff
@@ -0,0 +1,13 @@
+---
+
+datadog_api_key: "PUT_YOUR_API_KEY_HERE"
```
e0d, line 3: This value actually comes from config secure, it's name differs there too.  Fixing the name would cause a var precedence issue, but something needs to happen here.

PR #245, opened 2013-09-09
Startup script to ensure nginx log dirs present

jrbl:
@jbau and @sefk please

### DIFF
--- a/playbooks/roles/nginx/tasks/nginx_site.yml
+++ b/playbooks/roles/nginx/tasks/nginx_site.yml
@@ -1,5 +1,14 @@
 # Requires nginx package
 ---
+- name: nginx | Install job to ensure log directories exist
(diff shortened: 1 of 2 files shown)


sefk:
What was the problem this was meant to fix @jrbl ?  This looks fine, of course, just don't remember what started it.

jrbl:
I don't remember either. This was a trello task you created a long time
ago. You and Jason manually created these startup tasks on the prod hosts,
and then made a card for ansible-izing it, which I took care of yesterday
in fresh-from-vacation mode.

I think this might have been connected to moving the logs to EBS volumes,
which we eventually gave up on but which I think is still a good idea when
we get a round tuit.

jarv:
going through old PRs and closing this, feel free to re-open if you think we still need this.

PR #247, opened 2013-09-11
Change inventory_file to inventory_path in Vagrantfiles

andersonvieira:
According to http://docs.vagrantup.com/v2/provisioning/ansible.html the property iventory_file is now iventory_path. Trying to run "vagrant up" with the current Vagrantfiles and the newest Vagrant version results in the following error:

ansible provisioner:
- The following settings don't exist: inventory_file


feanil:
Looks like this was fixed by https://github.com/edx/configuration/pull/251

PR #251, opened 2013-09-12
vagrant provisioning updates

jarv:
- updates vagrant files for Vagrant 1.3.1 compatibility
- attempts to fix race condition where rabbitmq guest user addition was
  failing


feanil:
You should also update the shortstack-xml vagrant file but otherwise looks good.

PR #253, opened 2013-09-13
nginx configuration for long site names

sefk:
Three fixes in here needed for Carnegie / CME
1. Add support for specifying a "template_subdir" var that will be used when setting up the nginx sites-available.  If it's absent then no problem, but if it's there, look in that subdirectory of the templates dir first.  
2. Install our own nginx.conf.  This way we can specify a wider site_name_hash line, required with really long site names (like `preview.class.claslite.carnegiescience.edu`)
3. Noticed that perms were wrong on sites-available files, fixed those while I was in there.

@jbau pls review

### DIFF
--- a/playbooks/edx-west/carnegie-prod-app.yml
+++ b/playbooks/edx-west/carnegie-prod-app.yml
@@ -15,7 +15,7 @@
     - "{{ secure_dir }}/vars/edxapp_prod_users.yml"
   roles:
     - common
-    - nginx
+    - {'role': 'nginx', 'nginx_conf': true} 
     - {'role': 'edxapp', 'openid_workaround': true, 'template_subdir': 'carnegie'}
     # run this role last
     # - in_production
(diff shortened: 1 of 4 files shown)


playbooks/edx-west/carnegie-prod-app.yml
```diff
@@ -1,25 +1,8 @@
...
-# this gets all running prod webservers
-#- hosts: tag_environment_prod:&tag_function_webserver
-# or we can get subsets of them by name
 - hosts: ~tag_Name_app(10|20)_carn
   sudo: True
-  vars_prompt:
-    - name: "migrate_db"
-      prompt: "Should this playbook run database migrations? (Type 'yes' to run, anything else to skip migrations)"
-      default: "no"
-      private: no
```
sefk: we never want to migrate from here

jbau: well, they're different dbs on the same db host (so I don't quite see the difference with the carnegie instance versus others)

sefk: duh, of course you're right @jbau.  restoring.

jrbl:
I just pushed my change onto your branch, @sefk. With that small change, this is the branch I tested, and it seemed to work ok for me. :+1:

PR #254, opened 2013-09-13
edx-west/{prod,stage}-jumpbox: barebones playbooks

sefk:
Just to facilitate user updates to the jumpboxes.

Please pull @jbau


jbau:
edx-west only. easy peasy

PR #262, opened 2013-09-17
add edx-west/.gitignore for symlink to config repo

sefk:
Not checking those symlinks in since they are symlinks to another repo,
but generally people will want these.  On my machine they look like
this:
    configuration-secure@ -> ../../../configuration-secure
    edx-secret@ -> ../../../configuration-secure

Just for the edx-west configuration files.

Please comment and review @jrbl and @jbau.


jbau:
lgtm

PR #265, opened 2013-09-19
support ansible 1.3

jarv:
These are some minimal changes to support 1.3, does not yet include role
dependencies.


feanil:
looks good.

PR #267, opened 2013-09-20
collectstatic cleanup

jarv:
rake gather_assets replaces collectstatic
so all references and logic to handle both
are removed

removes collect_static.yml since now there are
only two tasks for rake gather_assets.


jarv:
tested against sandbox.edx.org

feanil:
:+1:

PR #271, opened 2013-09-21
java-link-var: Replace `java_link` var by `oraclejdk_link`

antoviaque:
This variable doesn't seem to be defined anywhere, and (in v1.3?) ansible
complains about the missing value. Since there is a `oraclejdk_link` which
seem to correspond, replacing.

(Had this after upgrading to ansible 1.3 & running the edx_sandbox from scratch with the current configuration master)


jarv:
seems to be already fixed in commit 1cda72cec7edd9ff9ca0f0e50c372dab98720430 , closing.

PR #275, opened 2013-09-24
Notifier install fixes / seen while installing at Stanford

sefk:
Hi @e0d, can you review these changes?  This is the branch I was working off of getting notifier installed out here.  There are some new config files in edx-west/\* that you shouldn't care about.  Aside from that there are tactical things in the code that I'll comment about inline.

It'd be great to get this reviewed soon so we could install on prod tomorrow.  Thanks!


playbooks/edx-west/carnegie-prod-app.yml
```diff
@@ -21,7 +18,7 @@
     - "{{ secure_dir }}/vars/edxapp_prod_users.yml"
   roles:
     - common
-    - {'role': 'nginx', 'nginx_conf': true} 
+    - {'role': 'nginx', 'nginx_conf': true}
```
sefk, line 21: unrelated cleanups

playbooks/edx-west/prod-worker.yml
```diff
@@ -14,6 +14,12 @@
...
     - "{{ secure_dir }}/vars/shib_prod_vars.yml"
+    - "{{ secure_dir }}/vars/notifier_prod_vars.yml"
   roles:
     - common
     - { role: 'edxapp', celery_worker: True }
+    - role: virtualenv
+      virtualenv_user: "notifier"
+      virtualenv_user_home: "/opt/wwc/notifier"
+      virtualenv_name: "notifier"
+    - notifier
```
sefk, line 35: adding the notifier to all our worker machines.  We currently have two, util10 and util20, both m1.smalls that aren't too busy.

playbooks/roles/notifier/files/git_ssh.sh
```diff
@@ -1,2 +0,0 @@
-#!/bin/sh
-exec /usr/bin/ssh -o StrictHostKeyChecking=no -i /etc/git-identity "$@"
```
sefk, line 2: removed this wrapper since not longer needed, now that the repo is open

playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -46,40 +19,14 @@
...
-    group={{ notifier_user }} mode=2750 recurse=yes
-  tags:
-  - notifier
-  - deploy
-  - install
-  - update
-
-#
-# TODO: Needed while this repo is private
-#
```
sefk, line 18: all the removals in here are just switching from a private to public repo.  Note that in my vars file I changed from a git: link to a https: link, maybe you'd like to do the same?  Not required I believe.

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -126,6 +126,14 @@
   - install
   - update
 
+- name: notifier | create notifier/db directory
+  file:
+    path="{{ notifier_home }}/db" mode=2775 state=directory
+  tags:
+  - notifier
+  - install
+  - update
```
sefk: got to put the local sqlite3 db somewhere

playbooks/roles/notifier/tasks/main.yml
```diff
@@ -134,7 +142,18 @@
...
+  copy: >
+    src=../files/opt/notifier/bin/forums_digest.sh
+    dest={{ notifier_home }}/bin/forums_digest.sh
+    owner={{ notifier_user }} 
+    group={{ notifier_user }}
+  tags:
+  - notifier
+  - install
+  - update
+
```
sefk: This file was getting its perms set below, and existed in the repo, but just wasn't getting copied to the dest.  Just an oversight I presume

e0d: hmm, probably a partial re-factor that worked on existing environments because the file was already dropped.  This file is cruft from when we were running the scheduler as a cron job.  This and the section setting permissions can both be removed.

playbooks/roles/notifier/tasks/deploy.yml
e0d, line 18: Yeah this all makes sense.

playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -1,43 +1,16 @@
...
-  tags:
-  - notifier
-  - deploy
-  - install
-  - update
-
+---
 - name: notifier | stop notifier-celery-beat
   supervisorctl: name=notifier-celery-beat state=restarted
+  ignore_errors: yes
```
e0d: why?

e0d:
Other than my question about ignoring supervisor errors, LGTM.

playbooks/roles/notifier/tasks/main.yml
sefk: ah, I figured this file was there for a reason.  I'll just remove the file and this copy business then?

playbooks/roles/notifier/tasks/deploy.yml
sefk: Boostrapping case: when you're installing for the first time on a machine, there's nothing to stop, so it fails.

e0d: Hmm, I think ansible should handle this -- in fact I thought it did.  I don't want conditional non-sense to support the first install, but I'd also like to know that the services have started.

I also just noticed that there are still reference to beat above which have been removed from master, need a rebase?

sefk: This was probably failing, not just because I was trying to stop something that wasn't running, but it was failing because ~/src wasn't there all.  Maybe related to a weirdness that I spent some time debugging that the PID file was getting written to ~notifier/src.  Even thought that's not the right place, I couldn't figure out how to fix.

Anyway, I couldn't spend all day debugging so this got me past.

sefk:
@e0d, OK to pull?  Want to install today.  Thx.

e0d:
Yeah, looks good.

sefk:
Thanks for the review and OK, @e0d

PR #278, opened 2013-09-24
Added task to make sure elasticsearch is started

carsongee:
I recently had a build where elasticsearch hadn't auto started so
I added that check here.


jarv:
looks good

PR #279, opened 2013-09-24
Forum role port test now properly uses host argument

carsongee:
It was using localhost for every test, and without a mongo instance running locally it failed. The host argument was already defined in defaults so it seemed like just a glitch.


e0d:
+1

PR #280, opened 2013-09-24
Add the dev common vars to the list of vars files.

feanil:


jarv:
@feanil if you still want this in master can you re-base?

PR #273, opened 2013-09-23
rabbitmq-atp-update: Fixes "Failed to find required executable rabbitmq-plugins"

antoviaque:
(Fix for another issue I've had since the upgrade to ansible 1.3)

Without the `apt-get update`, the apt cache is out of date while attempting to install
rabbitmq. An older version from the Ubuntu repo is installed instead of the one from
the rabbit repository, which makes the role fail later on with:

```
TASK: [rabbitmq | install plugins] ********************************************
[...]
failed: [10.145.2.22] => {"failed": true, "item": ""}
msg: Failed to find required executable rabbitmq-plugins
```

PR #283, opened 2013-09-25
Modified vagrant-fullstack by using sandbox role definitions - Tested

carsongee:
This plus PR #273 should make everything 1.3 compatible. I built shortstack and fullstack vagrant with them.


e0d:
+1

PR #286, opened 2013-09-26
Notifier config improvements: db dir, email vars

sefk:
- specify notifier_db_dir as a setting.  Had previously just stuck
  under $HOME, but Jim preferred as a distinct setting.
- pass through email settings env variables
- plumb through django's EMAIL_USE_TLS setting.  Required to be
  able to send mail using Amazon SES.  They speak smtp, but only
  over TLS.

Pls review @e0d.


e0d:
+1

sefk:
Thanks for the review @e0d, proceeding w/ merge.

PR #288, opened 2013-09-27
edx-west prod: don't install notifiers yet

sefk:
The notifier isn't ready yet, needs theming to be done before it can be installed.  Affects the edx-west machines only.

@jrbl -- make sure you pull this in before this week's install.

### DIFF
--- a/playbooks/edx-west/prod-worker.yml
+++ b/playbooks/edx-west/prod-worker.yml
@@ -18,18 +18,22 @@
     - common
     - { role: 'edxapp', celery_worker: True }
 
+#
+# COMMENT OUT THE NOTIFIER UNTIL IT IS READY
+#
+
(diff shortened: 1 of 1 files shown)


jrbl:
Do we not need the workers for anything but notification? I thought @jbau had stuff that ran on the workers.  I mean, I'm happy to have one less machine to install - but I want to make sure that I really do. @jbau?

jbau:
no. nothing else runs on the workers
but again, we should install workers whenever we update the prod machines, to keep edx-platform consistent between app and worker

jrbl:
ok, understood, thanks.

jrbl:
lgtm @sefk

jrbl:
merged, you can delete your branch whenever you'd like.

PR #289, opened 2013-09-27
Slight refactor to make this more like edxapp and reuse nginx role

carsongee:


jarv:
I think the direction we want to go with nginx configurations is to have the roles include their own nginx config like we were doing here (edxapp was done first which is why it is different than some of the others).
With role includes what we can probably do is let all the roles that need nginx depend on it for the base configuration.

carsongee:
That is fine, but as it is this breaks a build where nginx and xserver aren't on the same machine. So if we want to use this method instead of the edxapp method then we at least need to check whether the nginx role has been run.  Should I close this and submit a PR with the relevant when statements then?

jarv:
Right - maybe we use role includes to make this do the right thing in ansible 1.3?  Typically what we have been doing for other roles is to put role dependencies in the comment block at the top of `tasks/main.yml`

carsongee:
I think using the new role dependency stuff added 1.3 is great, but in this case I explicitly don't want nginx installed and running on the same machine as xserver (same for edxapp servers) since I'm running that on a dedicated load balancer.  So, we could maybe make a new role called nginx-site or similar and do dependencies on that with each task depending on the nginx role variable that is set or not.  That should probably solve it both ways somewhat elegantly.  Should I go ahead and do that for all these roles that have nginx items in them to replace the file include?

jarv:
I see, running gunicorn without nginx wasn't really a use-case we have considered since all of our static files are still being served from the instances behind the ELB. 
It sounds like a conditional nginx dependency in our roles is what we want then which will require some refactoring.  Do you know if it's possible to conditionally include a role dependency?  I was also going to suggest to leave nginx installed and configured but have an override to turn it off during the run though that isn't ideal.

carsongee:
Yeah, I just rsync the static files up to the load balancer from one of the app instances and I haven't seen any issues (still not in production).  The role dependencies don't appear to have anyway to make them conditional from any of the reading I've seen on ansible (just like includes can't be conditional).  So we have to make tasks conditional, and I'd rather not have to add a new role to app server playbook that just turns off nginx.  I think the way it is now for edxapp and the consistency this change adds to make it behave the same way is probably the most flexible.  

If we really don't like that though, I don't see another option other than having an "nginx-site" role that effectively is just what nginx_site.yml is in the existing nginx role, but I'm glad to help and I'm certainly open to other ideas.

jarv:
>  The role dependencies don't appear to have anyway to make them conditional from any of the reading I've seen on ansible (just like includes can't be conditional).

I'm not sure about role dependencies but I know you can do conditional includes since we do them now:

```
- include: lms.yml
  when: "'lms' in service_variants_enabled"
```

All this does is attach the when clause to all of the included tasks so that they are skipped if it's false (which unfortunately means it still needs to run through all of them).

Why not something like:

```
- service: name=nginx state=stopped
  when: edxapp_no_nginx == True
```

or something similar at the end of the play?

We were thinking about doing away with the nginx role entirely and just including the nginx requirements into the edxapp role itself.  We will also be moving static content from the servers (eventually) and then those nginx tasks can go away.

carsongee:
I must have missed that change in ansible.

So are you planning on getting rid of the role entirely so that every role that has nginx has to install it independently?  That seems a bit redundant since you need nginx for xserver and xqueue as well right? So now you have three roles that have to install nginx, remove the default site, make sure the service is restarted, and then have a notifier to handle config changes instead of one include in each of those three roles.

Just my two cents.  I think it honestly makes it more flexible and less redundant to keep it the way it is, but maybe just add a when clause to the include of nginx-site.yml.

jarv:
After a discussion this morning at standup we are thinking that a parameterized nginx role (similar to what we have now) is what we want except that we don't want to include it from the role even though that would encapsulate the logic.  This means that in the play we would have something like

```
roles:
    - role: nginx
      nginx_servers:
        - lms
        - cms
     - edxapp
     ...
```

And eventually do the same thing with gunicorn:

```
roles:
    - role: nginx
      nginx_servers:
        - lms
        - cms
        - xqueue
     - role: gunicorn
       gunicorn_servers:
         - lms
         - cms
         - xqueue
     - edxapp
     ...
```

How does that sound?
If you want to do a hangout sometime today we can find time to discuss.

-John

carsongee:
Nice. That sounds good.  I don't know how soon I'll be able to get to it, but I'd be happy to work towards getting everything refactored towards that style.

PR #291, opened 2013-09-27
Consolidate requirements.txt into a single file.

feanil:
The only conflict was that fabric was using an older version of paramiko but I verified that the version of fabric we are using is compatible with the newer version of paramiko.


e0d:
+1

jarv:
looks good, may need some wiki updates wherever we reference the old file

feanil:
good call, wiki updated.

PR #292, opened 2013-09-27
Jarv/edxapp mixedmodule

jarv:


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -73,128 +76,151 @@ EDXAPP_LOGGING_ENV:  'sandbox'
...
   <<: *edxapp_generic_env
 lms_xml_auth_config:
   <<: *edxapp_generic_auth
-  'MODULESTORE':
-    'default':
-      'ENGINE': 'xmodule.modulestore.xml.XMLModuleStore'
-      'OPTIONS':
-        'data_dir': '/opt/wwc/data'
-        'default_class': 'xmodule.hidden_module.HiddenDescriptor'
+  MODULESTORE:
```
feanil: We should remove the lms_xml configs since they won't be used anywhere.

feanil:
Once the lms_xml configs are removed I think this is good to merge.

PR #298, opened 2013-10-01
Feanil/notifier updates

feanil:


e0d:
Looks good at a high level, but I'm not sure what the best was to test this is.  Would we expect an ansible role to be a no-op and so indicated by the terminal coloration?

feanil:
I would say that looking at the summary of how many things change would be a good check but yea, there are things that are 'modified' just because the task doesn't know how to verify that nothing changed.  Is there an environment where we can test this?

e0d:
Sure the stage environment is the sensible place, but we should coordinate with the forums team.

feanil:
@e0d I tested this on stage, and it looks good there, can you give it a once over when you get a sec.

e0d:
+1

PR #299, opened 2013-10-01
Added ability for mongo role to do simple replica set clusters

carsongee:
By default this should keep the role the same.  The only catch with this implementation
is that the ansible mongodb_user module doesn't handle slaves particularly well. So if you
have a cluster than any task that creates a user will succeed on the master and fail on all
slaves. As a result the play has been changed to ignore failure at user creation.

### DIFF
--- a/playbooks/roles/mongo/defaults/main.yml
+++ b/playbooks/roles/mongo/defaults/main.yml
@@ -4,6 +4,9 @@ mongo_logappend: true
 mongo_version: 2.4.7
 mongo_bind_ip: 127.0.0.1
 mongo_extra_conf: ''
+mongo_key_file: '/etc/mongodb_key'
+mongo_repl_set: rs0
+mongo_cluster_members: 
 
 MONGO_USERS:
   - user: cs_comments_service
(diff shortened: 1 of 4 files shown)


e0d:
@carsongee Have you been using the 10gen package or the ubuntu package?  We may need to extend the work that you've done here in the near future, but we've been using using the 10gen package recently and were planning to standardize on it.

carsongee:
@e0d I just asked @jarv in relation to the edxlocal role. I'm think I'm fine with moving to the 10gen package.  It should work just the same.  The only concern would be with updates over time since that repo is always the latest and thus subject to possible instabilities so I would prefer we lock it to a specific version in the code (it looks like the repo supports that) that we could update as we go.  I just don't know what that version is at this point.

carsongee:
I rebased this for the changes from @feanil.

feanil:
I will probably be updating this soon so that, we can generate the list of replica set members dynamically like we do for the rabbit cluster but this is a great starting point for that.  I think you are good to go ahead and merge this.

carsongee:
Cool, and yeah there is a lot of room to make it better and less naive about the cluster members.  It would also be nice to fix the ansible module to have slaveOk = true.

PR #300, opened 2013-10-02
rsyslog should log to {{log_bas_dir}}

jrbl:
- /mnt/logs crept into the common recipes, when logs should be written
  to wherever {{log_base_dir}} says

This broke our log rotation for the month it took us to notice, so it would be nice to get this fix into configuration's master.

@feanil and @jbau please.


feanil:
looks good.

PR #303, opened 2013-10-02
Jarv/wheel

jarv:
Add wheel support for python libraries
- Introduces a new standalone role that creates wheel archives from requirements files and uploads them to s3
- edxapp now downloads a wheel archive and installs prior to installation from requirements
- cleans up the base package installs


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -226,7 +226,107 @@ lms_preview_auth_config:
...
+  - python_openid
+  - python_termstyle
+  - pytz
+  - pytz
+  - PyYAML
+  - rednose
+  - requests
+  - scipy
+  - selenium
+  - selenium
```
e0d: dupe

jarv: fixed right as you commented :)

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -291,75 +387,29 @@ sandbox_post_requirements:  "{{ edx_platform_code_dir }}/requirements/edx-sandbo
...
+  # apparmor
+  - apparmor-utils
+  # misc
+  - aspell
+  - curl
+  - ipython
   - npm
   - ntp
-  - openjdk-7-jdk
-  - openjdk-7-jre
```
e0d, line 338: Does this need to be here?  Isn't it going to pull a bunch of other stuff into system python?

aspell?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -89,7 +89,27 @@
...
+
+- name: Untar the wheel archive
+  shell: >
+    tar -xzf {{edxapp_wheel_archive}}.tar.gz 
+      chdir=/var/tmp 
+      creates=/var/tmp/{{edxapp_wheel_archive}}
+    
+- name: Install from wheel files
+  shell: >
+    {{venv_dir}}/bin/pip install --use-wheel --no-index --find-links=/var/tmp/{{edxapp_wheel_archive}} {{item}}
```
feanil, line 121: Why do this instead of setting the wheel parameters for the existing plays?

jarv, line 121: The wheel package list is not created dynamically

playbooks/roles/edxapp/defaults/main.yml
jarv, line 338: removed aspell

jbau:
very interesting, @jarv  Is there anything preventing us (or other instances) from sharing a common wheel file?

jarv:
As long as you are using the Ubuntu precise base image you can.  The wheel files need to be compiled on the same architecture/OS which means we should probably organize them that way in S3.

PR #304, opened 2013-10-02
add a ansible_debug role that just dumps vars into json

jbau:
@jrbl


jbau:
also @jarv and @feanil @e0d as FYI.

jrbl:
I like this.

jarv:
would to_nice_yaml work better for this?

jbau:
how about both!  I'll change that
I think the idea is that you could diff json with the prod version, or yaml with your overrides.

playbooks/roles/ansible_debug/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Dump all vars
+  action: template src=dumpall.j2 dest=/tmp/ansible.all
```
jarv: I would set paranoid permissions on this to be safe.
It might also might be nice to fetch ansible.all for when you are running ansible remotely.

playbooks/roles/ansible_debug/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+- name: Dump all vars to json
+  action: template src=dumpall.json.j2 dest=/tmp/ansible.all.json
+  tags:
+  - dumpall
+  - debug
+
+- name: Dump all vars to yaml
+  action: template src=dumpall.yml.j2 dest=/tmp/ansible.all.yml
+  tags:
+  - dumpall
```
jarv, line 58: I would set paranoid permissions on these

playbooks/roles/ansible_debug/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
...
+  - debug
```
jarv: might be helpful to fetch them from the server so you can see them locally too.

playbooks/roles/ansible_debug/tasks/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+- name: Dump all vars to json
+  action: template src=dumpall.json.j2 dest=/tmp/ansible.all.json
```
jarv: If we add more complexity to the template files for auth/env it might be better to use the existing template files as they exist in the edxapp role (dumped to a different location).  This works great for what we have now though.

playbooks/roles/ansible_debug/templates/dumpall.yml.j2
```diff
@@ -0,0 +1,51 @@
...
+{{ environment | to_nice_yaml }}
+
+GROUP NAMES Variables ("group_names"):
+--------------------------------
+{{ group_names | to_nice_yaml }}
+
+GROUPS Variables ("groups"):
+--------------------------------
+{{ groups | to_nice_yaml }}
+
```
jbau, line 16: HOSTVARS didn't dump nicely to yaml (threw an error), so omitted

jbau:
ok @jarv, added your feature requests.

jarv:
lgtm

PR #305, opened 2013-10-03
remove direct modulestore from preview config

jbau:
lms-preview (I know it's going away soon, hopefully) does not need a direct modulestore.

Having this setting stick around leads in some cases to the default settings (localhost, etc) showing up in our prod config.

@jarv @feanil


jarv:
lgtm

PR #306, opened 2013-10-03
a potporri of edx-west fixes

jbau:
@jrbl this is basically a bunch of small playbook changes in edx-west

the only substantive change is removing persistent-id and targeted-id from being assigned to REMOTE_USER (this is to fix the account cross-wiring)


jrbl:
lgtm :+1:

PR #307, opened 2013-10-04
Symlink /var/lib/jenkins to /mnt/jenkins to avoid running out of disk space

wedaly:
When I started archiving reports, Jenkins ran out of disk space in /var/lib/jenkins.

@feanil


feanil:
@wedaly is /mnt ephemeral storage or is it an EBS volume?  If it's ephemeral then on reboots you will lose everything there.

wedaly:
It's EBS.

feanil:
looks good.

PR #309, opened 2013-10-04
E0d/analytics server

e0d:
Includes the analytics-server role, and ancillary roles for s3fs and automation.  The automated role is include by roles that need to be controlled from the automation server.  What can constrained via rbash and sudo.


playbooks/roles/analytics-server/defaults/main.yml
```diff
@@ -0,0 +1,88 @@
...
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+# Vars for role analytics-server
+# vars are namespace with the module name.
+#
```
jarv, line 13: This is great - did it come from your template generation script?

playbooks/roles/analytics-server/defaults/main.yml
```diff
@@ -0,0 +1,88 @@
...
+DATABASES:
+  default: &databases_default
+    ENGINE: 'django.db.backends.mysql'
+    NAME: 'wwc'
+    USER: 'analytics001'
+    PASSWORD: 'CHANGEME!'
+    HOST: 'CHANGEME'
+    PORT: 3306
+
+analytics_auth_config:
```
jarv, line 32: maybe rename the role to "analytics" and use that as the prefix instead of mixing AS with ANALYTICS?

playbooks/roles/analytics-server/files/etc/sudoers.d/99-automator-analytics-server
```diff
@@ -0,0 +1 @@
+automator  ALL=(www-data) NOPASSWD:SETENV:/opt/wwc/analytics-server/virtualenvs/analytics-server/bin/django-admin.py	run_all_queries	*
```
jarv, line 1: might want to switch this to a template and use `{{ as_web_user }}` and `{{ as_home }}`

playbooks/roles/analytics-server/tasks/deploy.yml
```diff
@@ -0,0 +1,87 @@
+#
+# TODO: Needed while this repo is private
+#
+- name: analytics-server | upload ssh script
+  copy: 
+    src=git_ssh.sh dest=/tmp/git_ssh.sh
```
jarv: we might want to use a role directory for git_ssh.sh for the case where we have multiple roles on the same server.

playbooks/roles/analytics-server/tasks/deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+    notify: analytics-server | restart the analytics service
+  notify: analytics-server | start the analytics service
+  tags:
+    - analytics-server
+    - deploy
+    - install
+    - update
+
+#
+# TODO: Needed while this repo is private
```
jarv, line 42: curious - are there any plans to make this repo public?

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+#
+#
+# Tasks for role analytics-server
+# 
+# Overview:
+#
+# Installs the edX analytics-server Django application which provides
+# basic analytics to the LMS instructor dashboard via service calls.
+#
+# Dependencies:
```
jarv, line 20: does this depend on the common role (I see a reference to it below)?

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+    dest={{ as_home }}/{{ as_env }} 
+    owner="{{ as_user }}" group="{{ as_user }}"
+  tags:
+  - analytics-server
+  - install
+  - update
+
+- name: analytics-server | drop a bash_profile
+  copy: >
+    src=../../common/files/bash_profile
```
jarv, line 68: I think I would rather include the bash_profile in this role unless there is already a common role dependency.

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+- name: analytics-server | drop a bash_profile
+  copy: >
+    src=../../common/files/bash_profile 
+    dest={{ as_home }}/.bash_profile 
+    owner={{ as_user }} 
+    group={{ as_user }}
+
+
+- name: analytics-server | ensure .bashrc exists
+  shell: touch {{ as_home }}/.bashrc
```
jarv: There is a new touch option for the file module in the current version of ansible which will be nice to user here so it doesn't report changed every time if you want to add a TODO. http://www.ansibleworks.com/docs/modules.html#file

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+  - analytics-server
+  - install
+  - update
+
+- name: analytics-server | add source venv to .bashrc
+  lineinfile:
+    dest={{ as_home }}/.bashrc
+    regexp='. {{ as_venv_dir }}/bin/activate'    
+    line='. {{ as_venv_dir }}/bin/activate'
+  tags:
```
jarv, line 107: we should probably have the default bashrc include `.bashrc.d/*` and use that directory for role specific configuration.

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+  - analytics-server
+  - install
+  - update
+
+- name: analytics-server | install global python requirements
+  pip: name={{ item }}
+  with_items: as_pip_pkgs
```
jarv, line 114: I was using the apt s3cmd package for the wheel role, I guess this is better since we get a more recent version

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,124 @@
...
+  tags:
+  - analytics-server
+  - install
+  - update
+
+- name: analytics-server | create config
+  template: 
+    src=opt/wwc/analytics.auth.json.j2
+    dest=/opt/wwc/analytics.auth.json 
+    owner="{{ as_web_user }}" group="{{ as_web_user }}"
```
jarv: I'm not sure what the default perms are but if they aren't 600 we should set them explicitly here.

playbooks/roles/s3fs/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+# - name: my_role | create s3fs mount points
+#   file:
+#     path={{ item.mount_point }} owner={{ item.owner }}
+#     group={{ item.group }} mode={{ item.mode }} state="directory"
+#   with_items: "{{ my_role_s3fs_mounts }}"
+#
+# - name: my_role | mount s3 buckets
+#   mount:
+#     name={{ item.mount_point }} src={{ item.bucket }} fstype=fuse.s3fs
+#     opts=use_cache=/tmp,iam_role={{ task_iam_role }} state=mounted
```
jarv: on the marketing site we needed to use the `allow_other` flag to allow non-root users to  write (and maybe read?)  the filesystem, not sure if it matters here but we may want to at least make it an option.

playbooks/roles/analytics-server/tasks/deploy.yml
e0d, line 42: No specific plans.

playbooks/roles/analytics-server/defaults/main.yml
e0d, line 13: Yeah, I have a couple fixes for it I need to commit, ansible 1.3 related changes and prompt before over-write.

playbooks/roles/analytics-server/tasks/main.yml
e0d, line 114: The reason for using the pip package here is that the latest version supports using IAM roles so we don't have to ship credentials.

playbooks/roles/analytics-server/files/etc/sudoers.d/99-automator-analytics-server
e0d, line 1: Yeah, the reason I didn't do this is because it's passed to the automator role.  Passing a template ref seems like it was more trouble than it was worth.  Simple cases are happy with a file, feel like would need to minimally support both.  Thoughts?

playbooks/roles/analytics-server/tasks/main.yml
e0d, line 107: I think this is a good suggestion, but I'd like to do it as a separate chunk of work.  The template is already in common.

playbooks/roles/analytics-server/files/etc/sudoers.d/99-automator-analytics-server
feanil, line 1: I think as long as we have role name prefixes for variables, passing a template should be fine since the variables won't be overridden. To support both it seems like you would have to pass in a third variable to the automator role to switch between the file rule and the template rule, or just always use the template module.

playbooks/roles/analytics-server/tasks/main.yml
```diff
@@ -0,0 +1,136 @@
...
+#     - "{{ secure_dir }}/vars/common/common.yml"
+#     - "{{ secure_dir }}/vars/stage/analytics-server.yml"
+#     - "{{ secure_dir }}/vars/users.yml"
+#   gather_facts: True
+#   roles:
+#     - common
+#     - analytics-server
+#
+- name: analytics-server | install system packages
+  apt: pkg={{','.join(as_debian_pkgs)}} state=present
```
feanil, line 41: Since the defaults provide a list of debian and redhat packages, we should do this conditionally on being on a debian based system and add a rule for redhat.

playbooks/roles/s3fs/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+#     - "{{ secure_dir }}/vars/common/common.yml"
+#     - "{{ secure_dir }}/vars/users.yml"
+#   gather_facts: True
+#   roles:
+#     - common
+#     - s3fs
+#
+
+- name: s3fs | install system packages
+  apt: pkg={{','.join(s3fs_debian_pkgs)}} state=present
```
feanil, line 59: same as with analytics main.yml since we have both redhat ad debian package lists, we should conditionally install the correct ones based on the OS.

playbooks/roles/analytics/tasks/main.yml
```diff
@@ -0,0 +1,136 @@
...
+  tags:
+  - analytics
+  - install
+  - update
+
+- name: analytics | add source of analytics_env to .bashrc
+  lineinfile:
+    dest={{ analytics_home }}/.bashrc
+    regexp='. {{ analytics_home }}/analytics_env'    
+    line='. {{ analytics_home }}/analytics_server_env'
```
feanil: Should this line also be 'analytics_env' instead of 'analytics_server_env'?

feanil:
I think we should address the installing of packages, for debian or redhat, as part of the boilerplate 'ansible_role' role.  We should go back and update all the roles if we're serious about it.

Just one question about the bashrc sourcing of the analytics_env but otherwise looks good.

PR #310, opened 2013-10-04
Feanil/splunk role

feanil:
@e0d @jarv @jbau @jrbl 

This PR adds a splunk role.  I had to make some permission updates so that various log folders now belong to group 'syslog' instead of 'adm'  not sure if this will break any of the Stanford logging setup.


playbooks/roles/splunkforwarder/defaults/main.yml
```diff
@@ -0,0 +1,48 @@
...
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Vars for role splunk
+# 
+
+#
+# vars are namespace with the module name.
+#
+splunk_role_name: 'splunk'
```
jarv, line 17: nitpick: I think we should put the all-caps variable definitions at the top

playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+# Dependencies:
+#
+# 
+# Example play:
+#
+#
+
+- name: splunkforwarder | stub ansible task
+  debug: msg="This is a stub task created by the ansible-role role"
+  notify: splunk | notify me
```
jarv: this is boilerplate right?

playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+    chdir="/tmp/" 
+    executable="/bin/bash" 
+    creates="/tmp/{{SPLUNKFORWARDER_DEB}}"
+    wget -q {{SPLUNKFORWARDER_PACKAGE_LOCATION}}{{SPLUNKFORWARDER_DEB}}
+  register: download_deb
+
+- name: splunkforwarder | install splunk forwarder
+  shell: gdebi -nq /tmp/{{SPLUNKFORWARDER_DEB}}
+  when: download_deb.changed
+
```
jarv, line 41: It might be better to use the get_url module here. The default behavior is that it will only report as changed if the file doesn't already exist on the file system.

playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+# Create splunk user
+- name: splunkforwarder | create splunk user
+  user: name=splunk group=splunk createhome=no state=present append=yes groups=syslog
+  when: download_deb.changed
+
+- name: splunkforwarder | create boot script
+  shell: creates=/etc/init.d/splunk {{splunkforwarder_output_dir}}/bin/splunk enable boot-start -user splunk --accept-license --answer-yes --no-prompt
```
jarv: nitpick: split this into two lines and put the shell command first

```
shell: >
  <cmd>
    creates=...
```

playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+  shell:
+    chdir="/tmp/" 
+    executable="/bin/bash" 
+    creates="/tmp/{{SPLUNKFORWARDER_DEB}}"
+    wget -q {{SPLUNKFORWARDER_PACKAGE_LOCATION}}{{SPLUNKFORWARDER_DEB}}
+  register: download_deb
+
+- name: splunkforwarder | install splunk forwarder
+  shell: gdebi -nq /tmp/{{SPLUNKFORWARDER_DEB}}
+  when: download_deb.changed
```
jarv, line 40: What I would do here is install the deb when get_url changes, and add a `creates=path/to/file/in/the/package` to the "install splunk forwarder" task and register a var.  Then, for subsequent tasks that should only be run on the first installation, key off of it instead of the download task.

playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -0,0 +1,92 @@
...
+
+- name: splunkforwarder | create outputs config file
+  template:
+    src=opt/splunkforwarder/etc/system/local/outputs.conf.j2
+    dest=/opt/splunkforwarder/etc/system/local/outputs.conf
+    owner=splunk
+    group=splunk
+    mode=644
+
+# Restart Splunk
```
jarv: It might be better to use notifies on the tasks above instead of the restart at the end (so it doesn't restart on every run).  It does however add a bunch of extra lines.

playbooks/roles/splunkforwarder/tasks/main.yml
feanil, line 40: If I put a creates on this command then it won't re-run for newer versions of splunk.

jarv:
lgtm after rebase

PR #312, opened 2013-10-05
Initial working version.

e0d:
Includable role for allowing automation.


playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,163 @@
...
+#
+# That list should be provided via role defaults
+#
+# role_automated_rbash_links:
+#   - /usr/bin/sudo
+#   - /usr/bin/scp
+#
+
+- fail: automated_rbash_links required for role
+  when: automated_rbash_links is not defined
```
jarv, line 55: default 1.3 behavior will error out on undefined variables if that's sufficient, another option is to default it to `!!null` and do a `when: not automated_rbash_links`

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,163 @@
...
+  
+- name: automated | create task user
+  user:
```
jarv, line 61: I think we want a convention to use line continuations `user: > ...` though it's fine in this case it's more explicit to use the line continuation. For example with and without the `>` :

```
foo:
  herp: derp

bar: >
  herp: derp
```

will be loaded as:

```
{'bar': 'herp: derp', 'foo': {'herp': 'derp'}}
```

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,163 @@
...
+# Dependencies:
+#
+# This role depends upon variables provided by the including roles
+# via the role/meta/main.yml.  Includes take the following form.
+#
+#  dependencies:
+#  - {
+#  role: automated,
+#  automated_rbash_links: $as_automated_rbash_links,
+#  automated_sudoers_file: 'roles/analytics-server/files/etc/sudoers.d/99-automator-analytics-server'
```
jarv: If 99-automator-analytics-server is specific to the automated role maybe have it live with the automated in role instead (like we currenlty do for nginx templates) and pass in the calling role name?

playbooks/roles/automated/tasks/main.yml
jarv, line 55: scratch that, behavior is different for vars as role arguments, this is the right thing to do.

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,163 @@
...
+    name={{ automated_user }} state=present shell=/bin/rbash 
+    home={{ automated_home }} createhome=yes
+  tags:
+  - automated
+  - install
+  - update
+
+- name: automated | create sudoers file
+  copy:
+    dest=/etc/sudoers.d/{{ automated_sudoers_file.split('/').pop() }}
```
jarv, line 71: yuck :/ Not sure if there is a better way to do this though if we are passing in the full path to the file.

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,163 @@
...
+
+- name: automated | update shell file mode
+  shell: chmod 640 .bash* .profile
+  tags:
+  - automated
+  - install
+  - update
+
+- name: automated | update shell file ownership
+  shell: chown root.{{ automated_user }} {{ automated_home }}/.bash* {{ automated_home }}/.profile
```
jarv: Why the file mode updates here?

e0d: The rbash'd user needs limited access these files or they can break the containment.

jarv: It would be better to use the file module if it's possible to do this without the glob, otherwise these tasks will report changed every time.

feanil: +1 if this is possible via the file module.

e0d:
@jarv @feanil So, I think that the behavior where nginx has the templates for other roles that use it, is exactly what we should avoid and I think we really can now with dependant roles. 

Also, I still think that passing a file into this role, rather than a template makes senses, despite being a little inconvenient, because templates that generate sudoers files are trouble.  With files, they can be edited using sudoers modes, for example.

feanil:
looks good.

jarv:
looks good

PR #317, opened 2013-10-08
Upgrade git version used on the Jenkins worker

wedaly:
GitHub requires version 1.7.10 or later, but we were using a slightly older version.  I think this may be responsible for the three times I've observed the Jenkins git plugin hanging on `git fetch`.  

By adding the git-core apt-repo, we will use git version 1.8.4.

Reviewer: @feanil


feanil:
looks good.

PR #318, opened 2013-10-08
Make easy to use host names.

feanil:
This change also puts both a jumpbox rule and a private ip address rule
for the jumpbox so that ansible can provision it.


e0d:
+1

PR #319, opened 2013-10-08
Jarv/provision

jarv:


playbooks/library/ec2
```diff
@@ -0,0 +1,679 @@
+#!/usr/bin/python
+# This file is part of Ansible
```
feanil, line 2: If this is a part of ansible why do we have to include it here?  Is it not available until a later version?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -293,77 +293,34 @@ sandbox_post_requirements:  "{{ edx_platform_code_dir }}/requirements/edx-sandbo
...
-  - vagrant
-  - yui-compressor
-  - zip
-  - zlib1g-dev
-
-# Ruby Specific Vars
+  # for shapely
+  - libgeos-dev
+
+# R by Specific Vars
```
feanil: missing 'u' in ruby.

playbooks/roles/gh_users/tasks/main.yml
```diff
@@ -0,0 +1,40 @@
+---
+
+# gh_users
```
feanil, line 3: Can we remove this from the common role once this exists?

jarv, line 3: Yeah, I would prefer to not do it in this pr but we can later.

util/jenkins/ansible-provision.sh
```diff
@@ -0,0 +1,113 @@
...
+EOF
+
+
+cat $extra_vars
+
+
+cd playbooks/edx-east
+# run the tasks to launch an ec2 instance from AMI
+ansible-playbook -vvvv edx_provision.yml  -i inventory.ini -e "@${extra_vars}"  --user ubuntu
+# run tasks to update application config files that
```
feanil: rest of the sentence?

feanil:
@jarv just a few nitpicks but otherwise looks good.  Please make an issue for pulling out the git_hub users content from the common role and then feel free to merge.

PR #321, opened 2013-10-09
E0d/merge problem

e0d:


feanil:
looks good.

playbooks/roles/analytics/tasks/deploy.yml
```diff
@@ -0,0 +1,87 @@
...
+  - install
+  - update
+
+- name: analytics | checkout code
+  git: 
+    dest={{ analytics_code_dir }} repo={{ analytics_source_repo }} 
+    version={{ analytics_version }} force=true
+  environment:
+    GIT_SSH: $analytics_git_ssh
+    notify: analytics | restart the analytics service
```
jarv, line 33: one of the patterns we discussed in the past (and I'm not really sure we are consistent everywhere) was that tasks in `deploy.yml` would stop the service in the beginning, do whatever tasks are necessary to update the code, virtualenv, etc. and then start it at the end.  Tasks in `main.yml` would use the notify handler to restart the service when there were changes.  I'm not sure what is best to do here, I think the same pattern would work though it would force the app to stop and start even if there isn't a change.  Using notifies in `deploy.yml` means that tasks are run while the app is running as well.

playbooks/roles/analytics/templates/tmp/analytics.git_ssh.sh.j2
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+exec /usr/bin/ssh -o StrictHostKeyChecking=no -i {{ analytics_git_identity_dest }} "$@"
```
jarv, line 2: nitpick: we probably make it part of our code conventions if for other roles we want to start creating sub-directories in role/templates, I don't think we are doing this in other places though I could be wrong.

PR #324, opened 2013-10-09
re-apply ec2.py fixes for inifile

sefk:
Our changes was undone by 80b5f3e1edf37676829bed98e5ee7887c79dabb8.  Reapplying.  Pair-programmed with @jrbl, can you pull Joe?

Re-enable set by command-line option and environment file.  We need this at Stanford so we can get private DNS names to use with VPC's.

The only interesting change is moving the option parsing up a bit.


jrbl:
Hooray looks good.

PR #327, opened 2013-10-10
Create the synced folder if it doesn't exist

wedaly:
Otherwise we get an error if the edx-platform folder doesn't already exist.


feanil:
:+1:

PR #328, opened 2013-10-10
Jarv/deploy

jarv:
- Creates new deploy_\* plays that serve the purpose of running the deploy tasks for a single role. 
- Tags the {auth,env}.json files with `deploy`, moves them to `deploy.yml`
- Tags the upstart templates with `deploy`


playbooks/edx-east/deploy_edxapp.yml
```diff
@@ -0,0 +1,6 @@
+- name: Deploy edxapp
+  hosts: all
+  sudo: True
+  gather_facts: True
+  roles:
+    - edxapp
```
jbau, line 6: don't you need common, also?

jarv, line 6: These plays are only for deploys, there aren't any tasks tagged with deploy in the common role so we don't need it.

jbau, line 6: oic, this is strictly meant to be run with `--tags=deploy`.  We currently do that with our existing playbooks (ie. edx-west/prod-app.yml)

how about the nginx role, though?  does it pick those up, because there are tasks tagged 'deploy' there.

playbooks/roles/edxapp/tasks/cms.yml
```diff
@@ -5,19 +5,19 @@
 ---
 - name: create cms application config
   template: src=cms.env.json.j2 dest=$app_base_dir/cms.env.json mode=640 owner=www-data group=adm
-  notify: restart edxapp
```
jbau, line 8: removing the notify because it gets done explicitly, correct?

playbooks/edx-east/deploy_edxapp.yml
jarv, line 6: I think it will since `deploy.yml` includes `{lms,cms,lms-preview}.yml` which then includes `nginx_site.yml` and all the tasks are tagged with `deploy`, I'll need to verify though.

playbooks/roles/edxapp/tasks/cms.yml
jarv, line 8: right, `cms.yml` is included now from `deploy.yml` which stops and starts the services in the beginning and end of `deploy.yml`

playbooks/edx-east/deploy_edxapp.yml
jbau, line 6: yeah, do verify.  I think I've had it not work before because the `nginx` role wasn't included, though I could be misremembering.

Also, are there default variables that are set up in the `common` role that might be needed and now missing?  Though I guess this is a good test of how well our variables are factored

jarv, line 6: ah yeah it does look like we need to add the role, sadly.
Here is the full `--tags deploy` run for edxapp where it appears the nginx tasks are being skipped.
https://gist.github.com/jarv/bbc92f3622d4ada2ee79

feanil:
looks good.

PR #331, opened 2013-10-11
Bugfix/pchiu33/missing xqueue rabbit hostname

pchiu33:
Some fixes to xqueue role default vars that were causing several issues:
- log flooding because of wrong rabbitmq hostname
- log flooding because of DEBUG loglevel


jarv:
@pchiu33 have you signed a contributor agreement yet? http://code.edx.org/individual-contributor-agreement.pdf

pchiu33:
Yes, of course.

Was validated by James Tauber and Jenifer. (just sent you the email with
confirmation)

PR #332, opened 2013-10-11
Jarv/nginx refactor

jarv:


playbooks/roles/nginx/templates/lms.j2
```diff
@@ -23,8 +28,19 @@ server {
     proxy_redirect off;
     proxy_pass http://lms-backend;
   }
+  location / {
```
feanil: Isn't this what the below include of basic-auth.j2 does?

jarv: yup, fixed.

feanil:
looks good.

jbau:
how will this interact with this playbook, e.g.:

https://github.com/edx/configuration/blob/2fd6d4d874cc947c59171e2213d014f5fe21cd00/playbooks/edx-west/carnegie-prod-app.yml

jbau:
I guess what I'm really asking is how do we get both {'nginx_conf': true}
`- {'role': 'nginx', 'nginx_conf': true}`
and the multiple service variants?

feanil:
@jbau I think it would look something like:

```
    - role: nginx
      nginx_conf: true
      nginx_sites:
      - cms
      - lms
      - lms-preview 
```

feanil:
looks good to me.

jarv:
@jbau since it's not a per-server thing feanil's example will work.
I'm curious though, what's in nginx.conf that is secure? Can we add it to the public repo and use variables for the secure bits?

playbooks/roles/nginx/templates/cms-backend.j2
```diff
@@ -1,4 +0,0 @@
-upstream cms-backend {
-    # For a TCP configuration:
-    server 127.0.0.1:8010 fail_timeout=0;
```
carsongee, line 3: Can we add a variable for server address(es) in all the upstreams too, now that this is decoupled from the app server.  I have something like:
    # For a TCP configuration:
    {% for app_server in app_servers %}
    server {{ app_server }}:{{ edxapp_lms_gunicorn_port }} fail_timeout=300;
    {% endfor %}
with a default app_server of 127.0.0.1?

playbooks/roles/nginx/tasks/main.yml
```diff
@@ -11,16 +11,36 @@
...
+  file: src=/etc/nginx/sites-available/edx-release dest=/etc/nginx/sites-enabled/edx-release state=link owner=root group=root
+  notify: nginx | reload nginx
+  tags:
+  - nginx
+
+- name: nginx | Copying nginx configs for {{ nginx_sites }}
+  template: src={{ item }}.j2 dest=/etc/nginx/sites-available/{{ item }} owner=root group=root mode=0600
+  notify: nginx | reload nginx
+  with_items: nginx_sites
+  tags:
```
carsongee, line 35: Can we add back the first_available_file stuff incase I need to override the templates (I do at this point for SSL stuff).

playbooks/roles/nginx/templates/cms-backend.j2
jarv, line 3: Sure, I've added `*_gunicorn_host` which defaults to 127.0.0.1.
See this commit - https://github.com/edx/configuration/commit/c3cfb13b2755dc3bfaefdb8a084e9107f96e2562

playbooks/roles/nginx/tasks/main.yml
jarv, line 35: Previously it was this:

```
- name: nginx | Copying nginx config {{ site_name }}
  template: src={{ item }} dest=/etc/nginx/sites-available/{{ site_name }} owner=root group=root mode=0600
  first_available_file:
  - "{{ local_dir }}/nginx/templates/{{ template_subdir }}/{{ site_name }}.j2"
  - "{{ local_dir }}/nginx/templates/{{ site_name }}.j2"
  # seems like paths in first_available_file must be relative to the playbooks dir
  - "roles/nginx/templates/{{ site_name }}.j2"
```
- Are you sure that these customizations can't be put in the template, using a variable condition?
- Can `local_dir` be replaced with `secure_dir`?

carsongee, line 35: local_dir can (and should) get replaced with secure_dir.  The template could possibly be fixed by just adding a generic extra_conf type variable that would allow me to stuff things in, but with nginx there a lot of places you may want to put in those extra_confs that would be hard to predict with a template so I think it is probably worth allowing the full file override.

jarv:
I see, I guess I was hoping that both the nginx role and the edxapp role (for example) would be able to share the same variable for {lms,cms}_gunicorn_host as the backend address.  If there are multiple backends we would probably need a variable in the nginx role for the backends instead, something like:

```
# nginx/defaults/main.yml
---

nginx_lms_backends:
  - ...
nginx_cms_backends:
  -...
```

Is that what you had in mind?

e0d:
I'm +1 on this change.

I still do worry about implicit cross role dependencies and variable sharing, but I think this is a big improvement over what we currently have.

carsongee:
Yeah, I think that gives you the most architectural freedom since if you want to split nginx and the app servers. If you want to do that they are gunicorn is going to have to listen on 0.0.0.0 and if you have 2 cms backends you can then balance on them using the nginx role.

jarv:
makes sense - added:

```
nginx_xserver_gunicorn_hosts:
  - 127.0.0.1
nginx_xqueue_gunicorn_hosts:
  - 127.0.0.1
nginx_ora_gunicorn_hosts:
  - 127.0.0.1
nginx_lms_gunicorn_hosts:
  - 127.0.0.1
nginx_lms_preview_gunicorn_hosts:
  - 127.0.0.1
nginx_cms_gunicorn_hosts:
  - 127.0.0.1
```

carsongee:
It looks like the first_available got missed, should I just a do a separate PR for that now that this is merged?

jarv:
Ah sorry, too many changes :/
Go ahead and submit a pull for it now that it's in master.

PR #334, opened 2013-10-11
Stanford: correcting prod variables

jrbl:


jrbl:
Trivial changes that only impact Stanford.

PR #337, opened 2013-10-15
Add github.com to ssh known_hosts for jenkins

wedaly:
I may need to switch to using ssh as the transport protocol for `git` with Jenkins.  If so, `github.com` needs to be in the ssh `known_hosts`; otherwise calling `git` on the command line will prompt Jenkins, causing the job to hang.

@e0d what do you think about this approach?

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/system.yml
+++ b/playbooks/roles/jenkins_worker/tasks/system.yml
@@ -44,3 +44,9 @@
 - name: jenkins_worker | Add script to set up environment variables
   template: src=jenkins_env.sh.j2 dest=/usr/local/bin/jenkins_env.sh
             owner=root group=root mode=0555
+
+# Need to add Github to known_hosts to avoid
+# being prompted when using git through ssh
+- name: jenkins_worker | Add github.com to known_hosts if it does not exist
+  shell: >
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -38,3 +38,15 @@
...
+- name: jenkins_worker | Check known_hosts for github.com
+  shell: ssh-keygen -f {{ jenkins_user_home }}/.ssh/known_hosts -H -F github.com | grep 'github.com'
+  register: github_in_known_hosts
+  ignore_errors: yes
+
+# Need to add Github to known_hosts to avoid
+# being prompted when using git through ssh
+- name: jenkins_worker | Add github.com to known_hosts
+  shell: ssh-keyscan -H github.com > {{ jenkins_user_home }}/.ssh/known_hosts
+  when: github_in_known_hosts|failed
```
jarv: hold off on merging this

wedaly: sure thing

jarv:
I don't really see a way to do this without the task reporting changed every time unfortunately unless we write a custom module. Instead of the two commands it can be condensed to the following:

```
- name: jenkins_worker | Add github.com to known_hosts if it doesn't exist
- shell: >
    ssh-keygen -f {{ jenkins_user_home }}/.ssh/known_hosts -H -F github.com | grep -q found || ssh-keyscan -H github.com > {{ jenkins_user_home }}/.ssh/known_hosts
```

wedaly:
@jarv Updated

jarv:
lgtm

PR #339, opened 2013-10-16
Allows override of nginx configuration templates via simple variable

carsongee:
The only downside I can see is that you have to override for all sites, but I don't really see that as a major problem.

### DIFF
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -29,6 +29,8 @@ nginx_debian_pkgs:
   - nginx
   - python-passlib
 
+nginx_template_dir: "edx/app/nginx/sites-available"
+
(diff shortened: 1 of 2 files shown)


jarv:
I think I'm ok with this and it's better than using try_files.
@jbau @feanil @e0d @jrbl  ok with you?

feanil:
seems good.

e0d:
I think this is going to be a use case that we have to support in multiple places.  Do we just want to have a convention for providing overrides?  I fine with overrides in general and think it's highly preferable to conditionals in the templates.

hkim823:
@carsongee is this still a valid PR?  Let us know

carsongee:
I'm still using it since my templates are different enough from stock.  I'll rebase it to get rid of merge conflicts.

hkim823:
:+1:

PR #340, opened 2013-10-16
adding simple play for role creation

e0d:


playbooks/ansible_role.yml
```diff
@@ -0,0 +1,6 @@
+- name: Configure instance(s)
```
feanil, line 1: Can you add a comment on how to pass in the variable for the role name and a sample ansible command.

jarv, line 1: it's handy to run plays like this without an inventory script. The way I have been running plays run on localhost is setting:
`hosts: all` in the play and then:

`ansible-playbook -c local -i "localhost," ansible_role.yml`

jarv:
completely forgot that there was already an open pull for this, I think this can be closed since I just merged in a play that does this.

jarv:
closing

PR #343, opened 2013-10-16
Bumping vagrant boxes RAM

carsongee:
I ran into failures to compile/install platform requirements and bumping the
RAM fixed them.


jarv:
lgtm

PR #345, opened 2013-10-19
Feanil/update integration

feanil:
@jarv @e0d  These changes should fix the continuous integration environment and the ansible sandboxes.  I updated jenkins to point to this branch for now so it should be changed back to master wen we merge this in.


jarv:
@feanil looks good
@jbau @jrbl I would really like to simplify our migration check here so that we have a default var 

`edxapp_migrate_db: False` in `edxapp/defaults/main.yml` and setting it to boolean true or anything will run migration, meaning that the when clause is simplified to `when: edxapp_migrate_db`

I would also like to remove the separate "migrate only" task unless there is a reason for it, thoughts?

PR #348, opened 2013-10-21
Will/jenkins custom git plugin

wedaly:
- Add forked git-plugin repo (resolves an issue with duplicate builds)
- Refactor plugin installation to install updated versions.
- Stop running tests after compiling forked plugins.
- Update plugin versions to the ones we're using in Jenkins now.

### DIFF
--- a/playbooks/roles/jenkins_master/defaults/main.yml
+++ b/playbooks/roles/jenkins_master/defaults/main.yml
@@ -13,7 +13,6 @@ jenkins_plugins:
     - { name: "credentials", version: "1.8.3" }
     - { name: "dashboard-view", version: "2.9.1" }
     - { name: "ec2", version: "1.19" }
-    - { name: "git", version: "1.5.0" }
     - { name: "github", version: "1.8" }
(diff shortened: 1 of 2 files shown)


feanil:
I don't like how we have to build the custom plugins every time, is there any check we can do to not run the build if that version of the plugin has already been built?

wedaly:
@feanil I don't know a clean way to do that without giving up the guarantee that the plugins get updated when we increment the git revision.

The compilation is pretty fast -- the git plugin, for instance, takes about 7.4 seconds.

Any suggestions for how to do this without adding a lot of complexity?

feanil:
@wedaly is it possible for you to specify the output name for the plugin file created by maven?  It would be nice to output a file with the version as part of the name and then add 'creates' to the maven command so that it doesn't run if that version has already been built.  If that is too much work, I think the current version looks good and you can go ahead an merge it.

wedaly:
I don't see a way to do that in maven without modifying the pom.xml file in the plugin repo.  That's more work than I want to do for this fix :)

PR #349, opened 2013-10-21
Jarv/common refactor

jarv:
# Summary
- The entire application is installed under a single directory, for
  now the default is /edx but it's easy to change.
- edxapp and forums now run in supervisor

```
/edx/bin/supervisorctl -c /edx/etc/supervisord.conf 
edxapp:cms                       RUNNING    pid 16884, uptime 0:04:16
edxapp:lms                       RUNNING    pid 16883, uptime 0:04:16
edxapp_worker:cms_default_4      RUNNING    pid 16902, uptime 0:04:16
edxapp_worker:cms_high_1         RUNNING    pid 16891, uptime 0:04:16
edxapp_worker:cms_low_3          RUNNING    pid 16886, uptime 0:04:16
edxapp_worker:lms_default_3      RUNNING    pid 16887, uptime 0:04:16
edxapp_worker:lms_high_4         RUNNING    pid 16907, uptime 0:04:16
edxapp_worker:lms_low_1          RUNNING    pid 16895, uptime 0:04:16
forum                            RUNNING    pid 16885, uptime 0:04:16
```
- Strips the common role of tasks that are no longer necessary
  (touching log files, changing permissions, etc)
- Takes the virtualenv creation out of the common role as well as
  other jobs that are now in edxapp
- Makes the edxapp role conform to variable naming conventions
- gunicorn and nginx run as www-data but the rbenv, virtualenv,
  staticfiles dir and code repo is now owned by the "edxapp" user.  This
  is important for security and should be the convention moving forward
  for other roles.
- This pull aggressively removes extra tags, for now I am assuming we don't want role tags on every task in a role.
- Application files and application configuration - `/edx/app/<role>`
- Application data - `/edx/var/<role>`
- Application logging - `/edx/var/log/<role>`
- Application python virtual environments - `/edx/app/<role>/venvs/<name>`
- Application ruby environment - `/edx/app/<role>/{.rbenv,.gem}`
- `/edx/bin` has symlinks as a convenience for running executables in the app/ dir, for now this directory is flat without any subdirectories
- `/edx/etc` has symlinks to configuration directories for convenience, for now this directory is flat without any subdirectories
- nginx configuration `{sites_enabled, sites_available}` has been relocated to `/edx/app/nginx`
# TODOs

These TODOs have been put on the backlog
- the user that the forum runs as (www-data) needs write access to its .gem/ directory, for this reason its rbenv is owned by www-data which should be fixed
- applications keep their configuration in the /app directory, this is necessary for now until we go and update repos not to use relative paths to load the json files.
- xqueue, xserver, ora are still running under upstart, they should be moved to supervisor after this pull is merged and checked out
- need to discuss whether other services, nginx, rabbit, mongo, etc. should be moved to supervisor


playbooks/roles/common/tasks/main.yml
```diff
@@ -1,93 +1,57 @@
...
   apt: pkg={{','.join(common_debian_pkgs)}} install_recommends=yes state=present update_cache=yes
   tags:
   - pre_install
   - update
-
-- name: common | Create log directory
-  file: path={{log_base_dir}} state=directory mode=2755 group=adm owner=syslog
-  tags:
-  - pre_install
   - update
```
feanil: Looks like you have this tagged with update twice.

playbooks/roles/common/tasks/main.yml
```diff
@@ -1,93 +1,57 @@
...
-  command: touch -a {{log_base_dir}}/edx.log creates={{log_base_dir}}/edx.log
-  tags:
-  - pre_install
-  - logging
-  - install
 
-- name: common | Set permissions on edx log file
-  # This is done for the benefit of the rake commands, which expect it
-  file: path={{log_base_dir}}/edx.log owner=syslog group=adm mode=640
+- name: common | Install logrotate configuration for tracking file
```
feanil: Shouldn't this be in the edxapp role? or are all services meant to generate tracking logs?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -292,22 +314,23 @@ edxapp_theme_version: 'HEAD'
...
-pre_requirements_file:    "{{ edx_platform_code_dir }}/requirements/edx/pre.txt"
-post_requirements_file:   "{{ edx_platform_code_dir }}/requirements/edx/post.txt"
-base_requirements_file:   "{{ edx_platform_code_dir }}/requirements/edx/base.txt"
-github_requirements_file: "{{ edx_platform_code_dir }}/requirements/edx/github.txt"
-repo_requirements_file:   "{{ edx_platform_code_dir }}/requirements/edx/repo.txt"
-
-sandbox_base_requirements:  "{{ edx_platform_code_dir }}/requirements/edx-sandbox/base.txt"
-sandbox_local_requirements: "{{ edx_platform_code_dir }}/requirements/edx-sandbox/local.txt"
-sandbox_post_requirements:  "{{ edx_platform_code_dir }}/requirements/edx-sandbox/post.txt"
+edx_platform_commit: 'jarv/static-root'
```
feanil: This should be set back to release before this is merged in.

playbooks/roles/edxapp/tasks/cms.yml
```diff
@@ -3,28 +3,32 @@
...
-  - cms
-  - update
   - deploy
 
 - name: Create CMS log target directory
-  file: path={{log_base_dir}}/cms state=directory owner=syslog group=syslog mode=2750
+  file: >
+    path={{ edxapp_log_dir }}/cms
+    state=directory
+    owner={{ common_log_user }}
```
feanil: It would be good to set group to syslog as well so that admin users can be a part of that group and view the logs without having to sudo as syslog.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -25,90 +25,61 @@
 
 
 # Do A Checkout
-- name: edxapp | checkout edx-platform repo into {{edx_platform_code_dir}}
-  git: dest={{edx_platform_code_dir}} repo={{edx_platform_repo}} version={{edx_platform_commit}}
+- name: edxapp | checkout edx-platform repo into {{edxapp_code_dir}}
+  git: dest={{edxapp_code_dir}} repo={{edx_platform_repo}} version={{edx_platform_commit}}
   register: edx_platform_checkout
```
feanil, line 33: It would be good to use this variable to only run the ruby, pip and npm tasks when the checkout has changed.  Not sure if that is easy to do I'm happy to have this be a future task if that makes more sense.

playbooks/roles/edxapp/tasks/lms-preview.yml
```diff
@@ -3,26 +3,32 @@
...
-  - lms-preview
-  - lms-preview-env
   - deploy
 
 - name: Create lms-preview log target directory
-  file: path={{log_base_dir}}/lms-preview state=directory owner=syslog group=syslog mode=2750
+  file: >
+    path={{ edxapp_log_dir }}/lms-preview
+    state=directory
+    owner={{ common_log_user }}
```
feanil: Same here, can we set the group to the common_log_user as well so that admins can read logs without having to sudo as syslog.

playbooks/roles/edxapp/tasks/lms.yml
```diff
@@ -1,27 +1,32 @@
...
-  - lms-env
-  - update
   - deploy
 
 - name: Create lms log target directory
-  file: path={{log_base_dir}}/lms state=directory owner=syslog group=syslog mode=2750
+  file: >
+    path={{ edxapp_log_dir }}/lms
+    state=directory
+    owner=syslog
```
feanil: Owner should be set to `common_log_user`

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -2,54 +2,57 @@
...
+    group="{{ common_web_group }}"
+  with_items:
+    - "{{ edxapp_app_dir }}"
+    - "{{ edxapp_venvs_dir }}"
+
+- name: edxapp | create edxapp log dir
+  file: >
+    path="{{ edxapp_log_dir }}"
+    state=directory
+    owner="{{ common_log_user }}"
```
feanil: set group to common_log_user as well.

playbooks/roles/common/tasks/main.yml
jarv: fixed - removed all tags in the common role

playbooks/roles/common/tasks/main.yml
jarv: I'm leaving the following tasks:

```
- name: common | Install rsyslog configuration for edX
- name: common | Install logrotate configuration for edX
```

Since they are system wide configuration though they do contain some specific directives for edxapp (like tracking logs)

I moved the following task:

```
- name: common | Install logrotate configuration for tracking file
```

To the edxapp role as well as the template since that is clearly edxapp specific.

carsongee:
Just a couple general comments.  It would be nice if we maybe keep a wiki page of variable changes like what is introduced here, or any of the other refactors since it breaks backwards compatibility.  I also can't say I'm a huge fan of putting the app on the devices root.  I still prefer something like /opt/edx/{var|app|...} as the default so that it allows easier relocation of the platform on the system with mounting and doesn't mess with file system hierarchy standards.

jarv:
wrt /edx vs /opt/edx :
We ended up on on /edx (after some debate) because /opt/edx/opt was weird though that would be now /opt/edx/app.  
Another reason is that we definitely want everything under one directory and everything on one ebs volume for snapshots and so that we can move /edx to a different server easily.  Putting a mounted volume under /opt is a little weird for me (though not unheard of).

jarv:
I agree with the wiki page, not sure yet about migrations or whether it is worth the effort to support it.
Let me try to get my head around what that would take first.

carsongee:
On the path thing, I can always change my default, so it isn't a big deal if you want to keep it that way.  That said I am totally in agreement on the single path and I am happy with the structure change, it is just that if I decide to end up putting the edx folder on shared storage for example so I only have to manage one install on many machines, putting the app in the root of an NFS or gluster volume for example is generally a bad practice for security and flexibility since you may want to use that volume for other purposes, and it would be bad to have to intermix that other content in with the application.  My two cents.

jbau:
no worries about the tags, i think you moved the tagged tasks we'd use into the "deploy" tag anyway.

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -22,21 +19,25 @@
...
 
-- name: mongo | move mongodb to {{ storage_base_dir }}
-  command: mv /var/lib/mongodb  {{ storage_base_dir}}/.  creates={{ storage_base_dir }}/mongodb
-  tags: mongo
+- name: mongo | move mongodb to {{ mongo_data_dir }}
+  command: mv /var/lib/mongodb  {{ mongo_data_dir}}/.  creates={{ mongo_data_dir }}/mongodb
 
-- name: mongo | create mongodb symlink
-  file: src={{ storage_base_dir }}/mongodb dest=/var/lib/mongodb state=link
-  tags: mongo
```
carsongee, line 39: This gets rid of the link, but there isn't a change in the mongodb.conf.j2 template to point to the data dir that I can see.  So either the link should come back from /var/lib/mongodb to {{ mongo_data_dir }}/mongodb or update the conf file to use the variable.

Also, it looks like mongo_app_dir and mongo_dbpath are created but never used anywhere...what purpose are those serving?

jarv, line 39: looks like bad conflict resolution.
In https://github.com/edx/configuration/blob/jarv/common-refactor/playbooks/roles/mongo/defaults/main.yml mongo_dbpath is set twice. I'll update it so that it is `mongo_dbpath: "{{ mongo_data_dir }}/mongodb"`

I don't think we need the symlink since in mongodb.conf.j2 we have `dbpath={{ mongo_dbpath }}`

mongo_app_dir isn't used anywhere so it can be removed, I was setting the boilerplate directories (app, data) for every role.

carsongee, line 39: Good call. I forgot I added that and forgot to remove the symlink...now I've forgotten twice.  whoops.

playbooks/roles/common/tasks/main.yml
```diff
@@ -1,93 +1,43 @@
...
-- include: create_github_users.yml
-  when: GITHUB_USERS is defined
-
 - name: common | Add user www-data
-  # This user should be created on the system by default
-  user: name=www-data
-  tags:
-  - pre_install
-  - update
+  # This is the default user for nginx
```
e0d, line 3: Not sure this is common. Would we want this on a box of role jenkins for example.

jarv, line 3: I am game for renaming this user to something else, right now `{{ common_web_user }}` is not only the user for nginx but the user for supervisord and every application that runs under supervisord.
I think the common role should own it since now the common role is very limited in scope and should be included for a standalone jenkins server.

playbooks/roles/discern/defaults/main.yml
```diff
@@ -1,15 +1,112 @@
...
+  - build-essential
+  - python-dev
+  - gfortran
+  - libfreetype6-dev
+  - libpng12-dev
+  - libxml2-dev
+  - libxslt1-dev
+  - libreadline6
+  - libreadline6-dev
+  - redis-server
```
e0d, line 86: @VikParuchuri is redis-server really a requirement?

jbau, line 86: It definitely is for ora.  So I'm assuming it's the same for discern.

playbooks/roles/discern/tasks/main.yml
```diff
@@ -1,111 +1,55 @@
...
-  template: src=celery.conf.j2 dest=/etc/init/celery.conf owner=root group=edx mode=0664
+  template: >
+    src=celery.conf.j2 dest=/etc/init/celery.conf
+    owner=root group=root
   notify: discern | restart celery
 
 - name: discern | render discern service from template
-  template: src=discern.conf.j2 dest=/etc/init/discern.conf owner=root group=edx mode=0664
+  template: >
+    src=discern.conf.j2 dest=/etc/init/discern.conf
```
e0d, line 36: Can we standardize on templates/os/path/to/file/discern.conf.j2?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -1,114 +1,90 @@
...
-  - npm
-  - update
   - deploy
 
 # Python plays that need to be run after platform updates.
 
 # Substitute github mirror in all requirements files
 #
-- name: Updating requirement files for git mirror
+- name: edxapp | Updating requirement files for git mirror
```
e0d, line 87: Shouldn't this be conditional?

playbooks/roles/discern/tasks/main.yml
jarv, line 36: I'll add it to the backlog unless you think it needs to be in this pull, should be self contained enough that we can do it across all of the roles for all templates.

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -2,54 +2,61 @@
...
+  with_items:
+    - "{{ edxapp_staticfile_dir }}"
+    - "{{ edxapp_theme_dir }}"
+
+
+- name: edxapp | create web-writable edxapp data dirs
+  file: >
+    path="{{ item }}" state=directory
+    owner="{{ common_web_user }}" group="{{ edxapp_user }}"
+    mode="0775"
```
e0d, line 41: Does other need any permission here?

jarv, line 41: it doesn't but I think we are going to want to tackle other permissions in a single shot as a role or a set of tasks that are run after deploy.  I'll add a backlog item for it.  For example our git checkouts have other-read enabled because we aren't setting a umask and there are some changes (like group permissions) that we will probably need to set as root in main.yml

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -0,0 +1,81 @@
...
+    chdir={{ edxapp_code_dir }}
+  sudo_user: "{{ edxapp_user }}"
+  when: celery_worker is not defined
+  with_items: service_variants_enabled
+  environment: "{{ edxapp_deploy_environment }}"
+  tags:
+  - deploy
+
+
+- name: edxapp | syncdb and migrate
```
e0d, line 69: There's at least on cms specific migration that needs to be handled.

playbooks/roles/edxapp/templates/cms.conf.j2
```diff
@@ -1,28 +1,12 @@
...
-env PID=/var/tmp/cms.pid
-#env NEW_RELIC_CONFIG_FILE={{app_base_dir}}/newrelic.ini
-#env NEWRELIC={{venv_dir}}/bin/newrelic-admin
+[program:cms]
 {% if ansible_processor|length > 0 %}
-env WORKERS={{ ansible_processor|length * worker_core_mult.cms }}
+command={{ edxapp_venv_dir }}/bin/gunicorn --preload -b 127.0.0.1:{{ edxapp_cms_gunicorn_port }} -w {{ ansible_processor|length * worker_core_mult.cms }} --timeout=300 --pythonpath={{ edxapp_code_dir }} cms.wsgi
 {% else %}
-env WORKERS={{ worker_core_mult.cms }}
+command={{ edxapp_venv_dir }}/bin/gunicorn --preload -b 127.0.0.1:{{ edxapp_cms_gunicorn_port }} -w {{ worker_core_mult.cms }} --timeout=300 --pythonpath={{ edxapp_code_dir }} cms.wsgi
```
e0d: shouldn't the bind host be a var?

playbooks/roles/edxapp/templates/lms.conf.j2
```diff
@@ -1,37 +1,12 @@
...
-respawn
-respawn limit 3 30
-
-env PID=/var/tmp/lms.pid
-#env NEW_RELIC_CONFIG_FILE={{app_base_dir}}/newrelic.ini
-#env NEWRELIC={{venv_dir}}/bin/newrelic-admin
+[program:lms]
 {% if ansible_processor|length > 0 %}
-env WORKERS={{ ansible_processor|length * worker_core_mult.lms }}
+command={{ edxapp_venv_dir }}/bin/gunicorn --preload -b 127.0.0.1:{{ edxapp_lms_gunicorn_port }} -w {{ ansible_processor|length * worker_core_mult.lms }} --timeout=300 --pythonpath={{ edxapp_code_dir }} lms.wsgi
```
e0d: same questions as above.

playbooks/roles/edxapp/templates/lms.conf.j2
```diff
@@ -1,37 +1,12 @@
...
-    if $(curl -s -i localhost:$PORT/heartbeat | egrep -q '200 OK'); then
-      break;
-    else
-      sleep 1;
-    fi
-  done
-end script
+user={{ common_web_user }}
+directory={{ edxapp_code_dir }}
+environment=PORT={{edxapp_lms_gunicorn_port}},ADDRESS={{edxapp_lms_gunicorn_host}},LANG=en_US.UTF-8,DJANGO_SETTINGS_MODULE={{ edxapp_lms_env }},SERVICE_VARIANT="lms"
```
e0d: I think that LANG should be a param.

playbooks/roles/edxapp/templates/cms.conf.j2
jarv: yes, fixed.

playbooks/roles/edxapp/templates/lms.conf.j2
jarv: fixed

playbooks/roles/edxapp/tasks/service_variant_config.yml
jarv, line 69: I'm going to add this to the backlog since we weren't doing this in master either.

playbooks/roles/edxapp/tasks/deploy.yml
jarv, line 87: leaving this as-is after discussing with Ed offline.

playbooks/roles/discern/defaults/main.yml
```diff
@@ -1,15 +1,112 @@
...
 discern_settings: discern.aws
-nltk_data_dir: /usr/share/nltk_data
-ease_branch: master
 discern_branch: dev
 discern_gunicorn_port: 8070
 discern_gunicorn_host: 127.0.0.1
-discern_user: discern
-site_name: discern
+
+discern_env_config:
```
feanil, line 41: These should probably have top level all caps variables to update them just like with the auth settings.

jarv, line 41: The discern and ora roles need to be brought inline with our variable conventions in general, I will add a backlog item.

playbooks/roles/discern/tasks/main.yml
```diff
@@ -1,111 +1,55 @@
...
-    - collectstatic
+- include: deploy.yml
 
-#Have this separate from the other three because it doesn't take the noinput flag
-- name: discern | django update_index for discern
-  shell: ${venv_dir}/bin/python {{discern_dir}}/manage.py update_index --settings={{discern_settings}} --pythonpath={{discern_dir}}
+- name: discern | create a symlink for venv python
+  file: >
+    src="{{ discern_venv_bin }}/python"
+    dest={{ bin_dir }}/python.discern
```
feanil: what uses this link?  Does some part of discern require it?

jarv: each role creates an `/edx/bin/python.<role>` link for convenience, not used for anything else.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -285,29 +343,30 @@ worker_core_mult:
...
-pre_requirements_file:    "{{ edx_platform_code_dir }}/requirements/edx/pre.txt"
-post_requirements_file:   "{{ edx_platform_code_dir }}/requirements/edx/post.txt"
-base_requirements_file:   "{{ edx_platform_code_dir }}/requirements/edx/base.txt"
-github_requirements_file: "{{ edx_platform_code_dir }}/requirements/edx/github.txt"
-repo_requirements_file:   "{{ edx_platform_code_dir }}/requirements/edx/repo.txt"
-
-sandbox_base_requirements:  "{{ edx_platform_code_dir }}/requirements/edx-sandbox/base.txt"
-sandbox_local_requirements: "{{ edx_platform_code_dir }}/requirements/edx-sandbox/local.txt"
-sandbox_post_requirements:  "{{ edx_platform_code_dir }}/requirements/edx-sandbox/post.txt"
+edx_platform_commit: 'HEAD'
```
feanil: Is there a reason this is pointing to HEAD instead of 'release'?

jarv: it should be release, fixed.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -120,161 +96,107 @@
...
-  service: name=edxapp state=restarted
+- name: edxapp | restart the edxapp services (supervisor)
+  supervisorctl: >
+    name="edxapp:{{ item }}"
+    supervisorctl_path={{ supervisor_ctl }}
+    config={{ supervisor_cfg }}
+    state=restarted
   when: celery_worker is not defined
+  with_items: service_variants_enabled
+  ignore_errors: true
```
feanil: Why are we ignoring errors on this?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -120,161 +96,107 @@
...
-  service: name=edx-workers state=restarted
+- name: edxapp | restart the celery worker services (supervisor)
+  supervisorctl: >
+    name="edxapp_worker:{{ item.service_variant }}_{{ item.queue }}_{{ item.concurrency }}"
+    supervisorctl_path={{ supervisor_ctl }}
+    config={{ supervisor_cfg }}
+    state=restarted
   when: celery_worker is defined
+  with_items: edxapp_workers
+  ignore_errors: true
```
feanil: same as above.

e0d:
:+1:

playbooks/roles/edxapp/templates/edx-worker-cms.conf.j2
```diff
@@ -10,15 +10,15 @@ respawn
 instance edx.${SERVICE_VARIANT}.core.${QUEUE}
```
feanil: Shouldn't this file be deleted since we have supervisor configurations for workers?

playbooks/roles/edxapp/templates/edx-worker-lms.conf.j2
```diff
@@ -10,15 +10,15 @@ respawn
 instance edx.${SERVICE_VARIANT}.core.${QUEUE}
```
feanil: same question as above.

playbooks/roles/nginx/templates/edx_logrotate_nginx_access.j2
```diff
@@ -1,6 +1,6 @@
 # Put in place by ansible
 
-{{log_base_dir}}/nginx/access.log {
+{{ nginx_log_dir }}/nginx/access.log {
```
feanil: don't need the nginx log directory since {{ nginx_log_dir }} already has this as part of it's path.

playbooks/roles/nginx/templates/edx_logrotate_nginx_error.j2
```diff
@@ -1,6 +1,6 @@
 # Put in place by ansible
 
-{{log_base_dir}}/nginx/error.log {
+{{ nginx_log_dir }}/nginx/error.log {
```
feanil: same as above.

feanil:
looks good after you update the nginx logrotate templates. :+1:

PR #354, opened 2013-10-22
Fix Jenkins version to last known stable.

wedaly:
Apparently the latest version of Jenkins has a show-stopping bug: https://issues.jenkins-ci.org/browse/JENKINS-20123

This change ensures that we install the same version we're using in the Jenkins cluster, which does not have this problem.

@jzoldak @feanil

### DIFF
--- a/playbooks/roles/jenkins_master/defaults/main.yml
+++ b/playbooks/roles/jenkins_master/defaults/main.yml
@@ -4,6 +4,9 @@ jenkins_group: "edx"
 jenkins_server_name: "jenkins.testeng.edx.org"
 jenkins_port: 8080
 
+jenkins_deb_url: "http://pkg.jenkins-ci.org/debian/binary/jenkins_1.530_all.deb"
+jenkins_deb: "jenkins_1.530_all.deb"
+
 jenkins_plugins:
     - { name: "build-name-setter", version: "1.3" }
     - { name: "build-pipeline-plugin", version: "1.4" }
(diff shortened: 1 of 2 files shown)


playbooks/roles/jenkins_master/defaults/main.yml
```diff
@@ -53,4 +56,5 @@ jenkins_debian_pkgs:
     - nginx
     - git
     - maven
+    - daemon
```
feanil, line 59: what does this package do?

feanil:
Instead of grabbing the file, I would continue to use the jenkins apt repo and just lock the correct version
`apt: pkg=jenkins={{jenkins_version}}`  where `jenkins_version` is set to the appropriate version.

jzoldak:
:thumbsup: tested the changes locally by cherry-picking against another branch

playbooks/roles/jenkins_master/defaults/main.yml
wedaly, line 59: The Jenkins debian package installs some scripts into /etc/init.d that depend on daemon.

wedaly:
@feanil As far as I can tell, they don't provide old versions in the apt repo.  I'm following the recommendations they provide here for installing older versions:

http://pkg.jenkins-ci.org/debian/

feanil:
Ah I see, looking at the sources list, it does not provide that version, really weird.  It provides 1.528, and 1.536 but nothing in between.

```
1.536 - 
1.528 - 
1.424.6+dfsg-1ubuntu0.2 - 
1.424.6+dfsg-1ubuntu0.1 - 
1.424.6+dfsg-1 - 
```

Change looks good to merge.

PR #357, opened 2013-10-23
Create "pinned" files to allow upgrading plugins bundled with Jenkins.

wedaly:
I noticed that some (but not all) of the Jenkins plugins were not being upgraded correctly by the Ansible scripts.

According to https://issues.jenkins-ci.org/browse/JENKINS-13129
- Jenkins "pins" built-in plugins.  From the Jenkins docs:
  
  Normally whenever Jenkins is upgraded or downgraded, its built-in plugins overwrite whatever versions of the plugins are currently installed in $JENKINS_HOME. This ensures that consistent versions of those plugins are used. However, this behavior also means that those plugins cannot be manually updated, as every time Jenkins is started they will be replaced by the bundled versions.
- The workaround is to create a ".pinned" file in the plugins directory for these built-in plugins.

@feanil

### DIFF
--- a/playbooks/roles/jenkins_master/defaults/main.yml
+++ b/playbooks/roles/jenkins_master/defaults/main.yml
@@ -41,6 +41,12 @@ jenkins_plugins:
     - { name: "multiple-scms", version: "0.2" }
     - { name: "timestamper", version: "1.5.7" }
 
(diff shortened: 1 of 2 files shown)


jarv:
lgtm

PR #359, opened 2013-10-23
openid workaround needs to happen before the syncdb/migrate

jarv:
this is due to the ordering change in https://github.com/edx/configuration/commit/73141bed534abf76532461b5d63794438f10950f


feanil:
looks good.

PR #361, opened 2013-10-23
Feanil/automate stack creation

feanil:


util/vpc-tools/create_stack.py
```diff
@@ -0,0 +1,94 @@
...
+    template_url = upload_file(template, temp_bucket, key_name)
+
+    # Reference the stack.
+    try:
+        stack_id = cfn.create_stack(stack_name,
+            template_url=template_url,
+            capabilities=['CAPABILITY_IAM'],
+            tags={'autostack':'true'},
+            parameters=[('KeyName', 'continuous-integration')])
+    except Exception as e:
```
jarv, line 48: Why catch all exceptions if you are just printing it out and raising it anyway?

feanil, line 48: For a lot of the boto exceptions, the message is more useful than the repr but it doesn't get printed out by default.

util/vpc-tools/create_stack.py
```diff
@@ -0,0 +1,94 @@
...
+    status = None 
+    while blocking:
+        sleep(5)
+        stack_instance = cfn.describe_stacks(stack_id)[0]
+        status = stack_instance.stack_status
+        print(status)
+        if 'COMPLETE' in status:
+            break
+
+    if status in FAILURE_STATES:
```
jarv, line 62: any reason why this isn't `if "COMPLETE" not in status:`?  Or more specifically what is a successful status?

feanil, line 62: I went with the failure states because I felt that 'ROLLBACK_COMPLETE' and 'DELETE_COMPLETE' were failure states.  if it makes more sense I can do not in success states but I think that will just be an inverse of the state list I have right now.

util/vpc-tools/vpc_dns.py
```diff
@@ -0,0 +1,126 @@
...
+    zone_name = zone.Name[:-1]
+    dns_name = dns_template.format(prefix=prefix,
+                                   zone_name=zone_name)
+
+    add_or_update_record(zone, dns_name, 'CNAME', 600, [elb.dns_name])
+
+
+if __name__ == "__main__":
+    description = "Give a cloudformation stack name, for an edx stack, setup \
+        DNS names for the ELBs in the stack."
```
jarv, line 95: Similar to what we do for looking up intances by tag with this module https://github.com/edx/configuration/blob/master/playbooks/library/ec2_lookup we could also filter and return a list of ELBs, after that is registered you could then feed it into route53.  Probably not worth the effort now though if this is already working.

util/vpc-tools/create_stack.py
jarv, line 48: huh that's silly behavior, can you add a comment indicating such since it strikes me as a little odd.

util/vpc-tools/create_stack.py
jarv, line 62: yeah that's fine, just concerned that it might fall through here on a failure state that we aren't catching.

jarv:
lgtm

PR #364, opened 2013-10-28
analytics vpc cloud formation

mulby:
@rocha (FYI)

Reviewers: @e0d, @jarv

### DIFF
--- a/cloudformation_templates/edx-analytics-vpc.json
+++ b/cloudformation_templates/edx-analytics-vpc.json
@@ -1,0 +1,289 @@
+{
+  "AWSTemplateFormatVersion" : "2010-09-09",
+
+  "Description" : "A simple VPC with a public subnet, a private subnet, a NAT and a Bastion host.",
+
+  "Parameters" : {
+
+    "KeyName" : {
+      "Description" : "Name of an existing EC2 KeyPair to enable SSH access to the instances",
+      "Type" : "String",
+      "MinLength": "1",
+      "MaxLength": "64",
+      "AllowedPattern" : "[-_ a-zA-Z0-9]*",
+      "ConstraintDescription" : "can contain only alphanumeric characters, spaces, dashes and underscores."
+    },
+
+    "SSHLocation" : {
+      "Description" : "Lockdown SSH access to the bastion host (default can be accessed from anywhere)",
+      "Type" : "String",
+      "MinLength": "9",
+      "MaxLength": "18",
+      "Default" : "0.0.0.0/0",
+      "AllowedPattern" : "(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})/(\\d{1,2})",
+      "ConstraintDescription" : "must be a valid CIDR range of the form x.x.x.x/x."
+    }
+
+  },
+
+  "Mappings" : {
+    "SubnetConfig" : {
+      "VPC"     : { "CIDR" : "10.1.0.0/16" },
+      "Public"  : { "CIDR" : "10.1.0.0/24" },
+      "Private" : { "CIDR" : "10.1.1.0/24" }
+    }
+  },
+
+  "Resources" : {
+
+    "VPC" : {
+      "Type" : "AWS::EC2::VPC",
+      "Properties" : {
+        "CidrBlock" : { "Fn::FindInMap" : [ "SubnetConfig", "VPC", "CIDR" ]},
+        "Tags" : [
+          {"Key" : "Application", "Value" : { "Ref" : "AWS::StackId"} },
+          {"Key" : "Network", "Value" : "Public" }
+        ]
+      }
+    },
+
+    "PublicSubnet" : {
+      "Type" : "AWS::EC2::Subnet",
+      "Properties" : {
+        "VpcId" : { "Ref" : "VPC" },
+        "CidrBlock" : { "Fn::FindInMap" : [ "SubnetConfig", "Public", "CIDR" ]},
+        "Tags" : [
+          {"Key" : "Application", "Value" : { "Ref" : "AWS::StackId"} },
+          {"Key" : "Network", "Value" : "Public" }
+        ]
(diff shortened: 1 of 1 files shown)


cloudformation_templates/edx-analytics-vpc.json
```diff
@@ -0,0 +1,289 @@
...
+    "NATSecurityGroup" : {
+      "Type" : "AWS::EC2::SecurityGroup",
+      "Properties" : {
+        "GroupDescription" : "Enable internal access to the NAT device",
+        "VpcId" : { "Ref" : "VPC" },
+        "SecurityGroupIngress" : [
+           { "IpProtocol" : "tcp", "FromPort" : "80",   "ToPort" : "80",   "CidrIp" : { "Fn::FindInMap" : [ "SubnetConfig", "Private", "CIDR" ]}} ,
+           { "IpProtocol" : "tcp", "FromPort" : "443",  "ToPort" : "443",  "CidrIp" : { "Fn::FindInMap" : [ "SubnetConfig", "Private", "CIDR" ]}} ,
+           { "IpProtocol" : "tcp", "FromPort" : "9418", "ToPort" : "9418", "CidrIp" : { "Fn::FindInMap" : [ "SubnetConfig", "Private", "CIDR" ]}} ,
+           { "IpProtocol" : "tcp", "FromPort" : "22",   "ToPort" : "22",   "CidrIp" : { "Ref" : "SSHLocation" }} ],
```
e0d: Should lock down ssh to the bastion only.

cloudformation_templates/edx-analytics-vpc.json
```diff
@@ -0,0 +1,289 @@
...
+
+    "PrivateSubnetRouteTableAssociation" : {
+      "Type" : "AWS::EC2::SubnetRouteTableAssociation",
+      "Properties" : {
+        "SubnetId" : { "Ref" : "PrivateSubnet" },
+        "RouteTableId" : { "Ref" : "PrivateRouteTable" }
+      }
+    },
+
+    "PrivateRoute" : {
```
e0d, line 142: I believe a route scoped for the address space of the VPC is created auto-magically, but we should confirm that next time this is run.

cloudformation_templates/edx-analytics-vpc.json
```diff
@@ -0,0 +1,289 @@
...
+    },
+
+    "PrivateSecurityGroup" : {
+      "Type" : "AWS::EC2::SecurityGroup",
+      "Properties" : {
+        "GroupDescription" : "Enable secure access from the bastion host and free communication between nodes.",
+        "VpcId" : { "Ref" : "VPC" },
+        "SecurityGroupIngress" : [
+          { "IpProtocol" : "-1", "FromPort" : "0",  "ToPort" : "65535",  "CidrIp" : { "Fn::FindInMap" : [ "SubnetConfig", "Private", "CIDR" ]}},
+          { "IpProtocol" : "-1", "FromPort" : "0",  "ToPort" : "65535",  "SourceSecurityGroupId" : { "Ref" : "BastionSecurityGroup" }} ],
```
e0d, line 226: Too permissive, should apply principle of least privilege. I feel better about it from inside the private subnet.

cloudformation_templates/edx-analytics-vpc.json
```diff
@@ -0,0 +1,289 @@
...
+    "SubnetConfig" : {
+      "VPC"     : { "CIDR" : "10.1.0.0/16" },
+      "Public"  : { "CIDR" : "10.1.0.0/24" },
+      "Private" : { "CIDR" : "10.1.1.0/24" }
+    }
+  },
+
+  "Resources" : {
+
+    "VPC" : {
```
e0d, line 39: We have found it useful to enable DNS in the VPC:

```
    "EnableDnsSupport" : "true",
    "EnableDnsHostnames" : "true",
```

e0d:
Generally looks good.  I think we should have a notion of a private subnet that is specifically for data at rest.  Mongo would live in there, but need to know more about Hadoops requirements or what an HDFS deployment would look like.

cloudformation_templates/edx-analytics-vpc.json
mulby, line 226: I'm not sure I understand your comment.  Are you saying the "allow all inbound traffic from other hosts in the private subnet" is too permissive?  Or "allow all traffic from the bastion host" is too permissive?  Or both?

I haven't enumerated all of the various ports that hadoop uses, but I use the bastion host as a SOCKS proxy, so it needs to pass along connections on odd ports (like 50030 for the hadoop web UI).

Don't we trust all traffic coming from the bastion since it only allows SSH in?  I was assuming that if one of our keys was compromised then the whole security scheme is shot anyway...

cloudformation_templates/edx-analytics-vpc.json
mulby: Done

cloudformation_templates/edx-analytics-vpc.json
mulby, line 142: Yeah I wasn't sure about that... we should experiment the next time we run this.

cloudformation_templates/edx-analytics-vpc.json
mulby, line 39: Done

cloudformation_templates/edx-analytics-vpc.json
mulby, line 142: I've confirmed the ACLs are not necessary, it creates an "allow everything" ACL by default.

cloudformation_templates/edx-analytics-vpc.json
e0d, line 226: Sorry, catching up here.  Generally, we should specifically enumerate ports unless that's impractical.  I don't think this need to block merging.  For something like your case, it would be more typical to allow > 1024.

mulby:
Moved to analytics-configuration

PR #367, opened 2013-10-29
Common refactor - Applies changes to edx_sandbox

antoviaque:
Adapts the `edx_sandbox.yml` playbook to match the common-refactor changes, and also adds a missing default variable `EDXAPP_XQUEUE_URL`.


jarv:
thanks!

PR #368, opened 2013-10-29
Certain tasks (like generating grades) require a great deal of

jarv:
memory that isn't properly freed up. This creates a queue for such
workers so that we can give them different configuration settings
-- in this case, setting CELERYD_MAX_TASKS_PER_CHILD=1 to force
workers to die after they process one job. Because these jobs tend
to run for hours, the overhead is not a big deal.


feanil:
looks good.

PR #369, opened 2013-10-29
adding role for importing the demo course

jarv:


playbooks/edx-east/edx_continuous_integration.yml
```diff
@@ -19,6 +19,8 @@
     - supervisor
     - mongo
     - edxapp
+    - role: demo
+      tags: ['demo']
```
feanil, line 23: What is this tag for?

e0d:
+1

playbooks/edx-east/edx_continuous_integration.yml
feanil, line 23: talked to John, makes sense for limiting playbook runs to just the demo role.

feanil:
looks good.

PR #371, opened 2013-10-29
Update the jenkins worker to use the mongo role.

feanil:
The common role also has to be included because vars that were previously being set in group_vars_all are now in
the slimmed down common.

Review: @wedaly @jarv


jarv:
lgtm

wedaly:
Should probably remove `mongodb` from `jenkins_worker/defaults/main.yml` so it doesn't try to install the Ubuntu system package.  Other than that, this looks good.

feanil:
@wedaly already did that.

wedaly:
sorry, didn't see that :)

:+1:

feanil:
@wedaly realized that I had left the mongo conf stuff in the worker but it should not bee needed anymore so I removed it.

wedaly:
that makes sense

PR #375, opened 2013-10-30
Jarv/certs role

jarv:
@jrbl @feanil @e0d


feanil:
Looks good.

playbooks/roles/certs/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+    dest={{ certs_app_dir }}/{{ CERTS_LOCAL_PRIVATE_KEY|basename }}
+    owner={{ certs_user }} mode=0600
+  register: certs_gpg_key
+  tags: deploy
+
+
+- name: certs | load the gpg key
+  shell: >
+    /usr/bin/gpg --homedir {{ certs_gpg_dir }} --import {{ certs_app_dir }}/{{ CERTS_LOCAL_PRIVATE_KEY|basename }}
+  sudo_user: "{{ certs_user }}"
```
jrbl, line 73: From my reading of the ansible docs, this won't work right; have you tested this behavior carefully?  All of our playbooks run sudo: true, which is a whole-playbook parameter set from the top. Do you mean for this book to be runnable without that, in which case how can you be creating users? Or do you mean to have it sudo again to some other user for just this task, right? Which I think is the intended use of the remote_user field which will come out in Ansible 1.4.

jarv, line 73: If you set "sudo_user" it will run the command as `sudo -u certs ...` which is what I want to do here.
It assumes that whoever is running the play has permission to do that, make sense?
At the toplevel play we have "sudo: True" though the end goal will be to run the tasks tagged "deploy" without full sudo, eventually.

jrbl, line 73: Yeah, makes sense, both in explanation and in execution. I'm just surprised
because when I asked about this last week, I was told this wasn't possible
in 1.3. I'm happy to find out about this, thanks.

PR #376, opened 2013-10-30
Fix paths to Mongo in jenkins_worker

wedaly:
@feanil While cutting a new Jenkins worker AMI, I noticed a mismatch between the Mongo directories created by the Jenkins worker upstart script and the directories in the Mongo role config.

Longer term, there's probably a better way to do this.  Maybe an optional play in the mongo role for using ephemeral storage?  Short term, though, I'd like to get this in so I can update the Jenkins workers.

### DIFF
--- a/playbooks/roles/jenkins_worker/defaults/main.yml
+++ b/playbooks/roles/jenkins_worker/defaults/main.yml
@@ -47,8 +47,8 @@ jscover_url: "http://superb-dca2.dl.sourceforge.net/project/jscover/JSCover-1.0.
 jscover_version: "1.0.2"
 
 # Mongo config
-mongo_dir: "{{ COMMON_DATA_DIR }}/mongodb"
-mongo_log_dir: "{{ COMMON_DATA_DIR }}/logs/mongodb"
+mongo_dir: "{{ COMMON_DATA_DIR }}/mongo/mongodb"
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_worker/defaults/main.yml
```diff
@@ -47,8 +47,8 @@ jscover_url: "http://superb-dca2.dl.sourceforge.net/project/jscover/JSCover-1.0.
 jscover_version: "1.0.2"
 
 # Mongo config
-mongo_dir: "{{ COMMON_DATA_DIR }}/mongodb"
-mongo_log_dir: "{{ COMMON_DATA_DIR }}/logs/mongodb"
+mongo_dir: "{{ COMMON_DATA_DIR }}/mongo/mongodb"
```
jarv, line 50: We shouldn't be defining mongo role vars in the jenkins_worker role, can we just use the mongo role defaults?

from `roles/mongo/defaults/main.yml`

```
mongo_data_dir: "{{ COMMON_DATA_DIR }}/mongo"
mongo_log_dir: "{{ COMMON_LOG_DIR }}/mongo"
mongo_logpath: "{{ mongo_log_dir }}/mongodb.log"
mongo_dbpath: "{{ mongo_data_dir }}/mongodb"
```

wedaly:
Closing in favor of #380

PR #379, opened 2013-10-30
turns off ora and discern by default since they are broken on the single

jarv:
@jrbl @jbau @carsongee @feanil turns off ora/discern by default since they are both broken on master on the single instance sandbox

### DIFF
--- a/playbooks/roles/discern/handlers/main.yml
+++ b/playbooks/roles/discern/handlers/main.yml
@@ -1,6 +1,8 @@
 ---
+# FIXME: discern is not running properly yet in the single-server
+# ansible environment, leaving it stopped until it is fixed
 - name: discern | restart discern
-  service: name=discern state=restarted
+  service: name=discern state=stopped
 
 - name: discern | restart celery
-  service: name=celery state=restarted
+  service: name=celery state=stopped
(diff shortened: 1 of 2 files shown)


carsongee:
I have this working in croissant on a single instance and this would break that, if I'm reading it right, by turning off those services.  Any reason you don't just pull that role out of sandbox for the short term rather than turning of the service off in the role itself, or am I missing how this change would work?

jarv:
Would really like some help seeing if you can get this to work on master if you have the time to provision a new instance. The reason to disable it for now is so that we can install it but leave it shutdown while we troubleshoot it (and so we won't have failures)
I don't have a strong preference though if you think the problems will be easy to resolve.

jarv:
another reason to leave it off is because it is spamming the logs with errors though that's probably not a big deal

antoviaque:
FYI https://github.com/edx/configuration/pull/381 contains fixes to get ora/xqueue to work on the current master. I haven't had a look at discern though, so that one is still likely worth deactivating.

carsongee:
I can't help with discern, but ora is broken because it is calling back to the wrong port (I see that is fixed in @antoviaque PR).  It does still spam the log (at least celery) every 5 seconds by default mind you.

jarv:
OK, I'll take a more optimistic view on life and close this PR

PR #380, opened 2013-10-30
Use the mongo path settings provided by the mongo role.

feanil:
These variables are already defined in the mongo role.

An alternative to https://github.com/edx/configuration/pull/376

### DIFF
--- a/playbooks/roles/jenkins_worker/templates/mongodb_upstart.j2
+++ b/playbooks/roles/jenkins_worker/templates/mongodb_upstart.j2
@@ -1,12 +1,12 @@
 # Ubuntu upstart file at /etc/init/mongodb.conf
 
 pre-start script
-    mkdir -p {{ mongo_dir }}
+    mkdir -p {{ mongo_data_dir }}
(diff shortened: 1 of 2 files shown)


playbooks/roles/jenkins_worker/templates/mongodb_upstart.j2
```diff
@@ -1,12 +1,12 @@
 # Ubuntu upstart file at /etc/init/mongodb.conf
 
 pre-start script
-    mkdir -p {{ mongo_dir }}
+    mkdir -p {{ mongo_data_dir }}
```
jarv, line 4: aren't these directories created by the mongo role?

feanil, line 4: they are in the upstart script because the jenkins worker AMIs don't run ansible when they startup so the upstart script has to create the directories on ephemeral storage when mongo starts up.

feanil:
@wedaly merging this unless you have any reservations

wedaly:
:+1:

PR #381, opened 2013-10-30
Move xqueue to supervisor

antoviaque:
@jarv A first attempt at moving a service to supervisor -- can you check if I didn't forget something important? Once this is good to merge I'll do the same thing with the remaining services.

Btw this pull request also includes a series of fixes to get xqueue & ora to work properly, as I used the peer grading feature to check that xqueue was properly installed and functioned correctly.


playbooks/roles/supervisor/templates/supervisor-upstart.conf.j2
```diff
@@ -4,5 +4,4 @@ start on runlevel [2345]
 stop on runlevel [!2345]
 
 respawn
-setuid {{ common_web_user }}
```
jarv: We were thinking that supervisord as well as all the processes it controls would be run as `{{ common_web_user }}`.
This allows us to connect to it as that user which allows us to manage the processes as www-data, instead of root.

playbooks/roles/xqueue/templates/xqueue.conf.j2
```diff
@@ -1,25 +1,16 @@
...
-env PORT={{ xqueue_gunicorn_port }}
-env ADDRESS={{ xqueue_gunicorn_host }}
-env LANG=en_US.UTF-8
-env DJANGO_SETTINGS_MODULE=xqueue.aws_settings
-env SERVICE_VARIANT="xqueue"
 
+user={{ xqueue_user }}
+directory={{ xqueue_code_dir }}
+
+environment=PID=/var/tmp/xqueue.pid,PORT={{ xqueue_gunicorn_port }},ADDRESS={{ xqueue_gunicorn_host }},LANG=en_US.UTF-8,DJANGO_SETTINGS_MODULE=xqueue.aws_settings,SERVICE_VARIANT="xqueue"
```
jarv: even though it wasn't before I think we want LANG to be a var that can be overridden, something like `{{ XSERVER_LANG }}` (we did the same for edxapp already)

playbooks/roles/xqueue/templates/xqueue.conf.j2
```diff
@@ -1,25 +1,16 @@
...
-respawn
-respawn limit 3 30
-
-env PID=/var/tmp/xqueue.pid
 {% if ansible_processor|length > 0 %}
-env WORKERS={{ ansible_processor|length * 2 }}
+command={{ xqueue_venv_dir }}/bin/gunicorn --preload -b {{ xqueue_gunicorn_host }}:{{ xqueue_gunicorn_port }} -w {{ ansible_processor|length * 2 }} --timeout=300 --pythonpath={{ xqueue_code_dir }} xqueue.wsgi
 {% else %}
-env WORKERS=2
+command={{ xqueue_venv_dir }}/bin/gunicorn --preload -b {{ xqueue_gunicorn_host }}:{{ xqueue_gunicorn_port }} -w 2 --timeout=300 --pythonpath={{ xqueue_code_dir }} xqueue.wsgi
```
jarv: for this and the consumer you can use `{{ xqueue_venv_bin }}/gunicorn`

jarv: maybe add a new var for `{{ xqueue_gunicorn_workers }}` ?

playbooks/roles/ora/tasks/ease.yml
```diff
@@ -54,7 +54,7 @@
 
 - name: ora | install ease python package
   shell: >
-    {{ ora_ease_venv_dir }}/bin/activate; cd {{ ora_ease_code_dir }}; python setup.py install
+    . {{ ora_ease_venv_dir }}/bin/activate; cd {{ ora_ease_code_dir }}; python setup.py install
```
jarv: Can the ora changes be moved into a different PR?
I think this is fine, `{{ ora_eas_venv_bin }}/python setup.py install` will also work I think without sourcing the activate script.

playbooks/roles/supervisor/templates/supervisor-upstart.conf.j2
antoviaque: Then that means ora needs to be run as `{{ common_web_user }}` too no? Without it supervisord can't suid and gives the error:

```
supervisor: couldn't setuid to 1004: Can't drop privilege as nonroot user
supervisor: child process was not spawned
```

jarv: that's correct and what I am currently doing for other services (forum, lms, cms) that are running under supervisord.

playbooks/roles/ora/tasks/ease.yml
antoviaque: Sure - I've put the relevant commits in this pull request: https://github.com/edx/configuration/pull/382

antoviaque:
@jarv Made the changes, let me know if you see anything else/left.

playbooks/roles/xqueue/templates/xqueue.conf.j2
```diff
@@ -1,25 +1,16 @@
...
-respawn
-respawn limit 3 30
-
-env PID=/var/tmp/xqueue.pid
 {% if ansible_processor|length > 0 %}
-env WORKERS={{ ansible_processor|length * 2 }}
+command={{ xqueue_venv_bin }}/gunicorn --preload -b {{ xqueue_gunicorn_host }}:{{ xqueue_gunicorn_port }} -w {{ ansible_processor|length * 2 }} --timeout=300 --pythonpath={{ xqueue_code_dir }} xqueue.wsgi
 {% else %}
-env WORKERS=2
+command={{ xqueue_venv_bin }}/gunicorn --preload -b {{ xqueue_gunicorn_host }}:{{ xqueue_gunicorn_port }} -w 2 --timeout=300 --pythonpath={{ xqueue_code_dir }} xqueue.wsgi
```
antoviaque, line 6: For your remark here, about adding a variable `xqueue_gunicorn_workers` instead of the test, I agree it would be better. But I'm not sure how it would be possible to declare this in `defaults/main.yml`, since we need to use a conditionals (`{% if ansible_processor|length > 0 %}`)?

jarv:
lgtm

PR #382, opened 2013-10-30
Common refactor - Fix ora & xqueue

antoviaque:
Cf https://github.com/edx/configuration/pull/381 and #379 - I've put only the fixes to get ora working from #381.


jarv:
looks good

PR #383, opened 2013-10-30
Feanil/mongo rs

feanil:
Be able to automatiaclly setup a replica set when in aws.


carsongee:
The mongo stuff looks good to me.

PR #387, opened 2013-10-30
Change the default cloudformation elb backend ports.

feanil:
Now they default to the same as what is set in ansible which should reduce the number of things
we have to override when running ansible on a VPC.


jarv:
lgtm

PR #390, opened 2013-10-31
Some environment changes to speed up deploys

jarv:
- Instead of using pypi.edx.org why not cache on localhost?
  Seems to work well and it does result in a speed improvement
  for when we need to download packages. This adds the devpi
  role as dependency for edxapp
- For edxapp we md5sum all fo the requirements files and only
  run the pip installs, bundle, npm tasks if any one of those
  files have changed (it treats them as a group)
- Adds convenience symlinks for pip


feanil:
looks good.

feanil:
looks good.

PR #392, opened 2013-10-31
Move ora to supervisor

antoviaque:
Including the same extra fixes as for xqueue, plus:
- Exposed extra individual variables to override, for consistency with other roles
- Renamed the services & configuration files from `edx-ora*` to simply `ora`, for consistency with the role name

@jarv I've ported ora first, because I had it setup from yesterday already. If the current pull request goes well, and if xserver is still up for grabs and I manage to install it, it will be the next up.


jarv:
this looks great, thanks!

PR #396, opened 2013-10-31
Introduce POST size limit for ORA submissions

jrbl:
The ORA machinery gets gummed up when it gets very large submissions and
throws a lot of 500s.

Co-Authored-By: Jason Bau jbau@stanford.edu

@jarv please review.
@VikParuchuri please confirm that you don't know of anyone counting on being able to send large files.

### DIFF
--- a/playbooks/roles/nginx/templates/ora.j2
+++ b/playbooks/roles/nginx/templates/ora.j2
@@ -34,6 +34,7 @@ server {
   }
 
   location @proxy_to_app {
+    client_max_body_size 75K;
     proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
     proxy_set_header X-Forwarded-Port $http_x_forwarded_port;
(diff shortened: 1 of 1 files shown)


jarv:
lgtm

VikParuchuri:
@jrbl Why 75k?  No files are sent to ORA (only links to files).  Once concern is that most browsers don't handle the client_max_body_size well and throw a cryptic error.  Can we add some corresponding JS on the ORA side to limit the max length of submissions and peer feedback?  What kind of length was causing issues for you?

jbau:
Just jumping in here, we found that when updating the score with the peer feedback (when we were using the ICE track changes feature), very long (>200k) feedback would basically kill mysql. (We'd get OperationalError(Mysql has gone away)).  When this happens the rest of the site is affected as well. (TransactionErrors, etc)

So this is a band-aid to stop peer evaluators from submitting very large feedback.  This might not be the only thing we do, but it's a necessary thing.

PR #399, opened 2013-11-01
Updated Vagrant for refactored roles

wedaly:
I think the `rbenv` and `virtualenv` roles in fullstack and shortstack were leftover from before the refactoring.

@jarv does this look right?


jarv:
lgtm

PR #400, opened 2013-11-01
edxapp-devpi-fix: Add dependency on supervisor & fix venv install

antoviaque:
Fixes two issues:
- devpi was missing the dependency on the parameterized supervisor role, which wasn't installed
- the binary executable for devpi's `supervisorctl` was linked from devpi's venv, when supervisor installs its own venv under its own user


jarv:
Sorry about this, I didn't realize that playbooks/roles/devpi/meta was left untracked.
looks good

antoviaque:
@jarv No worries, I figured that was what happened : ) Thanks for the quick merge!

PR #404, opened 2013-11-02
Move xserver to supervisor

antoviaque:
@jarv As promised, the patch to move xserver from upstart to supervisor.

Note that it's my first experience with xserver, so it would be worth checking that I haven't overlooked anything important. Especially, there is no publicly released code grader, right? To get xserver running, and to be able to test, I have mocked one up at https://github.com/antoviaque/xserver-grader . It allowed me to test sending a code submission going through the LMS, xqueue and xserver, but this mock grader doesn't do much with the submission itself, especially it doesn't execute what's submitted.

If you have pointers to things I could use to test more thoroughly, it would be most useful : )

### DIFF
--- a/playbooks/roles/ora/tasks/deploy.yml
+++ b/playbooks/roles/ora/tasks/deploy.yml
@@ -56,8 +56,15 @@
   tags:
   - deploy
 
-- name: ora | reload supervisor to allow to start the new services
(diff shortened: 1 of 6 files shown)


jarv:
looks good, thanks!

PR #409, opened 2013-11-04
Jenkins worker refactor

wedaly:
Refactor of Jenkins worker role:

1) Use rbenv and edxlocal roles.  This required a slight change to the rbenv role: when using the rbenv environment from the shell, it defaults to using rake version 0.9.2 (installed with ruby in the rbenv).  The Ansible script removes that version and installs a later version (10.0.3).  This is the only way I've found to prevent rbenv from preferring the system installation to the gem installation of rake.

2) Remove upstart scripts that set up mongo/python/jenkins workspace in ephemeral storage.  This fixes a race condition between the upstart scripts and Jenkins jobs that would sometimes cause build failures.

3) Create a directory of wheel archives at AMI creation, which Jenkins jobs can use to create virtualenvs with pre-installed requirements.  This is less flexible than the previous implementation (which downloaded .egg files from S3), but much faster and more reliable.

If we later upload wheel files to S3, it should be easy to update the "wheel_venv.sh" script to download the archive instead of assuming it's part of the AMI.

Suggested reviewer: @feanil


playbooks/jenkins_worker.yml
```diff
@@ -7,9 +7,10 @@
   sudo: True
   gather_facts: True
   vars:
-      COMMON_DATA_DIR: '/mnt'
-      COMMON_LOG_DIR: '/mnt/log'
+      COMMON_DATA_DIR: '/var'
+      COMMON_LOG_DIR: '/var/log'
```
jarv: Why not use the defaults here so it is consistent with our other roles?

playbooks/roles/jenkins_worker/defaults/main.yml
```diff
@@ -1,34 +1,23 @@
...
-jenkins_user_home: /home/jenkins
+jenkins_home: /home/jenkins
 
 # System packages
 jenkins_debian_pkgs:
-    - ack-grep
     - build-essential
-    - git
     - gfortran
+    - git
```
jarv: If jenkins_worker depends on the common role some of these base packages will be provided.

wedaly: I think `git` is the only duplicate; the reason is that jenkins_worker explicitly upgrades git to a newer version.

I could roll that update into the common role, but I'd need to install `python-pycurl` in common so I can use Ansible's apt_repository module.  Is that something you want me to add to this PR?

playbooks/jenkins_worker.yml
wedaly: Would the default locations persist as part of the AMI?  Or would they be in ephemeral storage?  I wasn't sure, which is why I used /var (which I know get persisted with the AMI)

wedaly:
@jarv Updated to address your comments:

1) Moved the git upgrade stuff into the common role to avoid re-installing git
2) Used the default directory structure for consistency with the other playbooks

feanil:
looks good to me.

PR #412, opened 2013-11-04
ssl-nginx: Add new LMS variant to serve SSL through nginx

antoviaque:
Allow to use nginx to serve the LMS through SSL, when an external proxy (like the elastic load balancer) doesn't already handle it.

Deactivated by default, add `lms-ssl` to `nginx_sites` to activate it.


playbooks/roles/nginx/tasks/ssl.yml
```diff
@@ -0,0 +1,7 @@
+---
+- name: nginx | Install SSL Certificate
+  copy: src="{{secure_dir}}/files/ssl/lms.crt" dest="/etc/ssl/certs/lms.crt" owner=root group=root mode=0640
+
+- name: nginx | Install SSL Certificate - Key
+  copy: src="{{secure_dir}}/files/ssl/lms.key" dest="/etc/ssl/private/lms.key" owner=root group=root mode=0640
+
```
carsongee, line 7: Can you add a CA cert here as well?

carsongee:
This is a good start.  We are using SSL at MITx (both client and server), and it would be nice to get that merged in.  I still think we need to add the nginx template overrides in though, since for us there is a whole lot more to configure to get ssl going that I'm sure is more specialized than is needed for everyone else.

playbooks/roles/nginx/tasks/ssl.yml
antoviaque, line 7: I currently include the CA cert(s) directly in the `.crt` file (appended, one after the other) - this wouldn't work in your case?

antoviaque:
If you want needed I can add extra options - as long as they can have sensible defaults it's good with me : )

carsongee:
It turns out I meant ssl_client_certificate and not CA, which 99.9% of people won't use and thus won't have.  So we'd have to add some conditional to check for that.  Because of that I'm still thinking we'll have to do a full override of the template as proposed with #339 to handle our specialized needs. It may be also be more appropriate to do this that way rather than adding an extra variant by having a template that has both the ssl and plain http in it for your lms nginx template, but having a split http/server side only https lms is going to be a way more common use case than what we're doing.  I'm happy to go either way honestly though, so I'll either leave this up to another reviewer, or we can discuss at the next hangout ;) We could always go both ways too.

sasasaftic:
Can we activate ssl on edx after edx setup... using  /edx/etc/server-vars.yml?
(as we can for runnig our edx fork - https://github.com/edx/configuration/wiki/edX-Managing-the-Production-Stack#using-an-edx-platform-fork)
Or do I have to run whole ansible again?

hkim823:
@antoviaque and @carsongee is this still a valid PR at this point?  Let us know

carsongee:
This should be good to close as https://github.com/edx/configuration/pull/590 added SSL support.

PR #414, opened 2013-11-04
forum-ports-config: Use configuration variables to set services ports

antoviaque:
When using a custom mongo host/port which doesn't use the default mongo port (like mongolab instances), a lookup of the default mongo port doesn't work. Same issue with elasticsearch port.

### DIFF
--- a/playbooks/roles/forum/defaults/main.yml
+++ b/playbooks/roles/forum/defaults/main.yml
@@ -52,7 +52,5 @@ forum_elasticsearch_url: "http://{{ forum_elasticsearch_host }}:{{ forum_elastic
 #
 forum_services:
   - {service: "sinatra", host: "localhost", port: "4567"}
-  - {service: "mongo", host: "{{ forum_mongo_host  }}", port: "27017"}
-  - {service: "mongo", host: "{{ forum_mongo_host }}", port: "28017"}
-  - {service: "elasticsearch", host: "{{ forum_elasticsearch_host }}", port: "9200"}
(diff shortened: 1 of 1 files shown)


feanil:
Why did you remove the other two ports?  I think they are where mongo and elasticsearch expose other services.  For mongo I know it is the http api endpoint.

antoviaque:
Sorry, I sent this PR a bit too fast yesterday -- I should have provided more details and asked directly about those two services. The main reason for proposing to remove them is that I looked like they aren't used by edX:

For mongo, I could not find any other reference to that service, and not every instance of mongo provides it (mongolab provides another API https://support.mongolab.com/entries/20433053-Is-there-a-REST-API-for-MongoDB- and the official mongo doc http://docs.mongodb.org/ecosystem/tools/http-interfaces/ says it's disabled by default since 1.4). No worries to keep it, I can re-add the line and add an option to set the port. This would however mean that instances in a similar situation to mine would systematically need to override the whole `forum_services` dict to remove it.

For elasticsearch I think the port 9300 is http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/modules-transport.html which seem to be only used between nodes of a given cluster. I can also re-add it and add an option to set the port if you want to keep it too.

Let me know what you prefer, and I'll update this PR.

feanil:
@e0d you were the ones to originally add those ports I think.  Can you say for sure if we need them?

hkim823:
@e0d and @antoviaque let's see if this is still needed and either merge or close this PR

e0d:
This PR is no-longer relevant. closing.

PR #420, opened 2013-11-04
Feanil/vpc deploys

feanil:


playbooks/edx-east/edx_vpc.yml
```diff
@@ -0,0 +1,144 @@
+---
+- hosts: first_in_tag_role_mongo
+  sudo: True
+  vars_files:
+    - "{{ secure_dir }}/vars/dev/feanilsandbox.yml"
+    - "{{ secure_dir }}/vars/users.yml"
```
jarv: do we still need to import `feanilsandbox.yml`, I wiould like to get rid of user.yml in favor of just using gh_users.

playbooks/edx-east/edx_vpc.yml
```diff
@@ -0,0 +1,144 @@
...
+  roles:
+    - common
+    - role: 'mongo'
+      mongo_create_users: yes
+#- hosts: tag_role_mongo:!first_in_tag_role_mongo
```
jarv, line 11: commented out lines?

feanil, line 11: Currently we don't have a good way of determining which mongo server is the primary so I'm only provisioning one with clustering off.  Since for now we plan on using mongohq and heroku, I only need this server to test basic connectivity and it can be removed before the final version of this is merge in.

jarv:
lgtm

PR #422, opened 2013-11-05
fixing default for THEME_NAME

jarv:
@feanil @antoviaque

If THEME_NAME is set to an empty string it will cause a sass error due
to this mako code:

```
% if env.get('THEME_NAME') is not None:
  // import theme's Sass overrides
  @import '${env.get('THEME_NAME')}';
% endif
```


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -386,7 +386,7 @@ worker_core_mult:
 #Theming
 #To turn off theming, specify edxapp_theme_name: ''
```
feanil: You should update this comment as well so people don't make the same mistake in the future.

PR #424, opened 2013-11-05
Jarv/jenkins var updates for deploy

jarv:


playbooks/edx-east/edx_provision.yml
```diff
@@ -19,8 +19,10 @@
...
   hosts: launched
   sudo: True
-  gather_facts: True
+  gather_facts: False
   roles:
     # gh_users hash must be passed
     # in as a -e variable
     - gh_users
+    # Sets hostname to edx-server (default)
+    - hostname
```
feanil, line 28: Is this still necessary if you have to re-run rabbit anyway?

jarv, line 28: Yeah it isn't, I'll do a quick test to make sure things still come up without it included.

playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -18,6 +18,9 @@
 - name: rabbitmq | stop rabbit cluster
   service: name=rabbitmq-server state=stopped
 
+- name: rabbitmq | send sigterm to any running rabbitmq processes
+  shell: killall beam rabbitmq-server epmd
```
feanil: Shouldn't this only happen on refresh?

jarv: I think this should run every time assuming that even if the service is
already stopped there may be lingering processes.

playbooks/edx-east/edx_provision.yml
jarv, line 28: Changing the hostname caused problems for rabbit even after clearing out the mnesia dir and restarting both rabbit and the app.  I am going to leave it in for now.

feanil:
looks good.

PR #426, opened 2013-11-06
DO NOT MERGE: Pure ansible provisioning and MITx style architecture

carsongee:
This adds roles and a play for _mostly_ building a stack using pure
ansible. The roles build the MITx style stack with split AZ simple
tiers.  There is also a config variable to have this build a non
HA stack.  There is also a "load balancer" role that just rsyncs
the static content from the first app server to the nginx roled server
using whatever ssh key it can find.

Mostly this is just to show what is going on and isn't what I would
call public ready.  Particularly because it relies on two of my
PRs to the ansible repo.  One is just a module, but the other is an
enhancement to an existing module that lets you assign public IPs to
instances in a non-default VPC so that you don't need all the NAT
stuff.

### DIFF
--- a/playbooks/roles/loadbalancer/defaults/main.yml
+++ b/playbooks/roles/loadbalancer/defaults/main.yml
@@ -1,0 +1,1 @@
+course_static_dir: /mnt/data/repos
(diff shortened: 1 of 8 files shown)


hkim823:
@carsongee i've been tasked with cleaning up our PRs, is this something you still need as a PR?  Or can it be closed?

PR #427, opened 2013-11-06
Will/vagrant devstack

wedaly:
This is a prototype for `devstack`, a Vagrant instance for local development.  It's basically a clone of shortstack, with the following modifications:

1) Nginx and supervisor are not installed.  It's way easier to develop/debug when using runserver as a foreground process.  This is also the "typical" Django development workflow.

2) Port-forwarding, folder-sharing, and automatic updating of /etc/host for LMS preview.

3) `edxapp` and `forum` users are given login shells that automatically activate the appropriate Python/Ruby environment.  This provides a flexible way to add more repos, even if they have conflicting requirements.

4) Separated a new `browser` role from `jenkins_worker` to install headless browsers for running the test suite.

The instructions in the README depend on `devstack` settings, which are available in this branch of edx-platform: https://github.com/edx/edx-platform/compare/master...will;local-dev-settings

~~This PR includes changes from #409  (Jenkins worker refactor).  Once that gets merged, I will rebase.~~  Update: done.

Suggested reviewer: @feanil


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -5,13 +5,15 @@
 # TODO: the supervisor ansible module does not support
 # stopping and starting services by group.
 
+- debug: msg="supervisor_ctl = {{ supervisor_ctl }}"
```
jarv: is this necessary?

playbooks/roles/jenkins_worker/defaults/main.yml
```diff
@@ -1,52 +1,117 @@
 ---
```
jarv: are the jenkins_worker changes supposed to be in this pull as well?

wedaly: Yes; this PR is a branch of #409, which contains two changes I need (consolidating browser tasks in jenkins_worker and configuring rake).  I'm waiting on feedback for #409 but once that gets merged I'll rebase this branch.

If you want to see the devstack-specific changes, they're in the newer of the two commits.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+Vagrant instance for local development.
+
+
+Overview
+--------
+
+``devstack`` is a Vagrant instance designed for local development.  The instance:
+
+- Uses the same system requirements as production.  This allows developers to discover and fix system configuration issues early in development.
+- Simplifies certain production settings to make development more convenient.
```
jzoldak: Maybe give an example?

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+
+The ``devstack`` instance is designed to run code and tests, but you can do most development in the host environment:
+
+- Git repositories are shared with the host system, so you can use your preferred text editor/IDE.
+- You can load pages served by Vagrant from browsers in the host system.
```
jzoldak: served by the running Vagrant instance

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+
+
+Prerequisites
+-------------
+
+This guide assumes:
+
+- You know how to open a terminal and use basic commands.
```
jzoldak: - You understand basic virtualization concepts.
- You know how to open a terminal and use basic Unix operating system commands.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+- You are familiar with `Sinatra`__ development (for Forums)
+
+__ http://www.djangoproject.com/
+__ http://www.sinatrarb.com/
+
+
+Installation
+------------
+
+1. `Install Vagrant`__
```
jzoldak: Version X or higher

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+2. Install the ``vagrant-hostsupdater`` plugin:
+
+.. code:: bash
+
+    vagrant plugin install vagrant-hostsupdater
+
+3. `Install Ansible`__
```
jzoldak: Version X or higher

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+4. Clone the edX configuration repo:
+
+.. code:: bash
+
+    git clone https://github.com/edx/configuration
+
+5. From the ``configuration/vagrant/devstack`` directory, create the Vagrant instance:
```
jzoldak: If you have never used Vagrant before, take a moment to read the Why Vagrant page: http://docs.vagrantup.com/v2/why-vagrant/index.html The Vagrant doc site explains in detail all the commands that we will be using.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+.. code:: bash
+
+    vagrant up
+
+A new Vagrant instance will be created, with requirements for these repositories installed:
+
+- `edx-platform`__ (LMS and Studio)
+- `cs_comments_service`__
+
+The first time you create the instance, ``vagrant`` will download the base box, which is about 2GB.  From then on, however, Vagrant will re-use the box it downloaded.
```
jzoldak: After downloading the base box, vagrant will automatically provision your new virtual server using the Ansible scripts from this repo.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+
+
+Getting Started
+---------------
+
+In the ``configuration/vagrant/devstack`` directory, ssh into the Vagrant instance:
+
+.. code:: bash
+
+    vagrant ssh
```
jzoldak: Explain what username / password this will be using, and what directory LMS / Studio / Forums will have been created in.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+For Django applications, we use ``runserver`` for local development.
+See `Django's documentation`__ for more information on ``runserver``.
+
+__ https://docs.djangoproject.com/en/dev/ref/django-admin/#runserver-port-or-address-port
+
+
+LMS Workflow
+------------
+
+1. Within the Vagrant instance, activate the ``edxapp`` account:
```
jzoldak: I would say switch to the 'edxapp' user, rather than activate the 'edxapp' account.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+
+    sudo su edxapp
+
+2. Compile sass and coffeescript:
+
+.. code:: bash
+
+    rake assets[lms,devstack]
+
+3. Update Python and Ruby requirements:
```
jzoldak: I'd split these steps out to be clear that pip install is for python and bundle install is for ruby

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+    pip install -r requirements/edx/base.txt
+    bundle install
+
+4. Start the LMS server:
+
+.. code:: bash
+
+    ./manage.py lms runserver --settings=devstack 0.0.0.0:8000
+
+5. Open a browser and navigate to ``localhost:8000`` to load the LMS.
```
jzoldak: open a browser on your host machine. Explain that port 8000 from the host server will forward the request to the running vagrant server.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+
+    ruby app.rb
+
+4. Access the API at ``localhost:4567``
+
+
+Switching Accounts
+------------------
+
+To switch between accounts, you will need to log out of the current account using the ``exit`` command.  For example, to log in as ``edxapp`` and then switch to the ``forum`` acount:
```
jzoldak: It's unclear to me why/when they would want to do this. Are you assuming that they would only be running one at a time? I can definitely see the use case where they want both Studio and LMS. (Also in this case they'd probably want to run both preview and live on the lms side.) Or lms/forums.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+    git pull
+
+2. From the ``configuration/vagrant/devstack`` directory, provision the Vagrant instance:
+
+.. code:: bash
+
+    vagrant provision
+
+
+This process will perform a ``git clean`` of the ``edx-platform`` and ``cs_comments_service`` repositories.
```
jzoldak: so make sure that any changes you had are checked in or stashed.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,285 @@
...
+--------------------------
+
+To destroy and recreate the environment:
+
+.. code:: bash
+
+    vagrant destroy
+    vagrant up
+
+This will perform a ``git clean`` of the ``edx-platform`` and ``cs_comments_service`` repositories.  You will also lose any work stored in the Vagrant instance.
```
jzoldak: as that instance will be destroyed and a new one created from scratch.

vagrant/devstack/README.rst
wedaly: edxapp --> forum was the use case I had in mind.  If you try to do that without first logging out of edxapp, `sudo` will prompt for a password, which you don't have access to.

Given that "basic Unix commands" are listed in the prereqs, though, I feel comfortable taking this section out.

wedaly:
@jzoldak Thanks for the detailed feedback!  Updated to include your suggestions.

playbooks/roles/local_dev/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
+---
+- name: local_dev | install useful system packages
+  apt:
+    pkg={{','.join(local_dev_pkgs)}} install_recommends=yes
+    state=present update_cache=yes
+
+- name: local_dev | set login shell for app accounts
+  user: name=${item.user} shell="/bin/bash"
```
jarv: use new variable syntax here

jarv: One thing to keep in mind here is that if any of the main.yml tasks are run for the corresponding roles the shell will flip back to /bin/false.  As long as in the dev environment we run with `--tags deploy` after initial bringup we should be fine

wedaly:
@jarv I'm making the assumption that the `local_dev` role always executes after the roles it depends on (as it does in the `vagrant-devstack.yml` playbook).  So I think even if the other roles switch the shell back to `/bin/false`, the `local_dev` role will always reset it to `/bin/bash`.  Most people will use `vagrant up` or `vagrant provision` to do this, which uses `vagrant-devstack.yml`, so I think this is a safe assumption.

antoviaque:
@wedaly Very cool, many thanks for this! I've just tried the install, it worked perfectly. I'll try to adopt it as my dev environment, to see how that goes. : )

I'll also post a few comments I had while going through the install, if that helps.

vagrant/devstack/Vagrantfile
```diff
@@ -0,0 +1,33 @@
...
+CPU_COUNT = 2
+
+Vagrant.configure("2") do |config|
+
+# For efficiency, we create custom base box from an Ubuntu 12.04 image
+# Once the custom base box is created, we use that as a starting point.
+#  config.vm.box     = "precise64"
+#  config.vm.box_url = "http://files.vagrantup.com/precise64.box"
+  config.vm.box    = "edx-devstack"
+  config.vm.box_url = "https://s3.amazonaws.com/vagrant.testeng.edx.org/edx-devstack.box"
```
antoviaque, line 11: Why not the default precise64 box? It would avoid having to re-download a full image when you already have precise64, and would make it clearer that the configuration is based on a pristine installation of 12.04.

vagrant/devstack/README.rst
```diff
@@ -0,0 +1,296 @@
...
+The first time you create the instance, Vagrant will download the base box, which is about 2GB.
+After downloading the base box, Vagrant will automatically provision your new virtual server using the Ansible scripts from this repo.
+If you destroy and recreate the VM, Vagrant will re-use the box it downloaded.
+
+__ http://docs.vagrantup.com/v2/installation/index.html
+__ http://www.ansibleworks.com/docs/intro_installation.html
+__ http://docs.vagrantup.com/v2/why-vagrant/index.html
+__ https://github.com/edx/edx-platform
+__ https://github.com/edx/cs_comments_service
+
```
antoviaque, line 73: Might also be worth mentioning the vagrant troubleshooting page of edx-platform, as there are a couple of FAQ that are likely to surface when people will start to use this, especially about the nfs sharing (https://github.com/edx/edx-platform/wiki/Simplified-install-with-vagrant#mounting-nfs-shared-folders and for Windows-specific issues https://github.com/edx/edx-platform/wiki/Simplified-install-with-vagrant#dealing-with-line-endings-and-symlinks-under-windows ).

Something like:

```
If something goes wrong, you can refer to the
[troubleshooting section](https://github.com/edx/edx-platform/wiki/Simplified-install-with-vagrant#troubleshooting).
```

vagrant/devstack/Vagrantfile
wedaly, line 11: The main reason is speed of installation.  Because the scripts are installing so many requirements, it can take up to 40 minutes to provision a box from scratch.

In addition, I want developers to be able to quickly destroy and recreate the box if they need to.  After they've downloaded the custom box once, this process is very fast (several minutes).

vagrant/devstack/README.rst
wedaly, line 73: Good call; I added the link below.

jarv:
lgtm after rebase

PR #428, opened 2013-11-06
Jarv/single instance updates

jarv:
adds the inet_http_server to the supervisor config, by default will listen on localhost:9001
override added so that you can open it up to the world which is not recommended but appropriate for development sandboxes.

added stopasgroup/killasgroup options to supervisord which helps resolve an issue where processes were not being sent the stop/term signals when supervisor is reloaded.


feanil:
looks good.

PR #413, opened 2013-11-04
edxapp-extra-vars: Expose extra variables for configuration of the edxapp role

antoviaque:
Expose extra variables for configuration overload. These were things that I was missing to migrate the codecoalition.com instance.

Of specific note, `EDXAPP_ENV_EXTRA` and `EDXAPP_AUTH_EXTRA` allow to add variables to `auth.json` and `env.json` without having to overload the whole generic_env_config/generic_auth_config dictionary. This is especially useful when adding custom djangoapps or code which define extra configuration variables.

PR #429, opened 2013-11-07
Added ansible.cfg symlink for shortstack and shortstack-xml

wedaly:
#413 introduced "do" tags into some of the Jinja templates, which requires a Jinja extension that shortstack and shortstack-xml hadn't enabled.

I've added symlinks to `ansible.cfg` which enables this extension.  This is the same setup as fullstack.

Suggested reviewer: @feanil


feanil:
looks good.

PR #432, opened 2013-11-07
Creates a glusterfs role

carsongee:
Works by turning all peers into members of all defined volumes
and any machine that isn't in the peers list mounts the
gluster share from the primary peer.

If you don't want this, no biggie, but this is how we are sharing static course content on the MITx servers instead of serving it out of mongo (but still using the mongo data store for everything else). It can also be used to get out of having to rsync static assets up to the nginx servers like you saw in my sample play.

### DIFF
--- a/playbooks/roles/gluster/tasks/main.yml
+++ b/playbooks/roles/gluster/tasks/main.yml
@@ -1,0 +1,83 @@
+---
+# Install and configure simple glusterFS shared storage
+
+- name: gluster | all | Install common packages
+  apt: name={{ item }} state=present
+  with_items:
+    - glusterfs-client
+    - glusterfs-common
+    - nfs-common
+  tags: gluster
+
+- name: gluster | all | Install server packages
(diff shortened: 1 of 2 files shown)


playbooks/roles/gluster/tasks/main.yml
```diff
@@ -0,0 +1,83 @@
...
+  tags: gluster
+
+- name: gluster | all | Install server packages
+  apt: name=glusterfs-server state=present
+  when: >
+    "{{ ansible_default_ipv4.address }}" "{{ gluster_peers|join(' ') }}"
+  tags: gluster
+
+- name: gluster | all | enable server
+  service: name=glusterfs-server state=restarted enabled=yes
```
jarv: is this necessary? I assume the service will be started after installation. Or maybe we just need a `state=started` here

carsongee: Good catch, fixed.

jarv:
lgtm

PR #433, opened 2013-11-07
Jarv/add hostname to common

jarv:
@feanil 

This PR does the following:
- Moves the logic to change the hostname into the common role, by default it won't do anything to change the hostname unless COMMON_HOSTNAME is set.
- Creates /etc/skel/.bashrc that adds an optional prefix to PS1


playbooks/roles/gh_users/tasks/main.yml
```diff
@@ -12,9 +12,10 @@
 #      - mark
 
 
-- name: gh_users | checking for gh_users list
```
feanil, line 15: Why did you remove this check and instantiate the gh_users to empty?

jarv, line 15: I want to include the gh_users role to update the .skel directory with the default .bashrc (and optionally not create an users).

feanil, line 15: I'd rather see the .bashrc settings in common I think.

feanil, line 15: talked in person, this makes sense because of how the .skel directory is tied to user creation.

feanil:
looks good.

PR #437, opened 2013-11-12
Added login shell for Jenkins worker account

wedaly:
In the last refactor, the `jenkins` account on the worker lost its login shell.  Jenkins uses `scp` to upload a JAR and other scripts to the worker, so Jenkins wasn't able to launch the worker.

This PR gives the Jenkins worker account a login shell so it will work correctly with Jenkins.

@feanil


jarv:
lgtm

PR #439, opened 2013-11-13
Jarv/wait for cloud init

jarv:
adds wait_for condition for /var/log/cloud-init which is necessary if you do package operations immediately after provisioning.  Module changes also submitted to ansible core


feanil:
looks good.

PR #441, opened 2013-11-13
Deleted pipeline directions from readme

wedaly:
Just tried to enable the pipeline and it didn't seem to work.  Deleting the instructions from the README before the release so I can take the time to trouble-shoot.

@jarv


jarv:
lgtm, though I plan to kill the README in a future pull with a link to the configuration wiki instead if that's ok with you.

wedaly:
@jarv Of course -- it makes sense to put it wherever people are most likely to find it.

PR #442, opened 2013-11-13
Jarv/vagrant releases

jarv:
- removes shortstack-xml
- removes shortstack
- creates a vagrant/release dir for devstack and fullstack


feanil:
looks good.

vagrant/devstack/README.rst
```diff
@@ -18,253 +18,5 @@ The ``devstack`` instance is designed to run code and tests, but you can do most
...
-
-To destroy and recreate the environment:
-
-.. code:: bash
-
-    vagrant destroy
-    vagrant up
-
-This will perform a ``git clean`` of the ``edx-platform`` and ``cs_comments_service`` repositories.  You will also lose any work stored in the Vagrant instance as that instance will be destroyed and a new one created from scratch.
+Installation instructions - https://github.com/edx/configuration/wiki#installing-edx-using-a-pre-built-vagrant-devstack-image
```
wedaly, line 21: Please delete this from the devstack installation instructions in the wiki:

```
 After installation connect your web browser to the following ports:

 LMS - 80
 Studio - 18010
```

vagrant/devstack/Vagrantfile
```diff
@@ -5,19 +5,14 @@ Vagrant.configure("2") do |config|
 
 # For efficiency, we create custom base box from an Ubuntu 12.04 image
 # Once the custom base box is created, we use that as a starting point.
```
wedaly: These comments should be deleted.

vagrant/devstack/README.rst
jarv, line 21: done, but also feel free to make any edits you want! :)

wedaly:
One consequence of cutting a new box from Ubuntu 12.04 (instead of the base box in testeng S3) is that Chrome got updated to version 31.  This has unfortunately broken the acceptance tests and no longer aligns with what's running in Jenkins.

As far as I can tell, Google doesn't maintain old versions of Chrome in its apt repositories (otherwise I would have explicitly installed that version).  

This isn't an issue if you provision from the testeng base box, because it the ansible `apt` module sees that `google-chrome-stable` is installed and doesn't upgrade it.

I don't think this should block the release, but I'm wondering if you have ideas about how to address this?

jarv:
> One consequence of cutting a new box from Ubuntu 12.04 (instead of the base box in testeng S3)

When you created the base box in S3 didn't you use a base precise image?  If not what version did you use?

wedaly:
@jarv I did use the precise base image, but I cut the box at the same time I upgraded the Jenkins workers.  At that time, `google-chrome-stable` in the apt repo was version 30.

Google released version 31 today and I found out from the blades team an hour ago that it's breaking things.  I haven't had a chance yet to figure out why.

jarv:
Ah I see - can we just lock the version to 30 in ansible for now?

wedaly:
@jarv Unfortunately no.  As far as I can tell, Google only provides the latest version in the apt repo.

The main reason this is a problem is that ChromeDriver (which is how Selenium interacts with Chrome) needs to be updated as well.  So maybe it makes sense to update ChromeDriver to the latest version in this release:

https://github.com/edx/configuration/pull/443

And I can upgrade Jenkins workers today as well so they match.

feanil:
@wedaly  would it make sense to save the .deb to an s3bucket we do this for
splunk packages(for slightly different reasons)?

wedaly:
@feanil Yeah, that's a really good idea.

wedaly:
https://github.com/edx/configuration/pull/446 uses S3 to grab the debian packages for Chrome and FF.  I'm building the Vagrant image now from base Ubuntu 12.04 to verify that it works correctly.  If so, I'd really like those versions to be in the release, since it's going to take some work to get the tests running under Chrome 31 :(

jarv:
@wedaly sure, I will take that change and cut a new official release.
in the meantime I believe this branch is ready to merge, I updated the official release vagrant file for devstack to 
only run with the deploy tag (so it's faster on bringup)  and I added an env variable to set the mount base directory.

https://github.com/edx/configuration/commit/5f234b0ef44fa43cadfea74f7cce098f467b9a5c

wedaly:
Awesome, looks good :+1:

PR #444, opened 2013-11-13
Moved the role to using the root s3fs as it supports IAM

carsongee:
I don't know if there were other reasons to use IAM, but I just tested the current s3fs package with an IAM role, and all seems well.


e0d:
Yeah I got a  notification recently that this feature was merged to master,  so makes sense.

carsongee:
So good to merge then?

e0d:
There was one other fix in the github fork related to a memory leak,  I wonder if that has been merged.

Would be good to know,  but I'm :+1:

PR #445, opened 2013-11-13
Stanford-oriented Datadog updates

jrbl:
- Datadog explicitly sets hostname in style Stanford uses (relies on
  /etc/hostname being accurate, requires reboot)
- Enables nginx status check module integration for datadog
- Installs nginx access log processor that counts HTTP response codes
  and emits events for the individual components
- Supporting updates to Stanford playbooks

@jbau and @sefk I can show you the testing dashboard I've been using to confirm this; all of this should be installed to staging.


jrbl:
@caesar2164 I think this can be merged and then you can fetch and pull and use this for the prod release, if you'd like. Otherwise I'll have to do the dog stuff separately today which might be a pain. What say you?

jrbl:
(@jbau already looked and said it seemed reasonable.)

jbau:
@jrbl I'm curious where you fixed the datadog streaming too many events problem

jbau:
:+1:

PR #446, opened 2013-11-13
Use debian packages from S3 to fix the browser versions for Chrome and FF

wedaly:
Downloads the debian packages for Chrome and FF from S3 instead of using the apt-repositories, since the apt-repos don't maintain older versions of the browsers.

For now I've uploaded the deb packages by hand to S3, but I'll automate this in Jenkins when I get a chance.

@feanil


feanil:
looks good.

PR #456, opened 2013-11-15
Add logrotation to Stanford datadog setup

jrbl:
- Datadog logs now get rotated like other system logs
- nginx logs get rotated to have group syslog, so the datadog agent will
  continue to be able to read them

@jbau please look over


jarv:
lgtm

PR #448, opened 2013-11-14
vagrant-devstack-fix: Fix inverted names for nfs-mounted directories

antoviaque:

PR #457, opened 2013-11-15
edx_platform and cs_comments_service were swapped

wedaly:
Finishes the fix started in this PR: #448

@jarv


feanil:
looks good.

PR #463, opened 2013-11-17
Disable mongo journaling on devstack and Jenkins

wedaly:
The Mongo journal files are taking up a lot of disk space on Jenkins workers and in devstack.  This PR provides an option to disable journaling in these environments.

According to the [docs](http://docs.mongodb.org/manual/reference/configuration-options/#journal), you need two options to disable journaling on 64-bit systems:

```
journal=false
nojournal=true
```

I also noticed that the `browsers` task was included in the Jenkins worker, leftover from devstack changes, so I removed it.

### DIFF
--- a/playbooks/jenkins_worker.yml
+++ b/playbooks/jenkins_worker.yml
@@ -6,6 +6,8 @@
   hosts: jenkins_worker
   sudo: True
   gather_facts: True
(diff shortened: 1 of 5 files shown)


feanil:
looks good to me

PR #465, opened 2013-11-18
Skip gather assets in devstack

wedaly:
Since the devstack settings use Django staticfiles, I think it's safe to disable asset collection while provisioning the VM.  This should greatly speed up the initial `vagrant up` and subsequent `vagrant provision` commands.

@jarv


jarv:
lgtm

PR #467, opened 2013-11-18
support creation of a datajam specific devstack image

mulby:
Including:
- LMS
- CMS
- Forums
- Insights

Note: this is very rudimentary, and kind of just hacked together so that we can repeatably build vagrant images for the datajam as we iterate.

Outstanding issues:
- It should probably check out the "datajam" branches of stuff instead of "master"
- I wasn't able to import a course without first creating the /edx/app/edxapp/data directory even though that directory remains empty after I import the course.  I suspect some part of the code is erroneously checking for it's existence (reading the wrong setting etc).

Reviewers: @rocha, @brianhw, @e0d

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,10 +5,6 @@
 \#*\#
 *~
 .#*
-vagrant/devstack/cs_comments_service
-vagrant/devstack/edx-platform
-vagrant/datajam/cs_comments_service
-vagrant/datajam/insights
-vagrant/datajam/edx-platform
-vagrant/release/*/devstack/cs_comments_service
-vagrant/release/*/devstack/edx-platform
+vagrant/*/cs_comments_service
+vagrant/*/insights
+vagrant/*/edx-platform
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,8 @@
 .#*
 vagrant/devstack/cs_comments_service
 vagrant/devstack/edx-platform
+vagrant/datajam/cs_comments_service
+vagrant/datajam/insights
+vagrant/datajam/edx-platform
 vagrant/release/*/devstack/cs_comments_service
 vagrant/release/*/devstack/edx-platform
--- a/playbooks/roles/insights/defaults/main.yml
+++ b/playbooks/roles/insights/defaults/main.yml
@@ -31,7 +31,7 @@ insights_debian_pkgs:
 insights_app_dir: "{{ COMMON_APP_DIR }}/insights"
 insights_venv_dir: "{{ insights_app_dir }}/venv"
 
-INSIGHTS_VERSION: datajam1
+INSIGHTS_VERSION: master
 INSIGHTS_DJEVENTSTREAM_VERSION: master
 INSIGHTS_LOGHANDLERPLUS_VERSION: master
 
--- a/playbooks/roles/insights/defaults/main.yml
+++ b/playbooks/roles/insights/defaults/main.yml
@@ -1,20 +1,53 @@
 ---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role insights
+#
+
+#
+# vars are namespace with the module name.
+#
+insights_role_name: insights
+
+#
+# OS packages
+#
+
+insights_debian_pkgs:
+  - git
+  - python-pip
+  - python-virtualenv
+  - redis-server
+  - mongodb
+  - nginx
 
 insights_app_dir: "{{ COMMON_APP_DIR }}/insights"
 insights_venv_dir: "{{ insights_app_dir }}/venv"
+
+INSIGHTS_VERSION: datajam1
+INSIGHTS_DJEVENTSTREAM_VERSION: master
+INSIGHTS_LOGHANDLERPLUS_VERSION: master
 
 insights_repos:
   insights:
     code_dir: "{{ insights_app_dir }}/insights"
     repo: https://github.com/edx/insights.git
-    version: master
+    version: "{{ INSIGHTS_VERSION }}"
   djeventstream:
     code_dir: "{{ insights_app_dir }}/djeventstream"
     repo: https://github.com/edx/djeventstream.git
-    version: master
+    version: "{{ INSIGHTS_DJEVENTSTREAM_VERSION }}"
   loghandlersplus:
     code_dir: "{{ insights_app_dir }}/loghandlersplus"
     repo: https://github.com/edx/loghandlersplus.git
-    version: master
+    version: "{{ INSIGHTS_LOGHANDLERPLUS_VERSION }}"
 
 insights_requirements_file: "{{ insights_repos.insights.code_dir }}/requirements.txt"
+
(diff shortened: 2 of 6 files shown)


feanil:
@wedaly might also be a good person to look at this since he has recently been working on the devstack

.gitignore
```diff
@@ -7,5 +7,8 @@
 .#*
 vagrant/devstack/cs_comments_service
 vagrant/devstack/edx-platform
+vagrant/datajam/cs_comments_service
+vagrant/datajam/insights
+vagrant/datajam/edx-platform
```
jarv: we probably want something like `vagrant/*/cs_comments_service` `vagrant/*/insights`, `vagrant/*/edx-platform`

playbooks/roles/insights/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
+---
+
+insights_app_dir: "{{ COMMON_APP_DIR }}/insights"
```
jarv: please add a comment block to the top of this or better use the create_role.yml play to create the role which will do it for you.

`ansible-playbook -c local --limit "localhost," ./create_role.yml -i "localhost," -e role_name=my_awesome_role`

playbooks/roles/insights/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
...
+    repo: https://github.com/edx/insights.git
+    version: master
+  djeventstream:
+    code_dir: "{{ insights_app_dir }}/djeventstream"
+    repo: https://github.com/edx/djeventstream.git
+    version: master
+  loghandlersplus:
+    code_dir: "{{ insights_app_dir }}/loghandlersplus"
+    repo: https://github.com/edx/loghandlersplus.git
+    version: master
```
jarv: We should have top-level vars for these versions, something like:

```
INSIGHTS_DJEVENTSTREAM_VERSION: master
...

insights_repos:
  insights:
  ...
    version: "{{ INSIGHTS_DJEVENTSTREAM_VERSION }}"

```

This will allow us to override the version easily when we create deploy jobs.

playbooks/roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+
+- name: insights | system packages
+  apt: pkg={{item}}
+  with_items:
+    - git
+    - python-pip
+    - python-virtualenv
+    - redis-server
+    - mongodb
+    - nginx
```
jarv: please move this list into defaults.yml

playbooks/roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+  tags:
+    - insights
```
jarv: there is no reason to add role tags to all tasks for the role, you can add the tag at the play level.

```
- roles:
  - { role: insights, tags: ['insights'] }
```

playbooks/roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    - deploy
+
+- name: insights | create code directories
+  file: >
+    path="{{item}}" state=directory
+    owner="{{ edxapp_user }}" group="{{ common_web_group }}"
+  with_items:
+    - "{{insights_repos.djeventstream.code_dir}}"
+    - "{{insights_repos.loghandlersplus.code_dir}}"
```
jarv: This would be much better if insights_repo was a list of dictionaries, this way you can just do a with_items: over it and reference the code_dir as `{{ item.code_dir }}`

playbooks/roles/insights/defaults/main.yml
```diff
@@ -0,0 +1,21 @@
...
+insights_repos:
+  insights:
+    code_dir: "{{ insights_app_dir }}/insights"
+    repo: https://github.com/edx/insights.git
+    version: master
+  djeventstream:
+    code_dir: "{{ insights_app_dir }}/djeventstream"
+    repo: https://github.com/edx/djeventstream.git
+    version: master
+  loghandlersplus:
```
brianhw: I thought we weren't using loghandlersplus.

playbooks/vagrant-datajam.yml
```diff
@@ -0,0 +1,22 @@
+- name: Configure instance(s)
+  hosts: vagrant
+  sudo: True
+  gather_facts: True
+  vars:
+    migrate_db: "yes"
+    openid_workaround: True
+    devstack: True
+    edx_platform_commit: 'master'
```
brianhw, line 9: Is this where we specify the datajam branch?

playbooks/roles/insights/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+  sudo_user: "{{ edxapp_user }}"
+  tags:
+    - insights
+    - deploy
+
+- name: insights | install projects
+  shell: >
+     . {{insights_venv_dir}}/bin/activate && cd {{item.code_dir}} && python setup.py install && deactivate
+  sudo_user: "{{ edxapp_user }}"
+  with_items: insights_repos.values()
```
brianhw, line 62: Are we also installing insights itself?  When I tested things out, I installed djeventstream and insights into the virtualenv, then ran out of the example code.  (And because I was installing, that's why I wanted to specify a different version in the setup, so that the datajam install was distinct from the one from master.)

playbooks/roles/insights/defaults/main.yml
mulby: We don't want to long term, but for now I'm pretty sure it's tightly integrated into insights.

playbooks/roles/insights/tasks/main.yml
mulby, line 62: Yes, note that "insights_repos" includes insights itself.

playbooks/vagrant-datajam.yml
mulby, line 9: Yup!

mulby, line 9: Obviously I haven't gone through and done it yet, I'd like to wait until the release this week since it includes the REST API for the modulestore.  I suppose I could make the datajam branch point at the commit that includes that stuff, just haven't gotten around to it yet.

.gitignore
mulby: Good call, done.

playbooks/roles/insights/defaults/main.yml
mulby: Didn't know this existed!  Done.

playbooks/roles/insights/defaults/main.yml
mulby: Done.

playbooks/roles/insights/tasks/main.yml
mulby: Done.

playbooks/roles/insights/tasks/main.yml
mulby: Awesome, didn't know you could do that.  Done.

playbooks/roles/insights/tasks/main.yml
mulby: I do this in other places by iterating over the values.  There are a couple of places that I have to reference individual repos, so I would like to keep it a dictionary if possible.  The script bombs out if I try to create the code_dir for the insights project since vagrant creates that directory and sets up the NFS mount etc, so I need to create all of the code_dir values except that one.

mulby:
we are treating the datajam branch as a WIP, so I'm merging this in now.  It should not be merged into master!

PR #469, opened 2013-11-18
Added gettext requirement for i18n

wedaly:
The edx-platform i18n rake tasks complain if `gettext` isn't installed with this error message:

```
    Cannot locate GNU gettext utilities, which are required by django for internationalization.
    (see https://docs.djangoproject.com/en/dev/topics/i18n/translation/#message-files)
    Try downloading them from http://www.gnu.org/software/gettext/
```


jarv:
lgtm

PR #475, opened 2013-11-20
Added deploy tags for some local_dev tasks

wedaly:
When provisioning vagrant devstack, we only run Ansible deploy tasks.  This means that some of the work done by the `local_dev` role can be reverted (namely, setting login shells and permissions).

This PR tags the `local_dev` tasks that should run when deploying.

@feanil


feanil:
looks good.

PR #477, opened 2013-11-21
use correct branches for the datajam

mulby:
Reviewers: @rocha, @wedaly, @brianhw


playbooks/roles/datajam/files/edx-cms-devserver
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+cd /edx/app/edxapp/edx-platform
+./manage.py cms --settings devstack runserver --traceback 0.0.0.0:8001
```
brianhw: Is line-ending an issue?  Looks like a number of devserver scripts are missing the last linefeed.

mulby: oops, I'll add those in.  Carlos and I also discussed replacing these with upstart scripts - which would probably be better.

brianhw:
LGTM :+1:

rocha:
:rocket:

PR #478, opened 2013-11-21
Renaming local modules to not overwrite the ansible's library names

carsongee:
I ran into some behavioral issues with the newer versions of libraries.  I like the idea of having a local library that we can update as needed, but I don't think we should override the installed versions so I changed their names.

Sorry about all the line ending changes. I think those just happened when editing the files, but they should be innocuous


jarv:
lgtm

PR #482, opened 2013-11-21
Stanford 'certificates' stopgap branch

jrbl:
Substantially similar to the certs role developed (in paralell) that
exists on master, but this one uses pre-Croissant directory structure
and variable conventions. Modeled on the croissant-era notifier role.
Useful as a bridge to midterms, when Stanford can afford to port to the
new ways of doing things.

Includes updates to logging-related tasks for ansible and datadog. These
are ongoing problems for us at the moment.

### DIFF
--- a/playbooks/edx-west/prod-app.yml
+++ b/playbooks/edx-west/prod-app.yml
@@ -1,6 +1,7 @@
 # this gets all running prod webservers
 #- hosts: tag_environment_prod:&tag_function_webserver
 # or we can get subsets of them by name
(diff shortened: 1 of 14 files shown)


jrbl:
@jbau please take a look so I can try to run this on prod utils; this is running on stage and seems to be ok.

hkim823:
@jrbl and @jbau in an effort to clean up the PRs do we still need this opened?

jrbl:
No, we can close this. I have (separate) open tasks to port us to the Injera cert changes and get Stanford's certs repo cleaned up for publication.

PR #484, opened 2013-11-22
New role for installing and configuring haproxy

carsongee:
This can be useful for managing rabbitmq clusters and possibly
as a replacement for ELB.

### DIFF
--- a/playbooks/roles/haproxy/files/haproxy.logrotate
+++ b/playbooks/roles/haproxy/files/haproxy.logrotate
@@ -1,0 +1,13 @@
+/var/log/haproxy/*.log {
+		weekly
+		missingok
+		rotate 7
+		compress
+		delaycompress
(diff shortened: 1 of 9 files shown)


playbooks/roles/haproxy/files/haproxy.logrotate
```diff
@@ -0,0 +1,13 @@
+/var/log/haproxy/*.log {
```
feanil: Can you use the common log dir here instead of an explicit path?

carsongee: good catch

carsongee:
I switched it to use common log dir, should also make a specific logdir, or is it ok to leave haproxy as the static log subfolder?

feanil:
looks good to me as is.  :+1:

PR #488, opened 2013-11-23
add host for mysql

jbau:
set up user host for phpmyadmin control user


jbau:
small change, to edx-west/release.  pulling

PR #491, opened 2013-11-25
using role dependencies for common/supervisor

jarv:
@feanil 
@carsongee 
@e0d 
@jrbl 

This removes `- common` and `- supervisor` from our plays and adds them as role dependencies to the roles that require them.  It should mean that we won't need to include them explicitly at the play level unless you just want to install 

supervisor depends on common so roles that depend on supervisor will automatically pull in common in.


jarv:
@feanil @e0d @carsongee @jrbl - this look ok?

feanil:
looks good to me

carsongee:
This hurts me a bit since I'm using that role name and symlink it in as edx-common, but if everybody prefers this I can always rename mine, or finally get around to merging it with this one.

PR #493, opened 2013-11-25
datajam: add halt and start actions to the utility script

mulby:
Convenient if you want to temporarily shutdown the VM and start it back up later.

@rocha @brianhw


brianhw:
LGTM. :+1:

PR #497, opened 2013-11-26
Make basic_auth a global nginx parameter

jarv:
Instead of enabling basic auth per role this makes it
an nginx parameter which makes more sense for sandboxes
and probably other servers where we want basic auth turned on


jarv:
@carsongee @feanil @e0d

carsongee:
Nice, I missed the htpasswd module existing in ansible. :+1:

feanil:
:+1:

PR #502, opened 2013-11-30
add readme file for edx-west, prod-log playbook (into edx-west/release)

sefk:
For @jrbl or @jbau 

While in here do a couple of cleanups:
- remove a log file
- add a simple playbook for refreshing roles on the prod log machine

I'll do a separate one into master so we don't lose track when we merge.

### DIFF
--- a/playbooks/edx-west/README.md
+++ b/playbooks/edx-west/README.md
@@ -1,0 +1,33 @@
+This doc is a useful place to keep commands in a handy place.  Keep
+in mind that this is a public repo so we shouldn't store anything
+confidential in here.
+
+Other install docs:
+
+- Giulio's install doc [here][1].
+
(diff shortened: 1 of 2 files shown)


playbooks/edx-west/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+Ansible Commands - Stage
+------------------------
+
+Verify that you're doing something reasonable:
+
+    ANSIBLE_CONFIG=stage-ansible.cfg ANSIBLE_EC2_INI=ec2.ini ansible-playbook stage-app.yml -u ubuntu -c ssh -i ./ec2.py --list-hosts
+
+Verify that you're doing something reasonable:
```
jrbl, line 28: This stanza isn't the verification, right, it's the actual run? So perhaps "Ok, now do something reasonable:" ?

jrbl:
Other than above, lgtm and I think it should merge.

PR #503, opened 2013-11-30
add readme file for edx-west, prod-log playbook (into master)

sefk:
For @jrbl or @jbau 

While in here do a couple of cleanups:
- remove a log file
- add a simple playbook for refreshing roles on the prod log machine

Same as PR 502, but into master.

### DIFF
--- a/playbooks/edx-west/README.md
+++ b/playbooks/edx-west/README.md
@@ -1,0 +1,33 @@
+This doc is a useful place to keep commands in a handy place.  Keep
+in mind that this is a public repo so we shouldn't store anything
+confidential in here.
+
+Other install docs:
+
+- Giulio's install doc [here][1].
+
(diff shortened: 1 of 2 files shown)


playbooks/edx-west/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+Ansible Commands - Stage
+------------------------
+
+Verify that you're doing something reasonable:
+
+    ANSIBLE_CONFIG=stage-ansible.cfg ANSIBLE_EC2_INI=ec2.ini ansible-playbook stage-app.yml -u ubuntu -c ssh -i ./ec2.py --list-hosts
+
+Verify that you're doing something reasonable:
```
jrbl, line 42: same comment as #502

jrbl:
In the master branch version in particular do we want a more visible/obnoxious note indicating this directory is a convenience for us and can serve as a sample for others but probably isn't useful to others out of the box?

Other than that and the nit above, I think this is good and should merge.

sefk:
Good suggestion on the overview info @jrbl, thanks for that.  I went ahead and added.  Since it's just a README file change, I'll go ahead and pull in as you said.

PR #506, opened 2013-12-02
Add notifier configuration options

gwprice:
This change corresponds to notifier commits 73c67a9 and 88aa593.

@e0d @jimabramson


jimabramson:
:+1:

playbooks/roles/notifier/defaults/main.yml
```diff
@@ -14,6 +14,12 @@ NOTIFIER_LOG_LEVEL: "INFO"
 NOTIFIER_RSYSLOG_ENABLED: "yes"
 NOTIFIER_DIGEST_TASK_INTERVAL: "1440"
 
+NOTIFIER_DIGEST_EMAIL_SENDER: ""
```
e0d: Aren't there sensible defaults for these?  Even if they cannot actually "work", say in the case of non-addressable email addresses, they have a didactic purpose for installers.

e0d:
Other than my comment about defaults, I'm :+1:

playbooks/roles/notifier/defaults/main.yml
gwprice: Empty strings here are filtered out by the environment template, so the defaults in the notifier settings file will be used.

PR #508, opened 2013-12-02
Brian/add workbench

brianhw:
This adds a script and a port for running XBlock workbench.  Doesn't deal with thumb dependency in workbench, which needs to be installed.  @mulby


mulby:
lgtm :rocket:

PR #511, opened 2013-12-02
Add thumbs to edxapp venv so XBlock workbench will work.

brianhw:
@mulby @rocha

### DIFF
--- a/playbooks/roles/datajam/tasks/main.yml
+++ b/playbooks/roles/datajam/tasks/main.yml
@@ -24,3 +24,12 @@
     - edx-workbench-devserver
   tags:
     - deploy
+
+
+- name: datajam | install workbench XBlock dependency
+  shell: >
+     . {{ edxapp_venv_dir }}/bin/activate && cd {{ edxapp_venv_dir }}/src/xblock/thumbs && python setup.py install && deactivate
+  sudo_user: "{{ edxapp_user }}"
+  tags:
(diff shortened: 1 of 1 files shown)


mulby:
lgtm :rocket:

playbooks/roles/datajam/tasks/main.yml
```diff
@@ -24,3 +24,12 @@
     - edx-workbench-devserver
   tags:
     - deploy
+
+
+- name: datajam | install workbench XBlock dependency
+  shell: >
+     . {{ edxapp_venv_dir }}/bin/activate && cd {{ edxapp_venv_dir }}/src/xblock/thumbs && python setup.py install && deactivate
```
rocha, line 31: You can also do:
`{{ dxapp_venv_dir }}/bin/python /src/xblock/thumbs/setup.py install`

Instead of activating and deactivating the virtualenv.

rocha:
:rocket:

playbooks/roles/datajam/tasks/main.yml
mulby, line 31: I seem to remember trying something like this and it not working... maybe now that we are using the edx-platform venv it can/will work?

rocha, line 31: It is probably not worth to try if it is already working.

PR #515, opened 2013-12-03
datajam: remove virtualenv requirement

mulby:
This change has two major implications:

1) Users will no longer have to install all of the configuration repo deps, and will not need to run ansible to start up the image.
2) If we change any of the ansible stuff, we will have to generate a new image (which is a multi-hour somewhat manual process).

@brianhw, @rocha


brianhw:
So this creates the edx-configuration venv as needed, but no longer removes it.  Good to add a comment about its existence, and its reuse if it exists.  Otherwise looks good. :+1:

rocha:
LGTM :rocket:

PR #518, opened 2013-12-03
Remove MITX from the name of the features dict.

feanil:
This is to correspond to the same fix in the edx platform.
https://github.com/edx/edx-platform/pull/1827


singingwolfboy:
Looks good to me. :+1:

PR #520, opened 2013-12-04
mitxmako => edxmako

fmyzjs:
mitxmako => edxmako


feanil:
made this change as part of https://github.com/edx/configuration/pull/518

PR #527, opened 2013-12-05
Update lms.j2

fmyzjs:
for ourselves


hkim823:
@jarv or @e0d is this something we want to rename?

jarv:
Going to close until we can come to an agreement on a convention here.

PR #528, opened 2013-12-05
Update cms.j2

fmyzjs:
for ourselves


hkim823:
@jarv or @e0d is this something we want to rename?

jarv:
let's hold off on making this change

PR #529, opened 2013-12-05
datajam: ensure correct mongodb version

mulby:
The correct version of mongo was being removed automagically by apt-get when installing the insights dependencies.

@wedaly - FYI, note devstack should be fine, this was a problem with our insights role.

@brianhw, @rocha


rocha:
:rocket:

brianhw:
:+1:

PR #532, opened 2013-12-05
Disable basic auth for 127.0.0.1

wedaly:
This change disables basic auth for 127.0.0.1 so that services within a sandbox can communicate.  Basic auth is still enabled for external requests.

I tested this change in my sandbox, verifying that:
- Basic auth is still enabled for external requests
- ORA/LMS communication works with basic auth enabled (not the case before this change)

@feanil This isn't quite the change we discussed, but I think it's a lot simpler than duplicating basic auth credentials across all the roles that communicate by HTTP.


feanil:
cool :+1:

PR #533, opened 2013-12-06
datajam: remove dependency on hostsupdater

mulby:
Note that now users will have to manually include the following line in their /etc/hosts file:

192.168.33.10  preview.localhost

We will have to update the readme accordingly.

@brianhw, @rocha


jtauber:
why was the dependency removed? was it causing problems?

mulby:
Yup.  It has failed for me, Brian and Carlos intermittently.  One user has also reported it failing for her quite consistently.

See this thread: https://github.com/cogitatio/vagrant-hostsupdater/issues/16

mulby:
@wedaly - any thoughts on this?

brianhw:
Looks good to me.  (This matters if people are using preview, and for the Xblock development we hope they will be creating some kind of course in studio.)

wedaly:
Weird, I haven't had an issue with it.  I wonder if it's a Linux/Mac OS X thing?  Definitely makes sense to remove from the datajam Vagrant image, but I'd like to get it working in devstack so preview works out of the box.

PR #537, opened 2013-12-09
Configure NLTK_DATA environment variable for ORA celery workers

wedaly:
The `nltk` library uses the `NLTK_DATA` environment variable to find data, which we store in `/edx/var/ora` instead of the default location.  We had this correctly configured for ORA, but not for the ORA celery workers.  As a result, ML model creation was failing in the sandbox.


feanil:
:+1:

PR #539, opened 2013-12-10
Updated devstack role to include ORA

wedaly:
1) Adds the ORA role to devstack
2) Creates an ora_env script that activates the virtualenv and sets up other useful environment variables (such as `NLTK_DATA`)

Unfortunately, this will break the Dosa release of devstack because the Dosa Vagrantfile runs only the deploy plays.  Until the next release, you'll need to create the VM from the base Ubuntu image using `configuration/vagrant/devstack`.  You can speed this up by manually updating `configuration/vagrant/devstack` to use the Dosa base image.

I verified that all the tests ran when logged in as `ora`, but I didn't verify that ORA/LMS communicate correctly.

Suggested reviewers: @feanil @stephensanchez

### DIFF
--- a/playbooks/roles/ora/tasks/main.yml
+++ b/playbooks/roles/ora/tasks/main.yml
@@ -56,5 +56,3 @@
   with_items:
   - python
   - pip
-
-
--- a/playbooks/roles/local_dev/defaults/main.yml
+++ b/playbooks/roles/local_dev/defaults/main.yml
@@ -3,6 +3,7 @@
 localdev_accounts:
     - { user: "{{ edxapp_user}}", home: "{{ edxapp_app_dir }}" }
     - { user: "{{ forum_user }}", home: "{{ forum_app_dir }}" }
+    - { user: "{{ ora_user }}", home: "{{ ora_app_dir }}" }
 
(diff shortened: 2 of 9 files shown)


playbooks/roles/ora/tasks/main.yml
```diff
@@ -47,6 +47,15 @@
     - ora | restart ora_celery
   with_items: ora_ease_debian_pkgs
 
+- name: ora | setup the ora env
```
feanil: This should go in ora/tasks/deploy.yml and be tagged with deploy because we want to update configuration as part of deploys.

feanil:
I don't think we should be updating the dosa release stack vagrant file as that release has already happened.  We should update the base devstack image.  Then if we want faster builds we should just update the base box manually to point to the dosa box.

wedaly:
@feanil Good call; updated.  I'm creating a new devstack from the base image now to verify that things are working correctly.

feanil:
Cool, changes look good. I think you're good to merge if the vagrant image comes up cleanly and correctly.

symbolist:
@wedaly Good to know I was on the right track!

PR #540, opened 2013-12-10
datajam: update image to include latest xblock

mulby:
@rocha, @brianhw


brianhw:
:+1:   I hope it works. (I'm not trying it right now...)

PR #541, opened 2013-12-12
Created Jenkins test scripts for edx-platform

wedaly:
Install Jenkins job scripts using Ansible.

This has the following advantages:

1) Version control for the job scripts
2) Reduce duplication between the (many) Jenkins jobs
3) Make it easier for open-source contributors to set up test jobs

Suggested reviewers: @jzoldak @jarv

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/main.yml
+++ b/playbooks/roles/jenkins_worker/tasks/main.yml
@@ -6,8 +6,9 @@
 # Parameters:
 #   `jenkins_user`: jenkins
 #   `jenkins_home`: /var/lib/jenkins
-#   `jenkins_user_home`: /home/jenkins
+#   `github_oauth_token`
 
 - include: system.yml
 - include: python.yml
 - include: jscover.yml
+- include: test_scripts.yml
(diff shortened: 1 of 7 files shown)


jarv:
I definitely think we should put these scripts in git but is the configuration repo the right place? What about a test repo that the job pulls down that has the scripts?  Would we ever call edx-acceptance.sh from something other than the jenkins-worker role?

feanil:
Is there a reason not to put these in the edx-platform repo?

wedaly:
The scripts are pretty closely tied to how the Jenkins worker is provisioned.  In particular, they assume that:

1) The jenkins_env script is in a particular location
2) Wheel archives are stored in a particular location
3) They have write permissions to the home directory of the "jenkins" user
4) There is a script available for posting status back to GitHub.

Because the script templates are in the configuration repo, I can parameterize these locations and assume that things are set up correctly.  If those assumptions change, then the scripts can also change.

If, on the other hand, the scripts were stored in the edx-platform repo, then things are messier.  If any of the above assumptions change, then every branch that doesn't have the latest version of the scripts will fail in Jenkins.

wedaly:
I thought about it some more and changed my mind :)  I'll move these scripts to edx-platform.

Issue #510, opened 2013-12-02
Install from 'master' branch is failing


solitaryr:
Opened this issue under edx-platform, https://github.com/edx/edx-platform/issues/1812, but have gotten past that point now.  Now I am getting what I can only guess is an ansible issue.  I am now getting this message and am having trouble tracking down where it may be coming from:

TASK: [edxapp | openid workaround - NOT FOR PRODUCTION] **********************\* 
changed: [localhost]

TASK: [create  application config] *******************************************\* 
fatal: [localhost] => {'msg': "Encountered unknown tag 'do'.", 'failed': True}
fatal: [localhost] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': "Encountered unknown tag 'do'.", 'failed': True}]}

FATAL: all hosts have already failed -- aborting

I started this just before the T-giving break.

solitaryr:
I believe I have tracked it down to this file: playbooks/roles/edxapp/tasks/service_variant_config.yml included from playbooks/roles/edxapp/tasks/deploy.yml.  The first line has '- name: "create {{ item }} application config"'.  Since the error has a double space between "create" and "application" (which failed to show up above), it would seem that the "item" variable is not being passed successfully.  Not sure how to fix this.

solitaryr:
Also, here is the command initially used:

sudo ansible-playbook -c local --limit "localhost:127.0.0.1" /var/tmp/configuration/playbooks/edx_sandbox.yml -i "localhost," -e 'EDXAPP_PREVIEW_LMS_BASE=preview.mooc.example.com  EDXAPP_LMS_BASE=sandbox.mooc.example.com  EDXAPP_LMS_PREVIEW_NGINX_PORT=80 EDXAPP_CMS_NGINX_PORT=80 EDXAPP_LMS_NGINX_PORT=80 edx_platform_commit=master'

the limit is being rerun with --limit @/home/ubuntu/edx_sandbox.retry for subsequent runs.  It is using ansible 1.3.2 (latest available for Ubuntu 12.04)

brainstorm:
I tried making those changes since the `config.update` shouldn't require output spilled into the jinja template (by the name, it seems like it just updates environment variables?):

```
diff --git a/playbooks/roles/edxapp/templates/cms.auth.json.j2 b/playbooks/roles/edxapp/templates/cms.auth.json.j2
index 6f1859f..522770b 100644
[...]

PR #543, opened 2013-12-13
Add ports 18000 and 18010 for edx server and edx studio, fixes issue #510

brainstorm:
I was a bit confused when I could not connect to the EC2 instance after successful deployment.


feanil:
@brainstorm This change looks good.  Have you signed the edx Contributor License Aggrement(https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.md)?  Please drop a link to a signed copy of that form here and we can merge this change.

hkim823:
@brainstorm this is an old PR but we need to know if you've signed the contributor agreement.  let us know.

brainstorm:
@hkim823, Yes, I did and I sent it 4 months ago.

PR #544, opened 2013-12-13
Lowered ORA minimum submissions for ML and peer grading

wedaly:
To make testing easier in the sandbox, require only 4 essays to start ML or peer grading.

The ML grader requires at least one positive/negative example for each rubric category, but I've been able to create models for the demo course with as few as 4 essays.

@stephensanchez @jarv


stephensanchez:
:+1:

feanil:
looks good to me.

PR #545, opened 2013-12-13
Feanil/monitor repos

feanil:
Jenkins scripts for
- Monitoring a list of repositories for new or updated release candidates
- Taking a set of commits and requesting ami builds for any servers that need to be rebuilt.


jarv:
:+1:

PR #547, opened 2013-12-15
Fix DNS issue on Ubuntu 12.10

wedaly:
Fixes an issue in which DNS would not work on Ubuntu 12.10.
http://askubuntu.com/questions/238040/how-do-i-fix-name-service-for-vagrant-client

@dianakhuang @jarv


jarv:
:+1:

PR #548, opened 2013-12-15
Devstack enables prereq install by default

wedaly:
Default to enabling prereq installation in devstack.
Move `edxapp_env` file creation to the deploy steps to ensure it gets updated.

@feanil


edx-secure:
Change looks good.  I don't know what the purpose of this flag is,  would
you mind adding some documentation about it as well?

wedaly:
Added a comment about what the env variable does.  Basically, it causes the "rake" tasks to skip installing Python/Node/Ruby requirements, which makes sense when the Ansible scripts already handle this.

feanil:
:+1:

PR #552, opened 2013-12-17
Jarv/ansible

jarv:


carsongee:
Ah the ouroboros role ;)

feanil:
Why do we need an ansible role instead of a configuration role?  Shouldn't we install ansible from pip but download the configuration repo?

jarv:
after discussing this with fenail I'm going to rename it to edx_ansible

jarv:
@feanil @e0d @carsongee @jrbl @wedaly 
In this pull I am cleaning up the playbooks/ dir, removing old plays that we shouldn't need at that level.

feanil:
:+1:

PR #553, opened 2013-12-17
Feanil/jarv abbey ami

feanil:


jarv:
I think we need to broadly communicate this one, also wiki updates.

carsongee:
Why are we changing the variable name? Is this just to be more consistent with the ansible docs?

feanil:
Also to be consistent across roles.  

I'll send an e-mail to the edx-code list to make sure they see it coming and update the wiki.

carsongee:
Sounds good.  I really like the idea of having a change log directly in the repo with variable name changes or other backwards compat changes...maybe call it COMPAT.md or something.  I end up only updating my repo once every week or so and forgot the changes sometimes.  just a thought.  The wiki would probably be adequate, it just means I'll have to start going to it ;)

feanil:
@jarv Wiki updated and mail sent to edx-code.

jarv:
I think we should merge this so we can start the next release tomorrow morning :+1:

PR #559, opened 2013-12-18
switch to s3 download for nltk

jarv:


playbooks/roles/discern/tasks/deploy.yml
```diff
@@ -71,17 +71,21 @@
...
-  shell: >
-    {{ discern_venv_dir }}/bin/python -m nltk.downloader -d {{ discern_nltk_data_dir }} all
-      creates={{ discern_nltk_data_dir }}/taggers/maxent_treebank_pos_tagger/english.pickle
+- name: discern | download and install nltk
+  shell: |
+      set -e
+      curl -o {{ discern_nltk_tmp_file }} {{ discern_nltk_download_url }}
+      tar zxf {{ discern_nltk_tmp_file }}
+      rm -f {{ discern_nltk_tmp_file }}
+        creates={{ discern_nltk_data_dir }}/taggers/maxent_treebank_pos_tagger/english.pickle
```
feanil: Without some sort of version in the name or path, this will be imposible to upgrade.  Why not use the get_url module to download the file and then do the rest of the tasks if the download changed?

jarv: Yeah that was my first thought though I wasn't sure if we wanted to leave a 700MB compressed tar on disk, if we do that we can use get_url or just switch the creates to use to the path to the tar file.

jarv: Using the basename of the url which seems to work fine to indicate installation.

wedaly:
I'm getting an undefined variable error when I try to deploy this branch in Jenkins:

http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/338/console

jarv:
@wedaly  Working now

feanil:
looks good.

PR #565, opened 2013-12-20
Set devstack base box to empanada release

wedaly:
Will work once permissions on the S3 bucket are set to public.

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -21,10 +21,9 @@ end
 
 Vagrant.configure("2") do |config|
 
-  # Creates a devstack from a base Ubuntu 12.04 image
-  # TODO -- replace this with the base box that has edx_ansible and devstack installed
-  config.vm.box     = "precise64"
-  config.vm.box_url = "http://files.vagrantup.com/precise64.box"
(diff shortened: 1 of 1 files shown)


wedaly:
I got an error from the `hostsupdater` plugin that I suspect is the same one @mulby ran into with the datajam box.  I suspect this is an interaction between the shell provisioner and the plugin.  For the time being, I've removed it and will update the wiki documentation once this gets merged.

yarko:
[update: this comment was in error / ignore]
Are you really intending to take away the full build on a particular
platform away from base?

Where will people do base builds for their intended targets then?

I think you don't want to do this here, unless there is someplace else in
the tree for this which I am not aware of.

wedaly:
@yarko This change affects the Vagrantfile in `vagrant/release`, not `vagrant/base`.  Users who want to build the box from scratch can use `vagrant/base/devstack`

feanil:
looks good.

mulby:
LGTM :rocket: 

As an aside - you will likely want to update the docs to indicate that you have to manually add the entry to /etc/hosts to get preview to work properly.

yarko:
yeah, sorry @wedaly - read it wrong from mobile (565/566 confusion);  LGTM indeed!

PR #566, opened 2013-12-20
Set devstack base box to empanada release (RC)

wedaly:
Same as #565 but this one is to the release candidate


feanil:
looks good.

Issue #564, opened 2013-12-20
release/devstack fails


yarko:
These comments refer to a completely clean directory / fresh install attempt (helping someone on the google group), based on wiki instructions of this date.

This is a gaggle of issues:

RE:  @wedaly commit 853792ae8197e882b62e07b195ff567b516bc174

(These comments refer primarily to vagrant/release/devstack/Vagrantfile, but also refer to the un-reviewed wiki instructions):
- new script @ line #4 not robust; should test assumptions / existence of directory;
- line #25 comment:  "TODO: replace with base box that has edx_ansible ..."
  This is a problem / the problem:  the script assumes this state, yet this commit does not fulfill that promise.
- line #49 comment:  "... assume edx_ansible role installed..."
  This should not assume:  either confirm, or add script to test / confirm.

As is, this devstack with existing wiki instructions completely fails installation:

``` ...
[default] Booting VM...
[default] Waiting for machine to boot. This may take a few minutes...
[default] Machine booted and ready!
[default] The guest additions on this VM do not match the installed version of
VirtualBox! In most cases this is fine, but in rare cases it can
cause things such as shared folders to not work properly. If you see
shared folder errors, please update the guest additions within the
virtual machine and reload your VM.

Guest Additions Version: 4.2.0
VirtualBox Version: 4.3
[default] Configuring and enabling network interfaces...
[default] Mounting shared folders...
[default] -- /vagrant
[default] Exporting NFS shared folders...
Preparing to edit /etc/exports. Administrator privileges will be required...
[default] Mounting NFS shared folders...
[default] Running provisioner: shell...
[default] Running: inline script
stdin: is not a tty
/tmp/vagrant-shell: line 1: /edx/app/edx_ansible/venvs/edx_ansible/bin/activate: No such file or directory
/tmp/vagrant-shell: line 2: cd: /edx/app/edx_ansible/edx_ansible/playbooks: No such file or directory
/tmp/vagrant-shell: line 3: ansible-playbook: command not found
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

chmod +x /tmp/vagrant-shell && /tmp/vagrant-shell

Stdout from the command:



Stderr from the command:

stdin: is not a tty
/tmp/vagrant-shell: line 1: /edx/app/edx_ansible/venvs/edx_ansible/bin/activate: No such file or directory
[...]

PR #572, opened 2013-12-22
Print error message if edx_ansible not installed in base box

wedaly:
In response to @yarko 's suggestion for issue https://github.com/edx/configuration/issues/564:

Added a check for the `edx_ansible` role in the release version of devstack.  If the base box doesn't have `edx_ansible` installed, then fail immediately with an error message.


jarv:
:+1:

PR #574, opened 2013-12-23
Unbuffer output of provisioning scripts for devstack release

wedaly:
A user emailed me thinking that the script was hung because the output was buffered.


feanil:
looks good.

PR #577, opened 2013-12-23
Clean up edxapp defaults

wedaly:
Fixed email default var for edxapp
Removed deprecated Pearson variables


feanil:
Change looks good but will affect people who are currently using the old 'EDXAPP_SERVER_EMAIL' override.  Can you announce on edx-code mailing list about this change so they know about it?

wedaly:
@feanil I don't think EDXAPP_SERVER_EMAIL was being pulled into the settings at all -- there was a mismatch between this variable name and the one used further down in the file.  So either people were overriding `EDXAPP_DEFAULT_SERVER_NAME`, in which case this change won't matter, or people were overriding `EDXAPP_SERVER_EMAIL`, which wasn't being used anywhere.

feanil:
ah, looking at the full file, that makes sense.  cool. Good to merge then.

PR #578, opened 2013-12-24
Create edxapp_bashrc.j2

martynovp:


jarv:
this was renamed to app_bashrc, I don't think there is anything to do here @wedaly

PR #579, opened 2013-12-24
Add a way to update legacy ora configuration.

feanil:
We can't use puppet on these legacy boxes but need a simple automated way to update the configuration and restart the services.

### DIFF
--- a/playbooks/roles/legacy_ora/tasks/main.yml
+++ b/playbooks/roles/legacy_ora/tasks/main.yml
@@ -39,10 +39,10 @@
   service:
     name=edx-ora
     state=restarted
-  when: env_state.changed and auth_state.changed
+  when: env_state.changed or auth_state.changed
 
 - name: legacy_ora | restart edx-ora-celery
   service:
     name=edx-ora-celery
     state=restarted
-  when: env_state.changed and auth_state.changed
+  when: env_state.changed or auth_state.changed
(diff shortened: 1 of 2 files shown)


playbooks/roles/legacy_ora/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
+#
+# Update config for a legacy ora installation.
+#
+# This role requires that ora_app_dir and ora_user both be defined.
+# There is no default for them.
```
wedaly, line 5: It might be worth adding an assertion for this to make it easier to debug.

```
fail: legacy_ora | check that required variables are defined
when: ora_app_dir is not defined or ora_app_user is not defined
```

playbooks/roles/legacy_ora/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+  register: auth_state
+  tags:
+  - deploy
+
+# Restart ORA Services
+- name: legacy_ora | restart edx-ora
+  service:
+    name=edx-ora
+    state=restarted
+  when: env_state.changed and auth_state.changed
```
wedaly: Wouldn't we want to restart if either file changed?  If so, it might make sense to use `creates=` for the file module and then have these be handlers.

feanil:
@wedaly added asserts to test the vars we're expecting and updating the restart logic.

wedaly:
lgtm

PR #584, opened 2013-12-27
Create log directories after rsyslog is installed

tusbar:
Those directories are created with 'syslog' as owner, if rsyslog is not
installed yet, the user does not exist.


jarv:
@tusbar have you signed a contributor agreement?

tusbar:
@jarv yep

PR #585, opened 2013-12-27
Add curl to rbenv’s debian packages

tusbar:
rbenv install requires curl or wget, if none of them is installed, the
role fails.

See https://github.com/sstephenson/ruby-build/blob/master/bin/ruby-build#L203


jarv:
@tusbar are you trying to extend compatibility to a different debian variant?

tusbar:
@jarv no, I’m just installing on an ubuntu 12.04 _box_ that has almost nothing installed on it.

PR #587, opened 2013-12-27
Adds ability to have multiple named access logs be rotated

carsongee:
This should still rotate the old logs, but I've split the log files by name, so the named ones weren't getting rotated, and thus adding a wildcard.


jarv:
:+1:

PR #588, opened 2013-12-27
skeleton role for db-grader-proxy

jbau:
adds a small skeleton configuration for db-grader-proxy.  basically just captures what I did, manually.

PR to edx-west/release-merge  @sefk @jrbl


jarv:
instead of defining common vars in the db-grader-proxy role can it include common as a dependency?
Would prefer if you used the role generate script - `ansible-playbook -c local --limit "localhost," ./create_role.yml -i "localhost," -e role_name=db-grader-proxy`

jbau:
Ah. that was a big hack.  I can’t include common because we’re using this manage a server running on ubuntu 11.04, so apt-get fails =(.  I don’t think we will ever PR this into master.

jbau:
But thanks for the tip on the ansible role generate script, that’s useful.

Jason

jarv:
ah ok, I didn't see that this wasn't going into master.  
You can also do something like

```
vars_files:
  - roles/common/default/main.yml
```

in your play if you want to pull in vars in common.

-John

hkim823:
@jbau this is going into your branch, stil need this to be opened or can we close it?

PR #589, opened 2013-12-27
Make devpi optional

tusbar:
To enable, set the `devpi_enabled` variable to `"yes"`.
The vagrant devstack installs devpi by default.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -209,6 +209,8 @@ edxapp_upload_dir: "{{ edxapp_data_dir }}/uploads"
 edxapp_theme_dir: "{{ edxapp_data_dir }}/themes"
 edxapp_git_identity: "{{ edxapp_app_dir }}/edxapp-git-identity"
 edxapp_git_ssh: "/tmp/edxapp_git_ssh.sh"
+edxapp_devpi_enabled: false
+
 edxapp_workers:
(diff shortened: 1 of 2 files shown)


tusbar:
I’m not sure if it needs to be enabled in other playbooks (edx-{we,ea}st).
Please let me know.

jarv:
A few things:
(1) `edxapp_pypi_local_mirror: "http://localhost:{{ devpi_port }}/root/pypi/+simple"` is in `edxapp/defaults/main.yml` so we will need to point that to pypi as the default (assuming that devpi off is default behavior)
(2) I would set a var `edxapp_devpi_enabled` in the edxapp role, default it to false 
(3) Assuming (2) you just need to do a `when: edxapp_devpi_enabled`

tusbar:
Sorry for the delays, I’ll work on this ASAP.

I was wondering, why do you sometimes use booleans and strings ("yes" when enabled)? What’s best to use?

tusbar:
@jarv I’ve updated this. Let me know what you think. :)

hkim823:
@jarv let me know if you still want this to be merged in, seems still like a good idea

jarv:
@hkim823 @tusbar I'm going to close this because the current code on master defaults to not using devpi (though it's still installed).

tusbar:
@jarv, the point of this pull request was to be able to install edxapp without installing devpi, not to stop using it.

PR #590, opened 2013-12-27
Jarv/nginx ssl

jarv:
@carsongee 
@antoviaque
@feanil 
@e0d 

By setting NGINX_ENABLE_SSL ssl this will be enabled in
the nginx configuration for those roles that support
it, currently just the lms and cms.

To use a custom key/cert set the following vars:
- NGINX_SSL_CERTIFICATE
- NGINX_SSL_KEY

Note that there is only one ssl certificate and key
per nginx install, this is because of the way nginx
is setup (independent of the other roles).

We could enable it for different roles though they would
share the same certificate.


carsongee:
:+1:

e0d:
:+1:

PR #593, opened 2014-01-01
Jarv/raspberrypi

jarv:
This PR will also give us wheezy support since we no longer require upstart for supervisor.


hkim823:
@jarv is this still something you want to keep open?

jarv:
Going to close, though I would like to do the upstart script conversion at some point for debian support.

PR #598, opened 2014-01-04
deploy gerrit server using pre-built binaries

mulby:
Reviewers: @e0d, @gwprice, @cpennington

### DIFF
--- a/playbooks/roles/gerrit/defaults/main.yml
+++ b/playbooks/roles/gerrit/defaults/main.yml
@@ -1,0 +1,27 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role gerrit
+#
+
+gerrit_app_dir: "{{ COMMON_APP_DIR }}/gerrit"
+gerrit_data_dir: "{{ COMMON_DATA_DIR }}/gerrit"
+
+gerrit_debian_pkgs:
+  - python-mysqldb
+  - python-boto
+gerrit_release: 2.8.1
+gerrit_user: gerrit2
+gerrit_db_name: reviewdb
+gerrit_http_port: 8080
+gerrit_sshd_port: 29418
+gerrit_jre_path: /usr/lib/jvm/java-7-oracle/jre
+gerrit_java_exe_path: "{{ gerrit_jre_path }}/bin/java"
+gerrit_repo_volume_os_device: /dev/xvdf
--- a/playbooks/roles/gerrit/handlers/main.yml
+++ b/playbooks/roles/gerrit/handlers/main.yml
@@ -1,0 +1,18 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Handlers for role gerrit
+
+- name: gerrit restarted
+  service: name=gerrit state=restarted
+
+- name: nginx restarted
+  service: name=nginx state=restarted
--- a/playbooks/roles/gerrit/tasks/main.yml
+++ b/playbooks/roles/gerrit/tasks/main.yml
@@ -1,0 +1,169 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Tasks for role gerrit
+#
+# Overview: Installs and configures Gerrit on the server.  Requires
+# several secure variables to be defined that are not defined in this
+# role.
+#
+#
+# Dependencies:
+#   - An existing running database.
+#   - An S3 bucket containing all of the necessary plugin jars.
+#   - In addition to the variables defined in defaults/main.yml, the following variables must be defined:
+#
+# gerrit_github_client_id: alskdjdfkjasdjfsdlfkj
+# gerrit_github_client_secret: 0938908450deffaaa87665a555a6fc6de5777f77f
+# gerrit_db_hostname: somedb.88374jhyehf.us-east-1.rds.amazonaws.com
+# gerrit_db_admin_username: adminuser
+# gerrit_db_admin_password: adminpassword
+# gerrit_db_password: gerrituserpassword
+# gerrit_artifact_s3_bucket:
+#   name: some-s3-bucket
+#   aws_access_key_id: "{{ lookup('env', 'AWS_ACCESS_KEY_ID') }}"
+#   aws_secret_access_key: "{{ lookup('env', 'AWS_SECRET_ACCESS_KEY') }}"
+# gerrit_hostname: "gerrit.example.com"
+# gerrit_smtp_enabled: false
+# gerrit_email: gerrit@example.com
+# gerrit_smtp_server: smtp.example.com
+# gerrit_smtp_encryption: none
+# gerrit_smtp_user: someuser
+# gerrit_smtp_pass: somepassword
+#
+#
+# Example play:
+#
+# - name: Deploy gerrit
+#   hosts: gerrit
+#   gather_facts: True
+#   sudo: True
+#   roles:
+#     - gerrit
+
+- name: system package pre-requisites installed
+  apt: pkg={{ item }}
+  with_items: gerrit_debian_pkgs
+
+- name: user
+  user: name={{ gerrit_user }} system=yes home={{ gerrit_app_dir }} createhome=no
+
+- name: directories created
+  file: path={{ item }} mode=700 owner={{ gerrit_user }} state=directory
+  with_items:
+    - "{{ gerrit_app_dir }}"
+    - "{{ gerrit_app_dir }}/etc"
+    - "{{ gerrit_data_dir }}"
+
+- name: repository volume fs exists
+  shell: file -s {{ gerrit_repo_volume_os_device }} | grep ext4
+  ignore_errors: yes
+  register: is_formatted
+
+- name: repository volume formatted
+  command: mkfs -t ext4 {{ gerrit_repo_volume_os_device }}
+  when: is_formatted | failed
+
+- name: fstab includes repository volume
+  lineinfile: >
+    dest=/etc/fstab
+    regexp="^{{ gerrit_repo_volume_os_device }}\s"
+    line="{{ gerrit_repo_volume_os_device }}       {{ gerrit_data_dir }}   ext4    defaults        0       2"
+
+# output will look roughly like:
+# /dev/foo on /some/mount/point type ext4 (options)
+- name: repository volume is mounted
+  shell: >
+    mount -l | grep '^{{ gerrit_repo_volume_os_device }} '
+  ignore_errors: yes
+  register: is_mounted
+
+- name: repository volume mounted
+  command: mount {{ gerrit_repo_volume_os_device }}
+  when: is_mounted | failed
+
+- name: war file downloaded
+  s3: >
+    bucket={{ gerrit_artifact_s3_bucket.name }}
+    object=gerrit-{{ gerrit_release }}.war
+    dest=/tmp/gerrit.war
+    mode=get
+    aws_access_key="{{ gerrit_artifact_s3_bucket.aws_access_key_id }}"
+    aws_secret_key="{{ gerrit_artifact_s3_bucket.aws_secret_access_key }}"
+  sudo_user: "{{ gerrit_user }}"
+
+- name: database created
+  mysql_db: >
+    name={{ gerrit_db_name }}
+    encoding=utf8
+    login_host={{ gerrit_db_hostname }} login_user={{ gerrit_db_admin_username }} login_password={{ gerrit_db_admin_password }}
+  register: db_created
+
+- name: database user created
+  mysql_user: >
+    name={{ gerrit_user }}
+    password={{ gerrit_db_password }}
+    host='%'
+    priv="{{ gerrit_db_name }}.*:ALL"
+    login_host={{ gerrit_db_hostname }} login_user={{ gerrit_db_admin_username }} login_password={{ gerrit_db_admin_password }}
+
+- name: configuration uploaded
+  template: src=gerrit.config.j2 dest={{ gerrit_app_dir }}/etc/gerrit.config mode=600
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit restarted
+
+- name: initialized
+  command: >
+    {{ gerrit_java_exe_path }} -jar /tmp/gerrit.war init -d {{ gerrit_app_dir }} --batch --no-auto-start
+    creates={{ gerrit_app_dir }}/bin
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit restarted
+
+- name: artifacts installed from s3
+  s3: >
+    bucket={{ gerrit_artifact_s3_bucket.name }}
+    object={{ item.jar }}
+    dest={{ item.dest }}/{{ item.jar }}
+    mode=get
+    aws_access_key="{{ gerrit_artifact_s3_bucket.aws_access_key_id }}"
+    aws_secret_key="{{ gerrit_artifact_s3_bucket.aws_secret_access_key }}"
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit restarted
+  with_items:
+    - { jar: "github-oauth-{{ gerrit_release }}.jar", dest: "{{ gerrit_app_dir }}/lib" }
+    - { jar: "github-plugin-{{ gerrit_release }}.jar", dest: "{{ gerrit_app_dir }}/plugins" }
+    - { jar: "singleusergroup-{{ gerrit_release }}.jar", dest: "{{ gerrit_app_dir }}/plugins" }
+
+- name: plugins installed from war
+  shell: unzip -p /tmp/gerrit.war WEB-INF/plugins/replication.jar > {{ gerrit_app_dir }}/plugins/replication.jar creates={{ gerrit_app_dir }}/plugins/replication.jar
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit restarted
+
+- name: setup ngnix vhost
+  template: >
+    src=nginx-gerrit.j2
+    dest={{ nginx_sites_available_dir }}/gerrit
+
+- name: enable gerrit vhost
+  file: >
+    src={{ nginx_sites_available_dir }}/gerrit
+    dest={{ nginx_sites_enabled_dir }}/gerrit
+    state=link
+  notify: nginx restarted
+
+- name: init script configured
+  template: src=gerritcodereview.j2 dest=/etc/default/gerritcodereview mode=644
+
+- name: init script installed
+  file: src={{ gerrit_app_dir }}/bin/gerrit.sh dest=/etc/init.d/gerrit state=link
+
+- name: starts on boot
+  service: name=gerrit enabled=yes
--- a/playbooks/roles/gerrit/templates/gerrit.config.j2
+++ b/playbooks/roles/gerrit/templates/gerrit.config.j2
@@ -1,0 +1,41 @@
+# {{ ansible_managed }}
+
+[gerrit]
+    basePath = {{ gerrit_data_dir }}
+    canonicalWebUrl = http://{{ gerrit_hostname }}/
+    changeScreen = CHANGE_SCREEN2
+[database]
+    type = MYSQL
+    hostname = {{ gerrit_db_hostname }}
+    database = {{ gerrit_db_name }}
+    username = {{ gerrit_user }}
+    password = {{ gerrit_db_password }}
+[auth]
+    type = HTTP
+    httpHeader = GITHUB_USER
+    logoutUrl = /oauth/reset
+[sendemail]
+    enable = {{ gerrit_smtp_enabled }}
+    smtpServer = {{ gerrit_smtp_server }}
+    smtpEncryption = {{ gerrit_smtp_encryption }}
+    smtpUser = {{ gerrit_smtp_user }}
+    smtpPass = {{ gerrit_smtp_pass }}
+[container]
+    user = {{ gerrit_user }}
+    javaHome = {{ gerrit_jre_path }}
+[sshd]
+    listenAddress = *:{{ gerrit_sshd_port }}
+[httpd]
+    listenUrl = http://*:{{ gerrit_http_port }}/
+    filterClass = com.googlesource.gerrit.plugins.github.oauth.OAuthFilter
+[cache]
+    directory = cache
+[github]
+    url = https://github.com
+    clientId = {{ gerrit_github_client_id }}
+    clientSecret = {{ gerrit_github_client_secret }}
+[user]
+    email = {{ gerrit_email }}
+    anonymousCoward = Anonymous User
+[suggest]
+    accounts = true
--- a/playbooks/secure_example/vars/gerrit.yml
+++ b/playbooks/secure_example/vars/gerrit.yml
@@ -1,0 +1,18 @@
+---
+gerrit_github_client_id: alskdjdfkjasdjfsdlfkj
+gerrit_github_client_secret: 0938908450deffaaa87665a555a6fc6de5777f77f
+gerrit_db_hostname: somedb.88374jhyehf.us-east-1.rds.amazonaws.com
+gerrit_db_admin_username: adminuser
+gerrit_db_admin_password: adminpassword
+gerrit_db_password: gerrituserpassword
+gerrit_artifact_s3_bucket:
+  name: some-s3-bucket
+  aws_access_key_id: "{{ lookup('env', 'AWS_ACCESS_KEY_ID') }}"
+  aws_secret_access_key: "{{ lookup('env', 'AWS_SECRET_ACCESS_KEY') }}"
+gerrit_hostname: "gerrit.example.com"
+gerrit_smtp_enabled: false
+gerrit_email: gerrit@example.com
+gerrit_smtp_server: smtp.example.com
+gerrit_smtp_encryption: none
+gerrit_smtp_user: someuser
+gerrit_smtp_pass: somepassword
--- a/playbooks/edx-east/gerrit_deploy.yml
+++ b/playbooks/edx-east/gerrit_deploy.yml
@@ -1,0 +1,17 @@
+---
+# Deploys gerrit on to a server.
+#
+# Usage:
+#   ansible-playbook gerrit_deploy.yml -i gerrit_inventory.ini -e "secure_dir=/path/to/secure/dir"
+
+- name: Install and configure gerrit
+  hosts: gerrit
+  sudo: True
+  gather_facts: True
+  vars_files:
+    - "{{ secure_dir }}/vars/gerrit.yml"
+  pre_tasks:
+    - name: update apt
+      apt: update_cache=yes
+  roles:
+    - gerrit
--- a/playbooks/roles/gerrit/meta/main.yml
+++ b/playbooks/roles/gerrit/meta/main.yml
@@ -1,0 +1,5 @@
+---
+dependencies:
+  - role: oraclejdk
+  - role: nginx
+    nginx_sites: []
--- a/playbooks/roles/gerrit/templates/gerritcodereview.j2
+++ b/playbooks/roles/gerrit/templates/gerritcodereview.j2
@@ -1,0 +1,1 @@
+export GERRIT_SITE={{ gerrit_app_dir }}
--- a/playbooks/roles/gerrit/templates/nginx-gerrit.j2
+++ b/playbooks/roles/gerrit/templates/nginx-gerrit.j2
@@ -1,0 +1,11 @@
+server {
+  listen 80;
+  server_name {{ gerrit_hostname }};
+
+  location / {
+    proxy_pass              http://localhost:{{ gerrit_http_port }};
+    proxy_set_header        Host $host;
+    proxy_set_header        X-Real-IP $remote_addr;
+    proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
+  }
+}


carsongee:
Is there any particular reason why you don't create the RDS in the role?  Something like:

``` yaml
- name: gerrit | create RDS instance
  local_action:
    module: rds
    command: create
    db_engine: "postgres"
    engine_version: "9.3.1"
    backup_window: "{{ gerrit_rds_backup_window }}"
    backup_retention: "{{ gerrit_rds_backup_retention }}"
    instance_name: "{{ gerrit_rds_instance_name }}"
    instance_type: "{{ gerrit_rds_instance_type }}"
    multi_zone: "{{ gerrit_rds_multi_zone }}"
    size: "{{ gerrit_rds_size }}"
    region: "{{ region }}"
    username: "{{ gerrit_db_admin_username }}"
    password: "{{ gerrit_db_admin_password }}"
    security_groups: "{{ gerrit_rds_security_group_name }}"
  register: gerrit_rds_info
  tags: rds
  ignore_errors: yes
```

The only catch there is that you won't get the instance name in register variable after the first run as the module lacks a bit of smarts there.

Also, out of curiosity, have you submitted those libraries as PRs to ansible?  They would be nice to have upstream. 

You should also include the secure_dir vars folder like our other plays so that it doesn't have to defined on the command line: 

``` yaml
vars_files:
    - "{{ secure_dir }}/vars/gerrit.yml"
```

In the deployment play.

carsongee:
I should also add that you could add our nginx role and a new template to serve gerrit on port 80 through a proxy, which is usually the best practice anyway.

mulby:
@carsongee - I have a few problems getting ansible to provision the RDS instance:

1) The boto version in requirements.txt is way too old to support the postgres engine (needs to be >2.21.0)
2) Ansible has not yet released their version that supports the postgres engine.  The PR is merged but it's not included in a released version.(https://github.com/ansible/ansible/commit/1dd2510e76531571be3fad37e91cbdfc2fca8a54#diff-967225cccebe139c0719d22ba8eb84aa)
3) AFAICT the RDS module will fail if called twice using the same instance name.  The EC2 module provides the handy "id" parameter which ensures the instance is created once and then reused on subsequent calls.

I've considered moving over to MySQL just so that I can provision the RDS instance in the playbook... maybe I'll do that.  I imagine my third concern will still be valid in that case though.  Also it feels like a weak reason for choosing the database engine.

RE: the secure vars - that makes a lot of sense.  I'll make that change.

RE: nginx and port 80.  Yeah, that's more or less what the gerrit docs recommend.  I was going to look in to that once we are more sure we want to move forward with using gerrit.

RE: ansible upstream.  I've submitted the alternatives module, will likely do the rds_group one as well.  See https://github.com/ansible/ansible/pull/5502.

Thanks for taking the time to review!

playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -0,0 +1,33 @@
...
+    hostname = {{ gerrit_db_hostname }}
+    database = {{ gerrit_db_name }}
+    username = {{ gerrit_db_username }}
+    password = {{ gerrit_db_password }}
+[auth]
+    type = HTTP
+    httpHeader = GITHUB_USER
+    logoutUrl = /oauth/reset
+[sendemail]
+    smtpServer = localhost
```
e0d: Would like to make this configurable and ideally use SES here.

playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -0,0 +1,33 @@
...
+[container]
+    user = {{ gerrit_user }}
+    javaHome = /usr/lib/jvm/java-7-openjdk-amd64/jre
```
e0d: Move to config

playbooks/edx-east/gerrit_inventory.ini
```diff
@@ -0,0 +1,5 @@
+[localhost]
+127.0.0.1
+
+[gerrit]
+gerrit.analytics.edx.org ansible_ssh_user=ubuntu
```
gwprice: We shouldn't deploy using a known name, since we shouldn't have our gerrit domain name pointing to the machine until it is properly configured.

cpennington: Plus, these playbooks are public, and other people might want to use them, and won't be using our domain name.

gwprice:
I don't think we should do RDS provisioning in this playbook. The use case for this is to re-deploy our Gerrit instance in the event of a failure or upgrade, so we want to point this at an existing RDS instance containing the Gerrit database and an existing EBS volume containing the repositories.

cpennington:
If the RDS provisioning were more along the lines of "if it exists, noop, otherwise create", then I think provisioning would be nice to have here, just to make it automated from the get go.

jarv:
In general I think the best way to handle provisioning in Ansible is to keep provisioning and configuration as separate plays.  One play to setup the resources, another play that configures them.

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: Good call, I'll just pull all of these variables up into the secure config.

playbooks/edx-east/gerrit_inventory.ini
mulby: Maybe I should just remove this file?  Then it's on the user to figure out what their inventory looks like and they can use ec2.py or /etc/ansible/hosts or whatever they want.

mulby:
@jarv - I'm fine doing this.  I guess I didn't see the distinction as that important if it's all idempotent and runs relatively quickly.

playbooks/roles/gerrit/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+#
+#
+# Handlers for role gerrit
+# 
+# Overview:
+# 
+#
+- name: gerrit | restarted
+  command: sudo -u {{ gerrit_user }} {{ GERRIT_HOME }}/bin/gerrit.sh restart
```
gwprice: Instead of sudo -u, use sudo_user

playbooks/edx-east/gerrit_provision.yml
```diff
@@ -0,0 +1,91 @@
+---
+# Provisions a gerrit review server on EC2.  Note that a postgres RDS instance
```
gwprice: Given that RDS for PostgreSQL is still in Beta, we may want to consider using MySQL instead.

playbooks/roles/gerrit/handlers/main.yml
mulby: IIRC that didn't work, I'll try it again and paste the error message in to a comment here so that it's clear why I did it this non-standard way (which is what I should have done when I made the change).

playbooks/edx-east/gerrit_provision.yml
mulby: I didn't know it's still in beta... seems reasonable to MySQL.

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+#   sudo: True
+#   roles:
+#     - gerrit
+
+- name: gerrit | system package pre-requisites installed
+  apt: pkg={{ item }}
+  with_items: gerrit_debian_pkgs
+
+- name: gerrit | user
+  user: name={{ gerrit_user }} shell=/bin/bash
```
gwprice: This should probably be a system user without a shell

mulby: Probably true, this is a remnant of my manual testing when I was logging in as this user fairly frequently.

gwprice: Also, the other tasks should all just use sudo_user, which will obviate things like the owner parameter

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+
+- name: gerrit | config directory created
+  file: path={{ GERRIT_HOME }}/etc owner={{ gerrit_user }} state=directory
+
+- name: gerrit | configuration uploaded
+  template: src=gerrit.config.j2 dest={{ GERRIT_HOME }}/etc/gerrit.config owner={{ gerrit_user }}
+  notify: gerrit | restarted
+
+- name: gerrit | initialized
+  command: sudo -u {{ gerrit_user }} java -jar /home/{{ gerrit_user }}/gerrit.war init -d {{ GERRIT_HOME }} creates={{ GERRIT_HOME }}/bin
```
gwprice: Add the --batch and --no-auto-start parameters

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+    mode=get
+    aws_access_key="{{ gerrit_artifact_s3_bucket.aws_access_key_id }}"
+    aws_secret_key="{{ gerrit_artifact_s3_bucket.aws_secret_access_key }}"
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit | restarted
+  with_items:
+    - { jar: "github-oauth-{{ gerrit_github_plugin_version }}.jar", dest: "{{ GERRIT_HOME }}/lib" }
+    - { jar: "github-plugin-{{ gerrit_github_plugin_version }}.jar", dest: "{{ GERRIT_HOME }}/plugins" }
+    - { jar: "singleusergroup-{{ gerrit_singleusergroup_plugin_version }}.jar", dest: "{{ GERRIT_HOME }}/plugins" }
+    - { jar: "replication-{{ gerrit_replication_plugin_version }}.jar", dest: "{{ GERRIT_HOME }}/plugins" }
```
gwprice: The replication plugin is included with Gerrit distributions, so we don't need to build and host it. The command I have been using to install it is `unzip -p {{ GERRIT_HOME }}/gerrit.war WEB-INF/plugins/replication.jar > {{ GERRIT_HOME }}/plugins/replication.jar`

playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -0,0 +1,33 @@
+# This file was automatically generated by ansible.  Do not edit manually.
+
+[gerrit]
+    basePath = {{ GERRIT_DATA }}
+    canonicalWebUrl = http://{{ gerrit_hostname }}:8080/
```
gwprice: We should find a way to use port 80; this will get really annoying

playbooks/roles/gerrit/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+##
+# Defaults for role gerrit
+# 
+
+GERRIT_HOME: /opt/gerrit
+GERRIT_DATA: /var/lib/gerrit/repositories
+
+gerrit_debian_pkgs:
+  - git
+  - openjdk-7-jdk
```
gwprice: We are using the Oracle JDK 7 elsewhere (see the oraclejdk role); we may want to just use that to be consistent with other playbooks

playbooks/roles/gerrit/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+gerrit_version: 2.8
+gerrit_war_url: "https://gerrit-releases.storage.googleapis.com/gerrit-{{ gerrit_version }}.war"
+gerrit_user: gerrit2
+gerrit_db_username: "{{ gerrit_user }}"
+gerrit_db_name: reviewdb
+gerrit_http_port: 80
+gerrit_sshd_port: 29418
+gerrit_github_plugin_version: 2.8
+gerrit_singleusergroup_plugin_version: 1.0-SNAPSHOT
+gerrit_replication_plugin_version: v2.8
```
gwprice: Add a gerrit_email parameter

playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -0,0 +1,33 @@
...
+[httpd]
+    listenUrl = http://*:8080/
+    filterClass = com.googlesource.gerrit.plugins.github.oauth.OAuthFilter
+[cache]
+    directory = cache
+[github]
+    url = https://github.com
+    clientId = {{ gerrit_github_client_id }}
+    clientSecret = {{ gerrit_github_client_secret }}
+
```
gwprice: Add the following:

```
[user]
    email = {{ gerrit_email }}
    anonymousCoward = Anonymous User
[suggest]
    accounts = true
```

playbooks/secure_example/vars/gerrit.yml
```diff
@@ -0,0 +1,15 @@
...
+gerrit_db_hostname: somedb.88374jhyehf.us-east-1.rds.amazonaws.com
+gerrit_db_admin_username: adminuser
+gerrit_db_admin_password: adminpassword
+gerrit_db_password: gerrituserpassword
+gerrit_artifact_s3_bucket:
+  name: some-s3-bucket
+  aws_access_key_id: "{{ lookup('env', 'AWS_ACCESS_KEY_ID') }}"
+  aws_secret_access_key: "{{ lookup('env', 'AWS_SECRET_ACCESS_KEY') }}"
+gerrit_provision_keypair_name: "{{ lookup('env', 'KEYPAIR_NAME') }}"
+gerrit_route53_zone: analytics.edx.org  # FIXME: s/analytics/testeng/
```
gwprice: Change this

playbooks/roles/gerrit/templates/gerrit.config.j2
e0d: Including the nginx role as @carsongee has proposed is one option to do this easily.  We can also put this behind and ELB and terminate SSL there, but that makes the role less complete.

playbooks/secure_example/vars/gerrit.yml
e0d: I'm not sure testeng makes much sense here either, what is this for?

playbooks/secure_example/vars/gerrit.yml
```diff
@@ -0,0 +1,15 @@
+---
+gerrit_github_client_id: alskdjdfkjasdjfsdlfkj
+gerrit_github_client_secret: 0938908450deffaaa87665a555a6fc6de5777f77f
+gerrit_db_hostname: somedb.88374jhyehf.us-east-1.rds.amazonaws.com
+gerrit_db_admin_username: adminuser
+gerrit_db_admin_password: adminpassword
+gerrit_db_password: gerrituserpassword
+gerrit_artifact_s3_bucket:
+  name: some-s3-bucket
+  aws_access_key_id: "{{ lookup('env', 'AWS_ACCESS_KEY_ID') }}"
```
e0d, line 10: Cannot Gerrit using iam server roles?  This is ideal if supported because the credentials are available via the EC2 meta-data service and can be updated auto-magically.

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: Done

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: Done

playbooks/edx-east/gerrit_inventory.ini
mulby: Done

playbooks/roles/gerrit/handlers/main.yml
mulby: Done

playbooks/edx-east/gerrit_provision.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Didn't work for me, I couldn't find the plugin in both the war I built and the one posted on the official downloads page.

playbooks/roles/gerrit/defaults/main.yml
mulby: that would require pulling common etc, I thought it wasn't worth it, but could be convinced otherwise

playbooks/roles/gerrit/defaults/main.yml
mulby: Done

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: Done

playbooks/secure_example/vars/gerrit.yml
mulby: In an ideal world... all of engineering.  We are going to take it slow, though, and start with some smaller projects to see how it goes.

I figured testeng made sense since it's a code quality related service, however, I could see it living elsewhere.  Suggestions are welcome!

playbooks/secure_example/vars/gerrit.yml
mulby, line 10: Lets discuss this.  I'm sure it does have access to that stuff, I just don't know much about it.  How do you guys handle this in your other secure var files?

playbooks/roles/gerrit/defaults/main.yml
mulby: (I added it to the secure vars)

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: I'd love to set this up, was going to try to tackle SSL and this all at once after I got the basics stable.

playbooks/roles/gerrit/tasks/main.yml
gwprice: Hrm, I guess they must have changed it for 2.8. It must be somewhere in the .war file, because there is an option in the interactive init step to install the "core" plugins, but using S3 is not so terrible.

playbooks/secure_example/vars/gerrit.yml
mulby, line 10: Hrmmm.  So cloud-init requires the system python-boto to be installed, which is much too old to allow authentication via IAM.  I would have to setup a venv and install a newer boto into it in order to make this work properly, is it worth that level of effort?

mulby:
Things that are still bugging me about this PR:

1) I had to fork the ansible "rds" module to prevent it from crashing.  I assume there is some kind of incompatibility with the version of boto that is in our requirements.txt.  While I was at it, I made the creation process idempotent (kinda).
2) I'm not entirely sold on the idea that this should live in the configuration repo.  This gerrit stuff doesn't really contain anything edx specific.  Have we thought about breaking it up a bit more to create more focused isolated playbooks w/ roles?  I realize there is a reuse trade-off here but I'm a bit concerned we are creating a monolith.
3) The EC2 volume bit probably doesn't do exactly what we want.  I want it to assign some kind of canonical ID to that exact volume and re-use it if it exists (even if the instance it was associated with is terminated).  The way it is setup now, it will create a new volume whenever it creates a new instance.  Yuck.
4) Gerrit should start on boot.
5) I still need to setup some kind of proxy (probably nginx).  It's unsatisfying to me that I have to pull in all of the edX common stuff just to use the nginx role, though... I kind of wish it was self contained and we had a separate role that added in all of the edX specific stuff.  But I suppose it's probably not worth the effort to tease apart.

@jarv, @e0d, @cpennington, @gwprice - thoughts?

gwprice:
Why are you particularly reluctant to use the common role? It seems like it would make life easier for the ops team in that it provides some consistency in tooling available on the boxes they manage. There's little platform-specific stuff in there that I see.

mulby:
So I removed all of the provisioning stuff from this PR.  In order to make it work properly I had forked 3 core ansible modules, which doesn't really seem like it's worth it.  Now this just contains the deployment stuff.

playbooks/roles/gerrit/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
...
+#
+##
+# Defaults for role gerrit
+#
+
+GERRIT_HOME: /opt/gerrit
+GERRIT_DATA: /var/lib/gerrit/repositories
+
+gerrit_debian_pkgs:
+  - git
```
gwprice: This is included in common

playbooks/roles/gerrit/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
...
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role gerrit
+#
+
+GERRIT_HOME: /opt/gerrit
+GERRIT_DATA: /var/lib/gerrit/repositories
```
gwprice: Is there any particular reason these are in caps?

playbooks/roles/gerrit/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
...
+
+gerrit_debian_pkgs:
+  - git
+  - python-mysqldb
+  - python-boto
+gerrit_version: 2.8
+gerrit_war_url: "https://gerrit-releases.storage.googleapis.com/gerrit-{{ gerrit_version }}.war"
+gerrit_user: gerrit2
+gerrit_db_username: "{{ gerrit_user }}"
```
gwprice: This seems unnecessary

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+# Tasks for role gerrit
+#
+# Overview: Installs and configures Gerrit on the server.  Requires
+# several secure variables to be defined that are not defined in this
+# role.
+#
+#
+# Dependencies:
+#   - An existing running postgres database.
+#   - An S3 bucket containing all of the necessary plugin jars.
```
gwprice: It would be nice to also document the variables that have to be set to use this role

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+
+- name: gerrit | database user created
+  mysql_user: >
+    name={{ gerrit_db_username }}
+    password={{ gerrit_db_password }}
+    host='%'
+    priv="{{ gerrit_db_name }}.*:ALL"
+    login_host={{ gerrit_db_hostname }} login_user={{ gerrit_db_admin_username }} login_password={{ gerrit_db_admin_password }}
+
+- name: gerrit | config directory created
```
gwprice: Fold this into the directory creation step above

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+  template: src=gerrit.config.j2 dest={{ GERRIT_HOME }}/etc/gerrit.config mode=700
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit | restarted
+
+- name: gerrit | initialized
+  command: >
+    {{ gerrit_java_exe_path }} -jar /tmp/gerrit.war init -d {{ GERRIT_HOME }} --batch --no-auto-start
+    creates={{ GERRIT_HOME }}/bin
+  sudo_user: "{{ gerrit_user }}"
+  notify: gerrit | restarted
```
gwprice: Don't start gerrit until after the plugins are installed

playbooks/roles/gerrit/tasks/main.yml
gwprice: I just checked, and the official release war file includes replication.jar at the same path.

playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -0,0 +1,41 @@
+# This file was automatically generated by ansible.  Do not edit manually.
```
gwprice: Remove this

playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -0,0 +1,41 @@
...
+
+[gerrit]
+    basePath = {{ GERRIT_DATA }}
+    canonicalWebUrl = http://{{ gerrit_hostname }}:{{ gerrit_http_port }}/
```
gwprice: Remove the port

playbooks/roles/gerrit/defaults/main.yml
jarv: Gerrit should include the common role as a dependency I think, unless there is something in the common role that you don't think we need for it. 

Create `gerrit/meta/main.yml` with the following contents:

```
---
dependencies:
  - common
```

then, you might set the following default directories:

```
gerrit_app_dir: "{{ COMMON_APP_DIR }}/gerrit"
gerrit_data_dir: "{{ COMMON_DATA_DIR }}/gerrit"
```

playbooks/secure_example/vars/gerrit.yml
```diff
@@ -0,0 +1,21 @@
...
+  aws_secret_access_key: "{{ lookup('env', 'AWS_SECRET_ACCESS_KEY') }}"
+gerrit_provision_keypair_name: "{{ lookup('env', 'KEYPAIR_NAME') }}"
+gerrit_route53_zone: testeng.edx.org
```
gwprice: Change the example data to refer to example.com

playbooks/secure_example/vars/gerrit.yml
```diff
@@ -0,0 +1,21 @@
...
+gerrit_route53_name: gerrit
+gerrit_hostname: "{{ gerrit_route53_name }}.{{ gerrit_route53_zone }}"
+gerrit_smtp_enabled: false
+gerrit_email: gerrit.edx.org
```
gwprice: This should be an email address (i.e. gerrit@example.com)

playbooks/roles/gerrit/defaults/main.yml
gwprice: We're going to want the gerrit data to be on an EBS volume (since that's where the repos live)

playbooks/roles/gerrit/tasks/main.yml
mulby: Ah, you are correct, I had accidentally been using my custom built jar (again).  I've fixed this so that it grabs the plugin from the war.

playbooks/roles/gerrit/defaults/main.yml
mulby: Done

playbooks/roles/gerrit/defaults/main.yml
mulby: Done.

playbooks/roles/gerrit/defaults/main.yml
mulby: Done

playbooks/roles/gerrit/defaults/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Notifying the handler simply tells ansible that it needs to restart gerrit at the end of the play if this handler has been notified at least once.  This means that the restart won't actually happen until the end of the play.

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: I thought this was an ops best practice...

playbooks/roles/gerrit/templates/gerrit.config.j2
mulby: Done

playbooks/secure_example/vars/gerrit.yml
mulby: Done

playbooks/secure_example/vars/gerrit.yml
mulby: Done

mulby:
@e0d, @gwprice, @jarv - addressed review comments

Still trying to figure out the best way to capture the expected AWS setup, options seem to be:
- Ansible provisioning role
- cloud formation
- README file describing the hardware requirements

let me know what you think

mulby:
@e0d, @gwprice - updated to support starting gerrit on boot

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,169 @@
...
+
+- name: gerrit | user
+  user: name={{ gerrit_user }} system=yes home={{ gerrit_app_dir }} createhome=no
+
+- name: gerrit | directories created
+  file: path={{ item }} mode=700 owner={{ gerrit_user }} state=directory
+  with_items:
+    - "{{ gerrit_app_dir }}"
+    - "{{ gerrit_app_dir }}/etc"
+    - "{{ gerrit_data_dir }}"
```
gwprice, line 64: This should be done after the repo volume is mounted

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,169 @@
...
+  shell: file -s {{ gerrit_repo_volume_os_device }} | grep ext4
+  ignore_errors: yes
+  register: is_formatted
+
+- name: gerrit | repository volume formatted
+  command: mkfs -t ext4 {{ gerrit_repo_volume_os_device }}
+  when: is_formatted | failed
+
+- name: gerrit | repository mount point available
+  file: dest={{ gerrit_data_dir }} state=directory
```
gwprice: This is redundant with "directories created"

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,169 @@
...
+
+- name: gerrit | fstab includes repository volume
+  lineinfile: >
+    dest=/etc/fstab
+    regexp="^{{ gerrit_repo_volume_os_device }}"
```
gwprice: Add \s to the end of this

playbooks/roles/gerrit/tasks/main.yml
gwprice, line 64: Never mind; I was thinking this would be a directory within the mount point, but this way makes more sense.

playbooks/roles/gerrit/tasks/main.yml
```diff
@@ -0,0 +1,169 @@
...
+    line="{{ gerrit_repo_volume_os_device }}       {{ gerrit_data_dir }}   ext4    defaults        0       2"
+
+- name: gerrit | repository volume is mounted
+  shell: mount -l | grep {{ gerrit_repo_volume_os_device }}
```
gwprice: Make the regex more strict

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

playbooks/roles/gerrit/tasks/main.yml
mulby: Done

gwprice:
One more general comment: The names of the tasks and handler are past tense, whereas the other stuff I've looked at in this repo are in present tense.

mulby:
@gwprice - hrmmm since ansible is declarative (or it should be), I try to describe the state that the task ensures the node is in instead of the action that would be required to put it into that state.  "create directories" doesn't make much sense if the directories are already created, however "directories created" makes sense in both contexts.

That said, if the rest of the repo describes tasks in that way I should probably follow suit.

mulby:
@gwprice - do you think it is worth changing all of that wording, or is it OK as is?

gwprice:
I can live with it if devops can

mulby:
@gwprice - are there other changes you would like to see before approving this change?

gwprice:
Looks good to me, but I think someone from ops should give final approval

mulby:
@e0d @jarv - anything else in here you would like to see changed before I merge it?

e0d:
:+1:

mulby:
Squashed and rebased on master.  Also removed the role prefix from the task names in preparation for the ansible 1.4 upgrade.

PR #602, opened 2014-01-06
adding a one step install script

e0d:
Allows simplified installs to Vagrant via

```
wget https://raw.github.com/edx/configuration/84beaafcfe2322fd73eef11942fa40a59ba45a93/util/install/vagrant.sh -O - | bash
```

### DIFF
--- a/util/install/vagrant.sh
+++ b/util/install/vagrant.sh
@@ -21,7 +21,7 @@ fi
 sudo apt-get install -y python-pip python-apt git-core build-essential python-dev libxml2-dev libxslt-dev curl
 sudo apt-get install -y software-properties-common python-software-properties
 sudo apt-get install -y python-pip python-dev build-essential
-
+wget https://bitbucket.org/pypa/setuptools/raw/0.8/ez_setup.py -O - | sudo python
 sudo pip install --upgrade pip
 sudo pip install --upgrade virtualenv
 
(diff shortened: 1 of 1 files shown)


util/install/vagrant.sh
```diff
@@ -0,0 +1,33 @@
+##
+## Installs the pre-requisites for running edX on a single Ubuntu 12.04
+## instance.  This script is provided as a convenience and any of these
+## steps could be executed manually.  
+## 
+## Note that this script requires that you have the ability to run 
+## commands as root via sudo.  Caveat Emptor!
+##
```
jarv, line 8: as a sanity check you could so an `lsb_release -i` to check for Ubuntu or an `lsb_release -r` to check the release as well.

util/install/vagrant.sh
```diff
@@ -0,0 +1,33 @@
...
+sudo apt-get install -y python-pip python-apt git-core build-essential python-dev libxml2-dev libxslt-dev curl
+sudo apt-get install -y software-properties-common python-software-properties
+sudo apt-get install -y python-pip python-dev build-essential
+
+wget https://bitbucket.org/pypa/setuptools/raw/0.8/ez_setup.py -O - | sudo python
```
jarv, line 24: What is this for?

util/install/vagrant.sh
```diff
@@ -0,0 +1,33 @@
+##
```
jarv, line 1: maybe add a `set -e` so it doesn't fall through on failures.

jarv:
:+1:  (sigh)

PR #603, opened 2014-01-06
Feanil/lock virtualenv

feanil:
Fixes for permission issues while making virtual envs.

This also fixes the an issues seen with NPM when the registry is not explicitly set.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -474,3 +474,6 @@ edxapp_cms_variant: cms
 # Worker Settings
 worker_django_settings_module: 'aws'
 
+# NPM rc file
+edxapp_npmrc: "/usr/share/npm/npmrc"
+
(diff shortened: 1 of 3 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -474,3 +474,6 @@ edxapp_cms_variant: cms
 # Worker Settings
 worker_django_settings_module: 'aws'
 
+# NPM rc file
+edxapp_npmrc: "/usr/share/npm/npmrc"
+
```
jarv: Shouldn't we use /etc/npmrc?

```
$ cat /usr/share/npm/npmrc
# DO NOT MODIFY THIS FILE - use /etc/npmrc instead.
globalconfig=/etc/npmrc
globalignorefile=/etc/npmignore
prefix=/usr/local
```

jarv:
:+1:

PR #608, opened 2014-01-07
adding course_data_path for ora

jarv:


feanil:
Does this need to be back ported to the current ORA stage and prod settings in any way?

jarv:
I don't think so, by default ora is trying to write to the `/edx/app/ora...` directory which isn't allowed with the new permission structure. this allows us to override the location in env.json, defaulting to a dir that www-data write perms.

feanil:
:+1:

PR #612, opened 2014-01-08
Sets up codejail

jbau:
@jarv @carsongee @nedbat 

This sets up the edxapp codejail using apparmor, which was previously not done in the ansible configuration.  Comments appreciated.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -116,6 +116,13 @@ EDXAPP_GRADE_ROOT_PATH: '/tmp/edx-s3/grades'
 # Configure rake tasks in edx-platform to skip Python/Ruby/Node installation
 EDXAPP_NO_PREREQ_INSTALL: 1
 
+# whether to setup the python codejail or not
+EDXAPP_PYTHON_SANDBOX: false
+# this next setting, if true, turns on actual sandbox enforcement.  If not true,
+# it puts the sandbox in 'complain' mode, for reporting but not enforcement
+EDXAPP_SANDBOX_ENFORCE: true
+
+
 #-------- Everything below this line is internal to the role ------------
 
 #Use YAML references (& and *) and hash merge <<: to factor out shared settings
@@ -162,6 +169,14 @@ edxapp_workers:
   - queue: high_mem
     service_variant: lms
     concurrency: 2
+
+# setup for python codejail
+edxapp_sandbox_venv_dir:  '{{ edxapp_venvs_dir }}/edxapp-sandbox'
+edxapp_sandbox_user: 'sandbox'  # I think something about the codejail requires hardcoding this to sandbox:sandbox
+edxapp_sandbox_group: 'sandbox'
+
+# apparmor command
+edxapp_aa_command: "{% if EDXAPP_SANDBOX_ENFORCE %}aa-enforce{% else %}aa-complain{% endif %}"
 
 # Requirement files we explicitely
 # check for changes before attempting
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -452,9 +452,6 @@ sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/base
 sandbox_local_requirements: "{{ edxapp_code_dir }}/requirements/edx-sandbox/local.txt"
 sandbox_post_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/post.txt"
 
-#do we want to install the sandbox requirements into the regular virtual env
-install_sandbox_reqs_into_regular_venv: true
-
 edxapp_debian_pkgs:
   - npm
   # for compiling the virtualenv
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -173,7 +173,6 @@ edxapp_workers:
 # setup for python codejail
 edxapp_sandbox_venv_dir:  '{{ edxapp_venvs_dir }}/edxapp-sandbox'
 edxapp_sandbox_user: 'sandbox'  # I think something about the codejail requires hardcoding this to sandbox:sandbox
-edxapp_sandbox_group: 'sandbox'
 
 # apparmor command
 edxapp_aa_command: "{% if EDXAPP_SANDBOX_ENFORCE %}aa-enforce{% else %}aa-complain{% endif %}"
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -376,6 +376,14 @@ lms_auth_config:
 
 lms_env_config:
   <<: *edxapp_generic_env
+  'CODE_JAIL':
+    # from https://github.com/edx/codejail/blob/master/codejail/django_integration.py#L24, '' should be same as None
+    'python_bin': '{% if EDXAPP_PYTHON_SANDBOX %}{{ edxapp_sandbox_venv_dir }}/bin/python{% endif %}'
+    'limits':
+      'VMEM': 0
+      'REALTIME': 5
+    'user': '{{ edxapp_sandbox_user }}'
+
 cms_auth_config:
   <<: *edxapp_generic_auth
 cms_env_config:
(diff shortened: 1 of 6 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -194,6 +197,63 @@
...
+  notify:
+  - "edxapp | restart edxapp"
+  - "edxapp | restart edxapp_workers"
+  tags:
+  - deploy
+  - edxapp-sandbox
+
+- name: edxapp | code sandbox | Install sandbox requirements into sandbox venv
+  shell: >
+    cd {{ edxapp_code_dir }} &&
```
jarv: can use `chdir=` here

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -194,6 +197,63 @@
...
+    - "apparmor_parser -R /etc/apparmor.d/code.sandbox"
+  when: EDXAPP_PYTHON_SANDBOX
+  notify:
+  - "edxapp | restart edxapp"
+  - "edxapp | restart edxapp_workers"
+  tags:
+  - deploy
+  - edxapp-sandbox
+
+- name: edxapp | code sandbox | Install sandbox requirements into sandbox venv
```
jarv: do we want the file ownership to be edxapp here?
For our other virtualenvs we do a 
`sudo_user: "{{ edxapp_user }}"`

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -194,6 +197,63 @@
   - "edxapp | restart edxapp_workers"
   tags: deploy
 
+# The next few tasks set up the python code sandbox
+
+# need to disable this profile, otherwise the pip inside the sandbox venv has no permissions
+# to install anything
```
jarv, line 204: why not do a `sudo /usr/sbin/aa-complain /etc/apparmor.d/code.sandbox` here?

playbooks/roles/edxapp/tasks/deploy.yml
jbau: i think line 239 takes care of this? `sudo_user: "{{ edxapp_sandbox_user }}"` (just pushed, maybe you need to refresh page?)

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -194,6 +197,63 @@
...
+- name: edxapp | code sandbox | disable apparmor profile for code sandbox
+  shell: "{{ item }}"
+  with_items:
+    - "ln -s /etc/apparmor.d/code.sandbox /etc/apparmor.d/disable/"
+    - "apparmor_parser -R /etc/apparmor.d/code.sandbox"
+  when: EDXAPP_PYTHON_SANDBOX
+  notify:
```
jarv: unfortunately shell commands always report changed=true which means notifies will always trigger here. 
Maybe we don't need the notify here.  
There is also `changed_when` which is handy for testing the output of shell commands:
example:

```
shell:...
register: cmd_output
changed_when: cmd_output != ''
```

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -194,6 +197,63 @@
...
+  - "edxapp | restart edxapp_workers"
+  tags:
+  - deploy
+  - edxapp-sandbox
+
+- name: edxapp | code sandbox | re-enable apparmor profile for code sandbox
+  shell: "{{ item }}"
+  with_items:
+    - "unlink /etc/apparmor.d/disable/code.sandbox"
+    - "apparmor_parser -r /etc/apparmor.d/code.sandbox"
```
jarv: use aa-enforce here instead?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -163,6 +170,14 @@ edxapp_workers:
     service_variant: lms
     concurrency: 2
 
+# setup for python codejail
+edxapp_sandbox_venv_dir:  '{{ edxapp_venvs_dir }}/edxapp-sandbox'
+edxapp_sandbox_user: 'sandbox'  # I think something about the codejail requires hardcoding this to sandbox:sandbox
+edxapp_sandbox_group: 'sandbox'
```
jarv: assuming we just want a group that has the same name as the user you can simplify this by just creating edxapp_sandbox_user and leaving out the creation of the edxapp_sandbox_group since by default adduser will create a group with the same name and associate it with the user.

playbooks/roles/edxapp/tasks/deploy.yml
jarv: I see it, nevermind.

playbooks/roles/edxapp/tasks/python_sandbox_env.yml
```diff
@@ -0,0 +1,96 @@
...
+  user: name={{ edxapp_sandbox_user }} group={{ edxapp_sandbox_group }} shell=/bin/false home={{ edxapp_sandbox_venv_dir }}
+  when: EDXAPP_PYTHON_SANDBOX
+  notify:
+  - "edxapp | restart edxapp"
+  - "edxapp | restart edxapp_workers"
+  tags:
+  - edxapp-sandbox
+  - deploy
+
+- name: edxapp | code sandbox | Create edxapp sandbox virtual env directory
```
jarv: I am pretty sure that the pip module when given edxapp_sandbox_venv_dir will create the directory if it doesn't exist.

playbooks/roles/edxapp/tasks/python_sandbox_env.yml
```diff
@@ -0,0 +1,96 @@
+- name: edxapp | code sandbox | Create edxapp sandbox group
```
jarv: if all of these tasks have a `when: EDXAPP_PYTHON_SANDBOX` can we leave it off and just add a `when:` to the include?

playbooks/roles/edxapp/tasks/python_sandbox_env.yml
```diff
@@ -0,0 +1,96 @@
...
+  template: src=code.sandbox.j2 dest=/etc/apparmor.d/code.sandbox mode=0644 owner=root group=root
+  when: EDXAPP_PYTHON_SANDBOX
+  notify:
+  - "edxapp | restart edxapp"
+  - "edxapp | restart edxapp_workers"
+  tags:
+  - edxapp-sandbox
+  - deploy
+
+- name: edxapp | code sandbox | write out sandbox user sudoers config
```
jarv, line 25: add `validate='visudo -c -f %s'` here to check for parsing errors that could potentially lock us out.

nedbat:
I don't know Ansible, so I'm not sure how I can help....

jbau:
Maybe just a quick sanity check on the app armor profile and the sudoer file?

playbooks/roles/edxapp/templates/95-sandbox-sudoer.j2 and
playbooks/roles/edxapp/templates/code.sandbox.j2

jbau:
@jarv the latest commit should address your comments.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -194,6 +198,57 @@
...
+  shell: >
+    {{ edxapp_sandbox_venv_dir }}/bin/pip install -i {{ edxapp_pypi_local_mirror }} --exists-action w --use-mirrors -r {{ item }}
+    chdir={{ edxapp_code_dir }}
+  with_items:
+  - "{{ sandbox_local_requirements }}"
+  - "{{ sandbox_post_requirements }}"
+  sudo_user: "{{ edxapp_sandbox_user }}"
+  when: EDXAPP_PYTHON_SANDBOX
+  register: sandbox_install_output
+  changed_when: 'installed' in sandbox_install_output
```
jarv: this will throw a parse error when it's read in as yaml.
`"'installed' in sandbox_install_output"` will work

playbooks/roles/edxapp/tasks/python_sandbox_env.yml
```diff
@@ -0,0 +1,57 @@
+- name: edxapp | code sandbox | Create edxapp sandbox user
+  user: name={{ edxapp_sandbox_user }} group={{ edxapp_sandbox_user }} shell=/bin/false home={{ edxapp_sandbox_venv_dir }}
```
jarv: You don't need to specify the `group={{ edxapp_sandbox_user }}` here (only use that option for extra groups)

jbau:
@jarv thx. got those 2 in latest push

jarv:
:+1:

jbau:
FYI all: just tested this again on our stage w/o issues.

PR #613, opened 2014-01-09
set startsecs param in xqueue supervisor conf

jbau:
@jarv 

quite simple, really.  when there are no push queues the `xqueue_consumer` will exit right away.  this might fall under the 1 second threshold for supervisor to consider a start to be healthy.  (see http://supervisord.org/configuration.html#program-x-section-values)

so set it to 0 instead.


jarv:
:+1:

PR #622, opened 2014-01-11
support building gerrit and plugins

mulby:
This is what I used to build the artifacts that are currently being used to deploy gerrit.  I ran this playbook on a bare metal ubuntu precise64 vagrant image.

### DIFF
--- a/playbooks/edx-east/gerrit_build.yml
+++ b/playbooks/edx-east/gerrit_build.yml
@@ -14,7 +14,4 @@
     - name: playbook pre-requisites
       apt: pkg=curl update_cache=yes
   roles:
-    - gerrit_source
-    - gerrit_github_api_source
-    - gerrit_singleusergroup_source
     - gerrit_github_plugin_source
--- a/playbooks/edx-east/gerrit_build.yml
+++ b/playbooks/edx-east/gerrit_build.yml
@@ -1,0 +1,20 @@
+---
+# Builds gerrit (and associated plugins) and uploads the artifacts to S3
+#
+# Usage:
+#   ansible-playbook gerrit_build.yml -i gerrit_inventory.ini -e "secure_dir=/path/to/secure/dir"
+
+- name: Build gerrit
+  hosts: gerrit_build
+  sudo: True
+  gather_facts: True
+  vars_files:
+    - "{{ secure_dir }}/vars/gerrit.yml"
+  pre_tasks:
+    - name: playbook pre-requisites
+      apt: pkg=curl update_cache=yes
+  roles:
+    - gerrit_source
+    - gerrit_github_api_source
+    - gerrit_singleusergroup_source
+    - gerrit_github_plugin_source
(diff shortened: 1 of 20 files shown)


mulby:
@gwprice @e0d @cpennington @wedaly 

Not sure if ansible is the right "home" for this automation, open to suggestions.  Also, I suppose we should probably fork the repos I modified, apply the patches to the forks and then clone those instead of the current clone & patch method.

hkim823:
@mulby this is pretty old, was this solved in another way since opening?

mulby:
Nope. This is the playbook I used to build gerrit and its associated
plugins. Just never got merged in.

playbooks/edx-east/gerrit_build.yml
```diff
@@ -0,0 +1,17 @@
...
+
+- name: Build gerrit
+  hosts: gerrit_build
+  sudo: True
+  gather_facts: True
+  vars_files:
+    - "{{ secure_dir }}/vars/gerrit.yml"
+  pre_tasks:
+    - name: playbook pre-requisites
+      apt: pkg=curl update_cache=yes
```
jarv, line 15: can we move this package into the role if it's required?

playbooks/roles/buck/defaults/main.yml
```diff
@@ -0,0 +1,22 @@
...
+# Defaults for role buck
+# 
+
+buck_debian_pkgs:
+  - git
+  - ant
+  - curl
+  - zip
+
+BUCK_HOME: /opt/buck
```
jarv, line 20: nit: put the all caps vars first

jarv, line 20: I would rather see this role depend on common if it's not too much overhead and have it installed in the common app dir. (/edx/app by default)

playbooks/roles/gerrit_github_api_source/defaults/main.yml
```diff
@@ -0,0 +1,20 @@
...
+#
+##
+# Defaults for role gerrit_github_api_source
+# 
+
+gerrit_github_api_source_debian_pkgs:
+  - git
+  - maven
+
+GERRIT_GITHUB_API_SOURCE_HOME: /opt/gerrit-github-api-source
```
jarv, line 18: also would like to this to depend on common and installed in the common app dir if possible

playbooks/roles/gerrit_github_api_source/tasks/main.yml
```diff
@@ -0,0 +1,36 @@
...
+# Dependencies: None
+#
+# 
+
+- name: gerrit_github_api_source | system packages
+  apt: pkg={{ item }}
+  with_items: gerrit_github_api_source_debian_pkgs
+
+- name: gerrit_github_api_source | source checked out
+  git: repo={{ gerrit_github_api_source_repo }} dest={{ GERRIT_GITHUB_API_SOURCE_HOME }} version={{ gerrit_github_api_source_version }}
```
jarv, line 26: nit: break this into multiple lines

e0d:
@mulby we're cleaning house; should this be closed?

mulby:
sure? I guess I can always revive it if/when we start using gerrit more broadly

PR #624, opened 2014-01-13
Add nginx roles to plays and add new plays.

feanil:
Added new plays for splunkforwarder and datadog and added
nginx to the xqueue, xserver, and edxapp plays.

### DIFF
--- a/playbooks/edx-east/edxapp.yml
+++ b/playbooks/edx-east/edxapp.yml
@@ -9,5 +9,4 @@
       nginx_sites:
         - lms
         - cms
-        - lms-preview
     - edxapp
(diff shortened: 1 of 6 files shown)


playbooks/edx-east/edxapp.yml
```diff
@@ -5,4 +5,9 @@
   vars_files:
     - ["{{ secure_vars }}", "dummy.yml"]
   roles:
+    - role: nginx
+      nginx_sites:
+        - lms
+        - cms
+        - lms-preview
```
jarv: don't need lms-preview

jarv:
:+1:  should probably do the others too, certs, ora, discern

feanil:
@jarv  Updated but I didn't see an nginx template for the certs role.

jarv:
https://github.com/edx/configuration/blob/master/playbooks/edx-east/certs.yml
:+1:

PR #625, opened 2014-01-13
Remove non-existent 'splunk' group when creating user

tusbar:
When 'splunk' group doesn’t exist yet, ansible fails when creating the
user. Leaving it empty will create the group anyway.


e0d:
This LGTM

PR #626, opened 2014-01-13
CMS should not be default.

feanil:
Since CMS and LMS share the same port by default, both cannot be the default app.

### DIFF
--- a/playbooks/edx-east/deploy_nginx_all_roles.yml
+++ b/playbooks/edx-east/deploy_nginx_all_roles.yml
@@ -16,3 +16,5 @@
       - ora
       - xqueue
       - xserver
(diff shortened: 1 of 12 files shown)


playbooks/roles/elasticsearch/templates/elasticsearch.yml.j2
```diff
@@ -0,0 +1,30 @@
+
```
e0d: Did this bleed in?

feanil: Yea, looks like it is.

e0d:
:+1:

PR #628, opened 2014-01-13
Update edx-west plays with new nginx variables.

feanil:
@jrbl  @jbau @sefk 

Updating the edx-west plays for  a new variable that determines which nginx app to set as the default app if multiple apps use the same port.


sefk:
Seems innocuous enough.  @jbau and/or @jrbl

jbau:
yeah, seems fine.

PR #631, opened 2014-01-13
Added mirrored queues for clusters and VHOST support

carsongee:
This enables the creation of multiple vhosts, but is quite naive in its auth work of just creating the same users in each virtual host. It also adds an HA policy that makes the queues mirrored across cluster members (if there is a cluster).


e0d:
:+1:

PR #633, opened 2014-01-14
Splunk improvements

tusbar:


playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -87,7 +87,7 @@
 # Drop template files.
 - name: splunkforwarder | drop input configuration
   template:
-    src=opt/splunkforwarder/etc/system/local/inputs.conf.j2
```
e0d: We've standardized on using the full path to the location the file will ultimately be placed in the template directory.  We haven't universally fixed all older templates, but we don't want to remove it where we have it in place.

feanil:
@tusbar  the reason for the seperation of the splunk package location and splunk package URL is so that we can easily seperate the location of the deb and the name/version of it.  This ensures that geturl will only run once per splunkforwarder version.

tusbar:
@feanil I don’t mind the separation, there still are two variables in this pull request, but the URL doesn’t have to end with the deb’s name (and that’s not the case when you’re downloading from Splunk’s website).

playbooks/roles/splunkforwarder/tasks/main.yml
tusbar: Alright. It just seemed to contradict the fact that there is a variable to specify splunkforwarder’s installation path.

hkim823:
@e0d and @feanil cleaning up older PRs that fell threw the cracks, any thoughts on this?

e0d:
I'm fine with it, but it will definitely need a rebase.

feanil:
Agreed.

tusbar:
@e0d rebased.

PR #640, opened 2014-01-15
Improving the sandbox update script

jarv:
This modifies the edx_ansible role to dump a clean var
file on disk that can be used for subsequent update runs
so that config vars will not be lost


e0d:
:+1:

PR #643, opened 2014-01-15
E0d/forum updates

e0d:
This pull updates vars for the forum role to make config work better for our "prod like" deployments.  It also adds nginx configuration for the forum for running the service in AWS.

This has been tested in the sandbox.

http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/477/console


playbooks/roles/forum/defaults/main.yml
```diff
@@ -15,33 +15,33 @@ FORUM_MONGO_USER: "cs_comments_service"
 FORUM_MONGO_PASSWORD: "password"
 FORUM_MONGO_HOST: "localhost"
 FORUM_MONGO_PORT: "27017"
+FORUM_MONGO_DATABASE: "cs_comments_service"
+FORUM_MONGO_URL: "mongodb://{{ FORUM_MONGO_USER }}:{{ FORUM_MONGO_PASSWORD }}@{{ FORUM_MONGO_HOST }}:{{ FORUM_MONGO_PORT }}/{{ FORUM_MONGO_DATABASE }}"
```
jarv: maybe `FORUM_MONGO_URL` should be lowercase if we are more likely to override the individual vars, I also wouldn't be opposed to getting rid of the other vars and just using `FORUM_MONGO_URL`

playbooks/roles/forum/defaults/main.yml
```diff
@@ -15,33 +15,33 @@ FORUM_MONGO_USER: "cs_comments_service"
...
 forum_source_repo: "https://github.com/edx/cs_comments_service.git"
 forum_version: "HEAD"
-forum_mongo_database: "cs_comments_service"
-forum_mongo_url: "mongodb://{{ FORUM_MONGO_USER }}:{{ FORUM_MONGO_PASSWORD }}@{{ FORUM_MONGO_HOST }}:{{ FORUM_MONGO_PORT }}/{{ forum_mongo_database }}"
-forum_rack_env: "development"
-forum_api_key: "password"
-forum_elasticsearch_host: "localhost"
-forum_elasticsearch_port: "9200"
-forum_elasticsearch_url: "http://{{ forum_elasticsearch_host }}:{{ forum_elasticsearch_port }}"
+forum_unicorn_port: "4567"
```
jarv, line 47: maybe use a port number that's closer to the existing ports we use for gunicorn?

playbooks/roles/forum/defaults/main.yml
feanil: @e0d  was explaining this to me earlier.  The other vars are used during the testing that happens later to make sure the various forum services are up.

jarv: makes sense

jarv: should be lowercase then?

jarv:
:+1:

playbooks/roles/forum/defaults/main.yml
feanil: Don't think so, we'll need to override this for replicasets.  @e0d  In the case of replica sets would the just test the primary?

e0d: At the moment, they really all need to be set until we enhance how we build the mongo url  The URL building code needs to support replica sets and doesn't currently.  The tests need single hostnames.

playbooks/roles/forum/defaults/main.yml
e0d, line 47: That's the sinatra default which I haven's changed, I've just removed the hard-code from a template.

feanil:
:+1:

feanil:
:+1:

PR #644, opened 2014-01-15
Removing vestigial fabric files

jbau:
@jarv we never got around to cleaning out the fabric stuff we used as a stopgap for releases.

This `git rm`s the `fabfile` and `version-script` dirs, and also `package_data.yaml`


jarv:
:+1:

PR #651, opened 2014-01-17
updates for notifier from stanford install

jbau:
A few updates from Stanford's install of notifier:
- makes logo parameterizable
- corrects permission on sqlite db (writeable by both `www-data` and `notifier` users)
- sets ownership of `bin` dir

@jarv @feanil @e0d


playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -30,6 +30,16 @@
   tags:
   - deploy
 
+- name: notifier | fix permissions on notifer db file
```
jarv, line 36: Is this to workaround an issue where the permissions weren't set properly to begin with?
If that's the case can you add a comment?

jbau:
@jarv I changed how I fixed permissions on the notifier db and added a comment.  Problem with the original deploy.task was that it could complete but leave a `notifier.db` with size 0 (failed syncdb) owned by `www-data:www-data`.  This now fixes both the permissions, and runs syncdb properly (it didn't have the right environment before)

jarv:
:+1:

PR #653, opened 2014-01-17
replacing deploy tags with taggged include

jarv:


carsongee:
Didn't know that trick, cool. :+1:

jarv:
Yeah I've also seen the pattern where in `main.yml` you have a single include

```
---
include <role>.yml tags=<role>
```

Which automatically tags every task in the role with the name of the role.

carsongee:
That is a cool pattern, can you do it with roles too? so something like:

``` yaml
- {role: edxapp, tags: edxapp }
```

so you can run specific roles of a play with tags instead of messy per task tags or having lots of plays?  I know that works with when:

jarv:
yeah you can do that as well at the play level. The only caveat is is that I don't believe the dependencies will be run since they won't have the tag.

e0d:
:+1:

feanil:
looks good to me.

PR #654, opened 2014-01-17
Feanil/pass mongo

feanil:


util/vpc-tools/abbey.py
```diff
@@ -71,19 +71,18 @@ def update_deployment(self, ami):
...
-        }
-        update = query.copy()
-        pprint(update)
-        update['plays'][args.play]['amis'][args.environment] = ami
-        self.mongo_deployment.update(query, update, True)
+        query = { '_id': args.jenkins_build }
+        deployment = self.mongo_deployment.find_one(query)
+        try:
+            deployment['plays'][args.play]['amis'][args.environment] = ami
+        except KeyError as e:
```
jarv, line 78: So does this mean you can't run the script against an empty db with the noop flag now?

feanil, line 78: Good point, yea, the DB would have to have a release before we can update it.  I'll add a conditional for noop that will write the new document without any checks if that makes sense?

jarv, line 78: yeah that's fine, though is there any harm in adding in keeping the query like you have it but including the structure in the update?

jarv:
:+1:

util/vpc-tools/abbey.py
feanil, line 78: Yea, if I always include the structure, we lose other parts of the top level dictionary.

```
Given: 
hash = {'a': {
  'b': 'c',
  'd': 'e',
  }
}

Running an update with the dict {'a': {'b':'f'}} would result in us losing the value of hash['a']['d']

hash= {'a': {
  'b': 'f',
  },
}
```

PR #658, opened 2014-01-19
Jarv/no deploy

jarv:
This combines the other PR to remove the deploy tags, replacing them a `- include: deploy.yml tags=deploy`. This also makes some additional tweaks so we can now run `--skip-tags deploy` on a fresh ubuntu 12.04 server which will mean that all the `main.yml` tasks will install without running any of the `deploy.yml` tasks. To do this there is a `<role>_installed` fact set at the end of deploy.yml that needs to be defined for notification handlers.


util/jenkins/ansible-provision.sh
```diff
@@ -75,7 +75,7 @@ if [[ -z $instance_type ]]; then
   if [[ $server_type == "full_edx_installation" ]]; then
     instance_type="m1.medium"
   elif [[ $server_type == "ubuntu_12.04" ]]; then
-    instance_type="m1.small"
```
e0d, line 78: Why not a parameter?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -325,4 +295,5 @@
   when: celery_worker is defined and not devstack
   with_items: edxapp_workers
   sudo_user: "{{ supervisor_service_user }}"
-  tags: deploy
+
+- set_fact: edxapp_installed=true
```
e0d, line 299: Seems like three service do this now.  Should this be universalized?

util/jenkins/ansible-provision.sh
jarv, line 78: I'll remove this logic since I want to default to m1.medium for both server types.

carsongee, line 78: Don't you want m3.medium now ;)

playbooks/roles/edxapp/tasks/deploy.yml
jarv, line 299: universalized how?

PR #660, opened 2014-01-20
forum-nginx-sites-playbook: Add missing forum config in nginx_sites

antoviaque:
Some playbooks are missing the new forum nginx reverse proxy configuration
file, which prevents the LMS to access the forum service.


feanil:
looks good.  @e0d can you take a look and merge if you agree.

e0d:
LGTM, will merge.

PR #662, opened 2014-01-21
E0d/forum vars

e0d:
Removes trailing whitespace in forum_env output.  Also adds default var for mongodb url tags query argument.


feanil:
looks good.

PR #665, opened 2014-01-21
Adds upstream address and request time to ssl logging

carsongee:
It turns out when using nginx as a load balancer it is handy to know which app server it is sending requests too. Go figure.


feanil:
looks good to me.

PR #668, opened 2014-01-22
Simple play for generating config locally.

e0d:
ansible-playbook -i local.ini generate-config.yml --extra-vars "secure_dir=/configuration-secure/ansible" --tags config

where local.ini is

```
[localhost]
127.0.0.1
```

### DIFF
--- a/playbooks/edx-east/generate-config.yml
+++ b/playbooks/edx-east/generate-config.yml
@@ -9,7 +9,7 @@
 # config across versions of the configuration
 # repository.
 #
-# Running the play assume you have a local.ini file
+# Running the play assumes you have a local.ini file
 # with the following content.
 #
 # [localhost]
--- a/playbooks/edx-east/generate-config.yml
+++ b/playbooks/edx-east/generate-config.yml
@@ -1,3 +1,26 @@
+# This play will generate local copies of edxapp service
+# variant json configuration files, exercising the complete
+# variable precedence path, and place the files in
+#
+# /tmp/$CONFIGURATION_GIT_HASH/[stage|prod]/
+#
+# This is useful for debugging variable precedence
+# and hash-merging as well as comparing generated
+# config across versions of the configuration
+# repository.
+#
+# Running the play assume you have a local.ini file
+# with the following content.
+#
+# [localhost]
+# 127.0.0.1
+#
+# The play is run as follows:
+#
+# ansible-playbook -i local.ini generate-config.yml \
+# --extra-vars "secure_dir=/path/to/configuration-secure/ansible" \
+# --tags config
+
 - name: Generate stage config
   hosts: localhost
   connection: local
--- a/playbooks/edx-east/generate-config.yml
+++ b/playbooks/edx-east/generate-config.yml
@@ -1,0 +1,39 @@
+- name: Generate stage config
+  hosts: localhost
+  connection: local
+  vars:
+    git_rev: $PIPE(git rev-parse --short HEAD)
+    edxapp_app_dir: /tmp/{{ git_rev }}/stage
+    edxapp_user: $PIPE(whoami)
+  vars_files:
+    - "{{ secure_dir }}/vars/common/common.yml"
+    - "{{ secure_dir }}/vars/stage/stage-edx.yml"
+  gather_facts: True
+  pre_tasks:
+    - file:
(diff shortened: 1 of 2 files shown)


playbooks/edx-east/generate-config.yml
```diff
@@ -0,0 +1,39 @@
+- name: Generate stage config
```
feanil, line 24: Can you add the ansible command to run as a comment at the top of this file.

jarv, line 24: maybe add a comment on what this should be used for as well

playbooks/edx-east/generate-config.yml
```diff
@@ -0,0 +1,39 @@
...
+    - "{{ secure_dir }}/vars/stage/stage-edx.yml"
+  gather_facts: True
+  pre_tasks:
+    - file:
+        path={{ edxapp_app_dir }}
+        state=directory
+      tags:
+        - config
+  roles:
+    - edxapp
```
jarv, line 42: Why create the `edxapp_app_dir` as a pre_tasks when the edxapp role will create it?
It will also create the data dir, run the common role, etc. Is that what you intend here?

e0d, line 42: I'm limiting what is run by tag, see to command above.  I could tag the command that creates the directory as well I suppose, but wanted this touch little in the roles and safely run locally.

jarv, line 42: hm, i think this will still try to notify the restart handlers when the json files are dropped then, does it?
Maybe it would be better to just bring the vars in the `vars_files` block and copy the tasks that drop the json files instead of including the role.

e0d, line 42: The handlers are exercised, but skipped:

```
NOTIFIED: [edxapp | restart edxapp] ******************************************* 
skipping: [127.0.0.1] => (item=lms)
skipping: [127.0.0.1] => (item=cms)
```

e0d, line 42: Copying the tasks is an option, but I wanted to avoid it because of the required duplication and because it wouldn't be exercising the same code path that is used in deployment.  It is simpler in some ways though.

jarv, line 42: Ah the reason why they aren't run is because we added the `edxapp_installed` fact recently that is set at the end of `deploy.yml` which is not getting set in your case.
I guess this is ok though I think if we just did something like

```

-  template: >
    src={{ item }}.env.json.j2
    dest=/var/tmp/{{ item }}.env.json
  with_items: service_variants_enabled

-   template: >
    src={{ item }}.auth.json.j2
    dest=/var/tmp}/{{ item }}.auth.json
  with_items: service_variants_enabled

```

In the play would overall be less lines of json since you won't need to worry as much about tags and overrides. Also a little worried that we lose track of this config tag and what it is used for in the role.

jarv:
actually now that I'm thinking more about this ansible_debug role might already give us what we need for this.

e0d:
Having a role to do this seems unnecessarily heavy-weight to me.  This version requires adding two tags to our roles.  It doesn't cover the remote fetching, which is a convenience, but I would like ultimately to remote the authoritative config from the hosts.

jarv:
I meant that there already is a role that does this, see https://github.com/edx/configuration/blob/master/playbooks/roles/ansible_debug/tasks/main.yml

just curious if that already meets your requirements.

I'm not worried as much about the tags as I am general tag pollution in the roles since it's easy to lose track of which ones are necessary. I guess we could add comments next to the tasks that are tagged saying why they are needed.

e0d:
Yeah, I know of that role.  It overlaps to a certain extent.  It could be altered to do the same thing.  Personally, the "feature" seems like more of a play than a role to me, but @jbau may disagree.  

I think we will need some mechanism for re-running config whether that's our typical flow or now.  Tags seemed like a pretty natural fit for that use case.

jarv:
Yeah a play seems fine for this, actually I would like to remove the ansible_debug role if we decide to go that route for cleanup if this is going to be used for the same thing, maybe move the play out of the edx-east dir in that case into the general playbook area.  Would we use this config tag for other roles too to dump the json files?

singingwolfboy:
Needs to be rebased.

PR #669, opened 2014-01-22
Added gettext to jenkins worker

wedaly:
We need `gettext` installed to run i18n tests in Jenkins: https://github.com/edx/edx-platform/pull/2253

@feanil


feanil:
:+1:

PR #671, opened 2014-01-22
removing role identifiers for ansible 1.4

jarv:


e0d:
Can you update the templates in ansible-role:

```
- name: {{ role_name }} | stub ansible task
```

Other than that LGTM.

playbooks/roles/gluster/tasks/main.yml
```diff
@@ -1,7 +1,7 @@
 ---
 # Install and configure simple glusterFS shared storage
 
-- name: gluster | all | Install common packages
```
feanil, line 4: The all/primary part should probably be left in for Gluster.

feanil:
Looks good other than the two comments above.

playbooks/roles/gluster/tasks/main.yml
jarv, line 4: I blame sed for not making non-greedy substitutions easy

carsongee, line 4: It is one of those annoying times I feel forced to use perl

jarv:
@carsongee @e0d @feanil I've verified a from-scratch install and this checks out.
OK to merge?

carsongee:
:+1:

PR #673, opened 2014-01-22
ensures sec groups is within the same VPC.

e0d:
Fix for:

```
Security group sg-5f0eef3a and subnet subnet-45c4b66d belong to different networks.
```

Needs testing from Jenkins.

### DIFF
--- a/util/vpc-tools/abbey.py
+++ b/util/vpc-tools/abbey.py
@@ -210,7 +210,7 @@ def get_instance_sec_group(vpc_id, security_group):
 
     if not security_group_id:
         print "Unable to lookup id for security group {}".format(
-            args.security_group)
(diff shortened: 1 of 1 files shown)


util/vpc-tools/abbey.py
```diff
@@ -193,27 +193,37 @@ def parse_args():
...
 
     for grp in grp_details:
-        if grp.name == args.security_group:
+        if grp.name == security_group:
             security_group_id = grp.id
             break
+
     if not security_group_id:
         print "Unable to lookup id for security group {}".format(
             args.security_group)
```
feanil: This should also be changed to just 'security_group'

jarv:
:+1:

PR #680, opened 2014-01-23
support gerrit AWS provisioning with cloud formation

mulby:
Note:
- Sets up a VPC.  Is that overkill?
- Deploys the app server and the ELB into the public subnet, RDS into the private.  I considered putting the app server in the private subnet, but that would require also running a NAT/bastion host, which seemed like a lot of overhead for a single app server.
- Removes the nginx reverse proxy since the ELB can now server that purpose.
- Completely untested.

Reviewers: @e0d, @jarv, @gwprice


gwprice:
I didn't look at the CloudFormation template at all, but the rest looks good to me

singingwolfboy:
Needs to be rebased.

jarv:
@mulby This is pretty old, please re-open after rebase if you still want to merge it

mulby:
heh, yeah I thought we already closed these.

PR #689, opened 2014-01-24
creating arg for configuration-secure and adding role path

jarv:


util/vpc-tools/abbey.py
```diff
@@ -271,9 +275,8 @@ def create_instance_args():
 $(curl http://169.254.169.254/latest/meta-data/instance-type 2>/dev/null)
 playbook_dir="$base_dir/configuration/playbooks/edx-east"
 git_repo_name="configuration"
-git_repo_secure_name="configuration-secure"
```
feanil, line 274: This name is used later so we know what directory to cd into and checkout the correct version.  See line 335

feanil:
looks good.

PR #690, opened 2014-01-24
a general role to DRY out specific roles.

e0d:


feanil:
looks good.

playbooks/roles/edx_service/meta/main.yml
```diff
@@ -0,0 +1,20 @@
...
+# Role includes for role edx_service
+# 
+# Example:
+#
+# dependencies:
+#   - {
+#   role: my_role 
+#   my_role_var0: "foo"
+#   my_role_var1: "bar"
+# }
```
jarv, line 20: maybe in a comment block include common and/or supervisor as a dependency so it's easy to comment out? Or make the default even.

playbooks/roles/edx_service/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role edx_service
+# 
+
+#
+# vars are namespace with the module name.
+#
+edx_service_role_name: edx_service
```
jarv, line 17: are we using the role name var anywhere in our existing roles right now?

playbooks/roles/edx_service/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+#
+#
+# Handlers for role edx_service
+# 
+# Overview:
+# 
+#
+- name: edx_service | notify me
+  debug: msg="stub handler"
```
jarv, line 18: maybe add the fact when condition to this,

```
when: edx_service_installed is defined
```

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -0,0 +1,49 @@
...
+
+- name: install a bunch of system packages on which edx_service relies
+  apt: pkg={{ item }} state=present
+  with_items: "{{ edx_service_name }}_debian_pkgs"
+  when: ansible_distribution in common_debian_variants
+
+- name: install a bunch of system packages on which edx_service relies
+  yum: pkg={{ item }} state=present
+  with_items: "{{ edx_service_name }}_redhat_pkgs"
+  when: ansible_distribution in common_redhat_variants
```
jarv, line 55: if we add the `when:` condition to the handler we could also include a default deploy.yml that adds the fact to indicate that it was run.

jarv:
:+1:

PR #697, opened 2014-01-27
Will/performance test sandbox

wedaly:
- Add feature flag for enabling auto_auth, which is required for JMeter load tests.
- Add `extra_vars` env variable to Jenkins provision script, which allows other Jenkins jobs to override arbitrary Ansible variables.  I use this to (a) enable auto auth, and (b) change the settings to `load_test.py`

@e0d

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -67,6 +67,7 @@ EDXAPP_FEATURES:
   PREVIEW_LMS_BASE: $EDXAPP_PREVIEW_LMS_BASE
   ENABLE_S3_GRADE_DOWNLOADS: true
   USE_CUSTOM_THEME: $edxapp_use_custom_theme
+  AUTOMATIC_AUTH_FOR_TESTING: $edxapp_enable_auto_auth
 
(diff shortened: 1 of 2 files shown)


wedaly:
I've been using this for our performance regression tests and would really like to get it merged so it stays up-to-date with the rest of the config scripts.

@e0d could you please review when you get a chance?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -67,6 +67,7 @@ EDXAPP_FEATURES:
   PREVIEW_LMS_BASE: $EDXAPP_PREVIEW_LMS_BASE
   ENABLE_S3_GRADE_DOWNLOADS: true
   USE_CUSTOM_THEME: $edxapp_use_custom_theme
+  AUTOMATIC_AUTH_FOR_TESTING: $edxapp_enable_auto_auth
```
feanil, line 70: Since this is a var that we expect to get overridden in some cases, I think it makes sense to make it an ALL CAPS var 'EDXAPP_ENABLE_AUTO_AUTH'.

jarv:
closing this in favor of https://github.com/edx/configuration/pull/944 which is more narrow in scope.

PR #699, opened 2014-01-27
Allow anonymous access to gerrit

gwprice:
Without the loginUrl setting, gerrit assumes the user is logged in; if
that is not the case, it then automatically attempts to start the OAuth
process and redirects to GitHub.

@mulby 

See https://groups.google.com/forum/#!topic/repo-discuss/fCYf4it4vsc

### DIFF
--- a/playbooks/roles/gerrit/templates/gerrit.config.j2
+++ b/playbooks/roles/gerrit/templates/gerrit.config.j2
@@ -13,6 +13,7 @@
 [auth]
     type = HTTP
     httpHeader = GITHUB_USER
+    loginUrl = /login
     logoutUrl = /oauth/reset
 [sendemail]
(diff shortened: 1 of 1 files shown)


playbooks/roles/gerrit/templates/gerrit.config.j2
```diff
@@ -13,6 +13,7 @@
 [auth]
     type = HTTP
     httpHeader = GITHUB_USER
+    loginUrl = /login
```
mulby, line 16: I just tested this and it works!

While we're at it we should add:

httpDisplayNameHeader = GITHUB_NAME
httpEmailHeader = GITHUB_EMAIL

it sounds like those will pre-populate your name and email in your gerrit user when it gets created from the github account.

mulby:
:+1:

PR #702, opened 2014-01-27
Supervisor doesn't resolve `hostname` so using ansible variable

carsongee:
I ended up with only one worker ever being able to run tasks since they were all running with the same hostname.


jarv:
:+1:

PR #708, opened 2014-01-28
Remove hash_behaviour=merge

carsongee:
I did some light testing and I don't think this is needed.  Switching the behavior on and off at least for edxapp with the deploy tag creates the same json configuration files, and I didn't think it was used anywhere else.  Having this  in has resulted in me being unable to replace the MixedModuleStore with just the MongoModuleStore in LMS since I can't get rid of the `stores` key which causes exceptions in the platform.

### DIFF
--- a/playbooks/ansible.cfg
+++ b/playbooks/ansible.cfg
@@ -5,5 +5,4 @@
 [defaults]
 
 jinja2_extensions=jinja2.ext.do
-hash_behaviour=merge
 host_key_checking = False
--- a/playbooks/edx-east/ansible.cfg
+++ b/playbooks/edx-east/ansible.cfg
@@ -5,6 +5,5 @@
 [defaults]
 
 jinja2_extensions=jinja2.ext.do
-hash_behaviour=merge
 host_key_checking=False
 roles_path=../../../ansible-roles


jbau:
Hi @carsongee,

Let’s coordinate on this one.  I think we may be relying on it, but can test to find out.

carsongee:
Yeah, it is a pretty major change, but probably the right one in the long run.

The other option I can think of is to have the xmodule django engine code try/catch options it doesn't understand.  Alternatively, if anyone knows any yaml tricks to remove keys from a hash, or ansible tricks to have a variable not be hash merged then I'd be happy to do that instead as it could be much lower impact.  Unfortunately, I ran out of my bag of tricks and google-fu on those fronts.

jarv:
:+1: I agree with this, we should not be using any hash merging since switching to the all-caps variable overrides.
@feanil you agree?

feanil:
agreed, as long as it doesn't cause issues for @jbau

jbau:
We’ll test today and report back.

Jason

sefk:
I'm the release master out here this week, so it's convenient for me to test.  I'll capture before/after configs, a la @carsongee's test, and report back.

sefk:
As expected this works fine here at Stanford.  I installed staging with `hash_behavior=merge` removed, compared configuration files, and only saw what I expected in this release:

```
for i in *.json; do echo "------ $i"; diff $i /tmp/pre-install-Jan30/$i; echo; done
------ cms.auth.json

------ cms.env.json

------ lms.auth.json

------ lms.env.json
116d115
<         "INDIVIDUAL_DUE_DATES": true, 

------ lms-preview.auth.json

------ lms-preview.env.json
```

Looks like with @jarv and @feanil's OK's you're good to click the green button @carsongee.  Thanks for checking with us.

PR #709, opened 2014-01-28
Jarv/consolidate dd splunk

jarv:


feanil:
looks good. can you also remove the existing datadog.yml and splunk.yml from the edx-east directory?

PR #711, opened 2014-01-28
Rollback boto to the latest version that works with ec2.py

feanil:


e0d:
Is this a better solution than grabbing the dev branch version of ec2.py?  I'm not sure what the bot fixes/features we'd be losing are.

feanil:
Just tried the dev ec2.py with the latest boto and that is also failing. Opened an issue on the ansible github page.

e0d:
In that case, this seems fine as long as we don't lose the ability to use ec2 roles. :+1:  if that's the case.

PR #713, opened 2014-01-28
Sandbox user is hardcoded in xserver, so setting default to value

carsongee:
This should probably get changed at some point in xserver, but for now, here is the quick default fix.  Here is the hard coded user value:
https://github.com/edx/xserver/blob/71d5423055b8670c8d5d58ac4c53167f6d757433/sandbox/sandbox.py#L30


feanil:
:+1:

PR #715, opened 2014-01-28
changing role to play

jarv:


feanil:
@jarv the vpc-tools.py script has some logic based on roles that should be updated to look for plays.

jarv:
@feanil updated

feanil:
Looks good.

PR #722, opened 2014-01-29
change to chmod as privileged user.

e0d:
Because we are chmod as a less privileged user than the owner of the .pyc files, this role files when run against an existing deployment.


carsongee:
I've run into this before too when I've run edxapp python venv commands as root.

feanil:
:+1:

jarv:
At one point I was thinking that the tasks in deploy.yml would all be done as the edxapp user which would allow us to give the user doing deploys limited sudo privs.  Unfortunately we are violating that in other places so I guess this is fine.  I've run into issues here as well when someone is running the server as root which will create pyc files owned by root which will cause subsequent runs to fail. 
:+1: :-1:

PR #724, opened 2014-01-29
E0d/automating edxapp

e0d:
Refactors and cleanup the automated role and puts the plumbing in place to allow automation of edxapp tasks.


feanil:
:+1:

playbooks/roles/edxapp/meta/main.yml
```diff
@@ -6,3 +6,8 @@ dependencies:
     rbenv_dir: "{{ edxapp_app_dir }}"
     rbenv_ruby_version: "{{ edxapp_ruby_version }}"
   - devpi
+  - role: automated
+    automated_rbash_links: "{{ edxapp_automated_rbash_links }}"
+    automated_sudoers_dest: '99-automator-edxapp'
+    automated_sudoers_template: 'roles/edxapp/templates/etc/sudoers.d/99-automator-edxapp-server.j2'
+    automated_authorized_keys: "{{ EDXAPP_AUTOMATOR_AUTHORIZED_KEYS }}"
```
feanil, line 13: We had talked about putting a when condition here.

playbooks/roles/automated/tasks/main.yml
```diff
@@ -156,36 +124,27 @@
...
-  - update
-
+    state=touch
+    
+- name: add authorized_keys
+  lineinfile:
+    line="{{ item }}" state=present
+    dest={{ automated_home }}/.ssh/authorized_keys mode=0600 
+    owner={{ automated_user }} group={{ automated_user }}
+  with_items: automated_authorized_keys
```
jarv: if an entry is removed from automated_authorized_keys will the corresponding entry be removed from the authorized_keys file?

carsongee:
I read the description, but I'm still not quite sure what automated does and why it should be a dependency for edxapp.  Mind filling me in?

playbooks/roles/edxapp/meta/main.yml
carsongee, line 13: Ah, I see that it isn't mandatory now, so I don't have to know why I want it anymore ;)

e0d:
@carsongee 

We have a number of tasks, mostly django-admin tasks, that we run on servers where the appropriate code and config is available.  One of our immediate needs is to automate running migrations when new version of the code lands.  This PR allows an explicit subset of commands to be run remotely over ssh, using rbash and sudo to conscript what can be run.

carsongee:
Ah, cool.  Thanks for the info, definitely would want that to be optional as it just got changed to be

playbooks/roles/automated/tasks/main.yml
```diff
@@ -156,36 +124,26 @@
...
-  tags:
-  - automated
-  - install
-  - update
-  
-- name: copy key to .ssh/authorized_keys
-  copy: 
-    src=home/automator/.ssh/authorized_keys
+
+- name: touch .ssh/authorized_keys
```
feanil: This doesn't need to happen now that we're building the file from a template.

carsongee:
:+1:

feanil:
:+1:

playbooks/roles/automated/tasks/main.yml
e0d: True, fixed.

PR #725, opened 2014-01-30
runs the worker role before the edxapp role

jarv:
Because the CMS uses the git hash of the edx-platform for the
static path it can be problematic if you update the edx-platform
repo for the worker AFTER you update the edx-platform repo
for edxapp.  This is because when running the edxapp role in worker mode
rake gather_assets is skipped.

This is a workaround but we may want to consider using separate
repos for worker and edxapp for situations where both are installed
on the same server.

### DIFF
--- a/playbooks/edx-east/edx_continuous_integration.yml
+++ b/playbooks/edx-east/edx_continuous_integration.yml
@@ -18,11 +18,11 @@
       - lms
     - edxlocal
     - mongo
+    - { role: 'edxapp', celery_worker: True }
     - edxapp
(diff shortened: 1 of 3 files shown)


carsongee:
Interesting, what kind of errors are you seeing?  I haven't noticed any trouble as of yet out of curiousity.
 :+1:

jarv:
The symptom I'm seeing is broken static links on the CMS (not the LMS).

carsongee:
Ah, I get it now

PR #727, opened 2014-01-30
edxapp-git-auth: allow authenticated edx-platform/edx-theme checkouts

antoviaque:
@jarv This should allow to use ssh auth conditionally on edxapp, for both the edx-platform and edx-theme repositories checkout, depending on the `EDXAPP_LOCAL_GIT_IDENTITY` variable. It's based on the same approach as in the certificate and xserver roles, ie a temporary deployment of a private key loaded through GIT_SSH.

Ticket: TKTS-799


jarv:
:+1:

antoviaque:
@jarv Added a fix - using `EDXAPP_LOCAL_GIT_IDENTITY` as a boolean expression was generating Jinja parsing errors on paths.

PR #731, opened 2014-01-31
Change how abby gets notified.

feanil:
We were previously using the jenkins api which was more complicated than the
buildByTokens plugin and its API.


jarv:
:+1:

PR #732, opened 2014-01-31
productionalization of forum role

e0d:
The PR updates the forum nginx config and the forum role to support running unicorn properly and configuring nginx to talk to unicorn over a unix socket.

This can't be merged yet as it relies on a branch of the forums code the pulls in worker config and paths for the pidfile and sock file from environment variables.


e0d:
@jbau @carsongee Please make sure to take a look at this. 

The performance test results versus Heroku are quite good.

jarv:
:+1:

carsongee:
Looks good code wise. Is there anything special we need to do to migrate from running it direct vs unicorn, or will it just switch over without hassle even though the supervisor script is changing quite a bit?

e0d:
@carsongee I think it should "just work" tm. I can test it against a sandbox instance built from master though to be certain.

I suspect this will substantially improve performance for your users.  It relies on a PR in the cs_comments_service repo:

https://github.com/edx/cs_comments_service/pull/83

feanil:
looks good.

PR #737, opened 2014-02-04
Jarv/migration config

jarv:
This will have no impact on existing deployments since the defaults for the new migration vars are the default mysql db settings.  
This also changes the migrate commands so that they use manage.py instead of django-admin.py

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -134,6 +134,14 @@ EDXAPP_AUTOMATOR_AUTHORIZED_KEYS: []
 EDXAPP_USE_GIT_IDENTITY: false
 # Example: "{{ secure_dir }}/files/git-identity"
 EDXAPP_LOCAL_GIT_IDENTITY: !!null
+# Configuration for database migration
+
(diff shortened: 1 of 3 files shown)


yarko:
I also notice you've removed "SERVICE_VARIANT=lms", use (or lack of) of SERVICE_VARIANT being an issue with rake (see https://github.com/edx/edx-platform/blob/master/rakefile  and https://github.com/edx/edx-platform/pull/2387).

What is the plan for SERVICE_VARIANT, and how it will be used?

jarv:
@yarko although this falls outside the scope of this PR an answer to your questions about `SERVICE_VARIANT`:
- it is not required when you invoke manage.py unless you want to override the default (which is lms). See the source code for `manage.py` and the `--service-variant` option which sets the `SERVICE_VARIANT` environment variable.
- We use this variable for when there are multiple instances of edx-platform (_variants_) running on the same server. It's used primarily for two things 1) logging, so that logs for cms, lms go to different directories. 2) loading the config, it is how we know what json file to load off of disk when aws.py is used.

yarko:
@jarv = thanks / wanted to clarify, understand its use.  multi- on same
server makes sense.

feanil:
looks good.

PR #738, opened 2014-02-04
E0d/es cluster

e0d:
PR to pull in a branch of @feanil 's branch with:
1. simpler template logic.
2. actual installation of the template.
3. support for /etc/default config
4. minor refactoring
5. ES upgrade to the latest production version


playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -8,12 +8,25 @@
 #   * oraclejdk
 # 
 # Example play:
-#
-#   roles:
-#   - common
-#   - oraclejdk
-#   - elasticsearch 
+# This play can be used to do a single server or clustered
```
feanil: You mean role not play?

playbooks/edx-east/elasticsearch.yml
```diff
@@ -0,0 +1,10 @@
+- hosts: tag_role_elasticsearch:&tag_environment_stage
```
jarv: If we have two deployments (like edge and edx) in the same account will this be problematic?

playbooks/roles/elasticsearch/defaults/main.yml
```diff
@@ -1,6 +1,14 @@
 ---
-
-elasticsearch_version: "0.90.2"
-elasticsearch_sha: "397227ce37f616734f39f6e93539c9eaa82bec9"
+elasticsearch_app_dir: "{{ COMMON_APP_DIR }}/elasticsearch"
+elasticsearch_data_dir: "{{ COMMON_DATA_DIR }}/elasticsearch"
+elasticsearch_log_dir: "{{ COMMON_LOG_DIR }}/elasticsearch"
+elasticsearch_version: "0.90.11"
+elasticsearch_sha: "8e81388d0ba7e427b42514d96e25ba6499024c24"
```
jarv, line 7: can we point this to a branch or tag instead?

playbooks/roles/elasticsearch/defaults/main.yml
```diff
@@ -1,6 +1,14 @@
...
 elasticsearch_file: "elasticsearch-{{ elasticsearch_version }}.deb"
 elasticsearch_url: "https://download.elasticsearch.org/elasticsearch/elasticsearch/{{ elasticsearch_file }}"
+elasticsearch_aws_stack_name: "stage-edx"
```
jarv: maybe use a more generic name here?

playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -22,15 +35,18 @@
 
 - name: install elasticsearch from local package
   shell: >
-    dpkg -i /var/tmp/elasticsearch-{{ elasticsearch_version }}.deb
+    dpkg -i  --force-confold /var/tmp/elasticsearch-{{ elasticsearch_version }}.deb
```
jarv, line 41: can we do something here so it isn't run every time, maybe only when `download elasticsearch` has changed?

playbooks/roles/elasticsearch/templates/etc/default/elasticsearch.j2
```diff
@@ -0,0 +1,38 @@
+# Run ElasticSearch as this user ID and group ID
```
jarv, line 4: can you add ansible_managed here?

playbooks/roles/elasticsearch/templates/etc/default/elasticsearch.j2
```diff
@@ -0,0 +1,38 @@
...
+#DATA_DIR=/var/lib/elasticsearch
+
+# ElasticSearch work directory
+#WORK_DIR=/tmp/elasticsearch
+
+# ElasticSearch configuration directory
+#CONF_DIR=/etc/elasticsearch
+
+# ElasticSearch configuration file (elasticsearch.yml)
+#CONF_FILE=/etc/elasticsearch/elasticsearch.yml
```
feanil: Let's put this under the '{{COMMON_CFG_DIR}}/elasticsearch' dir  which by default should be '/edx/etc/elasticsearch'

playbooks/roles/elasticsearch/templates/etc/elasticsearch/elasticsearch.yml.j2
```diff
@@ -0,0 +1,39 @@
+
```
jarv: add ansible_managed

playbooks/roles/elasticsearch/tasks/main.yml
feanil, line 41: +1 to this idea.

playbooks/edx-east/elasticsearch.yml
e0d: With this example yes.  However, the play tag will be added here when it exists.

playbooks/roles/elasticsearch/defaults/main.yml
e0d, line 7: this is the signature of the deb archive, not a git sha

playbooks/roles/elasticsearch/defaults/main.yml
e0d: This is legacy, will remove.

playbooks/roles/elasticsearch/tasks/main.yml
e0d, line 41: This take 2 seconds and runs infrequently.  The vast majority of the time this is run it will be to upgrade elasticsearch.  I'm -1 to this idea.  The code that was here simply prevented upgrades, which is why I removed it.  We could drop a file with the version in it in /edx/var/elasticsearch, which I considered, but I don't think it is worth the trouble.

feanil, line 41: since the version of the file is now part of the destination file name, downloads should happen only when new versions are available so it seems that keying off of the download would be better then re-installing.  I think it helps because the fewer things that ansible shows as always changing, the less noise there is for figuring out what happened during an ansible run.

jarv:
:+1:

PR #741, opened 2014-02-05
E0d/commoncluster

e0d:
This changes the CF template creates a new LaunchConfiguation, ASG for co-hosting Rabbit and ElasticSearch on a single server.  Additionally, there will be three instances rather than two in the cluster to avoid split brain problems.


e0d:
Accidentally merged.

feanil:
Looks good!

PR #742, opened 2014-02-05
add users without sudo (gh_user_no_sudo)

sefk:
It's handy to be able to create users that _don't_ have sudo rights. Here at Stanford we use this to add users to our VPC bastion box (ssh proxy) so these users can use this machine for ssh tunneling, but I
don't want to give those users the keys to the kingdom.

This let's me configure a playbook like this.

```
roles :
  - common
  - supervisor
  - role: gh_users
    gh_users:
      - su1
      - su2
      - su3
    gh_users_no_sudo:
      - normal1
      - normal2
```

The new gh_users_no_sudo list can be empty.

We've used this at Stanford already and it seems to be working.

### DIFF
--- a/playbooks/edx-west/prod-jumpbox.yml
+++ b/playbooks/edx-west/prod-jumpbox.yml
@@ -8,3 +8,18 @@
   roles:
     - common
     - supervisor
+    - role: gh_users
+      gh_users:
+        - sefk
+        - jbau
+        - jrbl
+        - ali123
+        - caesar2164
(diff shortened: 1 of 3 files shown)


sefk:
@feanil, can you to take a look?  I see you on a lot of commits in the roles/gh_users part of this repo.

feanil:
looks good to me.

sefk:
Thanks @feanil, I'll merge then.

PR #743, opened 2014-02-05
bug fixes relate to re-location to /edx

e0d:
- Updates the play for the new play name.
- Creates a logging config file so that ES, well, logs.
- Fully specified the config file name, which is apparently necessary.


feanil:
looks good.

PR #745, opened 2014-02-05
Feanil/proper extra vars

feanil:


util/vpc-tools/abbey.py
```diff
@@ -636,6 +640,14 @@ def launch_and_configure(ec2_args):
         extra_vars_yml = "---\n"
         extra_vars = {}
 
+    if args.refs:
+        with open(args.refs) as f:
+            git_refs_yml = f.read()
+            git_refs = yaml.load(git_refs_yml)
+    else:
+        git_refs_yml = "---\n"
```
jarv: this won't work since extra_vars_yml also sets the yml header.
I suggest setting both `git_refs_yml` and `extra_vars_yml` to an empty string and setting the yml header in the bash heredoc instead

jarv:
:+1: once the yaml header issue is resolved

jarv:
:+1:

PR #747, opened 2014-02-05
Installing the bigdesk monitoring plugin.

e0d:
Bigdesk is a nice monitoring dashboard that provides a lot of detail about cluster health, etc.

https://raw.github.com/lukas-vlcek/bigdesk/master/bigdesk-2.0.0-SNAPSHOT.jpg

### DIFF
--- a/playbooks/roles/elasticsearch/tasks/main.yml
+++ b/playbooks/roles/elasticsearch/tasks/main.yml
@@ -70,5 +70,17 @@
     mode=0744
   when: ELASTICSEARCH_CLUSTERED  
     
+  # Plugin installation fails hard when the plugin already
+  # exists. This is problematic of this is upgraded.
+- name: check if the bigdesk plugin is installed
+  command: test -d /usr/share/elasticsearch/plugins/bigdesk
+  ignore_errors: true
+  register: bigdesk_present
+  
+- name: install bigdesk plugin
+  shell: >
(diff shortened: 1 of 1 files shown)


playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -63,6 +63,18 @@
     src=edx/etc/elasticsearch/elasticsearch.yml.j2 dest={{ elasticsearch_cfg_dir }}/elasticsearch.yml
     mode=0744
   when: ELASTICSEARCH_CLUSTERED
-    
+
+  # Plugin installation fails hard when the plugin already
+  # exists. This is problematic of this is upgraded.
```
jarv: typo

playbooks/roles/elasticsearch/tasks/main.yml
```diff
@@ -63,6 +63,18 @@
...
+- name: check if the bigdesk plugin is installed
+  command: test -d /usr/share/elasticsearch/plugins/bigdesk
+  ignore_errors: true
+  register: bigdesk_present
```
jarv: would prefer to use the stat module instead, so it doesn't show a red error when it fails which will be skipped with ignore_errors but is still ugly.

feanil:
looks good.

jarv:
:+1:

PR #748, opened 2014-02-06
Changes to update dhclient search domain.

e0d:
Allows updating domain-search in /etc/dhcp/dhclient.conf.

Generally useful for AWS internal domain resolution.  Best solution for rabbitmq clustering.


feanil:
looks good.

PR #749, opened 2014-02-06
E0d/commoncluster

e0d:
Adding common vars.
Template simplification following the ElasticSearch pattern.


feanil:
Looks good.

carsongee:
I was wondering if maybe we could look at changing the way we are doing cluster a little bit again. Using hostvars.keys() returns all the hosts in the play, so if I have a play running the common cluster, and the app servers it ends up trying to add all of them to the rabbit/elasticsearch clusters.  Can we maybe just have something like:

```
{%- if RABBITMQ_CLUSTERED -%}
    {%- set hosts= [] -%}
    {%- for host in groups[RABBIT_CLUSTER_GROUP] -%}
        {% do hosts.append("rabbit@ip-" + host.replace('.','-')) %}
   {%- endfor %}
{%- endif -%}
```

?  That would make these cluster roles more flexible in usage and doesn't tie them to any particular inventory script (e.g ec2.py or nova.py).  I'll write the code if we like it, just wanted to discuss the options a bit.

feanil:
How would `RABBIT_CLUSTER_GROUP` get set?

carsongee:
In defaults.  It looks like you can even specify all, so that could be the default, and the existing use  case wouldn't require changes I don't think

carsongee:
Strike that, all is all in the real sense of all your inventoried hosts

carsongee:
Too bad the set presented by `hosts: tag_group_...` isn't stored anywhere I can seem to find...

e0d:
Yeah, I had investigated that as an option and came to the same conclusion, sorry to be late to the party.

PR #759, opened 2014-02-07
check out the release branch for devstack

jarv:


wedaly:
Makes sense, but we should probably make sure the vagrant-devstack.yml file is also using the release branch (since it includes edx_ansible)

jarv:
@wedaly not sure what you mean, do you think the vagrant-devstack.yml file should set the edx_platform_version to release as well? We were thinking it would remain master.

yarko:
... so this is a little reminiscent of hard-coded constants which need to vary, and coupling (copy/pasting) across files.  It's something I'm working to remove for a development environment which is manageable by developer and devops alike.   For example, because of the coupling which @wedaly points to, and because of the hierarchical nature by which ansible works (and other config setups in edx, e.g. `{cms, lms}/envs`) - think about how I (an independent developer) should test potential changes to the configuration repo w/ this hard-coded constants approach (`Vagrantfile` vs. a less problematic `vagrant-devstack.yml`).

At present, a developer would simply hand-modify `Vagrantfile` (and perhaps not be aware of any coupling effect from `vagrant-devstack.yml`).  Is this going to present problems?

And if update is updated to accept a `build_version.yml`,  is this going to create a race? (when would I do that?  what do I need to think about, as an independent developer?)

As I'm preparing changes to `Vagrantfile` (current state of affairs looks like https://gist.github.com/yarko/8865534 - used in conjunction with https://gist.github.com/yarko/8818633) - but consider this:

An Open edX developer, working on a non-edx.org product during development has spec'd a different repo / version for configuration.

...something about this change, it's motivation doesn't feel right...

So - as @wedaly points up, but with the more general consideration:   if this repo+branch does not match the repo+branch spec which is affected (e.g. overridden by extra-vars spec) for `vagrant-devstack.yml`  --- what is the effect?

That needs to be thought about.

yarko:
to put is more succinctly:   If you can step thru a few use-cases that show how this will affect things:
- build_version.yml used;  what does this say for this change?  when `vagrant-devstack.yml` doesn't have the change?  when it's overridden elsewhere by ansible `extra-vars`?
- developer wants to change the version (only) of configuration used - what steps should they take?
- developer wants to change version & repo of configuration - what steps do they need to take now?

wedaly:
@jarv edx_ansible is included in the vagrant-devstack role, but it doesn't override the default branch for the config repo:

https://github.com/edx/configuration/blob/master/playbooks/roles/edx_ansible/defaults/main.yml#L37

It won't make a difference for provisioning the release Vagrantfile (since you're explicitly updating to the release branch with /edx/bin/update), but could make a difference when creating the base box if release is different than master at the time of creation.

jarv:
@wedaly update vagrant-devstack so the version is set to release

wedaly:
lgtm :+1:

jarv:
@yarko PRs are not the place for these types of concerns and questions, please use the edx-code list.

PR #768, opened 2014-02-10
Feanil/pre release changes

feanil:


e0d:
:+1:

playbooks/edx-east/demo.yml
```diff
@@ -0,0 +1,15 @@
...
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    enable_datadog: True
+    enable_splunkforwarder: True
+  vars_files:
+    - ["{{ secure_vars }}", "dummy.yml"]
+  roles:
+    - demo
```
jarv, line 11: This won't work unfortunately because demo d uses the vars COMMON_APP_DIR and COMMON_GIT_MIRROR and doesn't currently have common as a dependency, maybe add it?

PR #772, opened 2014-02-11
edx-west: clean up users, remove two unused plays (*-log.yml)

sefk:
@jrbl please.  Three changes in here:
1. Added Sherif Halawa to the jumpbox, he needs prod-ro DB access
2. I got annoyed seeing inconsistent sets of users on our boxes, so cleaned up to a standard set, and
3. Noticed two plays that are no longer used, {prod,stage}-log.yml.  Those were supposed to be our logging box when we were considering writing logs to a live mongo instance.  But we never ended up doing that.  Removing.


jrbl:
lgtm.

sefk:
Thanks for the review @jrbl

PR #781, opened 2014-02-12
Kibana and logstash roles

carsongee:
These obviously are a bit old and need some work, but here are the kibana and logstash roles we've been using.


jarv:
very nice! I'll see if we can adapt this, are you running your syslog server and logstash on the same server? how large is it?

carsongee:
I run it on a medium for our production environment (and probably need to bump it), and a small seems to work fine for our smaller course development stacks.  We also attach an EBS volume to store all the logs for persistence on the provisioning side.

carsongee:
I left out the nginx template to do the proxying and auth. I've added that (it will need some adapting since it is using ssl certificates for auth)

carsongee:
I realized I only half answered your question.  Logstash is acting as my syslog server on that machine, so instead of running rsyslogd listen on 514 for logs I'm having logstash do that.  rsyslogd is still running on a socket, but forwarding to localhost:514 so local logs make it into logstash as well)

carsongee:
I've gotten this updated and think it is ready for review.  I've tested it on the vagrant boxes and all seems well.  I've added a sample play, but not sure we want to keep that in there, and I've also thrown in the general dashboard I use in Kibana to help get people started with it.

hkim823:
@jarv thoughts on merging this in?

jarv:
@carsongee sorry for not looking at this sooner. :+1:

carsongee:
No problem. I kind of forgot about it too.

PR #783, opened 2014-02-12
sync application and tracking logs to s3

jarv:
Uploads log paths to S3 every hour and on shutdown.

Configuration:

```
# Rotate logs to S3
# Only for when edX is running in AWS since it organizes
# logs by security group.
# !! The buckets defined below MUST exist prior to enabling !!
# this feature and the instance IAM role must have write permissions
# to the buckets
COMMON_S3_LOGS: false
# If there are any issues with the s3 sync an error
# log will be sent to the following address.
# This relies on your server being able to send mail
COMMON_S3_LOGS_NOTIFY_MAIL: dummy@example.com
# Separate buckets for tracking logs and everything else
# You should be overriding the environment and deployment vars
COMMON_S3_LOG_PATHS:
  - bucket: "{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-app-logs"
    paths:
    # globs are ok, do not use brace expansion
    # everything except tracking logs
    - "{{ COMMON_LOG_DIR }}/!(*tracking*)"
    - "/var/log/*"
  - bucket: "{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-tracking-logs"
    paths:
    - "{{ COMMON_LOG_DIR }}/*tracking*"
```

This generates a bash script that does the following:
- creates an hourly cron that syncs the AWS_S3_LOG_PATHS to the corresponding buckets
- runs the script on shutdown
- emails AWS_S3_LOGS_NOTIFY_MAIL on any failure with a copy of the log

Note that we probably should stagger the cron runs but there is not currently an easy way to do this in ansible.

@e0d @feanil  @hkim823  @carsongee @sefk


playbooks/roles/common/tasks/main.yml
```diff
@@ -55,6 +55,10 @@
 - name: Install logrotate configuration for edX
   template: dest=/etc/logrotate.d/edx-services src=edx_logrotate.j2 owner=root group=root mode=644
 
+  # This is in common to keep all logrotation config
+  # in the same role
+- name: Install logrotate configuration for tracking file
+  template: dest=/etc/logrotate.d/tracking.log src=edx_logrotate_tracking_log.j2 owner=root group=root mode=644
```
e0d: I thought that we had agreed to have the path of the file represented in the template directory, so that's what I've been doing.  I'd like to be consistent.

playbooks/roles/common/templates/send-logs-to-s3.j2
```diff
@@ -0,0 +1,106 @@
...
+set -e
+
+sec_grp=unset
+instance_id=unset
+s3_path=unset
+
+onerror() {
+  if [[ -z $noop ]]; then
+    echo "ERROR: syncing $s3_path on $instance_id"
+    cat /var/log/s3-log-sync.log | mail -s "Error syncing $s3_path on $instance_id" $NOTIFY_EMAIL
```
feanil: In the VPC we'll need to open up port 25 for this to work correctly right?  Is it possible to send the mail using SES instead?

playbooks/roles/common/templates/send-logs-to-s3.j2
```diff
@@ -0,0 +1,106 @@
...
+  else
+    echo "Error syncing $s3_path on $instance_id"
+  fi
+}
+
+trap onerror ERR SIGHUP SIGINT SIGTERM 
+
+# first security group is used as the directory name in the bucket
+sec_grp=$(curl -s http://169.254.169.254/latest/meta-data/security-groups | head -1)
```
e0d: Do we know this is deterministic?

playbooks/roles/common/templates/send-logs-to-s3.j2
```diff
@@ -0,0 +1,106 @@
...
+instance_id=$(curl -s http://169.254.169.254/latest/meta-data/instance-id)
```
e0d: > ec2metadata --instance-id

is simpler.

playbooks/roles/common/templates/send-logs-to-s3.j2
```diff
@@ -0,0 +1,106 @@
...
+ip=$(curl -s http://169.254.169.254/latest/meta-data/local-ipv4)
```
e0d: ec2metadata --local-ipv4

playbooks/roles/common/templates/send-logs-to-s3.j2
```diff
@@ -0,0 +1,106 @@
...
+# s3cmd needs $HOME defined even though it
+# doesn't use a config file
+export HOME=/
+
+exec > >(tee /var/log/s3-log-sync.log) 
+exec 2>&1
+
+shopt -s extglob
+
+NOTIFY_EMAIL={{ COMMON_S3_LOGS_NOTIFY_MAIL }}
```
e0d: I'd love to not have to run a mail server on each of our hosts.  If we installed the aws cli, either of these would be options and could be controlled via IAM.  Is AWS specific, however, but so is s3.

```
aws ses send-email 
aws sns publish 
```

e0d:
A general comment, repeated from hipchat, I think it would be a useful feature to not sync un-rotated logs.  We end up duplicated lines  by doing so.

playbooks/roles/common/templates/send-logs-to-s3.j2
jarv: A mail server isn't necessary, we just need a mail client (in this case the mail program). It does assume though that outgoing smtp is not blocked.
It will make sense to go this route if we switch to the aws cli

playbooks/roles/common/templates/send-logs-to-s3.j2
jarv: yeah, outgoing smtp (see comment above)
We can switch to the aws cli though I won't have time to do that until I get back on Tuesday unless someone else wants to make the change and test it.

playbooks/roles/common/templates/send-logs-to-s3.j2
jarv: our current vpc instances have only one security group. for ones that more than one it isn't ideal. I think it makes sense to switch to tags eventually.

playbooks/roles/common/tasks/main.yml
jarv: the common role is using a combination of both right now actually.
what do we do for template files that go under the /edx directory? it seems like that doesn't work when the path is configurable.

feanil: I think it's fine under /etc but what ed was talking about was the src should follow the same pattern 'templates/etc/logrotate.d/tracking.log.j2'

jarv: yeah i just it at the same level as the other logrotate configs, I will clean it up.

playbooks/roles/common/templates/send-logs-to-s3.j2
sefk: personally I'm fine opening outgoing smtp and not worrying about integration with ses.  Doesn't seem worth the dependency and complication just to send a little email like this.

edx-secure: I think I lean more towards ses (didn't consider the vpc use case) Maybe an
option to toggle it?

jarv:
unfortunately for now we will need to sync the unrotated logs since `logrotate --force` won't always force a rotation.

jarv:
@feanil @e0d this is ready for a final look-over

@sefk @carsongee we decided to move the s3 logrotate from common to a new aws role which means this is only going to be for aws deployments.  Using ses to send mail so that we do not have to adjust egress rules in the vpc

feanil:
looks good.

PR #788, opened 2014-02-14
install awscli from userdata.

e0d:


feanil:
looks good.  This will update the base autoscaling groups, when Asgard copies them, will the new ASGs also have this as the default userdata for instances?

PR #792, opened 2014-02-18
default to old change screen

mulby:
The new one lacks many very useful features and is still in beta, we should
be using the old one pretty much exclusively.

Note that users can override this setting in their own preferences if they
really want to use the new change screen.

Reviewers: @gwprice, @e0d


gwprice:
:+1:

PR #793, opened 2014-02-18
Put some tops on the letters.

nedbat:
@jarv better?


jarv:
:+1:

nedbat:
As of johnnycake, this seems to have reverted. :(

nedbat:
Oh, no it hasn't, there's just another file to also get the tops added to.

PR #794, opened 2014-02-19
Change boxname on fullstack, and remove ansible play from devstack

carsongee:
I wasn't sure on the devstack "fix", but don't know why you would need to run ansible again on a prebuilt box. The fullstack box isn't downloading a new box if you already have the previous release installed, so I changed the name.

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -16,7 +16,7 @@ cd /edx/app/edx_ansible/edx_ansible/playbooks
 # this can cause problems (e.g. looking for templates that no longer exist).
 /edx/bin/update configuration master
 
-ansible-playbook -i localhost, -c local vagrant-devstack.yml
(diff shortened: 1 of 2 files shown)


jarv:
I think the idea at the time was that you would re-run ansible so that you would be caught up with any configuration updates (specifically json updates) on the master branch.
@wedaly  may have more input.

wedaly:
We need to re-run the deploy scripts during provisioning to checkout repos into the shared folders, which don't exist on the VM.  There are some other set-up steps that need to occur as well (such as creating binstubs for the rake command in edx-platform).

We might be able to get away with running just the steps tagged --deploy instead of every step to speed things up.

carsongee:
I'll test it out with --tags deploy and if it works I'll update it or remove my change.

carsongee:
Switching to deploy tag worked for me, and was quite a bit quicker

carsongee:
ping this came up on the code list and reminded me about it.

jarv:
:+1:

PR #798, opened 2014-02-19
Jarv/nginx redirect

jarv:


playbooks/roles/aws/tasks/main.yml
```diff
@@ -49,10 +49,14 @@
 # Remove them if they are regular files
 # This can be removed once we don't need to worry
 # about backwards compatibility.
-- stat: path={{ COMMON_BIN_DIR }}/{{ aws_s3_sync_script|basename }}
+- local_action:
```
feanil: Why are these local actions, shouldn't they happen on the server doing the syncing?

jarv: they come right before the copy, which is copying a file from the local system to the remote.

feanil: That seems true for the SSL cert related ones, but doesn't make sense here for the s3_sync_script.  The {{COMMON_BIN_DIR}} path isn't going to exist on the local system.

jarv: you are right, updated.

feanil:
:+1:

hkim823:
@jarv still want this to be merged?

jarv:
These changes were made in another branch, closing

PR #801, opened 2014-02-21
Download JSCover from S3

wedaly:
Use an S3 URL for JSCover instead of downloading from a SourceForge mirror.  This should be more reliable, as SourceForge download URLs can change.

@feanil


jarv:
can we use files.edx.org instead?

wedaly:
@jarv Sure.  I don't think I have permissions to upload to that bucket, though.

feanil:
@wedaly uploaded it, can you use this url instead: 'http://files.edx.org/testeng/JSCover-1.0.2.zip'

Thanks.

wedaly:
sure thing

feanil:
looks good.

PR #802, opened 2014-02-21
adding aws role to the edx-east plays

jarv:
sending logs to s3 won't be enabled until we update
the var file for the vpc


feanil:
looks good.

PR #804, opened 2014-02-21
Refactoring

e0d:
- Adds new utility class for DNS entries
- Delays adding records to the end as one batch
- Alters the noop flag to output XML
- Fails if the run would require a delete
- Adds a force flag to make delete explicit


feanil:
looks good.

PR #806, opened 2014-02-22
use local temp dir for sockets and ansible cache

sefk:
@jbau or @jrbl please.

I was having problems where concurrent installs could trample on each
other.  The instance that immediately affected me was output caching
from ec2.py: the output of that command is different between staging and
prod, and both were being written to /tmp/ansible_ec2.cache and .index.

Fix here is to write to a local temp directory.  This creates empty temp
dirs to ensure that they are created in all repos.

While less likely, you could have collisions on named ssh sockets.
Those are named with just the instance name, which could be re-used
across VPC's.  Putting those in the ./tmp dir too prevents that.

Note that for consistency I did away with just the plain ec2.ini file,
and instead now there are prod- and stage- variants.  This is clean but
now means that you'll need to change your install command to look
something like this:

```
ANSIBLE_EC2_INI=prod-ec2.ini ANSIBLE_CONFIG=prod-ansible.cfg ansible-playbook -c ssh -u ubuntu -i ./ec2.py prod-app.yml
```

### DIFF
--- a/playbooks/edx-west/prod-ansible.cfg
+++ b/playbooks/edx-west/prod-ansible.cfg
@@ -1,1 +1,22 @@
-../ansible.cfg
+# config file for ansible -- http://ansible.github.com
+# nearly all parameters can be overridden in ansible-playbook or with command line flags
+# ansible will read ~/.ansible.cfg or /etc/ansible/ansible.cfg, whichever it finds first
+
+[defaults]
+
+jinja2_extensions=jinja2.ext.do
+hash_behaviour=merge
+host_key_checking = False
+
+# These are environment-specific defaults
+forks=10
+transport=ssh
+hostfile=./ec2.py
+extra_vars='key=deployment region=us-west-1'
+user=ubuntu
+
+
+[ssh_connection]
+# example from https://github.com/ansible/ansible/blob/devel/examples/ansible.cfg
+ssh_args= -o ControlMaster=auto -o ControlPersist=60s -o ControlPath=tmp/prod/ansible-ssh-%h-%p-%r
+scp_if_ssh=True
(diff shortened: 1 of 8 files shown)


sefk:
To reassure you that the .gitignores looks reasonable, here's what these dirs look like after some installing:

```
sefk@deploy {sef/edx-west/local_playbook_tempdir *} tmp> find .
.
./stage
./stage/ansible-ssh-ip-10-0-31-99.us-west-1.compute.internal-22-ubuntu
./stage/ansible-ec2.index
./stage/.gitignore
./stage/ansible-ec2.cache
./README
./.gitignore
./prod
./prod/ansible-ssh-ip-10-0-10-21.us-west-1.compute.internal-22-ubuntu
./prod/ansible-ec2.index
./prod/.gitignore
./prod/ansible-ec2.cache
./prod/ansible-ssh-ip-10-0-11-101.us-west-1.compute.internal-22-ubuntu
sefk@deploy {sef/edx-west/local_playbook_tempdir *} tmp> git status
# On branch sef/edx-west/local_playbook_tempdir
# Changes not staged for commit:
#   (use "git add <file>..." to update what will be committed)
#   (use "git checkout -- <file>..." to discard changes in working directory)
#
#       modified:   ../prod-app.yml
#
no changes added to commit (use "git add" and/or "git commit -a")
```

That one changed file (prod-app.yml) is just the normal changing install targets thing, nothing to worry about there.

jbau:
lgtm

sefk:
OK thanks for the review @jbau.  I'll go ahead and pull, especially since this is what I released with today. :)

sefk:
@e0d -- take note of this little change.  Not sure if it's something worth you changing for everyone.  This only seems applicable when you get into VPC land, which I believe you'll be in soon enough.

PR #814, opened 2014-02-25
E0d/newrelic role

e0d:
Basic role for installing the Newrelic system agent.


playbooks/roles/newrelic/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
...
+# vars are namespace with the module name.
+#
+newrelic_role_name: newrelic
+
+NEWRELIC_REPO: 'deb http://apt.newrelic.com/debian/ newrelic non-free'
+NEWRELIC_KEY_ID: '548C16BF'
+NEWRELIC_KEY_URL: 'https://download.newrelic.com/{{ NEWRELIC_KEY_ID }}.gpg'
+NEWRELIC_LICENSE_KEY: 'NEW-RELIC-KEY'
+
+NEWRELIC_APP_NAME: 'e0d-sandbox'
```
jarv: change the default?

playbooks/roles/newrelic/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+#
+#
+# Handlers for role newrelic
+# 
+# Overview:
+# 
+#
+- name: notify me
+  debug: msg="stub handler"
```
jarv: i would just remove the handlers dir if there aren't any, up to you though

playbooks/roles/newrelic/meta/main.yml
```diff
@@ -0,0 +1,20 @@
...
+# Role includes for role newrelic
+# 
+# Example:
+#
+# dependencies:
+#   - {
+#   role: my_role 
+#   my_role_var0: "foo"
+#   my_role_var1: "bar"
+# }
```
jarv: same comment for meta

jarv:
:+1:

playbooks/edx-east/forum.yml
```diff
@@ -5,6 +5,7 @@
   vars:
     enable_datadog: True
     enable_splunkforwarder: True
+    enable_newrelic: True
```
feanil, line 8: This should be disabled on the single instance sanboxes by default.  https://github.com/edx/configuration/blob/master/util/jenkins/ansible-provision.sh#L95-L96

feanil:
How will upgrades of the newrelic agent work?

e0d:
Per discussion, upgrades need to be handled in a more consistent way in general.  This role should be included in the edX base AMI play as it will be generally used.

feanil:
looks good.

PR #820, opened 2014-02-26
Jarv/user refactor

jarv:


e0d:
This generally looks good to me, though I prefer keeping the templates for things like the analytics or edxapp specific sudoers requirements with the owning role.   In this case, the config is closely related to the owning role, not the "utility" role.  Reuse of the templates would mostly be a bad idea.  I'm still :+1:

jarv:
Yeah I was feeling the same way up until I had to use `{{ item.name }}` in the template since we are looping over user_config to generate them.  For that reason I thought it would be better to pull it into the role, I don't have a strong preference though.

playbooks/edx-west/prod-jumpbox.yml
```diff
@@ -8,18 +8,35 @@
...
+          admin: true
+        - name: jinpa
+          github: true
+        - name: gbruhns
+          github: true
+        - name: paepcke
+          github: true
+        - name: akshayak
+          github: true
       tags: users
```
sefk, line 42: Thanks for the cleanup to our config while you're at it.

sefk:
@jarv did you consider making it so the default behavior was github?  So instead of making the plays conditional on `when item.github is defined` you could do `when item.authorized_keys is not defined`.  I like the github mechanism and I predict that'd be the default for most people.

Also: being able to support explicit (and >1) authorized keys is a nice feature.  We have a user at Stanford, because of his particular setup, wants multiple keys and moving to gh_users kind of left him in the lurch.  That's nice to support again!

jarv:
@sefk There is probably a use case for creating a user that doesn't have authorized keys but is also a github user (for example a non-login user)
I thought about github being the default and no-github being an option but it seems like now that we have a more generic role github might be the special case.

playbooks/roles/common/meta/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 dependencies:
-  - gh_users
+  - user
+  - automated
```
feanil: This role no longer exists.

playbooks/roles/edxapp/meta/main.yml
```diff
@@ -6,9 +6,10 @@ dependencies:
...
-    automated_sudoers_template: 'roles/edxapp/templates/etc/sudoers.d/99-automator-edxapp-server.j2'
-    automated_authorized_keys: "{{ EDXAPP_AUTOMATOR_AUTHORIZED_KEYS }}"
-    when: EDXAPP_INCLUDE_AUTOMATOR_ROLE
+  - role: user
+    user_info:
+      name: automator
+      type: restricted
+      sudoers_template: '99-edxapp-manage-cmds.j2'
+      user_authorized_keys: "{{ EDXAPP_AUTOMATOR_AUTHORIZED_KEYS }}"
+    when: EDXAPP_AUTOMATOR_AUTHORIZED_KEYS|length != 0
```
feanil, line 15: The automator user is going to need and rbash link for sudo so it can run the django admin commands.

feanil:
looks good.

sefk:
Thanks @jarv I understand.

PR #822, opened 2014-02-27
Added HTTP Strict Transport Security headers to ensure clients always

davestgermain:
use SSL (helps prevent MITM attacks).

### DIFF
--- a/playbooks/roles/nginx/templates/cms.j2
+++ b/playbooks/roles/nginx/templates/cms.j2
@@ -20,6 +20,8 @@ server {
 
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
+  # request the browser to use SSL for all connections
+  add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
 
   {% else %}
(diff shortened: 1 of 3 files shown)


davestgermain:
@feanil @e0d I'm not sure if these are the only places to set the header (or if there's an exception to the rule somewhere).

feanil:
@davestgermain can you also add this to the cms template?  Also, have you tested this setting on a single instance sandbox?

davestgermain:
@feanil done. How do I test it on a sandbox?

feanil:
You can bring up a single instance sandbox at 'http://jenkins.edx.org:8080/job/ansible-provision/' which will bring up an ec2 instance and then you can run '/edx/bin/update configuration <hash>' to update the configuration repo to be your branch and verify that it doesn't cause any problems.

davestgermain:
@feanil I don't have access to that jenkins instance, as far as I can tell

davestgermain:
@feanil well, the header seems to have no effect one way or the other.  but the bug report recommended adding it... [HTTP Strict Transport Security](http://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security)

feanil:
Hmm. Ok.  If the header is being passed through to the client, I think that is fine.  Can you see the header being set?

davestgermain:
yup

feanil:
cool, merge it.

PR #829, opened 2014-02-28
Add a role that will build an admin jenkins.

feanil:


playbooks/roles/jenkins_admin/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+#
+#
+# Handlers for role jenkins_admin
+# 
+# Overview:
+# 
+#
+- name: notify me
+  debug: msg="stub handler"
```
jarv: my personal preference is to remove the stub directories if they aren't used

playbooks/roles/jenkins_admin/defaults/main.yml
```diff
@@ -0,0 +1,51 @@
...
+
+# JENKINS_ADMIN_S3_PROFILE: !!null
+# JENKINS_ADMIN_CONFIGURATION_REPO: !!null
+# JENKINS_ADMIN_CONFIGURATION_SECURE_REPO: !!null
+# 
+# # Local path to the key to use to checkout secure repos on jenkins and in abbey
+# JENKINS_ADMIN_GIT_KEY_LOCAL_PATH: !!null
+# 
+# # EC2 Key to use when bringing up the abbey instance in ec2 (aws key-pair)
+# JENKINS_ADMIN_EC2_KEY: !!null
```
jarv: nitpick: put the all caps vars at the top

playbooks/roles/jenkins_admin/tasks/main.yml
```diff
@@ -0,0 +1,122 @@
...
+  when: JENKINS_ADMIN_CONFIGURATION_REPO is not defined
+
+- fail: "JENKINS_ADMIN_CONFIGURATION_SECURE_REPO is not defined."
+  when: JENKINS_ADMIN_CONFIGURATION_SECURE_REPO is not defined
+
+- fail: "JENKINS_ADMIN_GIT_KEY_LOCAL_PATH is not defined."
+  when: JENKINS_ADMIN_GIT_KEY_LOCAL_PATH is not defined
+
+- fail: "JENKINS_ADMIN_EC2_KEY is not defined."
+  when: JENKINS_ADMIN_EC2_KEY is not defined
```
jarv, line 46: for this one and maybe some of the others you have the default set to null which means that this will never fail (since the var will be defined)

feanil, line 46: That's why I commented them all out, I just left them in defaults so people could see what to set.

jarv, line 46: ah ok i missed that they were commented out

jarv:
:+1:

PR #831, opened 2014-02-28
Enable X11 forwarding in devstack release versions

jzoldak:
This is in base/devstack/Vagrantfile but not in release/devstack/Vagrantfile.


feanil:
Looks good.

PR #836, opened 2014-03-03
Updates to apache role to make it more flexible and inline with nginx

carsongee:
This removes the include file so that apache sites are more like nginx sites, allows more than one port in case you want to do ssl for example, and untags the tasks.

@jbau and @sefk You will want to double check that I didn't mess anything up for your installs.

PR #837, opened 2014-03-03
Update to shibboleth role and allow changing the template directory

carsongee:
I found when trying to implement this role for us, our attributes and shibboleth2 configs were to different to reconcile, so I added a flag to allow overriding the template directory to provide further customization. This is similar to #339 and #836 

@jbau and @sefk to double check there aren't any side effects introduced for you.


carsongee:
I've also changed the metadata backup task a bit to make it configurable and optional since that step is currently failing for me and is hard coded to Stanford's IDP

sefk:
Certainly looks like a reasonable thing to do, and _shouldn't_ cause a problem.  But @jbau needs to take a look since he knows the shib stuff 100x better than I do.  

Curious why you removed the tags.  They might not be that useful, but maybe a bit.

carsongee:
I don't mind adding them back in (at least the install ones), but I've generally switched to doing role tags in my plays so I don't have to have them for every task or using the https://github.com/edx/configuration/blob/94bde0ec44aae9b860479ff2766cfc4dc624c338/playbooks/run_role.yml play

playbooks/roles/shibboleth/tasks/main.yml
```diff
@@ -32,24 +29,15 @@
     - key
     - pem
   notify: restart shibd
-  tags:
-    - shib
-    - install
 
 - name: writes out configuration files
-  template: src={{item}}.j2 dest=/etc/shibboleth/{{item}}  group=_shibd owner=_shibd mode=0644
+  template: src={{ shib_template_dir }}/{{item}}.j2 dest=/etc/shibboleth/{{item}}  group=_shibd owner=_shibd mode=0644
```
jbau, line 34: is this supposed to be `shib_template_path`?

jbau:
I'm fine with removing the tags.  Role tags make more sense.

playbooks/roles/shibboleth/tasks/main.yml
carsongee, line 34: whoops, good catch. Should be dir in defaults for parity

jbau:
ok. this looks good to me. :+1:

PR #839, opened 2014-03-04
Remove vagrant version requirement

valera-rozuvan:
**Description**

Many people have complained that the 1.4 version of Vagrant is not necessary. We can use a lower version of Vagrant. However, the `Vagrant.require_version` is only available starting with Vagrant 1.4. So, for now this requirement should be removed.

**Reviewers**
- @hkim823


hkim823:
:thumbsup:

PR #840, opened 2014-03-04
Add devpi_host setting to devpi role

tusbar:
Allow specifying a different host than localhost.


feanil:
looks good.

@jarv or @e0d

tusbar:
Btw, I haven’t changed https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L175 because of #589.

hkim823:
Closing this since we're not really using Devpi anymore

PR #849, opened 2014-03-05
edx-west: use local temp dir for sockets and ansible cache

sefk:
@feanil -- this is a change that we've been running with for the last few weeks on edx-west/release.  It was pulled into our config branch in #806.  This just gets the same change onto master, anticipating that we'll be picking up master at some point soon.

Given that we're doing this already and only affects our plays should be a quick one.  Thanks for reviewing.  Original commit msg below.

---

I was having problems where concurrent installs could trample on each
other.  The instance that immediately affected me was output caching
from ec2.py: the output of that command is different between staging and
prod, and both were being written to /tmp/ansible_ec2.cache and .index.

Fix here is to write to a local temp directory.  This creates empty temp
dirs to ensure that they are created in all repos.

While less likely, you could have collisions on named ssh sockets.
Those are named with just the instance name, which could be re-used
across VPC's.  Putting those in the ./tmp dir too prevents that.

Note that for consistency I did away with just the plain ec2.ini file,
and instead now there are prod- and stage- variants.  This is clean but
now means that you'll need to change your install command to look
something like this:

```
ANSIBLE_EC2_INI=prod-ec2.ini ANSIBLE_CONFIG=prod-ansible.cfg ansible-playbook -c ssh -u ubuntu -i ./ec2.py prod-app.yml
```

Conflicts:

```
playbooks/edx-west/ansible.cfg
```

### DIFF
--- a/playbooks/edx-west/ec2_cache/data/.gitignore
+++ b/playbooks/edx-west/ec2_cache/data/.gitignore
@@ -1,0 +1,3 @@
+*
+
+!.gitignore
--- a/playbooks/edx-west/ansible.cfg
+++ b/playbooks/edx-west/ansible.cfg
@@ -1,1 +1,0 @@
-../ansible.cfg
--- a/playbooks/edx-west/ec2_cache/.gitignore
+++ b/playbooks/edx-west/ec2_cache/.gitignore
@@ -1,0 +1,6 @@
+*
+
+!prod
+!stage
+!data
+!.gitignore
--- a/playbooks/edx-west/ec2_cache/README
+++ b/playbooks/edx-west/ec2_cache/README
@@ -1,0 +1,3 @@
+This temp directory created here so that we can make sure it doesn't
+collide with other users doing ansible operations on the same machine;
+or concurrent installs to different environments, say to prod and stage.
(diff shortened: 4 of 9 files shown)


feanil:
@sefk a couple of things, I don't like the temp directories as part of the repo.  Why not just set the cache path to /tmp/ansibel-stage-cache or /tmp/ansible-prod-cache?

sefk:
Having temp dirs right there gets them created for you.  If the dir doesn't exist, I believe ansible fails, and I don't know of a way to ensure they are in place for you net of some weird thing outside the play (makefile, git magic).

hkim823:
@sefk and @feanil i've been asked to help cleanup older PRs, what's the status on this one?

feanil:
@sefk I don't think there are any ways to have them automatically created.  I would suggest maybe calling them something more informative like ec2_cache instead of 'tmp' but that's a nitpick. :+1:

sefk:
@feanil changing the dir name from tmp to ec2_cache is a good one.  I made that change and repushed my branch.  Can you give one more review pls?

playbooks/edx-west/ec2_cache/data/.gitignore
```diff
@@ -0,0 +1,3 @@
+*
+
+!.gitignore
```
sefk, line 3: This is a new VPC for us where we're doing the analytics work.

feanil:
looks good to me. :+1:

sefk:
woot

PR #852, opened 2014-03-05
adding bastion role and task to setup access to the dbs from it

jarv:
should be merged with https://github.com/edx-ops/mckinsey-secure/pull/37


feanil:
looks good.

carsongee:
After applying this, I'm getting

```
TASK: [user | create the users] ***********************************************
fatal:  => One or more undefined variables: 'str object' has no attribute 'name'
```

debug info was:

```
TASK: [user | debug var=user_info] ********************************************
ok: => {
    "item": "",
    "user_info": "{{COMMON_USER_INFO}}"
}
```

PR #855, opened 2014-03-06
Added CAS settings to edxapp

carsongee:
Change related to https://github.com/edx/edx-platform/pull/2702/ for adding CAS server configuration to ansible.


feanil:
@carsongee just built this with the defaults and got the following in my env.json

```
    "CAS_EXTRA_LOGIN_PARAMS": "$EDXAPP_CAS_EXTRA_LOGIN_PARAMS",
    "CAS_SERVER_URL": "$EDXAPP_CAS_SERVER_URL",
```

carsongee:
Crazy, no matter what I stick in for the value it never ends up null in the json hash.  I guess I can just give them safe default values unless someone know some yaml/jinja magic

feanil:
Yea, I think because the dollar sign reference is a ansible 'feature' that's gonna go away in the future, for other similar cases we set the default to empty string, and we have logic in aws.py  that defaults it to none if the value is empty string.

eg. https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L290-L292

carsongee:
Cool, I've updated it

feanil:
Cool, so you'll need to update the platform code as well so that this works properly.  Do you already have a PR for the platform update?

carsongee:
It was merged last night.

feanil:
:+1:

PR #858, opened 2014-03-07
Feanil/always update cache prefixes

feanil:


e0d:
Mentioning what we discussed here, this is :+1: for expediency, but we should have a cache key that is traceable back to the associated builds, otherwise we lose the ability to trace which versions are in the cache.

PR #860, opened 2014-03-07
Have elastic search start, but not restart each time

carsongee:
With a large index on a poorly resourced instance, restarting the service causes me to lose logs and have tasks further down fail (logstash stuff particularly that needs elastic search to be fully running to install a template).  If this needs to be this way I can always work around it, but thought started should be sufficient.

### DIFF
--- a/playbooks/roles/elasticsearch/tasks/main.yml
+++ b/playbooks/roles/elasticsearch/tasks/main.yml
@@ -84,3 +84,7 @@
   
 - name: Ensure elasticsearch is enabled and started
   service: name=elasticsearch state=started enabled=yes
+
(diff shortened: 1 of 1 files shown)


feanil:
@carsongee this doesn't handle upgrades.  If the version of elasticsearch gets upgraded, this change would prevent the new version from getting started. Perhaps use the existing `elasticsearch_reinstall.changed` artifact to only restart if we installed a new version.

carsongee:
Cool, added it with a when clause and kept the started task as well in case the service is dead or disabled and we didn't download a new version

feanil:
sweet. :+1:

PR #854, opened 2014-03-06
parameterize machine names in edx-west playbooks -- to edx-west/release

sefk:
@jbau and @jrbl -- this is the new way I recommend running ansible plays, using the "-e machine=foo" option.  this is a pull request to our release branch.  If it lands then I'll do the same to master just for cleanliness sake (I've already put our README up there).

PR #861, opened 2014-03-07
Parameterize machine names in edx-west playbooks

sefk:
This is pulling into master a few changes we've made in the playbooks we use to drive our own Stanford instance.  Mostly want to keep master in sync so others can see what we're doing and our upcoming merges are cleaner.

@feanil -- can you take a look?  Just asking you since you've been looking at other of my changes, but this one isn't very exciting. 

Look at PR #854 for discussion with Joe and Jason about the over all feature.  Thanks.

### DIFF
--- a/playbooks/edx-west/carnegie-prod-app.yml
+++ b/playbooks/edx-west/carnegie-prod-app.yml
@@ -1,10 +1,14 @@
+<<<<<<< HEAD
 - hosts: ~tag_Name_app(10|20)_carn
+=======
+# This uses variable expansion so you can select machine(s) from the command line
+# using the -e flag.  See README for instructions on how to use.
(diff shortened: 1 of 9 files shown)


playbooks/edx-west/carnegie-prod-app.yml
```diff
@@ -1,10 +1,14 @@
+<<<<<<< HEAD
```
feanil: Looks like a merge conflict that wasn't resolved.

sefk:
whoops @feanil sorry i missed that.  Fixed now.

feanil:
looks good.

sefk:
Appreciate the extra set of eyes @feanil -- pulling.

PR #863, opened 2014-03-10
Addition of production tcp socket configuration

carsongee:
This adds an option to listen on a TCP socket and control host and port with ansible.  See https://github.com/edx/cs_comments_service/pull/90 for related pull to that repo.


e0d:
:+1:

PR #864, opened 2014-03-10
Added option to specify migration user and changed to admin user and password vars

carsongee:
I'm honestly not 100% sure what is happening in this part of the code, but it seems like we should be able to set the DB user and that we should use the admin password and user vars for migrations instead of the app user.  Feel free to correct me though, but I for example don't have a 'root' user in MySQL.


jarv:
I can help explain what is going on here.

Initially when this play was created the `EDXAPP_MYSQL_USER` had permissions to do migrations. In production we don't like giving the user that much access so what we did is create some new user definitions:
- _read_only_: SELECT on all databases, all RDS instances in the VPC
- _admin_: CREATE USER on all RDS instances in the VPC
- {_edxapp_, _xqueue_, _ora_} application users: SELECT, INSERT, UPDATE, DELETE access on their respective databases.  For each db we also create corresponding _<name>_test_ dbs and give them full access to it, this is so we can run "fake" migrations for building AMIs

When we want to do a real migration we use the root credentials, we have discussed creating a separate user just for that as well.  The admin user is misleading since he can only add users right now and nothing else.

We have a play that creates all of these users for the different dbs, see https://github.com/edx/configuration/blob/master/playbooks/edx-east/create_dbs.yml

jarv:
@e0d @feanil I think this change is ok. It doesn't apply to us since we are not currently using ansible to run migrations. It assumes the mysql admin user has permission to do migrations which we may allow later. Agree? The other option is to introduce a new class of user for migrations.

carsongee:
Have you guys had a chance to decide on this?

feanil:
@jarv @carsongee Just yesterday we talked about introducing a seperate user for migrations that is not the ADMIN_USER.  So we would have 3 users total

app_user:  can only add or remove data
admin_user: can only create or delete mysql users
migration_user: can run migrations on schema

Since all of our migrations are outside of ansible right now we can reconcile differences if we move them in here.

:+1:

carsongee:
It sounds like it would make more sense to change these to MIGRATE_USER and MIGRATE_PASSWORD. I'm not committed to these as is and should make it easier for you down the road if you decide to use these tasks.  Sound good?

feanil:
Sounds good to me.

carsongee:
Updated. I left the variable to define the ADMIN user/pass in there even though that var isn't used anywhere

PR #870, opened 2014-03-12
Adding rds control libraries

carsongee:
This kind of completes the missing part for us to be able to use pure ansible to configure an entire VPC.  I'm hoping to get some time adapt that old PR so that it will work here, but requesting to merge these in now while I'm thinking of it.


jarv:
:+1:

PR #874, opened 2014-03-13
adding user and edx_ansible

jarv:


feanil:
@jarv how about naming the new file vpc_admin.yml since it has both jenkins and asgard?

jarv:
done

PR #876, opened 2014-03-14
adding gem for status.edx.org

jarv:


feanil:
There is nothing in jenkins_master or jenkins_admin that install ruby on these servers.  Is ruby on these boxes by default?  It would be good to install it explicitly to make the role independent of our base AMI even if that's the case.

jarv:
done

PR #888, opened 2014-03-18
Feanil/disable supervisor runs

feanil:
@sefk @jbau @carsongee @jarv @e0d 

This a big looking change which is mostly a refactoring.
- The 'devstack' flag is now only used to start up some apps in developer mode.
- The new 'disable_edx_services' flag will allow you to install everything but keep supervisor from starting any of the services by default.
- All the supervisor configs will go into a confs.available.d folder and be symlinked into the conf.d folder.

The defaults have all remained the same but this allows us to build AMIs without starting up all the services at AMI build time.


playbooks/roles/certs/handlers/main.yml
```diff
@@ -21,3 +21,4 @@
     config={{ supervisor_cfg }}
     state=restarted
   when: certs_installed is defined
+  when: not disable_edx_services
```
jarv: I don't think this is valid syntax (the keys will clash when the yaml is read).

playbooks/roles/common/defaults/main.yml
```diff
@@ -73,6 +73,10 @@ common_git_ppa: "ppa:git-core/ppa"
 
 # Skip supervisor tasks
 # Useful when supervisor is not installed (local dev)
```
jarv: Can you add to this comment to explain the new reason for the flag?

jarv:
Can you test this branch using the http://jenkins.edx.org:8080/job/ansible-provision-alpha job and also squash the commits since there are so many of them?  On the jenkins job I would like to try the from-scratch server option as well as the full installation option.

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -88,11 +90,31 @@
   pip: name=supervisor virtualenv="{{supervisor_venv_dir}}" state=present
   sudo_user: "{{ supervisor_user }}"
 
+- name: install supervisor in its venv
+  pip: name=boto virtualenv="{{supervisor_venv_dir}}" state=present
+  sudo_user: "{{ supervisor_user }}"
+  when: supervisor_service == "supervisor"
```
jarv: what's the reason for this when condition?

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -88,11 +90,31 @@
...
 - name: create supervisor upstart job
   template: >
     src=supervisor-upstart.conf.j2 dest=/etc/init/{{ supervisor_service }}.conf
     owner=root group=root
 
+- name: create pre_supervisor upstart job
+  template: >
+    src=pre_supervisor.conf.j2 dest=/etc/init/pre_supervisor.conf
+    owner=root group=root
+  when: supervisor_service == "supervisor" and disable_edx_services
```
jarv: Maybe a comment here since it could be confusing to someone who isn't using AWS?
Also will this conflict with the non-AWS devstack environment?

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -1,9 +1,17 @@
...
-    src={{ item }}.conf.j2 dest={{ supervisor_cfg_dir }}/{{ item }}.conf
+    src={{ item }}.conf.j2 dest={{ supervisor_available_dir }}/{{ item }}.conf
     owner={{ supervisor_user }} group={{ common_web_user }} mode=0644
   with_items: ['xqueue', 'xqueue_consumer']
 
+- name: "enabling supervisor scripts - xqueue, xqueue consumer"
+  file: >
+    src={{ supervisor_available_dir }}/{{ item }}.conf
+    dest={{ supervisor_cfg_dir }}/{{ item }}.conf
+    owner={{ supervisor_user }} group={{ common_web_user }} mode=0644 state=link
```
jarv: add force option to all symlinks

requirements.txt
```diff
@@ -8,7 +8,4 @@ ecdsa==0.10
 paramiko==1.12.0
 pycrypto==2.6.1
 wsgiref==0.1.2
-GitPython==0.3.2.RC1
-pymongo==2.4.1
-requests==2.2.1
```
jarv, line 13: what's the reason for this change?

playbooks/roles/certs/handlers/main.yml
carsongee: I've found this will work, but it is worth doing a |bool filter after the var to make sure it works when passed in from command line or other interpolated value

carsongee: oh, sorry, missed the double when, but comment is true for the |bool on all these

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -68,10 +68,12 @@
     state=directory
     owner={{ supervisor_user }}
     group={{ supervisor_service_user }}
+    mode="775"
```
carsongee: Is this for the pyc thing?  I'd rather we do the precompile thing than give the service user write access here if that is reason.

playbooks/roles/supervisor/tasks/main.yml
feanil: I only wanted to install boto for the main supervisor but I see that we should also add an `and disable_edx_services` to this because we don't need it in the general case.

requirements.txt
feanil, line 13: These libraries were installed so that we could get to mongo and for some of the stuff that the prepare-release was doing.  We're not using them anymore.

playbooks/roles/supervisor/templates/pre_supervisor.conf.j2
```diff
@@ -0,0 +1,7 @@
+description     "Tasks before supervisord"
```
e0d: Can we add the path under the template directory so it's clear at a glance where this stuff goes.

e0d:
:+1:

PR #890, opened 2014-03-18
If you have a branch name with .py in it, compileall will fail with bad syntax

carsongee:
This was one of the fun ones to track down:

```
SyntaxError: ('invalid syntax', ('/edx/app/edxapp/edx-platform/.git/refs/remotes/origin/oliviermarquez/update-tabs.py', 1, 40, '925133409b1d81997e04e9c6d9ff58f01e309f46\n'))
```

Turns out python will try and compile everything, even git branches, and it isn't so good at git branches.

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -253,7 +253,7 @@
   - edxapp-sandbox
 
 - name: compiling all py files in the edx-platform repo
-  shell: "{{ edxapp_venv_bin }}/python -m compileall {{ edxapp_code_dir }}"
+  shell: "{{ edxapp_venv_bin }}/python -m compileall -x .git.* {{ edxapp_code_dir }}"
   sudo_user: "{{ edxapp_user }}"
   notify:
   - "restart edxapp"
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -253,7 +253,7 @@
   - edxapp-sandbox
 
 - name: compiling all py files in the edx-platform repo
-  shell: "{{ edxapp_venv_bin }}/python -m compileall {{ edxapp_code_dir }}"
+  shell: "{{ edxapp_venv_bin }}/python -m compileall -x .git.* {{ edxapp_code_dir }}"
```
feanil: Should that be `.git/*`?

carsongee: It is a regex and not a glob, so it will catch .git/, but I can make go ahead and make it more specific if you want to be able to compile .git_blah.py

carsongee:
Changed, although I do wonder if we maybe want to be more selective and just compile the venv and edx-platform dirs

feanil:
That's not a bad idea, there are sometimes other repos in platform-code(themes,microsites) but I don't think they will have python in them.  Do you want to do that instead?

carsongee:
Just realized this is just the edx-platform directory already (misread the edxapp_code_dir var), so this is right. It looks like we don't do this for the venv and doesn't appear we need to I guess. I tried running it against the venv and apparently there is a syntax error in the zendesk library, sympy, diff-cover, lettuce, and xblock so we would have to do an "ignore_errors: yes" if we wanted to compile those.  It also takes a while to compile that whole env.

feanil:
:+1:

PR #891, opened 2014-03-19
Update main.yml for Localization

fmyzjs:
Localization


feanil:
EDXAPP_LANG is not the same as the LANGUAGE_CODE.  The LANG setting is used by the environment that does asset pre-processing and is a different format from the default platform LANGUAGE which should be of the formats provided here(http://www.i18nguy.com/unicode/language-identifiers.html) but in all lower case.  It's a good idea to support this so please add a new variable called 'EDXAPP_LANGUAGE_CODE' and default it to 'en'.

fmyzjs:
I got it,thx.  @feanil

fmyzjs:
https://github.com/edx/configuration/pull/894

PR #896, opened 2014-03-20
changes to match currently what's in production for hourly tracking.log

hkim823:


playbooks/roles/common/templates/edx_logrotate_tracking_log.j2
```diff
@@ -1,11 +1,14 @@
-{{ COMMON_LOG_DIR }}/tracking.log {
+{{ COMMON_LOG_DIR }}/hourly/tracking.log {
```
feanil: This shouldn't be hourly because it's supposed to be the location of the file being rotated.

playbooks/roles/common/templates/etc/logrotate.d/edx_logrotate_tracking_log.j2
```diff
@@ -1,11 +1,14 @@
-{{ COMMON_LOG_DIR }}/tracking.log {
+{{ COMMON_LOG_DIR }}/hourly/tracking.log {
```
feanil: Same comment as above.

playbooks/roles/common/templates/edx_logrotate_tracking_log.j2
```diff
@@ -3,9 +3,12 @@
   compress
   delaycompress
   dateext
+  dateformat -%Y%m%d-%s
```
feanil: This file looks like it's not actually being used.  We should delete it.  I think it was meant to be moved to templates/etc/... but got copied instead.

hkim823:
Tested working in my sandbox.

playbooks/roles/common/tasks/main.yml
```diff
@@ -51,14 +51,23 @@
...
   template: dest=/etc/logrotate.d/edx-services src=etc/logrotate.d/edx_logrotate.j2 owner=root group=root mode=644
 
   # This is in common to keep all logrotation config
   # in the same role
+- name: Create hourly subdirectory in logrotate.d
+  file: path=/etc/logrotate.d/hourly state=directory
+
 - name: Install logrotate configuration for tracking file
-  template: dest=/etc/logrotate.d/tracking.log src=etc/logrotate.d/edx_logrotate_tracking_log.j2 owner=root group=root mode=644
+  template: dest=/etc/logrotate.d/hourly/tracking.log src=etc/logrotate.d/hourly/edx_logrotate_tracking_log.j2 owner=root group=root mode=644
```
jarv, line 63: nitpick: add a line break here, somethign like

```
template: >
  dest=...
  src=...
  owner=...  group=...
```

jarv:
:+1:

PR #903, opened 2014-03-21
Nginx X-Forwarded-For header configuration.

mtyaka:
The nginx configuration was set to pass `X-Forwarded-For` and related headers
unmodified to backend services.
This works fine when nginx is deployed behind another load balancer
(for example ELB), which already properly sets the `X-Forwarded-*` for headers.

But in simpler deployment scenarios, where nginx is directly facing the end user,
nginx should discard any existing `X-Forwarded-For` header and set it to remote IP of the request,
the `X-Forwarded-Port` header to the nginx server port, and the `X-Forwarded-Proto`
to the scheme of the current request.

The behavior is controlled by the `nginx_set_x_forwarded_headers` configuration variable.
When set to False, nginx will forward existing X-Forwarded-\* unmodified.
When set to True, nginx will discard any existing `X-Forwarded-*` headers and set them itself.

`nginx_set_x_forwarded_headers` is set to `False` by default, but set to `True` in the `edx_sandbox` playbook, where nginx is usually not deployed behind another load balancer.

The variable name is probably not ideal and I would welcome any better suggestions.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,3 +31,4 @@ Wang Peifeng <pku9104038@hotmail.com>
 Ray Hooker <ray.hooker@gmail.com>
 David Pollack <david@sologourmand.com>
 Rodolphe Quiedeville <rodolphe@quiedeville.org>
+Matjaz Gregoric <mtyaka@gmail.com>
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -13,6 +13,8 @@
     openid_workaround: True
     EDXAPP_LMS_NGINX_PORT: '80'
     edx_platform_version: 'master'
+    # Set to false if deployed behind another proxy/load balancer.
+    NGINX_SET_X_FORWARDED_HEADERS: True
     # These should stay false for the public AMI
     COMMON_ENABLE_DATADOG: False
     COMMON_ENABLE_SPLUNKFORWARDER: False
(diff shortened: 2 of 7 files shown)


hkim823:
Hi there @mtyaka can you rebase this to current master?  There's merge conflicts with this PR.

mtyaka:
@hkim823 Thanks for looking into this. I rebased against master.

feanil:
@hkim823 looks good to me, can you do the second review.

hkim823:
I would like to make this set to false by default, only because I could see people using this play behind a load balancer.  Thoughts?

mtyaka:
@hkim823 I set it to true in `edx_sandbox` because I imagined that playbook was more often used in deployment scenarios without an extra load balancer, but I may be completely wrong.
I wouldn't mind setting it to false if you think that's a better default.

carsongee:
Isn't there a way to do this with nginx if statements?  e.g 

``` perl
if ( $http_x_forwarded_proto != '') { 
  proxy_set_header X-Forwarded-Protocol $http_x_forwarded_proto
} else {
  proxy_set_header X-Forwarded-Protocol $scheme
}
```

Not exactly sure what the performance implications would be of that, but would make it more adaptable.  Just a thought.

mtyaka:
@carsongee The problem with that approach is that it allows IP spoofing. When not deployed behind another load balancer, nginx should not simply pass the X-Forwarded-For header along, because then the IP can easily be spoofed by sending a fake value for the X-Forwarded-For header.

carsongee:
Ah, good point @mtyaka. :+1 from me.  No preference on the default, but I would like it to be an all caps var since I don't use a load balancer in front of nginx.

feanil:
@mtyaka I think this looks good as well. Can you update the variable name so it's in all caps to indicate that it is meant to be overridden?

mtyaka:
@feanil Thanks! I changed the variable name to all caps and added a comment describing its purpose.

feanil:
That looks great.  Can you also add yourself to the AUTHORS file, sorry I forgot to mention that earlier.

mtyaka:
@feanil Added myself to AUTHORS.

PR #908, opened 2014-03-24
Add discern_local_pypi_mirror variable for use a local pypi mirror easil...

pku9104038:
Add a variable discern_local_pypi_mirror as the extra_args for pip install python requirmente, and define the default value as pypi official site. This will help to make it easy if someone want to use a local pypi mirror to speed up the deployment.


feanil:
Discern already has a dependency on Common which defines 'COMMON_PYPI_LOCAL_MIRROR'(https://github.com/edx/configuration/blob/master/playbooks/roles/common/defaults/main.yml#L20)  Can you update this to use that?

PR #911, opened 2014-03-24
Be able to paramatrize the admin subnets.

feanil:


jarv:
:+1:

e0d:
We had talked about using something smaller than a class b, could we address that now?

feanil:
It would actually be 2 class C's right now(under 10.254.0.0/16), though it could be even smaller if you like. Having it be 2 class Cs per admin will mean we will run out of them around the same time as we run out of unique subnets for VPCs.

PR #920, opened 2014-03-25
Improperly formed json makes AWS sad

fredsmith:


jarv:
I'll add more linting to travis! :) :+1:

PR #922, opened 2014-03-25
changing from ubuntu package to pip

jarv:


feanil:
looks good once testing is done.

PR #932, opened 2014-03-26
Cleaning up dollar vars

hkim823:


playbooks/roles/aws/templates/send-logs-to-s3.j2
```diff
@@ -5,7 +5,7 @@
 #  This script can be called from logrotate
 #  to sync logs to s3
 
-if (( $EUID != 0 )); then
+if (( $EUID != 0 ));git  then
```
jarv: typo?

jarv:
I think I would remove these assignments altogether form the bash script if isntead it will be better to use the template vars:

```
S3_LOGFILE="{{ aws_s3_logfile }}"
NOTIFY_EMAIL={{ AWS_S3_LOGS_NOTIFY_EMAIL }}
FROM_EMAIL={{ AWS_S3_LOGS_FROM_EMAIL }}
AWS_CMD={{ aws_cmd }}
```

feanil:
looks good, I agree with john, we should remove the variable assignments since they are no longer used anywhere.

jarv:
given that there are still dollar variables in this template will it still throw warnings? @feanil I wasn't escaping these previously, are you surprised this ever worked?

hkim823:
In my testing it seems to ignore $ variables if they aren't defined, it's only the defined ones it complains about.  

This is similar to how the update script in the edx_ansible is templated it and doesn't complain about that.  I just ran it on my sandbox looking for it to throw errors but it didn't seem to when I manually ran the aws play.

hkim823:
I've updated a bunch, looking for another code review @feanil @jarv

feanil:
looks good.

PR #934, opened 2014-03-27
creates the migration user, defaults the maintenance user passwords to

jarv:
null


carsongee:
Don't we need to add something so devstack migrations will work with this change, or is that working some other way?

jarv:
@carsongee we should probably do a quick hangout to discuss these changes. Do you have time after lunch?

carsongee:
Cool, :+1:

PR #938, opened 2014-03-27
a few tags for task limits.

e0d:
Allows doing things like:

```
ansible-playbook -i ec2.py commoncluster.yml --tags=vhosts --extra-vars "deployment=hello env=world" -e@/path/to/secrets.yml -T 30
```


feanil:
looks good.

PR #940, opened 2014-03-28
Han/jenkins admin

hkim823:
Cleaning up var names with spaces, added {{ vars }} instead of $vars


playbooks/roles/jenkins_admin/templates/edx/var/jenkins/jobs/build-ami/config.xml.j2
```diff
@@ -27,7 +27,7 @@
           <defaultValue></defaultValue>
         </hudson.model.TextParameterDefinition>
         <hudson.model.TextParameterDefinition>
-          <name>vars</name>
+          <name>ls</name>
```
feanil: This should not change.

feanil:
looks good.

PR #941, opened 2014-03-28
Fix forum LISTEN_{HOST,PORT} env variable names

tusbar:
https://github.com/edx/cs_comments_service/blob/master/config/unicorn_tcp.rb#L6


feanil:
@e0d this looks good to me.

e0d:
:+1:  @tusbar if you are running the forums in production, I would recommend using the nginx/unicorn config that used a unix socket rather than TCP to either webbrick or unicorn.  Performance is substantially better.

PR #943, opened 2014-03-28
The jenkins_worker roles needs to include the role defaults.

feanil:
These roles now set various default passwords that the edxlocal role
needs.


feanil:
@jzoldak  did you not need this change to make your stuff work?  If that's the case, I'll close this PR.

jzoldak:
Oh yeah, I did need those changes too. I just hadn't narrowed down to what they were exactly. 
My other PR #950 was just to make sure that I got in the newest python requirements correctly.

Thx for this. :thumbsup:

PR #951, opened 2014-03-28
adding defaults for CC_PROCESSOR

jarv:


feanil:
Why not use `EDXAPP_AUTH_EXTRA` instead?  Doesn't make sense to have a default for this since it will never work.

https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/templates/lms.auth.json.j2#L1

jarv:
I forgot about that var, I guess we can do that. Like the comment says the default is just copied from common.py. We could also make the default blank?
I think I do want to keep as much in defaults/main.yml as possible to show what is possible to configure.

feanil:
:+1:

PR #955, opened 2014-03-31
Feanil/fix paramaterized jobs

feanil:


util/jenkins/build-ami.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash -x
+# This script is meant to be sourced from elswhere(jenkins) and expects the
+# following variables to be set:
+#   - refs - repo revisions to pass to abbey
+#   - vars - other vars to pass to abbey
+#   - deployment
+#   - environment
+#   - base_ami
+#   - BUILD_ID
```
e0d: What's the thinking around CAPS versus non caps vars here?

util/jenkins/build-ami.sh
```diff
@@ -0,0 +1,57 @@
...
+if [[ -n "$base_ami" ]]; then
+  base_params="-b $base_ami"
+fi
+
+blessed_params=""
+if [[ "$use_blessed" == "true" ]]; then
+  blessed_params="--blessed"
+fi
+
+cd configuration
```
e0d, line 86: Doesn't managing this here limit options and maybe make things worse than letting jenkins manage virtualenvs?

util/jenkins/build-ami.sh
```diff
@@ -0,0 +1,57 @@
...
+#   - configuration_secure - the version of the secure repo to use
+#   - jenkins_admin_ec2_key - location of the ec2 key to pass to abbey
+#   - jenkins_admin_configuration_secure_repo - the git repo to use for secure vars
+#   - use_blessed - whether or not to use blessed AMIs
+
+if [[ "$play" == "" ]]; then
+    echo "No Play Specified. Nothing to Do."
+    exit 0
+fi
+
```
e0d, line 61: I think this should fail if all required params are not available.  Hmm, is play even in that list.  Even better would be if it could support positional args or env vars so this wasn't a total pain to run from outside Jenkins.

util/jenkins/build-ami.sh
feanil: CAPS vars are jenkins builtins where as the rest are defined in the job.  I'll update the comment to say that.

util/jenkins/build-ami.sh
feanil, line 86: Jenkins creates the virtual env and sets your path so that pip/python come from that virtualenv but you still have to do the pip install yourself.

feanil:
@e0d comments addressed or replied to.

e0d:
:+1:

util/jenkins/build-ami.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash -x
+# This script is meant to be run from jenkins and expects the
+# following variables to be set:
+#   - BUILD_ID - set by jenkins, Unique ID of build
+#   - BUILD_NUMBER - set by jenkins, Build number
+#   - refs - repo revisions to pass to abbey
+#   - vars - other vars to pass to abbey
```
jarv: might want to mention here that these revisions and vars should be yaml syntax, one variable per line.

util/jenkins/build-ami.sh
```diff
@@ -0,0 +1,93 @@
...
+  configuration=`git rev-parse HEAD`
+  cd ..
+fi
+
+if [[ -z $configuration_secure ]]; then
+  cd configuration-secure
+  configuration_secure=`git rev-parse HEAD`
+  cd ..
+fi
+
```
jarv, line 75: nitpick: check to see if all of these are defined in one if statement, if not print usage().

util/jenkins/build-ami.sh
```diff
@@ -0,0 +1,93 @@
...
+  base_params="-b $base_ami"
+fi
+
+blessed_params=""
+if [[ "$use_blessed" == "true" ]]; then
+  blessed_params="--blessed"
+fi
+
+cd configuration
+pip install -r requirements.txt
```
jarv, line 87: It seems more natural to put the pip install in the jenkins job but I could go either way, it certainly doesn't hurt putting it here.

jarv:
:+1:

PR #968, opened 2014-04-02
Make LC_ALL default to 'en_US.UTF-8'.

mtyaka:
The `LC_ALL` environment variable is set to `en_US` in `/etc/default/locale`
in the precise32 vagrant box, see:

https://github.com/mitchellh/vagrant/issues/1188

When running the python unit tests in the devstack machine, several of the
tests fail with a `UnicodeEncodeError` when trying to create folders with names
that contain non-ascii characters.

Explicitly setting `LC_ALL`to `en_US.UTF-8` before running the tests fixes this issue,
but it would be nice if the box was set up with the `en_US.UTF-8` locale by default.

### DIFF
--- a/vagrant/base/devstack/Vagrantfile
+++ b/vagrant/base/devstack/Vagrantfile
@@ -45,6 +45,10 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
   end
 
+  # Make LC_ALL default to en_US.UTF-8 instead of en_US.
+  # See: https://github.com/mitchellh/vagrant/issues/1188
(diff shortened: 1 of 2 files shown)


feanil:
@mtyaka @jarv Any reason not to do this on the release Vagrant files as well?

mtyaka:
@feanil @jarv I don't think it is necessary to repeat this in the release Vagrant files, at least once a new base box has been packaged with this patch, and the release Vagrant files have been updated to be based on the new base box.
I am not really familiar with the packaging process so I may be wrong.

I'd be happy to add this to the release files as well, if you think it is needed.

feanil:
@mtyaka that makes sense to me.  We'll leave it in the base box for now.

PR #969, opened 2014-04-02
Allow users to be removed from systems

mulby:
This will allow us to keep a list of users that are allowed explicitly not
allowed to access systems.

```
user_info:
 - name: someonewholeft
   state: absent
 - name: activeaccount
   type: admin
   github: true
```

The above settings would ensure that the user named "someonewholeft" is not
able to access the system and "activeaccount" is.

### DIFF
--- a/playbooks/roles/user/tasks/main.yml
+++ b/playbooks/roles/user/tasks/main.yml
@@ -66,13 +66,20 @@
 #
 # By default for restricted users we only allow sudo, if you
 # want to provide more binaries add them to user_rbash_links
-# which can be passed in as a paramter to the role.
+# which can be passed in as a parameter to the role.
 #
 
 - debug: var=user_info
 
 - name: create the edxadmin group
   group: name=edxadmin state=present
+
+# some AMIs (such as EMR master nodes) don't read the config files out of /etc/sudoers.d by default
+- name: ensure sudoers.d is read
+  lineinfile: >
+    dest=/etc/sudoers state=present
+    regexp='^#includedir /etc/sudoers.d' line='#includedir /etc/sudoers.d'
+    validate='visudo -cf %s'
 
 # give full sudo admin access to the edxadmin group
 - name: grant full sudo access to the edxadmin group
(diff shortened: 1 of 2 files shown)


mulby:
@e0d @jarv not sure if you want this, but I made these modifications on our fork of this role

feanil:
looks good to me.

playbooks/roles/user/tasks/main.yml
```diff
@@ -74,6 +74,12 @@
 - name: create the edxadmin group
   group: name=edxadmin state=present
 
+- name: ensure sudoers.d is read
```
jarv: why is this necessary? afaik reading sudoers.d is the default.

playbooks/roles/user/tasks/main.yml
```diff
@@ -85,19 +91,21 @@
...
 - name: create .ssh directory
   file:
     path=/home/{{ item.name }}/.ssh state=directory mode=0750
     owner={{ item.name }}
+  when: item.get('state', 'present') == 'present'
   with_items: user_info
 
 - name: assign admin role to admin users
-  user:
+  user: >
```
jarv, line 106: I would go ahead and fix the other tasks as well.

playbooks/roles/user/tasks/main.yml
mulby: Unfortunately on the EMR master nodes it is not the default. AFAICT they are actually running some version of Debian instead of LTS ubuntu. They only ship with python 2.6 as well... its frustrating.

If you do decide to merge this we should probably test it on the boxes you typically run against to make sure the regex matches etc.

jarv: Can you provision a sandbox using this branch then and add a comment saying why it's necessary?

playbooks/roles/user/tasks/main.yml
mulby, line 106: Ah, I didn't notice the other ones.

mulby, line 106: Done.

playbooks/roles/user/tasks/main.yml
mulby: Sure thing, firing one up now.

mulby:
Sandbox deploy succeeded: http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/118/console

feanil:
:rocket:

jarv:
@mulby did you ssh in and verify you can run sudo? if so :+1:

mulby:
Yeah I tested sshing in and using sudo, both worked. What's up with travis?

jarv:
maybe rebase? The .travis file was fixed and merged into master today

PR #973, opened 2014-04-02
Build from release instead of master

hkim823:


util/install/vagrant.sh
```diff
@@ -27,7 +28,7 @@ sudo pip install --upgrade virtualenv
 ## Clone the configuration repository and run Ansible
 ##
 cd /var/tmp
-git clone https://github.com/edx/configuration
+git clone -b https://github.com/edx/configuration
```
jarv: doesn't there need to be a branch specifier for the `-b` option?

hkim823: yeah sorry, forgot to type in release

jarv:
:+1:

PR #979, opened 2014-04-03
Remove update ssh config tasks

pku9104038:
Although AWS EC-2 always use key to loggin, but in local deployment, use password to loggin ssh is common. Update the system sshd_config without any warning might cause people can not loggin to their host any more, and that should be a disaster for host manager. It should be more safe to give people manage their sshd_config.


feanil:
@pku9104038 I don't think this is a good change.  It would be better to update ssh config template to make password auth optional with the default set to what it is now.  Although, I think this is less secure than using ssh keys.

PR #982, opened 2014-04-04
Enable tunnelled clear text passwords for sshd_config template

pku9104038:
Make it enabled by sshd_config password option, not by remove the update template task according to Feanil suggestion.


feanil:
@pku9104038 I was suggesting making this an ansible paramater not changing its value.  Something like `{{ COMMON_SSH_PASSWORD_AUTH }}`  with the default value of `COMMON_SSH_PASSWORD_AUTH` set to no.

PR #986, opened 2014-04-04
Add COMMON_SSH_PASSWORD_AUTH for sshd_config template

pku9104038:
An variable for using a ssh password auth in local deployment.


feanil:
@jarv this change seems good to me, can you please review as well.

jarv:
:+1:

PR #987, opened 2014-04-04
For static content, only match static at the root of the URI.

feanil:


e0d:
:+1: I wonder if there are any hidden bugs that will be found once we enable this.

jarv:
ok when do we decide to get rid of lms-preview? :) :+1:

PR #990, opened 2014-04-07
check before installing rabbit

jarv:


feanil:
won't this break our ability to do upgrades?

feanil:
How about doing the install when there is a change to the get_url task to download the deb file?

jarv:
I don't this it's really safe to upgrade in a cluster scenario without some thought put into it since upgrading will prevent the node from joining the cluster

jarv:
> How about doing the install when there is a change to the get_url task to download the deb file?

Will break the base AMI unfortunately since rabbit is installed but doesn't have the deb file on disk yet

feanil:
I thought we were always using base ubuntu for the common cluster?

jarv:
we are, the breakage is happening in the single instance sandbox case that uses the release AMI

feanil:
:+1:

PR #994, opened 2014-04-08
Initial wheel conversions.

e0d:
- Updates to pip calls to use wheel
- removing redundant pips.
- Removing use of local mirroring from edxapp
- Wheel goodness for paver.
- Removing the distribute.
- changing to new bucket

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -231,11 +231,11 @@
     "{{ PIP_EXTRA_ARGS }}"
     --exists-action w --use-mirrors -r {{ item }}
     chdir={{ edxapp_code_dir }}
-    with_items:
-  - "{{ sandbox_base_requirements }}"
-  - "{{ sandbox_local_requirements }}"
-  - "{{ sandbox_post_requirements }}"
-  sudo_user: "{{ edxapp_user }}"
+  with_items:
+    - "{{ sandbox_base_requirements }}"
+    - "{{ sandbox_local_requirements }}"
+    - "{{ sandbox_post_requirements }}"
+    sudo_user: "{{ edxapp_user }}"
   when: "not EDXAPP_PYTHON_SANDBOX"
   notify:
   - "restart edxapp"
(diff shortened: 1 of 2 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -121,6 +97,19 @@
   - "restart edxapp"
   - "restart edxapp_workers"
 
+- name: Upgrade pip
+  pip: >
+    name={{ item }}
+    virtualenv="{{edxapp_venv_dir}}"
+    state=latest
```
jarv: shouldn't we pin these versions?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -128,28 +117,52 @@
     requirements="{{pre_requirements_file}}"
     virtualenv="{{edxapp_venv_dir}}"
     state=present
-    extra_args="-i {{ COMMON_PYPI_MIRROR_URL }}"
+    extra_args="--use-wheel --no-index --find-links=http://edx-wheels.s3-website-us-east-1.amazonaws.com"
```
jarv: let's make the url a var in case we want to put the bucket behind a cdn later.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -128,28 +117,52 @@
...
   notify:
   - "restart edxapp"
   - "restart edxapp_workers"
-  when: not inst.stat.exists or new.stat.md5 != inst.stat.md5
 
-# Install the python modules into {{ edxapp_venv_dir }}
 - name : install python base-requirements
+  pip: >
+    requirements="{{base_requirements_file}}"
+    virtualenv="{{edxapp_venv_dir}}"
```
jarv, line 130: nit pick - missing spaces in jinja var

playbooks/roles/edxapp/tasks/deploy.yml
feanil: I would also like to make all of the extra_args a var so that people can override it to not use wheel.

hkim823:
Hey @e0d still need this PR opened?

PR #995, opened 2014-04-10
Add LTI stub forwarded port.

olmar:
In order to start a LTI stub service manually and use in local development.

@feanil please review.


feanil:
:+1:

PR #998, opened 2014-04-11
include AWS instanceID in prompt

fredsmith:


playbooks/roles/user/templates/default.bashrc.j2
```diff
@@ -53,10 +53,12 @@ if [ -n "$force_color_prompt" ]; then
     fi
 fi
 
+INSTANCEID=$(curl -s http://169.254.169.254/latest/meta-data/instance-id);
```
jarv: I would do something like 

```
command -v ec2metadata >/dev/null 2>&1 && {  INSTANCE_ID=$(ec2metadata --instance-id); }
```

so that users who aren't on AWS can still use this profile.

playbooks/roles/user/templates/default.bashrc.j2
```diff
@@ -53,10 +53,12 @@ if [ -n "$force_color_prompt" ]; then
     fi
 fi
 
+command -v ec2metadata >/dev/null 2>&1 && {  INSTANCE_ID=$(ec2metadata --instance-id); }
+
 if [ "$color_prompt" = yes ]; then
-    PS1='{{ USER_CMD_PROMPT }}${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u@\h\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
+    PS1='{{ USER_CMD_PROMPT }}${debian_chroot:+($debian_chroot)}\[\033[01;32m\]\u@\h $INSTANCEID\[\033[00m\]:\[\033[01;34m\]\w\[\033[00m\]\$ '
```
jarv, line 59: my bad, I think I gave you an example earlier that had an underscore for the instance id var. One of them will need to change.

jarv:
:+1: once the var is fixed

PR #1001, opened 2014-04-15
Add a task to install paver requirements.

jarv:


singingwolfboy:
Why not install paver globally?

hkim823:
looks fine to me

jarv:
@singingwolfboy  I think the issue with that is going to be if other code requires paver (and I assume it will eventually) what we do about multiple versions in multiple virtualenvs. If we always want to use the global paver then I think the thing to do is _not_  install this requirements file via ansible and instead have a global paver installed in the common role.

singingwolfboy:
@jarv Sounds reasonable to me.

PR #1004, opened 2014-04-15
fixing pypi mirror

jarv:


hkim823:
looks good, strange that it wasn't replace globally

PR #1005, opened 2014-04-15
Small improvements to json_lint.sh

ciupicri:
Note: I'm wondering why is it necessarily to keep track of the directory
from which the script was launched. The working directory of the parent
process isn't modified by its children.

### DIFF
--- a/util/json_lint.sh
+++ b/util/json_lint.sh
@@ -9,18 +9,16 @@ cd $(git rev-parse --show-toplevel)
 
 # Do very basic syntax check of every json file to make sure it's valid format
 for file in `find . -iname '*.json'`; do 
-    cat $file | python -m json.tool 1>/dev/null 2>json_complaint.err; 
+    errors=$(python -m json.tool "$file" 2>&1 1>/dev/null)
     retval=$?
     if [ $retval != 0 ]; then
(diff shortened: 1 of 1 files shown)


ciupicri:
@jrbl, could you please have a look at this?

jarv:
I completely forgot about this script.
We are currently doing some basic json and yaml syntax checking in our travis-ci config here -> https://github.com/edx/configuration/blob/master/.travis.yml

hkim823:
@ciupicri otherwise it looks good to me. thanks for your contribution.  I can't seem to find it so if you have already signed out code contributor contribution agreement, thanks again and I will merge this in.  If not, please let us know when you complete and return this.  http://code.edx.org/individual-contributor-agreement.pdf

jarv:
I think we should probably remove this script since we have the travis file that does the same thing.

feanil:
+1 to removing this file.

ciupicri:
I haven't signed the Contributor Agreement yet, but I declare the patches public domain.

Though I guess removing the file seems to be the best option.

hkim823:
If that's the case, what I'll do is close this PR, and setup a PR to just remove the file all together, but thank you again for bringing this to our attention.

PR #1011, opened 2014-04-16
Han/devstack

hkim823:
Allowing password based ssh logins, making sure full ansible is run on current release to allow for this to happen.  This is to allow PyCharm to work properly


fredsmith:
:+1:

PR #1013, opened 2014-04-16
Allow override of edx-ora2 version installed

wedaly:
I'd like to create an integration sandbox that gets updated each night with the latest version of edx-ora2, even if edx-platform has an older version in its requirements.

This PR adds a variable to the `edxapp` role that allows users to `pip install` a version of `edx-ora2` after other edx-platform requirements are installed, but before database migrations are run.

@e0d Does this seem like a reasonable approach?


e0d:
If we're going to do it this way, I'd prefer to have a general concept of EDX_REQUIREMENTS_OVERRIDES and remove the ORA2 specifics.

wedaly:
That makes sense.   I'll close this and open another PR.

PR #1016, opened 2014-04-17
make `bundle install` stop requiring a password in /edx/app/forum

jimabramson:
@e0d i think this will fix the problem...


hkim823:
:+1: tested on my own using a fresh devstack

PR #1019, opened 2014-04-17
Jarv/ansible 1.5 and certs

jarv:


playbooks/roles/certs/tasks/main.yml
```diff
@@ -43,7 +39,7 @@
     shell=/bin/false
   notify: restart certs
 
-- name: create certs app and data dirs
+- name: create certs app dirs
```
e0d, line 42: use edX_service role?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/certs.j2
```diff
@@ -0,0 +1,10 @@
+server {
+  listen {{ CERTS_NGINX_PORT }} default_server;
```
e0d, line 2: Doesn't the default here need to be configurable?

e0d:
Couple small comments, but :+1:

playbooks/roles/certs/tasks/main.yml
jarv, line 42: It's a little bit different than the current service_role setup but I like the idea of using it, maybe take a look at it later?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/certs.j2
jarv, line 2: In this case it doesn't since there are no other vhosts that will be listening on the same port.
We do the same for xserver.

PR #1020, opened 2014-04-18
allow abbey to use additional repos and customizable playbook locations.

fredsmith:
This PR will allow abbey and build-ami to:
- check out a third repo, ansible-private, which contains private playbooks and roles.  
- run playbooks out of a configurable directory, not hard coded to configuration/playbooks/edx-east

### DIFF
--- a/util/vpc-tools/abbey.py
+++ b/util/vpc-tools/abbey.py
@@ -213,7 +213,7 @@ git_repo="https://github.com/edx/$git_repo_name"
 git_repo_secure="{configuration_secure_repo}"
 git_repo_secure_name="{configuration_secure_repo_basename}"
 git_repo_private="{configuration_private_repo}"
-git_repo_private_name="{configuration_private_repo_basename}"
+git_repo_private_basename=$(basename $git_repo_private)
 secure_vars_file="$base_dir/$git_repo_secure_name/{secure_vars}"
 instance_id=\\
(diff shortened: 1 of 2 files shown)


util/vpc-tools/abbey.py
```diff
@@ -286,6 +299,14 @@ def create_instance_args():
     cd $base_dir
 fi
 
+if $config_private; then
```
jarv: where is the `config_private` bash var being set?

util/vpc-tools/abbey.py
```diff
@@ -201,6 +212,8 @@ def create_instance_args():
 git_repo="https://github.com/edx/$git_repo_name"
 git_repo_secure="{configuration_secure_repo}"
 git_repo_secure_name="{configuration_secure_repo_basename}"
+git_repo_private="{configuration_private_repo}"
+git_repo_private_name="{configuration_private_repo_basename}"
```
jarv: It might be less error prone to keep the bash names the same as the python var names.
Instead of passing in the basename you can do something like:

```
git_repo_private_basename=$(basename $git_repo_private)
```

fredsmith: This was mostly copy+paste from the secure repo.  I'll do it the better way instead.

util/vpc-tools/abbey.py
fredsmith: typo - should be git_repo_private or configuration_private as above.

PR #1021, opened 2014-04-18
Update Vagrantfile

yarko:
VirtualBox 4.3.10 has a known bug;  save people from needing to find this, or (worse) forgetting the "sudo" in the "ln -s..." command.   The comment refers to VBox bug report, so it should be simple to remove this 6-12 months down the road.

(This only applies to this Vagrantfile I believe;  does not seem to be pertinent to base/devstack/Vagrantfile, which has no shell provisioning, nor either of the */fullstack Vagrantfiles)

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -8,6 +8,12 @@ if [ ! -d /edx/app/edx_ansible ]; then
     echo "Error: Base box is missing provisioning scripts." 1>&2
     exit 1
 fi
+# VBox 4.3.10 has a bug; make it not a problem
+#  - see https://www.virtualbox.org/ticket/12879
+if [ ! -d /usr/lib/VBoxGuestAdditions ] \
+  && [ -d /opt/VBoxGuestAdditions-4.3.10/lib/VBoxGuestAdditions ]; then
+    sudo ln -s /opt/VBoxGuestAdditions-4.3.10/lib/VBoxGuestAdditions /usr/lib/VBoxGuestAdditions
+fi
 export PYTHONUNBUFFERED=1
 source /edx/app/edx_ansible/venvs/edx_ansible/bin/activate
 cd /edx/app/edx_ansible/edx_ansible/playbooks
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -8,12 +8,7 @@ if [ ! -d /edx/app/edx_ansible ]; then
     echo "Error: Base box is missing provisioning scripts." 1>&2
     exit 1
 fi
-# VBox 4.3.10 has a bug; make it not a problem
-#  - see https://www.virtualbox.org/ticket/12879
-if [ ! -d /usr/lib/VBoxGuestAdditions ] \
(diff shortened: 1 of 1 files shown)


jarv:
Good idea! This appears to be a problem for the fullstack as well. So there are a bunch of vagrantfiles we should update.

```
./base/devstack/Vagrantfile
./base/fullstack/Vagrantfile
./release/devstack/Vagrantfile
./release/fullstack/Vagrantfile
```

Also I don't know if I like running a sudo command like this since it's not a given that the user running it will have full sudo permissions.
How about checking to see if the symlink exists and echo'ing a workaround that the user can paste into his terminal if it doesn't?

hkim823:
Just running a test, I've noticed that I still error out on my initial vagrant up.  Investigating this further.

yarko:
@jarv: on vagrant devstack I think sudo should be OK, convenient, fast (and the place where destroy/up cycles likely to be most frequent). Other spots, could test + message + exit.

Curious to see outcome of @hkim823's tests first.

jarv:
@yarko ah right this is on the guest not the host. I think it's fine. What I have now on our jarv/rc-injera branch is `config.vm.synced_folder  ".", "/vagrant", disabled: true` which I think is working around this bug.

hkim823:
@yarko the new injera release is also going to have the updated version of the VB Guest Additions to match 4.3.10 and things seem great.

yarko:
two things about this PR:

I edited online, modeling manual behavior - but there is a problem which explains @hkim823's perceived problem:  the sh provisioning happens _after_ NFS mounts (which is why they fail).

Therefore, this PR - in the state it's in - is incorrect.

Also, if @jarv's injera release walks around this, then it's unecessary.
Either way, I"m closing this and will look at again after injera release (assuming it's relatively soon, as an RC).

hkim823:
@yarko i noticed that the script only runs only after VBGuests are updated and tries to do the mount as a test (which of course fails) and currently the auto_reboot flag doesn't seem to work or at least not work as I expect it, I put in an issue into the vagrant-vbguest team.  

I think injera is right around the corner.

yarko:
I've re-opened this.   This form looks to work:  place the link shell test & command in a separate script, and run it early (before nfs mounting).

@jarv @hkim823 - would this be useful now and in injera?

aside:  perhaps as part of this VBox 4.3.10 problem, I had (as a result?) a corrupted /etc/entries file on my mac host; clearing helped.

yarko:
FYI - looking over injera RC,  I'm not clear why /vagrant sync'ing disabled (and it would block what I've had in the wings, so want to understand).

In any case, I think this PR (as it is now, working) is good - it looks like even for injera (?), so please have another look, @jarv, @hkm823 -  let me know if you think it's ok, and belongs in the other devstack Vagrantfile also (don't think it applies to fullstack, but not fully sure about that).

yarko:
Ok - now that injera is released, I've been running & testing, and see why disabling NFS on "." helps (since bunches of directory contents of "." are involved in other mount-points).

What I want is simple access to files in ".", and it turns out that rsync option for vm.synced_folder works well (performance and update wise), and also avoids the problem which disabling NFS on "." does.   This is different enough, I'm closing this, and will provide a clean, separate PR - with the build-version.yml I am seeking to put in place.

PR #1029, opened 2014-04-22
adding play to set the hostname to the deploy host

jarv:


fredsmith:
:+1:

util/jenkins/ansible-provision.sh
```diff
@@ -179,4 +179,7 @@ fi
 # deploy the edx_ansible role
 ansible-playbook edx_ansible.yml -i "${deploy_host}," -e@${extra_vars_file} -e@${WORKSPACE}/configuration-secure/ansible/vars/developer-sandbox.yml --user ubuntu
 
+# set the hostname
+ansible-playbook set_hostname.yml -i "${deploy_host}," -e hostname_fqdn=${deploy_host} --user ubuntu
```
hkim823, line 183: This assumes an EC2 instance with the ubuntu user have sudo access, what about other scenarios?  will it just run without the --user?

jarv, line 183: yeah whatever user this is run as will require sudo.

PR #1032, opened 2014-04-23
Fixes to abbey external playbook support

fredsmith:


util/vpc-tools/abbey.py
```diff
@@ -213,7 +213,7 @@ def create_instance_args():
 git_repo_secure="{configuration_secure_repo}"
 git_repo_secure_name="{configuration_secure_repo_basename}"
 git_repo_private="{configuration_private_repo}"
-git_repo_private_basename=$(basename $git_repo_private)
+git_repo_private_name=$(basename $git_repo_private .git)
```
jarv, line 216: why not call it `git_repo_private_basename`?

jarv:
:+1:  other than the nitpick

util/vpc-tools/abbey.py
fredsmith, line 216: to make the naming consistent with git_repo_secure_name.

PR #1034, opened 2014-04-23
Jarv/clean ami

jarv:
These changes allow for easy creation of a public AMI for when we do configuration releases.
ansible-provision.sh reads a new boolean "edx_internal" that when unchecked will not use the sandbox secret vars or install the xserver.

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -115,14 +115,14 @@ if [[ $basic_auth == "true" ]]; then
 
     if [[ $public_ami == "true" ]]; then
         cat << EOF_AUTH >> $extra_vars_file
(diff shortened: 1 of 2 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -43,6 +43,14 @@ fi
 
 extra_vars_file="/var/tmp/extra-vars-$$.yml"
 
+if [[ $edx_internal == "true" ]]; then
```
e0d, line 47: nit: clearer to me to set 

extra_var_arg="-e@${extra_vars_file}" 

outside the conditional and append the additional arg if the condition is met.

e0d:
LGTM

PR #1038, opened 2014-04-24
Updating DNS script for multiple accounts in the edx.org zone

jarv:
This PR makes the following changes:
- Defaults to the edx.org zone instead of a vpc zone
- Takes an AWS id and secret to query another account for instance info
- Adds some safe guards around duplicate dns entries
- Creates an ban list for elbs
- pep8 cleanup


e0d:
:+1:

PR #1041, opened 2014-04-25
update port forward for comments service and add 9200 for elasticsearch

jimabramson:
devstack needs comments service running on 18080, so updated the port forward to make it directly accessible from the host.  in addition, made elasticsearch accessible from the host as well.

I know there are several other Vagrantfiles where these directives exist, let me know which of these would need the same changes applied and I'll add.

### DIFF
--- a/vagrant/base/devstack/Vagrantfile
+++ b/vagrant/base/devstack/Vagrantfile
@@ -27,7 +27,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.network :private_network, ip: "192.168.33.10"
   config.vm.network :forwarded_port, guest: 8000, host: 8000
   config.vm.network :forwarded_port, guest: 8001, host: 8001
-  config.vm.network :forwarded_port, guest: 4567, host: 4567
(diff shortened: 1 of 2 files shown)


e0d:
:+1:

hkim823:
isn't 4567 used as the api port for forums?

e0d:
4567 is the sinatra webbrick port, we are now running under unicorn and have the service fronted  by nginx.

jimabramson:
if making edxapp look for forums on 4567 in devstack is a thing that makes sense, we could do that instead (would be a little more user-friendly too).  not sure how to implement that though.

jimabramson:
should the Vagrantfile under vagrant/base/devstack get the same changes?
any of the fullstack ones?

hkim823:
i'd make the same changes to the devstack file in base, they should match what's in release

probably can leave fullstack be

jimabramson:
re-pushed with port forwards synced to base devstack Vagrantfile as well.
also planning to fix port forward instructions on [the devstack wiki](https://github.com/edx/configuration/wiki/edX-Developer-Stack#forum-workflow) so they align with this change.

hkim823:
:+1:

PR #1043, opened 2014-04-25
adding play mappings for the commoncluster

jarv:


util/vpc-tools/vpc_dns.py
```diff
@@ -218,17 +226,25 @@ def update_elb_rds_dns(zone):
...
-                dns_records.add(DNSRecord(zone, fqdn, 'CNAME', 600, [elb.dns_name]))
-        else:
-            env_tag, deployment_tag, play_tag = get_dns_from_instances(elb)
-            fqdn = "{}-{}-{}.{}".format(env_tag, deployment_tag, play_tag, zone_name)
-            if elb.dns_name not in ELB_BAN_LIST:
-                dns_records.add(DNSRecord(zone, fqdn, 'CNAME', 600, [elb.dns_name]))
+        # Override the play tag if a substring of the elb name
+        # is in ELB_PLAY_MAPPINGS
+
+        for key in ELB_PLAY_MAPPINGS.keys():
```
e0d, line 240: Should we consider it an error if there is no match?

e0d:
Other than the one question, LGTM.

util/vpc-tools/vpc_dns.py
jarv, line 240: This ELB_PLAY_MAPPINGS dict is only for exceptions, specifically elbs that have have common cluster server in them. For most cases there will not be a match.

PR #1045, opened 2014-04-26
Update dumpall.yml.j2

fmyzjs:
add some common variables to update repo by using local mirror


hkim823:
I'm not sure this is the right place for this, you can update the COMMON variables inside of https://github.com/edx/configuration/blob/master/playbooks/roles/common/defaults/main.yml and your settings are already in that file.

fmyzjs:
Thx，I will try it later

PR #1049, opened 2014-04-28
Archive the current virtualenv state as a starting point for new builds

jzoldak:
This goes with [edx-platform/pull/3498](https://github.com/edx/edx-platform/pull/3498), though they can be merged independently.


feanil:
looks good.

PR #1050, opened 2014-04-28
A script to clone a database into a vpc using stack-name

jarv:
There is some work outstanding before this is ready but I believe
it's complete enough to merge.  There will be more logic added
to sanitize and change the db password after the clone is complete


feanil:
looks good.

PR #1052, opened 2014-04-29
adding an sql script for sanitizing the prod mysql db

jarv:


util/vpc-tools/sanitize-db.sql
```diff
@@ -0,0 +1,86 @@
...
+UPDATE wwc.auth_user
+    set email = concat('success+',cast(id AS CHAR),'@simulator.amazonses.com')
+        where email not like ('%@edx.org'),
+    set username = concat('user-',cast(id AS CHAR)
+        where email not like ('%@edx.org'),
+    set first_name = concat('user-',cast(id AS CHAR)
+        where email not like ('%@edx.org'),
+    set last_name = concat('user-',cast(id AS CHAR)
+        where email not like ('%@edx.org'),
+    set password = null
```
e0d: I'm a little surprised this is nullable.

jarv: That's the default, maybe set it to an empty string instead?

util/vpc-tools/sanitize-db.sql
```diff
@@ -0,0 +1,86 @@
+SET FOREIGN_KEY_CHECKS=0;
+
+/* 
+  Truncate the courseware_studentmodulehistory table since
+  it is only needed for analytics    
+*/
+
+TRUNCATE courseware_studentmodulehistory;
```
e0d: Should probably put some conditional logic around this in case we handle leaving this out upstream, say by using manual export commands.

jarv: Since it's not part of the sanitation I'll remove it and run the truncate separately.

util/vpc-tools/sanitize-db.sql
e0d: If it's nullable, I'm fine with this; I'm just surprised.

e0d:
Other than the small comment, :+1:

PR #1055, opened 2014-04-29
Modify jenkins worker task to run the wheel script 

jzoldak:
to populate the python virtualenv.

I didn't realize this was being done by the Jenkins master config, and needs to be moved here when I submitted #1049.

Needed to get [edx-platform/pull/3498](https://github.com/edx/edx-platform/pull/3498) working, though it can be merged independently.

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/python.yml
+++ b/playbooks/roles/jenkins_worker/tasks/python.yml
@@ -45,11 +45,26 @@
     src=wheel_venv.sh.j2 dest={{ jenkins_home }}/wheel_venv.sh
     owner={{ jenkins_user }} group={{ jenkins_group }} mode=700
 
+# Run the wheel_venv.sh script for the first time
+# This was previously done in the Jenkins global
+# configuration as part of the AMI Init script.
+# Moving here so that we can archive a clean snapshot
+# of the virtualenv with only the defined packages
+# from jenkins_wheels.
+- name: Run the wheel_venv.sh script
+  command: >
+    ./wheel_venv.sh edx-venv
+    chdir={{ jenkins_home }}
+    creates={{ jenkins_home }}/edx-venv
+  sudo_user: "{{ jenkins_user }}"
+
 # Archive the current state of the virtualenv
-# as a starting point for new builds
+# as a starting point for new builds.
+# The edx-venv directory is deleted and then recreated
+# cleanly from the archive by the jenkins build scripts.
 - name: Create a clean virtualenv archive
   command: >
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -45,11 +45,26 @@
     src=wheel_venv.sh.j2 dest={{ jenkins_home }}/wheel_venv.sh
     owner={{ jenkins_user }} group={{ jenkins_group }} mode=700
 
+# Run the wheel_venv.sh script for the first time
+# This was previously done in the Jenkins global
```
feanil, line 49: Does this need to be removed from the global init or does it still make sense to run it there as well?

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -45,11 +45,26 @@
...
+  sudo_user: "{{ jenkins_user }}"
+
 # Archive the current state of the virtualenv
-# as a starting point for new builds
+# as a starting point for new builds.
+# The edx-venv directory is deleted and then recreated
+# cleanly from the archive by the jenkins build scripts.
 - name: Create a clean virtualenv archive
   command: >
-    tar -cpzWf {{ jenkins_venv }}_clean.tar.gz {{ jenkins_venv }}
```
feanil, line 52: why use an explicit name instead of the jenkins_venv variable?

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak, line 49: It should be removed. Won't cause problems if it's not, just it is redundant will cause the workers to take more time to spin up. I will remove it at the same time as I update it to the new AMI that I create after this PR gets merged.

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak, line 52: My last PR was incorrect. That variable isn't the correct value. From main.yml:
jenkins_venv: "{{ jenkins_home }}/wheel_venv"

That's the wheel venv. The one we want to archive is {{ jenkins_home }}/edx-venv

Would you prefer I make a new variable and stick that in main.yml too?

feanil, line 52: No this is fine.

playbooks/roles/jenkins_worker/tasks/python.yml
feanil, line 49: huh, for some reason I thought we were dropping that as a template file for the worker role but that doesn't seem to be the case.

feanil:
:+1:

PR #1057, opened 2014-04-29
Stanford only: move tmp dir from edx-west/tmp to edx-west/ec2_cache

sefk:
Per feedback from @feanil on PR #849, want our release branch to line up with what we're pulling into master.  I'm making that change separately on that PR.  @jrbl -- this is just the same change on edx-west/release.


jbau:
@sefk can you add a "data" directory too for the analytics vpc

sefk:
good catch @jbau, added and repushed

playbooks/edx-west/prod-ansible.cfg
```diff
@@ -18,5 +18,5 @@ user=ubuntu
 
 [ssh_connection]
 # example from https://github.com/ansible/ansible/blob/devel/examples/ansible.cfg
-ssh_args= -o ControlMaster=auto -o ControlPersist=60s -o ControlPath=tmp/prod/ansible-ssh-%h-%p-%r
+ssh_args= -o ControlMaster=auto -o ControlPersist=60s -o ControlPath=/tmp/ansible-ssh-%h-%p-%r
```
sefk, line 21: From commit msg:

> Going back to putting ssh connections in /tmp.  They have globally
> unique names, putting in /tmp is consisten with other things, and having
> them live in a dir named ec2_cache feels odd now.

jrbl:
lgtm.

sefk:
Thanks @jrbl, merged

PR #1064, opened 2014-05-05
adding stage config

jarv:


feanil:
looks good.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -390,17 +411,19 @@ generic_env_config:  &edxapp_generic_env
...
     celery:
       <<: *default_generic_cache
       KEY_PREFIX:  'integration_celery'
+      TIMEOUT: "7200"
   CELERY_BROKER_TRANSPORT: 'amqp'
   CELERY_BROKER_HOSTNAME: $EDXAPP_RABBIT_HOSTNAME
   COMMENTS_SERVICE_URL: $EDXAPP_COMMENTS_SERVICE_URL
   LOGGING_ENV: $EDXAPP_LOGGING_ENV
-  SESSION_COOKIE_DOMAIN:  !!null
+  SESSION_COOKIE_DOMAIN:  $EDXAPP_SESSION_COOKIE_DOMAIN
```
feanil, line 428: This is going to result in the default being set to blank instead of null is that going to cause any issues?

feanil:
:+1:  other than question above.

playbooks/roles/edxapp/defaults/main.yml
jarv, line 428: Yeah setting this to `!!null` doesn't work, setting it to a blank string instead.

PR #1065, opened 2014-05-05
Jenkins should save the state of the rbenv

singingwolfboy:
Companion to #1049, but for Ruby instead of Python. The companion edx-platform pull request is https://github.com/edx/edx-platform/pull/3603. @feanil @jzoldak, can you two review?

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/main.yml
+++ b/playbooks/roles/jenkins_worker/tasks/main.yml
@@ -10,4 +10,5 @@
 
 - include: system.yml
 - include: python.yml
+- include: ruby.yml
 - include: jscover.yml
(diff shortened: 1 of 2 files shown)


feanil:
:+1:

jzoldak:
@singingwolfboy did you run the playbook to verify that this works correctly?
I think you'd need to run a bundle install the first time to populate the rbenv. Does this get done somewhere?

feanil:
The bundle install gets done as part of the rbenv role which is a dependency(https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_worker/meta/main.yml)

singingwolfboy:
@jzoldak I just successfully tested the playbook locally, with a fair amount of help from @feanil. It works the way I expected.

PR #1068, opened 2014-05-06
Han/fix motd

hkim823:
Not sure why this wasn't being caught before or throwing errors, but it did trying to build from scratch testing ansible 1.6


jarv:
:+1:

PR #1070, opened 2014-05-07
optionally override cms hostname in nginx

fredsmith:


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
```diff
@@ -27,8 +27,11 @@ server {
   listen {{EDXAPP_CMS_NGINX_PORT}} {{default_site}};
   {% endif %}
 
-
+  {% if cms_hostname is defined %}
```
feanil: If this is an override, we should make it an all caps variable.

e0d:
why not just override the var rather than have conditional logic in the template.

jarv:
:+1:

jarv:
@fredsmith Yeah I think I agree with @e0d and would put `~^((stage|prod)-)?studio\..*` in a var

fredsmith:
Ah, that's a good idea.  thanks!

fredsmith:
can I get another :+1: on this?

feanil:
:+1:

PR #1071, opened 2014-05-07
creating separate options to clean the prod_grader and wwc dbs

jarv:
Needed for edge which doesn't have a prod_grader db


feanil:
looks good.

PR #1082, opened 2014-05-12
Added pyexpat to whitelisted shared objects, and flipped apparmor mode to enforce.

davestgermain:
@e0d


playbooks/roles/edxapp/templates/code.sandbox.j2
```diff
@@ -1,6 +1,6 @@
 #include <tunables/global>
 
-{{ edxapp_sandbox_venv_dir }}/bin/python flags=(complain) {
+{{ edxapp_sandbox_venv_dir }}/bin/python flags=(enforce) {
```
e0d: This is the default

e0d:
:+1:

PR #1083, opened 2014-05-12
Removing similar lines, updating mask

e0d:
@davestgermain 

Missed that you had added new lines for this, they were at the top of the file and just needed a mask change.


davestgermain:
:+1: case of the Mondays... :coffee:

PR #1086, opened 2014-05-13
fix NOTIFIER_DD_API_KEY to follow convention

jbau:
@feanil @hkim823 @jarv 

A simple bugfix, I think.  DATADOG_API_KEY was not following convention in notifier playbook


e0d:
I don't think this is the right thing to do as typically there would be a single dd key that would be used by all of your applications.  Doing 

FOO_DD_API_KEY: "{{ DD_API_KEY }}" 

in each role seems like a bit much.  Either way, this will definitely break our current installation without secrets changes.

jbau:
ok. well, it seemed a bit inconsistent with the rest of notifier, and it was the other way at somepoint, but nbd.

e0d:
Upon reflection, the right thing to do here is probably to merge this and use this convention:

FOO_DD_API_KEY: "{{ DD_API_KEY }}"

In our private vars files, though it is clutter, it makes each individual role stand on its own.

jbau:
OK.  I'll leave the merging up to you guys, then, (i.e. when you're done modifying the secrets files)

jarv:
The datadog api key has been moved into common so I don't think we want to merge this

PR #1093, opened 2014-05-14
upgrading boto

jarv:
The new version of boto doesn't except 'all' as a region, defaulting our
ini to use us-east-1


feanil:
:+1:

PR #1098, opened 2014-05-19
Add update_theme bash to edx_ansible role

dangtrinhnt:
This bash script will help you to update the theme independently without updating the whole stack (using update script)

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,3 +29,4 @@ Ker Ruben Ramos <xdiscent@gmail.com>
 Fred Smith <derf@edx.org>
 Wang Peifeng <pku9104038@hotmail.com>
 Ray Hooker <ray.hooker@gmail.com>
+Trinh Nguyen <dangtrinhnt@gmail.com>
--- a/playbooks/roles/edx_ansible/tasks/deploy.yml
+++ b/playbooks/roles/edx_ansible/tasks/deploy.yml
@@ -22,6 +22,17 @@
     dest={{ COMMON_BIN_DIR }}/update
     state=link
 
+- name: create update theme script
+  template: >
+    dest={{ edx_ansible_app_dir}}/update_theme
+    src=update_theme.j2 owner={{ edx_ansible_user }} group={{ edx_ansible_user }} mode=755
+
+- name: create a symlink for update_theme.sh
+  file: >
+    src={{ edx_ansible_app_dir }}/update_theme
+    dest={{ COMMON_BIN_DIR }}/update_theme
+    state=link
+
 - name: create a symlink for ansible-playbook
   file: >
     src={{ edx_ansible_venv_bin }}/ansible-playbook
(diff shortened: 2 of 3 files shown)


feanil:
I think a better approach would be to tag the ansible tasks that are needed for themes to get re-loaded and be able to use the existing update script.  Add a new option for themes that only runs the deploy tags for tasks that are also tagged for themes(https://github.com/edx/configuration/blob/master/playbooks/roles/edx_ansible/templates/update.j2).

dangtrinhnt:
I just dont want to mess up the update script and be able to do the theme reloading easily without remembering any parameter.

yarko:
If you need to just update a theme, you have little more to do than you would to run your script:

As edxapp user (environment & shell all setup, as per any maintenance task on fullstack, as per
https://github.com/edx/configuration/wiki/edX-Managing-the-Production-Stack#compile-assets-manually):

all you need to is:

```
git -C ../themes/out_theme  pull
paver update_assets lms --settings=aws
```

and / or if you want:

```
paver update_assets cms --settings=aws
```

If anything, for convenience (and there's some point to it), updating
`edx-platform/paverlib/assets::update_assets`  would be the way to do this.
I do not recommend adding a a shell script to the repository,
and an ansible task / updating ansible seems at the wrong level.

Here's what I suggest:
- do this manually for now (it's simply too easy w/ paver / rake / whatever-may-come-next);
- wait for others to chime in (I suspect there will be more theme usage);

If people agree, then the way I suggest is add a `--update_theme` flag
to `paver update_assets`, and conditioned on that being set,
referring to the appropriate env file:  `/edx/app/edxapp/{lms,cms}.env.json`:

```
  if `edxapp_use_custom_theme` is true, then:
      run:  `"git -C ../themes" / edxapp_theme_name + " pull"`
```

before the other asset commands.

This is simple enough, consistent with the level of work being done
(_a superset of asset_updating_), and presupposes a naturally useful constraint:
- You have already at least once done an "update" which included your desired theme repo & version/branch pull.

If you want to change the branch or repo, then to ensure other commands reading
`lms.env.json` are consistent, the prerequisite in those cases would be to update.

yarko:
As for Feanil's suggestion to modify update.j2:   please don't.

For one thing, right now this is unnecessary `update.j2` already works correctly when you add the appropriate variable values to server-vars.yml.

And I am still updating `update.j2`  so that you are not editing `server-vars.yml` (which is generated by ansible) but rather so that is also looks for (appropriately named) files:  `server-site-vars.yml` (?) (which would be for your edited site & server config values / overrides you want to set for your deployment) and `build-version.yml`  (which would control which repositories and versions your site is built from).

feanil:
I assume the reason for the script is because @dangtrinh wants to iterate over the theme faster while developing, in which case adding a few tags and a new parameter to the update script would make it pretty easy.

yarko:
@feanil @dangtrinh  - it would be, but it's not a provisioning dancestep as much as an extended assets update activity - that, and the implied complexity is why I'd prefer paverlib/assets.py be the action place (if anyplace).

dangtrinhnt:
@feanil : Yes. It's just a quick solution I could think of.
@yarko : You're right. I will dig into the paverlib.

So, I think I will close this PR.

yarko:
Paver lib/assets.py should be trivial to add an option flag ( just mimic
what's there).

PR #1099, opened 2014-05-19
Upgrade setuptools with pip upgrade to prevent pkg_resources errors

carsongee:
These two packages need to roughly sync up or newly installed package dependencies aren't found by the system's crusty setuptools.


feanil:
:+1:

PR #1103, opened 2014-05-19
Changes to Common Role

hkim823:


feanil:
What's the motivation for these changes?

hkim823:
When setuptools was set to 3.6, i looked up what version matches virutalenv and came up with https://pypi.python.org/pypi/virtualenv/1.11.6 which then led me to say we should go to pip 1.5.6.

feanil:
Ok, that makes sense, what about the MOTD changes?

hkim823:
For the change to motd, my initial testing with Ansible 1.6 and Ubuntu 14.04 throws an error when those files don't exist (but in 12.04 it didn't seem to care if they don't exist), it seems the purpose of that task is to just set file permissions on files in that directory, the change just makes it all files within that directory.

hkim823:
One more change, the commit labeled Typo isn't really a typo correction, it's adding 99-footer back since in Ubuntu 14.04 it's taken out but the functionality still and will continue to work with /ec/update-motd.d/ for exactly what we're using it for, a static MOTD.  

https://launchpad.net/ubuntu/trusty/+source/base-files/+changelog
https://bugs.launchpad.net/ubuntu/+source/manpages/+bug/785901

playbooks/roles/common/tasks/main.yml
```diff
@@ -105,19 +105,22 @@
...
 - name: update the ssh motd on Ubuntu
   file: >
-    mode=0644
-    path={{ item }}
-  with_items:
-  - "/etc/update-motd.d/10-help-text"
-  - "/usr/share/landscape/landscape-sysinfo.wrapper"
-  - "/etc/update-motd.d/51-cloudguest"
-  - "/etc/update-motd.d/91-release-upgrade"
+    mode=0755
```
feanil: I still don't understand why this is necessary.  Also setting recursive 0755 on the directory will mean that the files will also be executable which we don't want.

hkim823:
Squashed commits into one

playbooks/roles/common/tasks/main.yml
```diff
@@ -112,12 +113,20 @@
   - "/usr/share/landscape/landscape-sysinfo.wrapper"
   - "/etc/update-motd.d/51-cloudguest"
   - "/etc/update-motd.d/91-release-upgrade"
+  ignore_errors: yes
```
feanil: Why are we ignoring errors here?

hkim823:
I'm going to close this PR for now, reopen a new one for pip requirements, look more into the ssh update scripts and how best to approach them without using an ignore_errors.

PR #1106, opened 2014-05-21
explicityly create_user for ubuntu 12.04 option

hkim823:
Ubuntu 12.04 base box isn't getting GitHub ssh user created by default, adding that.


feanil:
:+1:

PR #1107, opened 2014-05-21
make URLs list overrideable

fredsmith:


feanil:
I think we should leave the comments sections we are editing but other than that this looks good.

PR #1109, opened 2014-05-21
Using built in VBox Shared Folders instead of nfs

hkim823:


feanil:
@hkim823 the reason we've been using NFS instead of VBox shared folders is because NFS is significantly faster in terms of IO so it's easier when people are developing in their IDEs on top of it.

hkim823:
I'm going to close this PR, there's no real good way to have both NFS and VBOXFS in the same Vagrantfile without using shell env which isn't ideal.

PR #1115, opened 2014-05-23
Working role for review.

e0d:


playbooks/roles/xqwatcher/defaults/main.yml
```diff
@@ -0,0 +1,57 @@
...
+
+xqwatcher_app_dir: "{{ COMMON_APP_DIR }}/{{ xqwatcher_service_name }}"
+xqwatcher_home: "{{ COMMON_APP_DIR }}/{{ xqwatcher_service_name }}"
+xqwatcher_venv_base: "{{ xqwatcher_home }}/venvs"
+xqwatcher_venv_dir: "{{ xqwatcher_venv_base }}/{{ xqwatcher_service_name }}"
+xqwatcher_code_dir: "{{ xqwatcher_app_dir }}/src"
+xqwatcher_conf_dir: "{{ xqwatcher_home }}/conf.d"
+xqwatcher_data_dir: "{{ xqwatcher_home }}/data"
+
+xqwatcher_source_repo: "git@github.com:edx/xqueue-watcher.git"
```
jarv: could replace `github.com` with `{{ COMMON_GIT_MIRROR }}` here

playbooks/roles/xqwatcher/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
...
+#
+#
+#
+# Handlers for role xqwatcher
+# 
+# Overview:
+# 
+#
+- name: notify me
+  debug: msg="stub handler"
```
jarv: would like to delete the stub if one isn't necessary

playbooks/roles/xqwatcher/defaults/main.yml
```diff
@@ -0,0 +1,57 @@
+---
```
jarv, line 1: Should `XQWATCHER_COURSES` be set to an empty list in defaults/main.yml with a pointer to the task file for an explanation?

playbooks/roles/xqwatcher/tasks/deploy.yml
```diff
@@ -0,0 +1,17 @@
...
+
+- name: upload ssh script
+  template: >
+    src=git_ssh.sh.j2 dest=/tmp/git_ssh.sh
+    owner={{ xqwatcher_user }} mode=750
+    
+- include: deploy_watcher.yml tags=deploy-watcher
+
+- include: deploy_courses.yml tags=deploy-courses
+
```
jarv: in order for the `deploy` tag to propagate on these includes you will need to do something like:

```
- include: deploy_courses.yml
  tags:
    - deploy-courses
    - deploy
```

playbooks/roles/xqwatcher/tasks/deploy_courses.yml
```diff
@@ -0,0 +1,10 @@
+- name: checkout grader code
+  git: >
+    dest={{ xqwatcher_data_dir }}/{{ item.COURSE }} repo={{ item.GIT_REPO }}
+    version={{ item.GIT_REF }}
+    accept_hostkey=yes
+  environment:
+    GIT_SSH: /tmp/git_ssh.sh
+  sudo: true
+  sudo_user: "{{ xqwatcher_user }}"
+  with_items: XQWATCHER_COURSES
```
jarv, line 10: no strong preference but this seems short to warrant a separate task file.

playbooks/roles/xqwatcher/tasks/deploy.yml
```diff
@@ -0,0 +1,17 @@
...
+  template: >
+    src=git_ssh.sh.j2 dest=/tmp/git_ssh.sh
+    owner={{ xqwatcher_user }} mode=750
+    
+- include: deploy_watcher.yml tags=deploy-watcher
+
+- include: deploy_courses.yml tags=deploy-courses
+    
+- name: remove read-only ssh key for the content repo
+  file: path={{ xqwatcher_git_identity }} state=absent
```
jarv, line 17: title needs an update.

playbooks/roles/xqwatcher/tasks/deploy_watcher.yml
```diff
@@ -0,0 +1,40 @@
+- name: checkout watcher code
+  git: >
+    dest={{ xqwatcher_code_dir }} repo={{ xqwatcher_source_repo }} version={{ xqwatcher_version }}
+    accept_hostkey=yes
+  environment:
+    GIT_SSH: /tmp/git_ssh.sh
```
jarv: no strong preference but we could try using the `ssh_opts` parameter to the git module that was added recently.

playbooks/roles/xqwatcher/defaults/main.yml
jarv, line 1: Actually we might want to default to the demo course, or at least include what it would look like for the demo course in a comment block.

playbooks/roles/xqwatcher/tasks/main.yml
```diff
@@ -0,0 +1,117 @@
...
+#
+# You would use a commone like the following to run the play.
+#
+# ansible-playbook -i ec2.py ./xqwatcher.yml -e@./example-config.yml
+#
+# The contents of the example-config.yml would include the queue
+# meta data and details related to the repository including the
+# grader code.
+#
+# XQWATCHER_COURSES:
```
jarv, line 50: update with demo course example?

playbooks/roles/xqwatcher/templates/edx/app/supervisor/conf.d/xqwatcher.conf.j2
```diff
@@ -0,0 +1,24 @@
...
+stopwaitsecs=10
+redirect_stderr=false
+stdout_logfile={{ xqwatcher_supervisor_log_dir }}/%(program_name)-stdout.log
+stdout_logfile_maxbytes=1MB
+stdout_logfile_backups=10
+stdout_capture_maxbytes=1MB
+stderr_logfile={{ xqwatcher_supervisor_log_dir }}/%(program_name)-stderr.log
+stderr_logfile_maxbytes=1MB
+stderr_logfile_backups=10
+stderr_capture_maxbytes=1MB
```
jarv: for our other services we are using the defaults. We could create a common config for all the services instead of settings these per service.

playbooks/roles/xqwatcher/tasks/main.yml
e0d, line 50: The existing demo course is grading Python using our existing xserver setup.  I think what we should do here is add a sample code grader to the demo course repo and install it as an xqwatcher grader on single instance sandboxes.

playbooks/roles/xqwatcher/tasks/deploy.yml
```diff
@@ -0,0 +1,25 @@
+- name: install read-only ssh key
+  copy: >
+    content="{{ XQWATCHER_GIT_IDENTITY }}" dest={{ xqwatcher_git_identity }}
+    owner={{ xqwatcher_user }} group={{ xqwatcher_user }} mode=0600
+
+# - name: upload ssh script
```
feanil: Shouldn't this be uncommented since we're accessing private repositories?

playbooks/roles/xqwatcher/tasks/deploy.yml
```diff
@@ -0,0 +1,25 @@
...
+#   template: >
+#     src=git_ssh.sh.j2 dest=/tmp/git_ssh.sh
+#     owner={{ xqwatcher_user }} mode=750
+
+- name: remove read-only ssh key for the content repo
```
feanil: If we're gonna use the git identity this should be moved to the section after all the code checkouts happen.

playbooks/roles/xqwatcher/tasks/deploy.yml
e0d: No, it should have been removed entirely in vafor of using ssh_opts in the actual git commands.

playbooks/roles/xqwatcher/handlers/main.yml
```diff
@@ -0,0 +1,16 @@
+---
```
jarv, line 1: left here on purpose?

jarv:
Other than the comment about `XQWATCHER_COURSES` :+1:

e0d:
I've added the XQWATCHER_COURSES block to defaults.  I'd like to integrated with the demo course, but will consider that a separate effort.

PR #1120, opened 2014-05-27
Set rbenv tar target to .rbenv instead of /home/jenkins

benpatterson:
This will make it consistent with how we pack and unpack the
virtualenv (in terms of directory structure).

CC @singingwolfboy @jzoldak For additional info on unpacking, see edx-platform/jenkins/all-tests.sh

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/ruby.yml
+++ b/playbooks/roles/jenkins_worker/tasks/ruby.yml
@@ -6,7 +6,7 @@
 # cleanly from the archive by the jenkins build scripts.
 - name: Create a clean rbenv archive
   command: >
-    tar -cpzf edx-rbenv_clean.tar.gz {{ jenkins_rbenv_root }}
+    tar -cpzf edx-rbenv_clean.tar.gz .rbenv
     chdir={{ jenkins_home }}
     creates={{ jenkins_home }}/edx-rbenv_clean.tar.gz
   sudo_user: "{{ jenkins_user }}"


singingwolfboy:
This seems like it's making the code less flexible. What's the reason for this change?

feanil:
agreed, why not just update the value of jenkins_rbenv_root?

benpatterson:
As it's currently coded, the tar will unpack to a directory /home/jenkins/.rbenv/<etc>. This means that the full path will actually look like /home/jenkins/home/jenkins/.rbenv/<etc>.

The python virtualenv is a hardcoded root directory in that tar as well (edx-venv), so this follows convention. The alternative is to change the scripts in the shell file, but I opted for keeping the consistency there.

For @feanil 's question, I see the jenkins_rbenv_root is used in a couple of other places, and those are under different contexts from what I can see, which means using the same var in both places will not be compatible.

singingwolfboy:
I don't understand. As far as I can tell, `jenkins_rbenv_root` is defined to be `{{ jenkins_home }}/.rbenv`, and `jenkins_home` is defined to be `/home/jenkins`. It looks to me like `jenkins_rbenv_root` is an absolute path, not a relative one -- so it shouldn't be appended to the current working directory. Are you actually seeing this behavior where it's unpacking to `/home/jenkins/home/jenkins/.rbenv/`? Can you print out the command to see if `jenkins_rbenv_root` is in fact a relative path rather than an absolute one? Because that seems like the root issue here.

benpatterson:
Yes, the resulting tar file includes /home/jenkins.

However, taking a second to look again at the ansible doc, I'm wondering if it is the use of the '>' character. Because the configuration of the command task uses the chdir setting; however, the resulting tar file includes /home/jenkins.

http://docs.ansible.com/command_module.html

Perhaps it's a matter of removing the '>'. I will give that a go to see if it corrects the problem.

feanil:
I see the issue now.  This is actually an artifact of the tar command. If passed an absolute path, it uses that path in the location of files it tars(stripping the leading slash).  The other way to fix this would be to unzip the tar at the root instead of within the jenkins_home directory.  This seems more secure. :+1:

benpatterson:
Thanks @feanil . 

BTW, removing the greater-than didn't alter behavior. 

I've verified on an EC2 instance that this change gets us past the path issue on future jenkins workers. I'm planning to merge once I'm done the rest of my test cycle.

benpatterson:
Actually, @singingwolfboy does this sound alright to you (i.e. :thumbsup:?)

At this point I've verified an AMI candidate using this PR.

singingwolfboy:
If you've verified that it works, then it's fine with me. :+1:

PR #1121, opened 2014-05-27
updating abbey to use a deployment.yml var file

jarv:


e0d:
:+1:

util/vpc-tools/abbey.py
```diff
@@ -48,11 +48,12 @@ def parse_args():
...
                         help="don't actually run the cmds",
                         default=False)
-    parser.add_argument('--secure-vars', required=False,
-                        metavar="SECURE_VAR_FILE",
+    parser.add_argument('--secure-vars-file', required=False,
+                        metavar="SECURE_VAR_FILE", default=None,
                         help="path to secure-vars from the root of "
-                        "the secure repo (defaults to ansible/"
-                        "vars/ENVIRONMENT-DEPLOYMENT.yml)")
+                        "the secure repo. By default <deployment>.yml and "
```
feanil, line 54: This seems to imply that if a file is provided here that will be used instead of the two files mentioned but in fact the code uses all 3 files with the secure-vars-file overriding the settings in the other two.  I would make that more obvious here.

PR #1122, opened 2014-05-27
Gems for everyone.

feanil:
Set user_install to `no` so that the gems are installed globally.

### DIFF
--- a/playbooks/roles/jenkins_admin/tasks/main.yml
+++ b/playbooks/roles/jenkins_admin/tasks/main.yml
@@ -133,5 +133,9 @@
     mode=0440 validate='visudo -cf %s'
 
 - name: install global gem dependencies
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_admin/tasks/main.yml
```diff
@@ -133,5 +133,9 @@
     mode=0440 validate='visudo -cf %s'
 
 - name: install global gem dependencies
-  gem: name={{ item.name }} state=present version={{ item.version }}
+  gem: >
```
e0d, line 136: Should we be using the _executable_ option?

feanil, line 136: We're not installing a specific rbenv for the admin jenkins right now.  Since we're using the global ruby,  the default gem executable is probably the best choice.

PR #1123, opened 2014-05-28
Add thin-backup plugin

benpatterson:
Although this can be used as part of any DR strategy, the immediate intent is
providing an ability to quickly spin up Jenkins master instances for testing
against. (A user would run the ansible script to create a test instance,
then restore a backup .zip which would contain all of the jobs' and master's
configuration.)

ThinBackup plugin doc: https://wiki.jenkins-ci.org/display/JENKINS/thinBackup


feanil:
:+1:   FYI, devops uses a custom backup job instead so that it uploads the backup to S3 instead of just keeping it on disk(https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_admin/templates/edx/var/jenkins/jobs/backup-jenkins/config.xml.j2)

PR #1126, opened 2014-05-28
Update edxapp ansible role for third party authorization

dhpollack:
I updated edxapp ansible role to include the ability to enable third party authorization via the server-vars.yml file.  A sample of the server-vars.yml file is as follows:

```
# Third Party Auth Additions
EDXAPP_ENABLE_THIRD_PARTY_AUTH: true
EDXAPP_THIRD_PARTY_AUTH: {
  'Google': {
    "SOCIAL_AUTH_GOOGLE_OAUTH2_KEY": "[GOOGLE-KEY]",
    "SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET": "[GOOGLE-SECRET]"
  }
}
```

Where the key ("Client ID" on Google's site) and secret are your own secret and key from [Google's Developer Console](https://console.developers.google.com/).

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -82,6 +82,7 @@ EDXAPP_FEATURES:
   ENABLE_S3_GRADE_DOWNLOADS: true
   USE_CUSTOM_THEME: $edxapp_use_custom_theme
   AUTOMATIC_AUTH_FOR_TESTING: $EDXAPP_ENABLE_AUTO_AUTH
+  ENABLE_THIRD_PARTY_AUTH: $edxapp_enable_third_party_auth
 
 EDXAPP_BOOK_URL: ''
 # This needs to be set to localhost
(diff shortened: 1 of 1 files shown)


feanil:
looks good to me, @jarv or @e0d can you do the second review?

dhpollack:
@feanil any update on this?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -82,6 +82,7 @@ EDXAPP_FEATURES:
   ENABLE_S3_GRADE_DOWNLOADS: true
   USE_CUSTOM_THEME: $edxapp_use_custom_theme
   AUTOMATIC_AUTH_FOR_TESTING: $EDXAPP_ENABLE_AUTO_AUTH
+  ENABLE_THIRD_PARTY_AUTH: $edxapp_enable_third_party_auth
```
feanil: Can you make `edxapp_enable_third_party_auth` an all caps variable?  Also set a default value of False for it.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -378,6 +379,7 @@ edxapp_generic_auth_config:  &edxapp_generic_auth
   CELERY_BROKER_USER: $EDXAPP_CELERY_USER
   CELERY_BROKER_PASSWORD: $EDXAPP_CELERY_PASSWORD
   GOOGLE_ANALYTICS_ACCOUNT: $EDXAPP_GOOGLE_ANALYTICS_ACCOUNT
+  THIRD_PARTY_AUTH: $edxapp_third_party_auth
```
feanil: Same with `edxapp_third_party_auth` it should be an all caps var and should default to `!!null`

dhpollack:
@feanil should I set the default values for both of these within this file or another file?  Specifically, are you refering to [edx-platform/lms/envs/common.py](https://github.com/edx/edx-platform/blob/5bcd4c54cc3851473e18701d2e01263fdf7e7793/lms/envs/common.py), which sets defaults for these two variables already to False and { } respectively.

---

Edit:  Ok, I see that I have to add it to this file.  But should EDXAPP_THIRD_PARTY_AUTH be { } to remain consistent with the file referenced above or !!null?

feanil:
Tested with our CI ansible run to make sure we don't see any issues.

PR #1127, opened 2014-05-28
updating the commoncluster plays so that they can be run from jenkins

jarv:


util/jenkins/run-ansible.sh
```diff
@@ -0,0 +1,37 @@
...
+
+if [ ! -z "$extra_vars" ]; then
+      ansible_extra_vars+=" -e $extra_vars"
+fi
+
+if [[ $first_in == "true" ]]; then
+    $ansible_limit+="first_in_"
+fi
+
+ansible_limit="tag_Name_${environment}-${deployment}-${play}"
```
feanil: This should be a `+=` so the first_in tag doesn't get lost.

util/jenkins/run-ansible.sh
```diff
@@ -0,0 +1,37 @@
...
+cd $WORKSPACE/configuration/playbooks/edx-east
+
+ansible_extra_vars+=" -e serial_count=$serial_count -e elb_pre_post=$elb_pre_post"
+
+if [ ! -z "$extra_vars" ]; then
+      ansible_extra_vars+=" -e $extra_vars"
+fi
+
+if [[ $first_in == "true" ]]; then
+    $ansible_limit+="first_in_"
```
feanil: What's the purpose of the `$` here?

feanil:
:+1:

PR #1128, opened 2014-05-28
Add vagrant capability for jenkins_master.

benpatterson:
I've used this locally for some Jenkins configuration work/testing.

Would appreciate any suggestions you might have on how to improve it.

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -15,6 +15,12 @@
     - "{{ COMMON_BIN_DIR }}"
     - "{{ COMMON_CFG_DIR }}"
 
+# Determine if machine is provisioned via vagrant
+# Some EC2-specific steps would need to be skipped
+- name: check if instance is vagrant
+  stat: path=/home/vagrant
+  register: vagrant_home_dir
+
 # Need to install python-pycurl to use Ansible's apt_repository module
 - name: Install python-pycurl
   apt: pkg=python-pycurl state=present update_cache=yes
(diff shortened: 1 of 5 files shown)


benpatterson:
@hkim823 this is running on ansible 1.6.1

@feanil I've changed the logic around the motd.d files like we discussed. (Did a push -f though, so our discussion was overwritten).

I had glanced at the jenkins_worker through the same process, and there are ec2-specific pieces in there, so the vagrant.stat could be useful on other files.

Thoughts/feedback? Thanks for the look.

playbooks/roles/common/tasks/main.yml
```diff
@@ -15,6 +15,12 @@
     - "{{ COMMON_BIN_DIR }}"
     - "{{ COMMON_CFG_DIR }}"
 
+# Determine if machine is provisioned via vagrant
+# Some EC2-specific steps would need to be skipped
+- name: check if instance is vagrant
+  stat: path=/home/vagrant
+  register: vagrant_home_dir
+
```
hkim823, line 23: I'm not sure doing it by version of Ubuntu spun up is a good way to do it, but doing a register / stat.exists works, but I think it should be done based on the items specifically listed.

vagrant/base/jenkins-standalone/requirements.txt
```diff
@@ -0,0 +1,4 @@
+# Assumes vagrant is installed globally
+
+ansible==1.6.1
+boto==2.28.0
```
hkim823, line 4: We have a global configuration requirements.txt file on the base level of the repo, which pins the versions of ansible and boto.

feanil:
@benpatterson I've got something like this that I've been using for a while just opened it up as a PR(https://github.com/edx/configuration/pull/1140)  it does the more general case of what's happening here allowing you to run any role on a vagrant image.  What do you think about closing this and incorporating the MOTD changes into that PR?

benpatterson:
Sure @feanil that sounds good to me. I'd personally wanted to eventually have something generic enough to run various different roles, and that takes us there :)

PR #1129, opened 2014-05-28
Feanil/enable sandbox

feanil:


e0d:
LGTM

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -230,24 +230,6 @@
...
-  with_items:
-  - "{{ sandbox_base_requirements }}"
-  - "{{ sandbox_local_requirements }}"
-  - "{{ sandbox_post_requirements }}"
-  sudo_user: "{{ edxapp_user }}"
-  when: "not EDXAPP_PYTHON_SANDBOX and (not inst.stat.exists or new.stat.md5 != inst.stat.md5)"
-  notify:
-  - "restart edxapp"
-  - "restart edxapp_workers"
-
```
carsongee: So we are just nuking the option for not executing in a sandbox?  Will that mess with devstack at all?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -495,13 +496,13 @@ lms_env_config:
   <<: *edxapp_generic_env
   PAID_COURSE_REGISTRATION_CURRENCY: $EDXAPP_PAID_COURSE_REGISTRATION_CURRENCY
   SITE_NAME:  $EDXAPP_LMS_SITE_NAME
-  'CODE_JAIL':
+  CODE_JAIL:
     # from https://github.com/edx/codejail/blob/master/codejail/django_integration.py#L24, '' should be same as None
-    'python_bin': '{% if EDXAPP_PYTHON_SANDBOX %}{{ edxapp_sandbox_venv_dir }}/bin/python{% endif %}'
-    'limits':
-      'VMEM': 0
-      'REALTIME': 5
```
carsongee, line 503: @jbau Is this ok?  You seemed to have added it

playbooks/roles/edxapp/tasks/deploy.yml
feanil: I think this is not a problem as long as the sandbox requirements are correct, it just doesn't add the sandbox requirements to the regular env.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -495,13 +496,13 @@ lms_env_config:
...
-    'user': '{{ edxapp_sandbox_user }}'
+    python_bin: '{% if EDXAPP_PYTHON_SANDBOX %}{{ edxapp_sandbox_venv_dir }}/bin/python{% endif %}'
+    limits:
+      VMEM: 0
```
carsongee: Are we sure we want to allow infinite memory?

playbooks/roles/edxapp/tasks/deploy.yml
carsongee: But if you aren't running with sandbox (EDXAPP_PYTHON_SANDBOX: false), that means the courseware won't have all the packages it could expect if there was a sandbox venv unless all the sandbox requirements are already inside the regular venv.

jbau: I agree with @carsongee,  I added this code specifically to handle the case he described.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -268,6 +268,7 @@ edxapp_workers:
 
 # setup for python codejail
 edxapp_sandbox_venv_dir:  '{{ edxapp_venvs_dir }}/edxapp-sandbox'
+edxapp_sandbox_venv_bin: '{{ edxapp_sandbox_venv_dir }}/bin'
```
jbau: why add this?  I don't see it used anywhere.

playbooks/roles/edxapp/tasks/xml.yml
```diff
@@ -27,6 +27,14 @@
   with_items: EDXAPP_XML_COURSES
   when: item.disposition == "on disk" or item.disposition == "no static import"
 
+- name: make symlinks so code works
```
jbau, line 30: Is this related the sandboxing?

playbooks/roles/edxapp/defaults/main.yml
jbau: It seems hard to figure out what to set memory to, though...

playbooks/roles/edxapp/tasks/deploy.yml
feanil: Good point. Putting this back.

playbooks/roles/edxapp/tasks/xml.yml
feanil, line 30: This is related to importing XML courses.  I can put this on a separate PR but it was a minor change while doing the same work which is why it's here.

playbooks/roles/edxapp/defaults/main.yml
carsongee: I was just thinking something really high but not infinite, like 128MiBs or similar so an instructor can't OOM a machine, but 0 is fine. Maybe it makes more sense to leave these as is here, and turn the params into capital vars so they can be changed easily and we don't have to worry about a universal setting. It might be nice to do this for the generic_env too so we can modify the params in preview/studio.

feanil:
I've pulled out enabling code jail by default as a seperate PR(https://github.com/edx/configuration/pull/1148) and I've pulled out the XML courses related change into its own PR as well(https://github.com/edx/configuration/pull/1147).  I'll rebase this PR once both of those have merged.

feanil:
Closing this. Opened a new PR(https://github.com/edx/configuration/pull/1159) that has the changes needed for matplotlib in the code sandbox.

PR #1130, opened 2014-05-29
Install NLTK for edxapp workers

wedaly:
For ORA2 AI grading, we will need to make the NLTK corpus available to the edxapp workers.  NLTK finds the corpus using the environment variable `NLTK_DATA`, so this PR: (a) downloads the NLTK corpus from S3 (same as the ORA role) and (b) adds the `NLTK_DATA` to the edxapp worker environment.

@feanil I set this up for testing purposes, but if this approach seems reasonable then it would resolve [TKS-1835](https://edx-wiki.atlassian.net/browse/TKTS-1835)

### DIFF
--- a/playbooks/roles/nltk/defaults/main.yml
+++ b/playbooks/roles/nltk/defaults/main.yml
@@ -1,0 +1,13 @@
+---
+NLTK_DATA_DIR: "/usr/local/share/nltk_data"
+
+# Once the file is downloaded, it won't be downloaded again,
+# so if you need to version the data files, you should upload
+# your own version of the files with the version appended to the filename.
+NLTK_DATA:
+  - { path: "taggers/maxent_treebank_pos_tagger",
+      url: "http://nltk.github.com/nltk_data/packages/taggers/maxent_treebank_pos_tagger.zip" }
+  - { path: "corpora/stopwords",
+      url: "http://nltk.github.com/nltk_data/packages/corpora/stopwords.zip" }
+  - { path: "corpora/wordnet",
+      url: "http://nltk.github.com/nltk_data/packages/corpora/wordnet.zip" }
--- a/playbooks/roles/nltk/tasks/main.yml
+++ b/playbooks/roles/nltk/tasks/main.yml
@@ -1,0 +1,25 @@
+---
+
+- name: create the nltk data directory and subdirectories
+  file: path={{ NLTK_DATA_DIR }}/{{ item.path|dirname }} state=directory
+  with_items: NLTK_DATA
+  tags:
+    - deploy
+
+- name: download nltk data
+  get_url: >
+    dest={{ NLTK_DATA_DIR }}/{{ item.url|basename }}
+    url={{ item.url }}
+  with_items: NLTK_DATA
+  register: nltk_download
+  tags:
+    - deploy
+
+- name: unarchive nltk data
+  unarchive: >
+    src={{ NLTK_DATA_DIR }}/{{ item.url|basename }}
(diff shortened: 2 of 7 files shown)


carsongee:
Isn't this ~1GiB? Do we really need another copy? Can we maybe just change old ORA, discern, and this to use a common location for this and set a variable in each role?

wedaly:
@carsongee I installed the NLTK data separately for each role in case the roles are installed on different VMs.  The edxapp data directory is the conventional place to put data used by the LMS and Studio.  I'm hoping that eventually we will deprecate the current ORA role so that NLTK will be installed only once.

carsongee:
I was just thinking if we used something like /edx/var/shared/nltk-data and do that in each of the roles, it should work for all use cases and cut down on the large sandbox size/time to build.

feanil:
If we were going to do something like that I think it would make more sense to move the NLTK related tasks into its own role and have both roles depend on the NLTK role.  @wedaly would you consider doing that?

wedaly:
Yeah, that makes sense.  I'll update the PR.

wedaly:
@feanil I've updated the PR.  The new version adds a `nltk` role that downloads the NLTK corpus from S3 and installs it in `/usr/local/share/nltk_data`.  Since this is one of the default locations that NLTK searches, we don't need to specify any environment variables.

playbooks/roles/nltk/defaults/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+nltk_group: "www-data"
+nltk_data_dir: "/usr/local/share/"
+nltk_tmp_file: "/tmp/nltk.tmp.tar.tz"
+nltk_download_url: "http://edx-static.s3.amazonaws.com/nltk/nltk-data-20131113.tar.gz"
```
feanil: I'm not sure that we have the right to redistribute the NLTK data.  Can we point to the canonical source for this by default and then we can override it for edX purposes?

playbooks/roles/nltk/tasks/main.yml
```diff
@@ -0,0 +1,15 @@
+---
+
+- name: download and install nltk corpus
+  shell: |
```
feanil: Instead of shelling out, I think we should do this as a set of Ansible tasks.  The file, get_url, [unarchive](http://docs.ansible.com/unarchive_module.html) modules should be used so that we can repeatedly run this role but skip tasks that have already been performed.

wedaly:
@feanil Thanks for the comments!  After some investigation, I believe that EASE is actually using a very small subset of the full NLTK data set (about 53 MB, uncompressed).  I've updated the PR to download just those archives directly from the NLTK data site.

I suspect that this will have the positive side-effect of making the Vagrant base boxes smaller in the next release.

wedaly:
Hmm, it appears that `unarchive` is showing up as "changed" when I run the play multiple times.  I tried adding the "creates" flag, but I get an error message about the flag not being supported (looks like it was added in Ansible 1.6).

Is there another way to skip the unarchive step once it's already run?

feanil:
We can add a stats step that sees if the directory exists or use shell with curl which does support the 'creates' parameter.  However, I'm worried that the data URLs don't have versions on them, how often do those data sets get updated?  If we try to not download them if we already have them, we won't be able to update them easily, is that going to be an issue?

wedaly:
@feanil Okay, I think this is ready for another round of review.  I added a comment explaining how to use filenames to version the archives, and I configured the unarchive step to run only if the download registers a change.

feanil:
:+1:

PR #1134, opened 2014-05-30
Vagrantfile dev utilize

dangtrinhnt:
Trying to use the pre-downloaded devstack box before downloading from edx server.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -30,3 +30,4 @@ Fred Smith <derf@edx.org>
 Wang Peifeng <pku9104038@hotmail.com>
 Ray Hooker <ray.hooker@gmail.com>
 David Pollack <david@sologourmand.com>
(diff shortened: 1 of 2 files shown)


hkim823:
I'm not sure why this is necessary, the box if it's downloaded once should be stored in ~/.vagrant.d/ and doing a vagrant box list should show the boxes available.

dangtrinhnt:
It's the matter of Internet connection. I always have to download the box with a download manager before running the vagrant up to fasten up the process. I think this is useful.

feanil:
Vagrant allows you to add the file manually from a URL or local disk. https://docs.vagrantup.com/v2/getting-started/boxes.html

I don't think it makes sense to add logic to the vagrantfile for this.

dangtrinhnt:
Thanks @feanil , just noted.

PR #1136, opened 2014-05-30
Add celery flower role

wedaly:
This PR adds a role to run [Celery Flower](http://flower.readthedocs.org/en/latest/), a monitoring tool for Celery tasks.

Initially, I hope to deploy this on single-instance sandboxes for testing.  I've tested it on my sandbox and verified that it connects successfully to rabbitmq and serves HTTP over port 5555.

@feanil please review.

### DIFF
--- a/playbooks/roles/flower/defaults/main.yml
+++ b/playbooks/roles/flower/defaults/main.yml
@@ -1,0 +1,27 @@
+---
+
+# By default, point to the RabbitMQ broker running locally
+FLOWER_BROKER_USERNAME: "celery"
+FLOWER_BROKER_PASSWORD: "celery"
+FLOWER_BROKER_HOST: "127.0.0.1"
+FLOWER_BROKER_PORT: 5672
+FLOWER_ADDRESS: "0.0.0.0"
+FLOWER_PORT: "5555"
+
+flower_user: "flower"
+flower_app_dir: "{{ COMMON_APP_DIR }}/flower"
+flower_data_dir: "{{ COMMON_DATA_DIR }}/flower"
+flower_log_dir: "{{ COMMON_LOG_DIR }}/flower"
+
+flower_venv_dir: "{{ flower_app_dir }}/venvs/flower"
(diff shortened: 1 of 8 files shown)


playbooks/roles/flower/defaults/main.yml
```diff
@@ -0,0 +1,19 @@
...
+
+flower_venv_dir: "{{ flower_app_dir }}/venvs/flower"
+flower_venv_bin: "{{ flower_venv_dir }}/bin"
+
+flower_python_reqs:
+  - "flower==0.7.0"
+
+flower_deploy_path: "{{ flower_venv_bin }}:/usr/local/sbin:/usr/local/bin:/usr/bin:/sbin:/bin"
+
+flower_broker: "amqp://celery:celery@{{ ansible_default_ipv4.address }}:5672"
```
feanil: Since flower can be deployed on a different machine than the one that rabbit is installed on, I think  you should externalize the various parts of this url to variables with the default being on localhost and the default celery username and password.

The new variables should be in all caps so that it's obvious that they are meant to be overridden.

playbooks/roles/flower/meta/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+dependencies:
+  - supervisor
+  - devpi
+  - rabbitmq
```
feanil: This doesn't need to be a dependency since rabbitmq could be on a different machine than flower.

wedaly:
@feanil Thanks for the feedback!  I've updated the PR to remove the rabbitmq dependency and to make the broker host/port/credentials configurable.

feanil:
This looks great, to make it so that this is automatically part of the single instance sandboxes, I would add it to (edx_continuous_integration.yml)[https://github.com/edx/configuration/blob/master/playbooks/edx-east/edx_continuous_integration.yml].  It won't get pulled into the current single instance sandbox AMI but the next time we do a release of configuration, that will become part of the base AMI.  Other than that, :+1:

PR #1137, opened 2014-05-30
changing source_type

e0d:
The edx source type doesn't work properly with supervisor generated logs.  We should default to syslog until we can do something better here.


feanil:
What about adding a more specific rules for the supervisor logs and set that to be syslog?  That way the standard edx logs can still have the edx sourcetype.

e0d:
It will probably need to be the other way around, I think that everything other than lms/cms is broken at the moment.  I'll take that approach.

feanil:
:+1:

PR #1145, opened 2014-06-02
Add new task to set a boto config file

rodo:
The .boto config file will only be created if the  JENKINS_ADMIN_USE_BOTO_CFG var is defined, by default it is not.


feanil:
looks good. @jarv can you do the second review.

hkim823:
@rodo this is going to require a rebase

feanil:
Jenkins Admin already supports this, the boto file will be empty if the defalut list(`JENKINS_ADMIN_AWS_CREDENTIALS`) of boto configs is not overridden otherwise it will create a boto file that has a profile for each item in the list.

PR #1147, opened 2014-06-02
Link course data to location where edxapp looks for code.

feanil:
The application looks for course code inta a directory named after the
course name so we are symlinking the course repo to that directory as
well.


jarv:
:+1:

PR #1148, opened 2014-06-02
Feanil/enable codejail

feanil:
Enable codejail by default.

The codejail settings are also updated so that tmp files work.

This also allows the codejail process to be able to rm the sandbox users's
temp directory using sudo since it would not have permission otherwise.


e0d:
LGTM

PR #1150, opened 2014-06-02
mongo_mms role

fredsmith:
New Mongo MSS role.  requires setting MMSAPIKEY in secure vars.

### DIFF
--- a/playbooks/roles/mongo_mms/tasks/main.yml
+++ b/playbooks/roles/mongo_mms/tasks/main.yml
@@ -1,0 +1,36 @@
+---
+# mongo_mms
+#
+# Example play:
+#
+#   roles:
+#   - mongo_mms
+
+- fail: MMSAPIKEY is required
+  when: MMSAPIKEY is not defined 
+
+
+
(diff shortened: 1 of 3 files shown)


playbooks/roles/mongo_mms/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+
+- name: install mongo mms agent
+  apt: >
+    deb=/edx/mongodb-mms-monitoring-agent.deb
+
+- name: add key to monitoring-agent.config
+  lineinfile: >
+    dest=/etc/mongodb-mms/monitoring-agent.config
+    regexp="^mmsApiKey="
+    line="mmsApiKey={{ MMSAPIKEY }}"
```
e0d, line 31: need default in play vars.

e0d:
If you use the role generation play:

ansible-playbook -i "localhost," ./ansible_role.yml --connection=local --extra-vars "role_name=newrelic"

You get the boiler plate for free.

Otherwise :+1:

PR #1153, opened 2014-06-02
Use the unzip function since.

feanil:
Unarchive is not the right function since it is for transferring archives from the local machine to the remote one.


jarv:
:+1:

PR #1158, opened 2014-06-03
Remove NLTK_DATA environment var from ORA celery

wedaly:
This is the issue that's breaking the CI build.  In a PR yesterday I moved the NLTK data to /usr/local/share, so there's no need to define the `NLTK_DATA` environment variable.  I removed this from the ORA supervisor config, but forgot to remove it from the worker config.

@feanil


feanil:
:+1:

PR #1160, opened 2014-06-03
Update the ORA2 deploy script to restart workers

wedaly:
This is an update to the ORA2 one-off deployment script.  We've been using this to update our ORA2 sandbox, but it isn't used anywhere else.

Now that we're using Celery workers, we need to restart them once we update ORA2.

@feanil


feanil:
This looks fine, FYI there is a supervisorctl [ansible module](http://docs.ansible.com/supervisorctl_module.html) that you can use as well.

PR #1162, opened 2014-06-03
E0d/xqwatcher fixes

e0d:


playbooks/edx-east/xqwatcher.yml
```diff
@@ -0,0 +1,16 @@
...
+    COMMON_APP_DIR: "/edx/app"
+    common_web_group: "www-data"
+    ENABLE_DATADOG: False
+    ENABLE_SPLUNKFORWARDER: False
+    ENABLE_NEWRELIC: False
+  roles:
+    - aws
+    - datadog
+    - splunkforwarder
+    - newrelic
```
feanil: This should have a conditional otherwise it will always run.(https://github.com/edx/configuration/blob/master/playbooks/edx-east/edxapp.yml#L14-L19)

feanil:
Other than adding the conditionals around datadog, splunk and newrelic this looks good.

Issue #656, opened 2014-01-19
ERROR: the playbook: forums.yml could not be found


ahmadio:
trying to update forums in a single ubuntu installation

`root@ip-10-33-211-161:/edx/bin# sudo ./update forums master
ERROR: the playbook: forums.yml could not be found
`
where should the forums.yml be?

jarv:
thanks for the bug report, if you change `forums.yml` to `forum.yml` it should work. I just fixed this on the master branch.

PR #1163, opened 2014-06-04
fixed update script commands

dhpollack:
The update script has a typo.  The command "ease" runs the ansible script for discern and the "xserver" command is duplicated with the second entry running "forums.yml", which does not exist.  I deleted the second entry instead of changing it to "forum.yml", which does exist, but can be executed with the "cs_comments_service" command.


playbooks/roles/edx_ansible/templates/update.j2
```diff
@@ -45,8 +45,7 @@ repos_to_cmd["edx-platform"]="$edx_ansible_cmd edxapp.yml -e 'edx_platform_versi
 repos_to_cmd["xqueue"]="$edx_ansible_cmd xqueue.yml -e 'xqueue_version=$2'"
 repos_to_cmd["xserver"]="$edx_ansible_cmd xserver.yml -e 'xserver_version=$2'"
 repos_to_cmd["cs_comments_service"]="$edx_ansible_cmd forum.yml -e 'forum_version=$2'"
-repos_to_cmd["xserver"]="$edx_ansible_cmd forums.yml -e 'xserver_version=$2'"
-repos_to_cmd["ease"]="$edx_ansible_cmd discern.yml -e 'discern_ease_version=$2' && $edx_ansible_cmd ora.yml -e 'ora_ease_version=$2'"
```
dhpollack, line 49: "ease" appears to be misnamed and the line above it appears to be either a duplicate of like 46 or a typo of a command for "ease".  I found ease.yml in the ora tasks folders, but this seems to be included in the ora ansible files (line 22 in https://github.com/edx/configuration/blob/master/playbooks/roles/ora/tasks/deploy.yml.

Finally, I found a closed issue (https://github.com/edx/configuration/issues/656) that may pertain to this problem, but since the problem doesn't seem fixed perhaps something went awry.

hkim823:
Thanks for the PR and catching the duplicate that is indeed wrong.  For ease though, I would leave it as is for now unless you feel strongly that it should be.  If that's the case you should change line 15 to address this change.

dhpollack:
@hkim823, I see what you are saying.  I didn't see that it actually ran two ansible commands and that ease is within both of those.  I changed the command back to "ease" and moved the "xserver" command below the "cs_comments_service" command so that the order is the same as in line 15.

Part of my confusion also comes from some misinformation in the edx-platform wiki:

https://github.com/edx/configuration/wiki/edX-Managing-the-Production-Stack#updating-versions-using-edx-repos

The commands listed there should be the same as the commands here, right?

hkim823:
Ideally yes that is true.  I'm going to make sure the script and wiki reflect what should be there and if things go right approve this PR and merge it in.  Thanks again.

PR #1166, opened 2014-06-04
make backup of previous server-vars.yml

dhpollack:
Added some code so the previous server-vars.yml is backed up when /edx/bin/update is run.  Previously only the variables in all caps were preserved.

### DIFF
--- a/playbooks/roles/edx_ansible/tasks/deploy.yml
+++ b/playbooks/roles/edx_ansible/tasks/deploy.yml
@@ -34,6 +34,12 @@
     dest={{ COMMON_CFG_DIR }}/playbooks
     state=link
 
+- name: create backup of server vars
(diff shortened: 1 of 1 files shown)


hkim823:
this makes sense to me @e0d @feanil @jarv or @feanil thoughts?

e0d:
Well the idea was that all caps vars would be the ones changing or subject to customization per environment.  Still, I think that it's wishful thinking that we could predict all var that would need customization, etc.  So I'm fine with this.

@dhpollack Have you signed a contributor agreement?

dhpollack:
@e0d Yes, I've signed the agreement and added myself to the AUTHORS from a previous pull request.

dhpollack:
@e0d anything else that I need to do for this?

PR #1172, opened 2014-06-05
Adding COMMON_SSH_PASSWORD_AUTH var for easy access to flip ssh password based login

hkim823:


playbooks/edx_sandbox.yml
```diff
@@ -14,8 +14,12 @@
     EDXAPP_LMS_NGINX_PORT: '80'
     edx_platform_version: 'master'
     # These should stay false for the public AMI
+
+    # Set COMMON_SSH_PASSWORD_AUTH to Yes if you need to log into your server with a password
```
feanil, line 18: Why would we want this?

hkim823, line 18: There are a lot of people in the Open edx community that use https://github.com/edx/configuration/wiki/edX-Ubuntu-12.04-64-bit-Installation or variations and get locked out of their installation, even after the warning.  I figure this will give them a place to set the variable to their choosing.

feanil, line 18: Is it just that the default was not set before?

hkim823, line 18: The default is set to No, but it's buried in the common role, having it in the play is just easier access to the role to overwrite it for those who want to.

feanil, line 18: I don't think this makes sense to put here if that's the case.  I think the best thing to do is to put a warning in the wiki to tell people the effect this will have and how they can change this setting. The other vars that are in here were originally in here because they existed outside any role, now that they are part of common, I think it makes sense to remove them from here as well.

hkim823:
Fair enough, I'll document changing this in the wiki then.

PR #1174, opened 2014-06-05
E0d/add course requirements

e0d:
Fixes some bugs and adds a feature to write out a per course requirements.txt file based on config and then install it.  Our config is using the wheelhouse.


jarv:
:+1:

PR #1175, opened 2014-06-05
Changing pip packages

hkim823:
Using https://pypi.python.org/pypi/virtualenv as a guide on what versions that and pip should be if setuptools = 3.6 (and vice versa)

### DIFF
--- a/playbooks/roles/common/defaults/main.yml
+++ b/playbooks/roles/common/defaults/main.yml
@@ -71,9 +71,9 @@ common_debian_pkgs:
   - curl
 
 common_pip_pkgs:
-  - pip==1.5.4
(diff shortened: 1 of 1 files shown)


feanil:
What's the reason for the update?  just to stay up to date?

hkim823:
It was part of the other PR that had changes to motd.d scripts but I took those out.  If we're pinning setuptools to 3.6 (they're up to 4.0.1), we might as well pin virtualenv to the version that supports setuptools 3.6 and pip 1.5.6.

feanil:
:+1:  right this is to match the version of pip virtualenv and setuptools
to each other.

PR #1140, opened 2014-05-31
Feanil/run role vagrant file

feanil:

PR #1176, opened 2014-06-06
Skip chown of motd.d files for vagrant machines.

benpatterson:
Follow-on to #1140.

I'm setting the vagrant fact with the intention that it could be used in other .yml files with ec2-specific stuff.


feanil:
:+1:

PR #1184, opened 2014-06-09
Upgrade `requests` module

singingwolfboy:
Pretty minor upgrade: http://docs.python-requests.org/en/latest/api/#migrating-to-2-x

Here's the companion PR for edx-platform: edx/edx-platform#4026

### DIFF
--- a/playbooks/roles/jenkins_worker/defaults/main.yml
+++ b/playbooks/roles/jenkins_worker/defaults/main.yml
@@ -69,7 +69,7 @@ jenkins_wheels:
     - { pkg: "mongoengine==0.7.10", wheel: "mongoengine-0.7.10-py27-none-any.whl" }
     - { pkg: "networkx==1.7", wheel: "networkx-1.7-py27-none-any.whl" }
     - { pkg: "nltk==2.0.4", wheel: "nltk-2.0.4-py27-none-any.whl" }
-    - { pkg: "oauthlib==0.5.1", wheel: "oauthlib-0.5.1-py27-none-any.whl" }
+    - { pkg: "oauthlib==0.6.3", wheel: "oauthlib-0.6.3-py27-none-any.whl" }
     - { pkg: "paramiko==1.9.0", wheel: "paramiko-1.9.0-py27-none-any.whl" }
     - { pkg: "path.py==3.0.1", wheel: "path.py-3.0.1-py27-none-any.whl" }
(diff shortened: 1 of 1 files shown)


hkim823:
What is the main driver to move to this version?

singingwolfboy:
Fixing this issue: https://groups.google.com/d/msg/edx-code/0TWdOF5bmVk/IqSg65iKpjcJ

feanil:
:+1:

jarv:
@singingwolfboy I'm worried about upgrading requests on edx-platform to 2.x because of requests api changes.

The fix for https://groups.google.com/forum/#!msg/edx-code/0TWdOF5bmVk/IqSg65iKpjcJ was just merged into master, https://github.com/edx/edx-platform/pull/4023

singingwolfboy:
@jarv Did you read [the migration guide for 2.x?](http://docs.python-requests.org/en/latest/api/#migrating-to-2-x) I don't think the backward-incompatible changes affect us.

jarv:
The 0.x->1.x was pretty bad it looks like this one will be much safer.
Just wanted to let you know the other issue is now resolved in edx-platform master if that was a driving reason for the change.

singingwolfboy:
@jarv It's good that the issue is resolved. However, we should still stay on the upgrade treadmill if possible, since other issues like this one may end up biting us in the ass in the future. Also, developers who want to work on the edx-platform code may want to use features present in `requests` 2.x that are not present in `requests` 1.x. So I still think we should do this upgrade. (It's not urgent, but it is important.)

feanil:
@jzoldak if the requests module is upgraded, do you need to build wheels for the new package first?

jzoldak:
The pip wheels that are created on the jenkins workers should be in step with the platform. I.e. https://github.com/edx/edx-platform/blob/master/requirements/edx/base.txt

So DO NOT MERGE this until after https://github.com/edx/edx-platform/pull/4026 (and whatever the PR for oauthlib is) make it through.

Every jenkins job for a commit with a different requirement will need to get and install the version of requests and oathlib that _it_ needs, which is at the moment an earlier version than will be in the local wheel.

tusbar:
@jzoldak I think this can be merged now?

singingwolfboy:
I mentioned this issue to @jzoldak yesterday, and he agreed that the `requests` module version in the configuration repo needed to be updated, so I'm interpreting that as a thumbs-up from him for this PR.

PR #1185, opened 2014-06-09
Jarv/migrations

jarv:


feanil:
:+1:  maybe squash it before committing?

PR #1186, opened 2014-06-09
E0d/analytics db

e0d:


jarv:
seems like this could be used to create any db and assign roles, maybe make the name more generic? :+1:

PR #1190, opened 2014-06-09
Updating local ec2 module

carsongee:
I don't recall why we have a local one, but I had use for the current version of the ec2 module.  Feel free to close if this will break whatever the original reason was to have a module different than stock.


jarv:
@carsongee If you want to use the stock one why not just "ec2" in your play instead of "ec2_local"?

The original reason for the fork was the `root_ebs_size` option so we can create sandboxes with 50GB root volumes (mostly for mongo).  Does the new `volumes` option cover us? Not clear if it can be used without having a pre-existing snapshot.

carsongee:
The reason for my update is to grab some options in the module from devel/1.6 (a bug fix to associate_public_ip).  Without some testing, I don't know if the volumes will work for you need, but seems like it should.  I can do some testing on that and get back to you, and we can go from there.

jarv:
For the `ec2_elb` module what I did was call it `ec2_elb_local_1.6.2` if you want to do something similar for grabbing the future version of the ec2 module.

carsongee:
I did just test that this works with ubuntu:

``` yaml
---
- hosts: all
  gather_facts: no
  sudo: no
  tasks:
  - local_action:
      module: ec2
      key_name: mitx-devops
      instance_type: t1.micro
      image: ami-bf582a8f
      group: sshonly
      module: ec2
      wait: yes
      wait_timeout: 500
      region: us-west-2
      volumes:
        - device_name: /dev/sda1
          volume_size: 100
      monitoring: yes
```

So do we maybe want to adapt your plays dependent on this and get rid of it, and then I can add the future one as you suggest?

jarv:
sounds good to me

carsongee:
Ok, switched that all out

jarv:
:+1:

hkim823:
@carsongee did you mean to place the ec2_local file inside of the root level and not playbook/library?

carsongee:
d'oh, that is a big no

PR #1191, opened 2014-06-09
clip stdout_lines

jarv:
This may be something new that wasn't used before in the pip
module. If stdout_lines isn't clipped you will get a payload
error from sqs


feanil:
:+1:

PR #1194, opened 2014-06-10
Patch 1

fmyzjs:
to  use local mongodb repo


playbooks/roles/mongo/defaults/main.yml
```diff
@@ -10,6 +10,9 @@ mongo_data_dir: "{{ COMMON_DATA_DIR }}/mongo"
 mongo_log_dir: "{{ COMMON_LOG_DIR }}/mongo"
 mongo_user: mongodb
 
+mongodb_apt_key: "http://docs.mongodb.org/10gen-gpg-key.asc"
```
feanil: If these are meant to be overridden the variable names should be in all caps.

feanil:
other than variable name update, this looks good.  @fredsmith can you do the second review.

fredsmith:
Change the variables to MONGODB_APT_KEY and MONGODB_REPO and it's good go go.

PR #1195, opened 2014-06-10
Allow restart handler to run when updating conf

tusbar:
Running the `edxapp` role with `-t edxapp_cfg` requires the app to be restarted.

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -418,3 +418,5 @@
   when: EDXAPP_USE_GIT_IDENTITY
 
 - set_fact: edxapp_installed=true
+  tags:
+    - edxapp_cfg


feanil:
@jarv this looks good to me,  are you using the edxapp_cfg tag anywhere where this might cause an issue?

jarv:
I was thinking about removing this fact anyway since its only purpose originally was so that we could run with `--skip-tags deploy`  Unfortunately I think this will break our current migration job since it writes out the config (without restarting the app). Maybe the right thing to do here is delete the fact and we can change the migration job to use a different flag to prevent the handler from triggering.

hkim823:
@jarv what's the situation with this PR?

jarv:
Unfortunately we can't merge this due to the migration jobs and the new config check jenkins jobs.
What we need is a different flag that disables notifications. I've added that to our backlog - https://edx-wiki.atlassian.net/browse/VPC-191

PR #1198, opened 2014-06-10
Jenkins-datadog configuration

benpatterson:
When using datadog on Jenkins, it should include datadog's provided Jenkins capabilities.

### DIFF
--- a/playbooks/roles/jenkins_master/meta/main.yml
+++ b/playbooks/roles/jenkins_master/meta/main.yml
@@ -1,3 +1,5 @@
 ---
 dependencies:
   - common
+  - role: datadog
+    COMMON_ENABLE_DATADOG: True
--- a/playbooks/roles/jenkins_master/tasks/datadog.yml
+++ b/playbooks/roles/jenkins_master/tasks/datadog.yml
@@ -1,0 +1,3 @@
+- name: enable jenkins datadog
+  shell: cp /etc/dd-agent/conf.d/jenkins.yaml.example /etc/dd-agent/conf.d/jenkins.yaml creates=/etc/dd-agent/conf.d/jenkins.yaml
+  notify: restart the datadog service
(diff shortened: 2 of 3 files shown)


benpatterson:
@jzoldak @e0d could you guys take a look?

playbooks/edx-east/jenkins_master.yml
```diff
@@ -8,6 +8,9 @@
   gather_facts: True
   vars:
     COMMON_DATA_DIR: "/mnt"
+    COMMON_ENABLE_DATADOG_JENKINS: True
   roles:
     - common
     - jenkins_master
+    - role: datadog
+      when: COMMON_ENABLE_DATADOG
```
jzoldak: How does this get set to True? I only see the default of False, unless it gets overridden by the ansible provisioning script that is triggered by jenkins (the other jenkins).

jarv: If you always want datadog installed I would just include the role here without the `when:` condition.

playbooks/roles/datadog/tasks/main.yml
```diff
@@ -41,6 +41,10 @@
   tags:
     - datadog
 
+- name: enable jenkins telemetry
```
feanil: Instead of doing this in the datadog role, why not drop this configuration as part of the jenkins_master role?

benpatterson:
@jarv @feanil @jzoldak 

I refactored based on your feedback. Now datadog would always be installed and enabled under the jenkins_master role. I think that simplifies things, too...thx for the suggestion.

I can squash the commits...leaving the history here now to help in your review.

wxt123:
Hello，
       I am studying  edX-platform  ,and I have many questions.Do you have some  technical documentation  for the edx-platform,such as    the framwork  of the platform  .I think  an  a nalysis of the platform  is necessary.

Yours  turly，
Thanks!

jzoldak:
@wxt123 you should start at http://code.edx.org/, with the docs and mailing lists referred to on that page.

feanil:
:+1:

PR #1200, opened 2014-06-11
Have roles reference the nginx role for basic auth

jarv:
This simplifies nginx basic auth by setting a flag to toggle it and changes the roles to use the nginx defaults


e0d:
:+1:

feanil:
Won't this break if the nginx role is not in the same play as the other roles?  Which I assume would be a problem for setups where nginx is on a different server than the application.  @carsongee thoughts?

carsongee:
Yeah, I think I'd prefer we move these vars out to common rather than putting in dependencies where they aren't needed

jarv:
yeah what a mess :/ @carsongee are you ok using default() which is the most recent commit?

carsongee:
I end up overriding these right now so the default isn't necessarily going to help me, but having basic auth variables that are accessible in multiple roles, including nginx would make an improvement for both of our use cases. I know that isn't quite as clean, but it seems like it should be a common pattern for vars that aren't really role dependent.  For example, we could swap out nginx for something else down the road, and we would still probably want to keep basic auth all the same regardless of what app/role is responsible for providing that auth mechanism.

jarv:
@carsongee @feanil @e0d @fredsmith I think I'm convinced that this should move to common. See updates.

carsongee:
Looks good to me

PR #1206, opened 2014-06-11
Update vagrant images to support vmware providers

carsongee:
Adds support for VMWare fusion/workstation providers in vagrant.  I changed the release files even though those images don't exist to show the changes, but if you don't want to make those until the next release I can pull those changes out. I did gen some off of master to verify you could build them from base.


jarv:
:+1:

PR #1211, opened 2014-06-13
allow ansible to be pulled from a private repo

trungpham:


feanil:
@e0d this seems good, but I seem to remember you mentioning a better way to deal with checking out private repositories, can you provide details here or review the changes made by @trungpham

e0d:
Missed this PR, @feanil did you test this when you originally reviewed it?

feanil:
I believe so but I'll do it again just to be sure.

feanil:
Looks like this also needs a rebase as well @trungpham

hkim823:
One thing that this PR doesn't take into account is changing repos all together, say you were working in the edX/platform repo and then wanted to checkout trungpham/edx-platform, from what I can tell these changes will fail because the directory already exists and the git module cannot do a checkout of a fork it doesn't know about.

feanil:
@hkim823 I think that is a known issue in the way we do this in all our roles right now.  This just allows you to provide private sources of the ansible repo in the way we do in some of the other roles.

@trungpham Closing this since we haven't heard back from you in a while.

trungpham:
I couldn't figure out how to reopen this PR after rebasing. So I created a new PR #1697.

PR #1220, opened 2014-06-19
Make the admin files more coherent.

feanil:


playbooks/edx-east/vpc_admin.yml
```diff
@@ -1,20 +1,19 @@
...
-  - aws
-  - edx_ansible
-  - jenkins_admin
-  - hotg
-  - newrelic
+    - aws
+    - edx_ansible
+    - user
+    - jenkins_admin
+    - hotg
```
e0d, line 12: I think it probably makes sense for now to have a separate play for hotg as it needs ~4 different config sets

e0d:
% the comment, :+1:

PR #1221, opened 2014-06-19
Fixing backup server-var.yml

hkim823:


hkim823:
Squashed commits

jarv:
needs rebase

playbooks/roles/edx_ansible/tasks/deploy.yml
```diff
@@ -34,11 +34,16 @@
     dest={{ COMMON_CFG_DIR }}/playbooks
     state=link
 
+- name: check to see if server vars already exists
+  stat: path={{ edx_ansible_var_file }}
+  register: edx_ansible_var_file_exists
+
 - name: create backup of server vars
   copy: >
     src={{ edx_ansible_var_file }}
```
feanil, line 43: This line should be deleted.

playbooks/roles/edx_ansible/tasks/deploy.yml
```diff
@@ -34,11 +34,16 @@
...
     dest={{ edx_ansible_var_file }}.old
```
feanil, line 44: Let's use the 'path' alias instead of the 'dest' alias so that this is more clear.

PR #1222, opened 2014-06-19
Pulling out backup server-var.yml

hkim823:


hkim823:
@feanil

feanil:
Looks good.

hkim823:
@dhpollack I had to take out your PR because it was causing issues, in particular if the file didn't exist , and the copy module is meant for remote files to copy to local servers, and if the file doesn't already exist in the computer running ansible, this will fail.

PR #1226, opened 2014-06-23
Feanil/generic admin

feanil:


playbooks/roles/jenkins_admin/tasks/main.yml
```diff
@@ -33,18 +33,6 @@
 - fail: "JENKINS_ADMIN_S3_PROFILE.secret_key is not defined."
   when: JENKINS_ADMIN_S3_PROFILE.secret_key is not defined
 
-- fail: "JENKINS_ADMIN_CONFIGURATION_REPO is not defined."
```
jarv, line 36: I believe the `fail` module requires a msg parameter

playbooks/roles/jenkins_admin/tasks/main.yml
```diff
@@ -72,6 +60,14 @@
     group={{ jenkins_group }}
     mode=0644
 
+- name: configure the boto profiles for jenkins
+  template: >
+    src="./{{ jenkins_home }}/boto.j2"
```
jarv, line 65: I don't really like this pattern, but I'll live with it I guess.

jarv:
:+1:

PR #1229, opened 2014-06-23
Feanil/optional datadog

feanil:
@jzoldak @benpatterson 

I moved around the datadog dependency because I don't need it for the admin jenkins we use in devops and so I wanted it to be optional.  I also updated the jenkins_master.yml play you're using to indicate that it is being used by testeng.

### DIFF
--- a/playbooks/edx-east/jenkins_testeng.yml
+++ b/playbooks/edx-east/jenkins_testeng.yml
@@ -1,4 +1,4 @@
-# Configure a Jenkins master instance
+# Configure a Jenkins instance for testeng.
 # This has the Jenkins Java app, but none of the requirements
 # to run the tests.
 
--- a/playbooks/edx-east/jenkins_testeng.yml
+++ b/playbooks/edx-east/jenkins_testeng.yml
@@ -12,4 +12,5 @@
   roles:
     - common
     - datadog
+    when: COMMON_ENABLE_DATADOG
     - jenkins_master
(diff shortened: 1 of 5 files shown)


playbooks/edx-east/jenkins_testeng.yml
```diff
@@ -8,6 +8,8 @@
   gather_facts: True
   vars:
     COMMON_DATA_DIR: "/mnt"
+    COMMON_ENABLE_DATADOG: True
   roles:
     - common
+    - datadog
```
jarv: The only place this `COMMON_ENABLE_DATADOG` is used is iat the play level so I think you would want to add a `when: COMMON_ENABLE_DATADOG` condition to the role include

feanil: As a part of this change, I'm now using it within the `jenkins_master/defaults/main.yml`

playbooks/edx-east/jenkins_testeng.yml
```diff
@@ -1,4 +1,4 @@
-# Configure a Jenkins master instance
+# Configure a Jenkins instance for testeng.
```
jzoldak: This is still the testeng master, right? If so, let's keep the word master in there too.

feanil:
@jzoldak any other comments?  good to merge?

jzoldak:
sure, can you pls change the comment in playbooks/edx-east/jenkins_testeng_master.yml to "Configure a Jenkins master instance for testeng", then :thumbsup:

PR #1232, opened 2014-06-24
Jarv/settings from yaml

jarv:


playbooks/roles/analytics-api/tasks/main.yml
```diff
@@ -27,9 +27,9 @@
 #     ENABLE_NEWRELIC: False
 #   roles:
 #     - aws
-#     - analytics_api
+#     - analytics-api
 #
-# ansible-playbook -i 'api.example.com,' ./analytics_api.yml  -e@/ansible/vars/deployment.yml -e@/ansible/vars/env-deployment.yml
+# ansible-playbook -i 'api.example.com,' ./analytics-api.yml  -e@/ansible/vars/deployment.yml -e@/ansible/vars/env-deployment.yml
```
feanil: play name here should be `analyticsapi.yml`

feanil:
Other than nitpick looks good. :+1:

rocha:
@jarv the api uses token authentication. To assign a token to a user, we made a django-admin command:

```
django-admin.py set-api-key username key
```

You can optionally create the user if it doesn't exist using the flag `--create-user`.

Does this need to be reflected in the deployment task?

mulby:
Also we will eventually need that same key to be passed into the LMS, so that the API client in the LMS can access the protected endpoints in the service. Not sure if that affects the deployment strategy at all.

jarv:
It definitely does, especially if we are deciding to manage the API key in
config files on disk rather than managing it in the database.

rocha:
@jarv can you make the required changes?

jarv:
I'm not sure if I will have time to get to it today, can you put a task in
the analytics queue and assign it to me?

rocha:
Done. Thanks!

PR #1234, opened 2014-06-24
add nginx for analytics-api

jarv:


playbooks/roles/analytics-api/defaults/main.yml
```diff
@@ -17,6 +17,7 @@ ANALYTICS_API_GIT_IDENTITY: !!null
 # and a key being provided via NEWRELIC_LICENSE_KEY
 ANALYTICS_API_NEWRELIC_APPNAME: "your Newrelic appname"
 ANALYTICS_API_PIP_EXTRA_ARGS: "-i {{ COMMON_PYPI_MIRROR_URL }}"
+ANALYTICS_API_NGINX_PORT: "18080"
```
feanil: This is the default forum NGINX port.

feanil:
With regards to the nginx rules, we should support ssl for the sandboxes and force redirect to https like we do in the lms/cms.

feanil:
looks good.

PR #1235, opened 2014-06-24
Feanil/abbey profiles

feanil:


util/jenkins/build-ami.sh
```diff
@@ -61,17 +61,9 @@ fi
...
-fi
-
-if [[ -z $configuration_secure ]]; then
-  cd configuration-secure
-  configuration_secure=`git rev-parse HEAD`
-  cd ..
-fi
+cd configuration
+configuration=`git rev-parse --short HEAD`
+cd ..
```
jarv: nit : I like using absolute paths instead of relative paths (use `$WORKSPACE` for jobs run from jenkins)

jarv:
:+1:

PR #1242, opened 2014-06-25
Jarv/migrate changes

jarv:
- removes the "fake" migrations against the fake db since we don't need them anymore (I doubt you guys were using this)
- remove the task to run migrations using the edx application user
- always run migrations using the COMMON_MYSQL_MIGRATE user


playbooks/roles/analytics-api/tasks/deploy.yml
```diff
@@ -37,6 +44,16 @@
   notify: restart the analytics service
   with_items: analytics_api_requirements
 
+- name: syncdb and migrate
+  shell: >
+    chdir={{ analytics_api_code_dir }}
+    DB_MIGRATION_USER={{ COMMON_MYSQL_MIGRATE_USER }}
```
feanil, line 43: Why not pass these in via the ansible environment parameter?

feanil, line 43: Addressed in person, It's not possible to pass both the migration envs and an ANSIBLE var that points to a list of environment settings.

feanil:
looks good.

PR #1244, opened 2014-06-25
Vagrantfiles for Johnnycake

hkim823:
Changes for Johnnycake, adding logic for VBOXFS as an alternative to NFS, making base and release devstack close in line with each other


jarv:
:+1:

PR #1254, opened 2014-06-27
Add and update vars.

feanil:
Add a new overridable var for the ORA2_FILE_PREFIX and
update the default value of the newrelic application names.


jarv:
:+1:

PR #1269, opened 2014-07-02
Some fixes required for mongodb to be deployed

fredsmith:


feanil:
:+1:

carsongee:
Why are we switching to  a string instead of copying a file?

fredsmith:
@carsongee as I understand it, the {{ secure_dir }}/files/ directory isn't used by any other ansible code anymore, and all the other recipes embed keys in a variable in the secure-vars file.  if it's problematic for you, I think I can work logic into the deployment steps to use either the variable or the file.

carsongee:
I guess I missed that decision though I have been feeling the effects of it.  I'll change, no biggie, just makes for uglier var include files with all these big keys in them.

PR #1270, opened 2014-07-02
new plays for roles

fredsmith:


jarv:
should we add optional newrelic and datadog inclusions for these?

fredsmith:
@jarv do you mean adding newrelic/datadog to these plays, or adding new plays for newrelic/datadog?

jarv:
Adding them to the plays, probably more important for newrelic since we use that for server monitoring.
For example we typically add

```
    - role: datadog
      when: COMMON_ENABLE_DATADOG
    - role: splunkforwarder
      when: COMMON_ENABLE_SPLUNKFORWARDER
    - role: newrelic
      when: COMMON_ENABLE_NEWRELIC
```

to the single role plays in the edx-east dir.

PR #1272, opened 2014-07-03
loosen basic auth rules

jarv:


carsongee:
mind adding 172.16.0.0/12 in there too?

jarv:
added

PR #1275, opened 2014-07-03
Adding vpc_subnet_id to allow for t2 instances

hkim823:


playbooks/edx-east/edx_provision.yml
```diff
@@ -4,12 +4,12 @@
...
-    instance_type: m1.medium
-    security_group: sandbox
+    instance_type: t2.medium
+    security_group: sandbox-vpc
     # ubuntu 12.04
-    ami: ami-d0f89fb9
+    ami: ami-f478849c
     region: us-east-1
-    zone: us-east-1b
+    zone: us-east-1c
```
e0d, line 12: What's the purpose of this change?

hkim823, line 12: The subnet I created for sandboxes lives in 1c instead of 1b (just my mistake when I setup the VPC)

jarv:
In general once we have the mongo server in the new vpc as well as the rds we will be connecting sandboxes directly to it so the only task needed in this play is to update server vars to point to the mongo server and rds and re-run config. There is no reason to grab backups or any of the current mongo logic.

hkim823:
I think that makes sense, I'll edit the play to reflect what will happen in the future.

hkim823:
@jarv this is ready for review for merging into master

jarv:
:+1:

PR #1279, opened 2014-07-05
change default of COMMON_ENABLE_BASIC_AUTH to False

dhpollack:
I changed the default behavior of basic authorization to false.  I found the default behavior unintuitive.  Upon installation of the default image via vagrant, going to the server from a remote host (including the vagrant host computer) prompts the user for a username/password.  The default behavior should be that the edx lms/cms "just work" without a username/password, right?  Furthermore, the documentation does not mention that the username/password is edx/edx, and one can only find this information by going into the source files.  Finally, I didn't see a great reason to use basic auth at all, and thus if someone wants to use it, they probably should explicitly enable it, instead of this being the default behavior.

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -131,7 +131,7 @@ EOF
 if [[ $basic_auth == "true" ]]; then
     # vars specific to provisioning added to $extra-vars
     cat << EOF_AUTH >> $extra_vars_file
-COMMON_ENABLE_BASIC_AUTH: True
+COMMON_ENABLE_BASIC_AUTH: False
 COMMON_HTPASSWD_USER: $auth_user
 COMMON_HTPASSWD_PASS: $auth_pass
 XQUEUE_BASIC_AUTH_USER: $auth_user
--- a/playbooks/roles/common/defaults/main.yml
+++ b/playbooks/roles/common/defaults/main.yml
@@ -4,7 +4,7 @@
 # where edX is installed
 
 # Set global htpasswd credentials
-COMMON_ENABLE_BASIC_AUTH: True
+COMMON_ENABLE_BASIC_AUTH: False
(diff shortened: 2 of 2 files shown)


jarv:
Yeah this change came recently when we did a refactor of the vars for basic auth. I'm fine with making basic auth turned off as the default if the rest of the team is ok with it.

jarv:
We should probably merge this into the release branch as well if we take it

dhpollack:
@jarv thanks for the quick reply.  Let me know if I need to anything else.  I have already signed the edx waiver form and added myself to the AUTHORS file for previous pull requests.

util/jenkins/ansible-provision.sh
```diff
@@ -131,7 +131,7 @@ EOF
 if [[ $basic_auth == "true" ]]; then
     # vars specific to provisioning added to $extra-vars
     cat << EOF_AUTH >> $extra_vars_file
-COMMON_ENABLE_BASIC_AUTH: True
+COMMON_ENABLE_BASIC_AUTH: False
```
hkim823, line 134: I don't think this should be set to False, this script is used to provision developer sandboxes, and should be set to True in this scenario.

jarv, line 134: @hkim823 good catch, I only saw the first diff.
@dhpollack can you resubmit this pull to our release branch with only the first change to defaults/main.yml?

dhpollack, line 134: alright, see pull #1286

hkim823:
Closing this PR due to new PR

dhpollack:
@hkim823 got it.  Would it make sense then to remove the same ppas from [this file](https://github.com/edx/edx-platform/blob/master/requirements/system/ubuntu/apt-repos.txt)?  Or perhaps that entire folder, what got me looking into all of this was the existence of these supposed "requirements".

PR #1292, opened 2014-07-08
In the configuration hash values set to "None" will be converted to None

jarv:


playbooks/roles/edxapp/templates/cms.env.json.j2
```diff
@@ -1,5 +1,7 @@
 {% do cms_env_config.update(EDXAPP_ENV_EXTRA) %}
-{% if EDXAPP_UPDATE_STATIC_FILES_KEY %}
```
feanil, line 2: We should stop setting this in abbey.py

feanil:
:+1:  @fredsmith FYI since the default for the cache key was changed to be host specific  I think we don't need to override it using the EDXAPP_UPDATE_STATIC_FILES_KEY flag.

PR #1294, opened 2014-07-09
Migrate all enabled service variants databases

tusbar:
Not just `lms`.
Also, it will work if `lms` is not enabled.


feanil:
@e0d this looks good to me can you please do the second review.

feanil:
@jarv or @hkim823  any objects to merging this?

hkim823:
No objections from me, seems legit

tusbar:
:+1:

PR #1297, opened 2014-07-09
Update jenkins_master package versions, add missing plugins.

benpatterson:
@jzoldak and anyone from devOps ... @feanil or @hkim823 ?


feanil:
:+1:

PR #1300, opened 2014-07-10
allows overriding ssh user

e0d:


jarv:
:+1:

playbooks/edx-east/splunk.yml
```diff
@@ -0,0 +1,14 @@
+- name: Deploy xqueue-watcher
```
tusbar: Probably not what you wanted. :)

PR #1318, opened 2014-07-15
Feanil/xml changes

feanil:
Need symlinks for both the course name and the repo name because the XML is inconsistent about how it refers to the static resources.


jarv:
:+1:

PR #1328, opened 2014-07-22
Add openid workaround to syncdb script.

feanil:


playbooks/edx-east/edxapp_migrate.yml
```diff
@@ -4,8 +4,11 @@
   gather_facts: False
   vars:
     db_dry_run: "--db-dry-run"
+    edxapp_venv_dir: "$VIRTUAL_ENV"
```
jarv, line 7: Is this valid syntax?

feanil:
Closing this since I don't think we need this.  Loadtest should be brought up with a snapshot of sanatized prod data generally.

PR #1336, opened 2014-07-24
Adding the paver completion template to ansible playbook

minhtuev:
adding tasks

adding credit notice

fixing the main task so that we do not overwrite bashrc

adding . to filename to make it hidden


minhtuev:
@hkim823

hkim823:
:+1: i like it

PR #1337, opened 2014-07-24
Revert "Use DraftMongoModuleStore."

feanil:
This reverts commit bd33e9a34ab63047fd56c52a1d5ba09f3267cd45.

This is to maintain backwards compatibility with older versions
of edx-platform.


jarv:
:+1:

jarv:
don't we want to revert this from release, not master?

feanil:
good point will close this and open a PR against release.

jbau:
hmm.  this is one case where maintaining configuration forward / backward compatibility seems tricky.  I wonder if there should be some state support period for edx-platform versions.

feanil:
Yea, we need to figure that out,  right now I'm reverting this so that we can unbreak solutions.

feanil:
Made https://github.com/edx/configuration/pull/1338 to revert this on release.

PR #1338, opened 2014-07-24
Revert "Use DraftMongoModuleStore."

feanil:
This reverts commit bd33e9a34ab63047fd56c52a1d5ba09f3267cd45.

This is to maintain backwards compatibility with older versions
of edx-platform.


fredsmith:
:+1:

PR #1347, opened 2014-07-28
Remove ORA1 role from devstack, in preparation for deprecation

wedaly:
[ORA-696](https://openedx.atlassian.net/browse/ORA-696): Update DevStack to not have ORA1 by default

This PR removes ORA1 from the devstack base image.  It does NOT remove it from the sandbox, vagrant-fullstack, or the release version of devstack (which assumes ORA1 folder sharing).

@feanil Please review.  When the next configuration release goes out, someone will need to update `vagrant/release/devstack/Vagrantfile` as well.

### DIFF
--- a/playbooks/roles/local_dev/tasks/main.yml
+++ b/playbooks/roles/local_dev/tasks/main.yml
@@ -27,6 +27,7 @@
     src=app_bashrc.j2 dest={{ item.home }}/.bashrc
     owner={{ item.user }} mode=755
   with_items: localdev_accounts
+  ignore_errors: yes
 
 # Default to the correct git config
 # No more accidentally force pushing to master! :)
@@ -35,6 +36,7 @@
     src=gitconfig dest={{ item.home }}/.gitconfig
     owner={{ item.user }} mode=700
   with_items: localdev_accounts
+  ignore_errors: yes
 
 # Configure X11 for application users
 - name: preserve DISPLAY for sudo
@@ -59,4 +61,5 @@
   template:
     src=paver_autocomplete dest={{ item.home }}/.paver_autocomplete
     owner={{ item.user }} mode=755
-  with_items: localdev_accounts    
+  with_items: localdev_accounts
+  ignore_errors: yes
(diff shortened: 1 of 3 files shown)


jarv:
@wedaly I'm pretty sure we need to keep ora on devstack until we no longer need to support it, isn't this premature?

stephensanchez:
@jarv I'll be drafting up a deprecation email for ORA1 to edx-code today (and should be sent out tomorrow / wednesday).  After we've stopped supported ORA1 as a code base, I think it'd be reasonable for this change to be in production, so people are not developing against ORA1.

jarv:
@stephensanchez I'm thinking more in terms of what sustaining will need to support assuming ORA will remain in production for a while longer.

wedaly:
As a compromise, would it make sense to keep ORA1 in devstack, but require an environment variable to enable folder sharing?  That way the escalation team could still use ORA1 (by setting the env variable), but it would be less visible to open source contributors.

@adampalay @jarv @stephensanchez What do you think?

adampalay:
Yeah, we still need to dip into ORA1 from time to time. I'm cool with an environment variable.

wedaly:
@jarv @adampalay I've updated the PR to install ORA only when `ENABLE_LEGACY_ORA` is set.

A few things I'd like to get your feedback on:

1) I've updated the base Vagrantfile, but not release.  When is the right time to update release?

2) When the image is built from the base box, ORA1 dependencies won't be installed, which means that the escalation team may suffer longer installation times, but everyone else's installation should speed up.

3) I've added `ignore_errors` to some of the `local_dev` tasks that were failing because `/edx/app/ora` wasn't mounted.  This is the cleanest way I could think of to handle things, but maybe there's some Ansible magic I don't know about?

adampalay:
@wedaly would it be possible to install ORA1 after you build an initial image without it?

wedaly:
@adampalay Yes, it'll just take longer since the requirements (like numpy/scipy/scikit) won't already be installed in the base image.

wedaly:
@jarv When you get a chance, could you please take another look at this?

stephensanchez:
@jarv thanks for feedback thus far, if this is good now we can wrap this up

jarv:
:+1:

PR #1349, opened 2014-07-29
Several fixes to xqwatcher role

carsongee:
- Missing dictionary value of PYTHON_REQUIREMENTS
- Mispelling and boolification of aa-enforce/aa-complain mode
- Changed supervisord port to 9003 tro avoid conflict with devpi role
- Added course list to with_items of aa-enforce task
- Corrected data directory
- Corrected supervisord service name


e0d:
All these changes look good to me :+1:

playbooks/roles/xqwatcher/defaults/main.yml
```diff
@@ -73,7 +75,7 @@ xqwatcher_venv_base: "{{ xqwatcher_home }}/venvs"
 xqwatcher_venv_dir: "{{ xqwatcher_venv_base }}/{{ xqwatcher_service_name }}"
 xqwatcher_code_dir: "{{ xqwatcher_app_dir }}/src"
 xqwatcher_conf_dir: "{{ xqwatcher_home }}"
-xqwatcher_data_dir: "{{ xqwatcher_home }}/data"
+xqwatcher_data_dir: "{{ COMMON_DATA_DIR }}/{{ xqwatcher_service_name }}/data"
```
jarv: originally the idea behind the COMMON_DATA_DIR was a place for compiled static files that are generated during the deploy process as well as a place that the www-data user can write to if the application needs to write to disk. I don't think the xqwatcher uses the xqwatcher_data_dir for these reasons. It is using it to (1) a place to put the grader code  (2) writing out the requirements.txt files for code-jail .  Anything else I am missing? I think for (1) and (2) we should probably put them under `/edx/app/xqwatcher` instead of `/edx/var/xqwatcher` which probably just means keeping what we had originally but renaming the var?

carsongee: I generally think of var as the place to put application data, and to me xqwatcher's data is the grader code, but no strong preference. The pragmatic reason for the change is just that that folder wasn't created by the role, but the `var` variant was by the service role.  It seemed to make more sense to me to just move where the courses ended up, but if you prefer `app` for that I can add the directory creation piece.

jarv: I think since the grade code is critical to the operation of the service it probably belongs in /app.
how about  `xqwatcher_app_data: {{ xqwatcher_app_dir }}/data`?

I believe `xqwatcher_app_dir` is the same as `xqwatcher_home` so we can probably collapse them.

carsongee: I went ahead and made the suggested changes.

jarv:
:+1:

PR #1351, opened 2014-07-29
More RAM for base Vagrantfiles

hkim823:
Helpful for base vagrantboxes to be able to build all the things.


e0d:
hmm, 3072 isn't sufficient?  If not, I'm :+1:

PR #1352, opened 2014-07-29
New Relic : Adding some logic for Amazon Linux

hkim823:
Adding some logic for Amazon Linux so we can install New Relic on those machines based on that AMI


jarv:
:+1:  works for this small task file but we may want to consider creating an `ubuntu.yml` and `amazon-linux.yml` and including them conditionally in the future to make it a little cleaner.

PR #1353, opened 2014-07-30
Fix xqwatcher sudoers

carsongee:
@e0d and @itsbenweeks I created jails for course numbers and those have dots.  It turns out sudoers.d ignores any file with a dot in it and doesn't give a warning of any kind so it wasn't an apparmor problem.  I also removed the sudoers files for the jail users since they don't need a password to execute any command as themselves.  I am also fairly certain we don't want xqwatcher or the codejail users to be able to execute kill and pkill as root.

### DIFF
--- a/playbooks/roles/xqwatcher/tasks/code_jail.yml
+++ b/playbooks/roles/xqwatcher/tasks/code_jail.yml
@@ -18,17 +18,10 @@
     mode=0644 owner=root group=root
   with_items: XQWATCHER_COURSES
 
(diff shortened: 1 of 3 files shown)


e0d:
:+1:

nedbat:
@e0d @carsongee is it clear who will merge this?

e0d:
Carson can merge, it's his PR and he has the bits.

carsongee:
yummy yummy merge bits

PR #1357, opened 2014-07-30
Allow future_builtins map module to memory map in xqwatcher apparmor profile

carsongee:
numpy, for example, uses `map` from future_builtins so adding it to allowed memory map locations.  I was curious if we just wanted to allow all the system python .so libraries to memory map and read to avoid this type of one off change, but I'm ok with one off if we prefer to continue to whitelist.

### DIFF
--- a/playbooks/roles/xqwatcher/templates/etc/apparmor.d/code.jail.j2
+++ b/playbooks/roles/xqwatcher/templates/etc/apparmor.d/code.jail.j2
@@ -21,7 +21,7 @@
     /usr/lib/python2.7/lib-dynload/datetime.so mr,
     /usr/lib/python2.7/lib-dynload/_elementtree.so mr,
     /usr/lib/python2.7/lib-dynload/pyexpat.so mr,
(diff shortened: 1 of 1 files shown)


nedbat:
@carsongee I'm not familiar with these things: what bad behavior is this change fixing? Also, I hadn't heard of future_builtins!

carsongee:
import numpy with numpy==1.8.1 (possibly earlier versions too) uses that function on import `from future_builtins import map` and the apparmor profile doesn't allow it, so changing the profile to allow memory mapping is required to even `import numpy`.  The more meta question seems to be that the profile allows this behavior from anything in the codejail venv, perhaps we should also just allow everything from the system python library as well.

nedbat:
@e0d can you weigh in?

e0d:
I'm fine with the PR as it :+1: Let's continue with the white list approach for now, we can re-visit if it becomes a burden.

carsongee:
Sounds good

PR #1359, opened 2014-08-01
Clytwynec/removing rake install on jenkins

clytwynec:
@jzoldak @benpatterson  Please review...

Is there anyone else that should review this? 

I've tested that devstack is fine with this (but there weren't really any changes to the devstack config).  How would I go about checking this in Jenkins?

This goes with https://github.com/edx/edx-platform/pull/4634

### DIFF
--- a/playbooks/roles/edxapp/tasks/service_variant_config.yml
+++ b/playbooks/roles/edxapp/tasks/service_variant_config.yml
@@ -93,7 +93,7 @@
 
 - name: gather {{ item }} static assets with paver
   shell: >
-    SERVICE_VARIANT={{ item }} paver update_assets {{ system }} --settings=aws
+    SERVICE_VARIANT={{ item }} paver update_assets {{ item }} --settings=aws
     executable=/bin/bash
     chdir={{ edxapp_code_dir }}
   sudo_user: "{{ edxapp_user }}"
--- a/playbooks/roles/edxapp/tasks/service_variant_config.yml
+++ b/playbooks/roles/edxapp/tasks/service_variant_config.yml
@@ -89,11 +89,11 @@
   - "restart edxapp"
   - "restart edxapp_workers"
 
-# Gather assets using rake if possible
+# Gather assets using paver if possible
 
-- name: gather {{ item }} static assets with rake
+- name: gather {{ item }} static assets with paver
   shell: >
-    SERVICE_VARIANT={{ item }} rake {{ item }}:gather_assets:aws
+    SERVICE_VARIANT={{ item }} paver update_assets {{ system }} --settings=aws
     executable=/bin/bash
     chdir={{ edxapp_code_dir }}
   sudo_user: "{{ edxapp_user }}"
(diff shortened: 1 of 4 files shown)


jzoldak:
someone from devops. @hkim823 ?

hkim823:
I'm fine with this, I wonder if https://github.com/edx/configuration/blob/master/playbooks/roles/rbenv/defaults/main.yml should also remove the rbenv_rake_version variable.

clytwynec:
@hkim823  I think that var will  still be needed for the forums rbenv

hkim823:
Makes sense to me :+1:

feanil:
I think we should also test this on a single instance sandbox is possible.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -89,11 +89,11 @@
...
   - "restart edxapp_workers"
 
-# Gather assets using rake if possible
+# Gather assets using paver if possible
 
-- name: gather {{ item }} static assets with rake
+- name: gather {{ item }} static assets with paver
   shell: >
-    SERVICE_VARIANT={{ item }} rake {{ item }}:gather_assets:aws
+    SERVICE_VARIANT={{ item }} paver update_assets {{ system }} --settings=aws
```
feanil: Where does the 'system' var get set?

clytwynec: ahhh. That was a typo, it should be {{ item }} as well.  Will fix.

hkim823: Good point, on a single instance sandbox {{ system }} is undefined.

clytwynec:
How do I go about testing this on a sandbox instance?

hkim823:
This is happy in my single instance sandbox @feanil

hkim823:
@clytwynec i just did a `sudo /edx/bin/update configuration clytwynec/removing_rake_install_on_jenkins && sudo /edx/bin/update edx-platform master` and things were happy with the recent change

Lastly on http://jenkins.edx.org:8080/job/ansible-provision you could probably change the parameters to build with your branch of configuration and see if it completes properly

clytwynec:
Thanks @hkim823 !

feanil:
Looks good to me.

clytwynec:
@jzoldak  @benpatterson  I've tested this via the manual testing process described [here](https://edx-wiki.atlassian.net/wiki/display/ENG/Jenkins+TestEng+Administration).  The tests are passing on the new instance and it has the expected changes in provisioning.   Is it ok to merge this and update the AMI-ID on jenkins? 

After that, I can merge https://github.com/edx/edx-platform/pull/4634.

benpatterson:
I say you're good to go. If there is no time pressure on it, I would wait until tomorrow morning to make the Jenkins Master change since load is generally a bit lower and you can more easily see how the new worker does. 

Regardless of timing, when you do make the change, you can manually provision a jenkins-worker by going to manage nodes and clicking the provision dropdown. Might want to do that so you easily know which node is running off of the new AMI (AWS console will tell you...but just to shortcut the lookup). Plus  you don't have to wait until Jenkins decides to provision a new one.

benpatterson:
:+1: by the way

jzoldak:
:thumbsup:

Along the lines of @benpatterson 's comment - After changing the AMI ID for workers in the jenkins config, the easiest way to get rid of the old nodes gracefully is to go to each of them and use "Mark this node temporarily offline." That way no new builds will run on the old node, and after it is idle for 5 minutes it'll automatically terminate on aws.

PR #1360, opened 2014-08-01
browsermob-proxy role added

clytwynec:
@jzoldak Please review.  Should someone from devops review this as well?

related to https://github.com/edx/bok-choy/pull/53 and https://github.com/edx/edx-platform/pull/4640

### DIFF
--- a/playbooks/roles/browsermob-proxy/defaults/main.yml
+++ b/playbooks/roles/browsermob-proxy/defaults/main.yml
@@ -1,0 +1,3 @@
+# browsermob-proxy
+browsermob_proxy_version: '2.0-beta-9'
+browsermob_proxy_url: 'https://s3-us-west-1.amazonaws.com/lightbody-bmp/browsermob-proxy-{{ browsermob_proxy_version }}-bin.zip'
--- a/playbooks/roles/browsermob-proxy/tasks/main.yml
+++ b/playbooks/roles/browsermob-proxy/tasks/main.yml
@@ -4,27 +4,33 @@
   get_url: >
     url={{ browsermob_proxy_url }}
     dest=/var/tmp/browsermob-proxy-{{ browsermob_proxy_version }}.zip
+  register: download_browsermob_proxy
 
 - name: unzip into /var/tmp/
   shell: >
     unzip /var/tmp/browsermob-proxy-{{ browsermob_proxy_version }}.zip
     chdir=/var/tmp
+  when: download_browsermob_proxy.changed
 
 - name: move to /etc/browsermob-proxy/
   shell: > 
     mv /var/tmp/browsermob-proxy-{{ browsermob_proxy_version }} /etc/browsermob-proxy
+  when: download_browsermob_proxy.changed
 
 - name: change permissions of main script
   file: >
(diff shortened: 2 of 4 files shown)


playbooks/roles/browsermob-proxy/tasks/main.yml
```diff
@@ -0,0 +1,30 @@
...
+---
+- name: get zip file
+  get_url: >
+    url={{ browsermob_proxy_url }}
+    dest=/var/tmp/browsermob-proxy-{{ browsermob_proxy_version }}.zip
+
+- name: unzip into /var/tmp/
+  shell: >
+    unzip /var/tmp/browsermob-proxy-{{ browsermob_proxy_version }}.zip
+    chdir=/var/tmp
```
feanil, line 12: All command below the geturl should be run only when the get URL task changes.

jzoldak:
@clytwynec can you pls test with a fresh devstack also?

playbooks/roles/browsermob-proxy/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+# browsermob-proxy
+browsermob_proxy_version: '2.0-beta-9'
+browsermob_proxy_url: 'https://s3-us-west-1.amazonaws.com/lightbody-bmp/browsermob-proxy-{{ browsermob_proxy_version }}-bin.zip'
```
clytwynec, line 3: Should we be hosting a version of this in our own s3 bucket? @jzoldak @feanil @benpatterson

benpatterson, line 3: FWIW that's what we're doing for JSCover.

benpatterson, line 3: I say goferit. It would protect us from any downtime on the browsermob servers (or wherever the original zip is).

feanil, line 3: Is the current s3 bucket associated with the maintainer of the plugin?  If that's the case I think it may be fine.  Though we may want to host it ourselves for security reasons.

jzoldak, line 3: Yes, I believe that s3 bucket is associated with the [official maintainer of the plugin](https://github.com/lightbody/browsermob-proxy). I think either way would be fine.

clytwynec, line 3: I don't know how to do this myself...  Is there a consensus about whether or not to change it?   I could also merge this and add a ticket to decide later?

jzoldak, line 3: All other things being equal, I'd recommend choosing the path of least resistance.
I.e. use theirs and if we need to change it later we can.
:)

clytwynec:
ok. I re-tested this on devstack with the latest commit.  Ok to merge?

playbooks/roles/browsermob-proxy/defaults/main.yml
benpatterson, line 3: My two cents, I agree we get the change in and get things working, and then consider our own S3 copy as an optimization.

PR #1363, opened 2014-08-04
Include common for jenkins_worker role.

benpatterson:
@hkim823 @jzoldak @clytwynec 

Running the jenkins_worker role via vagrant results in errors because the common_web_user is not set. This takes care of it.

### DIFF
--- a/playbooks/edx-east/jenkins_worker.yml
+++ b/playbooks/edx-east/jenkins_worker.yml
@@ -15,7 +15,6 @@
     - roles/xserver/defaults/main.yml
     - roles/forum/defaults/main.yml
   roles:
-    - common
     - edxlocal
     - mongo
     - browsers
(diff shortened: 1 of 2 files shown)


jzoldak:
:thumbsup: 

Did this break due to some other merge? And if so, how do we prevent this from happening in the future?

hkim823:
:+1:

benpatterson:
@jzoldak @hkim823 

This was missed because the jenkins_worker was never run as a role in & of itself, but always part of the jenkins_worker.yml play. So in part this is more about setting dependencies in the correct place (a role vs a play).

See updated commit.

This update will make the worker role something that can be run on vagrant, assuming the github oauth token is provided in a separate yml file.

clytwynec:
:+1:

jzoldak:
oh, good one @benpatterson I get what you're doing - setting it up to be able to be provisioned and tested in an automated way. I like it. :)

benpatterson:
Bingo @jzoldak !

PR #1364, opened 2014-08-04
Feanil/nat monitor

feanil:


feanil:
@jarv @e0d, Please review.

I'll be squashing this before merging.

playbooks/roles/jenkins_admin/files/nat-monitor.sh
```diff
@@ -0,0 +1,98 @@
...
+#        "ec2:StopInstances"
+#      ],
+#      "Effect": "Allow",
+#      "Resource": "*"
+#    }
+#  ]
+# }
+
+# Get the primary NAT instance's IP
+PRIMARY_NAT_IP=`aws ec2 describe-instances --instance-ids $PRIMARY_NAT_ID | jq -r ".Reservations[].Instances[].PrivateIpAddress"`
```
e0d, line 57: Can't the three api calls for describe-instances be combined?

playbooks/roles/jenkins_admin/files/nat-monitor.sh
```diff
@@ -0,0 +1,98 @@
...
+echo `date` "-- Running NAT monitor"
+while [ . ]; do
+  # Check the health of both instances.
+  primary_pingresult=`ping -c $Num_Pings -W $Ping_Timeout $PRIMARY_NAT_IP| grep time= | wc -l`
+  
+  if [ "$primary_pingresult" == "0" ]; then
+    backup_pingresult=`ping -c $Num_Pings -W $Ping_Timeout $BACKUP_NAT_IP| grep time= | wc -l`
+    if [ "$backup_pingresult" == "0" ]; then
+      echo `date` "-- Both NAT devices un reachable."
+      exit 1
```
e0d: An option would be to do cli sns calls.

playbooks/roles/jenkins_admin/tasks/nat_monitor.yml
```diff
@@ -0,0 +1,45 @@
...
+- fail: msg="NAT_MONITORS is not defined."
+  when: NAT_MONITORS is not defined
+
+- name: upload the monitor script
+  copy:
+    dest="{{ jenkins_admin_scripts_dir }}/nat-monitor.sh"
+    src="nat-monitor.sh"
+    owner="{{ jenkins_user }}"
+    group="{{ jenkins_group }}"
+    mode="755"
```
e0d, line 26: world perms?

e0d:
Generally looks good to me, can you provide sample config to make it go?

playbooks/roles/jenkins_admin/files/nat-monitor.sh
feanil, line 57: I'm not sure I fully understand what you mean.  Above the describe instances is getting the ID of the backup where as the calls here are getting the Private IPs of the instances.  You mean look up each instance once and then grab the IP and ID from same JSON?

e0d, line 57: Yes, I meant could you grab them both from the same JSON.

playbooks/roles/jenkins_admin/files/nat-monitor.sh
jarv: If you want to simply copy what I did for the s3 sync script see the `onerror()` function that is used to send mail via sns on any error https://github.com/edx/configuration/blob/master/playbooks/roles/aws/templates/send-logs-to-s3.j2#L63  you can also add a `trap onerror ERR SIGHUP SIGINT SIGTERM` on any signal or any time a child process of your script has a non-zero exit code.

playbooks/roles/jenkins_admin/files/nat-monitor.sh
feanil, line 57: I tried to do this but I wasn't able to find a good way to pull out the ip by instance_id using jq, which is what I would need to do.

jarv:
:+1:

PR #1366, opened 2014-08-05
Making settings a settable var by convention.

e0d:
This is something I think that should be easily overridden.  Loadtesting environment is one example of why we would do this now.


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -607,9 +610,6 @@ service_variants_enabled:
   - lms
   - cms
 
-edxapp_lms_env: 'lms.envs.aws'
```
feanil, line 653: Can we add a changelog entry for this since it may break people that are already overriding this var.

We may want to also just deprecate this for now and reference the new var from the old var?

feanil:
:+1:  needs rebase

PR #1371, opened 2014-08-07
Feanil/admin plugins

feanil:
@jarv @e0d @fredsmith @hkim823 

Updates to the jenkins role to not re-build plugins if they are already installed.  It also simplifies the admin jenkins role and speeds up re-runs of the jenkins roles in general.

@jzoldak @benpatterson FYI

### DIFF
--- a/playbooks/roles/jenkins_master/tasks/main.yml
+++ b/playbooks/roles/jenkins_master/tasks/main.yml
@@ -77,7 +77,7 @@
   register: jenkins_plugin_downloads
 
 - name: install Jenkins plugins
-  command: cp {{ item.dest}} {{ jenkins_home }}/plugins/{{ item.item.name }}.hpi
+  command: cp {{ item.dest }} {{ jenkins_home }}/plugins/{{ item.item.name }}.hpi
   with_items: jenkins_plugin_downloads.results
   when: item.changed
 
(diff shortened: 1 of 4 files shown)


e0d:
:+1: Although I think there's always some risk when updating that many plugins.

playbooks/roles/jenkins_master/tasks/main.yml
```diff
@@ -67,15 +74,18 @@
...
 - name: install Jenkins plugins
-  command: cp /tmp/{{ item.name }}_{{ item.version }} {{ jenkins_home }}/plugins/{{ item.name }}.hpi
-  with_items: jenkins_plugins
+  command: cp {{ item.dest}} {{ jenkins_home }}/plugins/{{ item.item.name }}.hpi
+  with_items: jenkins_plugin_downloads.results
+  when: item.changed
 
 - name: set Jenkins plugin permissions
-  file: path={{ jenkins_home }}/plugins/{{ item.name }}.hpi
+  file: path={{ jenkins_home }}/plugins/{{ item.item.name }}.hpi
```
jarv, line 85: is the `item.item` correct here?

jarv:
one comment, otherwise :+1:

playbooks/roles/jenkins_master/tasks/main.yml
feanil, line 85: Yes, because we are using the registered results of the download task.  Each item in the download task has its own results which contain the original item they were run with.

feanil:
@benpatterson Ideally yes but right now we always re-install the deb for the core jenkins package because ansible 1.5 doesn't support providing a path to the deb file.  Once we can know for sure that the version of jenkins didn't change we can probably not restart unless we have to.

PR #1374, opened 2014-08-07
Periodically refresh from truth.

feanil:
Periodically refresh what the primary nat is from AWS it changes
out from under the script.


fredsmith:
:+1:

Issue #1382, opened 2014-08-08
Upgrade ansible version to 1.7


rohitkode:
I used the ansible version specified in the requirements file and hit this bug:
https://github.com/ansible/ansible/issues/7413

Basically, the module apt_repository fails behind a proxy. I tried applying the patch to 1.5.5 but that did not fix the problem. 
However, after upgrading to 1.7, the playbook (common.yml in this case) and the task add-apt-repository succeeded.

The ansible version needs to be bumped up to the latest stable version (1.7) in the requirements.txt

hkim823:
We cannot upgrade at this time above Ansible 1.5.x due to $ vars that are inside may plays.  All of those vars need to be refactored and then tested with newer versions of Ansible before we can move forward.

PR #1384, opened 2014-08-08
Bump ansible to latest stable release 1.7.0

rohitkode:
This patch fixes issue/1382 https://github.com/edx/configuration/issues/1382.
Ansible version 1.5.5 is very old now and many important fixes have gone in since then.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-ansible==1.5.5
+ansible==1.7.0
 PyYAML==3.11
 Jinja2==2.7.2
 MarkupSafe==0.23


hkim823:
Thanks for the PR.  Currently we cannot do that because there are $ vars inside of many plays and that needs to happen before we move above Ansible 1.5.x

rohitkode:
Ok. Is there any other way that I can work around the issue explained in https://github.com/edx/configuration/issues/1382 ? (besides executing that step separately)

hkim823:
It doesn't seem like there is to be honest.

PR #1385, opened 2014-08-08
Create ~/.ssh directory when using Git via SSH

stvstnfrd:
Previously, if the `${HOME}/.ssh` directory did not already exist when
attempting to checkout a Git repository via SSH, the
`checkout edx-platform repo` and `checkout theme` tasks would fail.

This issue doesn't appear when checking out code as the `edxapp` user
(default: presumably this directory is created elsewhere), but will
surface if run as another user who lacks a `${HOME}/.ssh` directory.

The code now looks before it leaps.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,3 +34,5 @@ Rodolphe Quiedeville <rodolphe@quiedeville.org>
 Matjaz Gregoric <mtyaka@gmail.com>
 Ben Patterson <bpatterson@edx.org>
 Jason Zhu <fmyzjs@gmail.com>
+Steven Burch <stv@stanford.edu>
+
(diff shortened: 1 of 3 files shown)


feanil:
@stvstnfrd  When do you see this error?  I have not run into it before?

stvstnfrd:
@feanil Yeah, I suspect edx.org doesn't run into this issue, as it appears to work for the default user/config.

```
edxapp_user: CUSTOM_USERNAME_HERE
EDXAPP_USE_GIT_IDENTITY: true
EDXAPP_GIT_IDENTITY: KEY GOES HERE
```

`${HOME}/.ssh` appears to already exist for the `edxapp` user, whereas it doesn't if you're attempting to checkout platform/theme via SSH under a different user account. Since the `accept_hostkey=yes` flag is set on the `git` module calls, it tries to save the hostkey file to your `${HOME}/.ssh` dir, which doesn't exist yet, in this case.

It appears that in the default case, this directory is created by some other task; this change explicitly ensures it exists, if the user is enabling a feature that depends on it.

Hopefully that makes sense.

feanil:
@stvstnfrd Can you elaborate a bit more,  What user are you trying to checkout the platform as?  It shouldn't be that different since the home directory is always set for the user you pass in: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/tasks/main.yml#L7

stvstnfrd:
After discussion, I'm closing this in favor of a better solution; will readdress, if needed.

PR #1387, opened 2014-08-08
adding extra config tasks

jarv:


playbooks/roles/nginx/defaults/main.yml
```diff
@@ -4,7 +4,10 @@
 # and should be overridden
 nginx_sites: []
 nginx_redirects: {}
-
+nginx_extra_sites: []
```
feanil, line 7: Why do we need two sets of variables here?  Why not just use NGINX_EXTRA_CONFIGS/SITES instead?

jarv, line 7: The idea here is that these extra templates and configurations are specific to the edxapp role which is why we name the private var `NGINX_EDXAPP_EXTRA_SITES`, there is another var `nginx_extra_sites` which is the parameter so you can assign it to your secret var. In theory you may have multiple sets of extra configs/sites for different clusters

feanil:
:+1:

PR #1388, opened 2014-08-09
Deploy default theme, by default, not Stanford

stvstnfrd:
We would like the community to adopt the "default" theme, as the
starting point for theme development, as opposed to the "Stanford"
theme, which carries our branding.


stvstnfrd:
@caesar2164 I figured we want this change reflected upstream; feel free to interject.

stvstnfrd:
@jbau and I may have a better idea here, which we'll discuss at Monday's planning and will re-open, if needed.

PR #1393, opened 2014-08-12
path fixes

e0d:
Makes script work, but a bit brittle. @jarv


jarv:
:+1: I like the layout change

feanil:
:+1:

PR #1394, opened 2014-08-12
Clytwynec/preview host

clytwynec:
@jzoldak please review

http://docs.ansible.com/playbooks_variables.html#information-discovered-from-systems-facts
This is where I found the ansible_hostname var...   on devstack it is 'precise64'.

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -85,18 +85,18 @@
 
 - name: update /etc/hosts
   template: src={{ COMMON_HOSTS_TEMPLATE }} dest=/etc/hosts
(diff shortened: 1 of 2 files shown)


playbooks/roles/common/tasks/main.yml
```diff
@@ -85,18 +85,18 @@
 
 - name: update /etc/hosts
   template: src={{ COMMON_HOSTS_TEMPLATE }} dest=/etc/hosts
-  when: COMMON_HOSTNAME
+  when: COMMON_HOSTNAME|length > 0
```
clytwynec, line 88: I added this because I was getting errors for this step without it:

"fatal: [localhost] => error while evaluating conditional: precise64"

PR #1397, opened 2014-08-12
Remove usage of wheels during jenkins worker provisioning.

benpatterson:
@jzoldak @clytwynec 

Since the wheels were only being used during the initial buildout of the AMI, I'm setting that aside and re-using  the edx-venv directory if it exists (or create it if it does not exist).


benpatterson:
@jzoldak @clytwynec can you guys have a look at this PR?

clytwynec:
:+1:  Looks good to me.

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -17,45 +17,28 @@
             owner=root group=root
             mode=755
 
-# Create wheelhouse to enable fast virtualenv creation
-- name: Create wheel virtualenv
-  command: /usr/local/bin/virtualenv {{ jenkins_venv }} creates={{ jenkins_venv }}
+# Create a virtualenv for edx-platform by installing the requirements
+# and packaging the virtualenv.
+- name: Create shallow clone of edx-platform
```
jzoldak, line 25: is the depth 1 specifically to get the contents of the requirements directory? If so can you put that in the comment?

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -17,45 +17,28 @@
...
-# of the virtualenv with only the defined packages
-# from jenkins_wheels.
-- name: Run the wheel_venv.sh script
-  command: >
-    ./wheel_venv.sh edx-venv
-    chdir={{ jenkins_home }}
-    creates={{ jenkins_home }}/edx-venv
+- name: Install edx-platform requirements
+  pip: >
+    requirements={{ jenkins_home }}/shallow-clone/requirements/edx/{{ item }}
```
jzoldak, line 35: after this is completed, does the shallow-clone directory hang around forever? Should it be removed?

playbooks/roles/jenkins_worker/tasks/python.yml
feanil, line 25: Note that if you want the version to ever be anything other than master, the shallow-clone will break.  Since master is hardcoded for now, this will work.

benpatterson, line 25: @jzoldak and I were discussing this morning. And agree...for now this is ok since we'll only pull from master. But once we need/want to grab other branches, this will need some changes.

JZ the depth setting is what's making it a shallow clone...we only ask for minimal ref-log data.

playbooks/roles/jenkins_worker/tasks/python.yml
benpatterson, line 35: It's staying there right now. I had been thinking we can leave it there in case we do end up needing it to not-be shallow. But, considering my previous comment, perhaps we should just clean this up after it's done. I'll update the PR when I get that done.

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak, line 25: @benpatterson oh sorry, I should have RTFM'd!
Though @feanil has a good point. It would be worth capturing that info (works because we're hard coding master) in a comment.

benpatterson:
@jzoldak @feanil @clytwynec here's a new patch set

jzoldak:
:thumbsup:

Issue #1381, opened 2014-08-08
while trying to install script for openedx, getting following error


catftp:
TASK: [mongo | install mongo server and recommends] ***************************
failed: [localhost] => {"failed": true, "item": ""}
stderr: E: There are problems and -y was used without --force-yes

stdout: Reading package lists...
Building dependency tree...
Reading state information...
The following packages will be DOWNGRADED:
  mongodb-10gen
0 upgraded, 0 newly installed, 1 downgraded, 0 to remove and 3 not upgraded.
Need to get 0 B/87.9 MB of archives.
After this operation, 420 kB disk space will be freed.

msg: 'apt-get install 'mongodb-10gen=2.4.7' ' failed: E: There are problems and -y was used without --force-yes

FATAL: all hosts have already failed -- aborting

hkim823:
Looking at https://github.com/edx/configuration/blob/master/playbooks/roles/mongo/tasks/main.yml looks like we need to add force=yes

Can you put in a PR for this?

catftp:
still having same error.

On Fri, Aug 8, 2014 at 7:05 PM, Han Su Kim notifications@github.com wrote:

> Looking at
> https://github.com/edx/configuration/blob/master/playbooks/roles/mongo/tasks/main.yml
> looks like we need to add force=yes
> 
> Can you put in a PR for this?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/edx/configuration/issues/1381#issuecomment-51606267.

## 

Thanks & Regards

Muhammad Ahsan

hkim823:
Here's the PR that should fix the issue https://github.com/edx/configuration/pull/1399

PR #1399, opened 2014-08-12
Adding force option for installing mongo

hkim823:


hkim823:
Based on https://github.com/edx/configuration/issues/1381

feanil:
So this is only needed if we want to force downgrades, is that something we want to do.  Seems like it is good to maintain the version we had in configuration but I just wanted to hear your thoughts on this as well.

hkim823:
I think if someone were to type sudo apt-get upgrade -y on a server for whatever reason then run sudo /edx/bin/update edx-platfor master, this is needed or the script is going to bomb out.

feanil:
:+1:

carsongee:
I know it is a little late, but I think a better way would be to hold the package instead of forcing the downgrade, so adding a task that does `apt-mark hold mongodb-10gen` so update will work normally since going back and forth can be problematic for mongodb and is why we forced a version to start with.

feanil:
I like that as well, @hkim823 thoughts?

PR #1402, opened 2014-08-13
Add tests for jenkins_worker role.

benpatterson:
@jzoldak @clytwynec

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/test.yml
+++ b/playbooks/roles/jenkins_worker/tasks/test.yml
@@ -57,15 +57,6 @@
 # Assert that it was modified at least within the hour
       - "{{ testfile.stat.mtime }} - {{ edxvenv.stat.mtime }} < 3600"
 
-# Verify the rbenv tar is newly-built
-- name: Get info on rbenv tar
-  stat: path={{ jenkins_home }}/edx-rbenv_clean.tar.gz
-  register: rbenv
-- assert:
-    that:
-# Assert that it was modified at least within the hour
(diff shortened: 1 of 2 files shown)


playbooks/roles/jenkins_worker/tasks/test_jenkins_worker.yml
```diff
@@ -0,0 +1,41 @@
+---
+# Tests for this role
+
+
+# Set up #
+- name: Create test file
```
jzoldak: Why do you need to create a test file? Explain in comments.

benpatterson:
@jzoldak I added a bit more description, as well as tests for the github python scripts.

playbooks/roles/jenkins_worker/tasks/test.yml
```diff
@@ -0,0 +1,71 @@
...
+  register: testfile
+
+
+# Tests #
+- name: Verify java cmd is using v 1.7
+  shell: java -version
+  register: java_version
+- assert:
+    that:
+      - "'1.7.0' in java_version.stderr"
```
clytwynec, line 22: This goes to stderr not stdout?

benpatterson, line 22: Believe it or not, yes. The java -version command exits with 0, but this data is fed through stderr.

jzoldak:
:thumbsup:

clytwynec:
:+1:

benpatterson:
@jzoldak @clytwynec Just a heads up I've removed the rbenv test since it'll fail. I added a ticket to fix the underlying issue.

But I'll merge this without that test today.

PR #1285, opened 2014-07-07
added ansible task to add chris-lea ppas for current version nodejs

dhpollack:
I added a task to use chris-lea's repos for nodejs.  In other parts of the program, it seems like these repos are preferred to the default ubuntu repos ([see this](https://github.com/edx/edx-platform/blob/master/requirements/system/ubuntu/apt-repos.txt)).  This will keep the nodejs version current as well.

Separately, I fixed a typo where "npm" was listed twice.  Although "nodejs" does not need to be explicitly listed because it is a dependency of npm, I added it for people that may not be familiar with npm.

PR #1406, opened 2014-08-13
fixing perm issue with npm registry

jarv:


feanil:
:+1:

benpatterson:
@jarv @feanil @hkim823 I don't suppose this was tested on devstack? (or, for that matter, #1285 ?)

I'm having issues provisioning a functional jenkins_worker play...running into the same npm chown issue.

CC @jzoldak

hkim823:
No these were testing to make sure edx-platform can be built basically, which is where we were seeing the errors.  Can you show us the error you are running into?  Try update your configuration branch to master before running the jenkins_worker play

benpatterson:
@hkim823 You can see some info on this over on this PR: https://github.com/edx/edx-platform/pull/4828; however we're considering other options for fixing the issue since that solution is still going to have errors the user sees.

PR #1408, opened 2014-08-14
Update the course ID for the demo course

wedaly:
A few weeks ago, we updated the demo course repository by exporting the course from Studio.  In production, the course ID is `edX/DemoX/Demo_Course`, so that's also the ID of the exported course.

This PR updates the configuration scripts so they install users for the course with the new ID.

@jarv


jarv:
@wedaly Can you submit a pr against the release branch and then merge release into master? :+1:

PR #1410, opened 2014-08-14
Increase how long supervisor waits for workers to stop.

feanil:
This is set as 5 days for now because we have seen "grade_course" tasks that
take about 4 days to complete for large classes and if we shutdown before
they are done we end up in a bad state in the database.


jarv:
:+1:

PR #1411, opened 2014-08-14
Add optimizely project ID to lms config.

feanil:


hkim823:
We should be use curly brackets no?

feanil:
Just trying to stay consistent for now. I think we should just have one PR that makes all those fixes.

hkim823:
:+1: for me, I'll put in the PR for the other changes then

PR #1412, opened 2014-08-14
We have to keep upstart from killing supervisor.

feanil:
Supervisor could wait up to 5 days for workers to finish jobs. So
Upstart needs to wait that long for supervisor to finish.


jarv:
:+1:

PR #1413, opened 2014-08-14
add connection max age to enable database connection pooling.  disabled by default

fredsmith:


feanil:
@fredsmith @chrisndodge  This was only added in Django 1.6

https://docs.djangoproject.com/en/1.6/ref/databases/#persistent-connections

fredsmith:
ah, you're right.  No mention of it on https://docs.djangoproject.com/en/1.4/ref/databases/#persistent-connections

PR #1414, opened 2014-08-14
Ability to create Jenkins worker AMIs via packer.

benpatterson:
@jzoldak @clytwynec 

In progress but wanted to post what I have so far. Please take a look?

### DIFF
--- a/util/packer/jenkins_worker.json
+++ b/util/packer/jenkins_worker.json
@@ -4,7 +4,7 @@
     "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
     "github_oauth_token": "{{env `GITHUB_TOKEN`}}",
     "playbook_remote_dir": "/tmp/packer-edx-playbooks",
-    "ami": "{{env `JENKINS_WORKER_AMI`}}",
+    "ami": "{{env `JENKINS_WORKER_AMI`}}"
   },
   "builders": [{
     "type": "amazon-ebs",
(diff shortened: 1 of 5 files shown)


benpatterson:
@jzoldak @clytwynec 

I've rebased and added the bit where we're testing by running the test_build_server role.

clytwynec:
@benpatterson  Can you describe the actual process that we'd have to go through to get a new ami set up with these changes?

benpatterson:
@clytwynec @jzoldak 

A new AMI would be created as follows:
- via shell (or shell script), 
  *\* set the needed environment variables (AMI id, AWS keys & secrets, github oauth ID)
  *\* cd into the util/packer directory
  *\* run 

```
packer build jenkins_worker.json
```
- If successful, a private AMI will be created with a timestamp
- we would then manually change the ami-id assigned to jenkins workers, and the new ami will begin to be used when the others expire, or we force-down the older workers.

Ultimately the AMI creation would be executed by the public jenkins instance in a job that is not visible to the public. We can re-use the AWS keys that are used for turning on/off workers.

If you'd rather, I can wiki-fy this as well.

cc @feanil @hkim823

util/packer/jenkins_worker.json
```diff
@@ -0,0 +1,46 @@
...
+  }, {
+    "type": "file",
+    "source": "../../../configuration/playbooks",
+    "destination": "{{user `playbook_remote_dir`}}"
+  }, {
+    "type": "file",
+    "source": "../../../configuration/requirements.txt",
+    "destination": "{{user `playbook_remote_dir`}}/requirements.txt"
+  },
+  {
```
clytwynec: I think this has an extra new line between the `,` and  `{`

benpatterson: Quite right. Fixed.

clytwynec:
:+1:

clytwynec:
https://github.com/edx/configuration/pull/1443 Is merged ...

benpatterson:
squashed commits and merging

PR #1417, opened 2014-08-15
Optionally add arguments to service calls.

feanil:
Adding this so we can do a non-blocking call to stop services
which will be useful when we call this as a part of the ansible
run that kicks off Minos.

@e0d @jarv


e0d:
Looks good to me :+1:

PR #1418, opened 2014-08-15
initial working version

e0d:
@feanil @jarv  Inventory script targeting the ASG.  Could be rolled into ec2.py, but I'd rather get our version to the latest and fix it to use profiles before integrating.

Output looks like so:

```
{
  "InService": [
    "10.0.47.127", 
    "10.0.30.137"
  ], 
  "Terminating:Wait": [
    "10.0.11.83"
  ], 
  "e-d-CommonClusterServerAsGroup": [
    "10.0.47.127", 
    "10.0.46.174", 
    "10.0.48.104"
  ], 
  "e-d-CommonClusterServerAsGroup_InService": [
    "10.0.47.127", 
    "10.0.46.174", 
    "10.0.48.104"
  ]
}
```

### DIFF
--- a/playbooks/edx-east/lifecycle_inventory.py
+++ b/playbooks/edx-east/lifecycle_inventory.py
@@ -1,0 +1,58 @@
+#!/usr/bin/env python
+
+import argparse
+import boto
+import json
+
+class LifecycleInventory():
+
+    profile = None
+
+    def __init__(self):
+        parser = argparse.ArgumentParser()
+        group = parser.add_argument_group()
+        group.add_argument('-p', '--profile', help='The aws profile to use when connecting.')
+        args = parser.parse_args()
+
+        self.profile = args.profile
+        self.run()
+
+    def push(self, the_dict, key, element):
+        if key in the_dict:
+            the_dict[key].append(element);
+        else:
+            the_dict[key] = [element]
+
+
+    def get_instance_dict(self):
+        ec2 = boto.connect_ec2(profile_name=self.profile)
+        reservations = ec2.get_all_instances()
+
+        dict = {}
+
+        for instance in [i for r in reservations for i in r.instances]:
+            dict[instance.id] = instance
+
+        return dict
+
+    def run(self):
+        autoscale = boto.connect_autoscale(profile_name=self.profile)
+        groups = autoscale.get_all_groups()
+
+        instances = self.get_instance_dict()
+        inventory = {}
+
+        for group in groups:
+
+            for instance in group.instances:
+
+                private_ip_address = instances[instance.instance_id].private_ip_address
+
+                self.push(inventory, group.name, private_ip_address)
+                self.push(inventory,group.name + "_" + instance.lifecycle_state,private_ip_address)
+                self.push(inventory,instance.lifecycle_state,private_ip_address)
+
+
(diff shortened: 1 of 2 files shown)


playbooks/roles/aws/templates/edx/etc/minos.yml
```diff
@@ -0,0 +1,18 @@
...
+aws_profile: !!null
+s3_bucket: 'edx-prod-edx'
+bucket_path: 'test'
+
+voters:
+  - BellwetherVoter:
+      config:
+  - ProccessQuienscenceVoter:
+      config:
+        process_name: 'gunicorn'
```
feanil, line 12: This is good for edxapp but won't work for celery workers which aren't run under gunicorn.  What's the right thing to do, add a second voter for celery?

playbooks/edx-east/lifecycle_inventory.py
```diff
@@ -0,0 +1,58 @@
...
+import argparse
+import boto
+import json
+
+class LifecycleInventory():
+
+    profile = None
+
+    def __init__(self):
+        parser = argparse.ArgumentParser()
```
feanil, line 41: Seems weird to parse args inside the class, I think it makes more sense to do the arg parsing before the class instantiation and pass in a set of configurations to the class. Which in this case is just the profile name.

playbooks/roles/aws/templates/edx/etc/minos.yml
e0d, line 12: I think the right thing to do is to refactor the config logic to use conf.d style configuration.

playbooks/edx-east/lifecycle_inventory.py
jarv, line 41: The standard python name to do for a script that's run instead of loaded is add a `if __name __ == '__main__'` and do argument parsing in that block (or a function)

playbooks/edx-east/lifecycle_inventory.py
```diff
@@ -0,0 +1,58 @@
...
+        group = parser.add_argument_group()
```
jarv: what is `add_argument_group` used for here?

playbooks/edx-east/lifecycle_inventory.py
```diff
@@ -0,0 +1,58 @@
...
+        group.add_argument('-p', '--profile', help='The aws profile to use when connecting.')
+        args = parser.parse_args()
+
+        self.profile = args.profile
+        self.run()
+
+    def push(self, the_dict, key, element):
```
jarv: this can be replaced with defaultdict - https://docs.python.org/2/library/collections.html#collections.defaultdict

playbooks/edx-east/lifecycle_inventory.py
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env python
+
+import argparse
+import boto
+import json
+
+class LifecycleInventory():
```
jarv, line 36: add a docstring or a usage()?

e0d:
@jarv @feanil comments addressed

feanil:
:+1:

jarv:
:+1:

PR #1419, opened 2014-08-15
jenkins_worker: Add pre-populated .nmprc file to jenkins home dir.

benpatterson:
@jzoldak @hkim823 @clytwynec 

This is the companion piece to https://github.com/edx/edx-platform/pull/4828. The npm profile should be on the worker already, and when the get registry call is made, the set registry call (which would result in errno 50) is unnecessary and therefore avoided.

### DIFF
--- a/playbooks/roles/jenkins_worker/templates/.npmrc.j2
+++ b/playbooks/roles/jenkins_worker/templates/.npmrc.j2
@@ -1,0 +1,1 @@
+registry=http://registry.npmjs.org/
--- a/playbooks/roles/jenkins_worker/templates/.npmrc.j2
+++ b/playbooks/roles/jenkins_worker/templates/.npmrc.j2
@@ -1,1 +1,1 @@
-registry=http://registry.npmjs.org/
+registry={{ COMMON_NPM_MIRROR_URL }}
--- a/playbooks/roles/jenkins_worker/tasks/system.yml
+++ b/playbooks/roles/jenkins_worker/tasks/system.yml
@@ -49,3 +49,11 @@
 - name: add preview.localhost to /etc/hosts
   shell: sed -i -r 's/^127.0.0.1\s+.*$/127.0.0.1 localhost preview.localhost/' /etc/hosts
   sudo: yes
+
+# Npm registry must be pre-loaded or else setting it
+# with the Jenkins user will fail
+# See https://github.com/npm/npm/issues/3565
+- name: Set npm registry
+  template:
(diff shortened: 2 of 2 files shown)


playbooks/roles/jenkins_worker/templates/.npmrc.j2
```diff
@@ -0,0 +1 @@
+registry=http://registry.npmjs.org/
```
feanil: Use the `COMMON_NPM_MIRROR` url instead of hard coding this.

https://github.com/edx/configuration/blob/4ea4e16a03a20e4cd179dce640a158e990af57d7/playbooks/roles/common/defaults/main.yml#L26

benpatterson:
Quite right. Thanks @feanil 

New commit.

playbooks/roles/jenkins_worker/templates/.npmrc.j2
```diff
@@ -0,0 +1 @@
+registry={{ COMMON_NPM_MIRROR_URL }}
```
jzoldak, line 1: the npm config set command leaves spaces around the equals sign, so you probably want to do the same. e.g. registry = https://registry.npmjs.org/

benpatterson, line 1: Ah, in this case, we are actually populating the .npmrc file itself. There it has no spaces.

benpatterson, line 1: @jzoldak ^

jzoldak, line 1: Oh, that was what I meant. It does have the spaces in mine when populated via the npm config set command.

benpatterson, line 1: Oh that's interesting. Mine doesn't have spaces. What distro are you testing on? Perhaps we should discuss...

jzoldak, line 1: No problem. If you test it and it works for you I'm good with no spaces.

benpatterson, line 1: Verified on vagrant (ubuntu precise machine).

```
---> pavelib.prereqs.install_node_prereqs
test `npm config get registry` = "http://registry.npmjs.org/" || (echo setting registry; npm config set registry http://registry.npmjs.org/)
npm install
coffee-script@1.6.3 node_modules/coffee-script
```

PR #1428, opened 2014-08-18
multi host supported

fmyzjs:
make sure mysqlserver can be connected by edxapp servers,ora servers and xqueue servers


hkim823:
Sorry that this fell through the cracks @fmyzjs 

Can you give us an explination of what your code change does, and why you want this to be merged back in?

feanil:
@fmyzjs This has been quiet for a while, closing for now.  It looks like you're trying to use the edxlocal role to setup some production like mysql servers(accessible from anywhere) that is not the purpose of this role, since it is meant to be run on local sandboxes.

PR #1434, opened 2014-08-19
Update CyberSource configuration

wedaly:
Update the CyberSource configuration to use the newer API implementation by default.  This updates the configuration to match the changes in: https://github.com/edx/edx-platform/pull/4880

@feanil Please review.


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -171,6 +171,7 @@ EDXAPP_GRADE_BUCKET: 'edx-grades'
 EDXAPP_GRADE_ROOT_PATH: '/tmp/edx-s3/grades'
 # Credit card processor
 # These are the same defaults set in common.py
+EDXAPP_CC_PROCESSOR_NAME: "CyberSource"
```
hkim823, line 174: Should this be CyberSource2?  Or maybe just set it as "" and we can set it in the secure vars?

wedaly, line 174: I was thinking this would be safer to deploy, since every environment should have "CyberSource" already defined.  When we're ready to cut over, we can make "CyberSource2" the default.

jarv:
:+1:

PR #1435, opened 2014-08-19
replacing colons in the state name

e0d:


playbooks/edx-east/lifecycle_inventory.py
```diff
@@ -52,6 +53,13 @@ def get_instance_dict(self):
 
         return dict
 
+    def to_safe(self, word):
+        ''' Converts 'bad' characters in a string to underscores so they can be
+        used as Ansible groups '''
+
+        return re.sub("[:]", "_", word)
```
jarv: probably better to do a `word = word.replace(":", "_")` instead of using the re module here.

jarv:
once comment(s) are addressed, :+1:

PR #1436, opened 2014-08-19
Let the common web user write to course data.

feanil:
This is beacuse of hardcoded brokenness in how we deal with xml courses
right now.  The course has JS that is compiled when common.py is loaded
and relies on the DATA_DIR(course_data_dir) being in a specific relative
location to the edx-platform code dir.


jarv:
:+1:

PR #1438, opened 2014-08-19
Jenkins Scripts to work with VPC

hkim823:
New script to add to cert whitelist, modification to generate cert to allow for just one user to be generated

### DIFF
--- a/util/jenkins/certificate-whitelist-edge.sh
+++ b/util/jenkins/certificate-whitelist-edge.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+cd configuration
+pip install -r requirements.txt
+env
+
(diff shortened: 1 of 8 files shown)


feanil:
It seems like a lot of these are mostly the same and the differences could be pull out as parameters.  Also, I think we should move to using only manage.py and not use django_admin.py directly as it won't load all the middleware.

hkim823:
I don't think I'm calling django_admin.py anywhere (jsut the name of the file which could be renamed), basically though it was reusing the current Jenkins plays that do work and refactoring them without using Fabric but using just plain ol ssh instead.

feanil:
Ah I see.  I still thing it's worth refactoring this out since there is a lot of repeated code.

hkim823:
I think at this time I just want to merge this into master, and make changes to it over time as needed.

PR #1439, opened 2014-08-19
updates to syncing s3 logs and cron jobs

jarv:
- Create an upstart job that is run after supervisor stops to sync logs
- Syncs trakcing logs before application logs
- stderr and stdout go to /dev/null so we don't fill up root's mailbox

### DIFF
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -84,18 +84,19 @@
     dest={{ COMMON_BIN_DIR }}/{{ aws_s3_sync_script|basename }}
   when: AWS_S3_LOGS
 
-- name: run s3 log sync script on shutdown
-  file: >
-    state=link
-    src={{ COMMON_BIN_DIR }}/send-logs-to-s3
-    path=/etc/rc0.d/S00send-logs-to-s3
+- name: run s3 log sync script on supervisor shutdown
+  template: >
+    src=etc/init/sync-on-stop.conf.j2
+    dest=/etc/init/sync-on-stop.conf
+    owner=root group=root mode=0644
   when: AWS_S3_LOGS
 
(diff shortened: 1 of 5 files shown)


playbooks/roles/aws/tasks/main.yml
```diff
@@ -84,18 +84,19 @@
...
   when: AWS_S3_LOGS
 
-- name: run s3 log sync script on shutdown
-  file: >
-    state=link
-    src={{ COMMON_BIN_DIR }}/send-logs-to-s3
-    path=/etc/rc0.d/S00send-logs-to-s3
+- name: run s3 log sync script on supervisor shutdown
+  template: >
+    src=etc/init/sync-on-stop.conf.j2
```
feanil, line 89: Don't see this template, did you check it in?

jarv, line 89: fixed

playbooks/roles/aws/tasks/main.yml
```diff
@@ -84,18 +84,19 @@
...
-  cron: >
-    name="cronjob for s3 log sync"
-    user=root
-    minute=0
-    job={{ aws_s3_sync_script }}
+  cron:
+    name: "cronjob for s3 log sync"
+    user: root
+    minute: 0
+    job: "{{ aws_s3_sync_script }} > /dev/null 2>&1"
```
feanil, line 101: Maybe log to a file instead?

jarv, line 101: The sync script does a

```
exec > >(tee "{{ aws_s3_logfile }}")
exec 2>&1
```

so everything is already logged.

I made the change to rabbitmq since it's really noisy with a `set -x` and is spamming root's mailbox.

feanil:
:+1: once Travis is done.

PR #1440, opened 2014-08-19
inital working version for review

e0d:
@jarv @feanil

### DIFF
--- a/util/vpc-tools/asg_lifcycle_watcher.py
+++ b/util/vpc-tools/asg_lifcycle_watcher.py
@@ -1,4 +1,4 @@
-__author__ = 'edward'
+__author__ = 'e0d'
 
 """
 Retrieves AWS Auto-scaling lifecycle messages from and SQS queue and processes them.  For
@@ -13,7 +13,7 @@ This script is meant to be run periodically via some process automation, say, Je
 
 It relies on some component applying the proper tags and performing pre-retirement activities.
 
-./sqs.py -q loadtest-edx_autoscaling-lifecycle -b ~/.virtualenvs/aws/bin
+./sqs.py -q autoscaling-lifecycle-queue -b /home/you/.virtualenvs/aws/bin --hook MyLifeCycleHook
 """
 
 import argparse
@@ -30,12 +30,12 @@ class LifecycleHandler:
     NUM_MESSAGES = 10
     WAIT_TIME_SECONDS = 10
 
-    def __init__(self, profile,queue, hook, bin, dry_run):
+    def __init__(self, profile, queue, hook, bin_directory, dry_run):
         logging.basicConfig(level=logging.INFO)
         self.profile = profile
         self.queue = queue
         self.hook = hook
-        self.bin = bin
+        self.bin_directory = bin_directory
         self.dry_run = dry_run
         self.ec2 = boto.connect_ec2(profile_name=self.profile)
 
@@ -46,13 +46,14 @@ class LifecycleHandler:
         # Needed to get unencoded message for ease of processing
         queue.set_message_class(RawMessage)
 
-        for sqs_message in queue.get_messages(self.NUM_MESSAGES,
-                                              wait_time_seconds=self.WAIT_TIME_SECONDS):
+        for sqs_message in queue.get_messages(LifecycleHandler.NUM_MESSAGES,
+                                              wait_time_seconds=LifecycleHandler.WAIT_TIME_SECONDS):
             body = json.loads(sqs_message.get_body_encoded())
             as_message = json.loads(body['Message'])
             logging.info("Proccessing message {message}.".format(message=as_message))
 
-            if 'LifecycleTransition' in as_message and as_message['LifecycleTransition'] == self.INSTANCE_TERMINATION:
+            if 'LifecycleTransition' in as_message and as_message['LifecycleTransition'] \
+                    == LifecycleHandler.INSTANCE_TERMINATION:
                 # Convenience vars, set here to avoid messages that don't meet the criteria in
                 # the if condition above.
                 instance_id = as_message['EC2InstanceId']
@@ -79,12 +80,15 @@ class LifecycleHandler:
                     self.record_lifecycle_action_heartbeat(asg, token,self.hook)
             # These notifications are send when configuring a new lifecycle hook, they can be
             # deleted safely
-            elif as_message['Event'] == self.TEST_NOTIFICATION:
+            elif as_message['Event'] == LifecycleHandler.TEST_NOTIFICATION:
                     if not self.dry_run:
                         logging.info("Deleting message with body {message}".format(message=as_message))
                         sqs_con.delete_message(queue,sqs_message)
                     else:
                         logging.info("Would have deleted message with body {message}".format(message=as_message))
+            else:
+                raise NotImplemented("Encountered message, {message_id}, of unexpected type.".format(
+                    message_id=as_message['MessageId']))
 
 
     def record_lifecycle_action_heartbeat(self, asg, token, hook):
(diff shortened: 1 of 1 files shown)


util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+
+    profile = None
+    queue = None
+    bin = None
+
+    def __init__(self, profile,queue, bin):
+        logging.basicConfig(level=logging.INFO)
+        self.profile = profile
+        self.queue = queue
+        self.bin = bin
```
jarv: if these args are required should an exception be raised here if they aren't passed in?

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+        sqs_con = boto.connect_sqs()
+        q = sqs_con.get_queue(self.queue)
+        q.set_message_class(RawMessage)
+
+        for sqs_message in q.get_messages(10,wait_time_seconds=10):
+            body = json.loads(sqs_message.get_body_encoded())
+            as_message = json.loads(body['Message'])
+            logging.info("Proccessing message {message}.".format(message=as_message))
+
+            if 'LifecycleTransition' in as_message and as_message['LifecycleTransition'] == 'autoscaling:EC2_INSTANCE_TERMINATING':
```
jarv: are these two conditions always true? Specifically will there be a `LifecycleTransistion` string in as_message where the transition won't be `autoscaling:EC2_INSTANCE_TERMINATING`?

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+
+    def __init__(self, profile,queue, bin):
+        logging.basicConfig(level=logging.INFO)
+        self.profile = profile
+        self.queue = queue
+        self.bin = bin
+
+    def process_lifecycle_messages(self):
+        sqs_con = boto.connect_sqs()
+        q = sqs_con.get_queue(self.queue)
```
jarv: nit: avoid single letter variable names

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+        q.set_message_class(RawMessage)
+
+        for sqs_message in q.get_messages(10,wait_time_seconds=10):
```
jarv: why 10? Given the logic below I assume that `get_messages` will leave messages in the queue right?

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+
+        self.run_subprocess_command(command)
+
+    def run_subprocess_command(self,command):
+        logging.info("Running command {command}.".format(command=command))
+
+        try:
+            output = subprocess.check_output(command.split(' '))
+            logging.info("Output was {output}".format(output=output))
+        except Exception, e:
```
jarv: better syntax here is `except Exception as e`

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+            logging.exception(e)
+            logging.error(output)
```
jarv: Is it intentional to log the error for any generic exception and continue here without raising it?

util/vpc-tools/asg_lifcycle_watcher.py
feanil: nit: pull out the value `autoscaling:EC2_INSTANCE_TERMINATING` to an all caps var at the top .

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,146 @@
...
+
+
+    def get_ec2_instance_by_id(self,id):
+        """
+        Simple boto call to get the instance based on the instance-id
+        """
+        ec2 = boto.connect_ec2(profile_name=self.profile)
```
jarv: nit: move this to constructor as self.ec2 so the same connection can be used for every method call

util/vpc-tools/asg_lifcycle_watcher.py
e0d: @jarv no, they're not always true 
- When you create a new hook test messages are sent to your queue
- There are other lifecycles

util/vpc-tools/asg_lifcycle_watcher.py
e0d: Ten is the largest acceptable value.

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,177 @@
...
+import logging
+
+class LifecycleHandler:
+
+    INSTANCE_TERMINATION = 'autoscaling:EC2_INSTANCE_TERMINATING'
+    TEST_NOTIFICATION = 'autoscaling:TEST_NOTIFICATION'
+    NUM_MESSAGES = 10
+    WAIT_TIME_SECONDS = 10
+
+    def __init__(self, profile,queue, hook, bin, dry_run):
```
jarv: nit: space

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,177 @@
...
+        logging.basicConfig(level=logging.INFO)
+        self.profile = profile
+        self.queue = queue
+        self.hook = hook
+        self.bin = bin
+        self.dry_run = dry_run
+        self.ec2 = boto.connect_ec2(profile_name=self.profile)
+
+    def process_lifecycle_messages(self):
+        sqs_con = boto.connect_sqs()
```
jarv, line 43: could also move your connection to the constructor for sqs

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,177 @@
...
+
+                    self.record_lifecycle_action_heartbeat(asg, token,self.hook)
+            # These notifications are send when configuring a new lifecycle hook, they can be
+            # deleted safely
+            elif as_message['Event'] == self.TEST_NOTIFICATION:
+                    if not self.dry_run:
+                        logging.info("Deleting message with body {message}".format(message=as_message))
+                        sqs_con.delete_message(queue,sqs_message)
+                    else:
+                        logging.info("Would have deleted message with body {message}".format(message=as_message))
```
jarv, line 88: should an exception be raised here if there is a message that cannot be identified?

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,177 @@
...
+    def continue_lifecycle(self, asg, token, hook):
+        command = "{path}/python " \
+                  "{path}/aws autoscaling complete-lifecycle-action --lifecycle-hook-name {hook} " \
+                  "--auto-scaling-group-name {asg} --lifecycle-action-token {token} --lifecycle-action-result " \
+                  "CONTINUE".format(
+              path=self.bin, hook=hook, asg=asg, token=token)
+
+        self.run_subprocess_command(command, self.dry_run)
+
+    def run_subprocess_command(self,command, dry_run):
```
jarv: nit: space after comma

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,177 @@
...
+
```
jarv, line 116: since self.dry_run is always passed into this maybe just default it to None and use self.dry_run in the function as the default?

util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -0,0 +1,177 @@
...
+import subprocess
+from boto.sqs.message import RawMessage
+import logging
+
+class LifecycleHandler:
+
+    INSTANCE_TERMINATION = 'autoscaling:EC2_INSTANCE_TERMINATING'
+    TEST_NOTIFICATION = 'autoscaling:TEST_NOTIFICATION'
+    NUM_MESSAGES = 10
+    WAIT_TIME_SECONDS = 10
```
jarv, line 31: nit: when referencing these class constants below it's more explicit to use `LifecycleHandler.WAIT_TIME_SECONDS` instead of `self.WAIT_TIME_SECONDS`.

jarv:
after comments :+1:

PR #1441, opened 2014-08-19
retirement play

e0d:
Play for running against autoscaling retirement candidates.

### DIFF
--- a/playbooks/edx-east/retire_host.yml
+++ b/playbooks/edx-east/retire_host.yml
@@ -1,0 +1,23 @@
+# ansible-playbook -i ./lifecycle_inventory.py ./retire_host.yml 
+# -e@/vars/env.yml --limit Terminating_Wait
+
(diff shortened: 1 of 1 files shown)


jarv:
:+1:

playbooks/edx-east/retire_host.yml
```diff
@@ -0,0 +1,23 @@
+# ansible-playbook -i ./lifecycle_inventory.py ./retire_host.yml 
+# -e@/vars/env.yml --limit Terminating_Wait
+
+- name: Stop all services
+  hosts: Terminating_Wait
+  sudo: True
+  gather_facts: False
```
feanil, line 7: We need to add `STOP_ALL_EDX_SERVICES_EXTRA_ARGS: "--no-wait"` here so that it won't block on the shutdown of supervisor.

PR #1449, opened 2014-08-20
minos role and play updates

e0d:


e0d:
@jarv @feanil @hkim823 

Adding new role for minos and updating the edxapp and worker play

playbooks/edx-east/worker.yml
```diff
@@ -4,6 +4,7 @@
   gather_facts: True
   roles:
     - aws
+    - minos
```
jarv: attach the when condition to the role

jarv:
after comments :+1:

playbooks/roles/minos/templates/edx/etc/minos/conf.d/TrackingLogVoter.yml.j2
```diff
@@ -0,0 +1,6 @@
+TrackingLogVoter:
+  config:
+    aws_profile: !!null
+    local_directory: '{{ COMMON_LOG_DIR }}/tracking'
+    s3_bucket: 'edx-{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}'
+    bucket_path: 'logs/tracking/{{ ansible_ec2_security_groups }}/{{ ansible_ec2_instance_id }}-{{ ansible_ec2_local_ipv4 }}'
```
feanil: This will result in the bucket path having the ID and IP of the abbey box instead of the instance that is running.

e0d:
@feanil these changes go with those to edx-minos to deal with the var expansion timing issues.

feanil:
:+1:

PR #1455, opened 2014-08-21
E0d/logrotate on retire

e0d:


hkim823:
:+1:

playbooks/edx-east/retire_host.yml
```diff
@@ -9,17 +9,31 @@
...
     - stop_all_edx_services
-
-- name: Sync the logs to s3
-  hosts: all
-  sudo: True
   tasks:
-    - name: run log script
+    - name: Force a log rotation
+      command: /usr/sbin/logrotate -f /etc/logrotate.d/{{ item }}
+      with_items:
```
jarv, line 15: any reason why we don't do something like
`shell: find /etc/logrotate.d/ -maxdepth 1 -type f | xargs -n1 /usr/sbin/logrotate -f`?

PR #1456, opened 2014-08-21
fix handler interaction/add pkill

e0d:
- Makes minos a separate play as stop all services users handlers which are run last.
- Defensively kills any existing sync log job

@jarv


feanil:
:+1:

PR #1460, opened 2014-08-22
use COMMON_ENABLE_NEWRELIC

fmyzjs:
use COMMON_ENABLE_NEWRELIC and  NEWRELIC_LICENSE_KEY in forum_env


feanil:
seems good to me.  @e0d can you do the second review?

hkim823:
Makes sense to me :+1:

PR #1462, opened 2014-08-22
fixes nginx access log rotate issue

e0d:


playbooks/edx-east/retire_host.yml
```diff
@@ -34,12 +43,13 @@
...
     - name: Terminate existing s3 log sync
-      command: /usr/bin/pkill send-logs-to-s3
+      command: /usr/bin/pkill send-logs-to-s3 || true
     - name: Send logs to s3
       command: /edx/bin/send-logs-to-s3
 
+      
 - name: Run minos verification
-  hosts: all
+  hosts: Terminating_Wait
```
feanil, line 52: Did you want to revert this change as well?  Or continue to use the limit tag?

e0d, line 52: I think not having this is potentially very dangerous.  There may be cases when we need to target hosts not in that state, but we should have to do something special in that case.

feanil, line 52: Makes sense to me just wanted to make sure that this wasn't getting pulled in accidentally.

feanil:
:+1:

PR #1464, opened 2014-08-22
enable the newrelic app as a separate flag

jarv:


jarv:
This new flag has been added to the *-private repos

feanil:
:+1:

PR #1465, opened 2014-08-22
packer jenkins_worker: use ansible 1.7.1 for testing build server.

benpatterson:
The test_build_server role was failing on acceptance tests because of unicode
static assets being gathered in collectstatic for django. This was because of
this issue https://github.com/ansible/ansible/issues/6655, which has been
resolved in recent ansible versions. We must override the version that is
being used by this repo, and upgrade ansible at provisioning time for
running the test_build_server role.

### DIFF
--- a/playbooks/roles/test_build_server/files/test-development-environment.sh
+++ b/playbooks/roles/test_build_server/files/test-development-environment.sh
@@ -38,8 +38,8 @@ paver test_js_run -s xmodule
 paver test_bokchoy -t test_lms.py:RegistrationTest
 
 # Run some of the lettuce acceptance tests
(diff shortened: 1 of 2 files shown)


benpatterson:
@jzoldak @clytwynec 

This resolves the unicode issue. More details in the commit message :)

jzoldak:
@benpatterson in a future PR, we should add in running a bok-choy test that uses studio and also make sure we choose one that specifically uses one of the [stub servers](https://github.com/edx/edx-platform/tree/master/common/djangoapps/terrain/stubs) or change the lms bok-choy test selected to one.

benpatterson:
Sounds good. New PR for that. 
https://github.com/edx/configuration/pull/1478

PR #1472, opened 2014-08-25
Jarv/retire host fix

jarv:


playbooks/roles/common/defaults/main.yml
```diff
@@ -7,7 +7,10 @@
 COMMON_ENABLE_BASIC_AUTH: False
 COMMON_HTPASSWD_USER: edx
 COMMON_HTPASSWD_PASS: edx
-
+# Turn on syncing logs on rotation for edx
+# application and tracking logs, must also
+# have the AWS role installed
+COMMON_AWS_SYNC: False
```
feanil, line 13: Since the AWS role dependes on common, what do you think about setting this to true in the AWS role meta dependency for common?

jarv, line 13: sounds great as long as the precedence works like you expect it.

playbooks/roles/common/templates/etc/logrotate.d/hourly/edx_logrotate.j2
```diff
@@ -12,5 +12,12 @@
   size 1M
   postrotate
     /usr/bin/killall -HUP rsyslogd
+    {% if COMMON_AWS_SYNC -%}
+    for dir in `ls {{ COMMON_LOG_DIR }}`; do
+        if echo $dir | grep -qv tracking; then
+            {{ aws_s3_sync_script }} -d {{ COMMON_LOG_DIR }}/$dir -b "edx-{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}/logs/application"
```
feanil: The base bucket name should be overridable because we need to for edge currently and may have other reasons to set this explictily in the future.

playbooks/roles/aws/templates/etc/init/sync-on-stop.conf.j2
```diff
@@ -1,5 +1,5 @@
 start on stopped supervisor
-description     "sync s3 logs on supervisor shutdown"
+description     "sync tracking logs on supervisor shutdown"
 script
-        /bin/bash {{ aws_s3_sync_script }} 
+        /usr/sbin/logrotate -f /etc/logrotate.d/hourly/tracking.log
```
feanil, line 4: Since we now sync application logs to s3 as well, do we want to sync those on shutdown as well?

jarv, line 4: added

feanil:
:+1:

PR #1475, opened 2014-08-26
Move environment updates to the deploy section.

feanil:


jarv:
@feanil what's the reason for this?

feanil:
So I can run an alton update with `--tags deploy`.  All of alton's config is in the environment so if we want to add or remove rooms, we have to update the environment.

jarv:
:+1:

PR #1477, opened 2014-08-26
Packer: set config to work with any clone name (host machine).

benpatterson:
@clytwynec @jzoldak 

This will allow us to set up Jenkins jobs to use this process. (The 'configuration' setting was some left over from local development.)


jzoldak:
:thumbsup:

PR #1478, opened 2014-08-26
test_build_server: additional tests.

benpatterson:
@jzoldak @clytwynec @muhammad-ammar 

Will these cover some of the smoke/critical areas for a new build server (in terms of bok-choy)?

### DIFF
--- a/playbooks/roles/test_build_server/files/test-development-environment.sh
+++ b/playbooks/roles/test_build_server/files/test-development-environment.sh
@@ -36,9 +36,9 @@ paver test_js_run -s xmodule
 
 # Run some of the bok-choy tests
 paver test_bokchoy -t test_lms.py:RegistrationTest
-paver test_bokchoy -t test_discussion.py:DiscussionTabSingleThreadTest
-paver test_bokchoy -t test_studio_with_ora_component.py
-paver test_bokchoy -t test_matlab_problem.py
+paver test_bokchoy -t test_discussion.py:DiscussionTabSingleThreadTest --fasttest
+paver test_bokchoy -t test_studio_with_ora_component.py:ORAComponentTest --fasttest
+paver test_bokchoy -t test_matlab_problem.py:MatlabProblemTest --fasttest
 
 # Run some of the lettuce acceptance tests
 paver test_acceptance -s lms --extra_args="lms/djangoapps/courseware/features/problems.feature -s 1"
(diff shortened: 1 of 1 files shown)


jzoldak:
Awesome. :+1:

muhammad-ammar:
Cool. :+1:

playbooks/roles/test_build_server/files/test-development-environment.sh
```diff
@@ -36,6 +36,9 @@ paver test_js_run -s xmodule
 
 # Run some of the bok-choy tests
 paver test_bokchoy -t test_lms.py:RegistrationTest
+paver test_bokchoy -t test_discussion.py:DiscussionTabSingleThreadTest
+paver test_bokchoy -t test_studio_with_ora_component.py
```
clytwynec: What is the likelihood of there being more tests added to this file (or the matlab problem one below)? 

I'm basically wondering if it is worth being more specific to prevent us unknowingly running a lot more tests than intended at some point.

benpatterson:
@clytwynec The intention should be something like a "system smoke test." Basically I'd like to cover things that would come out of a provisioned machine, like having mongo running, making sure jasmine can run, etc. I think making sure stubs are working is a good idea as well. But, to your point, I also would rather avoid it getting too heavy. I added matlab because I thought that included a sort of code jail, but I added it a bit blindly. If you suggest other tests I'm happy to make changes.

To your point, we should avoid making this too cumbersome a process...it should really be testing to be sure the build server has everything it needs for executing builds.

clytwynec:
These seem like good additions.  We should just be conscious that if more tests are added to any of these files, then we may be running them unknowingly and the testing part of the ami creation will get slower.

jzoldak:
@clytwynec I was thinking the same thing, then wondered in the other direction - what if the tests get refactored and either the testcase class or the test itself gets renamed?

I think you're right though, all considered it is probably preferable to specify the test case class and a single testcase name.

benpatterson:
@jzoldak @clytwynec @muhammad-ammar New commit, getting more specific and also reducing some overhead.

jzoldak:
oh, good one with the --fasttest!
:+1: :thumbsup:

clytwynec:
Ah yea,  definitely good call on --fasttest.

:+1:

PR #1479, opened 2014-08-26
New defaults for social network footer

hkim823:
Look at https://github.com/edx/edx-platform/pull/4855 for more info


stephensanchez:
:+1:

feanil:
These don't get referenced by the env files, this won't work.  The 'EDXAPP_' files need to be referenced within the env file to work correctly.  But it looks like edx/edx-platform#4855 doesn't update aws.py so this won't have any effect anyway.

PR #1480, opened 2014-08-26
Add registration extra fields setting to configuration.

wedaly:
This exposes a Django setting to configure the registration form.  The values in this PR are identical to the [current defaults](https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L1425-L1441)

Eventually, we'll need to override the defaults to set `country` to "optional" for [ECOM-128](https://openedx.atlassian.net/browse/ECOM-128).

@jarv please review.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -288,6 +288,17 @@ EDXAPP_SUBDOMAIN_BRANDING: {}
 EDXAPP_WORKERS: !!null
 EDXAPP_ANALYTICS_DATA_TOKEN: ""
 EDXAPP_ANALYTICS_DATA_URL: ""
+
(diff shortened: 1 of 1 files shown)


jarv:
Did we investigate putting these settings into the configuration model instead of settings?

jarv:
:+1:

wedaly:
The configuration was added back in January by Xavier -- I don't know if they discussed using a configuration model instead of settings.  I'm expecting that the ecommerce team will rework the registration dialog down the road, and we can certainly revisit that decision.

jbau:
just curious, what do you guys mean by configuration model?  do you mean django model fields, or is this something else?

jarv:
@jbau we would like to migrate most of the settings into the database, see https://github.com/edx/edx-platform/blob/master/common/djangoapps/config_models/models.py

jbau:
ah ok, thx.

PR #1481, opened 2014-08-26
Verbose output for ansible-provision jenkins job

hkim823:


feanil:
Can we have this as a flag?  It will make the output pretty un-readable.

hkim823:
I'm seeing that the output is pretty un-readable yeah.

PR #1490, opened 2014-08-27
Add var for extra gunicorn options

carsongee:
IN case people need to add installation specific gunicorn options


e0d:
I'm :+1:  General question, is there any value in making your extra args the defaults?

jarv:
what's this for? Does it make sense to use it for max requests instead of setting it separately?

carsongee:
The specific use case I'm looking to meet is to specify `--forwarded-allow-ips` since I need that for my architecture, but the thought occurred to me that it made sense to give people flexible access to their gunicorn conf.  Another option I considered was migrating our gunicorn switches into config files instead since our gunicorn switch count is getting pretty high, but that may be too much work for little gain.  I think something like `--max-requests` makes sense to go in directly and not in this `EXTRA` var since memory leak protections is probably a general good thing, whereas my specific config is going to change for me for each environment I use it in, and I'm probably the only one not using a localhost nginx to proxy gunicorn.

jarv:
:+1:

PR #1492, opened 2014-08-28
allow auth_extra override for CMS/LMS independently

fredsmith:


feanil:
Removing `EDXAPP_AUTH_EXTRA` will break people who are currently using it.  Maybe leave it in and override it with `EDXAPP_AUTH_EXTRA` first and then LMS/CMS specific ones?  Also, add a change log entry for this.

fredsmith:
@feanil -  I fall back and use EDXAPP_AUTH_EXTRA if EDXAPP_CMS_AUTH_EXTRA or EDXAPP_LMS_AUTH_EXTRA aren't defined.  I could also make them both default to EDXAPP_AUTH_EXTRA in defaults.yml, if you think that's the better way to do it

feanil:
:+1:  once travis is happy.

PR #1494, opened 2014-08-28
Making profile explicit

e0d:
This changes the AWS shell outs to use an explicit profile.  I think the indirection of the env var is confusing and it also seem to not work.

```
15:33:14 A client error (ValidationError) occurred when calling the CompleteLifecycleAction operation: No Lifecycle Hook found with name 'e-d-GetTrackingLogs' for group 'e-d-EdxappServerASGroup-Q9BAAVZLU6K7-v011'
```

This hook does exist, but is not being found because the profile isn't being used for CLI commands.

### DIFF
--- a/util/vpc-tools/asg_lifcycle_watcher.py
+++ b/util/vpc-tools/asg_lifcycle_watcher.py
@@ -37,7 +37,7 @@ class LifecycleHandler:
         logging.basicConfig(level=logging.INFO)
         self.queue = queue
         self.hook = hook
-        os.environ['AWS_PROFILE'] = profile
+        self.profile = profile
         if bin_directory:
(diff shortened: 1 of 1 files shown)


util/vpc-tools/asg_lifcycle_watcher.py
```diff
@@ -97,27 +97,26 @@ def process_lifecycle_messages(self):
                 raise NotImplemented("Encountered message, {message_id}, of unexpected type.".format(
                     message_id=as_message['MessageId']))
 
+    def get_base_cli_command(self):
```
jarv: instead of a function can this be set in the constructor as `self.aws_command`?

jarv:
after comment (leave it up to you) :+1:

PR #1495, opened 2014-08-28
Release...rc/kifli

hkim823:
Merging in changes from rc/kifli to the release branch of configureation


.travis.yml
```diff
@@ -3,7 +3,7 @@ language: python
 python:
   - "2.7"
 install:
-  - "sudo apt-get install -y npm python-demjson"
+  - "sudo apt-get install -y nodejs python-demjson"
```
feanil, line 6: This should probably have a changelog entry as it may break people who don't uninstall the old npm before installing the new nodejs package right?

wedaly:
@hkim823 A while back, I made some changes to make ORA1 optional in devstack.  I made the change to the base Vagrantfile, but not to the release version (because I wanted to coordinate the change with the actual release).

I'm not in the office for the first part of this week, but could you please make sure release/devstack/Vagrantfile gets updated (same changes as in #1347).  Otherwise, folks using the release Vagrantfile will have to wait a long time on startup while everything for ORA1 gets installed.

hkim823:
@wedaly thanks for checking on that, those changes are a part of the kifli release

jarv:
:+1:

PR #1506, opened 2014-08-29
cleanup notifier default variables

fredsmith:


playbooks/edx-east/notifier.yml
```diff
@@ -0,0 +1,9 @@
+- name: Configure notifier instance
+  hosts: all
+  sudo: True
+  gather_facts: True
+  roles:
+    - role: aws
+      when: COMMON_AWS_SYNC
```
e0d: We need to figure out what to do here. Including the AWS role shouldn't hinge on enabling s3 log sync.

jarv: For now I think the right thing to do is to include aws unconditionally since that is what we do for other plays like edxapp https://github.com/edx/configuration/blob/master/playbooks/edx-east/edxapp.yml .

PR #1513, opened 2014-09-03
Add rubygems dependency to rbenv role

kluo:
@jarv


kluo:
Turns out this isn't a bug, Rubygems is installed with ruby. I just had a incomplete ruby installation

PR #1524, opened 2014-09-10
updating jenkins_worker to include browsermob-proxy

clytwynec:
TE-471

@benpatterson  Please review.  I've created a new AMI and confirmed this change in it.

### DIFF
--- a/playbooks/edx-east/jenkins_worker.yml
+++ b/playbooks/edx-east/jenkins_worker.yml
@@ -18,4 +18,5 @@
     - edxlocal
     - mongo
     - browsers
+    - browsermob-proxy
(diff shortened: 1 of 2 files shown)


benpatterson:
Think you could add a test in test_build_server role? Remember that sh script will run as the jenkins user so arguably things should be in place...

clytwynec:
@benpatterson  any suggestions for how to test this automatically?  Testing that it works in context depends on the edx-platform repo having things in place, and just confirming that it exists seems excessive.

benpatterson:
BTW as noted offline before the merge, the change is :+1: for me.

PR #1527, opened 2014-09-11
Jarv/migration check for sandbox clone

jarv:


playbooks/edx-east/edxapp_migrate.yml
```diff
@@ -16,3 +17,14 @@
       with_items:
         - lms
         - cms
+    - name: syncdb
+      shell: >
+        chdir={{ edxapp_code_dir }}
+        python manage.py {{ item }} syncdb --noinput --settings=aws_migrate
```
e0d, line 23: should we do syncdb --migrate to catch everything here?

e0d:
Maybe overkill, but should we check django version?

:+1: after one code comment.

playbooks/edx-east/edxapp_migrate.yml
jarv, line 23: I think we are covered since the migrate happens immediately before this task

PR #1529, opened 2014-09-11
Adding aws vars for common role to work

hkim823:
The AWS vars need to be added because the common/templates/etc/logrotate.d/hourly/edx_logrotate_tracking.j2 calls the variable {{ aws_s3_sync_script }}


jarv:
This has come up in other areas, I vote we change this var to `{{ common_aws_s3_sync_script }}`
and maybe do something like this in common/defaults/main.yml:

```
common_aws_s3_sync_script: {{ aws_s3_sync_script|default('/bin/true') }}

```

feanil:
I would agree with that as well, this way, the common role doesn't have to depend on the aws role.

playbooks/roles/common/defaults/main.yml
```diff
@@ -12,6 +12,7 @@ COMMON_HTPASSWD_PASS: edx
 # have the AWS role installed
 COMMON_AWS_SYNC: False
 COMMON_AWS_SYNC_BUCKET: "edx-{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}"
+COMMON_AWS_S3_SYNC_SCRIPT: "{{ COMMON_APP_DIR }}/aws/send-logs-to-s3"
```
jarv: If we do hardcode the path I think we should probably use `{{ COMMON_BIN_DIR }}/send-logs-to-s3` and ensure that a symlink is created from there to the location in the aws role, how does that sound?

hkim823:
If we want to make the break between the roles, I think we just hard code it into common then.

playbooks/roles/common/defaults/main.yml
hkim823: The symlink is being created in the aws role itself when it's run.  

```
- name: create symlink for s3 log sync script
  file: >
    state=link
    src={{ aws_s3_sync_script }}
    dest={{ COMMON_BIN_DIR }}/{{ aws_s3_sync_script|basename }}
  when: AWS_S3_LOGS
```

jarv:
:+1:

PR #1530, opened 2014-09-11
Removing ora from script - DO NOT MERGE

hkim823:


jarv:
I think sustaining wanted to keep ora installed for when they need to troubleshoot it, what's the motivation for the change?

hkim823:
yeah sorry this wasn't supposed to be a PR, I just wanted a branch to test CI since ORA is failing since https://github.com/edx/configuration/commit/6a6871c30ac76c49813ef437db38fe0f77a1ffe3 there are different versions of libraries that's conflicting for matplotlib to work in code jail properly

PR #1531, opened 2014-09-11
makes an all-caps var for celery workers

jarv:


feanil:
Can you add a change log entry for this as well? :+1: once that is done.

PR #1538, opened 2014-09-15
Added vagrant-aspen playbook and vagrant conf

singingwolfboy:
@jarv, can you review this? I'm in the process of testing, to be sure that this works the way I expect. (I mostly copied the vagrant file from `vagrant-fullstack.yml`.)

### DIFF
--- a/vagrant/base/fullstack/Vagrantfile
+++ b/vagrant/base/fullstack/Vagrantfile
@@ -39,7 +39,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :ansible do |ansible|
     # point Vagrant at the location of your playbook you want to run
     ansible.playbook = "../../../playbooks/vagrant-fullstack.yml"
-    ansible.verbose = "vvvv"
+    ansible.verbose = "vvv"
     # set extra-vars here instead of in the vagrant play so that
     # they are written out to /edx/etc/server-vars.yml which can
     # be used later when running ansible locally
--- a/vagrant/base/fullstack/Vagrantfile
+++ b/vagrant/base/fullstack/Vagrantfile
@@ -39,9 +39,12 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :ansible do |ansible|
     # point Vagrant at the location of your playbook you want to run
     ansible.playbook = "../../../playbooks/vagrant-fullstack.yml"
+    ansible.verbose = "vvvv"
     # set extra-vars here instead of in the vagrant play so that
     # they are written out to /edx/etc/server-vars.yml which can
     # be used later when running ansible locally
-    ansible.verbose = "vvvv"
+    if ENV['OPENEDX_RELEASE']
+      ansible.extra_vars = { OPENEDX_RELEASE: ENV['OPENEDX_RELEASE']}
+    end
   end
 end
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -23,7 +23,12 @@ cd /edx/app/edx_ansible/edx_ansible/playbooks
 # this can cause problems (e.g. looking for templates that no longer exist).
 /edx/bin/update configuration release
 
-ansible-playbook -i localhost, -c local vagrant-devstack.yml --tags=deploy -e configuration_version=release
+# Did we specify an openedx release?
+if [ -n "$1" ]; then
+  EXTRA_VARS="-e OPENEDX_RELEASE=$1"
+fi
+
+ansible-playbook -i localhost, -c local vagrant-devstack.yml --tags=deploy -e configuration_version=release $EXTRA_VARS
 SCRIPT
 
 edx_platform_mount_dir = "edx-platform"
@@ -105,5 +110,5 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Assume that the base box has the edx_ansible role installed
   # We can then tell the Vagrant instance to update itself.
-  config.vm.provision "shell", inline: $script
+  config.vm.provision "shell", inline: $script, args: ENV['OPENEDX_RELEASE']
 end
--- a/playbooks/roles/forum/defaults/main.yml
+++ b/playbooks/roles/forum/defaults/main.yml
@@ -29,7 +29,7 @@ FORUM_ELASTICSEARCH_URL: "http://{{ FORUM_ELASTICSEARCH_HOST }}:{{ FORUM_ELASTIC
 
 # This needs to be a string, set to 'false' to disable
 FORUM_NEW_RELIC_ENABLE: '{{ COMMON_ENABLE_NEWRELIC }}'
-FORUM_NEW_RELIC_LICENSE_KEY: "{{ NEWRELIC_LICENSE_KEY }}"
+FORUM_NEW_RELIC_LICENSE_KEY: '{{ NEWRELIC_LICENSE_KEY | default("") }}'
 FORUM_NEW_RELIC_APP_NAME: "{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-forum"
 
 FORUM_WORKER_PROCESSES: "4"
@@ -72,4 +72,3 @@ forum_version: "master"
 #
 forum_services:
   - {service: "elasticsearch", host: "{{ FORUM_ELASTICSEARCH_HOST }}", port: "{{ FORUM_ELASTICSEARCH_PORT }}"}
-
--- a/playbooks/vagrant-devstack.yml
+++ b/playbooks/vagrant-devstack.yml
@@ -7,12 +7,17 @@
     openid_workaround: true
     devstack: true
     disable_edx_services: true
-    edx_platform_version: 'master'
     mongo_enable_journal: false
     EDXAPP_NO_PREREQ_INSTALL: 0
     COMMON_MOTD_TEMPLATE: 'devstack_motd.tail.j2'
     COMMON_SSH_PASSWORD_AUTH: "yes"
     ENABLE_LEGACY_ORA: !!null
+    # used for releases
+    edx_platform_version: '{{ OPENEDX_RELEASE | default("master") }}'
+    ora2_version: '{{ OPENEDX_RELEASE | default("master") }}'
+    certs_version: '{{ OPENEDX_RELEASE | default("master") }}'
+    forum_version: '{{ OPENEDX_RELEASE | default("master") }}'
+    xqueue_version: '{{ OPENEDX_RELEASE | default("master") }}'
   vars_files:
     - "group_vars/all"
   roles:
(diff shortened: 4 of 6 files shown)


benpatterson:
I realize I haven't been privvy to any discussion here, but if someone wanted a specific tagged release, wouldn't  they just...

git checkout <tag>
then go to the appropriate directory and vagrant up?

(then, git checkout <tag> on whatever repo, like platform, they care about?)

playbooks/vagrant-aspen.yml
```diff
@@ -0,0 +1,47 @@
...
+    EDXAPP_LMS_NGINX_PORT: '80'
+    EDX_ANSIBLE_DUMP_VARS: true
+    CERTS_DOWNLOAD_URL: 'http://192.168.33.10:18090'
+    CERTS_VERIFY_URL: 'http://192.168.33.10:18090'
+    edx_platform_version: 'openedx/rc/aspen-2014-09-10'
+    ora2_version: 'openedx/rc/aspen-2014-09-10'
+    certs_version: 'openedx/rc/aspen-2014-09-10'
+    forum_version: 'openedx/rc/aspen-2014-09-10'
+    xqueue_version: 'openedx/rc/aspen-2014-09-10'
+    XQWATCHER_VERSION: 'openedx/rc/aspen-2014-09-10'
```
jarv: my understanding is that xq-watcher is unable to function without also including grader code which is not open source

singingwolfboy:
@benpatterson we're trying to release _all_ of our repositories at a specific version. This simply helps us get everything up and running more easily, without having to go around to each one and checkout the proper version.

playbooks/vagrant-aspen.yml
```diff
@@ -0,0 +1,47 @@
+- name: Configure instance(s)
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    migrate_db: 'yes'
+    openid_workaround: true
+    EDXAPP_LMS_NGINX_PORT: '80'
+    EDX_ANSIBLE_DUMP_VARS: true
```
jarv: To see all of the version vars in the configuration repo try this:
`grep _version: playbooks/roles/*/defaults/main.yml`

there isn't an `ora2_version` because its version is specified in edx-platform's requirements file, so it will be at whatever version is specified at edx-platform's commit

playbooks/vagrant-aspen.yml
singingwolfboy: How do I make ansible not install xq-watcher, then? It isn't listed under "roles", either here or in `vagrant-fullstack.yml`, but gets installed anyway in fullstack.

vagrant/release/aspen/ansible.cfg
```diff
@@ -0,0 +1,9 @@
+# config file for ansible -- http://ansible.github.com
+# nearly all parameters can be overridden in ansible-playbook or with command line flags
+# ansible will read ~/.ansible.cfg or /etc/ansible/ansible.cfg, whichever it finds first
+
+[defaults]
+
+jinja2_extensions=jinja2.ext.do
+host_key_checking = False
+roles_path=../../ansible-roles/roles:../../ansible-private/roles:../../ansible-roles/
```
feanil: It doesn't do any harm but we don't need to include the ansible-private path for this configuration.

feanil:
Rather than having a vagrant file for Aspen, we should just have one for openedx which defaults to the Aspen RC for now but can be overridden later if necessary.  You can have the default value of all the versions be a single new Ansible var.  `OPEN_EDX_VERSION` perhaps. And then set that from the vagrant file using the [extra_vars](https://github.com/edx/configuration/blob/master/vagrant/base/test_role/Vagrantfile#L36-L38) construct for vagrant ansible.

playbooks/vagrant-aspen.yml
```diff
@@ -0,0 +1,47 @@
...
+    - demo
+    - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
+    - oraclejdk
+    - elasticsearch
+    - forum
+    - { role: "xqueue", update_users: True }
+    - certs
+    - role: analytics-api
+      when: ANALYTICS_API_GIT_IDENTITY
+    - edx_ansible
```
hkim823: Not sure if edx_ansible needs to be repeated here

playbooks/vagrant-aspen.yml
jarv: I missed the part where it isn't included as a role.
Since it isn't included as a role there is no reason to set a version here.

vagrant/base/fullstack/Vagrantfile
```diff
@@ -39,9 +39,12 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :ansible do |ansible|
     # point Vagrant at the location of your playbook you want to run
     ansible.playbook = "../../../playbooks/vagrant-fullstack.yml"
+    ansible.verbose = "vvvv"
```
feanil: This is great for debugging but should be turned off by default.  Ah I see that this was here from before.  @hkim823 can you think of any good reason for the very verbose output?

hkim823: I agree, I went with a full 4 v to diagnose something, but I think it can go down to 1 v and if need to move to more verbose output later.

feanil:
@singingwolfboy that looks good for the base files, but the corresponding changes to the release vagrant files still need to be made.

vagrant/base/fullstack/Vagrantfile
```diff
@@ -39,9 +39,12 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :ansible do |ansible|
     # point Vagrant at the location of your playbook you want to run
     ansible.playbook = "../../../playbooks/vagrant-fullstack.yml"
+    ansible.verbose = "v"
```
feanil: I talked to @hkim823 and the reason for this to be super verbose was to make it easier to debug when building the base boxes for a release.  I think `vvvv` was too much but `vvv` is probably good for the base full stack file.

singingwolfboy:
@feanil I've made the changes for the release/devstack Vagrantfile, but I can't set the box for the release/fullstack Vagrantfile until the aspen release is totally ready to go.

vagrant/release/devstack/Vagrantfile
```diff
@@ -105,5 +110,5 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Assume that the base box has the edx_ansible role installed
   # We can then tell the Vagrant instance to update itself.
-  config.vm.provision "shell", inline: $script
+  config.vm.provision "shell", inline: $script, s.args: ENV['OPENEDX_RELEASE']
```
feanil: Were you able to verify that this works as expected, with and without the environment variable being set?

singingwolfboy: Just gave it a try, and caught a bug. Thanks for pointing that out. Should be fixed in this PR now.

feanil:
@singingwolfboy what about the fullstack release vagrant file?

singingwolfboy:
@feanil as I commented above, I can't set the box for the release/fullstack Vagrantfile until the aspen release is totally ready to go.

feanil:
I think this PR is ready to go other than that change then.  We should figure out when we should freeze the aspen branches so that we can make base boxes and fully release aspen.  I would like to have the final box URLs before we merge this branch into master so we can just tag the point at which this branch is merged.  Also don't forget to rebase before building any base boxes.

singingwolfboy:
@feanil you and I talked in person about merging this now, and you said you were fine with it. Can you just add a thumbs-up comment verifying that, before I hit the merge button?

feanil:
@singingwolfboy  yes, we'll plan on a follow on PR once the Aspen RCs are frozen and we can build base boxes.

:+1:

vagrant/base/fullstack/Vagrantfile
```diff
@@ -39,9 +39,12 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provision :ansible do |ansible|
     # point Vagrant at the location of your playbook you want to run
     ansible.playbook = "../../../playbooks/vagrant-fullstack.yml"
+    ansible.verbose = "vvv"
     # set extra-vars here instead of in the vagrant play so that
     # they are written out to /edx/etc/server-vars.yml which can
     # be used later when running ansible locally
-    ansible.verbose = "vvvv"
+    if ENV['OPENEDX_RELEASE']
```
clytwynec, line 46: Why don't we need the release version of this file updated too? (as in https://github.com/edx/configuration/blob/master/vagrant/release/fullstack/Vagrantfile)

singingwolfboy, line 46: @clytwynec Please see the previous discussion in this pull request. We _will_ update the release version of this file, but that requires creating a Vagrant box, which we can't do until aspen is actually released.

clytwynec, line 46: Ok.  So to test fullstack, we will have to set the `OPENEDX_RELEASE` var and use this Vagrantfile?

clytwynec, line 46: @jzoldak  ^

singingwolfboy, line 46: I believe that's correct. If anyone in devops has any better ideas, I'm happy to hear them.

jzoldak, line 46: @singingwolfboy (I also just added this as a comment in @nedbat 's google doc). I believe that we need an aspen release branch on the configuration repo also. That way we can verify vagrant/base/devstack and /fullstack Vagrantfiles, then create the vagrant images for those and upload to s3, and update the Vagrantfiles in vagrant/release with the new .vbox names.
Also if we later find that we need to apply hotfixes to the configuration repo we will need to do that on the aspen release branch, in order to avoid pulling in more (possibly breaking) changes that might have gotten merged to master in the configuration repo in the interim.

clytwynec, line 46: fyi ---  I just created a branch off of current master named 'openedx/rc/aspen-2014-09-10'.

vagrant/release/devstack/Vagrantfile
```diff
@@ -23,7 +23,12 @@ cd /edx/app/edx_ansible/edx_ansible/playbooks
 # this can cause problems (e.g. looking for templates that no longer exist).
 /edx/bin/update configuration release
 
-ansible-playbook -i localhost, -c local vagrant-devstack.yml --tags=deploy -e configuration_version=release
+# Did we specify an openedx release?
+if [ -n "$1" ]; then
+  EXTRA_VARS="-e OPENEDX_RELEASE=$1"
+fi
+
+ansible-playbook -i localhost, -c local vagrant-devstack.yml --tags=deploy -e configuration_version=release $EXTRA_VARS
```
clytwynec, line 31: Sorry... one more question about this.  Does `configuration_version` in this line also need to be updated to use the release name (defaulting to 'release' or 'master' depending on how we want to handle it)?

vagrant/release/devstack/Vagrantfile
```diff
@@ -23,7 +23,12 @@ cd /edx/app/edx_ansible/edx_ansible/playbooks
 # this can cause problems (e.g. looking for templates that no longer exist).
 /edx/bin/update configuration release
```
clytwynec, line 24: I suppose same question here...

PR #1539, opened 2014-09-15
adding libfreetype6-dev

hkim823:
matplotlib is a new requirement in the sandbox, and it needs libfreetype6-dev before it complies properly


feanil:
:+1:

PR #1542, opened 2014-09-15
Change the sudo permissions for sandbox.

feanil:
Instead of rm directly we now do a find.


feanil:
https://github.com/edx/codejail/pull/27

jarv:
:+1:

PR #1543, opened 2014-09-15
Don't run pre-supervisor on abbey builds.

feanil:


playbooks/roles/supervisor/files/pre_supervisor_checks.py
```diff
@@ -96,6 +96,16 @@ def edp_for_instance(instance_id):
     instance_id = get_instance_metadata()['instance-id']
     prefix = instance_id
 
+
+    ec2 = boto.connect_ec2()
+    reservations = ec2.get_all_instances(instance_ids=[instance_id])
+    instance = reservations[0].instances[0]
+    if instance.instance_profile['arn'].endswith('/abbey'):
+        print("Running an abbey build. Not starting any services.")
```
jarv, line 104: you could do this without using the api with something like:
`curl http://169.254.169.254/latest/meta-data/iam/security-credentials/abbey`
Which will return a 404 except on servers where the IAM role is abbey

jarv:
I think I like the metadata approach better but this is fine :+1:

PR #1546, opened 2014-09-15
New changes for openedx vagrant release

hkim823:


hkim823:
Also moving aspen Vagrantfile into "base" vagrant folder

feanil:
:+1: to the new-relic change, the open-edx vagrant file should not have a release name as part of its path.

hkim823:
Maybe this should be part of this PR as well.  I'll work on this branch for more changes.

playbooks/vagrant-openedx.yml
```diff
@@ -45,4 +45,3 @@
     - certs
     - role: analytics-api
```
hkim823: Should this even be here?

PR #1547, opened 2014-09-16
this needs to default to false because it currently depends on

jarv:
analytics-server

analytics-server is a private repo


feanil:
:+1:

e0d:
:+1:

PR #1548, opened 2014-09-16
E0d/service builder

e0d:
@jarv @feanil 

initial thoughts?

The point is to provide tools to allow processing of ansible module output so that dependant modules can take the output as input.  This improves composability of modules.

For example

```
    - name: Create scaling policies
      ec2_scaling_policy:
        state: "{{ item.state }}"
        profile: "{{ item.profile }}"
        region: "{{ item.region }}"
        name: "{{ item.name }}"
        adjustment_type: "{{ item.adjustment_type }}"
        asg_name: "{{ item.asg_name }}"
        scaling_adjustment: "{{ item.scaling_adjustment }}"
        min_adjustment_step: "{{ item.min_adjustment_step }}"
        cooldown: "{{ item.cooldown }}"
      with_items: asg_policies
      register: created_policies

    - name: Apply function to policy data
      util_map:
        function: 'zip_to_dict'
        input: "{{ created_policies.results }}"
        args:
          - "name"
          - "arn"
      register: policy_data

    - name: Create metric alarms
      ec2_metric_alarm:
        state: "{{ item.state }}"
        region: "{{ aws_region }}"
        name: "{{ item.name }}"
        metric: "{{ item.metric }}"
        namespace: "{{ item.namespace }}"
        statistic: "{{ item.statistic }}"
        comparison: "{{ item.comparison }}"
        threshold: "{{ item.threshold }}"
        period: "{{ item.period }}"
        evaluation_periods: "{{ item.evaluation_periods }}"
        unit: "{{ item.unit }}"
        description: "{{ item.description }}"
        dimensions: "{{ item.dimensions }}"
        alarm_actions: "{{ policy_data.function_output[item.target_policy] }}"
      with_items: asg_alarms 
```


feanil:
The plugin seems pretty straightforward but I'm having a bit of trouble trying to think of use case where we would want this.  I guess the idea is to be able to converts data structures so that we can use keys to access them?

e0d:
The usecase is in the example above.  There are two ec2_modules that are interdependant, alarms need the ARNs of scaling policies.

However, the output from the scaling policy creation isn't easy to work with, especially when used with with_items.

This module maps a function onto that output in order to make it consumable by the metric alarm module which can be seen here:

alarm_actions: "{{ policy_data.function_output[item.target_policy] }}"

The target policy still needs to be explicitly provided in config.

feanil:
:+1:

feanil:
:+1:

PR #1551, opened 2014-09-18
Adding new env config value

e0d:
@hkim823 @jarv


jarv:
may be worth adding a changelog entry :+1:

PR #1555, opened 2014-09-18
Adding libfreetype6-dev

hkim823:
For matplotlib in sandbox to work

### DIFF
--- a/util/install/vagrant.sh
+++ b/util/install/vagrant.sh
@@ -25,7 +25,7 @@ sudo apt-get upgrade -y
 ##
 ## Install system pre-requisites
 ##
-sudo apt-get install -y build-essential software-properties-common python-software-properties curl git-core libxml2-dev libxslt1-dev python-pip python-apt python-dev
+sudo apt-get install -y build-essential software-properties-common python-software-properties curl git-core libxml2-dev libxslt1-dev libfreetype6-dev python-pip python-apt python-dev
 sudo pip install --upgrade pip
 sudo pip install --upgrade virtualenv
 


util/install/vagrant.sh
```diff
@@ -25,7 +25,7 @@ sudo apt-get upgrade -y
 ##
 ## Install system pre-requisites
 ##
-sudo apt-get install -y build-essential software-properties-common python-software-properties curl git-core libxml2-dev libxslt1-dev python-pip python-apt python-dev
+sudo apt-get install -y build-essential software-properties-common python-software-properties curl git-core libxml2-dev libxslt1-dev libfreetype6-dev python-pip python-apt python-dev
```
feanil, line 28: Why do this at all if ansible is going to get run?

hkim823, line 28: I ran into a scenario that ansible wasn't picking up the requirement building from a clean ubuntu box, just to be sure I thought I would add this in here.

feanil, line 28: @hkim823  I would opt for removing this line entirely and fixing the issue that was causing ansible to not install all system packages correctly.  Is this easily re-producible?

hkim823, line 28: Unsure how reproducible it is and will require longer term testing to see how ansible installs debian pacakges based on roles.

singingwolfboy:
I'm reproducing this failure: https://gist.github.com/singingwolfboy/390df7e9da3b6f51f6d1 and it sounds like @mattdrayer is as well.

benpatterson:
CC @clytwynec who was troubleshooting this as well.

hkim823:
Closing this PR, this should be handled by ansible, not by relying on manual pre install intervention

PR #1556, opened 2014-09-19
Jarv/newrelic log checks

jarv:
This adds an optional plugin for newrelic to grep for strings in any file on disk and report the occurrences of the match to newrelic.

The extra tasks are only run if `NEWRELIC_LOGWATCH` is defined which is set to a list of templates for watching agents that are checked into the newrelic role template directory.

### DIFF
--- a/playbooks/roles/newrelic/templates/opt/newrelic_platform_plugins/newrelic_logwatcher_agent/config/logwatch-certs-errors.j2
+++ b/playbooks/roles/newrelic/templates/opt/newrelic_platform_plugins/newrelic_logwatcher_agent/config/logwatch-certs-errors.j2
@@ -1,0 +1,15 @@
+{% if SANDBOX_USERNAME is defined %}
+{% set server_name = SANDBOX_USERNAME %}
+{% else %}
+{% set server_name = COMMON_ENVIRONMENT|default('unknown-env') + '-' + COMMON_DEPLOYMENT|default('unknown-deployment') %}
+{% endif %}
+
+  {{ server_name }}-certs:
+    # Full path to the the log file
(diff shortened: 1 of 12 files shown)


playbooks/roles/newrelic/templates/opt/newrelic_platform_plugins/newrelic_logwatcher_agent/config/logwatch-certs-errors.j2
```diff
@@ -0,0 +1,15 @@
+{% if SANDBOX_USERNAME is defined %}
+{% set server_name = SANDBOX_USERNAME %}
+{% else %}
+{% set server_name = COMMON_ENVIRONMENT|default('unknown-env') + '-' + COMMON_DEPLOYMENT|default('unknown-deployment') %}
```
feanil, line 4: There are default values for both COMMON_ENVIRONMENT and COMMON_DEPLOYMENT in the `common` role, why not just use those?

fredsmith:
:+1:

playbooks/roles/newrelic/templates/opt/newrelic_platform_plugins/newrelic_logwatcher_agent/config/logwatch-certs-errors.j2
jarv, line 4: This is a decision on whether we want this to bomb out here or use default values if those vars are not defined. 
The `newrelic` role doesn't depend on common (though it's often included with common) which is why I set defaults if those vars are not defined

feanil:
:+1:

PR #1557, opened 2014-09-19
E0d/insights role

e0d:
@jarv @feanil

### DIFF
--- a/playbooks/roles/insights/tasks/deploy.yml
+++ b/playbooks/roles/insights/tasks/deploy.yml
@@ -1,8 +1,8 @@
 ---
 
-- name: create notifier/.ssh directory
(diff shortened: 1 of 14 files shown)


playbooks/roles/insights/tasks/deploy.yml
```diff
@@ -0,0 +1,115 @@
+---
+
+- name: create notifier/.ssh directory
+  file:
+    path="{{ insights_home }}.ssh" mode=2700 state=directory owner={{ insights_user }} group={{ insights_user }}
```
jarv: `insights_home` doesn't end with a `/`, will this work?

jarv:
after comment :+1:

PR #1561, opened 2014-09-22
whitelisting 10.0.0.0/8

e0d:


feanil:
:+1:

jarv:
Hm, there was a reason why we removed this - https://github.com/edx/configuration/commit/7ea2fbeb2ab657453a92f366c4f5d715115aedf8#diff-b10b3eecfbcafe1bc698b53a5c470f9e

I think it was because it was disabling basic auth in all cases for traffic through an ELB

e0d:
@jarv I've made tis specific for now until we can decide upon a better solution.

jarv:
:+1:

PR #1565, opened 2014-09-22
whitelisting /oath2

e0d:
@jarv @feanil 

The initlal strategy won't work because of our proxy configurtion.  This covers the oauth2 case.


jarv:
:+1:

PR #1572, opened 2014-09-23
Use boto 2.32.1

singingwolfboy:
To match edx-platform

### DIFF
--- a/playbooks/roles/aws/defaults/main.yml
+++ b/playbooks/roles/aws/defaults/main.yml
@@ -47,7 +47,7 @@ aws_debian_pkgs:
 aws_pip_pkgs:
   - https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz
   - awscli==1.4.2
-  - boto==2.32.0
+  - boto==2.32.1
 
 aws_redhat_pkgs: []
 aws_s3cmd_version: s3cmd-1.5.0-beta1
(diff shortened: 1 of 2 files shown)


playbooks/roles/aws/defaults/main.yml
```diff
@@ -47,7 +47,7 @@ aws_debian_pkgs:
 aws_pip_pkgs:
   - https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-latest.tar.gz
   - awscli==1.4.2
-  - boto==2.32.0
+  - boto==2.32.1
```
singingwolfboy, line 50: I'm not sure if this actually needs to be updated -- I just did a search for `boto==`, found this reference, and figured I'd try to keep the versions as uniform as possible. If this change needs to be removed from the pull request, I'm fine with that.

feanil, line 50: I'm happy to upgrade it if it doesn't break things, so if you would like to keep everything up to date, go ahead and post a link to the changelog and I can take a look, otherwise, this will get updated at some point in the future when one of us has more time to look at it.

singingwolfboy, line 50: Release notes for 2.32.1: https://github.com/boto/boto/blob/develop/docs/source/releasenotes/v2.32.1.rst

feanil:
:+1:

PR #1575, opened 2014-09-23
Change to grep -F for get-rc-branches script

hkim823:


hkim823:
grep -P is taken out of OS X

util/jenkins/get-rc-branches.sh
```diff
@@ -62,7 +62,7 @@ fi
 
 $noop cd "/var/tmp/$repo_basename"
 if [[ -z $noop ]]; then
-    for branch in $(git branch -a | sort -r | tr -d ' ' | grep -P "$filter" ); do
+    for branch in $(git branch -a | sort -r | tr -d ' ' | grep -F "$filter" ); do
```
jarv: Did you  mean to switch this to `-E`?

hkim823: Up to you, -F for just glob match which I thought is what we're doing here, but if we want to expand the script to actually take a regex pattern then yes -E is better.  -P though is gone unfortunately in OS X

jarv: We pass a regex in, not a match string.
`get-rc-branches.sh -r https://github.com/edx/edx-platform -f ^rc/`

Can you test to see if this works with `-E`, if it does :+1:

PR #1576, opened 2014-09-24
setting minimum chunk size

jarv:


e0d:
:+1:

jarv:
This options is not working with s3cmd and I don't see a way to force the chunk size in the aws cli.

PR #1578, opened 2014-09-24
Add play for fixing and checking bash vulnerability

carsongee:


edx-jira:
Thanks for the pull request, @carsongee! I've created [OSPR-56](https://openedx.atlassian.net/browse/OSPR-56) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. Supporting information that is good to add there is anything that can help Product understand the context for the PR - supporting documentation, edx-code email threads, timeline information ('this must be merged by XX date', and why that is), partner information (this is for a course on edx.org, for example), etc. 

All technical communication about the code itself will still be done via the Github pull request interface. As a reminder, [our process documentation is here](http://edx.readthedocs.org/projects/userdocs/en/latest/process/overview.html).

We can't start reviewing your pull request until you've added yourself to the [AUTHORS](https://github.com/edx/edx-platform/blob/master/AUTHORS) file. Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst) file for more information.

jarv:
Maybe include this conditionally in tasks/main.yml?

carsongee:
In common? and what condition?

carsongee:
I should add, feel free to just do as you prefer and submit a new PR, probably easier than explaining ;)

jarv:
Yeah I was thinking that we would add `roles/common/tasks/security-ubuntu.yml` and do something like 

```
- include security-ubuntu.yml
  when: 
    - apply_security_updates is defined
    - ansible_distribution == 'Ubuntu'
```

though if the tasks are idempotent and not too expensive maybe we don't need `apply_security_updates`

carsongee:
sounds good, I'll shuffle it there. I think because update_cache needs to be set it isn't super fast so probably worth the condition

carsongee:
Updated.  I can remove the play if you want, but I thought it would be handy. Just let me know your preference if it differs

jarv:
actually if we are going to bake it into common how about we use `COMMON_SECURITY_UPDATES` as the var, default it to False in common/defaults/main.yml?

carsongee:
updated again

PR #1579, opened 2014-09-24
Update security check to be smarter

carsongee:

PR #1580, opened 2014-09-24
executable misspelled

fredsmith:


hkim823:
:+1:

carsongee:
Good catch, updated #1579 with fix...oddly enough it still worked right...

PR #1581, opened 2014-09-24
Changing out edxapp apt module is being called

hkim823:


hkim823:
@clytwynec can you see if this fixes bringing in the right apt dependencies for edxapp?

clytwynec:
I'm seeing a different error now: https://gist.github.com/clytwynec/58cdd0272167b2919a46
Unfortunately,  I get this error on 'release' too. I have no idea why.

PR #1582, opened 2014-09-24
switches the security task file to a role that is included by common

jarv:


jarv:
@carsongee @feanil @e0d This moves the security task file into a separate role so it is easier to apply by itself. thoughts?

feanil:
Why not just a security tag?

jarv:
Having it in its own role does a couple things

1) gives us the ability to run it from a play without using tags
2) allows us to run it outside of common

e0d:
Makes sense ot me, :+1:

carsongee:
:+1: from me

PR #1585, opened 2014-09-26
Fix check-migrations

hkim823:


util/jenkins/check-migrations.sh
```diff
@@ -62,9 +62,9 @@ extra_var_args+=" -e edxapp_user=jenkins"
 extra_var_args+=" -e syncdb=$syncdb"
 
 # Generate the json configuration files
-ansible-playbook -c local $extra_var_args --tags edxapp_cfg -i localhost, -s -U jenkins edxapp.yml
+ansible-playbook -c local "$extra_var_args" --tags edxapp_cfg -i localhost, -s -U jenkins edxapp.yml
```
feanil: Why are quotes needed for this? etxra_var_args can be multiple arguments not just one.

util/jenkins/get-rc-branches.sh
```diff
@@ -54,17 +54,20 @@ repo_basename=$(basename "$repo")
...
     for branch in $(git branch -a | sort -r | tr -d ' ' | grep -E "$filter" ); do
-      echo "origin/${branch}"
+        echo "origin/${branch}"
     done
-else
-    echo "Would have checked for branches using filter $filter"
+    for tag in $(git tag -l | sort -r | tr -d ' ' | grep -E "$filter"); do
+        echo "$tag"
+    done
+    else
```
feanil: remove leading space.

hkim823: Although it doesn't really matter, I think there was a mix of 2 and 4 space soft tabs in the previous script, normalizing it to 4 space soft tabs.

util/jenkins/check-migrations.sh
hkim823: It could I was worried about spaces not being escaped properly but since it's already working without I'll leave them out

util/jenkins/get-rc-branches.sh
feanil: The `else` line shouldn't have any spaces in front of it.  It should match the `if` above.

fredsmith:
:+1:

PR #1589, opened 2014-09-29
Add a list of available boto profiles to the alton environment.

feanil:


playbooks/roles/alton/templates/alton_env.j2
```diff
@@ -6,3 +6,5 @@ export {{ name }}="{{ value }}"
 {% endif %}
 {%- endfor %}
 
+export BOTO_PROFILES="{{ ALTON_AWS_CREDENTIALS|join(';') }}"
```
jarv, line 9: `ALTON_AWS_CREDENTIALS` is a dictionary of dictionaries, does this work?

jarv, line 9: nm, it appears to work

jarv:
after comments :+1:

playbooks/roles/alton/templates/alton_env.j2
jarv, line 9: personally I would have done something like `{{ ';'.join(ALTON_AWS_CREDENTIALS.keys()) }}`

feanil, line 9: I was just trying to stick to jinja syntax in there.

PR #1593, opened 2014-09-29
Use OPENEDX_RELEASE instead of release branch if specified

singingwolfboy:
@clytwynec Is this what you were looking for?

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -13,6 +13,7 @@ if [ ! -d /edx/app/edx_ansible ]; then
     echo "Error: Base box is missing provisioning scripts." 1>&2
     exit 1
 fi
+OPENEDX_RELEASE=$1
 export PYTHONUNBUFFERED=1
 source /edx/app/edx_ansible/venvs/edx_ansible/bin/activate
 cd /edx/app/edx_ansible/edx_ansible/playbooks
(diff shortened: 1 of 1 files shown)


clytwynec:
Yea.  Do you know if we will be continuing to use the 'release' branch for other reasons after named releases start?  If not, we should make the default 'master'.

singingwolfboy:
We will continue to use the `release` branch, even after we've pushed out our first named release. The `release` branch tracks the weekly releases to edx.org, while the named releases are going to be much less frequent than that.

fredsmith:
:+1:

vagrant/release/devstack/Vagrantfile
```diff
@@ -21,14 +22,14 @@ cd /edx/app/edx_ansible/edx_ansible/playbooks
 # The vagrant-devstack.yml playbook will also do this, but only
 # after loading the playbooks into memory.  If these are out of date,
 # this can cause problems (e.g. looking for templates that no longer exist).
-/edx/bin/update configuration release
+/edx/bin/update configuration ${OPENEDX_RELEASE-release}
```
feanil, line 25: Don't you need a `:` before the dash?

singingwolfboy, line 25: [Nope.](http://www.tldp.org/LDP/abs/html/parameter-substitution.html) "${parameter-default} and ${parameter:-default} are almost equivalent. The extra : makes a difference only when parameter has been declared, but is null."

jarv, line 25: I prefer extra if statements to inscrutable shell syntax personally, but this works.

PR #1594, opened 2014-09-29
automatic upgrades on image creation, enable unattended upgrades daily

fredsmith:


playbooks/roles/security/tasks/security-ubuntu.yml
```diff
@@ -1,3 +1,17 @@
+#### Enable periodic security updates
+
+- name: install security packages
+  apt: pkg={{','.join(security_debian_pkgs)}} state=latest
```
hkim823: I'm wondering if we should move to the convention of using name instead of pkg, and using `{{ item}}` and `with_items: {{ security_debian_pkgs }}`

carsongee:
Can we make the unattended stuff optional behind a when (ok to default to on)?  I've lost a few too many servers to unattended ;)

fredsmith:
@carsongee - default is now 'false' on unattended upgrades and on running updates during ansible run.  In addition, you can set unattended upgrades to only upgrade from the security repo (default) or from all repos.

carsongee:
Cool, thanks

feanil:
:+1:   One other thing we may want to consider is being able to turn off un-attended upgrades( right now this isn't possible but I'm also not 100% about whether or not we need that in ansible.

fredsmith:
@feanil you mean, making the file for /etc/apt/apt.conf.d/02periodic state=absent when someone sets SECURITY_UNATTENDED_UPGRADES: false?

feanil:
yea

PR #1595, opened 2014-09-30
Upgrading S3 pluging to 0.6

hkim823:


e0d:
:+1: assuming tested.

playbooks/roles/jenkins_master/defaults/main.yml
```diff
@@ -36,7 +36,7 @@ jenkins_plugins:
     - { name: "postbuild-task", version: "1.8" }
     - { name: "PrioritySorter", version: "2.8" }
     - { name: "sauce-ondemand", version: "1.61" }
-    - { name: "s3", version: "0.5" }
+    - { name: "s3", version: "0.6" }
```
feanil, line 39: We should leave this as is unless @jzoldak you want to upgrade this.  We did so on the admin since the newer version can deal with folders instead of uploading everything in a flat structure.

playbooks/roles/jenkins_admin/defaults/main.yml
```diff
@@ -119,7 +119,7 @@ jenkins_admin_plugins:
   - { name: "jquery", version: "1.7.2-1" }
   - { name: "dashboard-view", version: "2.9.4" }
   - { name: "build-pipeline-plugin", version: "1.4.3" }
-  - { name: "s3", version: "0.5" }
+  - { name: "s3", version: "0.6" }
```
feanil, line 122: :+1:

feanil:
Upgrading jenkins_master might break test eng expectations about that plugin so we should wait till they can review it or pull that part out for now.

jzoldak:
@benpatterson do we have a way to test this out on Jenkins Master?

benpatterson:
Although we could test it on our test jenkins instance, this change is n/a for us. We have the plugin on testeng jenkins, but we are not using it (i.e. no S3 profiles, etc...)

jzoldak:
:+1: :ship:

PR #1596, opened 2014-09-30
Fixup config version in vagrant devstack

singingwolfboy:
@stephensanchez does this work for you?


stephensanchez:
:+1: yup! Is release what we used to do, or master? Release makes sense, just was curious.

singingwolfboy:
@stephensanchez it used to be release. The `${OPENEDX_RELEASE-release}` syntax in bash _should_ mean "Use the contents of the `$OPENEDX_RELEASE` variable, or if its unset, use the string `release`", but I guess something was getting confused. This syntax, while more verbose, should also be clearer.

stephensanchez:
oh, I get ya. OK well looks good to me then!

fredsmith:
I agree the old syntax should have worked.  This is clearer, though.  Looks good to me. :+1:

PR #1600, opened 2014-10-01
edxapp user should have .npmrc

singingwolfboy:
Work-around for [OPS-116](https://openedx.atlassian.net/browse/OPS-116)

### DIFF
--- a/playbooks/roles/edxapp/tasks/main.yml
+++ b/playbooks/roles/edxapp/tasks/main.yml
@@ -38,7 +38,6 @@
     state=link
     owner="{{ edxapp_user }}"
     group="{{ common_web_group }}"
-    
 
 - name: create edxapp log dir
   file: >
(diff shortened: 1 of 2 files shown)


playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -70,6 +69,11 @@
   - "restart edxapp"
   - "restart edxapp_workers"
 
+- name: set up edxapp .npmrc
+  template:
+    src=.npmrc.j2 dest={{ edxapp_app_dir }}/.npmrc
+    owner={{ edxapp_user }} mode=0600
```
feanil: Please set the group to {{edxapp_user}} as well and set the mode to 644 so that it is world readable.

feanil: Actually lets set the group to '{{ common_web_group }}'

singingwolfboy:
I'm setting the mode to 600 because [that's what `npm` wants it to be](https://github.com/npm/npm/blob/48cda625b39ce4f76e99288ca02f907ba534a354/node_modules/npmconf/npmconf.js#L262). It would be nice if I could make it world-readable, but `npm` will just set it back to 600, so I figured I'd just go with what `npm` wants.

feanil:
That's annoying but your decisions make sense. :+1:

PR #1601, opened 2014-10-01
Fixing faulty defaults from 3f98d8f

singingwolfboy:
We can't currently bring up a devstack from scratch, because these variables are undefined. This should avoid those errors, introduced in 3f98d8f

@e0d can you review?


playbooks/roles/analytics-api/defaults/main.yml
```diff
@@ -50,18 +50,18 @@ ANALYTICS_API_EMAIL_HOST_USER: 'mail_user'
...
-  EMAIL_HOST: '{{ ANALYTICS_API_EMAIL_HOST }}'
-  EMAIL_HOST_PASSWORD: '{{ ANALYTICS_API_EMAIL_HOST_PASSWORD }}'
-  EMAIL_HOST_USER: '{{ ANALYTICS_API_EMAIL_HOST_USER }}'
-  EMAIL_PORT: '{{ ANALYTICS_API_EMAIL_PORT }}'
-  API_AUTH_TOKEN: '{{ ANALYTICS_API_AUTH_TOKEN }}'
+  EMAIL_HOST: '{{ ANALYTICS_API_EMAIL_HOST|default("smtp.example.com") }}'
+  EMAIL_HOST_PASSWORD: '{{ ANALYTICS_API_EMAIL_HOST_PASSWORD|default("") }}'
+  EMAIL_HOST_USER: '{{ ANALYTICS_API_EMAIL_HOST_USER|default("") }}'
+  EMAIL_PORT: '{{ ANALYTICS_API_EMAIL_PORT|default(587) }}'
+  API_AUTH_TOKEN: '{{ ANALYTICS_API_AUTH_TOKEN|default('dummy-token') }}'
```
jarv, line 64: I'm not opposed to this but another way is to define these at the top like we do other vars.

e0d, line 64: Yes, please just move the definitions to the top.

singingwolfboy:
Looks like the problem was misspelled variable names. Fixed in #1602.

PR #1606, opened 2014-10-03
Switch from legacy variable templating system to Jinja2 everywhere

singingwolfboy:
Just updating the templates to use Jinja2 everywhere, without changing the version of ansible.


playbooks/edx-west/cloudformation.yml
```diff
@@ -31,14 +31,14 @@
...
       template_parameters:
-        KeyName: $key
+        KeyName: "{{key}}"
         InstanceType: m1.small
-        GroupTag: $group
+        GroupTag: "{{group}}"
     register: stack
   - name: show stack outputs
-    debug: msg="My stack outputs are ${stack.stack_outputs}"
+    debug: msg="My stack outputs are {{stack.stack_outputs}}"
```
e0d, line 44: @jbau are you using the files in edx-west?  Can you review?

playbooks/roles/discern/defaults/main.yml
```diff
@@ -53,23 +53,23 @@ discern_env_config:
...
-      PASSWORD: $DISCERN_MYSQL_PASSWORD
-      PORT: $DISCERN_MYSQL_PORT
-      USER: $DISCERN_MYSQL_USER
-  GOOGLE_ANALYTICS_PROPERTY_ID: $DISCERN_GOOGLE_ANALYTICS_PROPERTY_ID
+      HOST: "{{DISCERN_MYSQL_HOST}}"
+      NAME: "{{DISCERN_MYSQL_DB_NAME}}"
+      PASSWORD: "{{DISCERN_MYSQL_PASSWORD}}"
+      PORT: "{{DISCERN_MYSQL_PORT}}"
+      USER: "{{DISCERN_MYSQL_USER}}"
+  GOOGLE_ANALYTICS_PROPERTY_ID: "{{DISCERN_GOOGLE_ANALYTICS_PROPERTY_ID}}"
```
e0d, line 72: I'd argue we should retire this role, @feanil any reason not to?

e0d:
@singingwolfboy Have you tested this?  Minimally, you should be able to build a sandbox using your branch of config and have everything run as expected.

Also, needs a rebase.

playbooks/roles/discern/defaults/main.yml
feanil, line 72: It may be worth an e-mail to edx-code before we do just to make sure no one in the open source community is using this but hopefully not.  I'm pro retiring this role.

singingwolfboy, line 72: It may be a good idea to retire this role, but I think that's out of scope for this pull request. This pull request is only focused on getting rid of the legacy templating, and that's all. Maybe a follow-up pull request can remove this role.

singingwolfboy:
@e0d I just tried creating a sandbox with this branch, and I got a very strange error: `error while evaluating conditional: ansible_distribution == 'Ubuntu'`. This doesn't seem related to my changes. Any ideas? https://gist.github.com/singingwolfboy/4b1a3b600000f850e0af#file-gistfile1-txt-L1349

e0d:
@singingwolfboy Don't have time to dig ATM, but I suspect it's a conditionals change related to update to ansible, I'm actually not certain that the ansible version used when building the sandbox respects what's defined in the config repos requirements.txt.  Need to confirm that.

singingwolfboy:
@e0d this pull request specifically does _not_ update the version of ansible, it _only_ changes the templates. Ansible 1.5.5 should be able to support the Jinja2 templating everywhere, and I figured it was a good transitional step to getting the upgrade to happen.

e0d:
Oh, then we must not be gathering facts on the sandbox runs, you need:
gather_facts: true

feanil:
closing this as it is part of https://github.com/edx/configuration/pull/1632

PR #1607, opened 2014-10-03
OPENEDX_RELEASE doesn't work properly with variable inheritance in ansible

singingwolfboy:
@feanil 

Works for me locally.


feanil:
:+1:

PR #1614, opened 2014-10-07
Don't use {{ }} for loop variables when it's unnecessary

singingwolfboy:


playbooks/edx-east/create_dbs.yml
```diff
@@ -129,55 +129,55 @@
         name: "{{ COMMON_MYSQL_ADMIN_USER }}"
         priv: "*.*:CREATE USER"
         password: "{{ COMMON_MYSQL_ADMIN_PASS }}"
-        login_host: "{{ item.db_host }}"
+        login_host: "{{ item.db_host|default('None') }}"
```
feanil, line 132: Why make this semantic change as a part of this PR?

singingwolfboy, line 132: In the `with_items` section below, we had to use Jinja templates so that we could use the `default` filter. However, by moving the `default` filter up to here, we can dispense with the Jinja templates in the `with_items` section.

e0d:
@feanil @singingwolfboy should this PR be closed?

singingwolfboy:
@e0d I still think it's worth merging. I've rebased the pull request.

PR #1616, opened 2014-10-07
Set none as the default.

feanil:
No one will have access to the fake default bucket causing
things to break if it isn't overridden.


fredsmith:
:+1:

PR #1619, opened 2014-10-07
Set the correct default for EDXAPP_MKTG_URL_LINK_MAP

feanil:
@hkim823 

FYI: @stephensanchez  @dianakhuang 

By setting it to a literal None here, it will allow us to use the defaults.


hkim823:
:+1:

feanil:
Turns out this change is not needed.  The settings are being overridden correctly. I just verified settings.MKTG_URL_LINK_MAP has the correct defaults.

PR #1620, opened 2014-10-08
Using ansible instead of ssh for jenkins scripts

hkim823:


util/jenkins/certificate-whitelist.sh
```diff
@@ -4,14 +4,28 @@ cd configuration
 pip install -r requirements.txt
 env
 
-ip=$(python playbooks/ec2.py | jq -r '."tag_Name_prod-edx-worker"[0] | strings')
-ssh="ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
-manage="cd /edx/app/edxapp/edx-platform && sudo -u www-data /edx/bin/python.edxapp ./manage.py"
+if [[ "$cluster" = "prod" ]]; then
```
feanil: This conditional can be simplified to just putting the cluster var in the string `first_in_tag_name_${environment}-${deployment}-worker`

util/jenkins/change-enrollment.sh
```diff
@@ -2,27 +2,18 @@ cd configuration
...
+  ansible="ansible first_in_tag_Name_stage-edx-worker -i playbooks/ec2.py -u ubuntu -s -U www-data -a"
+elif [[ "$cluster" = "edge" ]]; then
+  ansible="ansible first_in_tag_Name_prod-edge-worker -i playbooks/ec2.py -u ubuntu -s -U www-data -a"
 fi
 
-ssh="ssh -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
-manage="cd /edx/app/edxapp/edx-platform && sudo -u www-data /edx/bin/python.edxapp ./manage.py lms change_enrollment"
+manage="/edx/bin/python.edxapp /edx/bin/manage.py lms change_enrollment --settings aws"
 
 if [ $"noop" ]; then
```
feanil: Does this work?(Quoted var name?)  I would think that noop would just never work.

feanil:
It looks like generally in all of these scripts you're using cluster when you mean environment.  Can we update everything so that we use environment and deployment explicitly since the cluster is usually 'worker'.

util/jenkins/certificate-whitelist.sh
hkim823: That's a good way to do it, I'll add that in and change the Jenkins jobs to reflect that

feanil:
looks good to me.  @e0d if it looks good to you, I think we should merge it.

e0d:
yeah :+1:

PR #1622, opened 2014-10-08
Remove ora from fullstack.

feanil:


e0d:
@adampalay you OK with this?

I'm :+1:

feanil:
I talked to @adampalay about this in hipchat. He was cool with this since it is still optionally available in devstack.

PR #1623, opened 2014-10-10
Make devstack release Vagrantfile use aspen-rc1 box if requested

singingwolfboy:


vagrant/release/devstack/Vagrantfile
```diff
@@ -54,11 +54,21 @@ if ENV['VAGRANT_MOUNT_BASE']
 
 end
 
+openedx_releases = {
```
feanil, line 60: So I'm not familiar with Ruby's hash syntax can you explain what's going on here?  Maybe add a comment to that effect?

singingwolfboy, line 60: It's just a mapping from <git branch name> to the name and URL path to use for the Vagrant box. I've added comment to that effect in the pull request. Is it clear enough, or is it still confusing?

vagrant/release/devstack/Vagrantfile
```diff
@@ -54,11 +54,24 @@ if ENV['VAGRANT_MOUNT_BASE']
...
 
+# map the name of the git branch that we use for a release
+# to a name and a file path, which are used for retrieving
+# a Vagrant box from the internet.
+openedx_releases = {
+  "openedx/rc/aspen-2014-09-10" => {
+    :name => "aspen-rc1", :file => "20141009-aspen-devstack-rc1.box",
+  }
+}
+openedx_releases.default = {
```
feanil, line 65: So is the behaviour of this default has just the settings to use if rel is 'nil'?

singingwolfboy, line 65: Correct. If the user doesn't have an `OPENEDX_RELEASE` environment variable set, or its set to an unknown value, then we'll fall back on this default.

feanil:
:+1:

PR #1625, opened 2014-10-10
Apache LMS template listen on all ports

carsongee:
@marcore and @jbau This is a workaround for the apache role changes I made earlier.  The original purpose of allowing multiple ports was to support having an 80 and 443, but this would cause the same VirtualHost to respond on all specified ports, so if you wanted to do that you would need to override the template I think.

### DIFF
--- a/playbooks/roles/apache/templates/lms.j2
+++ b/playbooks/roles/apache/templates/lms.j2
@@ -1,7 +1,7 @@
 WSGIPythonHome {{ edxapp_venv_dir }}
 WSGIRestrictEmbedded On
 
-<VirtualHost *:{{ apache_port }}>
+<VirtualHost *:*>
     ServerName https://{{ lms_env_config.SITE_NAME }}
     ServerAlias *.{{ lms_env_config.SITE_NAME }}
     UseCanonicalName On


openedx-webhooks:
Thanks for the pull request, @carsongee! I've created [OSPR-136](https://openedx.atlassian.net/browse/OSPR-136) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. Supporting information that is good to add there is anything that can help Product understand the context for the PR - supporting documentation, edx-code email threads, timeline information ('this must be merged by XX date', and why that is), partner information (this is for a course on edx.org, for example), etc. 

All technical communication about the code itself will still be done via the Github pull request interface. As a reminder, [our process documentation is here](http://edx.readthedocs.org/projects/userdocs/en/latest/process/overview.html).

playbooks/roles/apache/templates/lms.j2
```diff
@@ -1,7 +1,7 @@
 WSGIPythonHome {{ edxapp_venv_dir }}
 WSGIRestrictEmbedded On
 
-<VirtualHost *:{{ apache_port }}>
+<VirtualHost *:*>
```
Gerst20051, line 4: when would apache not be on 80?

marcore:
@Gerst20051 In a sanbox when Shibboleth auth is needed you should have nginx on 80 and 443 (for all open edx stuff) and apache on 8080 (or whatever) installed only for the mod_shib module.
Nginx proxy the shib-login request to apache to enable Shibboleth authentication

Gerst20051:
@marcore ahhhhh thanks that makes sense now :+1:

jbau:
if apache is listening to `*`, won't it fight with nginx listening e.g. on 80?  @carsongee

marcore:
@jbau Should be ok because the socket port is created by directive
`Listen` in the template `ports.j2` that correctly iterate over apache_ports variable.
In the VirtualHost you "select" one of the port specified by Listen directive

jbau:
got it. sounds good.

jbau:
:+1:

PR #1634, opened 2014-10-15
Add a default cache version.

feanil:


e0d:
:+1: We should send mail to the ops mailing list about this.

feanil:
Mail sent.

PR #1637, opened 2014-10-17
move from mongo 10gen to org

fredsmith:


feanil:
:+1:

carsongee:
So have you guys been able to actually add users with this role since the upgrade to 2.6?  I'm getting ansible telling me it's working but getting:

``` terminal
rs0:PRIMARY> show users
rs0:PRIMARY> use edxapp
switched to db edxapp
rs0:PRIMARY> db.getUsers()
[ ]
rs0:PRIMARY>
```

I tried moving pymongo in the role to the latest and that caused the user creations steps to fail entirely.

feanil:
I saw a case of this yesterday, but I need to dig into it more.

carsongee:
Just as an update as I did some digging.  The entire auth model changes in 2.6 and it requires you to set a super user account as your first account, and then the localhost auth exception no longer happens.  I started looking into modifying the play to create this user as an early task, but haven't finished it.  It looks like we may need to do a future grab of the module, but not 100% sure yet. http://docs.mongodb.org/manual/tutorial/add-admin-user/ for more details.

fredsmith:
i haven't had issues after running the user authentication schema migration:

http://docs.mongodb.org/manual/release-notes/2.6-upgrade-authorization/

```
 db.getSiblingDB("admin").runCommand({authSchemaUpgrade: 1 });
```

carsongee:
Well this appears to be a replica set thing with new clusters (via the new vagrant cluster) and fresh installs of 2.6. Tried a lot of stuff and still haven't had any luck doing anything with the ansible module to create users. Thanks for tips.  I'll hopefully figure something out for clustered and have a PR sometime in the near future.

PR #1641, opened 2014-10-20
validate certs=no on apt_repository due to ansible bug

fredsmith:

PR #1642, opened 2014-10-20
Revert "validate certs=no on apt_repository due to ansible bug"

feanil:
Reverts edx/configuration#1641


feanil:
Reverting because this is not available in ansible 1.5.5

PR #1643, opened 2014-10-20
Activating nodeenv to use bower

clintonb:
The nodeenv needs to be active in order to add `node` to the `PATH`. The `bower` command makes calls to `node`.

@e0d @feanil

### DIFF
--- a/playbooks/roles/insights/tasks/deploy.yml
+++ b/playbooks/roles/insights/tasks/deploy.yml
@@ -62,7 +62,7 @@
 - name: install bower dependencies
   shell: >
     chdir={{ insights_code_dir }}
-    {{ insights_node_modules_dir }}/.bin/bower install --production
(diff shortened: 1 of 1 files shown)


feanil:
What settings does this need to set, is it just the NODE_PATH?  If so, I would prefer to set that as an environment setting.  If it's more complicated then this is fine but I would rather you say `source` than `.`

clintonb:
The activate script updates PATH as well. `source` failed on both my sandbox and during the AMI build. I'll make the environment setting change.

clintonb:
@feanil The activate script sets a few other variables as well. It's not complicated, just a bit tedious. I'd prefer to stick with using it. Let me know if you have thoughts on `source` not working.

clintonb:
@e0d @feanil Can this be merged?

feanil:
:+1:

PR #1644, opened 2014-10-20
Feanil/ignore cert validations

feanil:
FYI: @fredsmith @jarv

### DIFF
--- a/playbooks/roles/jenkins_admin/tasks/main.yml
+++ b/playbooks/roles/jenkins_admin/tasks/main.yml
@@ -34,7 +34,7 @@
   when: JENKINS_ADMIN_S3_PROFILE.secret_key is not defined
 
 - name: add admin specific apt repositories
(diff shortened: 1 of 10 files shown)


fredsmith:
:+1:

playbooks/roles/jenkins_admin/tasks/main.yml
```diff
@@ -34,7 +34,7 @@
   when: JENKINS_ADMIN_S3_PROFILE.secret_key is not defined
 
 - name: add admin specific apt repositories
-  apt_repository: repo="{{ item }}" state=present update_cache=yes
+  apt_repository_1.8: repo="{{ item }}" state=present update_cache=yes validate_certs=no
```
carsongee, line 37: Didn't we want to use _local_ver like the other modules we have grabbed from the future?

PR #1653, opened 2014-10-22
Create default mongo users.

feanil:
Because the application expects the users to exist and tries to use them for auth.  In 2.6.4 if you provide auth and the mongo database has no auth setup, it will fail to authenticate.  This is a change in behaviour.


fredsmith:
:+1:

PR #1655, opened 2014-10-22
Add libjpeg-dev to edxlocal play

singingwolfboy:
Because running the tests for thumbnail manipulation requires it: see https://github.com/edx/edx-platform/pull/5685

### DIFF
--- a/playbooks/roles/edxlocal/defaults/main.yml
+++ b/playbooks/roles/edxlocal/defaults/main.yml
@@ -3,3 +3,4 @@ edxlocal_debian_pkgs:
   - python-mysqldb
   - mysql-server-5.5
   - postfix
+  - libjpeg-dev


feanil:
Is this a dependency of edxapp? If so I would put it in the requirements for edxapp.

singingwolfboy:
@feanil: [it's there already](https://github.com/edx/configuration/blob/0c0b0b462dc804112d6d6646473b6ce299925062/playbooks/roles/edxapp/defaults/main.yml#L750-L751).

feanil:
So why is this needed in edxlocal?

singingwolfboy:
@feanil: because it apparently isn't installed for the Jenkins worker. [Here's the pull request](https://github.com/edx/edx-platform/pull/5685) where I re-enable a test that relies on libjpeg, and [here's a failed Jenkins run for that pull request](https://jenkins.testeng.edx.org/job/edx-all-tests-auto-pr/17176/SHARD=1,TEST_SUITE=unit/testReport/junit/contentstore.tests.test_contentstore/ImportRequiredTestCases/test_asset_import/) with the error message `IOError: decoder jpeg not available`.

benpatterson:
@feanil @singingwolfboy depending on when that was added to edxapp, the jenkins worker AMI we're using is ~1 mo. old

Just FYI depending on when those were added

Unfortunately we couldn't test whether or not a new run of the worker would install these today because of the CA certs problem

feanil:
That would have been there for a while.  If it's not being installed to the worker, it's probably worth a deeper investigation as to why that isn't happening...looking at the jenkins_worker and play, it doesn't install any of the debian packages from the edxapp list as far as I can tell.  I'm surprised we haven't run into more bugs...I think our best option would be to have a task for the jenkins_worker that will install the edxapp debian packages on the box.

benpatterson:
@feanil @singingwolfboy I pulled up an instance of Jenkins-worker and validated that installing libjpeg8-dev gets the job done. (Though, I also installed libpng12-dev before validating...could that be added to this PR?)

Prior to these packages being installed, indeed, the test Dave mentioned was skipped because libjpeg support was not on the system.

feanil:
@benpatterson my suggestion would be that rather than running the edxlocal role to install this, you put a tag on the installation of the edxapp system packages so that they can be installed on the jenkins workers.  That way people don't have to update two places when new system dependencies are added or old ones are removed.

feanil:
Or perhaps reference the edxapp system packages var from edxlocal if you want to avoid having the edxapp role in here, though we have generally tried to avoid this sort of thing, I think we already do something like this for the jenkins workers in other cases.

benpatterson:
I agree we should revisit edxlocal vs edxapp for the jenkins_worker role, but that would be a large undertaking at this point. I suggest we defer that piece of it so we can get the Pillow upgrade done.

benpatterson:
As per discussion, we can move forward with this change and look at refactoring how we manage packages between edxlocal and edxapp in a separate effort.

:+1: from me.

feanil:
:+1:

benpatterson:
@singingwolfboy I think we're good to merge this as-is. Think that could happen today? I'd like to get moving on this train if we can.

PR #1662, opened 2014-10-24
E0d/s3 log perms

e0d:
@jarv @feanil @fredsmith  

related to https://openedx.atlassian.net/browse/DEVOPS-291.  Includes some general re-factoring too.  I'd like to fixate on this service directory creation pattern if you like it.

Still need to test on my sandbox.

### DIFF
--- a/playbooks/roles/aws/defaults/main.yml
+++ b/playbooks/roles/aws/defaults/main.yml
@@ -28,11 +28,26 @@ AWS_S3_LOGS_FROM_EMAIL: dummy@example.com
 # vars are namespace with the module name.
 #
 aws_role_name: aws
-aws_data_dir: "{{ COMMON_DATA_DIR }}/aws"
-aws_app_dir: "{{ COMMON_APP_DIR }}/aws"
-aws_s3_sync_script: "{{ aws_app_dir }}/send-logs-to-s3"
-aws_s3_logfile: "{{ aws_log_dir }}/s3-log-sync.log"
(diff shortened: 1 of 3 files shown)


playbooks/roles/aws/defaults/main.yml
```diff
@@ -28,11 +28,26 @@ AWS_S3_LOGS_FROM_EMAIL: dummy@example.com
...
+    path: "{{ COMMON_LOG_DIR }}/{{ aws_role_name }}"
+    owner: "syslog"
+    group: "syslog"
+    mode: "0700"
+  data:
+    path: "{{ COMMON_DATA_DIR }}/{{ aws_role_name }}"
+    owner: "root"
+    group: "root"
+    mode: "0700"
+
```
jarv, line 48: why are these directories in the aws role instead of a more generic service role?

feanil:
:+1:  Once sandbox testing is done.

playbooks/roles/aws/defaults/main.yml
feanil, line 48: Pulling them out further seems hard since different roles have differing number of directories that they need to create and track.

e0d, line 48: I have another branch in the works that will pull this stuff up further, I wanted to make this better without fixing everything.

e0d:
Testing went well.

jarv:
:+1:

PR #1665, opened 2014-10-24
Updates to Make Insights Configuration Sandbox-Friendly

clintonb:
I have copied the behavior of analytics-api. I am unable to fully test this since `/edx/app/edx_ansible/server-vars.yml` is not created with `INSIGHTS_GIT_IDENTITY`. I added it manually and confirmed via curl that the server starts; however, I cannot access it externally (security group block perhaps?).

@e0d @feanil


e0d:
:+1:

PR #1668, opened 2014-10-28
Skip validating cert :(

benpatterson:
This seems to be necessary to get our jenkins-workers building again. I don't entirely understand since this worked a month ago and we were running it with ansible 1.7.1 anyway.

@fredsmith I don't suppose you could suggest a different solution? 

CC @singingwolfboy

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/packer.yml
+++ b/playbooks/roles/jenkins_worker/tasks/packer.yml
@@ -1,6 +1,6 @@
 ---
 - name: Download packer
-  get_url: url={{ packer_url }} dest=/var/tmp/packer.zip
+  shell: "curl -L {{ packer_url }} -o /var/tmp/packer.zip"
 
 - name: Unzip packer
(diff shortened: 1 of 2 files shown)


singingwolfboy:
:+1: :cry:

playbooks/roles/jenkins_worker/tasks/packer.yml
```diff
@@ -1,6 +1,6 @@
 ---
 - name: Download packer
-  get_url: url={{ packer_url }} dest=/var/tmp/packer.zip
+  get_url: url={{ packer_url }} dest=/var/tmp/packer.zip validate_certs=no
```
feanil: An alternative would be to use CURL instead of get_url which seems to be validating certs correctly.  Take a look at what @fredsmith did for the get_url calls in common.

benpatterson:
@fredsmith @feanil here is the curl implementation. This fixes the problem as well.

I can squash before merge but am keeping the commit separate for your review... Thx for the look.

fredsmith:
:+1:

feanil:
:+1:

PR #1671, opened 2014-10-29
new relic s3 plugin

benpatterson:
@e0d 

Still a work in progress but if you have time, it may be worthwhile to review.

Big TODO's at this writing:
- daemonize it via supervisor (I'm going to do that last, potentially in a separate PR)
- I was not using common because of everything that comes with it. Rbenv does not have an explicit dependency on common; however, it is unable to function without it because it needs to use git. So I'm either going to establish a dependency on common, or solve it some other way.

### DIFF
--- a/playbooks/roles/newrelic/defaults/main.yml
+++ b/playbooks/roles/newrelic/defaults/main.yml
@@ -43,3 +43,13 @@ newrelic_logwatch_repo: https://github.com/railsware/newrelic_platform_plugins
 newrelic_logwatch_repo_dir: /opt/newrelic_platform_plugins
 newrelic_logwatch_dir: "{{ newrelic_logwatch_repo_dir }}/newrelic_logwatcher_agent"
 newrelic_logwatch_version: "8edd6d214e462b27fdd07d41712eb7b4fff2f7d8"
+
+# s3_watcher
+newrelic_plugin_user: "{{ common_web_user }}"
+newrelic_s3watcher_repo: https://github.com/edx/analytics-gasket
+newrelic_s3watcher_version: "e33cf920aa172707b788834e176350f423c7c38d"
+newrelic_s3watcher_repo_dir: /edx/app/s3watcher
+newrelic_s3watcher_dir: "{{ newrelic_s3watcher_repo_dir }}/newrelic/s3-newrelic-plugin"
+
+rbenv_root_dir: "/opt/rbenv"
+
--- a/playbooks/roles/newrelic/tasks/s3-watcher.yml
+++ b/playbooks/roles/newrelic/tasks/s3-watcher.yml
@@ -1,0 +1,75 @@
+---
+# This task file is for the newrelic s3watcher plugin
+# which is an agent that obtains metrics from S3 for
+# a given bucket
+
+- name: Install newrelic plugin related packages
+  apt: pkg={{ item }} install_recommends=yes state=present
+  with_items: newrelic_debian_plugin_pkgs
+
+- name: Install s3cmd
+  apt: pkg="s3cmd" install_recommends=yes state=present
+
+- name: create {{ newrelic_s3watcher_repo_dir }}
+  file: >
+    path={{ newrelic_s3watcher_repo_dir }} state=directory
+    owner="{{ newrelic_plugin_user }}" group="{{ common_web_group }}"
+    mode=0755
+
+- name: check out the s3 newrelic plugin
+  git: >
+    dest={{ newrelic_s3watcher_repo_dir }}
+    repo={{ newrelic_s3watcher_repo }} version={{ newrelic_s3watcher_version }}
+    accept_hostkey=yes
+  sudo_user: "{{ newrelic_plugin_user }}"
+
+- name: create bundle install shell script
+  template:
+    owner: "{{ newrelic_plugin_user }}"
+    src: s3-watcher-bundle-install.sh.j2
+    dest: "{{ newrelic_s3watcher_repo_dir }}/s3-watcher-bundle-install.sh"
+    mode: 0755
+  sudo_user: "{{ newrelic_plugin_user }}"
+
+- name: execute bundle install via rbenv
+  shell: ./s3-watcher-bundle-install.sh chdir={{ newrelic_s3watcher_repo_dir }}
+
+- name: create agent configuration
+  template: >
+    src=opt/newrelic_platform_plugins/s3_watcher/config/newrelic_plugins.yml.j2
+    dest={{ newrelic_s3watcher_dir }}/config/newrelic_plugin.yml
+  sudo_user: "{{ newrelic_plugin_user }}"
+
+- name: set agent as executable
+  file: >
+    path={{ newrelic_s3watcher_dir }}/newrelic_s3_agent
+    mode=0755
+
+- name: create supervisor wrapper
+  template:
+    src: s3-watcher-supervisor.sh.j2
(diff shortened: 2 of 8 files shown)


playbooks/roles/newrelic/tasks/s3-watcher.yml
```diff
@@ -0,0 +1,38 @@
...
+    dest={{ newrelic_s3watcher_repo_dir }}
+    repo={{ newrelic_s3watcher_repo }} version={{ newrelic_s3watcher_version }}
+    accept_hostkey=yes
+
+- name: create bundle install shell script
+  template:
+    owner: root
+    src: s3-watcher-bundle-install.sh.j2
+    dest: /opt/newrelic_platform_plugins/s3-watcher-bundle-install.sh
+    mode: 777
```
benpatterson: Can you suggest a proper setting here?

playbooks/roles/newrelic/tasks/s3-watcher.yml
```diff
@@ -0,0 +1,38 @@
...
+- name: create agent configuration
+  template: >
+    src=opt/newrelic_platform_plugins/s3_watcher/config/newrelic_plugins.yml.j2
+    dest={{ newrelic_s3watcher_dir }}/config/newrelic_plugin.yml
+  notify: restart newrelic-logwatch-agent
+
+- name: set agent as executable
+  file: >
+    path={{ newrelic_s3watcher_dir }}/newrelic_s3_agent
+    mode=777
```
benpatterson: Same here

benpatterson:
Update: the common TODO is done. This is officially PR-worthy as a non-daemonized install.

CC @clytwynec for any feedback you can offer as well.

playbooks/roles/newrelic/defaults/main.yml
```diff
@@ -19,7 +19,7 @@ newrelic_role_name: newrelic
 NEWRELIC_DEBIAN_REPO: 'deb http://apt.newrelic.com/debian/ newrelic non-free'
 NEWRELIC_DEBIAN_KEY_ID: '548C16BF'
 NEWRELIC_DEBIAN_KEY_URL: 'https://download.newrelic.com/{{ NEWRELIC_DEBIAN_KEY_ID }}.gpg'
-NEWRELIC_LICENSE_KEY: "SPECIFY_KEY_HERE"
+NEWRELIC_LICENSE_KEY: "12345678901234567890ABCDEFGHIJABCDEFGHIJ"
```
clytwynec: This is ok to check in to the repo?

clytwynec: Looking closer, this looks like a dummy key?

benpatterson: Yeah, I was getting exceptions that it needed to be 40 characters long. However, newrelic won't start with a dummy key anyway. I'll remove this particular line change.

playbooks/roles/newrelic/tasks/s3-watcher.yml
clytwynec: I'm curious what @e0d suggests for this.   Other changes look ok, best I can tell.

e0d: 750

benpatterson:
@clytwynec @e0d @jzoldak 

Here's the new relic plugin -- daemonized. Thoughts?

Ed, everything is set to root as far as supervisor is concerned. Seemed odd to run it as common_web_user...I'm open to any suggestions!

benpatterson:
@e0d @clytwynec @jzoldak 

Perms are all tightened up on this.

I suggest we deploy with it set as this, and then, as a follow-on, we use IAM as Ed suggests. It will be a bit time-consuming for me to test right now (and---except for the hash being pulled---the changes would be in a different repo). It'll be easier to adjust to IAM once we have some stats being collected. 

My commit message will have text describing the vars that need to be set when running this role.

benpatterson:
@fredsmith @feanil If either of you has time to look over this today, I'd appreciate the review.

playbooks/roles/newrelic/defaults/main.yml
```diff
@@ -43,3 +43,15 @@ newrelic_logwatch_repo: https://github.com/railsware/newrelic_platform_plugins
 newrelic_logwatch_repo_dir: /opt/newrelic_platform_plugins
 newrelic_logwatch_dir: "{{ newrelic_logwatch_repo_dir }}/newrelic_logwatcher_agent"
 newrelic_logwatch_version: "8edd6d214e462b27fdd07d41712eb7b4fff2f7d8"
+
+# s3_watcher
+newrelic_plugin_user: "{{ common_web_user }}"
+newrelic_s3watcher_repo: https://github.com/benpatterson/analytics-gasket
```
jzoldak: Is the plan to move this under the edx org?

playbooks/roles/newrelic/defaults/main.yml
```diff
@@ -43,3 +43,15 @@ newrelic_logwatch_repo: https://github.com/railsware/newrelic_platform_plugins
...
+newrelic_s3watcher_repo_dir: /edx/app/s3watcher
+#newrelic_s3watcher_repo_dir: /opt/newrelic_platform_plugins/s3-newrelic-plugin
+newrelic_s3watcher_dir: "{{ newrelic_s3watcher_repo_dir }}/newrelic/s3-newrelic-plugin"
+newrelic_s3watcher_version: "e33cf920aa172707b788834e176350f423c7c38d"
```
jzoldak: I would move this up to right under the repo URL because they are related.

playbooks/roles/newrelic/handlers/main.yml
```diff
@@ -1,3 +1,7 @@
 ---
 - name: restart newrelic-logwatch-agent
   service: name=newrelic-logwatch-agent state=restarted
+
+# Inactive for now
```
jzoldak: What do you mean by "Inactive for now?"

playbooks/roles/newrelic/tasks/s3-watcher.yml
```diff
@@ -0,0 +1,77 @@
...
+    mode: 0755
+    owner: "{{ newrelic_plugin_user }}"
+  sudo_user: "{{ newrelic_plugin_user }}"
+
+- name: create supervisor config
+  template:
+    src: s3watcher.conf.j2
+    dest: "{{ supervisor_cfg_dir }}/s3watcher.conf"
+    owner: "{{ supervisor_user }}"
+#    owner: root
```
jzoldak: leftover from debugging?

jzoldak:
Reviewing for anything obvious, it lgtm :+1:

playbooks/roles/newrelic/defaults/main.yml
benpatterson: Yes; I'll put in a ticket for that

playbooks/roles/newrelic/defaults/main.yml
benpatterson: Done

playbooks/roles/newrelic/handlers/main.yml
benpatterson: This is left over from development. I've removed that section.

playbooks/roles/newrelic/tasks/s3-watcher.yml
benpatterson: Yep. Removed.

PR #1672, opened 2014-10-29
Feanil/vpc play

feanil:
@e0d @jarv

### DIFF
--- a/playbooks/edx-east/edx_vpc.yml
+++ b/playbooks/edx-east/edx_vpc.yml
@@ -21,9 +21,9 @@
         wait: yes
       register: created_vpc
 
-    # A public network acl is created
-    # when a vpc is created so each VPC
-    # should have one.
+    # A default network acl is created when a vpc is created so each VPC
+    # should have one but we create one here that allows access to the
+    # outside world using the internet gateway.
     - name: create public network acl
       ec2_acl:
         profile: "{{ profile }}"
--- a/playbooks/edx-east/edx_vpc.yml
+++ b/playbooks/edx-east/edx_vpc.yml
@@ -41,11 +41,7 @@
         region: "{{ aws_region }}"
         state: "present"
         name: "{{ vpc_name }}-public"
-        routes:
-          - cidr: "{{ vpc_cidr }}"
-            gateway: 'local'
-          - cidr: '0.0.0.0/0'
-            gateway: "{{ created_vpc.igw.id }}"
+        routes: "{{ vpc_public_route_table }}"
       register: created_public_rt
 
     - name: create public subnets
@@ -108,11 +104,7 @@
         region: "{{ aws_region }}"
         state: "present"
         name: "{{ vpc_name }}-private"
-        routes:
-          - cidr: "{{ vpc_cidr }}"
-            gateway: 'local'
-          - cidr: '0.0.0.0/0'
-            instance: "{{ created_nat_instance.instances[0].id }}"
+        routes: "{{ vpc_private_route_table }}"
       register: created_public_rt
       when: potential_existing_nat.instances|length == 0
       
--- a/playbooks/edx-east/edx_vpc.yml
+++ b/playbooks/edx-east/edx_vpc.yml
@@ -26,7 +26,7 @@
     # should have one.
     - name: create public network acl
       ec2_acl:
-#        profile: "{{ profile }}"
+        profile: "{{ profile }}"
         name: "{{ vpc_public_acl.name }}"
         vpc_id: "{{ created_vpc.vpc_id }}"
         state: "present"
--- a/playbooks/edx-east/edx_vpc.yml
+++ b/playbooks/edx-east/edx_vpc.yml
@@ -1,134 +1,154 @@
 ---
-- hosts: first_in_tag_role_mongo
-  sudo: True
-  vars_files:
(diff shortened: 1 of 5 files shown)


playbooks/edx-east/edx_vpc.yml
```diff
@@ -1,134 +1,154 @@
...
+        internet_gateway: yes
+        wait: yes
+      register: created_vpc
+
+    # A public network acl is created
+    # when a vpc is created so each VPC
+    # should have one.
+    - name: create public network acl
+      ec2_acl:
+#        profile: "{{ profile }}"
```
jarv: remove comment?

jarv:
:+1:

playbooks/edx-east/edx_vpc.yml
```diff
@@ -1,134 +1,154 @@
...
+        module: 'ec2_vpc_1.8'
+        resource_tags: '{{ vpc_tags }}'
+        cidr_block: '{{ vpc_cidr }}'
+        region: '{{ aws_region }}'
+        state: '{{ vpc_state }}'
+        internet_gateway: yes
+        wait: yes
+      register: created_vpc
+
+    # A public network acl is created
```
e0d: Do you mean that one is created by default so that this should be a NOOP?

playbooks/edx-east/edx_vpc.yml
```diff
@@ -1,134 +1,154 @@
...
+      register: created_public_acl
+
+    - name: create public route table
+      ec2_rt:
+        profile: "{{ profile }}"
+        vpc_id: "{{ created_vpc.vpc_id }}"
+        region: "{{ aws_region }}"
+        state: "present"
+        name: "{{ vpc_name }}-public"
+        routes:
```
e0d: I think I'd like this pulled out.  I realize that these routes are generic for the public route table, but we do add static routes is some places and it would be nice to be able to handle that case.

playbooks/edx-east/edx_vpc.yml
```diff
@@ -1,134 +1,154 @@
...
+            gateway: 'local'
+          - cidr: '0.0.0.0/0'
+            instance: "{{ created_nat_instance.instances[0].id }}"
+      register: created_public_rt
+      when: potential_existing_nat.instances|length == 0
+      
+
+#        
+
+#    - name: create public subnets
```
e0d: Remove?

playbooks/library/ec2_acl
```diff
@@ -60,7 +60,7 @@ except ImportError:
 
 from boto.exception import NoAuthHandlerFound
 
-PROTOCOL_NUMBERS = {"ICMP": 1, "TCP": 6, "UPD": 17}
+PROTOCOL_NUMBERS = {"ICMP": 1, "TCP": 6, "UDP": 17, "ALL": -1 }
```
e0d, line 63: You're not down with UPD?

playbooks/library/ec2_rt
```diff
@@ -58,14 +54,139 @@ try:
...
+            elif "gateway" in route:
+                call_args['gateway_id'] = route['gateway']
+            elif "instance" in route:
+                call_args['instance_id'] = route['instance']
+            else:
+                msg = "No gateway or instance provided for route with" + \
+                      "CIDR {}"
+                raise InconsistentRouteError(msg.format(route['cidr']))
+
+            if route['cidr'] in existing_routes:
```
e0d, line 143: Doesn't need to be solved here, but think we should have a pattern for dealing with "orphaned" routes.  My current thought for any of these sorts of things were multiple mechanisms are possibly creating entities is that the ansible modules should have an argument called, prune or canonical or something.  This would cause un-provided routes to be deleted.

playbooks/library/ec2_rt
```diff
@@ -58,14 +54,139 @@ try:
...
-    module.exit_json(id="-1")
+class DuplicateRouteTableError(Exception):
+    pass
+
+class InconsistentRouteError(Exception):
+    pass
+
+class RTManager():
+
+    def __init__(self, connection, vpc_id, route_name, routes, tags):
```
e0d, line 68: Now that boto has univeralized the dryrun named argument, I think we should try to thread that through via checkmode.

playbooks/edx-east/edx_vpc.yml
feanil: The comment should have been updated, there is a network ACL created but getting and changing it was not straight forward also it was the default one which in the past we've left alone.  So the approach I ended up going with is to create a new ACL that is explicitly public.  The Default ACL can only communicate within the VPC which I thought was a safe way to leave things in case we accidentally create infrastructure that uses the default ACL it doesn't immediately get exposed to the world.

playbooks/library/ec2_acl
feanil, line 63: Yea, you know me.

playbooks/library/ec2_rt
feanil, line 143: Yea, ansible is starting to do this with 'purge' we can try to follow that.

playbooks/library/ec2_rt
feanil, line 68: This is unfortunately not super straight forward since the commands throw an exception when called with 'dry_run=True' and we would need a lot more logic to deal with all the exceptions and make sure they are just dry_run exceptions.

feanil:
@e0d I've made the updates you asked for, the rest of the comments I'm not going to deal with unless you think I should before merging this bit onto the service builder branch.

e0d:
:+1:

PR #1676, opened 2014-10-31
Feanil/update datadog agent

feanil:
Update to deal with the deprecation of support for the older version of Datadog.


fredsmith:
:+1:

PR #1679, opened 2014-11-03
Addition of gitreload role for github Web hook course imports

carsongee:
It also supports updating repositories such as python graders
from xqwatcher/xserver.


openedx-webhooks:
Thanks for the pull request, @carsongee! I've created [OSPR-181](https://openedx.atlassian.net/browse/OSPR-181) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. Feel free to add information to the ticket which can help Product understand the context for the PR - supporting documentation, edx-code email threads, timeline information ('this must be merged by XX date', and why that is), partner information (this is for a course on edx.org, for example), etc. 

playbooks/roles/gitreload/defaults/main.yml
```diff
@@ -0,0 +1,45 @@
...
+# Defaults for role gitreload
+# 
+
+GITRELOAD_COURSE_CHECKOUT: false
+GITRELOAD_GIT_IDENTITY: !!null
+GITRELOAD_REPOS:
+  - name: "edx-demo-course"
+    url: "https://github.com/edx/edx-demo-course.git"
+    commit: "master"
+GITRELOAD_REPODIR: "{{ edxapp_data_dir }}/course_repos"
```
feanil: This will fail if the edxapp role is not included in the play that runs this.  Should edxapp be a dependency for this?

feanil: Is there a reason this needs to be under the edxapp_data dir at all? Why not put it in a `GITRELOAD_DATA_DIR`?

playbooks/roles/gitreload/tasks/course_pull.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: change group on repos if using devstack
+  file: >
+    path={{ GITRELOAD_REPODIR }}
+    state=directory
+    group={{ edxapp_user }}
+    recurse=yes
+  when: devstack
+
+- name: change mode on repos with using devstack
+  command: chmod -R o=rwX,g=srwX,o=rX {{ GITRELOAD_REPODIR }}
```
feanil, line 33: Why not use the `file` module for this?

playbooks/roles/gitreload/tasks/main.yml
```diff
@@ -0,0 +1,146 @@
+---
```
feanil, line 1: Can you move the tasks with deploy tags to a `tasks/deploy.yml` to be consistent with the other roles.

playbooks/roles/gitreload/tasks/course_pull.yml
carsongee, line 33: file module isn't smart enough to do capital X chmod and since I am setting the group sticky bit so this works whether you run it as edxapp or www-data, group id + unnecessary execute permissions with the file module would be bad for least privilege.

playbooks/roles/gitreload/defaults/main.yml
carsongee: That is because that folder is set in edx-platform settings file so this syncs it with the default value there.  I'll go ahead define that setting in the edxapp role and reference it here instead of using the default.  The edxapp dependency isn't in meta though because this app gets used in different ways, and only one way is dependent on edxapp.  The first is importing courses into platform, and the other is just updating repos (we use it for xserver graders for example, but there are other utility uses).  To solve that I include edxapp role vars in tasks: https://github.com/edx/configuration/pull/1679/files#diff-036ed9abea637f04ba8f9b5665ded842R35

carsongee:
I moved deploy tasks to a deploy.yml and think I have addressed your other concerns thus far.  I also moved it to using a gunicorn configuration file since I wrote this before that change.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -119,6 +119,8 @@ EDXAPP_THIRD_PARTY_AUTH: {}
 EDXAPP_MODULESTORE_MAPPINGS:
     'preview\.': 'draft-preferred'
 
+EDXAPP_GIT_REPO_DIR: '/edx/var/edxapp/course_repos'
```
feanil, line 122: Is there a reason that this needs to be under `/edx/var/edxapp` instead of `/edx/var/gitreload`?

feanil:
Can you also add a change log about the new role and its purpose.

playbooks/roles/edxapp/defaults/main.yml
carsongee, line 122: It is the default in platform: https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L251 and since that is a directory that the platform writes to and reads from in order to do imports of the courses, it made sense to leave it there.  i.e. the sysadmin dashboard git course import and management command use this directory regardless of gitreload being installed or not, so defaulting it somewhere else means adding gitreload as a dependency of edx-platform.

feanil:
:+1:

carsongee:
Do you want two reviews, or are we good to merge?

feanil:
I think this is good to merge.

carsongee:
Great, thanks.

PR #1680, opened 2014-11-03
E0d/maintenance site

e0d:
@feanil @jarv my branch is against release as I was testing in sandboxes, but this is what I'm thinking of for the maintenance config.

### DIFF
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -62,6 +62,13 @@ nginx_debian_pkgs:
   - nginx
   - python-passlib
 
+NGINX_EDXAPP_ENABLE_S3_MAINTENANCE: False
+nginx_default_error_page: "/server/server-error.html"
+NGINX_EDXAPP_ERROR_PAGES:
+  "500": "{{ nginx_default_error_page}}"
+  "502": "{{ nginx_default_error_page}}"  
+  "504": "{{ nginx_default_error_page}}"
+  
 CMS_HOSTNAME: '~^((stage|prod)-)?studio.*'
 
 nginx_template_dir: "edx/app/nginx/sites-available"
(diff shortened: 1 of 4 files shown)


jarv:
:+1:

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -62,6 +62,13 @@ nginx_debian_pkgs:
   - nginx
   - python-passlib
 
+NGINX_EDXAPP_ENABLE_S3_MAINTENANCE: False
+nginx_default_error_page: "/server/server-error.html"
```
feanil, line 66: How does this file get generated? is it by gathering assets in the edx-platform?  Is there a PR for that yet?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
```diff
@@ -13,8 +13,22 @@ upstream cms-backend {
 server {
   # CMS configuration file for nginx, templated by ansible
       
-  # 500 error pages
-  error_page 500 502 504 /server/server-error.html;
+  # Proxy to a remote maintanence page
+  {% if NGINX_EDXAPP_ENABLE_S3_MAINTENANCE %}
+
+  # Do not include a 502 error in NGINX_ERROR_PAGES when
```
feanil, line 19: Do you mean that 502s should also point to the maintenance page when the maintenance flag is enabled?

e0d, line 19: Yes, the backend server will be shutdown, which will cause a 502 from nginx's perspective.  The flag doesn't enable maintenance mode alone.  It installs the location, but will only server from it when the lms and/or cms are unavailable.

playbooks/roles/nginx/defaults/main.yml
e0d, line 66: I've just made this a var where it was hard-coded previously, but the file is part of edx-platform.

PR #1684, opened 2014-11-05
Harprofiler repo is now under edx. Disable test cruft.

benpatterson:
The test was under dev and shouldn't have been merged. (This
approach is incompatible with ansible's design.)

cc @cgoldberg

### DIFF
--- a/playbooks/roles/harprofiler/defaults/main.yml
+++ b/playbooks/roles/harprofiler/defaults/main.yml
@@ -7,7 +7,7 @@
 harprofiler_role_name: harprofiler
 
 harprofiler_user: "harprofiler"
-harprofiler_github_url: https://github.com/cgoldberg/harprofiler
-harprofiler_version: 2958a910cc0c33967f7d7faf4519fa03271fdaf1
+harprofiler_github_url: https://github.com/edx/harprofiler
+harprofiler_version: master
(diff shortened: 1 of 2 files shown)


clytwynec:
Why don't we want the test?

benpatterson:
The test never worked, and ansible does not currently support sourcing in the context of a shell command (which is what we would need to use for a venv --- ansible uses a session-less user). I can't find the perfect link to describe this, but here's a concise one: https://groups.google.com/forum/#!searchin/ansible-project/bashrc/ansible-project/44Lb72cPMmg/Acton7JTJbcJ

clytwynec:
Ahh ok. Thanks.  I was just curious what the reason was.  

Changes look good. :+1:

cgoldberg:
FWIW, you must be in a bash shell to use 'source'.  you can't source in a
dash/bourne/etc shell.
can ansible run things under bash explicitly?

benpatterson:
Best, recent example of accomplishing this is using the rbenv on the new relic plugin. That was done with separate shell scripts (basically, one for each command that would need to be independently run under a given sourced environment).

https://github.com/edx/configuration/blob/master/playbooks/roles/newrelic/tasks/s3-watcher.yml#L34

PR #1686, opened 2014-11-06
Revert "Compiling Insights Translations"

clintonb:
Reverts edx/configuration#1685

Build failed and I have no logs to determine why.

@e0d merge this if we need to build an AMI before I return.


clintonb:
Reverting so that we can continue testing other work.

PR #1687, opened 2014-11-06
harprofiler ami creation via packer.

benpatterson:
TE-583

### DIFF
--- a/playbooks/roles/harprofiler/tasks/main.yml
+++ b/playbooks/roles/harprofiler/tasks/main.yml
@@ -38,3 +38,17 @@
     regexp="browsermob_dir"
     line="browsermob_dir: /usr/local"
     state=present
+
+- name: create validation shell script
+  template:
+    owner: "{{ harprofiler_user }}"
+    src: validate_harprofiler_install.sh.j2
+    dest: "{{ harprofiler_dir }}/{{ harprofiler_validation_script }}"
+    mode: 0755
+  sudo_user: "{{ harprofiler_user }}"
+
+
+- name: test install
+  shell: >
+    ./{{ harprofiler_validation_script }} chdir={{ harprofiler_dir }}
+  sudo_user: "{{ harprofiler_user }}"
--- a/playbooks/roles/harprofiler/templates/validate_harprofiler_install.sh.j2
+++ b/playbooks/roles/harprofiler/templates/validate_harprofiler_install.sh.j2
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+
+# This script confirms that harprofiler can successfully run on the
+# target machine.
+source {{ harprofiler_venv_dir }}/bin/activate
+cd {{ harprofiler_dir }}
+python harprofiler.py
+
--- a/util/packer/harprofiler.json
+++ b/util/packer/harprofiler.json
@@ -1,0 +1,38 @@
+{
+  "variables": {
+    "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
+    "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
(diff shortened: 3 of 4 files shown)


benpatterson:
@jzoldak @cgoldberg @clytwynec

playbooks/roles/harprofiler/tasks/main.yml
```diff
@@ -38,3 +38,17 @@
     regexp="browsermob_dir"
     line="browsermob_dir: /usr/local"
     state=present
+
+- name: create validation shell script
+  template:
+    owner: "{{ harprofiler_user }}"
+    src: validate_harprofiler_install.sh.j2
```
jzoldak, line 45: where is this template?

util/packer/harprofiler.json
```diff
@@ -0,0 +1,39 @@
...
+    "playbook_remote_dir": "/tmp/packer-edx-playbooks",
+    "ami": "{{env `HARPROFILER_BASE_AMI`}}"
+  },
+  "builders": [{
+    "type": "amazon-ebs",
+    "access_key": "{{user `aws_access_key`}}",
+    "secret_key": "{{user `aws_secret_key`}}",
+    "region": "us-east-1",
+    "source_ami": "{{user `ami`}}",
+    "instance_type": "m3.large",
```
jzoldak: I thought that the harprofiler has a pretty small footprint, can we use a micro?

playbooks/roles/harprofiler/tasks/main.yml
benpatterson, line 45: I hadn't checked it in. Fixed.

util/packer/harprofiler.json
benpatterson: Can't use a micro, but we can use an m1.small. FWIW, this is just the size of machine for doing the AMI building itself. Once we have the AMI, we can put it on whatever class of machine we'd want.

playbooks/roles/harprofiler/templates/validate_harprofiler_install.sh.j2
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+
+# This script confirms that harprofiler can successfully run on the
+# target machine.
+source {{ harprofiler_venv_dir }}/bin/activate
+cd {{ harprofiler_dir }}
```
clytwynec, line 7: Why is this line needed if `chdir={{ harprofiler_dir }}` is used?

benpatterson, line 7: I left it in there in case someone ends up calling the shell not-through ansible...

util/packer/harprofiler.json
```diff
@@ -0,0 +1,39 @@
...
+    "type": "file",
+    "source": "../../requirements.txt",
+    "destination": "{{user `playbook_remote_dir`}}/requirements.txt"
+  }, {
+    "type": "shell",
+    "inline": ["cd {{user `playbook_remote_dir`}}",
+      "virtualenv packer-venv",
+      ". packer-venv/bin/activate",
+      "pip install -q -r requirements.txt",
+      "echo '[jenkins_worker]' > inventory.ini",
```
clytwynec: Should this really say 'jenkins_worker' here?

playbooks/roles/harprofiler/templates/validate_harprofiler_install.sh.j2
clytwynec, line 7: Ok. Makes sense.

util/packer/harprofiler.json
benpatterson: Good catch. Fixed it.

clytwynec:
:+1:

PR #1688, opened 2014-11-06
Hadoop configuration proof of concept (DO NOT MERGE)

mulby:
@e0d here is the ansible code we discussed. I wrote this a long time ago, really helped ramp me up on ansible.

Notable stuff in here:
- I was always testing this on a set of dedicated vagrant VMs. So it does scary stuff like clobber the /etc/hosts file.
- I don't remember what version of ansible I was running this stuff with... I think 1.4.4?
- There are some defaults for variables that look secure. They aren't and are not actually in use anywhere.
- I had half-extracted the security stuff, I don't think this is in a working state right now. It could also be vastly simplified if we removed all of the stuff related to kerberos and the "cdh-secure" tag.
- The "alternatives" module I actually got merged into ansible trunk a while ago, so we wouldn't actually need the custom module in the library.
- The way this is currently setup you can really only deploy a single hadoop cluster. The inventory isn't smart enough to distinguish between clusters.
- I have a bunch of other setup code for Hive etc, but this gives you the general idea.
- This is all using a fairly old version of cloudera hadoop.

### DIFF
--- a/playbooks/authorize-user.yml
+++ b/playbooks/authorize-user.yml
@@ -1,0 +1,20 @@
+---
+
+- hosts: kerberos_server
+  sudo: True
+  vars:
+    - username: ubuntu
+  tasks:
+    - name: principal created
+      kerberos_principal: name={{ username }}
+
+    - name: keytab created
+      command: /usr/sbin/kadmin.local -q 'xst -norandkey -k /home/{{ username }}/credentials.keytab {{ username }}'
+
+    - name: keytab permissions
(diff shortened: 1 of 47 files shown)


e0d:
Seems generally reasonsble and worth refining.  I do have a bunch of specific questions, but I think some of them are things you already know about.  Without having your knowledge of the toolset, it's a little difficult to get my head around everything that's happening here.

mulby:
Closing for now, can use as future reference.

PR #1696, opened 2014-11-11
Add roles

e0d:
@feanil I'm hopeful this will fix the Mongo issues that @wedaly was seeing, see hipchat.  Will try to test from here and see how it goes.

### DIFF
--- a/playbooks/roles/mongo/tasks/main.yml
+++ b/playbooks/roles/mongo/tasks/main.yml
@@ -1,18 +1,23 @@
-
 ---
-- name: check to see that MongoDB 2.4 isn't installed
+- name: check to see that MongoDB 2.4 is not installed
   stat: path=/etc/init.d/mongodb
   register: mongodb_needs_upgrade
 
 - name: verify 2.4 not installed
-  fail: msg="MongoDB 2.4 is currently installed. If on a stand alone host (devstack), apt-get remove mongodb-10gen and re-run ansible. if on a cluster, read http://docs.mongodb.org/manual/release-notes/2.6-upgrade/#upgrade-considerations and upgrade to 2.6."
-  when: mongodb_needs_upgrade.stat.exists 
-
-
+  fail: msg="MongoDB 2.4 is currently installed and cannot be safely upgraded in a clustered configuration.  Please read http://docs.mongodb.org/manual/release-notes/2.6-upgrade/#upgrade-considerations and upgrade to 2.6."
+  when: mongodb_needs_upgrade.stat.exists and MONGO_CLUSTERED
+  
+- name: remove mongo 2.4 if present
+  apt: >
+    pkg=mongodb-10gen
+    state=absent purge=yes
+    force=yes
+  when: mongodb_needs_upgrade.stat.exists and not MONGO_CLUSTERED
+  
 - name: install python pymongo for mongo_user ansible module
   pip: >
     name=pymongo state=present
-    version=2.6.3 extra_args="-i {{ COMMON_PYPI_MIRROR_URL }}"
+    version={{ pymongo_version }} extra_args="-i {{ COMMON_PYPI_MIRROR_URL }}"
 
 - name: add the mongodb signing key
   apt_key: >
@@ -46,8 +51,9 @@
   service: name=mongod state=stopped
 
 - name: move mongodb to {{ mongo_data_dir }}
-  command: mv /var/lib/mongodb  {{ mongo_data_dir}}/.  creates={{ mongo_data_dir }}/mongodb
-
+  command: >
+    mv /var/lib/mongodb  {{ mongo_data_dir}}/.
+    creates={{ mongo_data_dir }}/mongodb
 
 - name: copy mongodb key file
   copy: >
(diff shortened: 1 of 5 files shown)


e0d:
This is not sufficient to fix the sandbox issues, though maybe still useful.  Although, the db name does seem to be getting passed to the module and to pymongo, it looks like the created users aren't being properly created or associated with their db. 

Was wondering if it was inttentional to use the system pymongo?

e0d:
@carsongee @feanil @dmitchell 

This is a possible alternative to calling JS directly, though I admit it's not a clear winner.  It has the advantage of using the module for normal users.  I needed to import the devel version until we can upgrade safely, but once we do our version can be deleted.

It has the disadvantage of needing to call mongo directly to create a superuser with a non empty password in the bootstrap case.  Pymongo won't allow 0 length passwords anymore, which is one of our issues.

The mongo_user module falls back to using a ~/.mongodb.cnf when login_user and login_password are not provided.  I think providing them makes sense in general, but given the pasword length constraint, we need to shell out to create an admin user with a password.  Using the cnf file might allow us to workaround the constraint, but I'd prefer not to enable bad practice.

I brought some of Carson's changes along, specifically upgrading the mongodb version and removing the create_user condition.

I also upgraded pymogo which was hard code to an older version for reasons I don't know.  It's worth noting that our config tasks are currently installing pymongo system-wide.

I plan to merge either this or https://github.com/edx/configuration/pull/1701 today, let me know if you see any reason to prefer one or the other.  Frankly, I don't love either.

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -1,18 +1,20 @@
 
 ---
-- name: check to see that MongoDB 2.4 isn't installed
+- name: check to see that MongoDB 2.4 is not installed
   stat: path=/etc/init.d/mongodb
   register: mongodb_needs_upgrade
 
-- name: verify 2.4 not installed
```
fredsmith: We were doing this this way so that we didn't lose any data or mess up anyone's clusters by doing an unannounced upgrade.  Should we put this behind a flag?

feanil: +1 to putting this behind a flag and still failing out if we are running on a cluster(If clustured flag is set.)

feanil:
Other than @fredsmith 's comment :+1:

playbooks/library/mongodb_user_1.8
```diff
@@ -0,0 +1,252 @@
+#!/usr/bin/python
```
dmitchell: Are you incl this b/c you made changes to it? If so, can you check in the dist version as a PR before this so this PR only shows your edits? Will you submit your edits back to ansible?

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -76,16 +78,28 @@
...
 
+# Need to bootstrap the creation of an admin user
+# By default, root can login from the host with an empty
+# password.  However, pymongo does not allow 0 lenth
+# passords on auth calls.  The ansible module falls
+# back to using, ~/.mongodb.cnf.  This will neeed to be
+# created and  contain a user with a non-empty password
+- name: create admin user
+  shell: >
+    mongo --eval "admin=db.getSiblingDB('admin'); if (admin.getUser('{{ MONGO_ADMIN_USER }}') == null) { admin.createUser( { user: '{{ MONGO_ADMIN_USER }}', pwd: '{{ MONGO_ADMIN_PASSWORD }}', roles: ['root']} )}"
```
carsongee: Can you move this into a js template so that we don't have super user passwords in history?

carsongee:
have you tested this with replica sets?  I think the admin creation step will cause failures without first checking if you are master. I can later this afternoon if you can't, but the vagrant cluster play should excerise this

playbooks/roles/mongo/tasks/main.yml
carsongee: I think you'll still will want to make the changes I did in `repset_init` in #1701 for this, and then you can just do `when: not MONGO_CLUSTERED` the super user creation here.

playbooks/roles/mongo/templates/create_root.js.j2
```diff
@@ -0,0 +1,21 @@
+// Add super user
+conn = new Mongo();
+db = conn.getDB("admin");
+db.auth( '{{ MONGO_ADMIN_USER }}', '{{ MONGO_ADMIN_PASSWORD }}');
+if(db.getUser('{{ MONGO_ADMIN_USER }}') == null) {
+db.createUser(
```
carsongee: indentation would be nice

playbooks/roles/mongo/templates/repset_init.js.j2
```diff
@@ -47,4 +47,4 @@ if(rs.isMaster().ismaster) {
             throw 'Could not add all members to cluster'
         }
     }
-}
+}
```
carsongee: This is missing the user creation stuff in #1701 from line 54 down.

e0d:
@carsongee whitespace and user creation restored.  Works for the sandbox case, if you an easy mechanism for testing a cluster that would be helpful, or let me know what you typically do.

carsongee:
just vagrant up in `vagrant/base/cluster` will stand up three machines and build them.  For quicker test you can comment out everything but the mongo role

e0d:
@carsongee I've added it back and things seems happy.  Part of my problem is definitely memory related, swapping like crazy with my meager 8G of memory.  Let me know your disposition on this, want to merge to build new dev stack boxes for the hack-a-thon.

carsongee:
I'm all for merging it as it is now. It is building great and woth multiple runs.

jarv:
:+1:

PR #1698, opened 2014-11-12
Work around for an issue in the Route53 module.

feanil:
See: https://github.com/ansible/ansible-modules-core/pull/310


e0d:
:+1:

PR #1700, opened 2014-11-13
Compiling Insights Translations

clintonb:
The virtualenv must be active in order for i18n_tool to run properly.

@e0d @feanil

### DIFF
--- a/playbooks/roles/insights/tasks/deploy.yml
+++ b/playbooks/roles/insights/tasks/deploy.yml
@@ -80,6 +80,12 @@
     - "collectstatic --noinput"
     - "compress"
 
+- name: compile translations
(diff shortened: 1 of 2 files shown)


playbooks/roles/insights/tasks/deploy.yml
```diff
@@ -1,5 +1,8 @@
 ---
 
+- name: install gettext
```
feanil: This shouldn't be part of the deploy task, it should be in the main.yml along with any other dependencies that are being installed.

clintonb: Fixed.

feanil:
:+1:

PR #1701, opened 2014-11-13
Replace mongodb_user module with javascripts

carsongee:
This add fullsupport of new authentication model in mongodb 2.6.

I know this isn't the most elegant fix, but using the latest module in ansible didn't work, and I was having issues modifying to work properly so I've fallen back to stripping that out and just using mongo javascripts.

I have tested this with the cluster build and devstack (including a "from scratch" on mongodb 2.6 devstack as well as an upgraded version.

One of the upshots is that it will actually work on a cluster and not fail on the slaves since the scripts check if they are master before attempting to modify anything so I removed the "create_users" flag and just run that each time.

### DIFF
--- a/playbooks/roles/mongo/defaults/main.yml
+++ b/playbooks/roles/mongo/defaults/main.yml
@@ -1,5 +1,5 @@
 mongo_logappend: true
-mongo_version: 2.6.4
+mongo_version: 2.6.5
 mongo_port: "27017"
 mongo_extra_conf: ''
 mongo_key_file: '/etc/mongodb_key'
@@ -14,13 +14,19 @@ MONGODB_APT_KEY: "http://docs.mongodb.org/10gen-gpg-key.asc"
 MONGODB_REPO: "deb http://downloads-distro.mongodb.org/repo/ubuntu-upstart dist 10gen"
 
 # Vars Meant to be overridden
+MONGO_USER_SUPER: 'admin'
+MONGO_PASSWORD_SUPER: 'password'
 MONGO_USERS:
   - user: cs_comments_service
     password: password
     database: cs_comments_service
+    roles:
+      - dbOwner
   - user: edxapp
     password: password
     database: edxapp
+    roles:
+      - dbOwner
 
 MONGO_CLUSTERED: !!null
 MONGO_BIND_IP: 127.0.0.1
@@ -29,10 +35,6 @@ MONGO_BIND_IP: 127.0.0.1
 
 mongo_logpath: "{{ mongo_log_dir }}/mongodb.log"
 mongo_dbpath: "{{ mongo_data_dir }}/mongodb"
-
-# Have to use this conditional instead of ignore errors
-# because the mongo_user module fails and doesn't ignore errors.
-mongo_create_users: true
 
 # If the system is running out of an Amazon Web Services
 # cloudformation stack, this group name can used to pull out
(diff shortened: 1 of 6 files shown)


openedx-webhooks:
Thanks for the pull request, @carsongee! I've created [OSPR-200](https://openedx.atlassian.net/browse/OSPR-200) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

e0d:
@carsongee I've been planning to look into this, can you characterize what the issues with the module are?

carsongee:
Updating to the newest module (as of three weeks ago), and the latest pymongo, it would just fail to create the users (at least in the cluster/replicaset mode).  The craziest thing was when I attached a remote debugger (epdb) and stepped through line by line, it had no problem but would always fail during a real play run.  There were a couple other issues with it that I can't remember right now, but it seemed like a lot less effort to use the dropped javascript files instead of digging in further.  

There were also a couple issues in the ansible repo before they split the modules referencing some of the issues that got closed and haven't been ported, so maybe there is a combination of 1.5 and the module that are causing the issue, notably https://github.com/ansible/ansible/issues/7157.

All that said, at this point it could have been something strange in my setup alone, since it worked in the debugger.  Some extra context from when I started looking at this a couple weeks ago: https://github.com/edx/configuration/pull/1637

e0d:
It also isn't working in a similar way in our single-instance-sandboxes, though we don't get an exception.  In fact, users are created if I log on to mongo as root from the box.  The user's are associated with the correct database.  I had debugged as well and to the point I got, things looked likel they were being done correctly.  We've been using a crappy temporary workaround, https://gist.github.com/e0d/4ce5f0de1d1fac459e75, but I had hoped to fix.  It's possible our issues is unrelated.  Do you have the error you see handy?

carsongee:
The specific error was just in the cluster case as the single instance would just silently not work.  It says "not master and slaveOk != true" even on the master node with the latest version that theoretically supports replicaSets. This branch fixes it for all types of mongo installs and is idempotent so I'd rather not have to put too much effort into fixing the module at this point, unless you have strong feelings about it.

e0d:
Not asking you to fix the module, I'm happy to do it, but wanted to collect what we know about the issue.

dmitchell:
:+1: the ansible script is way out of date. `authenticate` raises `PyMongoError` if the user/pswd is invalid (since pymongo 2.5 I believe). The ansible script doesn't catch this when checking if the user already exists. The ansible script uses undocumented `admin` methods.

e0d:
@dmitchell Unless I'm missing something OperationFailure is what is raised in this case, with the 2.7.x driver in any case.

```
[e0d-sandbox] e0d@e0d i-e6741c0c:~$ pip freeze | grep pymongo
pymongo==2.7.2

In [5]: from pymongo import MongoClient

In [6]: inspect.getfile(MongoClient)
Out[6]: '/usr/local/lib/python2.7/dist-packages/pymongo/mongo_client.pyc'

In [7]: c = MongoClient("localhost",27017)

In [8]: c['edxapp'].authenticate("no-such-user","password")

... snip ...

OperationFailure: command SON([('authenticate', 1), ('user', u'no-such-user'), ('nonce', u'4818224770da08f7'), ('key', u'0f6a97df250ef7aa473583b657eeeeda')]) failed: auth failed
```

dmitchell:
Hmm, I suppose empirical evidence trumps documentation. The doc string on `Database.authenticate` says

```
        .. versionchanged:: 2.5
           Added the `source` and `mechanism` parameters. :meth:`authenticate`
           now raises a subclass of :class:`~pymongo.errors.PyMongoError` if
           authentication fails due to invalid credentials or configuration
           issues.
```

However, now that I check `OperationFailure` is a subclass of `PyMongoError`; so, it is consistent. Now I'm just wondering if the other auth mechanisms also raise `OperationFailure`.

carsongee:
No longer needed thanks to #1696

PR #1702, opened 2014-11-13
Added Support for Django 1.7 for Analytics API and Insights

clintonb:
@e0d @feanil

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+- Role: analytics-api, insights
+  - Using Django 1.7 migrate command.
+
 - Role: edxapp
   - A new var was added to make it easy ot invalidate the default
     memcache store to make it easier to invalidate sessions. Updating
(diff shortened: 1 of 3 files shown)


e0d:
Do we need to support backward compatibility?  If so, I'd prefer to have separate tasks with conditionals.  This is simple and terse, but maybe too so.

clintonb:
@e0d We only need backward compatibility in the event that we do something admittedly silly like merge this configuration before actually upgrading to Django 1.7. My plan was to remove the Django 1.6 command once we were certain we would stick with Django 1.7. I'm fine with removing that now since the odds of us getting to a point of mixed versions is pretty low.

clintonb:
@e0d Updated.

e0d:
@jbau you may already be aware that this is coming. 

@clintonb we should add change log entries for this as well.

clintonb:
@e0d Changelog updated.

clintonb:
@e0d Are we good to merge?

e0d:
:+1:

PR #1705, opened 2014-11-17
Put all keys in the authorized_keys file.

feanil:


e0d:
Seems reasonable, what's the motivation?

jarv:
no strong preference here

feanil:
authorized_keys2 was not getting picked up on ec2 linux.

benpatterson:
Just noticed this, @feanil . If we run a second user with this playbook, would that erase any previous authorized_keys?

feanil:
Re-running this play for the same user multiple times will replace their authorized_keys file each time.  Running it for one user will not affect the authorized keys for a different user.

PR #1707, opened 2014-11-18
Make s3sync error message uniform.

benpatterson:
Nitpicky but it'll help for any case-sensitive log searches.

@e0d @mulby


benpatterson:
@e0d @mulby this is a small thing so I'm just going to merge it.

PR #1709, opened 2014-11-19
Allow s3sync log dir to be seen by Splunk.

benpatterson:
This will match perms for things like rabbitmq's logs, which are being indexed successfully.

@e0d @mulby @brianhw 

If you open a sandbox and look at the dirs, you can also see the perms difference.

### DIFF
--- a/playbooks/roles/aws/defaults/main.yml
+++ b/playbooks/roles/aws/defaults/main.yml
@@ -39,7 +39,7 @@ aws_dirs:
     path: "{{ COMMON_LOG_DIR }}/{{ aws_role_name }}"
     owner: "syslog"
     group: "syslog"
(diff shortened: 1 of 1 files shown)


benpatterson:
@e0d I've set permissions to 650 on my sandbox, but it still doesn't appear to index the log. (640/644 would not allow access to the splunk user, but 650 does.)

This has been a timeconsuming problem and if I/we can't confirm it's fixed in a sandbox, I'd like to merge it today at 755. That matches all the other logs in that directory. We can tighten restrictions on those logs in a separate effort.

benpatterson:
I should also point out that the underlying log file has the tighter perms already. This is the dir that holds the log.

e0d:
:+1:

PR #1716, opened 2014-11-24
[WIP] Use pip-accel for building jenkins workers

jzoldak:
@benpatterson @fredsmith FYI

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/packer.yml
+++ b/playbooks/roles/jenkins_worker/tasks/packer.yml
@@ -1,6 +1,8 @@
 ---
 - name: Download packer
   shell: "curl -L {{ packer_url }} -o /var/tmp/packer.zip"
+  args:
+      creates: /var/tmp/packer.zip
 
 - name: Unzip packer
   unarchive: src=/var/tmp/packer.zip dest=/usr/local/bin copy=no
--- a/playbooks/roles/jenkins_worker/tasks/python.yml
+++ b/playbooks/roles/jenkins_worker/tasks/python.yml
@@ -30,12 +30,24 @@
     depth=1
   sudo_user: "{{ jenkins_user }}"
 
-- name: Install edx-platform requirements
+- name: Pip installs that are needed for pip-accel to work for us
   pip: >
-    requirements={{ jenkins_home }}/shallow-clone/requirements/edx/{{ item }}
-    extra_args="--exists-action w"
+    name="{{ item }}"
     virtualenv={{ jenkins_home }}/edx-venv
     virtualenv_command=virtualenv-2.7
+  sudo_user: "{{ jenkins_user }}"
+  with_items: pip_accel_reqs
+
+# Install the platform requirements using pip-accel.
+# This will allow the binary distributions to be downloaded from S3
+# rather than compiled each time.
+- name: Install edx-platform requirements using pip-accel
+  pip: >
+    requirements={{ jenkins_home }}/shallow-clone/requirements/edx/{{ item }}
+    extra_args="--exists-action=w"
+    virtualenv={{ jenkins_home }}/edx-venv
+    virtualenv_command=virtualenv-2.7
+    executable=pip-accel
   with_items:
     - pre.txt
     - github.txt
--- a/playbooks/roles/jenkins_worker/tasks/system.yml
+++ b/playbooks/roles/jenkins_worker/tasks/system.yml
@@ -57,3 +57,13 @@
   template:
     src=.npmrc.j2 dest={{ jenkins_home }}/.npmrc
     owner={{ jenkins_user }} group={{ jenkins_group }} mode=0664
+
+# Set up configuration for pip-accel for caching python requirements
+- name: Create directory for pip-accel config file
+  file: path={{ jenkins_home }}/.pip-accel state=directory
+    owner={{ jenkins_user }} group={{ jenkins_group }} mode=0777 recurse=yes
+
+- name: Create pip-accel config file
+  template:
+    src=pip-accel.conf.j2 dest={{ jenkins_home }}/.pip-accel/pip-accel.conf
+    owner={{ jenkins_user }} group={{ jenkins_group }} mode=0664
--- a/playbooks/roles/jenkins_worker/templates/pip-accel.conf.j2
+++ b/playbooks/roles/jenkins_worker/templates/pip-accel.conf.j2
@@ -1,0 +1,7 @@
+[pip-accel]
+auto-install = no
+data-directory = ~/.pip-accel
+download-cache = ~/.pip/download-cache
+s3-bucket = edx-platform.pip-accel-cache
+s3-prefix = precise64
+s3-readonly = no
--- a/util/packer/README.rst
+++ b/util/packer/README.rst
@@ -1,0 +1,21 @@
+Packer
+=======
+
+``jenkins_worker.json`` is the `packer configuration template`_ that tells packer how to build the image.
+
+- `template variable`_ pattern ``"foo": "{{env `BAR`}}"``
+
+  - What this does is take the value of the environment variable BAR which was set in the shell that kicks off the ``packer build jenkins_worker.json`` command (for example in a jenkins job) and pass it through to the user variable "foo".
+  - This the user variable "foo" will now be available globally within the template.
+  - If the environment variable is not set in the shell that kicks off the packer build command, the user variable value will be the empty string.
+
+- Regarding the `ansible-playbook command`_ that is used to run the jenkins_worker role's -e (--extra-vars) option
+
+  - `playbook variable`_ pattern ``-e 'bar={{ user `foo` }}'``
+  - Packer has determined the value of the template user variable "foo" from the local environment variable (see above)
+  - Ansible will use this as the value for the playbook variable "bar" when running the play.
+
+.. _packer configuration template: http://www.packer.io/docs/templates/introduction.html
+.. _template variable: http://www.packer.io/docs/templates/user-variables.html
+.. _ansible-playbook command: http://docs.ansible.com/playbooks_intro.html#executing-a-playbook
+.. _playbook variable: http://docs.ansible.com/playbooks_variables.html#passing-variables-on-the-command-line
--- a/util/packer/jenkins_worker.json
+++ b/util/packer/jenkins_worker.json
@@ -10,11 +10,14 @@
     "type": "amazon-ebs",
     "access_key": "{{user `aws_access_key`}}",
     "secret_key": "{{user `aws_secret_key`}}",
+    "ami_name": "jenkins_worker-{{isotime | clean_ami_name}}",
+    "instance_type": "m3.large",
     "region": "us-east-1",
     "source_ami": "{{user `ami`}}",
-    "instance_type": "m3.large",
     "ssh_username": "ubuntu",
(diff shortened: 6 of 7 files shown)


playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,18 +30,50 @@
...
+# rather than compiled each time.
+- name: Install pip-accel for caching python requirements
+  pip: >
+    name='git+https://github.com/paylogic/pip-accel.git@0.18#egg=pip-accel'
+    virtualenv={{ jenkins_home }}/edx-venv
+    virtualenv_command=virtualenv-2.7
+  sudo_user: "{{ jenkins_user }}"
+
+# pip-accel only makes the s3 functionality available
+# if you have installed boto yourself.
```
benpatterson: Is this because you need the .boto file? Perhaps we just add the .boto file and install boto via the usual requirements? Would that work?

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,18 +30,50 @@
...
+- name: Install boto for caching python requirements
+  pip: >
+    name='boto' version='2.34.0'
```
benpatterson: FWIW this version number is different than platform right now. I think when we run the installs on platform we would need to use the wipe setting if we don't already.

util/packer/README.rst
```diff
@@ -0,0 +1,29 @@
+Packer
```
benpatterson, line 1: Thanks a lot for adding doc.

util/packer/jenkins_worker.json
```diff
@@ -35,7 +41,7 @@
       "pip install -q -r requirements.txt",
       "echo '[jenkins_worker]' > inventory.ini",
       "echo 'localhost' >> inventory.ini",
-      "ansible-playbook edx-east/jenkins_worker.yml -i inventory.ini -c local -e 'github_oauth_token={{user `github_oauth_token` }}' -vvvv"]
+      "ansible-playbook edx-east/jenkins_worker.yml -i inventory.ini -c local -e '@extra_vars.json' -vvvv"]
```
benpatterson: I like this particular change as well.

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak: No, you need more than just the .boto file. You need to have boto installed in the virtualenv for pip-accel to take advantage of the s3 cache, and the maintainer of pip-accel didn't want to include boto as a requirement. So you either need to install it yourself, or you need to install pip-accel like this: `pip install pip-accel[s3]`

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak: ah, good one. I will need to figure out what to pass through to pip-accel as an extra arg then, I had briefly tried it but it was giving some error message that I didn't feel like tracking down at that time. :(

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,18 +30,50 @@
     depth=1
   sudo_user: "{{ jenkins_user }}"
 
-- name: Install edx-platform requirements
+# Install the platform requirements using pip-accel.
+# This will allow the binary distributions to be downloaded from S3
+# rather than compiled each time.
+- name: Install pip-accel for caching python requirements
+  pip: >
+    name='git+https://github.com/paylogic/pip-accel.git@0.18#egg=pip-accel'
```
feanil: Can we install this from PyPI instead of github please.  https://pypi.python.org/pypi/pip-accel/0.19

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,18 +30,50 @@
...
+  pip: >
+    name='boto' version='2.34.0'
+    virtualenv={{ jenkins_home }}/edx-venv
+    virtualenv_command=virtualenv-2.7
+  sudo_user: "{{ jenkins_user }}"
+
+- name: Install edx-platform requirements using pip-accel
   pip: >
     requirements={{ jenkins_home }}/shallow-clone/requirements/edx/{{ item }}
-    extra_args="--exists-action w"
```
feanil, line 36: Why was this removed?

feanil:
Why does pip need AWS credentials?

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak: Update: figured out the wipe exists-action. This boto install should be fine now.

playbooks/roles/jenkins_worker/tasks/packer.yml
```diff
@@ -1,6 +1,8 @@
 ---
 - name: Download packer
   shell: "curl -L {{ packer_url }} -o /var/tmp/packer.zip"
+  args:
```
jzoldak, line 4: review note: this is unrelated to the other changes. It just speeds up the "vagrant provision" when I was developing.

playbooks/roles/jenkins_worker/tasks/python.yml
jzoldak, line 36: Was stuck on that. Turned out to be a bug in pip 1.4.1. Added back in. 

Also I'm not sure why there was no `=` sign originally, but I have confirmed that it works with it.

jzoldak:
@benpatterson @feanil 
All comments addressed. Please re-review.
Looking to merge this in.

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,12 +30,61 @@
     depth=1
   sudo_user: "{{ jenkins_user }}"
 
-- name: Install edx-platform requirements
+# Install the platform requirements using pip-accel.
+# This will allow the binary distributions to be downloaded from S3
+# rather than compiled each time.
+- name: Install pip-accel for caching python requirements
+  pip: >
+    name='pip-accel' version='0.21.1'
```
feanil: I would rather see a var that defines all the reqeuirements that need to be installed in defaults along with their versions, and then a single task to install both boto and pip-accel.  This will keep things DRY and make it easier to add other requirements in the future.

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,12 +30,61 @@
...
+    name='boto' version='2.34.0'
+    virtualenv={{ jenkins_home }}/edx-venv
+    virtualenv_command=virtualenv-2.7
+  sudo_user: "{{ jenkins_user }}"
+
+# WORKAROUND for pip-accel
+# Install a version of pip that will work with the exists-action flag
+- name: Install pip 1.4.1 patched
+  pip: >
+    name='git+https://github.com/jzoldak/pip.git@v1.4.1patch772#egg=pip'
```
feanil: Is there an issue for this berakage that we can also link to here?  What is the plan to not have to do this in the future?  Is this a bug in pip 1.5.6?  More comments here would be generally useful.

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,12 +30,61 @@
...
+    virtualenv={{ jenkins_home }}/edx-venv
+    virtualenv_command=virtualenv-2.7
+  sudo_user: "{{ jenkins_user }}"
+
+# WORKAROUND for pip-accel
+# Install Shapely which doesn't install cleanly with pip-accel
+- name: Install Shapely==1.2.16
+  pip: >
+    name='Shapely' version='1.2.16'
```
feanil: this can also go into the list of pre_pip_accel requirements.

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,12 +30,61 @@
...
+    virtualenv={{ jenkins_home }}/edx-venv
+    virtualenv_command=virtualenv-2.7
+    executable=pip
+  sudo_user: "{{ jenkins_user }}"
+
+# WORKAROUND for pip-accel
+# Install unittest2 which is needed by lettuce
+- name: Install unittest2
+  pip: >
+    name='unittest2'
```
feanil: same as other pre pip-accel requirements.  Also why doesn't this have a version?

playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -57,3 +63,19 @@
...
+
+- name: Create pip-accel config file
+  template:
+    src=pip-accel.conf.j2 dest={{ jenkins_home }}/.pip-accel/pip-accel.conf
+    owner={{ jenkins_user }} group={{ jenkins_group }} mode=0664
+
+# Set up configuration for pip-accel to use S3
+- name: Create boto config file with AWS credentials for the pip cache
+  template:
+    src=.boto.j2 dest={{ jenkins_home }}/.boto
```
feanil: If pip accel is using boto, why not use IAM instead of passing around credentials?

playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -34,6 +34,12 @@
   apt: pkg={{','.join(jenkins_debian_pkgs)}}
        state=present update_cache=yes
 
+- fail: AWS access key id was not passed in
```
benpatterson: Might want to specify which kind of key (since we have different S3 keys and we have EC2 keys...)

playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -34,6 +34,12 @@
...
+  when: pip_aws_access_key_id is not defined
+
+- fail: AWS secret access key was not passed in
```
benpatterson: Same here

util/packer/jenkins_worker.json
```diff
@@ -28,6 +30,10 @@
     "source": "../../requirements.txt",
     "destination": "{{user `playbook_remote_dir`}}/requirements.txt"
   }, {
+    "type": "file",
+    "source": "extra_vars.json",
```
benpatterson: Does this need to be added to the commit? Or, built by packer before this step?

benpatterson: Oh I see now...maybe this file provisioner should just be removed?

jzoldak:
@benpatterson passed the jenkins build-jenkins-worker packer tests. All comments addressed. Can I get a final review pls?

playbooks/roles/jenkins_worker/templates/pip-accel.conf.j2
```diff
@@ -0,0 +1,7 @@
+[pip-accel]
+auto-install = no
+data-directory = ~/.pip-accel
+download-cache = ~/.pip/download-cache
+s3-bucket = edx-platform.pip-accel-cache
```
benpatterson, line 5: should 'edx-platform.pip-accel-cache' be a var (so non-platform implementations can use)?

benpatterson:
LGTM. :+1: 

I have a nit, but not a must before merging.

playbooks/roles/jenkins_worker/templates/pip-accel.conf.j2
jzoldak, line 5: good point. will refactor on the next go-around.

PR #1724, opened 2014-11-28
E0d/optionally update ssh config

e0d:
Optionally checks if sshd is installed -- it isn't normally on docker -- before trying to edit the MOTD and bounce sshd.


fredsmith:
:+1: LGTM

PR #1732, opened 2014-12-08
Remove repos for courses we are about to checkout.

feanil:


jarv:
Why not a `git clean -xdf`?

feanil:
The repos have their .git directories removed before they are zipped so they are no longer git repos when the checkout tries to update the course.

e0d:
:+1:

jarv:
:+1:

PR #1735, opened 2014-12-09
moving sites-enabled to /etc/nginx/sites-enabled

jarv:
This is a small change that moves our sites-enabled directory to `/etc/nginx/sites-enabled`.

It does not remove the old directory `/edx/app/nginx/sites-enabled` which will no longer be used, it also keeps the sources in `/edx/app/nginx/sites-available`

### DIFF
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -54,7 +54,7 @@ nginx_server_static_dir: "{{ nginx_data_dir }}/server-static"
 nginx_conf_dir: "{{ nginx_app_dir }}/conf.d"
 nginx_log_dir: "{{ COMMON_LOG_DIR }}/nginx"
 nginx_sites_available_dir: "{{ nginx_app_dir }}/sites-available"
-nginx_sites_enabled_dir: "{{ nginx_app_dir }}/sites-enabled"
+nginx_sites_enabled_dir: "/etc/nginx/sites-enabled"
 nginx_user: root
 nginx_htpasswd_file: "{{ nginx_app_dir }}/nginx.htpasswd"
 nginx_default_sites: []
@@ -66,9 +66,9 @@ NGINX_EDXAPP_ENABLE_S3_MAINTENANCE: False
 nginx_default_error_page: "/server/server-error.html"
 NGINX_EDXAPP_ERROR_PAGES:
   "500": "{{ nginx_default_error_page}}"
-  "502": "{{ nginx_default_error_page}}"  
+  "502": "{{ nginx_default_error_page}}"
   "504": "{{ nginx_default_error_page}}"
-  
+
 CMS_HOSTNAME: '~^((stage|prod)-)?studio.*'
 
 nginx_template_dir: "edx/app/nginx/sites-available"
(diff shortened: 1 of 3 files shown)


carsongee:
Why are moving back to the system default, and don't we now need to delete the default site somewhere with this change that the deb package drops?

jarv:
@carsongee talking with @e0d yesterday we would like to propose that we by default not mess with the nginx.conf file and the nginx directories. Instead, link into the /etc/nginx/sites-enabled directory from our /edx/app/nginx dir which plays nicer with the nginx system package.

The reason for this change is for a packaging proof of concept that doesn't divert the nginx.conf.

carsongee:
That sounds fine to me, but I think we should remove `default` from `sites-enabled` so we don't get surprised with a brand new default domain web server.  Also, it looks like we have the dropping of the root conf only reloading the configuration, but I think we would need to restart for a change like this to take place. I also generally think that changing the root conf should trigger a restart instead of a reload.

jarv:
@carsongee any other concerns before merging?  I'll send a general announcement to openedx-ops as well to give folks a heads up.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -62,7 +62,7 @@ http {
         # Virtual Host Configs
         ##
 
-        include {{ nginx_sites_enabled_dir }}/*;
         include {{ nginx_conf_dir }}/*.conf;
+        include /etc/nginx/sites-enabled/*;
```
carsongee: I kind of like being explicit on the location, but since you already changed `{{ nginx_sites_enabled_dir }}` why not just leave this template as is?

carsongee:
Looks good to me. :+1: to this and the idea. I'm guessing there is a lot to undo in other roles similar to this one.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
jarv: fixed, I think what we want to do is just get rid of the nginx configuration diversion altogether. the only difference between the template and the OS package version is the log directory which I think we are ok to move back to /var/log to keep things standard

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -66,9 +66,9 @@ NGINX_EDXAPP_ENABLE_S3_MAINTENANCE: False
 nginx_default_error_page: "/server/server-error.html"
 NGINX_EDXAPP_ERROR_PAGES:
   "500": "{{ nginx_default_error_page}}"
-  "502": "{{ nginx_default_error_page}}"  
+  "502": "{{ nginx_default_error_page}}"
```
feanil: styling: these vars have a space before the var name but not after.

jarv: will fix but the change here was just the trailing space

feanil:
:+1:  other than what looks to be un-related styling issue.

playbooks/roles/nginx/defaults/main.yml
feanil: Yea, it looks like you linters just did a bit of cleanup and this was wrong previously, but if it's easy to fix, may as well do it while you're in here.

PR #1736, opened 2014-12-09
Add analytics-pipeline as a mounted dir.

benpatterson:
This commit does not introduce users; the vagrant user would run tasks.

@mulby @rocha @e0d 

Thoughts?

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -103,7 +103,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.synced_folder "#{ora_mount_dir}", "/edx/app/ora/ora",
       create: true, owner: "ora", group: "www-data"
     config.vm.synced_folder "#{analytics_pipeline_mount_dir}", "/edx/app/pipeline",
-      create: true, owner: "ora", group: "www-data"
+      create: true, owner: "edxapp", group: "www-data"
   else
     config.vm.synced_folder "#{edx_platform_mount_dir}", "/edx/app/edxapp/edx-platform",
       create: true, nfs: true
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -44,6 +44,7 @@ edx_platform_mount_dir = "edx-platform"
 themes_mount_dir = "themes"
 forum_mount_dir = "cs_comments_service"
 ora_mount_dir = "ora"
+analytics_pipeline_mount_dir = "edx-analytics-pipeline"
 
 if ENV['VAGRANT_MOUNT_BASE']
 
@@ -51,6 +52,7 @@ if ENV['VAGRANT_MOUNT_BASE']
   themes_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + themes_mount_dir
   forum_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + forum_mount_dir
   ora_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + ora_mount_dir
+  analytics_pipeline_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + analytics_pipeline_mount_dir
 
 end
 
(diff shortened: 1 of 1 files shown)


vagrant/release/devstack/Vagrantfile
```diff
@@ -109,6 +111,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       create: true, nfs: true
     config.vm.synced_folder "#{ora_mount_dir}", "/edx/app/ora/ora",
       create: true, nfs: true
+    config.vm.synced_folder "#{analytics_pipeline_mount_dir}", "/edx/app/pipeline",
```
feanil, line 116: Don't forget to add this to the VBOXFS section above as well, or refactor this to something cleaner.

benpatterson, line 116: Done.

benpatterson:
Addressed @feanil 's comment, but perhaps I should more directly ask:
- Do we consider having a pipeline user a requirement here?
- If so, do we want the requirements to be installed as well? I want to say no since that will be an expensive process that will only be useful to a certain subset of users
- Does the path /edx/app/pipeline make sense? I haven't tested the underlying mkdir functionality and am not sure if it'll give us subdirs for free (e.g., /edx/app/analytics/pipeline)

Thx for any comments here. I hope to pick away at this while I'm doing some dev in this space.

feanil:
I only see the vagrant file changes, are there changes to the vagrant-devstack.yml file as well?  If it is a top-level standalone app then I think it belongs under '{{ COMMON_APP_DIR }}/<app_name>'.  Each top-level app should have it's own user and requirements file.  Is there a PR for the work of that role?

rocha:
@benpatterson what is the goal with this PR? Given that the pipeline will not work without all its dependencies, the only thing that the PR does is bring the code in. Is that a requisite for some of the work you are doing?

benpatterson:
@rocha I suppose the goal of the PR would be to ultimately include the pipeline in devstack, but I'm not sure what the requirements for that would be. I was hoping I could start with this, then hear what else is needed and work its way into something that is mergeable.

Personally, I mounted my dir so I could run jobs on devstack. It'll shorten the cycle a little between creating events and running the job. I am configured to look in /edx/log/tracking. I only need to rename the log file on devstack to be sure it's found. (It will also save me from encountering any OSX-only weirdness).

rocha:
@benpatterson are you able to run the jobs you need without using Hadoop?

benpatterson:
@rocha I use the emulator

benpatterson:
@rocha I had been imagining checking in a devstack.cfg file into the pipeline repo as part of the deal

vagrant/release/devstack/Vagrantfile
```diff
@@ -100,6 +102,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       create: true, owner: "forum", group: "www-data"
     config.vm.synced_folder "#{ora_mount_dir}", "/edx/app/ora/ora",
       create: true, owner: "ora", group: "www-data"
+    config.vm.synced_folder "#{analytics_pipeline_mount_dir}", "/edx/app/pipeline",
+      create: true, owner: "ora", group: "www-data"
```
brianhw: You have the owner here as "ora".  Is that appropriate?

benpatterson: Fixed. Thanks for catching that.

e0d:
@benpatterson @feanil What's the status of this PR?

benpatterson:
I'd need some feedback from the team on how to proceed.

Excepting that, I think we should create another user so a developer can just go sudo su pipeline and land at the new $HOME with an environment activated. If we fully build the virtualenv, it will add some cycles to the initial devstack provisioning. We could also pare this down to have an empty environment that a user would need to create by running the make scripts.

@mulby do you happen to have an opinion here?

benpatterson:
CC @brianhw not sure if you have an opinion on devstacky stuff....

rocha:
@benpatterson is there a way to make the pipeline optional, so it does not have to be included and build into the stack unless specified? The list of requirements for the pipeline can be quite long, and like you mentioned it could add unnecessary cycles to the provisioning when the pipeline is not required.

mulby:
Closing since we now support an analytics devstack

PR #1738, opened 2014-12-10
Use common packages between edxlocal and edxapp.

benpatterson:
@feanil @singingwolfboy @clytwynec 

As per discussion for getting libjpeg onto jenkins_workers with some refactoring. 

To be tested once you give it a look.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -744,20 +744,13 @@ edxapp_chrislea_ppa: "ppa:chris-lea/node.js"
 edxapp_debian_pkgs:
   # for compiling the virtualenv
   # (only needed if wheel files aren't available)
-  - build-essential
   - s3cmd
   - pkg-config
   - graphviz-dev
-  - graphviz
-  - libmysqlclient-dev
   # for scipy, do not install
   # libopenblas-base, it will cause
   # problems for numpy
-  - gfortran
-  - liblapack-dev
   - g++
-  - libxml2-dev
-  - libxslt1-dev
   # apparmor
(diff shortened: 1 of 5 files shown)


benpatterson:
Intended as an alternative solution to #1655.

feanil:
@benpatterson  I thought from the e-mail that we would not be pursuing this now.  Did something change or is this just a kickstart of the future work?

benpatterson:
Just a kickstart, @feanil 

I don't think we could implement this right now as it is, because of the testing cost. But I would like to have some discussion on the right overall approach. Let's move discussion to the Jira ticket.

I'll leave this PR open in case we go in this direction, but let's chat first.

feanil:
Cool got it.

e0d:
@benpatterson I think that ultimately, I prefer a convention for specifying OS level requirements in the repo for each project rather than having them managed in ansible because the project is a complete thing without any dependencies on ansible if someone chooses a different configuration management tool.

This does add some additional risk, because dependencies can be added without devops knowing.  Dependencies may conflict between services and we won't know that until later than ideal.  Despite this, I still think I prefer defining them in our projects.  Thoughts?

benpatterson:
@e0d my goal here was to reduce upkeep and duplication. 

I think what you're suggesting merits a longer conversation. This PR can wait for that conversation at this point, before proceeding.

benpatterson:
@e0d could we discuss with @feanil ? 

I'll shoot you guys an e-mail.

benpatterson:
@e0d @feanil @clytwynec after yesterday's discussion I believe we are GTG on this. 

I'll plan to merge later today unless you'd like me to hold back on it.

PR #1739, opened 2014-12-10
When overrides are set they are set by service variant name.

feanil:
@carsongee


feanil:
See [this](https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L325-L330) for context.

carsongee:
Totally missed that.  :+1:

PR #1740, opened 2014-12-10
Jarv/create pkg role WIP

jarv:


playbooks/roles/create_pkg/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
...
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Handlers for role create_pkg
+# 
+# Overview:
+# 
+#
+- name: notify me
```
e0d: Let's remove.

playbooks/roles/create_pkg/meta/main.yml
```diff
@@ -0,0 +1,20 @@
+---
```
e0d: Let's remove.

playbooks/roles/edxlocal/defaults/main.yml
```diff
@@ -4,3 +4,4 @@ edxlocal_debian_pkgs:
   - mysql-server-5.5
```
e0d: Think this needs a re-base, need to confirm.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -3,66 +3,93 @@ worker_processes 4;
...
+	##
+	# Virtual Host Configs
+	##
+
+	include /etc/nginx/conf.d/*.conf;
+	include /etc/nginx/sites-enabled/*;
 }
 
+
+#mail {
```
e0d, line 76: Oh, is this intended to mirror what is installed by the deb package?  Why wouldn't we remove this entirely from our templates?

jarv, line 76: @e0d At the time I was thinking that we need to include it so that the configuration file reverts to the package version, I think ideally we would want to not template it out at all so that just the package manages it

playbooks/roles/edxlocal/defaults/main.yml
jarv: This commit is part of the PR, I'm not sure why memcached was ever installed as a separate task.

e0d: I'm happy with the memcached was/am confused by the mysql-server dep, which is gone on master https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/defaults/main.yml

jarv: ah was looking at the wrong diff, yeah that does seem change. I'll add it back.

jarv:
rebased from master

playbooks/roles/create_pkg/defaults/main.yml
```diff
@@ -0,0 +1,36 @@
...
+  - ruby-dev
+  - rubygems
+# global excludes
+create_pkg_excludes:
+  - '**/.git'
+  - '*.pyc'
+
+create_pkg_local_server_debian_pkgs:
+  - python-mysqldb
+  - mysql-server-5.5
```
e0d, line 34: sigh, ideally we want this to be 5.6 which means a new PPA.  The logic is in edxlocal, but this seems to complicate distribution a fair bit.

e0d:
I'm in favor of merging this, but we need to test the nginx changes.  Do you think we should take on removing the ngnix.conf now?  Seems like it problematic if a package update the same config.

jarv:
The changes to nginx are to move the symlinks to /etc/nginx/sites-enabled and to move the edx specific configuration to the conf.d/ directory which we can't do without changing the existing nginx.conf. Do you want to back out both or just the conf.d/ change? If we just back out the conf.d change the template won't change from what we have on master.

e0d:
I was unclear; I like the nginx changes.  My only question is whether we could remove the template for playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2 and rely on the version the package drops.

jarv:
I see - the issue with removing the template altogether is that these changes will then not be backwards compatible with servers that have already been configured.

jarv:
@e0d or @feanil any more thoughts on this?

e0d:
@jarv my only concern is the MySQL 5.5 versus 5.6 issue.  Though I think that trying to install 5.6 on 12.04 from a package is definitely swimming against the stream -- or worse.  Maybe a mention in the blob of text that upgrading to 5.6 is recommended, if possible.  I'd also like another sanity check.

@fredsmith or @maxrothman can  you have a look?

jarv:
Should be pretty easy to add the install script for 5.6 like it does now for mongo

maxrothman:
:+1: but don't take my word for it.

playbooks/roles/edxapp/templates/edx/bin/edxapp-django-shell-cms.j2
```diff
@@ -0,0 +1,3 @@
+{% include "edxapp_common.j2" %}
```
feanil, line 1: I don't see the `edxapp_common.j2` file as a part of this PR is that defined elsewhere?

feanil:
One question about a missing template but otherwise looks good to me.

playbooks/roles/edxapp/templates/edx/bin/edxapp-django-shell-cms.j2
jarv, line 1: `roles/edxapp/templates/edx/bin/edxapp_common.j2` already on master

feanil, line 1: Ah, cool.

feanil:
:+1:

openedx-webhooks:
Thanks for the pull request, @jarv! I've created [OSPR-389](https://openedx.atlassian.net/browse/OSPR-389) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

e0d:
There's still the MySQL version issue with this PR.  I'm OK with merging as
long as we have something in the backlog to get this to use MySQL 5.6.

jarv:
OK I'll look into the rebase so we can do one final review, will you guys hit the merge button or should I?

jarv:
@e0d @feanil @fredsmith rebased, ready to merge

PR #1741, opened 2014-12-11
adding helper scripts in /edx/bin

jarv:
This adds some helper scripts for managing the edx server.


feanil:
Can you also update the configuration to use these scripts instead of calling migrate directly?

fredsmith:
:+1:  I think these would be useful to have merged in.  is /edx/bin/ added to $PATH automatically?

jarv:
@fredsmith not yet but we could do that pretty easily

feanil:
:+1:

PR #1746, opened 2014-12-11
change the repo for certificates and fix checkout

jarv:
When there isn't an identity file provided we should not use GIT_SSH


playbooks/roles/certs/defaults/main.yml
```diff
@@ -35,7 +35,7 @@ CERTS_LOCAL_PRIVATE_KEY: "example-private-key.txt"
 
 # This defaults to the public certificates repo which is
 # used for open-edx
-CERTS_REPO: "https://github.com/edx/read-only-certificate-code"
+CERTS_REPO: "https://github.com/edx/edx-certificates"
```
jarv, line 38: note - these repositories are the same, the old one was renamed

feanil:
:+1:

PR #1749, opened 2014-12-15
Work for mounting currently detached volumes and inspecting them to tag them.

feanil:


util/vpc-tools/tag-old-ebs.py
```diff
@@ -0,0 +1,196 @@
...
+            tags.update(value)
+
+    return tags
+
+def potential_devices(root_device):
+    device_dir = dirname(root_device)
+    relevant_devices = lambda x: x.startswith(basename(root_device))
+
+    all_devices = os.listdir(device_dir)
+    all_devices = filter( relevant_devices, all_devices)
```
e0d: whitespace

util/vpc-tools/tag-old-ebs.py
```diff
@@ -0,0 +1,196 @@
...
+                vol.add_tag("devices_on_volume", str(devices_on_volume))
+                # Don't tag in this case because the different devices
+                # may have conflicting tags.
+                logging.info("Skipping {} because it has multiple mountpoints.".format(vol.id))
+                logging.info("{} has mountpoints {}".format(vol.id, str(devices_on_volume)))
+            else:
+                device = devices_on_volume[0]
+                try:
+                    # Mount the volume
+                    subprocess.check_call(["sudo", "mount", device, mountpoint])
```
e0d, line 175: Mixed feelings about making folks run the script as root or via sudo and doing this in an isolated way.

util/vpc-tools/tag-old-ebs.py
```diff
@@ -0,0 +1,196 @@
...
+            is_attached = lambda vol_id: ec2.get_all_volumes(vol_id)[0].status != "available"
+            while exists(root_device) or is_attached(vol.id):
+                if is_attached(vol.id):
+                    try:
+                        # detach the volume
+                        ec2.detach_volume(vol.id)
+                    except boto.exception.EC2ResponseError as e:
+                        logging.warning("Failed to detach volume.  Will try again in a bit.")
+                time.sleep(2)
+                logging.debug("Waiting for {} to be detached.".format(vol.id))
```
e0d: Should we give up eventually?

e0d:
Couple small comments, but :+1: after squash.

util/vpc-tools/tag-old-ebs.py
feanil, line 175: I figured sudo was better than root, but if you have suggestions for how to approach this, I'm happy to change it.  One other option I can think of is to update fstab to register the mountpoint so that non privileged users can mount the device.

util/vpc-tools/tag-old-ebs.py
feanil: I'll add a 5 minute timeout after which, I'll exit with an error. In most cases it does eventually detach but in the process of detaching the API calls are sometimes inconsistent.

PR #1750, opened 2014-12-16
Turn off basic auth for all endpoints below '/api'.

feanil:
Both basic auth and OAUTH2 use the same authorization header so a client can't auth both ways.


e0d:
:+1:

PR #1755, opened 2014-12-19
Fix notifier role defaults never working

maxrothman:
Refactored `NOTIFIER_USE_GIT_IDENTITY` to be represented by `NOTIFIER_GIT_IDENTITY` being an empty string.

### DIFF
--- a/playbooks/roles/notifier/tasks/deploy.yml
+++ b/playbooks/roles/notifier/tasks/deploy.yml
@@ -16,19 +16,19 @@
   template: >
     src=git_ssh_noauth.sh.j2 dest={{ notifier_git_ssh }}
     owner={{ NOTIFIER_USER }} mode=750
-  when: not NOTIFIER_USE_GIT_IDENTITY
+  when: NOTIFIER_GIT_IDENTITY == ""
 
 - name: create ssh script for git (authenticated)
   template: >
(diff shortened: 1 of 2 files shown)


playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -16,27 +16,19 @@
...
-  when: NOTIFIER_USE_GIT_IDENTITY
+  when: NOTIFIER_GIT_IDENTITY != ""
 
 - name: install read-only ssh key
   copy: >
     content="{{ NOTIFIER_GIT_IDENTITY }}" dest={{ notifier_git_identity }}
     force=yes owner={{ NOTIFIER_USER }} mode=0600
-  when: NOTIFIER_USE_GIT_IDENTITY
-
-- name: checkout theme
```
fredsmith: We need to leave theming in.  This shouldn't be removed.  Perhaps NOTIFIER_THEME_NAME just needs a default?

maxrothman: Whoops, I don't think I meant to remove all of that. I'll fix.

fredsmith:
:+1:

e0d:
:+1:

PR #1756, opened 2014-12-22
Fix edxapp not restarting

bdero:
This fixes an issue where edxapp doesn't restart when deploying using playbooks that run the edxapp role twice (once for the celery workers, and once for the LMS and CMS).

When a role is run in ansible, it seems that the state of handlers is kept in the scope of the play. Once a `when` clause for a handler has been evaluated (which seems to happen at the end of each role by default, even though they're not flushed until the end of the play), it is never evaluated again during that play, even when the handler is notified and executed after already having been flushed.

For instance, in the vagrant-fullstack playbook, the edxapp role is executed twice as follows:

``` yaml
    - { role: 'edxapp', celery_worker: True }
    - edxapp
```

..which results in the edxapp and edxapp_worker restart handlers being executed only once at the end of the play:

```
NOTIFIED: [edxapp | restart edxapp] ******************************************* 
skipping: [192.168.30.10] => (item=lms)
skipping: [192.168.30.10] => (item=cms)

NOTIFIED: [edxapp | restart edxapp_workers] *********************************** 
changed: [192.168.30.10] => (item={'queue': 'low', 'service_variant': 'cms', 'concurrency': 3})
changed: [192.168.30.10] => (item={'queue': 'default', 'service_variant': 'cms', 'concurrency': 4})
changed: [192.168.30.10] => (item={'queue': 'high', 'service_variant': 'cms', 'concurrency': 1})
changed: [192.168.30.10] => (item={'queue': 'low', 'service_variant': 'lms', 'concurrency': 1})
changed: [192.168.30.10] => (item={'queue': 'default', 'service_variant': 'lms', 'concurrency': 3})
changed: [192.168.30.10] => (item={'queue': 'high', 'service_variant': 'lms', 'concurrency': 4})
changed: [192.168.30.10] => (item={'queue': 'high_mem', 'service_variant': 'lms', 'concurrency': 2})
```

Well, what we really want is to have both the apps and the workers restart, so we'll need to flush the handlers at the end of the edxapp role. So let's say we add the task `meta: flush_handlers` to the end of deploy.yml in the edxapp role.
The result is that the restart handlers now run twice, but both as if celery_worker is defined (like above), even though it's not while the second role is executing.

The fix is to just turn the two handlers into tasks, since they're _always_ executed whenever the edxapp role is run anyhow.


openedx-webhooks:
Thanks for the pull request, @bdero! I've created [OSPR-318](https://openedx.atlassian.net/browse/OSPR-318) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

benpatterson:
Hi! I'm curious if flush_handlers would do the same? http://docs.ansible.com/playbooks_intro.html#handlers-running-operations-on-change

e0d:
We're testing some on this end to make sure we understand the implications, but, generally, I think this looks good.

carsongee:
This looks good to me.  The upshot being that I don't have to manually restart services after a double run and it cuts down on the code quite a bit.  It is a bummer that this doesn't work right though.  Especially `flush_handlers`

e0d:
@carsongee can you confirm that @benpatterson suggestion to use flush_handlers doesn't address the root cause?

carsongee:
Just from what @bdero wrote in the description, it appears there is a bug in the logic of flush_handlers with how it handles conditionals, but I haven't personally confirmed it:

> Well, what we really want is to have both the apps and the workers restart, so we'll need to flush the handlers at the end of the edxapp role. So let's say we add the task meta: flush_handlers to the end of deploy.yml in the edxapp role.
> The result is that the restart handlers now run twice, but both as if celery_worker is defined (like above), even though it's not while the second role is executing.

e0d:
@feanil please weigh in.  I've never been the biggest fan an handlers, so maybe I should recuse myself for this PR.

e0d:
@fredsmith thoughts?

maxrothman:
Boo handlers :+1:

feanil:
It's unfortunate that we would be restarting even if things haven't changed, but I think this is fine for the common case especially because of the bug when both services are setup on the same machine with a single ansible play.  That's pretty unexpected behaviour.  @bdero have you submitted a bug to ansible about this?  If not, I think it's worth it to track that to this PR.

:+1:

e0d:
@bdero @carsongee think we're good to merge

PR #1759, opened 2014-12-29
Update the settings for the root disk for abbey.

feanil:
Use SSD backed EBS instead of standard.
Delete the volume upon abbey termination.


e0d:
:+1:

PR #1762, opened 2014-12-30
Fix the thing.

feanil:
Set params on BlockDeviceType instead of BlockDeviceMapping.


fredsmith:
that looks better :+1:

PR #1763, opened 2014-12-30
Only run dependencies conditionally.

feanil:


e0d:
Why not have deps lists per OS family as we have stubbed out elsewhere?

feanil:
I didn't see anything like that, can you point me to an example?

fredsmith:
@feanil like https://github.com/edx/configuration/blob/2129abc370abf13ce221c87764a32266d44a74f4/playbooks/roles/edx_service/tasks/main.yml#L69 ?

I don't really see anything functionally different from what you're doing vs. that, though. 

I also don't really understand why we include specific roles based on the distro.

feanil:
The reason is that those roles are not distribution agnostic and assume debian variants(apt), which prenvents us from running the newrelic role on non debian boxes(ec2-linux).

e0d:
The PR seems to me to violate the principal of least astonishment.  I realize we don't support RH variants in any useful way, but shouldn't we move the check to the roles that are being included rather than in the meta include?  Minimally, I'd like comments and a TODO.

PR #1773, opened 2015-01-06
Use the V1 Hipchat token.

feanil:
It doesn't have the same rate limiting limitations as the V2 Token which we need since we have
so many rooms.


e0d:
Are there other implications to switching to the v1 key?

feanil:
Turns out we need both the V1 and V2 tokens.  The v1 tokens are used to look up rooms and user info because if we have a lot of rooms, we hit the rate limit of the v2 tokens.  With a V1 token there is the danger that Alton can also delete rooms and users.

fredsmith:
:+1:

PR #1774, opened 2015-01-09
Fix working dir.

e0d:


fredsmith:
:+1:  

This behavior of manage.py in the LMS/CMS, where it requires being in the working directory, seems new.   Is this a bug?

PR #1747, opened 2014-12-14
Use COMMON_GIT_MIRROR

fmyzjs:
**Ticket:** [OSPR-297](https://openedx.atlassian.net/browse/OSPR-297)

**Background:**
Some times ,because of the GFW or worse network,people in China can not use github services.

 I open this pull  to make sure we can switch our local edx  repo mirrors.

**Studio Updates:**  None.

**LMS Updates:** None

**Test:** [travis](https://travis-ci.org/eduStack/configuration/builds/43984602)

PR #1785, opened 2015-01-14
Fmyzjs common git mirror

maxrothman:
Found this on my computer. Not sure where it came from, but it refactors a bunch of roles to use `COMMON_GIT_MIRROR`.


maxrothman:
@feanil @e0d @fredsmith thoughts? We might want to take this a little further once [#1789](https://github.com/edx/configuration/pull/1789) lands, since it'll give us `COMMON_GIT_PROTOCOL` and `COMMON_GIT_PATH` as well.

fredsmith:
Copy of #1747

PR #1789, opened 2015-01-20
Refactor git checkout into edx_service role

maxrothman:
### Notes
- edx_service can now accept multiple git repos, but it has a strong dependency on there being a single primary one. Some more work is needed to refine what the use cases for multiple repos are and then build the needed functionality.
- supervisor has not been refactored into edx_service yet, but once it has, a task needs to be added to restart the relevant service at the end of the tasks.
- I've only tested this as far as the end of the edx_service tasks. More testing is needed.
- The changes to edx_service might not be backwards-compatible with other roles


playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -88,3 +88,64 @@
...
+- name: checkout code over https
+  git: >
+    repo=https://{{item.DOMAIN}}/{{item.PATH}}/{{item.REPO}}
+    dest={{ item.DESTINATION }}  version={{ item.VERSION }}
+  sudo_user: "{{ edx_role_name }}"
+  register: code_checkout
+  when: item.PROTOCOL == "https"
+  with_items: edx_service_config.GIT
+
+- debug: var=code_checkout
```
e0d: remove

playbooks/roles/edx-notes-api/defaults/main.yml
```diff
@@ -65,16 +64,24 @@ edx_notes_api_service_config:
       ENGINE: 'notesserver.highlight.ElasticsearchSearchEngine'
       URL: "{{ EDX_NOTES_API_ELASTICSEARCH_URL }}"
       INDEX_NAME: '{{ EDX_NOTES_API_DATASTORE_NAME }}'
-  DISABLE_TOKEN_CHECK: True  
+  DISABLE_TOKEN_CHECK: True
+  GIT:
```
e0d: Better name, EDX_NOTES_REPOS?

playbooks/roles/edx-notes-api/defaults/main.yml
```diff
@@ -11,35 +11,34 @@
 # Defaults for role edx-notes-api
 # 
 
-EDX_NOTES_API_VERSION: "master"
 EDX_NOTES_API_GUNICORN_WORKERS_EXTRA: ""
 EDX_NOTES_API_WORKERS_EXTRA_CONF: ""
-EDX_NOTES_API_LANG: "en-us"
+EDX_NOTES_API_LANG: en-us
```
e0d, line 16: Let's do significant whitespace/convention changes are independent PRs.  We typically quote values in our configs.  I'm not a zealot either way, but we should decide what the convention is going to be.

maxrothman, line 16: Ok, sure. For reference, quotes are only required when the value begins with a "{" (e.g. `value: "{{some_var}}"`).

playbooks/roles/edx-notes-api/defaults/main.yml
maxrothman: Unfortunately, this can't have a unique name because it's in the `edx_notes_api_service_config`, which is getting passed into `edx_service`. `edx_service` has to have a static name to access this stuff at.

playbooks/roles/edx-notes-api/defaults/main.yml
feanil, line 16: My vote would be to quote all string values so we don't have to be thinking about the special cases where it is necessary.  It's a simpler rule to follow and will have more predictable behavior when the value contains whitespace.

playbooks/roles/edx-notes-api/defaults/main.yml
```diff
@@ -82,16 +89,15 @@ edx_notes_api_home: "{{ COMMON_APP_DIR }}/{{ edx_notes_api_service_name }}"
...
 edx_notes_api_conf_dir: "{{ edx_notes_api_home }}"
 edx_notes_api_log_dir: "{{ COMMON_LOG_DIR }}/{{ edx_notes_api_service_name }}"
 
-edx_notes_api_gunicorn_host: "127.0.0.1"
-edx_notes_api_gunicorn_port: "8120"
-edx_notes_api_gunicorn_timeout: "300"
-edx_notes_api_wsgi: "notesserver.wsgi:application"
-edx_notes_api_nginx_port: "18120"
+edx_notes_api_gunicorn_host: 127.0.0.1
+edx_notes_api_gunicorn_port: 8120
```
feanil: This is not the same as removing quotes from a string since this will now be treated as a number.  It may not have any impact if it's just dropped into templates but can change behavior if the value is output as JSON for example.

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -88,3 +88,62 @@
...
+  register: "{{edx_service_name}}_code_checkout"
+  when: item.PROTOCOL == "ssh"
+  with_items: edx_service_config.GIT
+
+- name: checkout code over https
+  git: >
+    repo=https://{{item.DOMAIN}}/{{item.PATH}}/{{item.REPO}}
+    dest={{ item.DESTINATION }}  version={{ item.VERSION }}
+  sudo_user: "{{ edx_role_name }}"
+  register: code_checkout
```
feanil, line 130: should this also be `{{ edx_service_name }}_code_checkout`?  Is that something that gets passed into the service role when you declare it as a dependency?

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -88,3 +88,62 @@
...
+  when: (item.PROTOCOL != "https") and (item.PROTOCOL != "ssh")
+  with_items: edx_service_config.GIT
+
+- name: checkout code over ssh
+  git: >
+    repo=git@{{item.DOMAIN}}:{{item.PATH}}/{{item.REPO}}
+    dest={{ item.DESTINATION }}  version={{ item.VERSION }}
+    accept_hostkey=yes key_file={{ edx_notes_api_home }}/.ssh/id_rsa
+  sudo_user: "{{ edx_role_name }}"
+  register: "{{edx_service_name}}_code_checkout"
```
feanil: By convention we put spaces on either side of the curly braces.

playbooks/roles/edx_service/tasks/main.yml
maxrothman, line 130: It was previously `{{ edx_service_name }}_code_checkout`, but declaring variable names with values of other variables didn't work with looping over multiple repos. Besides, doing so is poor form anyways. I could put a prefix on the front if you think there's the potential for a namespace problem, but it's just a temporary variable for the "tag instance" task to use. It's incorrect above.

 `{{ edx_role_name }}` is passed in as an argument to the `edx_service` role. In the case of the notes api, it'll be `{{ edx_notes_api_role_name }}`

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -88,3 +88,62 @@
   yum: pkg={{ item }} state=present
   with_items: "{{ edx_service_name.replace('-', '_') }}_redhat_pkgs"
   when: ansible_distribution in common_redhat_variants
+
+
+- name: check if git repo exists before pruning
+  stat: path={{item.DESTINATION}}/.git
```
feanil: nitpick: spacing

playbooks/roles/edx-notes-api/defaults/main.yml
feanil, line 16: +1 to doing the whitespace/quoting change independently as this PR looks good otherwise.

playbooks/roles/edx_service/tasks/main.yml
maxrothman: That was intentional, actually. I was intending to separate the different conceptual bits of the file.

feanil: I meant inside of the curly braces for `item.DESTINATION`

playbooks/roles/edx-notes-api/defaults/main.yml
e0d: Per in person conversation, the GIT stuff should be pulled out of this dictionary, which is specifically for app configuration.

e0d:
:+1: Looks good.

PR #1790, opened 2015-01-21
No longer in alpha, comments update.

benpatterson:
I was working on something different when I noticed these nits. Fair to say we're no longer in alpha stage.


e0d:
:+1:

PR #1791, opened 2015-01-21
Add JS Cover for both local dev and jenkins_worker.

benpatterson:
@e0d @clytwynec @singingwolfboy @feanil @maxrothman @jzoldak 

Currently JSCover is not installed on devstack by default. This change will put it there and make it available to the edxapp user. This way, when running the javascript tests on platform, a coverage report will be generated. This can be used for examining coverage, but it can also be used by running diff-cover. (A subsequent change, when I have the time, would be wrapping diff-cover (for javascript) into our existing paver commands).

This change also keeps the JSCover as an installable for jenkins_workers, which is still a requirement today.

Please let me know what you think. My hope/goal is to get this in for Birch.

### DIFF
--- a/playbooks/roles/jscover/defaults/main.yml
+++ b/playbooks/roles/jscover/defaults/main.yml
@@ -1,0 +1,11 @@
+---
+# Installs JSCover jar.
+# Java is a pre-requisite for JSCover. This role is not responsible
+# for installing Java.
+#
+jscover_role_name: jscover
+
+# JSCover direct download URL
+jscover_url: "http://files.edx.org/testeng/JSCover-1.0.2.zip"
+jscover_version: "1.0.2"
+
--- a/playbooks/roles/jscover/defaults/main.yml
+++ b/playbooks/roles/jscover/defaults/main.yml
@@ -6,6 +6,5 @@
 jscover_role_name: jscover
 
 # JSCover direct download URL
-jscover_url: "http://files.edx.org/testeng/JSCover-1.0.2.zip"
 jscover_version: "1.0.2"
-
+jscover_url: "http://files.edx.org/testeng/JSCover-{{ jscover_version }}.zip"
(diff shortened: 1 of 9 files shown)


singingwolfboy:
Seems like a good idea to me! However, I'm not familiar with our ansible configs, and so can't really review this PR.

e0d:
:+1:

playbooks/roles/jscover/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+# Installs JSCover jar.
+# Java is a pre-requisite for JSCover. This role is not responsible
+# for installing Java.
+#
+jscover_role_name: jscover
+
+# JSCover direct download URL
+jscover_url: "http://files.edx.org/testeng/JSCover-1.0.2.zip"
+jscover_version: "1.0.2"
```
jzoldak, line 9: I think I'm missing something - I'm confused why this is hard coded here but put in a variable in playbooks/roles/local_dev/meta/main.yml?

benpatterson, line 9: Yes, by default, when running the JSCover role, you'll get v 1.0.2. 

In the case of the local_dev role, it's telling JSCover what version it wants. (Which at this point, is the same version.) That happens here: https://github.com/edx/configuration/pull/1791/files#diff-d482f581f35bbf9f7eab8e30e7fd9420R4

jzoldak, line 9: But say I change localdev_jscover_version to "1.0.3", then wouldn't it will point JSCOVER_JAR to /usr/local/bin/JSCover-all-1.0.3.jar, which won't exist?

benpatterson, line 9: Oh I get you now. You're right. We should use that same version string in the .zip download, too. 

This is a bug inherited from the previous iteration of JSCover installs on the jenkins_worker. Nice catch. I'll fix that.

benpatterson:
@jzoldak Addressed your comment

jzoldak:
:+1:

PR #1797, opened 2015-01-22
boto needs to be at least this version to be able to tag volumes

fredsmith:


feanil:
:+1:  once tests pass.

PR #1801, opened 2015-01-26
Benp/fix jenkins worker

benpatterson:
@clytwynec 
also @feanil since he's aware of the edxapp_common role from previous changes


feanil:
looks good to me. :+1:

PR #1802, opened 2015-01-27
Convert dashes to underscores

maxrothman:
I've split this work into 3 commits. All of the changes in the first should be fairly safe, while changes in the second could potentially break some things, so please review those carefully.


e0d:
```
(let
[ sad-feelings (make-sad ed)
  review-disposition (str ":+" (inc 0) ":")
]
(prn review-disposition)
)

:+1:

user>
```

:+1:

playbooks/roles/edx_notes_api/defaults/main.yml
```diff
@@ -15,17 +15,17 @@ EDX_NOTES_API_GUNICORN_WORKERS_EXTRA: ""
...
 EDX_NOTES_API_WORKERS: !!null
-EDX_NOTES_API_DATASTORE_NAME: edx-notes-api
+EDX_NOTES_API_DATASTORE_NAME: "{{ edx_notes_api_service_name }}"
 EDX_NOTES_API_MYSQL_DB_USER: notes001
 EDX_NOTES_API_MYSQL_DB_NAME: "{{ EDX_NOTES_API_DATASTORE_NAME }}"
 EDX_NOTES_API_MYSQL_DB_PASS: secret
 EDX_NOTES_API_MYSQL_HOST: localhost
 EDX_NOTES_API_ELASTICSEARCH_URL: http://127.0.0.1:9200
-EDX_NOTES_API_NEWRELIC_APPNAME: "{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-notes"
+EDX_NOTES_API_NEWRELIC_APPNAME: "{{ COMMON_ENVIRONMENT }}_{{ COMMON_DEPLOYMENT }}_notes"
```
feanil: New relic app names should stay consistent with existing cluster names so that they are easy to find.

feanil:
Other than the new relic app name, :+1:

maxrothman:
@feanil @e0d You're totally sure about the changes in the second commit? They include changing things like `DATASTORE_NAME` and the git repo name.

maxrothman:
Blocked by https://github.com/edx/edx-notes-api/pull/19

e0d:
I think there's still some testing that needs to happen around the --depth=1, but generally I'm :+1:  The depth argument has some nice properties in general.  We should verify that:
- it works from bare metal
- upgrades work
- the /edx/bin/update script still works.
- that it solves the deleted branch issue -- though maybe this is very unlikely to happen with --depth=1.

maxrothman:
It works from bare metal and upgrades work, but /edx/bin/update probably won't work since you can't check out other branches.

I found **yet another** potential solution in [git config fetch.prune](http://git-scm.com/docs/git-config), and I'll patch when I've got it working.

feanil:
:+1:

PR #1803, opened 2015-01-28
Correct when statement for git prune

singingwolfboy:
`git_dir.isdir` is never defined, even if `/edx/app/edxapp/edx-platform/.git` exists and is a directory.

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -46,7 +46,7 @@
   notify:
   - "restart edxapp"
   - "restart edxapp_workers"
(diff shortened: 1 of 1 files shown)


fredsmith:
:+1:

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -46,7 +46,7 @@
   notify:
   - "restart edxapp"
   - "restart edxapp_workers"
-  when: git_dir.stat.exists and (git_dir.isdir is defined) and git_dir.stat.isdir
+  when: git_dir.stat.exists and git_dir.stat.isdir
```
feanil, line 49: Seems like this is a typo and the actual correction should be `git_dir.stat.isdir is defined`, `gitdir.stat.isdir` is not defined when the location is not a directory.

PR #1805, opened 2015-01-28
Add the ability to embargo users based on CIDR blocks.

feanil:
This can be seen in action on my sandbox right now feanil.m.sandbox.edx.org

Normal traffic is fine, traffic from the MIT vpn will get the embargo page.


fredsmith:
:+1:  LGTM

PR #1807, opened 2015-01-29
apply glibc update if vulnerable to CVE-2015-0235

fredsmith:


playbooks/roles/security/tasks/security-ubuntu.yml
```diff
@@ -48,3 +48,24 @@
...
+  shell: gcc -o /tmp/GHOST /tmp/GHOST.c
+
+- name: Check if we are vulnerable
+  shell: /tmp/GHOST
+  register: test_ghost_vuln
+  ignore_errors: yes
+
+- name: Apply glibc security update if we are vulnerable
+  apt: name=libc6 state=latest update_cache=true
+  when: "'vulnerable' in test_ghost_vuln.stdout"
```
singingwolfboy, line 70: The code either prints `vulnerable` or `not vulnerable`. "vulnerable" will occur in stdout in either case. This doesn't seem like a good test.

e0d, line 70: Agree that will need to be addressed.  Otherwise :+1:

fredsmith, line 70: GHOST now prints "OK" instead of "not vulnerable"

thanks!

carsongee:
This won't reboot if the system is vulnerable.  Also, since you are checking with a fresh process instead of things like udev that are linked to glibc on boot, you will only know if new processes aren't vulnerable, but won't catch ones that still are (which on a system is pretty much everything).  That is why I went more with a one shot update for this since it is basically impossible to check a running system. I guess you could run your check and then also stat if /var/run/reboot-required exists, and reboot on that, but I'm not 100% sure that works...

PR #1809, opened 2015-01-29
Fix ssh key checkout being broken

maxrothman:


playbooks/roles/edx-notes-api/defaults/main.yml
```diff
@@ -46,6 +46,7 @@ EDX_NOTES_API_REPOS:
     REPO: edx-notes-api.git
     VERSION: e0d/migration-overrides
     DESTINATION: "{{edx_notes_api_code_dir}}"
+    SSH_KEY: !!null
```
feanil: Not having this be set to a variable will make it very hard to override this key without overriding the entire repos dictionary?  Why not just have a top level var called `EDX_NOTES_API_REPOS_SSH_KEY` which this var can reference?

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -106,17 +106,23 @@
...
+  fail: msg='REPOS.PROTOCOL must be "https" or "ssh"'
   when: (item.PROTOCOL != "https") and (item.PROTOCOL != "ssh")
   with_items: edx_service_repos
 
+- name: install read-only ssh key
+  copy: >
+    content="{{ item.SSH_KEY }}" dest={{ item.DESTINATION }}/id_rsa.pub
+    owner={{ edx_service_user }} group={{ edx_service_user }} mode=0600
+  when: item.PROTOCOL == "ssh"
+  with_items: edx_service_repos
```
feanil: This task will only ensure that the key for the last repo in the list is saved to `id_rsa.pub`.  I think we don't need to support seperate keys per repo.  Maybe just one per service?

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -106,17 +106,23 @@
...
-- name: validate GIT.PROTOCOL
-  fail: msg='GIT.PROTOCOL must be "https" or "ssh"'
+- name: validate git protocol
+  fail: msg='REPOS.PROTOCOL must be "https" or "ssh"'
   when: (item.PROTOCOL != "https") and (item.PROTOCOL != "ssh")
   with_items: edx_service_repos
 
+- name: install read-only ssh key
+  copy: >
+    content="{{ item.SSH_KEY }}" dest={{ item.DESTINATION }}/id_rsa.pub
```
feanil: This should `id_rsa` not `id_rsa.pub` right?  Since this is the private key?

feanil:
:+1:

feanil:
:+1:

PR #1813, opened 2015-01-31
Test with LanguageTest class; RegistrationTest class no longer exists.

benpatterson:
Currently this test file is used when building Jenkins workers.


feanil:
:+1:

PR #1815, opened 2015-02-02
Deal with the fact that the AWS API is eventually consistent.

feanil:
Sometimes we get an instance id back from the creation call but then
get an exception saying the ID was not found when we do a get_all_instances.


maxrothman:
:+1:

PR #1818, opened 2015-02-04
Vagrantfile and Ansible playbook for ecomstack

rlucioni:
This adds a Vagrantfile and Ansible playbook used for creating an "ecomstack," a Vagrant machine which the eCommerce Architecture team can use for development. I'm able to run edx-platform's Python and JavaScript unit tests on the box, as well as acceptance tests. This PR doesn't include the "ecom" (or "ecommerce") role needed to install our new ecommerce service on the new machine.

The new Vagrantfile mounts a synced folder for [django-oscar-extensions](https://github.com/edx/django-oscar-extensions) on the machine so that active development of django-oscar-extensions doesn't require re-installing it after every change. Oscar itself should be static enough that we'll be able to get away with pip installing it.

I also modified the "local_dev" role so that provisioning doesn't fail when trying to set up user accounts on machines without forums, ora, or notifier.

@feanil, could you please review this when you're able? @stephensanchez, @dianakhuang, and @wedaly, FYI.

### DIFF
--- a/playbooks/roles/local_dev/defaults/main.yml
+++ b/playbooks/roles/local_dev/defaults/main.yml
@@ -6,16 +6,16 @@ localdev_xauthority: "{{ localdev_home }}/.Xauthority"
 localdev_xvfb_display: ":1"
 
 localdev_accounts:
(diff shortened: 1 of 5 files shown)


feanil:
:+1:

PR #1829, opened 2015-02-09
Add receipt vars to lms_env_config

maxrothman:
As per https://openedx.atlassian.net/browse/DEVOPS-1084

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -124,6 +124,30 @@ EDXAPP_GIT_REPO_DIR: '/edx/var/edxapp/course_repos'
 
 EDXAPP_TENDER_DOMAIN: "None"
 
+#Only address should have newlines
+EDXAPP_PDF_RECEIPT_BILLING_ADDRESS: |
+  Fake Billing Services
+  123 Fake St.
+  Notaplaceshire, XX 00000
+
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -124,6 +124,26 @@ EDXAPP_GIT_REPO_DIR: '/edx/var/edxapp/course_repos'
 
 EDXAPP_TENDER_DOMAIN: "None"
 
+EDXAPP_PDF_RECEIPT_BILLING_ADDRESS: |
```
feanil: Why pipe here and `>` below?

maxrothman: In the example given in the ticket, the address had newlines, while the disclaimer text did not.

feanil: Can you add a comment to that effect so someone doesn't try to fix it for consistency.

feanil:
:+1:

maxrothman:
According to @wedaly, the `EDXAPP_PDF_RECEIPT_COBRAND_LOGO_PATH` should be blank on edx.org, so I set it to an empty string. Should `EDXAPP_PDF_RECEIPT_LOGO_PATH` also be empty by default and be overridden, or is there a functional default?
@e0d @feanil @fredsmith opinions?

feanil:
:+1:

PR #1835, opened 2015-02-11
Refactor insights to use edx_service

maxrothman:


playbooks/roles/insights/defaults/main.yml
```diff
@@ -59,7 +57,7 @@ INSIGHTS_DATABASES:
 #
 # This block of config is dropped into /edx/etc/insights.yml
 # and is read in by analytics_dashboard/settings/production.py
-INSIGHTS_CONFIG:
+INSIGHTS_SERVICE_CONFIG:
```
feanil: There will have to be a corresponding change in our secure vars.

playbooks/roles/insights/tasks/main.yml
```diff
@@ -21,7 +21,115 @@
...
-  when: not INSIGHTS_GIT_IDENTITY
+- name: setup the insights env file
+  template: >
+    src="edx/app/insights/insights_env.j2"
+    dest="{{ insights_app_dir }}/insights_env"
+    owner={{ insights_user }}
+    group={{ insights_user }}
+    mode=0644
 
-- include: deploy.yml tags=deploy
```
feanil, line 27: This means that running with '--tags deploy' will break.  @e0d is this something we're ok with?

playbooks/roles/insights/tasks/tag_ec2.yml
```diff
@@ -1,11 +0,0 @@
----
```
feanil, line 1: Was this removed because the service role does this?

maxrothman, line 1: Yep. See [this](https://github.com/edx/configuration/blob/f37c01a8e64ef6ea5598d2910b2dc16c3de09c51/playbooks/roles/edx_service/tasks/main.yml#L151).

playbooks/roles/insights/defaults/main.yml
```diff
@@ -103,29 +101,36 @@ INSIGHTS_CONFIG:
...
-insights_user: "{{ insights_role_name }}"
+insights_service_name: insights
+insights_user: "{{ insights_service_name }}"
 insights_app_dir: "{{ COMMON_APP_DIR }}/{{ insights_service_name }}"
 insights_home: "{{ COMMON_APP_DIR }}/{{ insights_service_name }}"
-insights_venv_base: "{{ insights_home }}/venvs"
-insights_venv_dir: "{{ insights_venv_base }}/{{ insights_service_name }}"
-insights_venv_bin: "{{ insights_venv_dir }}/bin"
-insights_code_dir: "{{ insights_app_dir }}/edx-analytics-dashboard"
+insights_code_dir: "{{ insights_app_dir }}/edx_analytics_dashboard"
```
feanil, line 133: Why are we changing the name of the repo, I understand having a convention for things we create, but it doesn't make sense to change the name of the repo.

playbooks/roles/insights/tasks/main.yml
e0d, line 27: I don't have a strong feeling about this, I don't think we've been disciplined enough about how we've been using that convention.  But we should discuss.

playbooks/roles/insights/tasks/tag_ec2.yml
e0d, line 1: In the case of multiple repos, last-in-wins.

playbooks/roles/insights/defaults/main.yml
maxrothman, line 133: We're not changing the name of the repo, just the name of the directory in which it's stored. The name of the repo is set [here](https://github.com/edx/configuration/blob/max/git-refactor/playbooks/roles/insights/defaults/main.yml#L117).

feanil:
:+1:

PR #1838, opened 2015-02-12
datetime tag fix for sandbox

arbabnazar:
@e0d , can you please check it, is it necessary to fix the error that we are getting relating to date_time tag on sandbox. Thanks


e0d:
Have you tested, I'd be a little surprised if this made any difference.

PR #1845, opened 2015-02-13
Initial version

e0d:
@maxrothman @feanil


playbooks/roles/edx_ecommerce/defaults/main.yml
```diff
@@ -0,0 +1,135 @@
...
+  DATABASE_OPTIONS:
+    connect_timeout: 10
+  DATABASES: '{{ EDX_ECOMMERCE_DATABASES }}'
+
+EDX_ECOMMERCE_REPOS:
+  - PROTOCOL: "{{ COMMON_GIT_PROTOCOL }}"
+    DOMAIN: "{{ COMMON_GIT_MIRROR }}"
+    PATH: "{{ COMMON_GIT_PATH }}"
+    REPO: edx-ecommerce.git
+    VERSION: "{{ EDX_ECOMMERCE_VERSION }}"
```
maxrothman, line 72: You might want define each version in separate external variables so they can be individually overridden.

playbooks/roles/edx_ecommerce/defaults/main.yml
```diff
@@ -0,0 +1,135 @@
...
+#
+edx_ecommerce_environment:
+  DJANGO_SETTINGS_MODULE: "settings.production"
+  EDX_ECOMMERCE_SERVICE_CONFIG: "{{ COMMON_CFG_DIR  }}/{{ edx_ecommerce_service_name }}.yml"
+
+edx_ecommerce_service_name: "edx_ecommerce"
+edx_ecommerce_user: "{{ edx_ecommerce_service_name }}"
+edx_ecommerce_home: "{{ COMMON_APP_DIR }}/{{ edx_ecommerce_service_name }}"
+edx_ecommerce_code_dir: "{{ edx_ecommerce_home }}/{{ edx_ecommerce_service_name }}"
+edx_ecommerce_conf_dir: "{{ edx_ecommerce_home }}"
```
maxrothman: This variable doesn't seem to be referenced anywhere.

playbooks/roles/edx_ecommerce/handlers/main.yml
```diff
@@ -0,0 +1,18 @@
+---
```
maxrothman: This notifier doesn't seem to be used anywhere.

playbooks/roles/edx_ecommerce/templates/edx/app/edx_ecommerce/edx_ecommerce.sh.j2
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/env bash
```
maxrothman, line 1: No ansible_managed?

playbooks/roles/edx_ecommerce/defaults/main.yml
```diff
@@ -0,0 +1,135 @@
...
+  # rw user
+  default:
+    ENGINE: 'django.db.backends.mysql'
+    NAME: 'edx_ecommerce'
+    USER: 'ecomm001'
+    PASSWORD: 'password'
+    HOST: 'localhost'
+    PORT: '3306'
+  # read-only user
+  reports:
```
feanil: ecommerce won't need a reports DB will it?

playbooks/roles/edx_ecommerce/defaults/main.yml
```diff
@@ -0,0 +1,135 @@
...
+edx_ecommerce_user: "{{ edx_ecommerce_service_name }}"
+edx_ecommerce_home: "{{ COMMON_APP_DIR }}/{{ edx_ecommerce_service_name }}"
+edx_ecommerce_code_dir: "{{ edx_ecommerce_home }}/{{ edx_ecommerce_service_name }}"
+edx_ecommerce_conf_dir: "{{ edx_ecommerce_home }}"
+
+edx_ecommerce_gunicorn_host: "127.0.0.1"
+edx_ecommerce_gunicorn_port: "8130"
+edx_ecommerce_gunicorn_timeout: "300"
+
+edx_ecommerce_django_settings: "production"
```
feanil: This one is not referenced anywhere either.

maxrothman:
Nitpick: why does the role name need to be prefixed by "edx"? We have things like forums, insights, and analytics_api that aren't prefixed. It seems overly verbose.

playbooks/roles/edx_ecommerce/defaults/main.yml
e0d: I assume it will need some DB and have left that stanza as a place-holder.

e0d:
@maxrothman tend to agree re verbosity, but I think it makes sense to have our roles match the name of the repo.  That's not always the case, but would like to make it the convention going forward.

maxrothman:
@e0d Agree on the verbosity point.

playbooks/roles/edx_ecommerce/defaults/main.yml
basetta, line 72: I know it is not my business :) But you can override a single value of a dictionary iif you specify in `ansible.cfg` the setting `hash_behaviour = merge`

feanil, line 72: @basetta  It's open for a reason, the more eyes the better.  One of the reasons we don't do that is because, hash merging behavior requires a better mental model of precedence.  It is harder for people just starting out with ansible to reason about it and it is harder to debug at which "merge" variables are being overridden.

e0d, line 72: It's an open source project, it's your business if you want it to be :)  That said, enabling `hash_behaviour = merge` is a bigger change and I personally don't like override vars using sparse hashes.  What we have is sub-optimal, but I think the alternative is worse.

carsongee, line 72: Just to add some history, we had that turned on not too long ago and I removed it. https://github.com/edx/configuration/commit/3b4193706b7655faaa11cf100f693b025d9f1bce#diff-c65aaeaf20dfc7d75fca73c229cb9e85

basetta, line 72: I see your point and it makes sense. Let's see how they people digests the sparse hashes here. :)

PR #1847, opened 2015-02-13
refactor minos to use new edx_service role

maxrothman:
NB: This includes some small changes to edx_service that allow it to be called without any repos. Please review carefully.


maxrothman:
Another way of implementing the edx_service change would've been to just pass an empty `edx_service_repos` list. Thoughts?

maxrothman:
Two notes:
- This role depends on variables set in `common`, though it does not have `common` as a dependency. Right now this isn't an issue since in most plays common gets run _somewhere_, but once `common` just sets variables, it should probably be added.
- `minos` now uses `MINOS_GIT_IDENTITY` to get stuff from git, so that'll need to be overridden to `COMMON_GIT_IDENTITY` somewhere.

feanil:
MINOS_GIT_IDENTITY is set to COMMON_GIT_IDENTITY in our secret vars.

feanil:
:+1:

PR #1848, opened 2015-02-13
Add Vagrant box links for Birch RC3

singingwolfboy:
@maxrothman


maxrothman:
Only the fullstack images exist so far. Is that intentional?

maxrothman:
:+1:

PR #1852, opened 2015-02-17
Updated Location of Insights manage.py

clintonb:
@dsjen @feanil @e0d


maxrothman:
This will conflict with the refactor on [e0d/edx-notes-api](https://github.com/edx/configuration/tree/e0d/edx-notes-api). Not a reason not to merge, but we should keep it in mind.

feanil:
:+1:   @maxrothman can you make a corresponding change to this on the notes branch?

clintonb:
Confirmed that server starts on sandbox and that AMI is built.

PR #1853, opened 2015-02-17
enable insights to be deployed via ansible-provision

fredsmith:


clintonb:
:+1:

feanil:
:+1:   Is this sufficient to run insights on sandbox?

fredsmith:
@feanil I have another PR against sandbox-secure that I'm working on.  any other concerns?

feanil:
If we don't update the base sandbox play, the default sandbox won't have insights by default.

https://github.com/edx/configuration/blob/master/playbooks/edx_sandbox.yml

fredsmith:
ah, I thought https://github.com/edx/configuration/blob/master/util/jenkins/ansible-provision.sh#L244 would deploy it, but that's only if $reconfigure != "true".

PR #1856, opened 2015-02-18
use pip instead of pip-accel for jenkins workers

clytwynec:
@jzoldak @benpatterson 

The version of pip was updated recently in the edx-platform and it now conflicts with the one needed for pip-accel, which was preventing provisioning of a new jenkins_worker ami.  This just reverts to using pip instead of pip-accel.

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/python.yml
+++ b/playbooks/roles/jenkins_worker/tasks/python.yml
@@ -30,24 +30,30 @@
     depth=1
   sudo_user: "{{ jenkins_user }}"
 
-- name: Pip installs that are needed for pip-accel to work for us
-  pip: >
-    name="{{ item }}"
-    virtualenv={{ jenkins_home }}/edx-venv
-    virtualenv_command=virtualenv-2.7
-  sudo_user: "{{ jenkins_user }}"
-  with_items: pip_accel_reqs
+# pip-accel skipped due to conflicting versions of pip required
+# by the pip-accel package and edx-platform
+# - name: Pip installs that are needed for pip-accel to work for us
+#   pip: >
+#     name="{{ item }}"
+#     virtualenv={{ jenkins_home }}/edx-venv
+#     virtualenv_command=virtualenv-2.7
+#   sudo_user: "{{ jenkins_user }}"
+#   with_items: pip_accel_reqs
 
-# Install the platform requirements using pip-accel.
-# This will allow the binary distributions to be downloaded from S3
-# rather than compiled each time.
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,24 +30,29 @@
...
+#   sudo_user: "{{ jenkins_user }}"
+#   with_items: pip_accel_reqs
 
-# Install the platform requirements using pip-accel.
-# This will allow the binary distributions to be downloaded from S3
-# rather than compiled each time.
-- name: Install edx-platform requirements using pip-accel
+# Install the platform requirements using pip.
+# Installing the platform requirements using pip-accel
+# woult allow the binary distributions to be downloaded from S3
```
jzoldak: typo "would"

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,24 +30,29 @@
...
+# rather than compiled each time. This was previously enabled,
+# but reverted back to pip because pip-accel is not compatible
+# with the recently updated version of pip in the platform.
```
jzoldak: Maybe be more specific like:
the current version of pip-accel (0.22.4) is only compatible with pip >= 1.4, < 1.5
and the current version of pip in the edx-platform is 1.5

playbooks/roles/jenkins_worker/tasks/python.yml
clytwynec: I had fixed this locally earlier but forgot to save :-(   Just pushed the fix

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -30,24 +30,29 @@
...
 
-# Install the platform requirements using pip-accel.
-# This will allow the binary distributions to be downloaded from S3
-# rather than compiled each time.
-- name: Install edx-platform requirements using pip-accel
+# Install the platform requirements using pip.
+# Installing the platform requirements using pip-accel
+# would allow the binary distributions to be downloaded from S3
+# rather than compiled each time. This was previously enabled,
+# but reverted back to pip because pip-accel is not compatible
```
jzoldak: Maybe be more specific like:
the current version of pip-accel (0.22.4) is only compatible with pip >= 1.4, < 1.5
and the current version of pip in the edx-platform is 1.5

jzoldak:
:+1:

benpatterson:
:+1:

PR #1859, opened 2015-02-19
Add a migration script for Birch

singingwolfboy:
@maxrothman, @feanil, can you two take a look at this? Both the contents of the script, and the general idea of having this script here in this repo.

### DIFF
--- a/util/vagrant/migrate.sh
+++ b/util/vagrant/migrate.sh
@@ -1,0 +1,100 @@
+#!/usr/bin/env bash
+
+# defaults
+CONFIGURATION="fullstack"
+TARGET="named-release/birch"
+INTERACTIVE=true
+
+read -d '' HELP_TEXT <<- EOM
+Attempts to migrate your Open edX installation to a different release.
+
+-c CONFIGURATION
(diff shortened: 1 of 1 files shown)


feanil:
Can we move this under 'util/devstack' instead of creating a new top level directory?

singingwolfboy:
@feanil: So, you want me to make a `devstack` directory under the `util` directory, and put this `migrate.sh` script there? It's designed to work with fullstack, too, so I'm not sure that a directory named `devstack` is a good choice. Maybe `util/vagrant`?

feanil:
`util/vagrant` sounds good to me.

singingwolfboy:
@feanil I've updated the path to the script. Better?

feanil:
:+1:

feanil:
nit: maybe rename it to 'migrate-aspen-to-birch.sh'?  As we'll have more in the future?

singingwolfboy:
@feanil I wrote the script with support for specifying the migration target, so that (in theory) you'll be able to reuse the same migration script to get to a future named release.

PR #1861, opened 2015-02-20
Explicitly set MTU for interfaces to avoid jumbo frame bug.

e0d:
@feanil prior commit didn't work for DHCP, this does and sets the MTU for running hosts as well.


feanil:
:+1:

PR #1865, opened 2015-02-20
One build only.

feanil:


fredsmith:
is that like "one ping only"?  :+1:

maxrothman:
Nitpick: you're using a 4-space indent, while the rest of the document uses a 2-space indent.

PR #1870, opened 2015-02-26
Updates to get ecom service up.

feanil:


feanil:
@e0d

playbooks/edx-east/edx_service.yml
```diff
@@ -3,7 +3,7 @@
 - name: Build application artifacts
   hosts: all
   connection: local
-  gather_facts: True
+  gather_facts: False
```
maxrothman, line 6: Are you sure that we don't use gathered facts in any of our plays?

feanil, line 6: Yes, I've run these plays with this flag to set to false, when I was testing.

maxrothman:
:+1:

PR #1872, opened 2015-02-27
Installing OpenJDK on Insights Servers

clintonb:
This is needed to make use of Closure Compiler for JS minification.


clintonb:
This is needed before merging edx/edx-analytics-dashboard#267.

@maxrothman @e0d @fredsmith @feanil

maxrothman:
Shouldn't a corresponding redhat package be added too?

clintonb:
@maxrothman Updated.

PR #1875, opened 2015-03-01
Increase RAM to avoid out of memory in stacks

yrchen:
After Birch release, the default services use over 2G RAM and 4xxMB swap after system boot. I suggest to increase RAM setting for devstack and fullstack environments to avoid out of memory issue.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -39,3 +39,4 @@ Brandon DeRosier <bdero@mit.edu>
 Arbab Nazar <arbab@edx.org>
 Nilesh Londhe <nilesh@cloudgeni.us>
 Riccardo Murri <riccardo.murri@uzh.ch>
+Xaver Y.R. Chen <yrchen@ATCity.org>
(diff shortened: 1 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @yrchen! I've created [OSPR-433](https://openedx.atlassian.net/browse/OSPR-433) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

sarina:
@singingwolfboy - this seem right to you? does ops need to review this as well?

singingwolfboy:
This looks right to me, but it should _definitely_ get an ops review as well.

feanil:
Agree with @e0d we need a better long term plan but this is a good solution to the immediate issue.

sarina:
@feanil can you or someone else from Ops merge this PR?

PR #1878, opened 2015-03-03
Add splunk forwarder role to jenkins

jzoldak:
@feanil @benpatterson

### DIFF
--- a/playbooks/edx-east/jenkins_testeng_master.yml
+++ b/playbooks/edx-east/jenkins_testeng_master.yml
@@ -9,8 +9,36 @@
   vars:
     COMMON_DATA_DIR: "/mnt"
     COMMON_ENABLE_DATADOG: True
+    COMMON_ENABLE_SPLUNKFORWARDER: True
+
+    SPLUNKFORWARDER_LOG_ITEMS:
+      - source: '/var/lib/jenkins/jobs/*/builds/.../junitResult.xml'
+        recursive: true
+        index: 'testeng'
+        sourcetype: junit
+        followSymlink: false
+      - source: '/var/lib/jenkins/jobs/*/builds/.../build.xml'
+        index: 'testeng'
+        recursive: true
+        sourcetype: build_result
+        followSymlink: false
+        initCrcLength: 1024
+      - source: '/var/lib/jenkins/jobs/*/builds/.../log'
+        index: 'testeng'
+        recursive: true
+        sourcetype: build_log
+        followSymlink: false
+        initCrcLength: 2048
+
   roles:
     - common
     - role: datadog
       when: COMMON_ENABLE_DATADOG
     - jenkins_master
+
+    # run just the splunkforwarder role by using '--tags "splunkonly"'
+    # e.g. ansible-playbook jenkins_testeng_master.yml -i inventory.ini --tags "splunkonly" -vvvv
+    - role: splunkforwarder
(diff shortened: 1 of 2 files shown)


benpatterson:
Hey...noticed this on my stream. Does the props.conf file need to include the build log, too? I forget if build.xml included the console output.

jzoldak:
@benpatterson Not in this PR. I need to add the output log to both inputs.conf (so it will start monitoring) and also to props.conf (if it needs anything other than the default parsing). I haven't figured that out yet, so I thought I'd get the framework moving with the two that I have nailed down so far.

benpatterson:
Ok makes sense...

playbooks/inventory.ini
```diff
@@ -1,2 +1,5 @@
 [localhost]
 127.0.0.1
+
```
benpatterson: I'm guessing this file shouldn't have been included in this commit...

playbooks/edx-east/jenkins_testeng_master.yml
```diff
@@ -9,8 +9,39 @@
   vars:
     COMMON_DATA_DIR: "/mnt"
     COMMON_ENABLE_DATADOG: True
+    COMMON_ENABLE_SPLUNKFORWARDER: True
+    SPLUNKFORWARDER_LOG_ITEMS:
```
benpatterson: I'm wondering if any of these settings belong in the jenkins_master role, under the meta/default.yml file. No strong preference so that's more a question for @feanil

benpatterson:
A couple questions but changes LGTM. :+1:

playbooks/inventory.ini
jzoldak: oh, snaps!
thx.

playbooks/edx-east/jenkins_testeng_master.yml
feanil: These are specific to the files and paths that are needed from the testeng jenkins.  We don't need to monitor the same locations on other jenkins instances so I think it is fine to put it here.

playbooks/edx-east/jenkins_testeng_master.yml
```diff
@@ -9,8 +9,39 @@
...
+      - source: '/var/lib/jenkins/jobs/*/builds/*/junitResult*.xml'
```
benpatterson: Not a must, but would it make sense (is it possible?) to re-use this setting for the paths? https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_master/defaults/main.yml#L1

jzoldak: @benpatterson I just looked into this and it's a bit tricky because of the sym link from /mnt/jenkins to /var/lib/jenkins. That is, the value of {{ jenkins_home }} is '/mnt/jenkins' but that endpoint does not exist on our current jenkins instances. Since it would take some effort troubleshooting, I think we should defer this refactoring.

benpatterson: Ok cool. Thanks for looking at it and I agree....defer! But thanks for the look...good to know.

jzoldak:
FYI I just back out the props.conf changes, as in testing it appears that these need to be applied on the indexer, not the forwarder.

feanil:
:+1:

PR #1882, opened 2015-03-04
Max/merge refactor

maxrothman:
### Summary:
- Add more functionality to the `edx_service` role, especially handling git checkouts and ec2 tagging
- Change `edx_service`'s interface with dependent roles to be more maintainable
- Refactor all dependent roles to use this functionality (`minos`, `xqwatcher`, `analytics_api`, and `insights`)
- Add a new role, `edx_notes_api` for the forthcoming notes api
- Bring the aforementioned roles into line with our coding standards, including removing unnecessary variables, false assumptions, and changing (almost) all dash word separators to underscores
- A host of other small fixes (properly handling missing remote git refs, etc.)
### Passing tests:
- insights & analyticsapi:
  - ok according to analytics team
- xqwatcher:
  - all jenkins jobs run
  - pulls stuff from xqueue
- minos
  - tracking logs are moved to s3
### Additional notes:
- adds a custom ec2_tag module, pending a PR against ansible
- minos task "maintain backwards-compatable config location" should be removed after merge
- pre_supervisor doesn't work with xqwatcher
- before prod released, change the previous ASG's services tag from "analytics-api" to "analytics_api"


maxrothman:
@fredsmith @feanil @e0d

playbooks/library/ec2_tag_local
```diff
@@ -0,0 +1,169 @@
+#!/usr/bin/python
```
feanil, line 1: What was the exact reason for pulling this in again?

playbooks/roles/common/defaults/main.yml
```diff
@@ -31,7 +31,9 @@ COMMON_PYPI_MIRROR_URL: 'https://pypi.python.org/simple'
 COMMON_NPM_MIRROR_URL: 'http://registry.npmjs.org'
 COMMON_UBUNTU_APT_KEYSERVER: "http://keyserver.ubuntu.com/pks/lookup?op=get&fingerprint=on&search="
 # do not include http/https
```
feanil: Can you add a comment here with some more details especially about what the 'COMMON_GIT_PATH' is so if open source users want to change these, they have a better idea of what they are?

playbooks/roles/edx-shared-vars/defaults/main.yml
```diff
@@ -0,0 +1,88 @@
+---
```
feanil: These look like the defaults for the `common` role are being pulled out to this new role.  But I don't see the corresponding removal of these vars from the `common` role.

CHANGELOG.md
```diff
@@ -1,6 +1,13 @@
 - Role: edxapp
   - Enabled combined login registration feature by default
 
+- Role: analytics_api
+  - Refactored name from `analytics-api` to `analytics_api`. This will require additional migrations when upgrading an existing server. While we recommend building from scratch, running the following command might work:
+  
+      ```
+      rm -rf /edx/app/analytics-api /edx/app/ /edx/app/nginx/sites-available/analytics-api.j2 /edx/app/supervisor/conf.d.available/analytics_api.conf
+      ```
+
```
feanil, line 16: Is this the only backwards incompatible change worth mentioning?
Also can we explicitly state that there was a backwards incompatible change here?
Can you also add more information about the refactor in general since it is a fairly large change?

playbooks/library/ec2_tag_local
maxrothman, line 1: It's needed for [this task](https://github.com/edx/configuration/blob/max/merge-refactor/playbooks/roles/edx_service/tasks/main.yml#L152) because ansible won't template YAML keys.

feanil, line 1: Ah, right.

playbooks/roles/edx-shared-vars/defaults/main.yml
maxrothman: I believe this change was made by @e0d

feanil:
:+1:  once we resolve the issue around the `edx-shared-vars` role.

e0d:
:+1:

PR #1889, opened 2015-03-09
Only run the migration with the migration credentials.

feanil:
Take a look at the task above this one which runs migrations using the migration credentials.
@e0d


e0d:
:+1:

PR #1890, opened 2015-03-09
Ansible fails in looking up keys inside the dictionary when the top leve...

feanil:
...l var does not exist.


maxrothman:
Are there other motivations for pulling the DB name out of that dict, or is it just to get this part functional?

feanil:
It is necessary for this to be functional.

maxrothman:
Is there a way this could be accomplished with an "is defined" conditional instead?

feanil:
Not that I know of.  That is the whole reason we use the pattern with defaults.

maxrothman:
:+1:

PR #1891, opened 2015-03-09
Adjust index.conf splunk settings for jenkins monitors

jzoldak:
Add the the full directory path to the source file to the CRC. This ensures that each file being monitored has a unique CRC. See splunk docs on [inputs.conf](http://docs.splunk.com/Documentation/Splunk/latest/Admin/Inputsconf)

@benpatterson @clytwynec


jzoldak:
@benpatterson or @clytwynec - let me know if you need further details for reviewing.
FWIW this is what I had already applied to our splunk server, so it's what's running in production.

benpatterson:
:+1:

PR #1898, opened 2015-03-12
Feanil/ecomstack updates

feanil:
Vagrant updates that I didn't want conflated with the other ecom changes.

Review: @maxrothman @fredsmith @e0d 

FYI: @rlucioni @clintonb 

Needs to be merged at the same time as https://github.com/edx/ecommerce/pull/21


playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -63,6 +63,7 @@
     owner="{{ edx_service_name }}"
     group="{{ common_web_group }}"
   with_items:
+    - ""
```
maxrothman, line 71: What's this for?

feanil, line 71: We are never explicitly creating the 'edx_service_home' directory and in vagrant at least, this directory was being created with owner and group `root`.  This insures that the service home directory exists and is owned by the correct service user and group.

PR #1905, opened 2015-03-17
Make ansible-provision.sh fail on any failed ansible runs

maxrothman:


feanil:
:+1: did you test this branch with the ansible-provision jenkins job?

maxrothman:
Yep, tested with no errors, and tested with errors. It succeeds with no errors and fails early with errors and marks the build as failed.

PR #1910, opened 2015-03-18
Add newrelic monitoring for edxapp workers

maxrothman:


feanil:
Is it easy to tell which queue a particular error is from or would it be better to give each worker a different app name?

maxrothman:
I think @e0d wanted to limit the number of applications for cost reasons.

e0d:
I don't think we would be charged for in this case, but an app per queue feels like overkill.  I'd vote for seeing how it goes and responding if there are issues.

feanil:
That sounds fine to me, I was more curious if we can get queue name info when we see stack traces or transactions.

playbooks/roles/edxapp/templates/workers.conf.j2
```diff
@@ -1,13 +1,13 @@
 {% for w in edxapp_workers %}
 [program:{{ w.service_variant }}_{{ w.queue }}_{{ w.concurrency }}]
 
-environment=CONCURRENCY={{ w.concurrency }},LOGLEVEL=info,DJANGO_SETTINGS_MODULE=aws,PYTHONPATH={{ edxapp_code_dir }},SERVICE_VARIANT={{ w.service_variant }}
+environment={% if COMMON_ENABLE_NEWRELIC_APP %}NEW_RELIC_APP_NAME={{ EDXAPP_NEWRELIC_WORKERS_APPNAME }},NEW_RELIC_LICENSE_KEY={{ NEWRELIC_LICENSE_KEY }},{% endif -%}CONCURRENCY={{ w.concurrency }},LOGLEVEL=info,DJANGO_SETTINGS_MODULE=aws,PYTHONPATH={{ edxapp_code_dir }},SERVICE_VARIANT={{ w.service_variant }}
```
feanil: This should be `COMMON_ENABLE_NEWRELIC`.

https://github.com/edx/configuration/blob/master/playbooks/roles/common/defaults/main.yml#L70

playbooks/roles/edxapp/templates/workers.conf.j2
```diff
@@ -1,13 +1,13 @@
...
 user={{ common_web_user }}
 directory={{ edxapp_code_dir }}
 stdout_logfile={{ supervisor_log_dir }}/%(program_name)s-stdout.log
 stderr_logfile={{ supervisor_log_dir }}/%(program_name)s-stderr.log
 
-command={{ edxapp_venv_bin}}/python {{ edxapp_code_dir }}/manage.py {{ w.service_variant }} --settings=aws celery worker --loglevel=info --queues=edx.{{ w.service_variant }}.core.{{ w.queue }} --hostname=edx.{{ w.service_variant }}.core.{{ w.queue }}.%%h --concurrency={{ w.concurrency }}
+command={{ edxapp_venv_dir + '/bin/newrelic-admin run-program ' if w.monitor and COMMON_ENABLE_NEWRELIC_APP else ''}}{{ edxapp_venv_bin }}/python {{ edxapp_code_dir }}/manage.py {{ w.service_variant }} --settings=aws celery worker --loglevel=info --queues=edx.{{ w.service_variant }}.core.{{ w.queue }} --hostname=edx.{{ w.service_variant }}.core.{{ w.queue }}.%%h --concurrency={{ w.concurrency }}
```
feanil: Same as above here. It should be `COMMON_ENABLE_NEWRELIC`

feanil:
:+1:  once this is validated in the sandbox environment.

PR #1914, opened 2015-03-23
antivirus doesnt need bastion users

fredsmith:


maxrothman:
Are you sure you don't need the user role at all? This should be tested.

feanil:
:+1:

jp-bpl:
Would it also be possible to make it AWS independent? Now the antivirus role is constructed for the use with S3 buckets only.

PR #1922, opened 2015-03-24
A play to do what fabric currently does for the certs servers.

feanil:


playbooks/edx-east/legacy_certificates.yml
```diff
@@ -0,0 +1,34 @@
...
+- name: Deploy legacy_certs
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    serial_count: 1
+    repo_url: "git@github.com/edx/certificates.git"
+    repo_path: "/opt/wwc/certificates"
+    certificates_version: "master"
+    git_ssh_script: "/tmp/git.sh"
```
e0d, line 11: What creates this?

e0d:
One question, but LGTM

playbooks/edx-east/legacy_certificates.yml
feanil, line 11: It was created by puppet long ago and is on the images we are using.

PR #1924, opened 2015-03-25
Add edxapp profile image config

maxrothman:
@jimabramson @dan-f

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -178,7 +178,7 @@ EDXAPP_BOOK_URL: ""
 EDXAPP_SITE_NAME: 'localhost'
 EDXAPP_LMS_SITE_NAME: "{{ EDXAPP_SITE_NAME }}"
 EDXAPP_CMS_SITE_NAME: 'localhost'
-EDXAPP_MEDIA_URL:  ""
+EDXAPP_MEDIA_URL:  "/media"
 EDXAPP_ANALYTICS_SERVER_URL:  ""
 EDXAPP_FEEDBACK_SUBMISSION_EMAIL: ""
 EDXAPP_CELERY_BROKER_HOSTNAME: ""
@@ -443,6 +443,35 @@ PERFORMANCE_GRAPHITE_URL: 'SetPerformanceGraphiteHostName'
 EDXAPP_ECOMMERCE_API_URL: 'http://localhost:18130'
 EDXAPP_ECOMMERCE_API_SIGNING_KEY: 'SET-ME-PLEASE'
 
+#To use AWS S3 as your backend, you need different kwargs:
+# EDXAPP_PROFILE_IMAGE_BACKEND_CONFIG:
+#   storage_class: storages.backends.s3boto.S3BotoStorage
+#   storage_kwargs:
+#     location: /path/to/images
+#     bucket: mybucket
+#     custom_domain: mybucket.s3.amazonaws.com
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+#     headers:
+#       Cache-Control: max-age-{{ EDXAPP_PROFILE_IMAGE_MAX_AGE }}
+#NB: access_key and secret_key are unneccessary if you use IAM roles
+#NB2: custom_domain is REQUIRED. Otherwise, boto will generate a 
+# temporary URL whenever asked for the URL of a specific file.
+EDXAPP_PROFILE_IMAGE_BACKEND:
+  class: storages.backends.overwrite.OverwriteStorage
+  options:
+    location: "{{ edxapp_media_dir }}/profile-images/"
+    base_url: "{{ EDXAPP_MEDIA_URL }}/profile-images/"
+
+EDXAPP_PROFILE_IMAGE_MAX_AGE: 31536000
+
+# used to salt hashed usernames for profile image filenames
+EDXAPP_PROFILE_IMAGE_SECRET_KEY: placeholder_secret_key
+
+# In bytes
+EDXAPP_PROFILE_IMAGE_MAX_BYTES: 1048576
+EDXAPP_PROFILE_IMAGE_MIN_BYTES: 100
+
 #-------- Everything below this line is internal to the role ------------
 
 #Use YAML references (& and *) and hash merge <<: to factor out shared settings
@@ -464,6 +493,7 @@ edxapp_node_bin: "{{ edxapp_code_dir }}/node_modules/.bin"
 edxapp_user: edxapp
 edxapp_deploy_path: "{{ edxapp_venv_bin }}:{{ edxapp_code_dir }}/bin:{{ edxapp_rbenv_bin }}:{{ edxapp_rbenv_shims }}:{{ edxapp_gem_bin }}:{{ edxapp_node_bin }}:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin"
 edxapp_staticfile_dir: "{{ edxapp_data_dir }}/staticfiles"
+edxapp_media_dir: "{{ edxapp_data_dir }}/media"
 edxapp_course_static_dir: "{{ edxapp_data_dir }}/course_static"
 edxapp_course_data_dir: "{{ edxapp_data_dir }}/data"
 edxapp_upload_dir: "{{ edxapp_data_dir }}/uploads"
@@ -678,7 +708,11 @@ generic_env_config:  &edxapp_generic_env
   WIKI_ENABLED: true
   SYSLOG_SERVER:  "{{ EDXAPP_SYSLOG_SERVER }}"
   LOG_DIR:  "{{ COMMON_DATA_DIR }}/logs/edx"
-  MEDIA_URL:  "{{ EDXAPP_MEDIA_URL }}"
+  
+  #must end in slash (https://docs.djangoproject.com/en/1.4/ref/settings/#media-url)
+  MEDIA_URL:  "{{ EDXAPP_MEDIA_URL }}/"
+  MEDIA_ROOT: "{{ edxapp_media_dir }}/"
+
   ANALYTICS_SERVER_URL:  "{{ EDXAPP_ANALYTICS_SERVER_URL }}"
   FEEDBACK_SUBMISSION_EMAIL: "{{ EDXAPP_FEEDBACK_SUBMISSION_EMAIL }}"
   TIME_ZONE: "{{ EDXAPP_TIME_ZONE }}"
@@ -758,6 +792,7 @@ lms_auth_config:
   CC_PROCESSOR_NAME: "{{ EDXAPP_CC_PROCESSOR_NAME }}"
   CC_PROCESSOR: "{{ EDXAPP_CC_PROCESSOR }}"
   TRACKING_SEGMENTIO_WEBHOOK_SECRET: "{{ EDXAPP_TRACKING_SEGMENTIO_WEBHOOK_SECRET }}"
+  PROFILE_IMAGE_SECRET_KEY: "{{ EDXAPP_PROFILE_IMAGE_SECRET_KEY }}"
 
 lms_env_config:
   <<: *edxapp_generic_env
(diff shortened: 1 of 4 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -443,6 +443,21 @@ PERFORMANCE_GRAPHITE_URL: 'SetPerformanceGraphiteHostName'
...
+# EDXAPP_PROFILE_IMAGE_BACKEND_CONFIG:
+#   storage_class: storages.backends.s3boto.S3BotoStorage
+#   storage_kwargs:
+#     location: /path/to/images
+#     bucket: mybucket
+#     custom_domain: mycdn.org
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+EDXAPP_PROFILE_IMAGE_BACKEND_CONFIG:
+  storage_class: storages.backends.overwrite.OverwriteStorage
```
feanil: Is this a valid storage class? ie. Will this work on sandboxes when using defaults?

fredsmith:
:+1:

playbooks/roles/edxapp/defaults/main.yml
maxrothman: Yep, it's taken directly from (here|https://github.com/edx/edx-platform/pull/7464). In fact, it matches what's in common.py exactly.

feanil:
:+1:

dan-f:
Holding my thumbs-up until we figure out what to do regarding default images and the media vs static discussion.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -443,6 +443,21 @@ PERFORMANCE_GRAPHITE_URL: 'SetPerformanceGraphiteHostName'
...
 
+#To use AWS S3 as your backend, you need different kwargs:
+# EDXAPP_PROFILE_IMAGE_BACKEND_CONFIG:
+#   storage_class: storages.backends.s3boto.S3BotoStorage
+#   storage_kwargs:
+#     location: /path/to/images
+#     bucket: mybucket
+#     custom_domain: mycdn.org
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
```
jimabramson, line 454: i don't think the app wants/needs the keys here (they would get picked up by whatever mechanism boto has already been using)

maxrothman, line 454: They're valid kwargs, at least according to what @dan-f showed me. I feel like it's better to have it there as documentation than not.

jimabramson:
:+1:

dan-f:
:+1:

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -443,6 +443,22 @@ PERFORMANCE_GRAPHITE_URL: 'SetPerformanceGraphiteHostName'
 EDXAPP_ECOMMERCE_API_URL: 'http://localhost:18130'
 EDXAPP_ECOMMERCE_API_SIGNING_KEY: 'SET-ME-PLEASE'
 
+#To use AWS S3 as your backend, you need different kwargs:
+# EDXAPP_PROFILE_IMAGE_BACKEND_CONFIG:
+#   storage_class: storages.backends.s3boto.S3BotoStorage
+#   storage_kwargs:
+#     location: /path/to/images
+#     bucket: mybucket
+#     custom_domain: mycdn.org
```
jimabramson: let's spell out explicitly that this value must be set, even if serving directly from s3 w/o a cdn.  in which case the value should be '{bucket-name}.s3.amazonaws.com'.
the reason is that the s3boto backend, unless a custom_domain is set, will consult with s3 to generate a temporary url whenever asked for the url to a specific file.

maxrothman:
@fredsmith @e0d @feanil rereview?
Related to https://github.com/edx-ops/sandbox-secure/pull/16, https://github.com/edx-ops/edge-secure/pull/100, https://github.com/edx-ops/edx-secure/pull/386

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -443,6 +443,35 @@ PERFORMANCE_GRAPHITE_URL: 'SetPerformanceGraphiteHostName'
...
+#     headers:
+#       Cache-Control: max-age-{{ EDXAPP_PROFILE_IMAGE_MAX_AGE }}
+#NB: access_key and secret_key are unneccessary if you use IAM roles
+#NB2: custom_domain is REQUIRED. Otherwise, boto will generate a 
+# temporary URL whenever asked for the URL of a specific file.
+EDXAPP_PROFILE_IMAGE_BACKEND:
+  class: storages.backends.overwrite.OverwriteStorage
+  options:
+    location: "{{ edxapp_media_dir }}/profile-images"
+    base_url: "{{ EDXAPP_MEDIA_URL }}/profile-images"
```
feanil: Do we need to worry about double slashes here since the MEDIA_URL terminates with a slash already?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -120,6 +120,12 @@ error_page {{ k }} {{ v }};
     try_files $uri @proxy_to_lms_app;
   }
 
+  # Need a separate location for the image uploads endpoint to limit upload sizes
+  location /api/profile_images/[^/]*/[^/]*/upload {
```
feanil: Needs a `~` since you're doing regex matching.  Also we should start it with '^' and end with '$'.

playbooks/roles/edxapp/defaults/main.yml
maxrothman: I'm fairly confident that django will handle path joins gracefully. @dan-f @jimabramson ?

dan-f: I actually think you should be defensive here and not include the redundant slash.

maxrothman: How would **not** including the slash be defensive?

jimabramson: how about using `os.path.join` - everybody wins

maxrothman: +1 go for it. Are you going to implement that?

jimabramson: @dan-f could you add that to https://github.com/edx/edx-platform/pull/7483 ?

dan-f: Because EDXAPP_MEDIA_URL ends in a slash, and I'm not sure what the
expected behavior is when you manually include an additional slash.

dan-f: Whoops sorry - didn't see the more recent messages in my inbox.

jimabramson: `os.path.join` cleans up any missing or duplicated slashes for you, so the
idea would be to use that function instead of string concatenation in
settings modules.

The caveat is, we definitely want to make sure the result ends with a
slash, because that is meaningful - but I'm 99% sure that `os.path.join`
won't remove it.

dan-f: `os.path.join` doesn't  clean up duplicate slashes, but it will insert slashes between its arguments if they're not provided. I'm not seeing how using `os.path.join`in the default declaration of `PROFILE_IMAGE_BACKEND` in my PR will solve this problem. That being said, I'm  happy to add it.

_edit_ some examples:

```
>>> os.path.join('foo/', 'bar')
'foo/bar'
>>> os.path.join('foo', 'bar')
'foo/bar'
>>> os.path.join('foo/', '/bar')
'/bar'
>>> os.path.join('foo', 'bar/')
'foo/bar/'
>>> os.path.join('foo//', 'bar')
'foo//bar'
```

maxrothman: Given `os.path.join`'s behavior, I'm going to establish a convention of not having paths contain trailing slashes. Therefore, I will remove the trailing slash from `EDXAPP_MEDIA_URL`.

Thoughts, @feanil @e0d ?

feanil: That seems reasonable to me, in other cases like this where it may be ambiguous, I've usually left a comment indicating that the trailing slash has be purposefully removed or added.

jimabramson: sorry, i think i had earlier conflated `os.path.join` and `os.path.normpath`

also, FWIW this particular case is one in which a trailing slash after
"profile-images" is actually important

maxrothman: @jimabramson why is the trailing slash important?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -779,6 +811,10 @@ lms_env_config:
   PDF_RECEIPT_TAX_ID_LABEL: "{{ EDXAPP_PDF_RECEIPT_TAX_ID_LABEL }}"
   PDF_RECEIPT_COBRAND_LOGO_PATH: "{{ EDXAPP_PDF_RECEIPT_COBRAND_LOGO_PATH }}"
   PDF_RECEIPT_LOGO_PATH: "{{ EDXAPP_PDF_RECEIPT_LOGO_PATH }}"
+  PROFILE_IMAGE_BACKEND: "{{ EDXAPP_PROFILE_IMAGE_BACKEND }}"
+  PROFILE_IMAGE_MIN_BYTES: "{{ EDXAPP_PROFILE_IMAGE_MIN_BYTES }}"
+  PROFILE_IMAGE_MAX_BYTES: "{{ EDXAPP_PROFILE_IMAGE_MAX_BYTES }}"
```
dan-f: I've been having issues uploading images to my sandbox. After debugging, turns out that PROFILE_IMAGE_MIN_BYTES is ending up as a string (rather than an int/number) in `lms.env.json`.

playbooks/roles/edxapp/defaults/main.yml
jimabramson: base urls in django settings must end with slashes if non-empty (see doc e.g. https://docs.djangoproject.com/en/1.4/ref/settings/#media-url)

the examples for filesystem root configurations also all end in slashes in the docs; while this isn't specified as being required in the docs i have observed incorrect behavior when they are omitted.

maxrothman: @jimabramson The latest commit resolves this by adding a slash to the structure that ends up in the env files ([see this](https://github.com/edx/configuration/pull/1924/files#diff-9bffb77ca37dcc0aead699b822b29095R713)). How's that look?

jimabramson: @maxrothman cool, works for me.  note that `location` and `base_url` (as in the above lines) still need to end with slashes as well, in case this is not yet accounted for.

jimabramson:
:+1:

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -443,6 +443,35 @@ PERFORMANCE_GRAPHITE_URL: 'SetPerformanceGraphiteHostName'
...
+EDXAPP_PROFILE_IMAGE_BACKEND:
+  class: storages.backends.overwrite.OverwriteStorage
+  options:
+    location: "{{ edxapp_media_dir }}/profile-images/"
+    base_url: "{{ EDXAPP_MEDIA_URL }}/profile-images/"
+
+EDXAPP_PROFILE_IMAGE_MAX_AGE: 31536000
+
+# used to salt hashed usernames for profile image filenames
+EDXAPP_PROFILE_IMAGE_SECRET_KEY: placeholder_secret_key
```
feanil, line 469: nitpick: quote this for consistency.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -150,6 +156,12 @@ error_page {{ k }} {{ v }};
     try_files $uri @proxy_to_lms_app;
   }
 
+location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
+    root {{ edxapp_media_dir }};
+    try_files /$file =404;
+    expires {{ EDXAPP_PROFILE_IMAGE_MAX_AGE }}s;
```
feanil, line 162: I'd rather have the `s` be a part of the max age variable and make a comment above the MAX age var to point to the nginx time syntax which will be valid for the values of `EDXAPP_PROFILE_IMAGE_MAX_AGE`

http://nginx.org/en/docs/syntax.html

feanil, line 162: Talked to Max in person, this var is used in multiple places and this was done to keep that number DRY.

feanil:
:+1:

PR #1931, opened 2015-03-30
modify the flower role for multiple venvs

arbabnazar:
@e0d can you please review it, I have modify the flower role to use it for multiple venvs. Thanks

### DIFF
--- a/playbooks/roles/flower/defaults/main.yml
+++ b/playbooks/roles/flower/defaults/main.yml
@@ -9,9 +9,9 @@ FLOWER_ADDRESS: "0.0.0.0"
 FLOWER_PORT: "5555"
 
 flower_user: "flower"
-flower_app_dir: "{{ COMMON_APP_DIR }}/flower"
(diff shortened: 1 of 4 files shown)


playbooks/roles/flower/meta/main.yml
```diff
@@ -1,4 +1,4 @@
 ---
-dependencies:
+dependencies:
```
e0d: Seems like an uneeded whitespace change.

playbooks/roles/flower/defaults/main.yml
```diff
@@ -9,9 +9,9 @@ FLOWER_ADDRESS: "0.0.0.0"
 FLOWER_PORT: "5555"
 
 flower_user: "flower"
```
e0d: Let's make this var all caps, FLOWER_USER since it will be expected to override it.

e0d:
:+1:

PR #1933, opened 2015-03-30
Don't terminate volumes on instance termination for now.

feanil:
This setting gets propogated to the launch configuration and is currently
causing unexpected behaviour.


fredsmith:
:+1:

PR #1934, opened 2015-03-30
Add and rearrange ecommerce settings.

feanil:
@rlucioni FYI


rlucioni:
Looks good to me. Are database settings like `ATOMIC_REQUESTS` and `CONN_MAX_AGE` only configured in the secure repo?

feanil:
They are set in secure, but I should update the defaults as well.

rlucioni:
Nice. :+1:

maxrothman:
:+1:

PR #1935, opened 2015-03-30
Set better defaults for LMS_BASE, CMS_BASE

singingwolfboy:
One of our contributors reported an issue in devstack where the "View Unit in Studio" button on the LMS wasn't working. Turns out the problem was that the `CMS_BASE` variable was blank, so that when the URL to Studio was constructed, it was missing a crucial component. This PR sets better defaults for `CMS_BASE`, `LMS_BASE`, and `PREVIEW_LMS_BASE` so that they work better for devstack.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -23,9 +23,9 @@ EDXAPP_DJFS:
   directory_root: '{{ edxapp_data_dir }}/django-pyfs/static/django-pyfs'
   url_root : '/static/django-pyfs'
 
-EDXAPP_LMS_BASE: ""
-EDXAPP_PREVIEW_LMS_BASE: ""
-EDXAPP_CMS_BASE: ""
+EDXAPP_LMS_BASE: "localhost:8000"
+EDXAPP_PREVIEW_LMS_BASE: "preview.localhost:8000"
+EDXAPP_CMS_BASE: "localhost:8001"
 
 EDXAPP_LMS_GUNICORN_EXTRA: ""
 EDXAPP_LMS_GUNICORN_EXTRA_CONF: ""


e0d:
Have you tested in an AWS sandbox?  Generally, this looks good, but I'd want to make sure it's not breaking anything in our AWS deployments.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -23,9 +23,9 @@ EDXAPP_DJFS:
   directory_root: '{{ edxapp_data_dir }}/django-pyfs/static/django-pyfs'
   url_root : '/static/django-pyfs'
 
-EDXAPP_LMS_BASE: ""
-EDXAPP_PREVIEW_LMS_BASE: ""
-EDXAPP_CMS_BASE: ""
+EDXAPP_LMS_BASE: "localhost:8000"
```
feanil, line 26: Let's use `EDXAPP_LMS_NGINX_PORT` `EDXAPP_LMS_PREVIEW_NGINX_PORT` and `EDXAPP_CMS_NGINX_PORT` instead of hardcoding the ports in the defaults here.

singingwolfboy, line 26: But `EDXAPP_LMS_NGINX_PORT` is 18000, `EDXAPP_LMS_PREVIEW_NGINX_PORT` is 18020, and `EDXAPP_CMS_NGINX_PORT` is 18010 -- those aren't the correct values. What variable should I use for these values?

openedx-webhooks:
Thanks for the pull request, @singingwolfboy! I've created [OSPR-1114](https://openedx.atlassian.net/browse/OSPR-1114) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

nedbat:
@e0d @feanil @singingwolfboy should we continue to pursue this?

singingwolfboy:
I still think it's worth doing, but I don't know what sort of effort will need to go into verifying the problem and the fix.

playbooks/roles/edxapp/defaults/main.yml
feanil, line 26: What do you mean they aren't correct? I don't understand.

singingwolfboy, line 26: When I wrote this code, I thought that port 8000 was the correct port for LMS, and port 8001 was the correct port for Studio. I don't know if the codebase still encodes this assumption, but I believe that it used to for the "View Unit in Studio" button.

feanil, line 26: I'm confused. Why not use the pre-defined variables like EDXAPP_LMS_NGINX_PORT.

https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L212

feanil:
@singingwolfboy Ping. There are vars you can use here.

singingwolfboy:
I no longer have permissions to modify this branch, but I've created a new pull request to continue this: #2773. I'm closing this pull request.

PR #1938, opened 2015-03-31
Tighten up splunk monitor recursion for jenkins TE-807

jzoldak:
Now running at 23 %CPU and 1.0 %MEM.

@benpatterson


benpatterson:
:+1:

PR #1942, opened 2015-04-07
Don't override static files.

feanil:
We want to use the static files directories from the base config, it doesn't make sense for configuration to override them.

FYI @rlucioni


fredsmith:
:+1:

PR #1946, opened 2015-04-12
Don't run `/edx/bin/update`  with `--tags deplay` by default.

feanil:
Having `--tags deploy` be part of the ansible-playbook call
that `/edx/bin/update` makes is not intuitive.  It's broken
us a few times recently and I don't think the benefit it
provides is worth the cost and confusion.  I feel that anyone
who wants to use tags will know how to do it without using
`/edx/bin/update`.

Review: @e0d @maxrothman @fredsmith


fredsmith:
:+1:

PR #1947, opened 2015-04-13
Minos should not be on sandboxes

maxrothman:
As per [this discussion](https://groups.google.com/forum/#!topic/openedx-ops/L_0orWLIc6Y), minos should not be on sandboxes.


fredsmith:
:+1:

PR #1950, opened 2015-04-14
Update syntax to match the new format.

feanil:
The old syntax was:
- causing the module name lookup to fail probably because it was having trouble dealing with the white space.
- using an older name for a module argument


maxrothman:
:+1:

PR #1693, opened 2014-11-09
 Disable dynamic scripting when not  ELASTICSEARCH_CLUSTERED

fmyzjs:
ticket :  [OSPR-191](https://openedx.atlassian.net/browse/OSPR-191)

Disable dynamic scripting on elasticsearch as it is insecure  when not ELASTICSEARCH_CLUSTERED
@feanil has disabled this but only work when ELASTICSEARCH_CLUSTERED is True.

So  people who install with the default conf (production/devstack vagrant included )will still contain the eval code !It is important to fix , our edx server was hacked by hacker due to this bug.

See: http://bouk.co/blog/elasticsearch-rce/
CVE: CVE-2014-3120

PR #1959, opened 2015-04-22
Use templated ElasticSearch config for single-instance deployment

mtyaka:
We use a custom elasticsearch.yml configuration file when deploying in clustered mode. The same configuration file should be used in single-instance mode. The templated file tweaks some settings and disables the insecure dynamic scripting, so it makes sense to use it in both deployment scenarios.

_Background_: The main motivation behind this change is to disable dynamic scripting when deploying to a single instance (it was already disabled in cluster mode).
_Partner Information_: Harvard, for a third-party Open edX installation
_Jira ticket_: https://openedx.atlassian.net/browse/OSPR-544

### DIFF
--- a/playbooks/roles/elasticsearch/tasks/main.yml
+++ b/playbooks/roles/elasticsearch/tasks/main.yml
@@ -1,14 +1,14 @@
 ---
 
 # elasticsearch
-# 
+#
 # Dependencies:
 #
 #   * common
 #   * oraclejdk
-# 
+#
 # Example play:
-#  
+#
 # This role can be used to do a single-server or clustered
 # installation of the elasticsearch service.  When a cluster
 # is being installed, there are two important things that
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: http://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=1959

openedx-webhooks:
Thanks for the pull request, @mtyaka! I've created [OSPR-544](https://openedx.atlassian.net/browse/OSPR-544) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams. 

- supporting documentation
- edx-code email threads

carsongee:
You might want to take a look at the comments in https://github.com/edx/configuration/pull/1693 which is a very similar change.  Basically we would like this change, but I think we would prefer the approach of using an entire template instead of lineinfile.  I think this shouldn't be much more work since we already have a templated `elasticsearch.yml` config when clustering is set.  So it probably shouldn't be much more than enabling the configuration template task for non-clustered runs and adding some more logic to the existing template to handle both use cases, i.e. single host vs. multi-machine cluster.

feanil:
+1 to what @carsongee  said,  I would rather we update the templated elasticsearch config file to work for non-clustered configurations as well.

mtyaka:
@carsongee @feanil Thanks for the feedback, makes sense. I'll try to update the patch to use the templated config file instead.

mtyaka:
@feanil @carsongee I changed the patch and description to use the templated config file instead of `lineinfile`. It was easier than expected - I didn't need to change the configuration template at all, it works fine in single-instance mode, too. All I had to do was to remove the `when: ELASTICSEARCH_CLUSTERED` checks in some of the tasks.

feanil:
:+1:   I'll try it out here and then merge it.

carsongee:
Looks good to me too, but I wonder if @e0d had a reason not to do that in the first place. Does anyone happen to remember?  Is there anything in the default config we are losing with this?

feanil:
I personally thought that the template file was always being dropped and that is why we weren't making the changes to add the line to the file.

mtyaka:
@carsongee The default elasticsearch.yml config file only contains comments for each available option - but doesn't actually set anything (all examples are commented out), so we aren't losing any settings by using the template.

carsongee:
cool :+1:

PR #1962, opened 2015-04-22
change default profile image

fredsmith:


andy-armstrong:
:+1: thanks for jumping on this, @fredsmith.

feanil:
@andy-armstrong is this image a valid default for the openedx community?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -716,6 +717,7 @@ generic_env_config:  &edxapp_generic_env
   WIKI_ENABLED: true
   SYSLOG_SERVER:  "{{ EDXAPP_SYSLOG_SERVER }}"
   LOG_DIR:  "{{ COMMON_DATA_DIR }}/logs/edx"
+  PROFILE_IMAGE_DEFAULT_FILENAME: "{{ EDXAPP_PROFILE_IMAGE_DEFAULT_FILENAME }}"
```
maxrothman: Can you move this to `lms_env_config` with the rest of the profile image stuff?

maxrothman:
I don't think this change won't do anything, because this value is already set in `common.py`.

Additionally, my impression was that the default image was going to be handled via the asset pipeline. Whatever ended up in that place would be the default image. I remember a specific discussion with @jimabramson and @dan-f where we decided NOT to expose the default image location in configuration.

andy-armstrong:
@maxrothman That is odd. I remember us agreeing that we could use the IS_EDX_DOMAIN configuration setting to dynamically choose which image was chosen. We did it this way:

https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2277

Did you have a subsequent meeting with @dan-f and @jimabramson?

Anyway, it turns out that this approach doesn't work, because IS_EDX_DOMAIN is still false when the line is executed, and is only overridden to true later.

andy-armstrong:
@feanil The setting in common.py was intended to work for both Open edX and edx.org, but it ended up always using the Open edX themed image. The idea of this PR was to override the default just for edx.org. Is there a better way to manage this?

fredsmith:
@andy-armstrong we should set defaults to be open-edx branding, and then override in the environment config.

playbooks/roles/edxapp/defaults/main.yml
andy-armstrong: Thanks, @maxrothman. In my testing I had this in `lms.env.json` which I believe is the same thing you are suggesting.

andy-armstrong:
@fredsmith Sorry, that is indeed what I intended. I don't know how these files map to the json files. This setting is needed to override the image only for edx.org and other edX branded sites.

andy-armstrong:
FYI, here's the PR where I introduced the default profile image:

https://github.com/edx/edx-platform/pull/7502

maxrothman:
@andy-armstrong is there a way to have IS_EDX_DOMAIN defined there, rather than adding new config?

EDIT: or better yet, could we do the same check in a different settings file?

andy-armstrong:
@maxrothman The problem is that IS_EDX_DOMAIN is defined at the time it gets used, but it hasn't yet been overridden by lms.envs.json. So at the time the PROFILE_IMAGE_DEFAULT_FILENAME is defined, it appears that the code is _not_ running on the edX domain.

It seems to me that we should have an edx-only configuration file (edx.py, say) that only gets included if IS_EDX_DOMAIN is true. We'd have to make sure that it gets included after the lms.envs.json settings get interpreted. I imagine if we did this that there could be a number of edX specific configurations that could go into such a file.

@feanil @maxrothman @cahrens what do you guys think? Should I invest in trying to implement this idea? Is there a better way to configure this?

maxrothman:
@andy-armstrong can you point to where `IS_EDX_DOMAIN` is defined?

andy-armstrong:
@maxrothman https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L224

maxrothman:
@andy-armstrong no, where does it get set to True?

andy-armstrong:
@maxrothman @feanil Maybe I have this wrong. I see there are several conditionals in common.py which would appear to have the same problem that I'm describing.

e.g. https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2197

How can these work in concert with lms.envs.json?

maxrothman:
Why not just move those to `aws.py` after `IS_EDX_DOMAIN` is already set?

andy-armstrong:
@maxrothman I don't know how IS_EDX_DOMAIN gets set in production. On devstack, I add a line to `lms.envs.json` to enable edX branding. I thought it was a configuration setting, but I don't see it in the configuration repo.

andy-armstrong:
@maxrothman explained to me that aws.py evaluates the JSON files before the configuration settings are specified, so all I needed to do was move the logic from common.py. At his suggestion, I've also left the Open edX default in common.py.

https://github.com/edx/edx-platform/pull/7784

If folks are happy with this approach, then this configuration PR can be closed.

maxrothman:
Superseded by https://github.com/edx/edx-platform/pull/7784. I think this change is no longer necessary.

PR #1965, opened 2015-04-24
We don't need it as a requirement it is in core as of 2.7.

feanil:
See https://pypi.python.org/pypi/argparse for more details.


maxrothman:
:+1:

PR #1969, opened 2015-04-28
'vars' var doesn't work anymore.

feanil:


e0d:
:+1: but more shell script.

PR #1972, opened 2015-04-29
Use ansible instead of scp.

feanil:
Ansible will use the correct ssh params so that we don't have to worry about host key checking
which will become a problem as we build sandboxes at the same domain more than once.

@maxrothman @e0d @fredsmith


e0d:
:+1:

PR #1975, opened 2015-04-29
Have CMS come up on the same port as the LMS.

feanil:


feanil:
I'm not sure how this worked without this line.  @e0d

e0d:
:+1:

PR #1989, opened 2015-05-07
Removed post.txt from the list of files that will have its github urls replaced with git mirror urls.

feanil:


singingwolfboy:
Seems reasonable to me. :+1:

fredsmith:
:+1:

PR #1996, opened 2015-05-11
Don't need pycurl for apt_repository anymore.

feanil:
We used to need to install python-pycurl so that the
apt_repository module would work correctly but the code
of that module has changed. It uses the apt python api
instead and installs it automatically if it is not available.

https://github.com/ansible/ansible-modules-core/blob/devel/packaging/os/apt_repository.py#L104-L118


e0d:
:+1:

fredsmith:
:+1:

PR #1998, opened 2015-05-11
Add an option to override max tasks pre child for edxapp workers.

feanil:
Update default tasks per child for high mem queue.

@e0d @fredsmith


jibsheet:
:+1:

PR #2003, opened 2015-05-13
Pass an option group for the new RDS

feanil:
This is needed now that the source RDS is also in the VPC and its option group is associated with the source VPC.


jibsheet:
:+1:

PR #2005, opened 2015-05-14
Add cassandra role

maxrothman:
Submitting for initial review

Uses JBOD, rather than RAID0 or lvm. It should be pretty simple to integrate either of those once the config code for them is written.

Default vars support single-instance deployment, but comments suggest alternatives for multi-node deployments.

Currently does not support authentication, because doing so is a little more complicated (involves creating a keyspace and changing its replication factor).


maxrothman:
Ready for review @e0d @jibsheet @feanil @fredsmith

playbooks/roles/cassandra/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+# cassandra.yaml basic configuration parameters
+cassandra_cluster_name: "Test Cluster"
+
+#Set this!
+cassandra_seeds: ["127.0.0.1"]
+
+#should eventually use EC2Snitch
+cassandra_snitch: "SimpleSnitch"
+
+#set this ONLY when initializing a new cluster with NO DATA
```
e0d, line 28: Given the warning, should this default to false?  I'd rather have new clusters be the exception than than to risk data loss.

playbooks/roles/cassandra/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
+
+- name: Make sure the cassandra group is present
+  group:
+    name: "{{cassandra_group}}"
+    state: present
+
+- name: Make sure the cassandra user is present
```
e0d: Isn't the package creating users?

playbooks/roles/cassandra/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
+
+- name: Make sure the cassandra group is present
+  group:
+    name: "{{cassandra_group}}"
```
e0d: white space, {{ cassandra_group }}

e0d:
This role should contain tasks for creating users in Cassandra or we should have a separate play like we do for MySQL, either way we need to ansible to do it and the dictionary that specifies users.

playbooks/roles/cassandra/defaults/main.yml
maxrothman, line 28: Good point.

maxrothman:
@e0d turns out the package does create the users. I've removed all of that logic and added a note to the variables warning people not to change them.

playbooks/roles/cassandra/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
+
+cassandra_version: "2.0.14"
+cassandra_source_url: "http://archive.apache.org/dist/cassandra/{{cassandra_version}}/apache-cassandra-{{cassandra_version}}-bin.tar.gz"
```
feanil: Looks like we don't use this source?

playbooks/roles/cassandra/templates/cassandra-env.sh.j2
```diff
@@ -0,0 +1,281 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
feanil: Looks like there are no changes in this file, do we need to override it?

feanil:
:+1:  once the above two comments are resolved.

playbooks/roles/cassandra/defaults/main.yml
e0d: If we need it fix the whitespace.

fredsmith:
LGTM :+1:

fredsmith:
:shipit:

feanil:
:shipit:   :anchor:  away!

PR #2018, opened 2015-05-19
Updates to db creation playbooks.

feanil:
- Delete the create_dbs.yml playbook since we don't use it.
- Update the create_db_and_users.yml playbook to use lists instead of dictionaries.

@e0d @jibsheet


jibsheet:
:+1:

PR #2021, opened 2015-05-19
E0d/service dbs

e0d:
Work in progress

### DIFF
--- a/playbooks/roles/edx_service_rds/defaults/main.yml
+++ b/playbooks/roles/edx_service_rds/defaults/main.yml
@@ -1,0 +1,125 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role edx_service_rds
+# 
+
+#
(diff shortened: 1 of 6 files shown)


playbooks/roles/edx_service_rds/meta/main.yml
```diff
@@ -0,0 +1,20 @@
+---
```
feanil: File Not Needed.

playbooks/roles/edx_service_rds/tasks/main.yml
```diff
@@ -0,0 +1,99 @@
...
+  rds:
+    command: "{{ db_module_command }}"
+    region: "{{ aws_region }}"
+    instance_name: "{{ e_d_c }}-db"
+    db_engine: "{{ db_engine }}"
+    engine_version: "{{ db_engine_version }}"
+    size: "{{ db_size }}"
+    instance_type: "{{ db_instance_type }}"
+    username: "{{ db_root_user }}"
+    password: "{{ db_root_password }}"
```
feanil: Would it be useful to put all the DB params into a dict?  It would make it easier to see what items are related.

playbooks/roles/edx_service_rds/defaults/main.yml
```diff
@@ -0,0 +1,125 @@
+---
```
feanil, line 1: The vars here should be updated to prefix with the role name as per convention.

feanil:
A few nits, but the logic makes sense.

PR #2024, opened 2015-05-20
Install phantomjs in the browser role TE-884

jzoldak:
@benpatterson

### DIFF
--- a/playbooks/roles/browsers/tasks/main.yml
+++ b/playbooks/roles/browsers/tasks/main.yml
@@ -15,20 +15,54 @@
   when: download_deb.changed
   with_items: browser_s3_deb_pkgs
 
-- name: Install ChromeDriver
+- name: download ChromeDriver
   get_url:
     url={{ chromedriver_url }}
     dest=/var/tmp/chromedriver_{{ chromedriver_version }}.zip
+  register: download_chromedriver
 
-- name: Install ChromeDriver 2
+- name: unzip ChromeDriver tarfile
   shell: unzip /var/tmp/chromedriver_{{ chromedriver_version }}.zip
          chdir=/var/tmp
+  when: download_chromedriver.changed
 
-- name: Install ChromeDriver 3
+- name: move ChromeDriver binary to /usr/local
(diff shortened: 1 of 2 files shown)


benpatterson:
LGTM.

playbooks/roles/browsers/tasks/main.yml
```diff
@@ -15,20 +15,54 @@
...
+
+- name: move PhantomJS binary to /usr/local
+  shell: mv /var/tmp/{{ phantomjs_tarfile }}/bin/phantomjs /usr/local/bin/phantomjs
+  when: download_phantom_js.changed
+
+- name: verify phantomjs location
+  stat: path=/usr/local/bin/phantomjs
+  register: phantomjs
+- assert:
+    that: "phantomjs.stat.exists"
```
benpatterson, line 65: Needs to be in quotes for ansible, eh? That's kinda weird.

benpatterson:
:+1:

playbooks/roles/browsers/tasks/main.yml
jzoldak, line 65: Actually looking at the examples again, it seems the quotes are optional if you don't have spaces or single quotes in the statement. Reference: [Testing Strategies in the Ansible docs](http://docs.ansible.com/test_strategies.html).

If we're establishing a style preference, what's your opinion - should I remove them? I don't have a strong opinion either way.

benpatterson, line 65: I don't have a strong preference, either. Sounds like "sometimes you need quotes". If that's the case, and we are landing on a style, might as well always use quotes.

jzoldak, line 65: "lolz." Alright I'll leave them. :)

jzoldak:
@benpatterson Jenkins worker ami build job [passed successfully](https://build.testeng.edx.org/job/build-jenkins-worker-ami/30/). Merging now.

benpatterson:
w00t

PR #2025, opened 2015-05-20
Fix the packer script that provisions jenkins workers

jzoldak:
- Selectively copy playbook dirs to avoid broken links
- Install git in packer script so that the edx ansible fork can be installed via requirements.txt

@benpatterson


benpatterson:
:+1:

PR #2041, opened 2015-05-26
Run an update before security installs.

feanil:
Need to do this so installs for security fixes don't fail when building from scratch.


e0d:
:+1:

PR #2042, opened 2015-05-26
Footer configuration changes

wedaly:
- Expose mobile store URLs setting to edxapp role.
- Remove deprecated social media URLs.

The related PR is: https://github.com/edx/edx-platform/pull/8175.

@feanil please review.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -183,7 +183,7 @@ EDXAPP_FEATURES:
   ENABLE_VIDEO_BEACON: false
   ENABLE_ONLOAD_BEACON: false
   ENABLE_EDXNOTES: "{{ EDXAPP_ENABLE_EDXNOTES }}"
-  
+
 EDXAPP_BOOK_URL: ""
 # This needs to be set to localhost
 # if xqueue is run on the same server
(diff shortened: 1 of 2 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -231,13 +231,8 @@ EDXAPP_PRESS_EMAIL: 'press@example.com'
 EDXAPP_PLATFORM_TWITTER_ACCOUNT: '@YourPlatformTwitterAccount'
 EDXAPP_PLATFORM_FACEBOOK_ACCOUNT: 'http://www.facebook.com/YourPlatformFacebookAccount'
 
-# Deprecated by EDXAPP_SOCIAL_MEDIA_FOOTER_URLS
-EDXAPP_PLATFORM_TWITTER_URL: "https://twitter.com/YourPlatformTwitterAccount"
-EDXAPP_PLATFORM_MEETUP_URL: "http://www.meetup.com/YourMeetup"
-EDXAPP_PLATFORM_LINKEDIN_URL: "http://www.linkedin.com/company/YourPlatform"
-EDXAPP_PLATFORM_GOOGLE_PLUS_URL: "https://plus.google.com/YourGooglePlusAccount/"
```
wedaly, line 238: These haven't been used in production for months now, so I'm removing them.  These aren't used by open source installations (unless they've explicitly enabled the edx.org footer), so this should be low-risk to remove.

feanil:
@wedaly please update the changelog to indicate that this change is backwards incompatible since it will break people who might be running older versions of edx platform.  Please include the link to the edx-platform PR in the changelog as well.

There will also need to be an update to our config-secure for this change, please link to the ticket for that work here when you've created it.

playbooks/roles/edxapp/defaults/main.yml
feanil, line 238: I would still add a changelog entry since they are ALL_CAPS vars which means they are part of the list of items we expect people to be able to configure.

wedaly:
@feanil Updated the changelog and created https://openedx.atlassian.net/browse/DEVOPS-1912 with the updates to config-secure.

feanil:
:+1:

PR #2044, opened 2015-05-26
sitespeed.io role

fredsmith:


fredsmith:
@cgoldberg FYI

playbooks/roles/sitespeedio/defaults/main.yml
```diff
@@ -0,0 +1,30 @@
...
+#
+sitespeedio_role_name: sitespeedio
+
+#
+# OS packages
+#
+sitespeedio_chrislea_ppa: "ppa:chris-lea/node.js"
+
+sitespeedio_debian_pkgs: 
+  - default-jre
```
feanil: Is there a reason to use this instead of the jre role which uses the same jre version as the rest of our applications?

feanil:
One small question otherwise seems pretty straightforward.

fredsmith:
@feanil updated to use oraclejdk role.   Tests out OK.

cgoldberg:
@fredsmith which version of Java will that install?  Sitespeed requires 1.7

fredsmith:
@cgoldberg Oracle 1.7u51

PR #2049, opened 2015-05-27
Update check-migrations.sh

feanil:
Add the EDXAPP_CFG_DIR so that yaml templates get created in the correct location.


fredsmith:
LGTM :sheep: it

PR #2055, opened 2015-05-28
Pull out docstore config and move modulestore config.

feanil:
We need to have different read preferences in the CMS and LMS for modulestore configs so that the lms can read from the nearest node.


jibsheet:
:+1:

maxrothman:
:+1:

PR #2057, opened 2015-05-28
Separate the read preference per docstore.

feanil:
This way you can change read preference for one of the module stores
without effecting the other.

Also fix whitespace


feanil:
@maxrothman @jibsheet @fredsmith

jibsheet:
:+1:

PR #2062, opened 2015-05-29
App monitoring is triggered on the _APP variable

jibsheet:
COMMON_ENABLE_NEWRELIC turns on server monitoring and is on for all
sandboxes.  App monitoring should not be enabled for all sandboxes and
since edxapp workers is on all sandboxes, surprise newrelic!


feanil:
:+1:

PR #2063, opened 2015-05-29
Tag the paver static assets task

jibsheet:
This lets you run ansible-playbook with --skip-tags gather_static_assets
so you can redeploy edx-platform on your sandbox to test configuration
changes without waiting for paver.

Likely wants a change to /edx/bin/update but I just ran things manually to test.


feanil:
:+1:

PR #2068, opened 2015-05-29
Update the source list after adding git ppa TE-907

jzoldak:
@benpatterson


benpatterson:
LGTM :+1:

jzoldak:
@benpatterson verified by spinning up a new jenkins test-worker.
git version 2.4.2

Did we want to pin it to some specific version?

benpatterson:
That's a tricky one to answer. The PPAs end up getting deleted when they are updated. At least, that's the behavior from December when it went from 2.1 to 2.2. What might be better is storing the ppa in a known place? I'd seek advice from @feanil and others.

Issue #2036, opened 2015-05-22
Named release install not working on Ubuntu 12.04 scratch install. 


taz-gnu:
The built script (configuration/master/util/install/sandbox.sh) runs a git clone -b (git clone -b $CONFIG_VER https://github.com/edx/configuration) to pull the named tag version of config. 

Unfortunately the version of git in 12.04 is 1.7.9.5, and checking out a tag with git clone is not supported before git v 1.7.10. see http://stackoverflow.com/questions/17216067/git-clone-b-tag-results-in-warning-remote-branch-not-found-in-upstream-origi

The clone completes with the current head. With $CONFIG set to Birch, this results in a Birch install with a current master config. If I remember right I got errors with the xqueue_watcher service.. 

You may want to try a  'git clone'  'git checkout'. I did a git upgrade and a 'git clone -b $CONFIG_VER --single-branch https://github.com/edx/configuration' with success. 

Also note (second issue); 
The 'manual installation' section of the instructions at https://github.com/edx/configuration/wiki/edX-Ubuntu-12.04-64-bit-Installation does not have the $EXTRA_VARS properly set. With out this, the playbooks run without a named release. 

This, with the above error also leaving the config at master, results in a successful master install instead of an intended named release installation.

feanil:
@taz-gnu I made an update to the script per your suggestion.  If you have any more suggestions, I'd be happy to review any according PRs.

taz-gnu:
thanks feanil. 

note the instruction page at https://github.com/edx/configuration/wiki/edX-Ubuntu-12.04-64-bit-Installation still has 'git clone -b' and '$EXTRA_VARS not set correctly' errors.

raelliss:
Hi All

Do you have the correct setting. Thanking you

taz-gnu:
The manual install step should mirror the steps in the script 

https://github.com/edx/configuration/blob/master/util/install/sandbox.sh

Its simply getting out of date.

PR #2071, opened 2015-05-30
Ubuntu 12.04 doesn't support 'git clone -b'

feanil:
This is to deal with the issue pointed out in #2036

@jibsheet @fredsmith @maxrothman


fredsmith:
:+1:

PR #2075, opened 2015-06-02
Add a variable for the new THEME_DIR

nedbat:
This is to support the sass-theming branch in edx-platform.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -510,6 +510,9 @@ EDXAPP_XBLOCK_SETTINGS: {}
 # the value is the 32-character key.
 EDXAPP_CREDIT_PROVIDER_SECRET_KEYS: {}
 
+# Comprehensive Theming
+EDXAPP_COMP_THEME_DIR: ""
+
 #-------- Everything below this line is internal to the role ------------
 
 #Use YAML references (& and *) and hash merge <<: to factor out shared settings
@@ -803,6 +806,7 @@ generic_env_config:  &edxapp_generic_env
   REGISTRATION_EXTRA_FIELDS: "{{ EDXAPP_REGISTRATION_EXTRA_FIELDS }}"
   XBLOCK_SETTINGS: "{{ EDXAPP_XBLOCK_SETTINGS }}"
   EDXMKTG_USER_INFO_COOKIE_NAME: "{{ EDXAPP_EDXMKTG_USER_INFO_COOKIE_NAME }}"
+  COMP_THEME_DIR: "{{ EDXAPP_COMP_THEME_DIR }}"
 
 lms_auth_config:
   <<: *edxapp_generic_auth
@@ -928,7 +932,7 @@ worker_core_mult:
   lms: 4
   cms: 2
 
-# Theming
+# Stanford-style Theming
 # Turn theming on and off with edxapp_use_custom_theme
 # Set theme name with edxapp_theme_name
 # Stanford, for example, uses edxapp_theme_name: 'stanford'


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -513,7 +513,7 @@ edxapp_media_dir: "{{ edxapp_data_dir }}/media"
 edxapp_course_static_dir: "{{ edxapp_data_dir }}/course_static"
 edxapp_course_data_dir: "{{ edxapp_data_dir }}/data"
 edxapp_upload_dir: "{{ edxapp_data_dir }}/uploads"
-edxapp_theme_dir: "{{ edxapp_data_dir }}/themes"
+edxapp_custom_theme_dir: "{{ edxapp_data_dir }}/themes"
```
nedbat: Hmm, this might be a good reason to have a name, so we can leave this as "edxapp_theme_dir".

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -743,6 +743,7 @@ generic_env_config:  &edxapp_generic_env
   COMMENTS_SERVICE_KEY: "{{ EDXAPP_COMMENTS_SERVICE_KEY }}"
   SEGMENT_IO_LMS: "{{ EDXAPP_SEGMENT_IO_LMS }}"
   SEGMENT_IO: "{{ EDXAPP_SEGMENT_IO }}"
+  THEME_DIR: "{{ edxapp_theme_dir }}"
```
fredsmith: I think this should be "{{ edxapp_custom_theme_dir }}" - or you should change the previous change on line 516.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -901,6 +902,9 @@ edxapp_theme_name: ""
 edxapp_theme_source_repo: 'https://{{ COMMON_GIT_MIRROR }}/Stanford-Online/edx-theme.git'
 edxapp_theme_version: 'master'
 
+# New theming.
+edxapp_theme_dir: ""
```
fredsmith: this seems to be wiping out the default on line 516.  - I think it should be removed, unless we really want the default to be blank.

nedbat: I changed the name of the variable on line 516.  A question I have is: is it OK to change that name there, or is it connected somehow to things outside these playbooks?

feanil:
@nedbat can you give us more info on what the new theme sass change actually is? And how does it relate to microsites?

nedbat:
@feanil @fredsmith @singingwolfboy the theming that db is leading uses a new setting, THEME_DIR to control where the theme is stored.   The plan for the first MVP is that microsites continue to work as they currently do.  So the thing we need here is a way to set THEME_DIR that will affect both the lms.env.json and the settings.py (which can simply read from the .json)

This new theming is also (for now) independent of the current "custom theming" (aka stanford theming), so the existing settings should all continue to work.

feanil:
Ok, I think this is going to get confusing very quickly,  I think that we should change the existing theme vars in all their uses to `edxapp_microsites_...`  and then create a new set of vars for the theme dir where the var names don't overlap with the old theme vars.

nedbat:
I've updated with a new commit. It leaves all the old names as-is, and uses "COMP_THEME" for the slug for this new comprehensive theming system.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -488,6 +488,11 @@ EDXAPP_PARSE_KEYS: {}
 EDXAPP_EDXNOTES_INTERFACE:
   url: http://localhost:18120/api/v1
 
+# Comprehensive Theming
+EDXAPP_USE_COMP_THEME: false
```
feanil: What is the goal of this flag?  Is it meant to go into the features dictionary?

singingwolfboy: @nedbat, I don't think we need this flag.

nedbat:
Ping.  Any concerns remaining about this PR?

feanil:
:+1:

PR #2077, opened 2015-06-03
update github_post_status.py for jenkins_worker role

clytwynec:
@benpatterson  I'm updating this script to send context when posting the github status. It looks like this script is still up to date otherwise and we can use it from a post build script for our master builds.

### DIFF
--- a/playbooks/roles/jenkins_worker/templates/github_post_status.py.j2
+++ b/playbooks/roles/jenkins_worker/templates/github_post_status.py.j2
@@ -25,7 +25,7 @@ def parse_args(arg_list):
     Parse the list of arguments, returning a dict.
     Prints an error message and exits if the arguments are invalid.
     """
-    if len(arg_list) != 7:
+    if len(arg_list) != 8:
         print USAGE.format(arg_list[0])
         exit(1)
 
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_worker/templates/github_post_status.py.j2
```diff
@@ -41,11 +41,12 @@ def parse_args(arg_list):
         'sha': arg_list[3],
         'status': arg_list[4],
         'target_url': arg_list[5],
-        'description': arg_list[6]
+        'description': arg_list[6],
+        'context': arg_list[7],
```
benpatterson, line 45: Could you give a description of what context means, either as a hashed comment on the right of this line, or in a docstring?

playbooks/roles/jenkins_worker/templates/github_post_status.py.j2
```diff
@@ -59,7 +60,8 @@ def post_status(org, repo, sha, status, target_url, description):
     params = {
         'state': status,
         'target_url': target_url,
-        'description': description
+        'description': description,
+        'context': context,
```
benpatterson, line 64: Perhaps the better place is here.... but you see where I'm going. Just some extra on the documentation.

playbooks/roles/jenkins_worker/templates/github_post_status.py.j2
clytwynec, line 45: The docstring has a link to the official github documentation that covers all of these (description, target_url, status, and context).   Should there be more than that?

benpatterson, line 45: Yeah I noticed that after my comment. Mmmm nah I think it's fine as-is. Doc is more straightforward to maintain this way as well. :)

benpatterson:
We've addressed the doc stuff. LGTM :+1:

PR #2078, opened 2015-06-04
Provide separate internal/public API endpoints.

jibsheet:
This allows us to route LMS requests over an internal ELB while public
JS direct requests to the API go over a separate ELB.

Implements changes from edx/edx-platform#8336


fredsmith:
:+1:  LGTM

PR #2080, opened 2015-06-04
fix broken test

clytwynec:


clytwynec:
@benpatterson  (FYI I've created a new AMI from this commit already too)

benpatterson:
:+1: solid

PR #2081, opened 2015-06-05
Switch to using the vpc_class_b for these subnets

jibsheet:
I suspect the third change is wrong - as is, it creates a route table,
but I cannot connect to the RDS.  However, there is no local gateway in
10.253.0.0 in loadtest, so the default isn't quite right either.

We should possible make SUBNET_[12]_CIDR easier to override, since I
ended up changing them anyway to use something other than .50.

@feanil - this is what I ran in the loadtest environment, along with the VPC file so
 ansible-playbook  -i localhost, edx_service_rds.yml -vv -e@/Users/kevinfalcone/git/ops/edx-secure/cloud_migrations/loadtest-edx.yml  -e@/Users/kevinfalcone/git/ops/edx-secure/ansible/vars/db/loadtest-edx-notes-rds.yml


feanil:
:+1:

PR #2085, opened 2015-06-10
Remove VMware Fusion references from Vagrantfile

singingwolfboy:
They're on kifli, rather than lavash. They don't support named releases. Realistically, anyone using VMware Fusion with this Vagrantfile is going to run into a lot of problems.

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -134,15 +134,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
   end
 
-  ["vmware_fusion", "vmware_workstation"].each do |vmware_provider|
-    config.vm.provider vmware_provider do |v, override|
-      override.vm.box     = "kifli-devstack-vmware"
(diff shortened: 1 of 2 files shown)


carsongee:
I can generate new boxes, I just hadn't been asked and we hadn't done an official lavash release.  They still inherit all the other settings, so they should work with named releases and shouldn't cause problems that I'm aware of.  That said I've been running on the virtualbox one for the last 6 months or so, so I haven't I've verified that.

singingwolfboy:
@carsongee In my opinion, if we aren't going to regularly generate boxes for VMware, then we shouldn't include it in the Vagrantfile at all. Maybe we can include a comment in there saying something along the lines if "if you want to run this on VMware, that's totally possible, but you're on your own". But it's worse to include old, outdated, possibly-broken Vagrant boxes, than to include no Vagrant boxes at all.

singingwolfboy:
@carsongee and I talked about this further: looks like he was building the VMware images previously, and wasn't notified that new images needed to be built. We dropped the ball on this one, sorry. Carson is going to build new images, send them to edX, and we're going to host them on S3 and alter the Vagrantfiles to reference these new images. Since it looks like we have people still interested in working with VMware, I'm closing this pull request.

PR #2088, opened 2015-06-11
Don't disable token checking on notes

jibsheet:
The default is to enforce token checking (in edx-notes-api's common.py)
and we should be doing the same thing in deployed configuration.

Make this into a variable in case other systems (sandboxes?) want to
disable it in the future.


feanil:
:+1:

PR #2091, opened 2015-06-12
Change replica host var names to namespace them with the roles that use them

maxrothman:
These vars are only used in the ad_hoc_reporting role, but their prefix indicates that they're part of their services' roles. This corrects the prefix.

Relates to https://github.com/edx-ops/edx-secure/pull/492


feanil:
:+1:

PR #2098, opened 2015-06-16
By deafult, nginx looks for ELB headers to forward

jibsheet:
In the case of sandboxes, nginx listens on 443/80 and needs to pass
along the scheme of the request so that django knows if the client is
connecting over HTTP or HTTPS
DEVOPS-2074


maxrothman:
Ticket # should be in the commit msg

feanil:
That name is misleading but you know this.  :+1:

jibsheet:
@maxrothman ticket # is in the commit message

maxrothman:
Ah, I think we're trying to do msgs like "DEVOPS-####: this is some stuff"

maxrothman:
Otherwise, :+1:

jibsheet:
Putting DEVOPS-####: in the subject of the git commit eats a lot of valuable space, but since I needed to fix a typo in the message, I moved it.

I have generally included it in the body of the commit message in recent work.

We should discuss this and write a policy, but not in this PR.

PR #2099, opened 2015-06-18
Stop using the RABBITMQ_CLUSTERED var and tooling.

feanil:
This method relied on the `hostvars` variable provided by ansible to always
be all the nodes in the rabbitmq cluster.  The hostvars var broke as a result
of the upgrade to ansible 1.9 but also even if it did work, this method
required that you always run rabbit plays on the whole cluster.  This is not
reasonable as it makes it very hard to heal a single node that may have fallen
over.

From here on, the RABBITMQ_CLUSTERED_HOSTS var should be used instead to
specify the rabbit hosts that are part of a cluster.

@edx/devops

### DIFF
--- a/playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
+++ b/playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
@@ -2,7 +2,7 @@
 
 [{rabbit, [
   {log_levels, [{connection, info}]},
-{# If RABBITMQ_CLUSTERED_HOSTS is set, use that instead assuming an aws stack.
+{# 
    Note: That these names should include the node name prefix. eg. 'rabbit@hostname'
 #}
   {cluster_nodes, {['{{ RABBITMQ_CLUSTERED_HOSTS|join("\',\'") }}'], disc}}
--- a/playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
+++ b/playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
@@ -2,19 +2,8 @@
 
 [{rabbit, [
   {log_levels, [{connection, info}]},
-{% if RABBITMQ_CLUSTERED -%}
-  {%- set hosts= [] -%}
-
-  {%- for host in hostvars.keys() -%}
-    {% do hosts.append("rabbit@ip-" + host.replace('.','-')) %}
-  {%- endfor %}
-
-  {cluster_nodes, {['{{ hosts|join("\',\'") }}'], disc}}
-
-{%- else -%}
-{# If rabbitmq_clustered_hosts is set, use that instead assuming an aws stack.
+{# If RABBITMQ_CLUSTERED_HOSTS is set, use that instead assuming an aws stack.
    Note: That these names should include the node name prefix. eg. 'rabbit@hostname'
 #}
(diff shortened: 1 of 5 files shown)


playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
```diff
@@ -2,19 +2,8 @@
...
-
-  {%- for host in hostvars.keys() -%}
-    {% do hosts.append("rabbit@ip-" + host.replace('.','-')) %}
-  {%- endfor %}
-
-  {cluster_nodes, {['{{ hosts|join("\',\'") }}'], disc}}
-
-{%- else -%}
-{# If rabbitmq_clustered_hosts is set, use that instead assuming an aws stack.
+{# If RABBITMQ_CLUSTERED_HOSTS is set, use that instead assuming an aws stack.
```
maxrothman: Please update this comment to not call out AWS explicitly.

maxrothman:
:+1:  after comment addressed

carsongee:
We are also doing this in MariaDB and ElasticSearch.  Are we going to change those soon as well?

feanil:
Good catch @carsongee yea, this PR should probably change all of them, one thing @e0d suggested is to replace the cluster option with an EC2 option that looks up the members of the cluster via tags and registers a variable.  I'll try to incorporate that and make the relevant changes to the other roles that use this logic.

carsongee:
I'd rather not marry these to EC2 since we have the vagrant cluster that is really helpful for testing (and we also run all of these clusters in OpenStack).  Using ansible groups would be fine though, but I remember that not being possible, and is part of the reason we went this way in the first place.

carsongee:
that said, that was with 1.5, and also my memory isn't super clear on that, so who knows.  It is probably worth a check

feanil:
Infact the issue is that in 1.5 groupvars worked like we would want but in
1.9 they no longer do making it harder for use that to automatically define
the cluster.  Another option might be to specify the group name that the
instance is part of as a part of the vars you pass in.  eg. `-e
cluster_group=tag_Name_stage-edx-commoncluster` in our case and then it
would be able to look up the hosts in that group.  The way we had it before
it would rely on the list of hosts in the actively running ansible run and
that would often break things.

carsongee:
If you can get it to use an inventory group, that would be great, but a static list is nice too when you want to be explicit.  Also, you'll need to change the vagrant-cluster play to switch to the new variable name:
https://github.com/edx/configuration/blob/1ae3121fb0f2450a9263e57f334961803affcbb7/playbooks/vagrant-cluster.yml#L31-L34

smaeda-ks:
In our case,
we deploy some servers have rabbitmq function only.
They also need clustering option.
so we use inventory group like this.

``` jinja
[{rabbit, [
{% if RABBITMQ_CLUSTERED -%}
  {%- set hosts= [] -%}

  {%- for host in groups['rabbitmq'] -%}
    {% do hosts.append("rabbit@" + hostvars[host]['ansible_hostname']) %}
  {%- endfor %}

  {cluster_nodes, {['{{ hosts|join("\',\'") }}'], disc}}

]}].
```

 Just for information.

feanil:
@edx/devops please have one more look, I'd like to merge this. I've tested that it works in one of our staging environments.

e0d:
I'm happy with this, though we should send out an email to the list describing the upgrade path.

:+1:

feanil:
e-mail sent.

PR #2102, opened 2015-06-18
Added Node for E-Commerce Asset Compilation

clintonb:
XCOM-367

@rlucioni @feanil @jimabramson @Nickersoft

### DIFF
--- a/playbooks/roles/ecommerce/defaults/main.yml
+++ b/playbooks/roles/ecommerce/defaults/main.yml
@@ -148,6 +148,11 @@ ecommerce_user: "{{ ecommerce_service_name }}"
 ecommerce_home: "{{ COMMON_APP_DIR }}/{{ ecommerce_service_name }}"
 ecommerce_code_dir: "{{ ecommerce_home }}/{{ ecommerce_service_name }}"
 
+ecommerce_nodeenv_dir: "{{ ecommerce_home }}/nodeenvs/{{ ecommerce_service_name }}"
+ecommerce_nodeenv_bin: "{{ ecommerce_nodeenv_dir }}/bin"
+ecommerce_node_modules_dir: "{{ ecommerce_code_dir }}/node_modules"
+ecommerce_node_bin: "{{ ecommerce_node_modules_dir }}/.bin"
+
 ecommerce_gunicorn_host: "127.0.0.1"
 ecommerce_gunicorn_port: "8130"
 ecommerce_gunicorn_timeout: "300"
--- a/playbooks/roles/ecommerce/meta/main.yml
+++ b/playbooks/roles/ecommerce/meta/main.yml
@@ -21,3 +21,4 @@ dependencies:
       debian: "{{ ecommerce_debian_pkgs }}"
       redhat: "{{ ecommerce_redhat_pkgs }}"
   - supervisor
+  - oraclejdk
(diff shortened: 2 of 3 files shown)


playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -171,5 +176,9 @@ ecommerce_debian_pkgs:
   - libjpeg-dev
   - libssl-dev
   - libffi-dev
+  - 'build-essential'
+  - gettext
+  - openjdk-7-jdk
```
jimabramson: whats this for?

clintonb: If/when we decide to use [Closure Compiler](https://developers.google.com/closure/compiler/), I'd like to avoid the need to once again edit configuration.

feanil: We actually have a separate play to install java(https://github.com/edx/configuration/tree/master/playbooks/roles/oraclejdk).  Also, I would rather not install things we don't need.  When we actually start using it, we can add it.

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -45,6 +45,22 @@
   with_items: ecommerce_dev_requirements
   when: devstack
 
+- name: create nodeenv
+  shell: >
+    creates={{ ecommerce_nodeenv_dir }}
+    {{ ecommerce_home }}/venvs/{{ ecommerce_service_name }}/bin/nodeenv {{ ecommerce_nodeenv_dir }}
```
feanil, line 51: Is node env part of the ecom requirements now?

clintonb, line 51: Yes.

playbooks/edx-east/ecommerce.yml
```diff
@@ -11,6 +11,7 @@
       nginx_sites:
         - ecommerce
     - aws
+    - oraclejdk
```
clintonb: @jimabramson @feanil Continuing our previous conversation: Closure Compiler is the bee's knees. "If/when we enable it" will probably be today as part of XCOM-367.

jimabramson: I have no objections if this is imminently to become a dependency, and we're using the right Java.

feanil: Yea, lets use the oracle jdk role instead of installing the version from the PPA. The version we are using is the role is newer and pinned.

Example of usage: https://github.com/edx/configuration/blob/096196a21441a8ba5342dd63588df6e8e3708e0a/playbooks/roles/cassandra/meta/main.yml

playbooks/roles/ecommerce/meta/main.yml
```diff
@@ -21,3 +21,4 @@ dependencies:
       debian: "{{ ecommerce_debian_pkgs }}"
       redhat: "{{ ecommerce_redhat_pkgs }}"
   - supervisor
+  - oraclejdk
```
clintonb, line 24: @feanil Correct?

feanil, line 24: yep, looks good.

feanil:
:+1:

jimabramson:
:+1:

clintonb:
Thanks guys. Will merge after a successful AMI build.

PR #2103, opened 2015-06-18
OPS-522 - abbey notifier callbacks

fredsmith:
https://openedx.atlassian.net/browse/OPS-522 MVP.  
- Does not provide an interface for notifications to happen for jobs that aren't started by alton.
- Does not clean up redis.


feanil:
:+1:

PR #2106, opened 2015-06-19
OPS-761: Rework db-clone.py

maxrothman:


util/vpc-tools/db-clone.py
```diff
@@ -1,181 +1,250 @@
...
+import boto, boto.route53, boto.rds2
+import pymysql, yaml
+from argparse import ArgumentParser, ArgumentDefaultsHelpFormatter
+from datetime import datetime
+import os, time, json, subprocess
 
-description = """
+#Config vars
+CREATE_DBS_AND_USERS = "../../playbooks/edx-east/create_db_and_users.yml"
+VENV_ACTIVATE = "../../venv/bin/activate"
```
feanil: These vars should be overridable from the commandline.

util/vpc-tools/db-clone.py
```diff
@@ -1,181 +1,250 @@
...
-
-def wait_on_db_status(db_name, region='us-east-1', wait_on='available', aws_id=None, aws_secret=None):
-    rds = boto.rds2.connect_to_region(region)
+def parse_args():
+    parser = ArgumentParser(description=description, formatter_class=ArgumentDefaultsHelpFormatter)
+
+    #Required args
+    parser.add_argument('-s', '--source-db', dest='source_db', help="Identifier of RDS instance to clone (requried)", required=True)
+    
+    parser.add_argument('-n', '--subnet-group', dest='subnet', required=True,
```
feanil, line 19: nit: I'd rather call it the `--destination-subnet-group`.

util/vpc-tools/db-clone.py
```diff
@@ -1,181 +1,250 @@
...
+                        'Equivalent to "--skip-tags dbs"'))
+
+    parser.add_argument('--skip-users', action='store_true', dest='ansible_skip_users',
+                        help=("don't run the part of the create_dbs_and_users play that creates users. "
+                        'Equivalent to "--skip-tags users"'))
+    
+
+    args = parser.parse_args()
+    if not args.target_db:
+        args.target_db = 'from-{}-{:%Y-%m-%d-%H-%M-%S}'.format(args.source_db, datetime.now())
```
feanil, line 80: RDS doesn't allow the length of the db_name to be longer than 63 characters, will this fail gracefully if the name is too long?  Should we have validation here?

util/vpc-tools/db-clone.py
```diff
@@ -1,181 +1,250 @@
...
+def run_sql(host, user, password, db, script):
+    '''Run a sql script file against a DB
+    Args:
+        host: hostname of an MySQL instance
+        user: user to run the script as
+        password: password to use
+        db: database to use
+        script: name of file containing SQL script to run
+    '''
+    print "Running {} against {}:{}".format(script, host, db)
```
feanil, line 127: nit: use a `/` instead of `:`

util/vpc-tools/db-clone.py
```diff
@@ -1,181 +1,250 @@
...
+            'login_user': user,
+            'login_password': password,
+        }
+    }
+
+    skip_tags = ','.join(tag for flag,tag in 
+        [[args.ansible_skip_users,'users'], [args.ansible_skip_dbs,'dbs']] if flag)
+
+    cmd = "source {}; ansible-playbook -c local -i localhost, {} -e@{} -e '{}'".format(
+        VENV_ACTIVATE, play, db_file, json.dumps(overrides), '--skip-tags ' + (skip_tags or ''))
```
feanil: can we do this append of `skip-tags` as a separate line for clarity?

util/vpc-tools/db-clone.py
```diff
@@ -1,181 +1,250 @@
...
+    
+    sp = subprocess.Popen(cmd, shell=True, executable='/bin/bash', cwd=play_path,
```
feanil, line 165: why Popen instead of `check_output`?

util/vpc-tools/db-clone.py
maxrothman, line 127: MySQL uses `:` for this sort of thing

util/vpc-tools/requirements.txt
```diff
@@ -1,3 +1,5 @@
 boto
 docopt
 python-simple-hipchat==0.2
+pymysql
```
feanil, line 4: Why use this instead of mysql-python which is what we use in Django on our apps?

util/vpc-tools/db-clone.py
maxrothman, line 165: This way I can output the results of the command in real-time and get the return code.

util/vpc-tools/db-clone.py
maxrothman, line 80: I haven't tested this case, I prefer letting the boto call fail noisily rather than reimplementing its error checking. Perhaps we should choose a naming convention less likely to overflow. @e0d you were interested in that, right?

util/vpc-tools/db-clone.py
maxrothman, line 19: Eh, that's really long, and it's in the help text.

util/vpc-tools/requirements.txt
maxrothman, line 4: I just needed something to be able to run a raw sql query. Django has a whole multiple-database-compatible ORM system.

feanil, line 4: yes but it uses 'mysql-python' as the underlying driver for the mysql implementation.

maxrothman, line 4: Yeah, that uses the binary mysql driver underneath, which is challenging cross-platform. This is a pure-python api-compatible reimplementation of that.

fredsmith:
:+1:  LGTM

e0d:
:+1:

maxrothman:
Closing this in favor of adding the script to the sysadmin repo: https://github.com/edx-ops/sysadmin/pull/53

PR #2108, opened 2015-06-22
add tags for newrelic server monitoring

fredsmith:


playbooks/roles/newrelic/tasks/main.yml
```diff
@@ -59,6 +59,12 @@
   shell: >
     nrsysmond-config --set license_key="{{ NEWRELIC_LICENSE_KEY }}"
     ssl=true
+ 
+- name: configure newrelic labels
+  lineinfile: >
+    dest=/etc/httpd/conf/httpd.conf
```
feanil: we can't use `nrsysmond-config` to do this?

fredsmith: @feanil no, it doesn't understand how to set labels:

```
fredsmith@ip-10-3-10-207 i-aa0f4603:/etc/newrelic$ sudo nrsysmond-config --set    labels='Environment:stage;Deployment:edx;Play:edxapp'
Error: invalid configuration key: 'labels'
```

fredsmith: ah, but clearly that's the wrong file to be editing.

PR #2111, opened 2015-06-23
Better for backward compatibility

e0d:
May be better from a backward compatibility perspective.


feanil:
:+1:   Maybe also add a changelog entry,  long term we probably want a feature flag around the platform feature.

maxrothman:
:+1:

PR #2116, opened 2015-06-25
Add CREDIT_PROVIDER_SECRET_KEYS configuration to edxapp role

wedaly:
This exposes a setting to the Ansible scripts that we will need to configure courses for credit.  The original setting is here:

https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L2531-L2536

@feanil Could you please review this when you have a moment?  This isn't urgent to get in or configure, since we're not launching credit courses until August, but it will make sandbox testing less painful.


jibsheet:
@wedaly Used by both CMS and LMS?

wedaly:
@jibsheet Yes, the app that uses this is installed in `openedx.core.djangoapps` and is used by both Studio and LMS.

jibsheet:
:+1:

PR #2118, opened 2015-06-28
Add edxapp_common as an admin_jenkins dependency

feanil:
Because admin_jenkins runs various management commands locally, it needs to be able to run the edxapp locally.  This means that it needs to be able to install the system requirements needed by the edxapp so that pip install to local venvs work correctly.


feanil:
@edx/devops

maxrothman:
:+1:

PR #2119, opened 2015-06-29
DEVOPS-2168 Set up this cookie name for overriding

jibsheet:
Since the cookie's domain is *.edx.org use different names to help with
testing in stage/prod/edge.


jibsheet:
@e0d @maxrothman 

FYI @alawibaba @wedaly

maxrothman:
Should this go in generic_env and not lms_env?

jibsheet:
It shows up in lms/envs/common.py and cms/envs/common.py and is used in a common/djangoapps so I believe generic is correct

maxrothman:
:+1:

PR #2121, opened 2015-06-30
Update list of forwarded ports

andy-armstrong:
Add ports for edX Notes and for Bok Choy test servers. This is to save developers time, as otherwise they have to manually add these new port forwarding rules.

@feanil This is my first time creating a pull request against configuration, so let me know if there's more that I should be doing (automated tests, tagging others). Also, I'm assuming that I don't update the `Vagrantfile` under release and that a separate release process handles that. Thanks.


e0d:
:+1:

maxrothman:
:+1:

PR #2122, opened 2015-06-30
Don't build devstack from the aged release branch

e0d:


singingwolfboy:
I'm not sure that we want to use `master` as the default branch. `master` breaks more frequently than `release` does, and we want people to have working devstack installations.

e0d:
I don't think your statement is accurate.  I don't think that release is currently more likely to be what you want, which precisely the problem in this case.

When people are mounting their local version of edx-platform, it is likely a branch not too far away from master, but under active development.  When the config repo needs to change to support, say, new OS libraries, those changes need to be available in devestack.

For consistency, folks can always used the named release.

For the bleeding edge, I think master of config is the way to go.

We have been discussing removing the release branch entirely.

e0d:
@edx/devops looking for other thoughts here.

singingwolfboy:
Oh, this is for the branch in the `configuration` repository, not in the `edx-platform` repository. (Should have been obvious from the fact that the variable is named `CONFIG_VER`, but somehow I missed that.) Sure, that seems reasonable to me. :+1:

jibsheet:
agreed, configuration's release branch is unused right now and having devstack use it seems wrong
:+1:

singingwolfboy:
I just tried making this change locally in my Vagrantfile, and creating a new VM. [It failed with the error `TypeError: template_from_file() got an unexpected keyword argument 'vault_password'`](https://gist.github.com/singingwolfboy/ebb0b1aaaf45e1c82680). Any idea what's up with that?

e0d:
@db that looks like an ansible version issue related to the upgrade of ansible itself, what version does devstack report?

singingwolfboy:
```
vagrant@precise64:~$ /edx/bin/ansible-playbook --version
ansible-playbook 1.9.1-edx.2
  configured module search path = None
```

feanil:
The issue DB is seeing is related to upgrading from old versions to a newer version: https://github.com/ansible/ansible/issues/7252

@singingwolfboy  if you run `vagrant provision` on the VM after the failure do you see the error again?

singingwolfboy:
Re-running `vagrant provision` gets me further along, but it still ultimately fails. :frowning: https://gist.github.com/singingwolfboy/3b16a668b429e3672352

feanil:
I think we should merge this and resolve any issues we find in master.  I would like to get rid of the release branch since we don't seem to have time to maintain it.

singingwolfboy:
As @fredsmith so eloquently stated a few days ago:
We'll do it live

singingwolfboy:
When should we merge this?

PR #2124, opened 2015-07-01
Remove --tags=deploy when calling ansible-playbook for devstack

singingwolfboy:
@maxrothman: maybe this will fix [this error](https://gist.github.com/singingwolfboy/3b16a668b429e3672352)?


singingwolfboy:
Another failure: https://gist.github.com/singingwolfboy/071e8ce6703eebf68df5

feanil:
I guess we missed this when we made the same change to the update script.  This change is generally a good idea.

:+1:

PR #2128, opened 2015-07-06
Update to use the dash name for studio.

feanil:
Use studio- instead of studio. so that ssl works correctly.


jibsheet:
:+1: but let's delay merge until we fix the sandbox side to use the right URL

feanil:
:+1:  to your changes @jibsheet

feanil:
@jibsheet I looked around the sandbox with this change and it looks good to me.  I think this is ready to merge and we should send the developers an e-mail after we update the jenkins job.

PR #2129, opened 2015-07-07
Corrected path to Commerce API

clintonb:
XCOM-312

@feanil @rlucioni @jimabramson @Nickersoft


rlucioni:
:+1:

Let's make sure we don't deploy new ecommerce AMIs until the corresponding edx-platform changes are deployed.

PR #2133, opened 2015-07-08
Stop celery workers using SIGINT instead of SIGTERM.

feanil:
According to the documentation TERM should do a graceful shutdown:
http://celery.readthedocs.org/en/latest/userguide/workers.html#process-signals

but this is not actually the case:
https://groups.google.com/forum/#!topic/celery-users/t8g5KvIvQZ8

SIGINT on the other hand seems to do what we want.

@edx/devops


e0d:
:+1:

maxrothman:
Are we sure their documentation is wrong? if so, have we notified them?

feanil:
@maxrothman please see the links in the description.

maxrothman:
Is there a way to formally file a bug other than posting in the google group?

feanil:
https://github.com/celery/celery/issues/2700

maxrothman:
:+1:

PR #2134, opened 2015-07-08
workaround invalid upstream certs when installing browsers

fredsmith:


e0d:
Seems like the simplest possible fix, but do we know why this has started now?

Still, :+1:

fredsmith:
My guess is that s3 changed their cert to one that's incompatible with ansible 1.5.5

jibsheet:
I assumed it was the SSLv3 deprecation.  I assumed those would all have broken on May 20th, but maybe they found more buckets to deprecate?

ghost:
Hi folks! Is there any way to fix it for the birch release also? Can't get my head on how to fix it myself.

PR #2135, opened 2015-07-09
Move db-clone.py to sysadmin repo

maxrothman:
See https://github.com/edx-ops/sysadmin/pull/53
@edx/devops


feanil:
Are jenkins jobs that depend on this updated or obseleted?

maxrothman:
Not yet, but we'll do that before merging.

PR #2138, opened 2015-07-09
Don't pkill some services.

feanil:
Certain services will shutdown gracefully and so they may not be dead by the
time this task is run.  Those services should not be explicitly killed since
they may need to do things like finish a grading run or other instructor tasks.

@edx/devops If you guys have other suggestions for how to do this let me know.  But this does what I need and prevents workers from getting killed prematurely.


fredsmith:
:+1:

jibsheet:
While worried about what will wedge and prevent shutdown, this is clearly better than what was before :+1:

PR #2141, opened 2015-07-10
Derf/mongo 3 0

fredsmith:


playbooks/edx-east/mongo.yml
```diff
@@ -3,11 +3,10 @@
   sudo: True
   gather_facts: True
   vars:
-    serial_count: 1
+    serial_count: 3
```
feanil, line 6: Should this be changed back to 1 based on what we talked about yesterday?

playbooks/edx-east/mongo.yml
```diff
@@ -3,11 +3,10 @@
...
   serial: "{{ serial_count }}"
   roles:
     - mongo
-    - mongo_mms
```
feanil, line 10: Should this be added back with a flag around it?  Something like `COMMON_ENABLE_MMS`

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -32,7 +32,8 @@
 
 - name: install mongo server and recommends
   apt: >
-    pkg=mongodb-org={{ mongo_version }}
+  apt: 
+    pkg={{','.join(mongodb_debian_pkgs)}}
```
feanil, line 45: Why not use `with_items` instead?

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -64,7 +66,7 @@
   when: MONGO_CLUSTERED
 
 - name: copy configuration template
-  template: src=mongodb.conf.j2 dest=/etc/mongod.conf backup=yes
+  template: src=mongodb-standalone.conf.j2 dest=/etc/mongod.conf backup=yes
```
feanil, line 78: Should we only do this when `MONGO_CLUSTERED` is false?

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -75,29 +77,51 @@
...
+  when: MONGO_CLUSTERED
+
+- name: wait for mongo server to start
+  wait_for: port=27017 delay=2
+  when: MONGO_CLUSTERED
 
 - name: Create the file to initialize the mongod replica set
   template: src=repset_init.js.j2 dest=/tmp/repset_init.js
-  when: MONGO_CLUSTERED
+  when: MONGO_CLUSTERED and mongo_primary == ansible_default_ipv4["address"]
```
feanil: End users are going to want to set the `mongo_primary` var so let's make it ALL_CAPS.

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -75,29 +77,51 @@
...
+  file: path=/tmp/repset_init.js state=absent
+  when: MONGO_CLUSTERED and mongo_primary == ansible_default_ipv4["address"] 
+
+- name: Create the file to add hosts to the mongod replica set
+  template: src=repset_add_slaves.js.j2 dest=/tmp/repset_add_slaves.js
+  when: MONGO_CLUSTERED and mongo_primary == ansible_default_ipv4["address"] 
+
+- name: Initialize the replication set
+  shell: >
+    /usr/bin/mongo /tmp/repset_add_slaves.js
```
feanil: Nit: Change this to `repset_add_secondaries.js`

feanil:
Looks like you haven't checked in the the new template files that you created.

playbooks/edx-east/mongo.yml
fredsmith, line 6: no, with the updates it works with 3 (or more) servers in serial.

playbooks/edx-east/mongo.yml
fredsmith, line 10: mms shouldn't be on every server, only on a central server that can see all hosts in the mms account.

playbooks/roles/mongo/tasks/main.yml
fredsmith, line 78: no, we always want to start up with standalone first, to create the users.   You can't create users once the replicaset has been started.

carsongee, line 78: But that is working fine as the role is now, does 3.0 change this?

carsongee:
Can we make an upgrade play or something this time around? 2.4 to 2.6 wasn't too bad, but 2.6 to 3.0 looks like it has quite a few changes with the storageEngine changes

fredsmith:
@carsongee this is _supposed_ to be a drop-in replacement, but 3.0 might eat all your data.  I think our suggestion should be to spin up a new cluster, make it a secondary to your current cluster, and then step down the primary.  We're testing that transition here.

Upgrading a cluster in place might just be a matter of timing and step-downs.  I'd be happy to chat about how that might work.

carsongee:
It should be drop in, but if this role is intended to do that, we need some steps to update to the new storageEngine: http://docs.mongodb.org/manual/release-notes/3.0-upgrade/#change-replica-set-storage-engine-to-wiredtiger

playbooks/roles/mongo/tasks/main.yml
fredsmith, line 78: yes, 3.0 changes some things regarding security.

carsongee, line 78: weird, the only mention I saw on security was the change of the hash to SCRAM-SHA-1

playbooks/roles/mongo/tasks/main.yml
fredsmith, line 45: with_items doesn't work with the packagename=version syntax.

fredsmith:
@carsongee an upgrade play is going to be a different PR/Story.   Let me know if you have any other concerns with this PR.

One thing to note is that mongo_extra_conf should be in YAML format instead of INI format, so that will be a breaking change for this.  However, it's an internal variable that we don't currently set anywhere else, so I'm not sure of the impact.

fredsmith:
blocked waiting on https://github.com/edx/edx-platform/pull/8692

PR #2142, opened 2015-07-10
Feanil/update minos version

feanil:
@edx/devops 

Pinning minos to a specific version correcting a template name.


jibsheet:
:+1:

PR #2143, opened 2015-07-10
Update the minos requirement var.

feanil:
Changed what version means but forgot to update the minos_requirement var.


jibsheet:
:+1:

PR #2146, opened 2015-07-14
update ec2 plugin version on testeng jenkins

clytwynec:
@benpatterson @jzoldak  Just updated this on the test jenkins.  None of it's dependencies updated, so it's just the ec2 plugin version that changed.


jzoldak:
:+1:

benpatterson:
:+1: but let's merge once you've tested it out.

PR #2151, opened 2015-07-15
Merge rbenv change to rc.

feanil:
@singingwolfboy FYI


singingwolfboy:
Can you remove the extra merge commit from this PR, and make sure it's reflected on https://openedx.atlassian.net/wiki/display/OPEN/Cypress+Bugfixes ?

feanil:
@singingwolfboy FYI,  I'm happy to remove the merge commits but why?  They show which PRs actually made the changes that got merged into the RC.

singingwolfboy:
I had assumed they were useless noise, but you're right, they do have valuable info. It looks like some more commits have crept into your PR though, which I'm not sure are intentional. Can you double-check?

feanil:
They are a fix for testeng, it doesn't effect the deployment of other parts of configuration, so I saw no reason to keep it out.

PR #2152, opened 2015-07-15
Use symbolic mode instead of explicit.

feanil:
Since we don't want to break other permissions we just want to give
the group write permission.

@jibsheet


jibsheet:
:+1:

PR #2157, opened 2015-07-20
Update sandbox base AMI ids.

feanil:
The new sandbox is built from master around cypress.

I also updated the base ubuntu images to the latest ones.

@edx/devops


jibsheet:
:+1:

PR #2161, opened 2015-07-20
Remove the 3rd party auth vars.

feanil:
3rd Party auth is now managed via django admin an not from json config.

https://github.com/edx/edx-platform/pull/8140/files#diff-8a23909bc8d201fba8525b1dd6396247

@edx/devops


jibsheet:
(grep fail, nm) :+1:

PR #2162, opened 2015-07-21
Update migration script for Cypress

singingwolfboy:
Ansible was upgraded between Birch and Cypress, so the migration script can no longer use the already-installed version of Ansible -- we need to create a new virtualenv, instead.

### DIFF
--- a/util/vagrant/migrate.sh
+++ b/util/vagrant/migrate.sh
@@ -80,23 +80,10 @@ fi
 if [[ $TARGET == *cypress* && $INTERACTIVE == true ]] ; then
   cat <<"EOM"
           WARNING WARNING WARNING WARNING WARNING
-The Cypress release of Open edX upgraded the version of Celery used
-by the edx-platform workers.  This change was introduced to fix the fact
-that when trying to restart the celery worker, the process hangs while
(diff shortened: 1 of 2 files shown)


util/vagrant/migrate.sh
```diff
@@ -70,6 +70,33 @@ EOM
   fi
 fi
 
+if [[ $TARGET == *cypress* && $INTERACTIVE == true ]] ; then
+  cat <<"EOM"
+          WARNING WARNING WARNING WARNING WARNING
+The Cypress release of Open edX upgraded the version of Celery used
+by edx-platform workers.  This change was introduced to fix the fact
+that or restart the celery worker sometimes blocks indefinitely and
```
singingwolfboy: @feanil the grammar is a bit weird here -- I'm not sure what you're trying to say. Can you clean it up?

singingwolfboy:
I just tried the migration script again, and it still hangs at the same point on devstack. Looks like [its trying to kill the workers, but having trouble?](https://gist.github.com/singingwolfboy/5117cebe0a1955a922e8#file-gistfile1-txt-L48)

PR #2165, opened 2015-07-22
Don't validate certs from S3.

feanil:
Ansible 1.5.5 doesn't support validating certs if SSL v3 is not supported.

@singingwolfboy is this the correct branch to PR this change into?


feanil:
@singingwolfboy any reason no to merge this?

singingwolfboy:
Looks good to me. :+1:

PR #2166, opened 2015-07-22
set YouTube API key for Birch

singingwolfboy:
#2164 for Birch. Needed for https://github.com/edx/edx-platform/pull/8992. @fredsmith, can you make sure this is OK?


fredsmith:
:+1:

PR #2168, opened 2015-07-22
Make the availability of the LMS Credit API configurable

rlucioni:
The Course Administration Tool needs access to a Credit API endpoint, but we don't want to enable the Credit API by default for all Open edX users. I'll follow this PR with support tickets requesting that the Credit API be enabled in stage and prod.

@feanil @wedaly @clintonb


rlucioni:
I've opened [DEVOPS-2357](https://openedx.atlassian.net/browse/).

feanil:
:+1: 

@fredsmith FYI since you are ticket overlord this week.

PR #2171, opened 2015-07-24
Create ecommerce MySQL user with the correct name

rlucioni:
Formerly, the database user was set up with username `ecommerce`, not the the user configured for the default ecommerce database, `ecomm001`.

@feanil

### DIFF
--- a/playbooks/roles/edxlocal/tasks/main.yml
+++ b/playbooks/roles/edxlocal/tasks/main.yml
@@ -111,7 +111,7 @@
 
 - name: setup users for ecommerce
   mysql_user: >
-    name="{{ ECOMMERCE_DEFAULT_DB_NAME }}"
+    name="{{ ECOMMERCE_DATABASES.default.USER }}"
     password="{{ ECOMMERCE_DATABASES.default.PASSWORD }}"
     priv='{{ ECOMMERCE_DEFAULT_DB_NAME }}.*:SELECT,INSERT,UPDATE,DELETE'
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxlocal/tasks/main.yml
```diff
@@ -111,10 +111,10 @@
 
 - name: setup users for ecommerce
   mysql_user: >
-    name="{{ ECOMMERCE_DEFAULT_DB_NAME }}"
+    name="{{ ECOMMERCE_DATABASES.default.USER }}"
     password="{{ ECOMMERCE_DATABASES.default.PASSWORD }}"
     priv='{{ ECOMMERCE_DEFAULT_DB_NAME }}.*:SELECT,INSERT,UPDATE,DELETE'
-  when: ECOMMERCE_DEFAULT_DB_NAME is defined
+  when: ECOMMERCE_DATABASES is defined
```
feanil: I would leave this keying off of the DB_NAME still as we do with other apps.  The DB_NAME is still used as the conditional key for most other database.  I think we want to standardize on that.

feanil:
:+1:  after you change the when condition back.

playbooks/roles/edxlocal/tasks/main.yml
rlucioni: Updated.

feanil:
@rlucioni :goat: :four: :it:

PR #2172, opened 2015-07-24
Ecommerce in devstack

rlucioni:
This is a work in progress which adds the ecommerce service to devstack. Remaining tasks include:
- ~~Automating OAuth2 Client creation (management command needs to be made idempotent)~~
  - ~~Also requires overriding local ecommerce OIDC settings with values from configuration~~
- ~~Documentation:~~
  - ~~Commands used to turn on ecommerce~~
  - ~~LMS configuration changes required to interface with the service~~

FYI @feanil @clintonb @jimabramson


clintonb:
:+1:

rlucioni:
Work to make the OAuth2 Client creation management command idempotent is up at https://github.com/edx/edx-oauth2-provider/pull/27.

rlucioni:
Devstack-specific settings module for ecommerce is up at https://github.com/edx/ecommerce/pull/227. Together with the changes in https://github.com/edx/edx-oauth2-provider/pull/27, automatic OAuth2 Client creation during devstack provisioning is working!

jimabramson:
looks great to me @rlucioni

rlucioni:
The commands used to turn on ecommerce and the modifications which need to be made to LMS configuration are documented on the [Open edX Operations wiki](https://openedx.atlassian.net/wiki/display/OpenOPS/How+to+Run+the+Ecommerce+Service+in+Devstack?flashId=-957986068).

rlucioni:
All PRs required to facilitate this work have been merged.

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -145,3 +135,18 @@
     name={{ ecommerce_service_name }}
   when: not disable_edx_services
   sudo_user: "{{ supervisor_service_user }}"
+
+- name: create OAuth2 Client
+  shell: >
+    {{ COMMON_BIN_DIR }}/manage.edxapp lms --settings=aws
```
feanil: You need to run this with `{{ COMMON_BIN_DIR }}/python.edxapp` so it uses the correct virtualenv.

feanil: Can we pull this out of the ecommerce role into the local_dev role, that role is expected to use vars from multiple roles but I'd rather not have the ecommerce role mention the `edxapp_user`

rlucioni: @feanil is there a difference between specifying the `environment` property as part of the task and running the management command with `{{ COMMON_BIN_DIR }}/python.edxapp`? I thought specifying `environment: "{{ edxapp_environment }}"` would cause the correct virtualenv to be used, but it sounds like that may not be the case.

I'll look into moving this to the local_dev role.

playbooks/vagrant-devstack.yml
```diff
@@ -8,6 +8,7 @@
     devstack: true
     disable_edx_services: true
     mongo_enable_journal: false
+    create_oauth2_clients: true
```
feanil: Why have this flag at all?  If the local_dev role is included I can't imagine a time when you wouldn't want to make sure the client connection is setup, especially now that it is idempotent.

feanil:
One last comment to address, otherwise :+1: 

As a result of some of the defaults that have been removed are there settings that should be removed from our secure var files?

Specifically it seems like the following two no longer exist:

```
ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY : 'some-secret'
ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_URL_ROOT : 'some-secret'
```

playbooks/vagrant-devstack.yml
rlucioni: Good call, removed.

playbooks/roles/local_dev/tasks/main.yml
```diff
@@ -76,3 +76,19 @@
...
+    create_oauth2_client
+    {{ item.url_root }}
+    "{{ item.url_root }}/complete/edx-oidc/"
+    confidential
+    --client_name {{ item.name }}
+    --client_id {{ item.id }}
+    --client_secret {{ item.secret }}
+    --trusted
+  sudo_user: "{{ edxapp_user }}"
+  environment: "{{ edxapp_environment }}"
```
rlucioni, line 92: @feanil is there a difference between specifying the `environment` property as part of the task and running the management command with `{{ COMMON_BIN_DIR }}/python.edxapp`? I thought specifying `environment: "{{ edxapp_environment }}"` would cause the correct virtualenv to be used, but it sounds like that may not be the case. Is specifying `environment` necessary?

rlucioni:
@feanil yes, `ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY` and `ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_URL_ROOT` can be removed from our secure var files. These settings are still required in order to configure ecommerce, but their values can be derived from other vars.

feanil:
@rlucioni I just merged the corresponding change in secure vars to remove `ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY`.

:+1:   :sheep: :it: :dog:

rlucioni:
@feanil great, thanks. What about `ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_URL_ROOT`?

feanil:
We were never overriding it in secure as far as I could tell.

rlucioni:
Cool. :fireworks:

PR #2175, opened 2015-07-27
Change when we tag instances.

feanil:
In all other roles, tagging is not done implicitly on EC2 instances, it is done when the tagging flag is set to true.  The service role should follow the same pattern.

This caused a problem on an open source deployment here: https://groups.google.com/forum/#!msg/openedx-ops/GTaZi6tAVOM/CCF_OSpxbsEJ


feanil:
@edx/devops

sarina:
@feanil is this something that will have to get into Cypress as well?

jibsheet:
:+1:

feanil:
@sarina it's a bug in analytics-api, insights and ecommerce roles(they use the same core logic) but we don't ship those right now in the open source release so my instinct would be that we don't need it in the cypress RC.

sarina:
Awesome :)

PR #2176, opened 2015-07-27
Install analytics stack dependencies

mulby:
This allows you to deploy the analytics stack to an existing devstack. The outstanding issues are highlighted in TODO.txt (which will be removed before merging) - thoughts on these would be greatly appreciated as I suspect I am not the first person to encounter them.

Primary Reviewers: @e0d @feanil 
Other Reviewers: @brianhw @jab5569 @bradenmacdonald


TODO.txt
```diff
@@ -0,0 +1,23 @@
+* First run of ansible always fails - I suspect a version bootstrapping issue
```
mulby: @e0d - Not sure what it takes to roll out a new vagrant box, but that may be the easiest way to address this issue, was something like this planned for cypress?

TODO.txt
```diff
@@ -0,0 +1,23 @@
...
+    * .rbenv is modified in forums (this may be related to the issue above)
+* fix auto-registration of new oauth2 client in LMS (insights role)
```
mulby: this happens to me every time I reprovision and I'm nervous it will affect the release. The observed behavior is that although the requirements file (github.txt) shows the proper library version (0.5.5) the virtualenv is still pointing at 0.3.1 after provisioning.

@rlucioni @e0d

TODO.txt
```diff
@@ -0,0 +1,23 @@
...
+    * the new version of edx-oauth2-provider is not being installed for some reason?
+* manual addition of "DEBUG: true" to /edx/etc/insights.yml in order to get django to serve static assets from "runserver"
```
mulby: @e0d this seemed like the most simple way to get "runserver" to serve up static assets... the other option is to tell users to start the service with runserver --insecure which also does the trick. Do you have a preference here?

TODO.txt
```diff
@@ -0,0 +1,23 @@
...
+* ensure hdfs://localhost:9000/data/ directory exists
+    * change permissions of /edx/var/log/tracking/tracking.log to 644
+
+
+* document instructions
+    * vagrant up
+    * as vagrant user run "/edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook -i localhost, -c loc
+al analytics_single.yml"
+    * sudo su hadoop
+        * edx-analytics-hadoop.sh start
```
bradenmacdonald: You could also consider making this automatic - I did that on the separate devstack and it appears to work well: https://github.com/open-craft/edx-analytics-devstack/commit/279892b7afa25619b3b5db8b787bb34d6d2796c2

TODO.txt
rlucioni: @mulby is the library stuck at 0.3.1 or 0.5.1? 0.5.2 switched away from installing the library as editable, which may be the root cause of the issue you're observing. It's worth noting that I haven't encountered this issue when provisioning fresh devstacks.

mulby: I would have to do a fresh run of devstack to see what version it shows for me. It's interesting that you don't see this when you provision fresh... Maybe it's an issue with the base vagrant box and we are using different Vagrantfiles?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -686,7 +686,7 @@ generic_env_config:  &edxapp_generic_env
   CROSS_DOMAIN_CSRF_COOKIE_NAME: "{{ EDXAPP_CROSS_DOMAIN_CSRF_COOKIE_NAME }}"
   VIDEO_UPLOAD_PIPELINE: "{{ EDXAPP_VIDEO_UPLOAD_PIPELINE }}"
   DEPRECATED_ADVANCED_COMPONENT_TYPES: "{{ EDXAPP_DEPRECATED_ADVANCED_COMPONENT_TYPES }}"
-  OAUTH_OIDC_ISSUER: "https://{{ EDXAPP_LMS_BASE }}/oauth2"
+  OAUTH_OIDC_ISSUER: "{{ EDXAPP_LMS_BASE_SCHEME | default('http') }}://{{ EDXAPP_LMS_BASE }}/oauth2"
```
mulby: We will have to update the production settings to set EDXAPP_LMS_BASE_SCHEME to `https`

playbooks/roles/insights/defaults/main.yml
```diff
@@ -11,21 +11,26 @@
...
 INSIGHTS_PRIVACY_POLICY_URL: '{{ INSIGHTS_MKTG_BASE }}/privacy-policy'
 INSIGHTS_TERMS_OF_SERVICE_URL: '{{ INSIGHTS_MKTG_BASE }}/terms-service'
 INSIGHTS_SUPPORT_URL: ''
 INSIGHTS_OAUTH2_SECRET: 'secret'
-INSIGHTS_OAUTH2_URL_ROOT: 'url_root'
+INSIGHTS_OAUTH2_URL_ROOT: '{{ INSIGHTS_LMS_BASE }}/oauth2'
+INSIGHTS_OAUTH2_APP_CLIENT_NAME: insights
+INSIGHTS_OAUTH2_APP_USERNAME: staff
+INSIGHTS_OAUTH2_APP_COMPLETE_URL: '{{ INSIGHTS_BASE_URL }}/complete/edx-oidc/'
+INSIGHTS_SOCIAL_AUTH_REDIRECT_IS_HTTPS: false
```
mulby, line 27: We will have to override this in our production settings to be `true`.

vagrant/base/devstack/Vagrantfile
```diff
@@ -117,5 +132,11 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         xqueue_version: ENV['OPENEDX_RELEASE'],
       }
     end
+    if ENV['CONFIGURATION_VERSION']
```
mulby, line 145: Open to ideas about how to do this better. I needed my branch to be deployed to `/edx/app/edx_ansible/` in order to be able to run the `analytics_single.yml` playbook once the normal devstack process was complete.

mulby:
@e0d this is ready for review

mulby:
@feanil this is ready for review

mulby:
FYI - @brianhw @jab5569 @bradenmacdonald

mulby:
Another note (for context):

Currently the way the pipeline works is that we provision a "bare metal" EMR cluster using boto and then run an ansible playbook from the jenkins box that is coordinating the whole operation against the master node of the cluster. Here's the playbook: https://github.com/edx/edx-analytics-pipeline/blob/master/share/task.yml

The playbook basically deploys and configures a specified version of the pipeline on the cluster.

This PR keeps that model. It does not actually deploy edx-analytics-pipeline to the vagrant box and instead expects the user to use the same tools Jenkins uses to deploy the pipeline to EMR clusters but instead the developer would be using their local machine to deploy the pipeline to the VM.

An alternative approach could be to deploy edx-analytics-pipeline the same way we deploy other services (using the NFS shared dir etc) which would mean the user wouldn't need to install _any_ other stuff on their host machine. To do this, though, we would have to either:
1. duplicate the pipeline deploy logic in both task.yml and the analytics_pipeline role (yuck)
   OR
2. create an EMR bootstrap script that downloads edx/configuration and deploys the pipeline using the same role that is used to deploy it to devstack to deploy it in prod - this would be a fairly risky and significant change. It would also have fairly dramatic impacts on our acceptance testing strategy.

(GM edit) - After some discussion with the team we are pushing out any changes here into a future story in order to manage scope creep. For the time being users will have to bootstrap the pipeline on their host before running `remote-task` pointing at the VM.

mulby:
Settings that will need to be added to the secure repo in order to not break prod:

``` yaml
EDXAPP_LMS_BASE_SCHEME: https
INSIGHTS_SOCIAL_AUTH_REDIRECT_IS_HTTPS: true
INSIGHTS_SESSION_COOKIE_NAME: sessionid
INSIGHTS_CSRF_COOKIE_NAME: crsftoken
INSIGHTS_LANGUAGE_COOKIE_NAME: language
```

mulby:
This PR details how this stuff is intended to be used:

https://github.com/edx/edx-documentation/pull/461

Note that it also depends on this PR: https://github.com/edx/edx-analytics-pipeline/pull/135

playbooks/roles/analytics_pipeline/defaults/main.yml
```diff
@@ -0,0 +1,49 @@
...
+ANALYTICS_PIPELINE_LUIGI_HADOOP_STREAMING_JAR: "{{ HADOOP_COMMON_HOME }}/share/hadoop/tools/lib/hadoop-streaming-{{ HADOOP_COMMON_VERSION }}.jar"
+
+#
+# vars are namespaced with the module name.
+#
+analytics_pipeline_role_name: analytics_pipeline
+
+analytics_pipeline_util_library:
+  path: /tmp/edx-analytics-hadoop-util
+  repo: https://github.com/mulby/edx-analytics-hadoop-util
```
mulby: We should probably move this into the edx organization.

playbooks/roles/analytics_pipeline/defaults/main.yml
```diff
@@ -0,0 +1,49 @@
...
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role analytics_pipeline
+#
+
+ANALYTICS_PIPELINE_OUTPUT_DATABASE:
+  username: pipeline001
+  password: password
```
brianhw, line 16: Should these be parameterized as well?

playbooks/roles/hadoop_common/defaults/main.yml
```diff
@@ -0,0 +1,58 @@
...
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role hadoop_common
+# 
+
+HADOOP_COMMON_VERSION: 2.3.0
```
brianhw: Just out of curiosity, is this also something that should be pulled out?  And is there a particular reason that 2.3.0 was selected?

playbooks/roles/insights/defaults/main.yml
```diff
@@ -105,13 +110,18 @@ INSIGHTS_CONFIG:
...
+  CSRF_COOKIE_NAME: "{{ INSIGHTS_CSRF_COOKIE_NAME | default('insights_csrftoken') }}"
+  LANGUAGE_COOKIE_NAME: "{{ INSIGHTS_LANGUAGE_COOKIE_NAME | default('insights_language') }}"
 
 INSIGHTS_NEWRELIC_APPNAME: "{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-analytics-api"
 INSIGHTS_PIP_EXTRA_ARGS: "-i {{ COMMON_PYPI_MIRROR_URL }}"
 INSIGHTS_NGINX_PORT: "18110"
 INSIGHTS_GUNICORN_WORKERS: "2"
 INSIGHTS_GUNICORN_EXTRA: ""
-INSIGHTS_COURSE_API_URL: "URL FOR COURSE API"
+INSIGHTS_COURSE_API_URL: "{{ INSIGHTS_LMS_BASE}}/api/course_structure/v0/"
```
brianhw: nit: space after LMS_BASE

playbooks/roles/insights/tasks/main.yml
```diff
@@ -132,4 +132,10 @@
...
   when: not disable_edx_services
-  sudo_user: "{{ supervisor_service_user }}"
\ No newline at end of file
+  sudo_user: "{{ supervisor_service_user }}"
+
+- name: register oauth2 application
+  shell: >
+    . /edx/app/edxapp/edxapp_env && {{ COMMON_BIN_DIR }}/manage.edxapp lms --settings=aws create_oauth2_client "{{ INSIGHTS_BASE_URL }}" "{{ INSIGHTS_OAUTH2_APP_COMPLETE_URL }}" confidential --username "{{ INSIGHTS_OAUTH2_APP_USERNAME }}" --client_name "{{ INSIGHTS_OAUTH2_APP_CLIENT_NAME }}" --client_id "{{ INSIGHTS_OAUTH2_KEY }}" --client_secret "{{ INSIGHTS_OAUTH2_SECRET }}" --trusted
+  sudo_user: edxapp
+  when: insights_register_oauth_app|bool
```
brianhw: nit: do we need newlines?

playbooks/roles/sqoop/defaults/main.yml
```diff
@@ -0,0 +1,39 @@
...
+SQOOP_HOME: "{{ HADOOP_COMMON_USER_HOME }}/sqoop"
+SQOOP_CONF: "{{ SQOOP_HOME }}/conf"
+SQOOP_LIB: "{{ SQOOP_HOME }}/lib"
+
+#
+# vars are namespace with the module name.
+#
+sqoop_role_name: sqoop
+sqoop_dist:
+  filename: "sqoop-{{ SQOOP_VERSION }}.bin__hadoop-2.0.4-alpha.tar.gz"
```
brianhw: Is this actually tied to a particular Hadoop version?  Or is this the version that is used for all Hadoop 2 versions?  Having alpha in the name looks suspicious....

playbooks/roles/sqoop/defaults/main.yml
```diff
@@ -0,0 +1,39 @@
...
+  url: "http://www.carfab.com/apachesoftware/sqoop/{{ SQOOP_VERSION }}/sqoop-{{ SQOOP_VERSION }}.bin__hadoop-2.0.4-alpha.tar.gz"
+  sha256sum: d582e7968c24ff040365ec49764531cb76dfa22c38add5f57a16a57e70d5d496
+sqoop_mysql_connector_dist:
+  filename: "mysql-connector-java-{{ SQOOP_MYSQL_CONNECTOR_VERSION }}.tar.gz"
+  url: "http://dev.mysql.com/get/Downloads/Connector-J/mysql-connector-java-{{ SQOOP_MYSQL_CONNECTOR_VERSION }}.tar.gz"
+  sha256sum: 04ad83b655066b626daaabb9676a00f6b4bc43f0c234cbafafac1209dcf1be73
```
brianhw: Should these be put as variables as well?  Since the sha should change from one url to another, and the url is parametrized...

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: distribution extracted
+  shell: >
+    chdir={{ HADOOP_COMMON_USER_HOME }}
+    tar -xzf /tmp/{{ sqoop_dist.filename }} && chown -R {{ hadoop_common_user }}:{{ hadoop_common_group }} sqoop-{{ SQOOP_VERSION }}.bin__hadoop-2.0.4-alpha
+  when: not installed.stat.exists
+
+- name: versioned directory symlink created
+  file: >
+    src={{ HADOOP_COMMON_USER_HOME }}/sqoop-{{ SQOOP_VERSION }}.bin__hadoop-2.0.4-alpha
```
brianhw: Should this be a variable?  Particularly the hadoop-2.0.4-alpha part....

brianhw:
LGTM, FWIW.

playbooks/roles/hadoop_common/defaults/main.yml
```diff
@@ -0,0 +1,58 @@
...
+HADOOP_COMMON_USER_HOME: /home/hadoop
```
feanil: For anything that we install that is more complicated than an apt-get install we have the homes be under `{{ COMMON_APP_DIR }}`  This should default to something like `{{ COMMON_APP_DIR }}/hadoop` unless there is a good reason for not doing that.

playbooks/roles/hadoop_common/defaults/main.yml
```diff
@@ -0,0 +1,58 @@
...
+HADOOP_COMMON_HOME: "{{ HADOOP_COMMON_USER_HOME }}/hadoop"
+HADOOP_COMMON_LOGS: "{{ HADOOP_COMMON_HOME }}/logs"
+HADOOP_COMMON_DATA: /var/lib/hadoop/hdfs
+HADOOP_COMMON_CONF_DIR: "{{ HADOOP_COMMON_HOME }}/etc/hadoop"
```
feanil, line 21: Config usually goes under `{{ COMMON_CFG_DIR }}/{{ appname }}`

playbooks/roles/hadoop_common/defaults/main.yml
```diff
@@ -0,0 +1,58 @@
...
+#
+##
+# Defaults for role hadoop_common
+# 
+
+HADOOP_COMMON_VERSION: 2.3.0
+HADOOP_COMMON_USER_HOME: /home/hadoop
+HADOOP_COMMON_HOME: "{{ HADOOP_COMMON_USER_HOME }}/hadoop"
+HADOOP_COMMON_LOGS: "{{ HADOOP_COMMON_HOME }}/logs"
+HADOOP_COMMON_DATA: /var/lib/hadoop/hdfs
```
feanil: Data usually goes under `{{ COMMON_DATA_DIR }}/{{ appname }}`

playbooks/roles/insights/tasks/main.yml
```diff
@@ -132,4 +132,10 @@
     config={{ supervisor_cfg }}
     name={{ insights_service_name }}
   when: not disable_edx_services
-  sudo_user: "{{ supervisor_service_user }}"
\ No newline at end of file
+  sudo_user: "{{ supervisor_service_user }}"
+
+- name: register oauth2 application
```
feanil: This should not live in the insights role.  It should live in the local_dev role.  Following a similar pattern to what ecommerce did: https://github.com/edx/configuration/blob/master/playbooks/roles/local_dev/tasks/main.yml#L80

You just need to add your vars to the list here: https://github.com/edx/configuration/blob/master/playbooks/roles/local_dev/defaults/main.yml#L33

playbooks/roles/sqoop/defaults/main.yml
feanil: +1 to this.  @mulby can you provide more info about sqoop versioning?

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+
+- name: check if downloaded and extracted
+  stat: path={{ SQOOP_LIB }}/mysql-connector-java-{{ SQOOP_MYSQL_CONNECTOR_VERSION }}-bin.jar
+  register: installed
+
+- name: distribution downloaded
+  get_url: >
+    url={{ sqoop_dist.url }}
+    sha256sum={{ sqoop_dist.sha256sum }}
+    validate_certs=no
```
feanil: Why not?  This should not be necessary with ansible 1.9

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+    dest=/tmp
+  when: not installed.stat.exists
+
+- name: distribution extracted
+  shell: >
+    chdir={{ HADOOP_COMMON_USER_HOME }}
+    tar -xzf /tmp/{{ sqoop_dist.filename }} && chown -R {{ hadoop_common_user }}:{{ hadoop_common_group }} sqoop-{{ SQOOP_VERSION }}.bin__hadoop-2.0.4-alpha
```
feanil: Why not use the [unarchive module](http://docs.ansible.com/ansible/unarchive_module.html)?

playbooks/roles/sqoop/tasks/main.yml
feanil: +1 to making this a variable in `defaults/main.yml`.

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+- name: check if downloaded and extracted
+  stat: path={{ SQOOP_LIB }}/mysql-connector-java-{{ SQOOP_MYSQL_CONNECTOR_VERSION }}-bin.jar
+  register: installed
+
+- name: distribution downloaded
+  get_url: >
+    url={{ sqoop_dist.url }}
+    sha256sum={{ sqoop_dist.sha256sum }}
+    validate_certs=no
+    dest=/tmp
```
feanil: If a machine gets re-started `/tmp` might get wiped.  Is it okay for this role to run multiple times on the same machine?

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+  file: >
+    src={{ HADOOP_COMMON_USER_HOME }}/sqoop-{{ SQOOP_VERSION }}.bin__hadoop-2.0.4-alpha
+    dest={{ SQOOP_HOME }}
+    owner={{ hadoop_common_user }} group={{ hadoop_common_group }} state=link
+
+- name: mysql connector distribution downloaded
+  get_url: >
+    url={{ sqoop_mysql_connector_dist.url }}
+    sha256sum={{ sqoop_mysql_connector_dist.sha256sum }}
+    validate_certs=no
```
feanil: Should no longer be needed.

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+    dest=/tmp
```
feanil: Same question about `/tmp` as above.

playbooks/roles/sqoop/tasks/main.yml
```diff
@@ -0,0 +1,84 @@
...
+  when: not installed.stat.exists
+
+- name: mysql connector distribution extracted
+  shell: >
+    chdir=/tmp
+    tar -xzf /tmp/{{ sqoop_mysql_connector_dist.filename }}
```
feanil: Use the [unarchive module](http://docs.ansible.com/ansible/unarchive_module.html).

vagrant/base/analyticstack/Vagrantfile
```diff
@@ -0,0 +1,142 @@
+Vagrant.require_version ">= 1.5.3"
```
feanil, line 1: We seem to be moving away from the model of having different devstack images for different services.  @e0d what is your thought on this?

playbooks/vagrant-devstack.yml
```diff
@@ -28,5 +30,3 @@
     - browsermob-proxy
     - local_dev
     - demo
-    - role: analytics_api
```
feanil, line 32: If you want insights/analytics_api on the standard devstack you will have to add the roles to this file.  Adding the folder mounts to the `Vagrantfile` is not sufficient.

playbooks/roles/analytics_pipeline/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+
+- name: store input database credentials for analytics pipeline
+  copy: >
+    content="{{ ANALYTICS_PIPELINE_INPUT_DATABASE | to_json }}"
+    dest={{ COMMON_CFG_DIR }}/edx-analytics-pipeline/input.json
+    mode=0644 owner={{ hadoop_common_user }} group={{ hadoop_common_group }}
+
+- name: luigi configuration directory created
+  file: >
+    path=/etc/luigi
```
feanil, line 60: Is this not configurable?  If it is, it should use a var and use `{{ COMMON_CFG_DIR }}` as the base for the location.

playbooks/roles/analytics_pipeline/tasks/main.yml
```diff
@@ -0,0 +1,128 @@
...
+  file: >
+    path={{ COMMON_LOG_DIR }}/tracking/tracking.log
+    mode=0644
+  ignore_errors: yes
+
+- name: cron job syncs tracking log file to hdfs
+  cron: >
+    user={{ hadoop_common_user }}
+    name="Sync tracking log to HDFS"
+    job="{{ HADOOP_COMMON_HOME }}/bin/hdfs dfs -put -f {{ COMMON_LOG_DIR }}/tracking/tracking.log {{ ANALYTICS_PIPELINE_HDFS_DATA_DIR }}/tracking.log"
```
feanil, line 128: No time settings on this.  Just making sure you want to run this every minute?

playbooks/roles/analytics_pipeline/tasks/main.yml
e0d, line 60: Hmm, unless this is the location specified in the deb package I would argue.

playbooks/roles/browsermob-proxy/defaults/main.yml
```diff
@@ -1,3 +1,3 @@
 # browsermob-proxy
-browsermob_proxy_version: '2.0-beta-9'
-browsermob_proxy_url: 'https://s3-us-west-1.amazonaws.com/lightbody-bmp/browsermob-proxy-{{ browsermob_proxy_version }}-bin.zip'
+browsermob_proxy_version: '2.0.0'
+browsermob_proxy_url: 'https://github.com/lightbody/browsermob-proxy/releases/download/browsermob-proxy-{{ browsermob_proxy_version }}/browsermob-proxy-{{ browsermob_proxy_version }}-bin.zip'
```
e0d, line 3: This change should be pulled into a separate PR.

playbooks/roles/edxapp/defaults/main.yml
e0d: I don't like this much.  While we have said we would make the defaults/main.yml settings target local environments, I'd like someone to have to explicitly decide to not use https.

playbooks/roles/hadoop_common/defaults/main.yml
```diff
@@ -0,0 +1,58 @@
...
+hadoop_common_debian_pkgs:
+  - llvm-gcc
+  - build-essential
+  - make
+  - cmake
+  - automake
+  - autoconf
+  - libtool
+  - zlib1g-dev
+  - maven
```
e0d, line 60: build-essential doesn't include make, cmake, automake, etc?

Do we really need maven and zlib1g-dev?

playbooks/roles/hadoop_common/defaults/main.yml
mulby, line 21: I'm not sure I can easily change this. Hadoop looks in `$HADOOP_HOME/etc/hadoop` for its configuration. Is it worth finding a way to make Hadoop look somewhere else?

vagrant/base/analyticstack/Vagrantfile
e0d, line 1: The meta question is still open I think, but I would not like to require different versions of Vagrant across projects.

playbooks/roles/insights/tasks/main.yml
mulby: Ah, cool, that makes sense. Side note: It is surprising to me that there are both `edxlocal` and `local_dev` roles.

vagrant/base/analyticstack/Vagrantfile
```diff
@@ -0,0 +1,142 @@
...
+      create: true, nfs: true
+    config.vm.synced_folder "#{forum_mount_dir}", "/edx/app/forum/cs_comments_service",
+      create: true, nfs: true
+
+    if ENV['ENABLE_LEGACY_ORA']
+        config.vm.synced_folder "#{ora_mount_dir}", "/edx/app/ora/ora",
+          create: true, nfs: true
+    end
+
+    config.vm.synced_folder "#{insights_mount_dir}", "/edx/app/insights/edx_analytics_dashboard",
```
e0d, line 77: Why are these services exempted from the conditional above?

playbooks/roles/sqoop/tasks/main.yml
mulby: Oh, cool - this is a holdover from an older version of these scripts, I'll remove throughout

vagrant/base/analyticstack/Vagrantfile
```diff
@@ -0,0 +1,142 @@
...
+
+VAGRANTFILE_API_VERSION = "2"
+
+MEMORY = 4096
+CPU_COUNT = 2
+
+edx_platform_mount_dir = "edx-platform"
+themes_mount_dir = "themes"
+forum_mount_dir = "cs_comments_service"
+ora_mount_dir = "ora"
```
e0d: This new file should not install ora1.

playbooks/roles/sqoop/tasks/main.yml
mulby: In other places where you download big packages, where do you store them? I could also use `/var/tmp`?

playbooks/roles/analytics_pipeline/tasks/main.yml
mulby, line 60: I believe this path is hardcoded in luigi - it would be rather difficult for us to change.

playbooks/vagrant-devstack.yml
mulby, line 32: The intent was that the image would be prepped to run insights and the API etc. Ports forwarded, NFS mounts etc, but it wouldn't ship with those services installed. The user would have to login to their devstack and run an ansible playbook in order to "install" the analytics stack extensions.

Open to better ideas about how to handle this.

playbooks/roles/edxapp/defaults/main.yml
mulby: Got it, is this the only place we break that norm? Either way I'll put a comment above it.

playbooks/roles/analytics_pipeline/defaults/main.yml
mulby, line 16: I don't think that's terribly important given that this role will only be used by devstacks.

playbooks/roles/insights/defaults/main.yml
mulby: done

playbooks/roles/hadoop_common/defaults/main.yml
feanil, line 21: No, if it's not something that is obviously in the hadoop config, I would just put a comment above it and leave it as is.

playbooks/roles/sqoop/defaults/main.yml
mulby: done

playbooks/roles/sqoop/tasks/main.yml
mulby: done

playbooks/roles/hadoop_common/defaults/main.yml
mulby: done

playbooks/roles/hadoop_common/defaults/main.yml
mulby: done

playbooks/roles/insights/tasks/main.yml
mulby: done

playbooks/roles/sqoop/tasks/main.yml
mulby: done

playbooks/roles/sqoop/tasks/main.yml
mulby: I remember trying to use this and having it not do what I expected it to do - replacing it with tar in a fit of rage.

I don't remember specifically what it was doing wrong, I can spend a little time figuring that out.

playbooks/roles/sqoop/tasks/main.yml
mulby: I moved everything to /var/tmp

playbooks/roles/sqoop/tasks/main.yml
mulby: done

playbooks/roles/sqoop/tasks/main.yml
mulby: done

playbooks/roles/analytics_pipeline/tasks/main.yml
mulby, line 128: Yeah that was the intent. I want you to be able to click around in the LMS, run the pipeline and see the updates in insights almost immediately afterwards.

playbooks/roles/browsermob-proxy/defaults/main.yml
mulby, line 3: Concur - I'll split this out

playbooks/roles/hadoop_common/defaults/main.yml
mulby, line 60: Not sure - I copied this list of packages from here:

https://github.com/mulby/docker-hadoop-single-node/blob/master/Dockerfile#L15

Since I based these roles on that Dockerfile

vagrant/base/analyticstack/Vagrantfile
mulby, line 77: That was a mistake, fixed.

vagrant/base/analyticstack/Vagrantfile
mulby: done

playbooks/roles/analytics_pipeline/tasks/main.yml
bradenmacdonald, line 60: You can change it by setting `LUIGI_CONFIG_PATH` as an OS environment variable for whatever user luigi runs as. If you do so, it has to point to a `.cfg` file, not a directory.

playbooks/roles/analytics_pipeline/tasks/main.yml
bradenmacdonald, line 128: Just wondering: couldn't you just make it so the pipeline task runner executes this sync command before running any pipeline task? That seems like it would be more efficient.

playbooks/roles/hadoop_common/tasks/main.yml
```diff
@@ -0,0 +1,194 @@
...
+# any role-specific actions such as formatting the namenode etc.
+#
+# Dependencies:
+#
+# oraclejdk: Not strictly required, but we tend to trust it more than openjdk.
+#
+
+- name: install system packages
+  apt: >
+    pkg={{ ','.join(hadoop_common_debian_pkgs) }}
```
e0d: This idiom should not longer be used, it was a work-around to install packages in a single transaction, but with_items should now do the right thing.

playbooks/roles/hadoop_common/tasks/main.yml
```diff
@@ -0,0 +1,194 @@
...
+
+- name: ensure user exists
+  user: >
+    name={{ hadoop_common_user }}
+    group={{ hadoop_common_group }}
+    home={{ HADOOP_COMMON_USER_HOME }} createhome=yes
+    shell=/bin/bash system=yes generate_ssh_key=yes
+    state=present
+
+- name: own key authorized
```
e0d, line 42: nit: you switch to the past tense here for names, why?

playbooks/roles/hadoop_common/templates/core-site.xml.j2
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0"?>
+<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
+
+<!-- Put site-specific property overrides in this file. -->
+
+<configuration>
+  <property>
+    <name>fs.default.name</name>
+    <value>hdfs://localhost:9000</value>
```
e0d, line 9: should this be a var?

playbooks/vagrant-analytics.yml
```diff
@@ -0,0 +1,34 @@
...
+    EDXAPP_OAUTH_ENFORCE_SECURE: false
+    EDXAPP_LMS_BASE_SCHEME: http
+  roles:
+    - edx_ansible
+    - edxlocal
+    - mongo
+    - edxapp
+    - oraclejdk
+    - elasticsearch
+    - forum
```
e0d, line 25: I question the value in have a lot of this on the analytics box, do you really need the notifier or browser packages?

vagrant/base/analyticstack/analytics_api
```diff
@@ -0,0 +1 @@
+Subproject commit d3a72344c0f1cc3dab05ca871bafa9a7a9effe1e
```
feanil: Is this a submodule?

vagrant/base/analyticstack/cs_comments_service
```diff
@@ -0,0 +1 @@
+Subproject commit 8d83377b7d93b69eebe2baebad57e9119f45f141
```
feanil: Looks like these repos got added as submodules accidentally?

playbooks/roles/analytics_pipeline/tasks/main.yml
mulby, line 60: Right, that would require changes to the way `remote-task` works (to ensure the environment variable is set properly when running launch-task), though, and I'm not sure it's worth the effort.

Also it would be unclear to people running `launch-task` directly on the VM that they need to set that environment variable. I suppose we could push it into a `.bashrc` file, but that would be a bigger change.

playbooks/roles/analytics_pipeline/tasks/main.yml
mulby, line 128: A good idea... I'm just not sure where to inject it in the process, particularly since I don't want to run this on production workflows.

We could put it into a wrapper script that we write out to the VM... but `remote-task` wouldn't run that wrapper script so you would get inconsistent behavior when running the pipeline from your local machine vs. inside the VM.

I could pull this command out into a separate script and just add another step to running the pipeline?

``` bash
# you've clicked around insights, now you want to run the pipeline
sudo su hadoop
/edx/bin/update-pipeline-tracking-log
# running it locally on the new data
launch-task ...
# alternately you could run it remotely
exit
exit
remote-task ...
```

Open to suggestions here...

playbooks/roles/hadoop_common/tasks/main.yml
mulby: done

playbooks/roles/hadoop_common/tasks/main.yml
mulby, line 42: I try to name them such that they describe the end state instead of describing the transformation. I think helps emphasize the fact that they should be idempotent.

I did this a while ago and am now realizing that it may not be consistent with the rest of your naming in this repo... but I'm thinking it's not worth changing - let me know if you disagree.

playbooks/roles/hadoop_common/templates/core-site.xml.j2
mulby, line 9: maybe?

This setting doesn't actually control the interface and port that the HDFS deamon will bind to, though, it simply tells any HDFS clients where to find the name node. This might lead to unexpected behavior if someone changed the setting in the ansible vars they would expect everything to "just work", however, they would really just be breaking the clients without changing the server setting.

I could also research the server setting if necessary, but I don't know it off hand.

playbooks/vagrant-analytics.yml
mulby, line 25: Yeah, I agree - I wasn't sure what was essential and what wasn't... so I just included everything. Guidance here would be appreciated! I don't know what some of these things are.

vagrant/base/analyticstack/ansible.cfg
```diff
@@ -0,0 +1,10 @@
+# config file for ansible -- http://ansible.github.com
```
mulby: this should be a symlink

vagrant/release/analyticstack/Vagrantfile
```diff
@@ -0,0 +1,91 @@
...
+  ora_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + ora_mount_dir
+  insights_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + insights_mount_dir
+  analytics_api_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + analytics_api_mount_dir
+
+end
+
+Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
+
+  # Creates an edX devstack VM from an official release
+  config.vm.box     = "analyticstack-beta"
```
mulby, line 32: include the box_url here once we upload it to files.edx.org

vagrant/base/analyticstack/ansible.cfg
mulby: done

playbooks/roles/browsermob-proxy/defaults/main.yml
mulby, line 3: done

mulby:
@e0d and @feanil - I think I have addressed or responded to all outstanding comments on this PR

playbooks/roles/hadoop_common/defaults/main.yml
feanil, line 60: @mulby did you get a chance to check the EMR boxes to see if we really needed `maven`?

mulby, line 60: Yes it is installed:

```
$ dpkg --get-selections | grep -v deinstall | grep maven
libmaven-archiver-java              install
libmaven-clean-plugin-java          install
libmaven-compiler-plugin-java           install
libmaven-dependency-tree-java           install
libmaven-file-management-java           install
libmaven-filtering-java             install
libmaven-install-plugin-java            install
libmaven-jar-plugin-java            install
libmaven-plugin-tools-java          install
libmaven-reporting-impl-java            install
libmaven-resources-plugin-java          install
libmaven-scm-java               install
libmaven-shade-plugin-java          install
libmaven-shared-io-java             install
libmaven2-core-java             install
maven2                      install
```

playbooks/vagrant-analytics.yml
feanil, line 25: If forums is going to be removed, you'd also have to update the features dictionary of the edxapp which enables it.  I think this is fine as is for simplicity.

feanil:
:+1:

playbooks/vagrant-analytics.yml
mulby, line 25: I wanted to leave forums since there are some events generated by them which are used by some reports.

playbooks/roles/hadoop_common/defaults/main.yml
bradenmacdonald: @mulby This change has broken the remote-task launcher, which is hard coded to look for a .bashrc file in `/home/{sudo_user}/`:

```
Running command = ['ssh', '-tt', '-o', 'ForwardAgent=yes', '-o', 'StrictHostKeyChecking=no', '-o', 'UserKnownHostsFile=/dev/null', '-o', 'KbdInteractiveAuthentication=no', '-o', 'PasswordAuthentication=no', '-o', 'User=vagrant', '-o', 'ConnectTimeout=10', '-p', '2222', '-i', '/Users/braden/.vagrant.d/boxes/cypress-devstack-rc4/0/virtualbox/vagrant_private_key', '127.0.0.1', "sudo -u hadoop /bin/bash -c 'cd /var/lib/analytics-tasks/devstack/repo && . /home/hadoop/.bashrc && /var/lib/analytics-tasks/devstack/venv/bin/launch-task --n-reduce-tasks 1 --local-scheduler InsertToMysqlUserProfilesTask'"]
Warning: Permanently added '[127.0.0.1]:2222' (RSA) to the list of known hosts.
/bin/bash: /home/hadoop/.bashrc: No such file or directory
Connection to 127.0.0.1 closed.
Exiting with status = 1
```

vagrant/release/devstack/Vagrantfile
```diff
@@ -122,6 +126,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.network :forwarded_port, guest: 18080, host: 18080
   config.vm.network :forwarded_port, guest: 8765, host: 8765
   config.vm.network :forwarded_port, guest: 9200, host: 9200
+  config.vm.network :forwarded_port, guest: 8100, host: 8100
+  config.vm.network :forwarded_port, guest: 8110, host: 8110
```
bradenmacdonald: @mulby Can you add port `50070` too? It's the web interface to HDFS, which includes a web-based HDFS file browser that is useful when debugging the pipeline.

bradenmacdonald:
@mulby Here's a request: I would love if you could tweak the config so analytics devstack users can use `hive` to run queries from the cli, which is extremely useful for developing pipeline tasks.

To get this working you just need to add (via template) a `/edx/app/hadooop/.hiverc` file like:

```
set hive.metastore.warehouse.dir=hdfs://localhost:9000/edx-analytics-pipeline/warehouse;
set javax.jdo.option.ConnectionURL=jdbc:derby:/var/lib/analytics-tasks/devstack/repo/metastore_db\;create=true;
```

With that in place, one can just do `vagrant ssh`, `sudo su hadoop`, `hive`, and then use the hive REPL to run queries on the hive tables. You may also want to [configure devstack hive to use MySQL for its metastore](https://github.com/open-craft/edx-analytics-devstack/commit/1ecd39cb6296f57cb57a81544571331e52a9f0ee) which should be more robust and have better performance compared to derby.

vagrant/release/devstack/Vagrantfile
mulby: Good idea, done.

mulby:
@bradenmacdonald - I considered pulling in your changes to use MySQL as the metastore, however, this is fairly different from the way the pipeline works in production. In production the workflows run on different clusters so luigi has to recreate the Hive metadata on each one, we haven't yet setup a centralized Hive metastore (there has been no need).

The more I think about this, though, I suppose if a developer keeps reusing the `devstack` version of the pipeline inside the VM they won't be getting a clean Hive metastore anyway, since Hive will just reuse the derby store from last time.

I'll pull in your changes.

mulby:
@bradenmacdonald I moved the metastore into mysql, I didn't include the warehouse setting since that is not set in prod and would result in rather different behavior locally.

I also opted to use hive-site.xml to store the settings, was there a reason you put them in .hiverc instead?

bradenmacdonald:
@mulby The main thing I care about is that if I ssh in to the devstack and sudo to the `hadoop` user, I want to be able to type `hive` and then run queries against the pipeline data in the Hive REPL shell (without having to specify any configuration settings / variables).

I wasn't able to get that to work with just hive-site.xml changes; I had to add some additional settings to .hiverc. But you might have figured out something I missed.

mulby:
@e0d - I'm just waiting on a thumb from you at this point.

Also we will have to sync up this PR with some secure repo changes, so we should probably make those changes before merging this.

mulby:
I will also squash and rebase before merging.

mulby:
Updated settings that have to be updated in secure configs:

``` yaml
INSIGHTS_SOCIAL_AUTH_REDIRECT_IS_HTTPS: true
INSIGHTS_SESSION_COOKIE_NAME: sessionid
INSIGHTS_CSRF_COOKIE_NAME: crsftoken
INSIGHTS_LANGUAGE_COOKIE_NAME: language
```

bradenmacdonald:
Is it possible something is still not right with the creation of users like `analytics_api` and `insights`? Because when I updated and re-ran this playbook, I was no longer able to `sudo` to those users without first doing e.g. `sudo usermod insights -s /bin/bash`

bradenmacdonald:
@mulby The Insights dashboard does not work out of the box, and the documentation doesn't mention how to get it working, unless I'm missing something. I had to create an OAuth2 client at http://localhost:8000/admin/oauth2/client/ and then edit `/edx/etc/insights.yml` (which I know should be edited via ansible) and set `SOCIAL_AUTH_EDX_OIDC_KEY`, `SOCIAL_AUTH_EDX_OIDC_SECRET` and `SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY`. I also added `DEBUG: true` while I was at it since I think everything should run in debug mode by default on a devstack. It would be really nice if that could all be done automatically on the devstack during provisioning. I know as it is my changes will be lost if I update the config... how am I supposed to change settings like that?

Also, another wishlist item: The data-api also isn't super useful out of the box. I could connect to swagger at http://localhost:8100/docs/ but attempting to run any commands gave a 401 error, and if any error occurred it just gave a generic 500 instead of showing the specific message. I got it to work by adding `DEBUG: true` to the `ANALYTICS_API_CFG` file (`/edx/etc/analytics_api.yml`) (and running `analytics_api/manage.py set_api_key edx edx` but I'm not sure if that matters)

mulby:
@bradenmacdonald it sounds like it somehow didn't run the most recent version of this PR, I fixed a few of the issues you described above at various points.

I have been running:

``` bash
cd vagrant/base/analyticstack
CONFIGURATION_VERSION=origin/gabe/analytics-devstack EDX_PLATFORM_VERSION=master vagrant up
```

It takes a while but I'm seeing Insights work "out of the box" as long as you access it via "http://127.0.0.1:8110",  using "localhost" causes oauth issues.

I definitely agree about the DEBUG issue. The suggested path forward there is to create new settings for the API and dashboard that introduce "devstack" settings that simply inherit everything from the production settings but set DEBUG to `True`.

mulby:
@bradenmacdonald also note that the latest version of the doc recommends using the release Vagrantfile, which isn't ready yet since it doesn't specify the correct box URLs etc.

playbooks/roles/hadoop_common/tasks/main.yml
e0d, line 42: Not enough to slow this down, but I think this differs from what's done everywhere else.

e0d:
:+1:

PR #2178, opened 2015-07-30
Allow configuring third party auth backends

tusbar:
Those are already used in [`aws.py`](https://github.com/edx/edx-platform/blob/35503ebda2ef0a34a9c53379e45e5f2b1c24d186/lms/envs/aws.py#L550) in edx-platform.


tusbar:
I’ve added this in the shared section and not just for the LMS as I wish that someday the third party auth authentication will be available for both LMS and Studio. :)

Also, in my fork, I’ve added it for Studio.

PR #2180, opened 2015-07-31
Clean rebuild of a sandbox base image without distribute

jibsheet:
Without this, sandbox builds fail when edx-platform updates because the
virtualenv contains distribute and we try to install setuptools.


jibsheet:
@e0d @maxrothman quick look so I can unbreak sandboxes (waiting for ansible-provision to finish)

maxrothman:
:+1:

maxrothman:
:+1:

PR #2183, opened 2015-08-03
Remove the 'm' from all DNS names and examples.

feanil:


jibsheet:
:+1:  
This reminds me that the connect to prod data, cname sandbox and ssh keys on sandbox all likely expose m.sandbox.edx.org in some way.

PR #2186, opened 2015-08-04
Update the course ID to match the split course_id

feanil:
We updated code elsewhere so that by default courses that are
imported are imported into the split modules store but that also
changes the course_id of this course and breaks the enrolling of
test students into the demo course.  This change should resolve
the issue.

@edx/devops 

FYI: @chrisndodge


jibsheet:
:+1:

PR #2188, opened 2015-08-06
Bad push

e0d:
Hmm, magit somehow had my local branch tracking named-release/cypress.rc


e0d:
@singingwolfboy @feanil FYI

PR #2193, opened 2015-08-06
OPS-748+OPS752: bring up instances without ASGs and manage disk layouts

fredsmith:
Related to https://github.com/edx-ops/edx-secure/pull/538

### DIFF
--- a/playbooks/edx-east/edx_service.yml
+++ b/playbooks/edx-east/edx_service.yml
@@ -225,19 +225,18 @@
       with_items: metric_alarms 
       when: auto_scaling_service
 
-    - name: See if instances already exist
-      local_action:
-        module: "ec2_lookup"
-        region: "{{ aws_region }}"
-        tags: "{{ asg_instance_tags }}"
-      register: potential_existing_instances
-
-    - name: Transform tags to the format ec2 module expects
+    - name: Transform tags into dict format for the modules that expect it
       util_map:
         function: zip_to_dict
         input: "{{ asg_instance_tags }}"
         args: ['key', 'value']
       register: reformatted_asg_instance_tags
+
+    - name: See if instances already exist
+      ec2_lookup:
+        region: "{{ aws_region }}"
+        tags: "{{ reformatted_asg_instance_tags.function_output }}"
+      register: potential_existing_instances
 
     #This task will create the number of instances requested (create_instances parameter). 
     # By default, it will create instances equaling the number of subnets specified.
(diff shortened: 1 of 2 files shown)


playbooks/edx-east/edx_service.yml
```diff
@@ -240,12 +249,70 @@
...
+  tasks:
+    #Wait in this play so it can multiplex across all launched hosts
+    - name: Wait for hosts to be ready
+      become: False
+      local_action:
+        module: wait_for
+        host: "{{ inventory_hostname }}"
+        port: 22
+
+    - name: Gather facts
```
feanil: Why gather facts here when you disable it on line 277?

playbooks/edx-east/edx_service.yml
```diff
@@ -240,12 +249,70 @@
...
+
+    - name: Unmount all specified disks that are currently mounted
+      mount:
+        name: "{{ item[0].mount }}"
+        src: "{{ item[0].device }}"
+        fstype: "{{ item[0].fstype }}"
+        state: absent
+      when: item[1].device_name == item[0].device
+      with_nested:
+        - ansible_mounts
```
feanil, line 300: Where is this var defined?

playbooks/edx-east/edx_service.yml
```diff
@@ -240,12 +249,70 @@
...
+        - volumes
+
+    - name: Create filesystems
+      filesystem:
+        dev: "{{ item.device_name }}"
+        fstype: ext4
+        #Since we don't do this task if any prior instances were found in the ec2_lookup task, it's safe to force
+        force: yes
```
feanil: Why is force required?

playbooks/edx-east/edx_service.yml
fredsmith, line 300: ansible_mounts is defined by ansible.

playbooks/edx-east/edx_service.yml
maxrothman: We can't gather facts until the instance is reachable, so we have to `wait_for` it first. The alternative is to `wait_for` each instance iteratively in the previous play, which would block on any particular instance being unreachable. I can add a comment to that effect.

playbooks/edx-east/edx_service.yml
maxrothman: If the disk came with a filesystem already, we want to overwrite it. The first ephemeral volume (/dev/xvdb) on boot is ext2 or 3 and is mounted at /mnt.

playbooks/edx-east/edx_service.yml
feanil: I think a comment would be a good idea.  Because setup takes no parameters it looks a bit weird.

playbooks/edx-east/edx_service.yml
feanil: cool, makes sense, can you add your note as a comment in this file as well.

feanil:
A couple more comments need to be added but then :+1:

feanil:
:clap: :shipit:

PR #2194, opened 2015-08-06
Change when we tag instances (Cypress)

singingwolfboy:
#2175 for Cypress. This is to address the issue reported here: https://groups.google.com/forum/#!msg/openedx-ops/GTaZi6tAVOM/CCF_OSpxbsEJ


e0d:
:+1: 

@feanil FYI back-porting for reported instance of the root problem.

PR #2198, opened 2015-08-07
We don't have this template

jibsheet:
It never got checked in and honestly we don't use it so it isn't worth
recreating.


jibsheet:
@edx/devops

maxrothman:
:+1:

feanil:
:+1:

PR #2199, opened 2015-08-08
E0d/devstack nightly

e0d:
Changes to support nightly builds of DevStack

### DIFF
--- a/vagrant/base/devstack/Vagrantfile
+++ b/vagrant/base/devstack/Vagrantfile
@@ -59,6 +59,19 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.ssh.forward_x11 = true
   end
 
+  # 
+  # By default no shared filesystem is configured between the guest and host.
+  # To enable filesystem sharing, which makes editing running code posible,
+  # you will need to enable either NFS of VBOXFS.
+  #
+  # To enable NFS:
+  #
+  # $ export VAGRANT_USE_NFS="true"
+  #
+  # To enable VBOXFS, which is typically less performant:
+  #
+  # $ export VAGRANT_USE_VBOXFS="true"
+  #
   if ENV['VAGRANT_USE_VBOXFS'] == 'true'
     MOUNT_DIRS.each { |k, v|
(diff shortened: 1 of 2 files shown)


e0d:
@singingwolfboy @feanil 

Opening to spur discussion.  I'd like to move this forward so that nightly builds of devstack are an option for edx developers.

vagrant/base/devstack/Vagrantfile
```diff
@@ -49,6 +49,19 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.ssh.forward_x11 = true
   end
 
+  # 
+  # By default no shared filesystem is configured between the guest and host.
+  # To enable filesystem sharing, which makes editing running code posible,
+  # you will need to enable either NFS of VBOXFS.
```
singingwolfboy, line 65: Why is this change necessary? It's going to need a lot of documentation and announcements.

feanil, line 65: This is only in the base devstack not on the release version that developers use.  I assume this is to make it easier to build these boxes @e0d ?

feanil:
:+1:

vagrant/base/devstack/Vagrantfile
carsongee, line 65: Why not just add a third option, like `VAGRANT_NO_SHARE` so that this isn't backwards incompatible?

e0d, line 65: I considered doing that way, this felt cleaner, but I'm happy to go the other way if that's majority opinion.  I'm not sure who all the customers of the base scripts are.  But not need to allow jenkins to sudo to mount NFS shares during builds seemed preferable to me.

This shouldn't prevent the release from automatically creating the mounts.

carsongee, line 65: I suppose I'm probably the primary user then, since I develop for both config and edx-platform dev it is convenient to use the one box. If I'm the only one, feel free to keep it as written, I can manage my environment variables.

e0d, line 65: I'm happy to fence it off in that case.

PR #2200, opened 2015-08-10
Remove old code for jenkins whitelisting

jzoldak:
@benpatterson I think we don't have any jobs that use the old whitelist any longer, right?


benpatterson:
@jzoldak good call. 

:+1:

PR #2201, opened 2015-08-10
new xsy role

fredsmith:


playbooks/edx-east/xsy.yml
```diff
@@ -0,0 +1,10 @@
+# Configure an admin instance with jenkins and asgard.
+- name: Configure instance(s)
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    serial_count: 1
```
maxrothman: Why serial_count?

playbooks/roles/xsy/defaults/main.yml
```diff
@@ -0,0 +1,65 @@
...
+XSY_V2_TOKEN: 'HIPCHAT_V2_TOKEN'
+XSY_ROOMS: 'Hammer'
+XSY_NAME: 'xsy'
+XSY_HANDLE: 'xsy'
+XSY_REDIS_URL: 'redis://fakeuser:redispassword@localhost:6379'
+XSY_HTTPSERVER_PORT: '8081'
+XSY_WORLD_WEATHER_KEY: !!null
+
+
+xsy_role_name: xsy
```
maxrothman: I think we swtiched to `service_name`.

playbooks/roles/xsy/handlers/main.yml
```diff
@@ -0,0 +1,23 @@
...
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Handlers for role xsy
+# 
+# Overview:
+# 
+#
```
maxrothman, line 16: I think it'd be better to just stick a restart at the end of tasks/main.yml. The when doesn't make sense either.

playbooks/roles/xsy/tasks/deploy.yml
```diff
@@ -0,0 +1,67 @@
+- name: setup the xsy env
+  template: >
+    src="xsy_env.j2" dest="{{ xsy_app_dir }}/xsy_env"
```
maxrothman: Nitpick: I think we're trying to switch to pure-YAML module invocations.

playbooks/roles/xsy/tasks/deploy.yml
```diff
@@ -0,0 +1,67 @@
...
+  sudo_user: "{{ xsy_user }}"
+  register: xsy_checkout
+  notify: restart xsy
+
+- name: install the requirements
+  pip: >
+    requirements="{{ xsy_requirements_file }}"
+    virtualenv="{{ xsy_venv_dir }}"
+    state=present
+    extra_args="-i {{ COMMON_PYPI_MIRROR_URL }}"
```
maxrothman: I think this doesn't work anymore with the most recent version of pip. Do we still add this arg for our other services?

maxrothman:
Since xsy isn't something we support for open-source users, it might be better to have this live in one of our other repos.

playbooks/roles/xsy/tasks/deploy.yml
```diff
@@ -0,0 +1,67 @@
+- name: setup the xsy env
```
maxrothman: In general, I think we're trying to get away from main.yml vs deploy.yml. One big file is fine.

playbooks/roles/xsy/templates/xsy_env.j2
```diff
@@ -0,0 +1,8 @@
+# {{ ansible_managed }}
+
+{% for name,value in xsy_environment.items() -%}
+{%- if value -%}
```
maxrothman, line 4: Is this a good idea? Then anything falsy will be excluded.

PR #2204, opened 2015-08-11
Removing xqwatcher as there's no opensource course exemplar

e0d:
@singingwolfboy @sarina @jbarciauskas 

OK, looking back at this, I think that removing xqwatcher is the right thing to do.  The issue is that we still don't have an open course that can be included with the xqwatcher.  I think that we could create one from prior-art without much difficulty.  But, for now, we should remove the watcher from full stack.

### DIFF
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -45,7 +45,6 @@
     - forum
     - { role: notifier, NOTIFIER_DIGEST_TASK_INTERVAL: "5" }
     - { role: "xqueue", update_users: True }
-    - xqwatcher
     - role: ora
       when: ENABLE_LEGACY_ORA
     - certs


sarina:
This makes sense to me. What's the right way to communicate this, for users who may be expecting xqwatcher -- they need to know they'll need to explicitly enable it.

feanil:
This also needs to be removed from the vagrant-devstack.yml and vagrant-fullstack.yml playbooks.

feanil:
@sarina I think making a note in the announcement should be sufficient, as Ed said, it is not likely that anyone is using it.

e0d:
@feanil xqwatcher was not included in the devstack and fullstack plays.  So, there's a lack of fidelity between the sandbox play the fullstack currently.

feanil:
In that case :+1:

carsongee:
:+1: though I would really like to get an open source grader out there in the future.

singingwolfboy:
:+1:

e0d:
@carsongee completely agree.

PR #2206, opened 2015-08-11
Remove insights from Cypress

singingwolfboy:
@feanil can you review this? We want to remove insights from the Cypress branch, but I assume that we want to leave it in this yaml file in master.


feanil:
:+1:

PR #2209, opened 2015-08-12
mongo 3.0 role

fredsmith:
This is a replacement for #2141 - We're making it a separate role so we don't need two mongo instances on devstacks until forums supports mongo 3.0.


maxrothman:
:+1: on my change

PR #2214, opened 2015-08-13
Add mongostat logging

maxrothman:


playbooks/roles/mongo/tasks/main.yml
```diff
@@ -160,3 +160,10 @@
     minute="0"
     day="{{ MONGO_S3_BACKUP_DAY }}"
   when: MONGO_S3_BACKUP
+
+- name: add mongodump logging command to cron
+  cron:
+    name: "mongodump"
+    job: >
+      /usr/bin/mongostat -u {{ MONGO_ADMIN_USER }} --authenticationDatabase admin
```
feanil: Let's put this in a shellscript that is actually run so that you can't see the username and password via top.

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -160,3 +160,10 @@
     minute="0"
     day="{{ MONGO_S3_BACKUP_DAY }}"
   when: MONGO_S3_BACKUP
+
+- name: add mongodump logging command to cron
+  cron:
+    name: "mongodump"
+    job: >
```
fredsmith: the default is to run every minute - is that what we want?

feanil: @fredsmith that is the default and I think that is what we want.

fredsmith: also, to run as root.

playbooks/roles/mongo/tasks/main.yml
maxrothman: Is sitting in a plain-text file significantly better than being visible in top?

EDIT: what I'd really prefer is some other sort of authentication, like Postgres' IDENT.

feanil:
You say `mongodump` in a bunch of places where I think you mean `mongostat`.  Once that is fixed, :+1:

playbooks/roles/mongo/tasks/main.yml
```diff
@@ -160,3 +160,16 @@
...
+    src="edx/bin/log-mongostat.sh.j2"
+    dest="{{ COMMON_BIN_DIR }}/log-mongostat.sh"
+    owner="{{ mongo_user }}"
+    group="{{ mongo_user }}"
+    mode=0700
+
+- name: add mongodump logging script to cron
+  cron:
+    name: "mongodump"
+    job: /edx/bin/log-mongostat.sh >{{ mongo_log_dir }}/mongostat.log 2>&1
```
feanil: you should become root to run this so it doesn't get put into the ubuntu user's crontab.

feanil:
:+1:

carsongee:
Is this rotated? Can we maybe make this optional?

maxrothman:
It is not currently rotated, but there's a [bug in for that](https://openedx.atlassian.net/browse/OPS-954). I'll make it optional.

maxrothman:
@carsongee I've made it optional. Looks good?

carsongee:
It might be nice to have it off by default so we don't fill up devstack/fullstack boxes hard drives, but I'm ok with it as is. Thanks for making the adaptation for me.

maxrothman:
My impression is that we tend to err on the side of **more** logging on devstack/fullstacks. Thanks for the feedback!

PR #2216, opened 2015-08-14
Added a dummy date in the date_joined

vkaracic:
It was causing 500 internal server errors because after
sanitization it would be set to null but a valid value would be
needed for a couple of actions on the website, like assigning
students to programs and assigning students to courses.

### DIFF
--- a/util/vpc-tools/sanitize-db-wwc.sql
+++ b/util/vpc-tools/sanitize-db-wwc.sql
@@ -24,7 +24,7 @@ UPDATE auth_user
         first_name = concat('user-',cast(id AS CHAR)),
         last_name = concat('user-',cast(id AS CHAR)),
         last_login = null,
-        date_joined = null
+        date_joined = "2010-01-01 00:00:00"
             where email not like ('%@edx.org');
 
 /*
--- a/util/vpc-tools/sanitize-db-wwc.sql
+++ b/util/vpc-tools/sanitize-db-wwc.sql
@@ -24,7 +24,7 @@ UPDATE auth_user
         first_name = concat('user-',cast(id AS CHAR)),
         last_name = concat('user-',cast(id AS CHAR)),
         last_login = null,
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @vkaracic! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request. 

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=2216

vkaracic:
It could be even better, as @mattdrayer suggested, to keep the original data and that way preserve real data as much as possible. Not sure if date_joined is a sensitive info that we want to sanitize. 
@fredsmith thoughts?

fredsmith:
@jibsheet @maxrothman -

what are your thoughts on this from a data sanitization perspective.   The solutions fork API throws an error when date_joined is Null.

vkaracic:
Details described here: https://edx-wiki.atlassian.net/browse/MCKIN-3444

util/vpc-tools/sanitize-db-wwc.sql
```diff
@@ -24,7 +24,7 @@ UPDATE auth_user
         first_name = concat('user-',cast(id AS CHAR)),
         last_name = concat('user-',cast(id AS CHAR)),
         last_login = null,
-        date_joined = null
+        date_joined = "2010-01-01 00:00:00"
```
fredsmith: can you make this "1970-01-01 00:00:00" instead?

jibsheet:
:+1:

fredsmith:
:+1:

PR #2221, opened 2015-08-17
Rotate mongo logs

maxrothman:
Explanation: if mongo is set with systemLog.logRotate='reopen', when it
receives a SIGURS1, it'll close and reopen its log file handle.


feanil:
:+1:   once testing has concluded.

PR #2222, opened 2015-08-17
Zoldak/jenkins sitespeed

benpatterson:
@jzoldak 

There is some repeated config/code in the meta/main.yml and tasks/main.yml. @fredsmith I don't suppose you have any suggestions on making that cleaner? Thanks for the look.


jzoldak:
@benpatterson Next up would be creating other kinds of lighter-weight, non-platform jenkins workers (e.g. for running the deduper, for being flow workers, etc.). Rather than refactor again, could we have a positive flag for platform-worker rather than using "when: not sitespeed..."?

Another question - why does the sitespeed worker need the OAuth key?

benpatterson:
@jzoldak for the OAuth key, I'm assuming that we will be doing github status updates from the sitespeed worker? 

In terms of lighter-weight workers, what do you imagine should be on those machines? Like, would just the common role be enough? For example, do we need the requests library? (Or perhaps that's handled in pip installs on those workers, etc).

jzoldak:
@benpatterson I guess I just really want to get rid of that wacky OAuth key in the script pattern that we have so I didn't want to proliferate it unless necessary, but I can also be convinced that this topic is out of scope as it's represented in TE-1008.

For the lighter weight workers, I'd have to look at the code again to see exactly what common does. But yeah the idea is something generic that has just enough to run the deduper or be a flow worker, etc. If it has python on it though I think it would be fine to have the requests library by default, that's not a lot of overhead and will most likely we would find it useful.

benpatterson:
@jzoldak ready for another round.

End-result is we have 3 packer files:
- jenkins_worker
- jenkins_worker_sitespeed
- jenkins_worker_simple

playbooks/edx-east/jenkins_worker_sitespeedio.yml
```diff
@@ -0,0 +1,21 @@
...
+  hosts: jenkins_worker
+  sudo: True
+  gather_facts: True
+  vars:
+    mongo_enable_journal: False
+    serial_count: 1
+    sitespeed_worker: True
+  serial: "{{ serial_count }}"
+  vars_files:
+    - roles/edxapp/defaults/main.yml
```
jzoldak, line 15: Does the sitespeed worker need these?

playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -38,6 +38,7 @@
   template:
     src=jenkins_env.j2 dest={{ jenkins_home }}/jenkins_env
     owner={{ jenkins_user }} group={{ jenkins_group }} mode=0500
+  when: not sitespeed_worker
```
jzoldak: Can this be when: platform_worker instead?

playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -62,8 +63,10 @@
 - name: Create directory for pip-accel config file
   file: path={{ jenkins_home }}/.pip-accel state=directory
     owner={{ jenkins_user }} group={{ jenkins_group }} mode=0777 recurse=yes
+  when: not sitespeed_worker
```
jzoldak: Can this be when: platform_worker instead?

playbooks/roles/jenkins_worker/tasks/system.yml
```diff
@@ -62,8 +63,10 @@
...
 
 - name: Create pip-accel config file
   template:
     src=pip-accel.conf.j2 dest={{ jenkins_home }}/.pip-accel/pip-accel.conf
     owner={{ jenkins_user }} group={{ jenkins_group }} mode=0664
+  when: not sitespeed_worker
```
jzoldak: Same here?

playbooks/roles/jenkins_worker/tasks/main.yml
```diff
@@ -1,16 +1,25 @@
...
+  when: platform_worker is defined
+- include: python_platform_worker.yml
+  when: platform_worker is defined
+
+# Run appropriate tests
 - include: test.yml
+- include: test_platform_worker.yml
+  when: platform_worker is defined
+- include: test_sitespeed_worker.yml
+  when: sitespeed_worker is defined
```
jzoldak, line 25: Should we have tests for the simple worker?

benpatterson, line 25: Agreed. That is handled with the plain-old `test.yml`.

playbooks/roles/jenkins_worker/tasks/system.yml
benpatterson: done

playbooks/roles/jenkins_worker/tasks/system.yml
benpatterson: done

playbooks/roles/jenkins_worker/tasks/system.yml
benpatterson: done

playbooks/edx-east/jenkins_worker_sitespeedio.yml
benpatterson, line 15: No, but it might be too noisy to have a bunch of when statements (one per file), so I left it as-is. They are just vars that will go unused by the sitespeed worker.

benpatterson:
@jzoldak I updated and answered feedback. Let me know what you think at this point.

playbooks/roles/jenkins_worker/tasks/main.yml
jzoldak, line 25: oic. cool!

jzoldak:
:+1:

PR #2223, opened 2015-08-17
Max/wrong mongo

maxrothman:
Will re-add these changes later, but felt it was important to unbreak the old mongo stuff first

@edx/devops


feanil:
:+1:  once tests pass.

PR #2227, opened 2015-08-19
Otto in Sandboxes

clintonb:
TODO
- [x] Ecom service starts, and OAuth setup properly
- [x] Set correct Nginx ports
- [ ] Verify devstack not broken
- [x] Move OAuth setup to either its own role or a new role for all local, post-edxapp, tasks
- ~~Migrate courses~~ (de-prioritized by @clintonb)

ECOM-2185

FYI @feanil @rlucioni @jimabramson @wedaly

### DIFF
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -39,6 +39,7 @@
     - insights
     - edx_notes_api
     - demo
+    - oauth_client_setup
     - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
     - oraclejdk
     - elasticsearch
--- a/playbooks/roles/launch_ec2/tasks/main.yml
+++ b/playbooks/roles/launch_ec2/tasks/main.yml
@@ -66,7 +66,7 @@
     value: "{{ item.public_dns_name }}"
   with_items: ec2.instances
 
-- name: Add DNS name studio
+- name: Add DNS names for services
   local_action:
     module: route53
     overwrite: yes
@@ -74,21 +74,11 @@
     zone: "{{ dns_zone }}"
     type: CNAME
     ttl: 300
-    record: "studio-{{ dns_name }}.{{ dns_zone }}"
-    value: "{{ item.public_dns_name }}"
-  with_items: ec2.instances
-
-- name: Add DNS name preview
-  local_action:
-    module: route53
-    overwrite: yes
-    command: create
-    zone: "{{ dns_zone }}"
-    type: CNAME
-    ttl: 300
-    record: "preview-{{ dns_name }}.{{ dns_zone }}"
-    value: "{{ item.public_dns_name }}"
-  with_items: ec2.instances
+    record: "{{ item[1] }}-{{ dns_name }}.{{ dns_zone }}"
+    value: "{{ item[0].public_dns_name }}"
+  with_nested:
+    - ec2.instances
+    - ['studio', 'ecommerce', 'preview']
 
 
 - name: Add new instance to host group
--- a/playbooks/edx-east/ecommerce.yml
+++ b/playbooks/edx-east/ecommerce.yml
@@ -10,6 +10,8 @@
     - role: nginx
       nginx_sites:
         - ecommerce
+      nginx_default_sites:
+        - ecommerce
     - aws
     - ecommerce
     - role: datadog
--- a/playbooks/edx-east/edx_continuous_integration.yml
+++ b/playbooks/edx-east/edx_continuous_integration.yml
@@ -16,6 +16,7 @@
       - xserver
       - certs
       - analytics_api
+      - ecommerce
       nginx_default_sites:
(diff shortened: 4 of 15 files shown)


clintonb:
FYI @fredsmith @jibsheet

clintonb:
@feanil any feedback here?

playbooks/roles/demo/defaults/main.yml
```diff
@@ -33,3 +33,7 @@ demo_edxapp_user: 'edxapp'
 demo_edxapp_venv_bin: '{{ COMMON_APP_DIR }}/{{ demo_edxapp_user }}/venvs/{{demo_edxapp_user}}/bin'
 demo_edxapp_course_data_dir: '{{ COMMON_DATA_DIR }}/{{ demo_edxapp_user }}/data'
 demo_edxapp_code_dir: '{{ COMMON_APP_DIR }}/{{ demo_edxapp_user }}/edx-platform'
+
```
feanil: Why is this in the demo role?   It should be in the local_dev role.

playbooks/roles/demo/tasks/main.yml
```diff
@@ -36,3 +36,18 @@
     owner="{{ demo_edxapp_user }}" group="{{ common_web_group }}"
 
 - include: deploy.yml tags=deploy
+
+- name: create OAuth2 Clients
```
feanil: same here.  Seems to be in the wrong role.

playbooks/roles/launch_ec2/tasks/main.yml
```diff
@@ -78,6 +78,18 @@
     value: "{{ item.public_dns_name }}"
   with_items: ec2.instances
 
+- name: Add DNS name ecommerce
```
feanil: @arbabnazar this DNS name should also be added to the list of removed DNS names for when sandboxes are cleaned up.

arbabnazar: @feanil current script will also cover it as it is looking for all the dns records that are associated with the sandbox public dns.

feanil: @arbabnazar A sandbox can have more than one public DNS name.  this adds a new one ecommerce-<dns_name>...  So we need to make sure we remove it.

playbooks/roles/launch_ec2/tasks/main.yml
```diff
@@ -78,6 +78,18 @@
...
+  local_action:
```
feanil: @clintonb this should be done as a 'with_items list and consolidate the tasks for creating DNS names into a single task.

playbooks/roles/launch_ec2/tasks/main.yml
jibsheet: @feanil the sandbox code removes anything which is a cname to a sandbox's ec2 dns name, which should cover this.  https://github.com/edx-ops/sysadmin/blob/master/jenkins/terminate-sandbox.py#L111

feanil: Ah, got it.

playbooks/roles/launch_ec2/tasks/main.yml
clintonb: Updated.

playbooks/roles/demo/defaults/main.yml
clintonb: Removed.

playbooks/roles/demo/tasks/main.yml
clintonb: Removed.

clintonb:
@feanil please take another look.

playbooks/roles/oauth_client_setup/meta/main.yml
```diff
@@ -0,0 +1,20 @@
+---
```
feanil: File can be deleted.

feanil:
@clintonb one nit but otherwise :+1:

playbooks/roles/oauth_client_setup/meta/main.yml
clintonb: Removed.

playbooks/edx_sandbox.yml
```diff
@@ -19,6 +19,8 @@
     COMMON_ENABLE_DATADOG: False
     COMMON_ENABLE_SPLUNKFORWARDER: False
     ENABLE_LEGACY_ORA: !!null
+    ECOMMERCE_NGINX_PORT: '80'
```
clintonb: @feanil any idea why these overrides are not showing up in the ecommerce Nginx file?

clintonb:
@feanil I confirmed that building a sandbox from scratch includes the ecommerce service. Building from the existing AMI does not include ecommerce; but, CMS and LMS work as expected. Ready for a new AMI!!!

PR #2232, opened 2015-08-21
Fix old mongo role

maxrothman:
@edx/devops

### DIFF
--- a/playbooks/roles/mongo/templates/repset_init.js.j2
+++ b/playbooks/roles/mongo/templates/repset_init.js.j2
@@ -1,6 +1,6 @@
 conn = new Mongo();
 db = conn.getDB("admin");
-db.auth( '{{ MONGO_ADMIN_USER }}', '{{ MONGO_ADMIN_PASSWORD }}');
+//no need to auth because of localhost exception: https://docs.mongodb.org/v2.6/core/authentication/#localhost-exception
 
 {# Generate a list of hosts if no cluster members are give. Otherwise use the
    hosts provided in the variable.
--- a/playbooks/roles/mongo/tasks/main.yml
+++ b/playbooks/roles/mongo/tasks/main.yml
@@ -73,31 +73,31 @@
 - name: wait for mongo server to start
   wait_for: port=27017 delay=2
 
-- name: drop super user script
-  template: src="create_root.js.j2" dest="/tmp/create_root.js"
-  when: not MONGO_CLUSTERED
-
-- name: create super user with js
-  shell: >
-    /usr/bin/mongo admin /tmp/create_root.js
-  when: not MONGO_CLUSTERED
-
-- name: delete super user script
-  file: path=/tmp/create_root.js state=absent
-  when: not MONGO_CLUSTERED
-
 - name: Create the file to initialize the mongod replica set
   template: src=repset_init.js.j2 dest=/tmp/repset_init.js
-  when: MONGO_CLUSTERED
+  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
 
 - name: Initialize the replication set
   shell: >
     /usr/bin/mongo /tmp/repset_init.js
-  when: MONGO_CLUSTERED
+  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
 
 #- name: delete repset script
 #  file: path=/tmp/repset_init.js state=absent
-#  when: MONGO_CLUSTERED
+#  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
+
+- name: drop super user script
+  template: src="create_root.js.j2" dest="/tmp/create_root.js"
+  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
+
+- name: create super user with js
+  shell: >
+    /usr/bin/mongo admin /tmp/create_root.js
+  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
+
+- name: delete super user script
+  file: path=/tmp/create_root.js state=absent
+  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
 
 - name: create a mongodb user
   mongodb_user: >
@@ -122,7 +122,7 @@
     state=present
     replica_set={{ mongo_repl_set }}
   with_items: MONGO_USERS
-  when: MONGO_CLUSTERED
+  when: MONGO_CLUSTERED and MONGO_PRIMARY == ansible_default_ipv4["address"]
 
 - name: install s3cmd
   pip: >


playbooks/roles/mongo/templates/repset_init.js.j2
```diff
@@ -1,6 +1,6 @@
 conn = new Mongo();
 db = conn.getDB("admin");
-db.auth( '{{ MONGO_ADMIN_USER }}', '{{ MONGO_ADMIN_PASSWORD }}');
+//no need to auth because of localhost exception: https://docs.mongodb.org/v2.6/core/authentication/#localhost-exception
```
carsongee, line 3: Does this really work without auth?  In my experience the cluster wouldn't form without it since as soon as you have one user the localhost exception goes away, so the second run of the play would fail.  Have you tested this with the vagrant cluster, and is it idempotent?

carsongee:
What is this fixing by the way?  I've run this on several clusters without any issues, so I'm not sure why it would be broken now

playbooks/roles/mongo/templates/repset_init.js.j2
maxrothman, line 3: I believe the role isn't idempotent anyways, since it attempts to initialize the replica set.

carsongee, line 3: It is actually, it reforms the replica set, but that doesn't cause any data loss, just a temporary loss of the replica set

maxrothman:
@carsongee I tried to run it yesterday in clustered mode and found that it broke in a couple ways:
- Auth failed when trying to initialize the replset because the admin user had not yet been created
- Creating users failed because the replset hadn't yet initialized, so none of the nodes knew who the master was

carsongee:
Strange, since there is a step you moved to create the admin user.  Also, I just ran a clean `vagrant up` in `vagrant/base/cluster.yml` and it provisioned fine.  Have you tried that?

carsongee:
Actually, it does skip those pieces for cluster, but it does them in the script that creates the replica set: https://github.com/edx/configuration/blob/master/playbooks/roles/mongo/templates/repset_init.js.j2#L59-L66

maxrothman:
Line 3 of repset_init.js used to be:

```
db.auth( '{{ MONGO_ADMIN_USER }}', '{{ MONGO_ADMIN_PASSWORD }}');
```

which would fail if the admin user had not yet been created. I moved the tasks to create the admin user before the repset init tasks to fix it.

I ran into these issues provisioning a fresh cluster on bare ec2 instances.

carsongee:
That should work just fine as it will pass if that user doesn't exist yet, so I'm not sure what is going on. Can you maybe paste the output?  I haven't ever seen that piece fail before, and I can't replicate with https://github.com/edx/configuration/blob/master/vagrant/base/cluster/Vagrantfile, and haven't seen it in AWS, or OpenStack in the past either.

maxrothman:
@carsongee When I run the mongo role without this PR against 3 fresh ubuntu instances in clustered mode, I get the following error:

```
TASK: [mongo | Initialize the replication set] ********************************
failed: [xx.xx.xx.xx] => {"changed": true, "cmd": "/usr/bin/mongo /tmp/repset_init.js", "delta": "0:00:30.049162", "end": "2015-08-24 18:32:13.649627", "rc": 253, "start": "2015-08-24 18:31:43.600465", "warnings": []}
stdout: MongoDB shell version: 2.6.11
connecting to: test
Error: 18 { ok: 0.0, errmsg: "auth failed", code: 18 }
{
    "startupStatus" : 3,
    "info" : "run rs.initiate(...) if not yet done for the set",
    "ok" : 0,
    "errmsg" : "can't get local.system.replset config from self or any seed (EMPTYCONFIG)"
}
2015-08-24T18:32:13.648+0000 Mongo Cluster Not Ok at /tmp/repset_init.js:13
failed to load: /tmp/repset_init.js

```

This indicates to me that the script failed to auth, hence my change.

carsongee:
That almost looks like maybe it is taking longer to form than normal and it needs to sleep longer.  The error seems to just be that the cluster isn't finished initializing yet since it is failing at [line 38](https://github.com/edx/configuration/blob/master/playbooks/roles/mongo/templates/repset_init.js.j2#L38) in the template, and not at the auth statement. I'm thinking maybe that [sleep statement](https://github.com/edx/configuration/blob/master/playbooks/roles/mongo/templates/repset_init.js.j2#L33) needs to wait longer for your systems.  Also, have you tried the vagrant cluster and gotten the same result?  It might also help to take a look at your play if you aren't using the [vagrant one](https://github.com/edx/configuration/blob/master/playbooks/vagrant-cluster.yml), or at least the mongo pieces out of it if the sleep extension doesn't end up helping.

maxrothman:
I'm using the [mongo play](https://github.com/edx/configuration/blob/master/playbooks/edx-east/mongo.yml), but I'll try extending that sleep and see what happens.

maxrothman:
Turned out to be partially PEBCAK: repset initiation was failing because of a security group issue. It ran successfully, but I still needed to change all of the `when` statements as per this PR (on [these](https://github.com/edx/configuration/blob/max/fix-old-mongo/playbooks/roles/mongo/tasks/main.yml#L76) [three](https://github.com/edx/configuration/blob/max/fix-old-mongo/playbooks/roles/mongo/tasks/main.yml#L80) [tasks](https://github.com/edx/configuration/blob/max/fix-old-mongo/playbooks/roles/mongo/tasks/main.yml#L114)). @carsongee would you be satisfied with a new PR with only those changes?

carsongee:
Can I see the new error output for how those are failing?  The repset_init script is primary aware, since any of them can become master during formation, so that shouldn't be required.  That is why it has https://github.com/edx/configuration/blob/master/playbooks/roles/mongo/templates/repset_init.js.j2#L42. The mongodb_user module is also replica set aware and shouldn't error out.

As implemented here with using primary ip addresses, I would also be a little concerned that that would be a bit fragile as the address ansible gets isn't always the right one, for example that would fail in the vagrant mongo cluster since they all report as having the same address as their primary one.

If we are having a race condition on rs.initiate() I think I would rather add some logic there to handle it, but I haven't seen that happen on any of the clusters we've built so far, and I couldn't find any docs on mongo about it failing when run on all members (but we might be a bit weird with this approach, so error messages would probably help in figuring it out).

maxrothman:
It's certainly not a race condition, because the play runs at most 1 host at a time.

The play fails with the same error as before, but it appears the the source is at `rs.initiate()` at line 6. Running the script line-by-line, I get the following error:

```
> rs.initiate(config)
{
    "ok" : 0,
    "errmsg" : "couldn't initiate : need all members up to initiate, not ok : 10.8.96.51:27017"
}
```

That's why rs.status().ok evaluates to false and the error gets thrown.

AFAIK this is the case in mongo3 as well. Perhaps the issue is that I'm running mongo 2.6, and mongo was more lenient in previous versions. What version are you running?

carsongee:
Oh, so you are running serial? The play is designed to run in parallel as per the [vagrant cluster play](https://github.com/edx/configuration/blob/master/playbooks/vagrant-cluster.yml#L1-L18), if you drop the serial it should work.  Sorry, I haven't paid too much attention to the edx-east play for this.

maxrothman:
Removing the serial option fixed it. It seems to have been added in https://github.com/edx/configuration/pull/1522, and it's probable we haven't run the mongo play in clustered mode since then. Thanks @carsongee!

I'll close out this PR.

PR #2233, opened 2015-08-21
Overrides for SAML public key and private key vars.

feanil:
@maxrothman

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -519,6 +519,10 @@ EDXAPP_PROCTORING_BACKEND_PROVIDER:
 # Comprehensive Theming
 EDXAPP_COMP_THEME_DIR: ""
 
+# SAML KEYS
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -519,6 +519,10 @@ EDXAPP_PROCTORING_BACKEND_PROVIDER:
 # Comprehensive Theming
 EDXAPP_COMP_THEME_DIR: ""
 
+# SAML KEYS
+EDXAPP_SOCIAL_AUTH_SAML_SP_PRIVATE_KEY: ''
+EDXAPP_SOCIAL_AUTH_SAML_SP_PUBLIC_CERT: ''
```
maxrothman, line 524: Any reason not to use "change me" default values?

feanil, line 524: Because this is the default in edx-platform.  Using something different might result in the platform trying to use it.

PR #2234, opened 2015-08-22
Benp/fix packer sitespeed aug2015

benpatterson:
@jzoldak 

Last packer ami build on Jenkins was successful against this branch. :smiley:


jzoldak:
:+1:

PR #2238, opened 2015-08-24
DEVOPS-2620 - DISPLAY_ANALYTICS_ENROLLMENTS

fredsmith:


jibsheet:
@dsjen  If we ship this as defaulting to False in configuration, is that going to conflict with the default in the LMS (so in devstacks?).  You mentioned that this will change something in production and I just want to understand what.

dsjen:
@jibsheet -- good question.  I wasn't picturing that this would change anything in devstack--just in our deployments.  The default setting in lms/env/common.py remains the same and the setting isn't overwritten in lms/envdevstack.py.

jibsheet:
@fredsmith looks like this should default to True to match common.py https://github.com/edx/edx-platform/blob/master/lms/envs/common.py#L338 and then we can override in secure.

jibsheet:
:+1:

PR #2243, opened 2015-08-25
OPS-976: Remove serial option from mongo play

maxrothman:
The mongo play doesn't work when run serially. It's run non-serially in
other places (vagrant-cluster.yml, for example). The serial option seems
to have been added mistakenly in 2632cb8 (#1522) and not tested.

@edx/devops 
FYI @carsongee

### DIFF
--- a/playbooks/edx-east/mongo.yml
+++ b/playbooks/edx-east/mongo.yml
@@ -2,9 +2,6 @@
   hosts: all
   sudo: True
   gather_facts: True
-  vars:
-    serial_count: 1
-  serial: "{{ serial_count }}"
   roles:
     - mongo
     - role: datadog


carsongee:
The [elasticsearch play](https://github.com/edx/configuration/blob/master/playbooks/edx-east/elasticsearch.yml) can also run in parallel as well (or serial), but not sure if we want to bother changing that (which might complicate the commoncluster play), :+1: regardless.

maxrothman:
I've filed [a bug](https://openedx.atlassian.net/browse/OPS-980) to look into elasticsearch and all other plays modified in #1522.

fredsmith:
:+1: :shipit:

PR #2244, opened 2015-08-25
Benp/jenkins worker java 8

benpatterson:

PR #2245, opened 2015-08-25
Use virtualenv install available on PATH.

benpatterson:
@jzoldak


benpatterson:
Closing in deference to #2244

PR #2251, opened 2015-08-31
Feanil/update sandbox workers

feanil:
Before this change, you couldn't actually update workers because their config never got updated and they never got restarted when you updated the edx-platform.  This will allow us to quickly update the workers.  We can't do this as part of the edxapp.yml play because that is used in multiple places.

@edx/devops @jibsheet


jibsheet:
:+1: http://i.imgur.com/SXUBN.gif

PR #2254, opened 2015-08-31
make logrotate not crash clusters

fredsmith:


maxrothman:
Whoops, guess I only put that in the standalone config.

maxrothman:
:+1:

PR #2256, opened 2015-09-01
WIP: simple working API client, sample play.

e0d:
@arbabnazar Basic direction that I'm thinking of for the module for managing MMS alerts.

@feanil early thoughts?  Any 2.0 concerns we should be keeping in mind?

### DIFF
--- a/playbooks/library/mms_alert.py
+++ b/playbooks/library/mms_alert.py
@@ -1,0 +1,63 @@
+#!/usr/bin/env python
+
+DOCUMENTATION = """
+---
+module: mms_alert
+short_description: 
+description:
+version_added: "1.9.3"
+author: Edward Zarecor
+options:
+"""
+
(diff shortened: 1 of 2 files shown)


playbooks/library/mms_alert.py
```diff
@@ -0,0 +1,63 @@
...
+EXAMPLES = '''
+'''
+
+from ansible.module_utils.basic import *
+import json
+import requests
+
+class CloudManagerClient(object):
+    def __init__(self, base_url, group_id, api_user, api_key):
+        self.base_url = base_url + "/api/public/v1.0"
```
feanil, line 23: Maybe start with a default value for base_url and allow overrides for custom installations.

feanil:
General direction seems good.  The interface to modules seems to be the same still as far as I can tell.

PR #2257, opened 2015-09-01
Separate installing datadog and new relic on sanboxes

arbabnazar:
@e0d can you please review it. Thanks


e0d:
We'll want to update the Jenkins jobs to support the new flags.

@edx/devops second look.

feanil:
:+1:

PR #2258, opened 2015-09-01
Whitelist the lti_provider endpoint

jibsheet:
This uses OAuth1 and configured tokens and shouldn't use basic auth.

@edx/devops


feanil:
:+1:

PR #2261, opened 2015-09-02
Added LTI_AGGREGATE_SCORE_PASSBACK_DELAY, default of 15 mins (900s).

ormsbee:
@jibsheet, @feanil: Companion piece to https://github.com/edx/edx-platform/pull/9582

@sarina, @mcgachey: FYI

An aggregate score is one derived from multiple problems (such as the
cumulative score for a vertical element containing many problems). Sending
aggregate scores immediately introduces two issues: one is a race condition
between the view method and the Celery task where the updated score may not
yet be visible to the database if the view has not yet returned (and committed
its transaction). The other is that the student is likely to receive a stream
of notifications as the score is updated with every problem. Waiting a
reasonable period of time allows the view transaction to end, and allows us to
collapse multiple score updates into a single message.

The time value is in seconds.


jibsheet:
:+1:

feanil:
:+1:

jibsheet:
@ormsbee any reason not to merge this?

ormsbee:
@jibsheet: None. I just lost forgot. :-P Thank you for the reminder.

PR #2262, opened 2015-09-02
Rerun local_dev whenever /edx/bin/update is run on devstack

bdero:
Most of the plays that are run from the update command interfere with the conveniences that are setup by local_dev at the time of provisioning. E.g. the shell of the user for the app which is updated is reset to `/bin/false`.

### DIFF
--- a/playbooks/local_dev.yml
+++ b/playbooks/local_dev.yml
@@ -1,0 +1,14 @@
+- name: Setup local development conveniences
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    edxapp_user: edxapp
+    forum_user: forum
+    ora_user: ora
+    notifier_user: notifier
+    ecommerce_user: ecommerce
+    analytics_api_user: analytics_api
+    insights_user: insights
+  roles:
+    - local_dev
--- a/playbooks/roles/edx_ansible/templates/update.j2
+++ b/playbooks/roles/edx_ansible/templates/update.j2
@@ -73,3 +73,12 @@ fi
 
 cd {{ edx_ansible_code_dir }}/playbooks/edx-east
 eval "sudo ${repos_to_cmd["$1"]} $verbose"
+
+{% if devstack %}
+if [ $? -eq 0 ]; then
+    cd {{ edx_ansible_code_dir }}/playbooks
+    echo
+    echo "Running local development configuration"
+    sudo {{ edx_ansible_venv_bin }}/ansible-playbook -i localhost, -c local local_dev.yml
+fi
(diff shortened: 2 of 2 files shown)


openedx-webhooks:
Thanks for the pull request, @bdero! I've created [OSPR-793](https://openedx.atlassian.net/browse/OSPR-793) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

nedbat:
@feanil we were just discussing this.

feanil:
@bdero I don't think we ever intended for people to use the update script to make updates to the platform in devstack.  The expectation is that devstack version management, migrations, requirements installs, etc would be handled by the developer.

bdero:
@feanil What's the intention of the update script, then? The devstack docs say to use it: https://github.com/edx/configuration/wiki/edX-Developer-Stack#updating-application-versions

bdero:
Although that instance is specifically for the config repo ^

feanil:
Yea, it is meant so that you can update the config repo and is used by the provision clause of the Vagrantfile.  If you look at the Vagrantfile you'll see that after it updates configuration it will run the devstack playbook to finish out the updates.  The devstack playbook will always run the local_dev role.

bdero:
I see that there have been steps taken to make jumping between edx-platform versions less work (installing requirements when running paver, for instance), which probably automates away most of the headache. I can see that while the devstack play takes a large amount of time to run, but it probably doesn't need to be run very often in the average case.

bdero:
Hmm, I thought about this a little more - while people can manage the config themselves, I don't think it hurts or is difficult to make it work with devstack. I think everyone's tried to use the update script and got confused by things not working, so maybe it should either be supported, or better document that it's specifically not supported. ;)

feanil:
I think it actually makes more sense to document that we don't support it.  I've also been toying with the idea of just not having it exist on devstacks by default. I think the current provision step that the devstack does is not actually necessary and blurs who should be responsible for managing the code on a devstack.

What do you think @bdero ?

sarina:
@bdero ping? Can you reply to Feanil's question above?

sarina:
I'm closing this PR, taking silence as agreement to Feanil's question from 18 days ago.

bdero:
I'm sort of agnostic about the `/edx/bin/update` script being used post-provision, but how would things work without the update script being ran during provisioning? Is the idea to hand out a pre-provisioned devstack? Cont. https://github.com/edx/configuration/pull/2271

PR #2265, opened 2015-09-04
Tagging tasks to tease out problems with scheme

e0d:
Tagging two roles with the notional lifecycle tags discussed in design time and "tweaked" by me.

@edx/devops

### DIFF
--- a/playbooks/roles/edx_service/tasks/main.yml
+++ b/playbooks/roles/edx_service/tasks/main.yml
@@ -54,6 +54,9 @@
     createhome=yes
     shell=/bin/false
     generate_ssh_key=yes
+  tags:
+    - install
+    - install:base
 
 # Assumes that the home directory has been created above.
 # In some cases(vagrant boxes) the home directory gets created
@@ -72,6 +75,9 @@
     - "venvs"
     - "data"
     - "staticfiles"
+  tags:
+    - install
+    - install:base
 
 - name: create /edx/var and /edx/etc dirs
   file: >
@@ -83,6 +89,9 @@
   with_items:
     - /edx/var
     - /edx/etc
+  tags:
+    - install
+    - install:base
 
 - name: create edx_service log dir
   file: >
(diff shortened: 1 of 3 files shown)


playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -101,27 +109,36 @@
   tags:
```
feanil: Already has a tags section.

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -12,6 +14,8 @@
     mode=0644 state=link force=yes
   with_items: ['xqueue', 'xqueue_consumer']
   when: not disable_edx_services
+  tags:
+    - install:configuration
```
maxrothman, line 20: This is more than just writing app configuration. Maybe it should have a separate 2nd-level tag?

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -19,18 +23,24 @@
   sudo_user: "{{ xqueue_user }}"
   notify:
     - restart xqueue
+  tags:
+    - install:configuration
```
maxrothman, line 30: Same as above

maxrothman:
In general, should we require that all tasks have some tag?

e0d:
@maxrothman @feanil

feanil:
:+1:

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -72,6 +75,9 @@
     - "venvs"
```
maxrothman, line 75: Why create this directory early under install:base rather than in install:app-requirements? (same comment as on [wiki page](https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle))

maxrothman:
Should [L156](https://github.com/edx/configuration/pull/2265/files#diff-366c95739562a5eeb748e2d3e4e03cafL156) be tagged with `to-remove`?

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -12,6 +15,9 @@
     mode=0644 state=link force=yes
   with_items: ['xqueue', 'xqueue_consumer']
   when: not disable_edx_services
+  tags:
+    - install
+    - install:configuration
```
maxrothman, line 20: I know we discussed whether activation should be part of the `manage` tag, but I don't think we reached a conclusion. That discussion is probably worth finishing.

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -91,7 +121,7 @@
   register: supervisor_update
   changed_when: supervisor_update.stdout is defined and supervisor_update.stdout != ""
   when: not disable_edx_services
-
+
```
maxrothman: Remove added whitespace

playbooks/roles/xqueue/tasks/deploy.yml
```diff
@@ -111,5 +144,8 @@
   with_items:
   - python
   - pip
+  tags:
+    - install
+    - install:base
```
maxrothman: If someone were to run only the `install:base` tag, this would fail because the virtualenv wouldn't have been created.

playbooks/roles/xqueue/tasks/main.yml
```diff
@@ -26,11 +29,17 @@
   with_items:
     - "{{ xqueue_app_dir }}"
     - "{{ xqueue_venvs_dir }}"
+  tags:
+    - install
+    - install:base
```
maxrothman, line 34: Same comment as with edxapp tasks

playbooks/roles/edx_service/tasks/main.yml
e0d, line 75: Because the value of having directory creation JIT is less than having a single-purpose task for building out app structures in my view.

playbooks/roles/xqueue/tasks/deploy.yml
e0d, line 20: I think this is something use will teach us better than speculation.

e0d:
@maxrothman @feanil updated based on CR comments.

feanil:
:+1:

PR #2266, opened 2015-09-07
Update Insights config to support enrollment API

bradenmacdonald:
This makes the config changes required to use the new user profiles feature of https://github.com/edx/edx-analytics-dashboard/pull/338 in production and on a devstack. (Specifically, it adds the URL of the enrollment API to the Insights dashboard settings)

**Dependencies**:
- https://github.com/edx/edx-analytics-dashboard/pull/338
- https://github.com/edx/edx-analytics-dashboard/pull/334

**Note**: This can be considered WIP, and I am not expecting any reviews until the dependency is reviewed and merged.

**Partner information**: Not an edX partner - 3rd party-hosted open edX instance


smarnach:
:+1: Looks good to me.

playbooks/roles/insights/defaults/main.yml
```diff
@@ -61,6 +61,7 @@ INSIGHTS_DATABASES:
     PORT: '3306'
 
 INSIGHTS_LMS_COURSE_SHORTCUT_BASE_URL: "URL_FOR_LMS_COURSE_LIST_PAGE"
+INSIGHTS_CMS_COURSE_SHORTCUT_BASE_URL: "http://127.0.0.1:8001/courses"
```
smarnach: Any reason to have 127.0.0.1:8001 hardcoded for Studio, but not for the LMS?

feanil: let's use `localhost` instead of 127.0.0.1 to be consistent with the other roles.

+1 to what @smarnach said.  Lets correct the LMS_COURSE_SHORTCUT url to work on sandboxes.

bradenmacdonald: Good points. Fixed. Note some of the settings in this file use `127.0.0.1` because that's how OAuth with the LMS is configured (so `localhost` won't work) but in this case we can use either one so I went with localhost now.

playbooks/roles/insights/defaults/main.yml
```diff
@@ -60,7 +60,8 @@ INSIGHTS_DATABASES:
     HOST: '127.0.0.1'
     PORT: '3306'
 
-INSIGHTS_LMS_COURSE_SHORTCUT_BASE_URL: "URL_FOR_LMS_COURSE_LIST_PAGE"
+INSIGHTS_LMS_COURSE_SHORTCUT_BASE_URL: "{{ INSIGHTS_LMS_BASE }}/courses"
+INSIGHTS_CMS_COURSE_SHORTCUT_BASE_URL: "http://localhost:8001/courses"
```
mtyaka: The CMS shortcut base should probably default to `/course` rather than `/courses`. Course URLs in the Studio have the form `/course/<course-id>`.

bradenmacdonald: Thanks @mtyaka :)

bradenmacdonald:
Closing since the related PR was rejected.

PR #2267, opened 2015-09-08
Remove plugins we don't use on Admin.

feanil:
These are a few plugins that I don't think we use on admin anymore and should be removed from this list.

@edx/devops

### DIFF
--- a/playbooks/roles/jenkins_admin/defaults/main.yml
+++ b/playbooks/roles/jenkins_admin/defaults/main.yml
@@ -94,9 +94,7 @@ jenkins_admin_plugins:
   - { name: "matrix-auth", version: "1.2" }
   - { name: "matrix-project", version: "1.3" }
   - { name: "mailer", version: "1.9" }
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_admin/defaults/main.yml
```diff
@@ -105,7 +103,6 @@ jenkins_admin_plugins:
   - { name: "parameterized-trigger", version: "2.25" }
   - { name: "multiple-scms", version: "0.3" }
   - { name: "git", version: "2.2.2" }
```
maxrothman, line 105: We don't use the git plugin anymore?

feanil, line 105: You're reading the diff incorrectly.

maxrothman, line 105: (doh)

maxrothman:
:+1:

PR #2269, opened 2015-09-08
Specify a default image for the forums notifier

sarina:
@feanil 

Default path should be https://github.com/edx/edx-platform/blob/master/lms/static/images/default-theme/logo.png


feanil:
:+1:

PR #2271, opened 2015-09-09
Ensure ansible is run on the release vagrant box

mulby:
Release the analyticstack Vagrantfile.

Note that for this to work the box file needs to be uploaded to files.edx.org. I submitted [a ticket for that work.](https://openedx.atlassian.net/browse/DEVOPS-2804)

Reviewers: @e0d, @feanil

### DIFF
--- a/vagrant/release/analyticstack/Vagrantfile
+++ b/vagrant/release/analyticstack/Vagrantfile
@@ -8,6 +8,19 @@ VAGRANTFILE_API_VERSION = "2"
 MEMORY = 4096
 CPU_COUNT = 2
 
+$script = <<SCRIPT
+if [ ! -d /edx/app/edx_ansible ]; then
+    echo "Error: Base box is missing provisioning scripts." 1>&2
+    exit 1
+fi
+export PYTHONUNBUFFERED=1
+source /edx/app/edx_ansible/venvs/edx_ansible/bin/activate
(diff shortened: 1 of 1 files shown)


feanil:
@mulby what is the motivation behind running ansible on startup, I know we currently do this for the release devstack but I've been seriously considering not doing so.

mulby:
The repos aren't being checked out into the NFS mounted folders, we could manually do those steps in the Vagrantfile?

mulby:
@feanil I didn't want to replicate the logic associated with the git checkout since it was non-trivial and governed by a bunch of variables... I suppose we could create a tag that only performed checkouts?

feanil:
Why not just let the user manage the checkouts themselves?  This is what I was thinking of doing with the release version of devstack.  This is already the expectation for day to day use so I figure the odd this is doing an update for them at all in the first place.

mulby:
We wanted a "one-click" setup. So you would just "vagrant up" and everything would work instead of "vagrant up" followed by "checkout 6 different repos into the correct folders".

mulby:
@feanil and @e0d - Note that there is a bit of time pressure here, since we want to release our documentation and this Vagrantfile along with a blog post next week. Everything is ready to go, but we need to merge this to master before we publish the docs since they instruct users to download this file.

feanil:
:+1:

e0d:
:+1:

PR #2274, opened 2015-09-09
Add missing shared folders and port forwards to Vagrantfile.

jimabramson:
@feanil @fredsmith @rlucioni 

My main priority with this PR is to get the ecommerce shared folders into the release version of the Vagrantfile.  I noticed that a bunch of port forwards were missing compared to the base Vagrantfile, so I pulled those in too.

Another difference I have NOT addressed is an env-driven switch around the ora shared folder.  I was confused by how this is implemented in the base Vagrantfile as it seems to be implemented twice, cf:
- https://github.com/edx/configuration/blob/master/vagrant/base/devstack/Vagrantfile#L70
- https://github.com/edx/configuration/blob/master/vagrant/base/devstack/Vagrantfile#L88

N.B. I have not looked at fullstack or analyticstack, so if there are corresponding changes needed in there, those are still outstanding.

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -43,6 +43,7 @@ edx_platform_mount_dir = "edx-platform"
 themes_mount_dir = "themes"
 forum_mount_dir = "cs_comments_service"
 ora_mount_dir = "ora"
+ecommerce_mount_dir = "ecommerce"
 insights_mount_dir = "insights"
 analytics_api_mount_dir = "analytics_api"
 
@@ -52,6 +53,7 @@ if ENV['VAGRANT_MOUNT_BASE']
   themes_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + themes_mount_dir
   forum_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + forum_mount_dir
   ora_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + ora_mount_dir
+  ecommerce_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + ecommerce_mount_dir
   insights_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + insights_mount_dir
   analytics_api_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + analytics_api_mount_dir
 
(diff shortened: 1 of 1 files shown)


vagrant/release/devstack/Vagrantfile
```diff
@@ -124,11 +126,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
-  config.vm.network :forwarded_port, guest: 8000, host: 8000
-  config.vm.network :forwarded_port, guest: 8001, host: 8001
-  config.vm.network :forwarded_port, guest: 18080, host: 18080
+  config.vm.network :forwarded_port, guest: 8000, host: 8000  # LMS
+  config.vm.network :forwarded_port, guest: 8001, host: 8001  # Studio
+  config.vm.network :forwarded_port, guest: 8002, host: 8002  # Ecommerce
+  config.vm.network :forwarded_port, guest: 8003, host: 8003  # LMS for Bok Choy
+  config.vm.network :forwarded_port, guest: 8031, host: 8031  # Studio for Bok Choy
+  config.vm.network :forwarded_port, guest: 8120, host: 8120  # edX Notes Service
   config.vm.network :forwarded_port, guest: 8765, host: 8765
```
jimabramson, line 135: This is the only port number I didn't recognize offhand, anyone know what's using it?

vagrant/release/devstack/Vagrantfile
```diff
@@ -149,6 +155,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       create: true, owner: "edxapp", group: "www-data"
     config.vm.synced_folder "#{forum_mount_dir}", "/edx/app/forum/cs_comments_service",
       create: true, owner: "forum", group: "www-data"
+    config.vm.synced_folder "#{ecommerce_mount_dir}", "/edx/app/ecommerce/ecommerce",
+      create: true, owner: "ecommerce", group: "www-data"
```
rlucioni, line 159: I think you also want to update this conditional's else clause, run when VAGRANT_USE_VBOXFS is false. See [here](https://github.com/edx/configuration/blob/master/vagrant/base/devstack/Vagrantfile#L85-L86).

jimabramson, line 159: thanks, nice catch, updated.

rlucioni:
Looks fine to me, :+1:. @feanil or @fredsmith, what's the difference between the Vagrantfiles in /base and /release? Is this PR tantamount to releasing a new version of devstack?

jimabramson:
@rlucioni I asked that in devops a few minutes ago:

> the ones under base are used to build the base boxes that the ones under release use, this means the ones under release come up a lot faster,
> another major difference is that the ones under base expect to be run from inside the config repo with a venv that has ansible, the ones under release are more free standing

rlucioni:
@jimabramson thank you, I missed that. Makes sense.

vagrant/release/devstack/Vagrantfile
feanil, line 135: It's for LTI testing on devstack: https://github.com/edx/configuration/pull/995

feanil:
:+1:

PR #2276, opened 2015-09-10
Proposing removing conditionals around monitoring and splunk

e0d:
@edx/devops 

Thoughts?

### DIFF
--- a/playbooks/edx-east/vpc_admin.yml
+++ b/playbooks/edx-east/vpc_admin.yml
@@ -12,8 +12,5 @@
     - hotg
     - alton
     - role: datadog
-      when: COMMON_ENABLE_DATADOG
     - role: splunkforwarder
-      when: COMMON_ENABLE_SPLUNKFORWARDER
     - role: newrelic
-      when: COMMON_ENABLE_NEWRELIC


maxrothman:
What's the motivation for this change?

e0d:
Seems to me they should just be installed on admin and we don't need to blindly follow the pattern of conditionalizing these roles everywhere.

maxrothman:
Is this a playbook that we're fairly sure is only used by us? If so, :+1: and maybe we should think about moving it out of this repo.

fredsmith:
I think we should leave these conditionals in - I think this play could be useful to other people, and those people might not want to run datadog/splunk/NR. 

Unless we have a different mechanisim for managing edX deployments, I think we should encourage others to use asgard/jenkins/alton as their deployment system on AWS.

feanil:
+1 to what @fredsmith said.

e0d:
I don't really think it is.  I'd actually lobby for moving it to
ansible-private.

Why do you think it's generally useful?

fredsmith:
@e0d I'll admit, it does seem a little far fetched that someone running openEdX would want an entire build/management system running ansible and jenkins just to keep their instances up to date if they're not doing intensive development.  Also, since our jenkins jobs are all private, all this really does is set up the base software and give people a blank configuration.  not too useful. We'd need to do a bunch more work to make our entire deployment system easily reproducable.   However, we attempt to be open and public by default, so I'd wonder what we gain by pulling this off the public repo if we're still using it.  It would make our public code simpler to pick up, because the repo won't be littered with unhelpful playbooks.

maxrothman:
The issue of public vs private to me seems orthogonal to the discussion of whether to leave the conditionals in or even to leave this playbook in the configuration repo. I think there's a separate discussion to be had as to whether the "ansible-private" repo should, in fact, be public.

As I see it, we have two options:
1. Do nothing
2. Remove the conditionals and move the playbook out of this repo

I think that choice should be informed by what sorts of things belong in the configuration repo, which I would argue are things that pertain to the deployment and maintenance of applications in the edX ecosystem. Since this does not, I would advocate for option 2. I would also advocate for making the "ansible-private" repo public (after taking the necessary precautions, of course).

fredsmith:
:microphone: is this thing on?

PR #2277, opened 2015-09-10
permit LTI auth requests through basic auth

fredsmith:


jibsheet:
Does this mean no third party auth worked on stage?
Should we make this specific to the LTI endpoint?
I wonder how we use 3rd party auth on stage at all (which I know people do)

jibsheet:
:+1:

feanil:
@jibsheet 3rd party auth would still work but just prompt the user for the basic auth creds on redirects.  I think that's how people are still able to use it.

PR #2278, opened 2015-09-11
Allow adding in of storage specific configuration

jibsheet:
This can be used to configure wiredTiger's memory usage away from "half
of the RAM on the box".

@edx/devops


fredsmith:
:+1:

PR #2280, opened 2015-09-11
Without this, users with a password of "pass word" fail

jibsheet:
It turns it into key/value arguments to the ansible module, causing
hilarious failures.


fredsmith:
:+1:  - I think there's more things here to fix, but this works for now.

PR #2281, opened 2015-09-15
Create nodeenv using prebuilt package

clintonb:
Using a prebuilt node package decreases the time to create a nodeenv from minutes to a couple seconds.

ECOM-2185

@feanil 

FYI @rlucioni @jimabramson @dsjen @mulby

### DIFF
--- a/playbooks/roles/ecommerce/tasks/main.yml
+++ b/playbooks/roles/ecommerce/tasks/main.yml
@@ -38,7 +38,7 @@
 - name: create nodeenv
   shell: >
     creates={{ ecommerce_nodeenv_dir }}
-    {{ ecommerce_home }}/venvs/{{ ecommerce_service_name }}/bin/nodeenv {{ ecommerce_nodeenv_dir }}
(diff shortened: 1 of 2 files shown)


rlucioni:
:+1:

feanil:
How does it do this?  Any concern about pre-built packages not existing or version issues?

clintonb:
Unless specified nodeenv always uses the latest version of nodejs. This is
fine for our purposes. Since we aren't specifying a version, there is no
concern about an older version not existing. Additionally, nodejs.org has
distributions going back to v0.5.1. We are using v4.0.0. (Note nodejs
switched versioning schemes. The version prior to v4.0.0 was v0.12.7.)

The prebuilt option instructs nodeenv to (a) pull the tarball for a
platform-specific version of nodejs from https://nodejs.org/dist/v4.0.0/
and (b) simply copy the built files to the appropriate directory instead of
building from source.

feanil:
:+1:

PR #2282, opened 2015-09-15
Enable overriding of the elastic search config set in aws.py

jibsheet:
DEVOPS-2816


feanil:
:+1:

lgfa29:
@feanil @jibsheet is there a reason why this variable is only set on `lms_env_config` and not on `edxapp_generic_env`? The current implementation doesn't allow to override this variable for the CMS.

jibsheet:
At the time, this was only used in the LMS, so that's what was configured.
If it's now used by the CMS please submit a PR that moves it into generic_env_config and I'm happy to look at it.

PR #2284, opened 2015-09-16
On vagrant curl is not available by default.

feanil:
This breaks running the user role on a fresh vagrant box where common has not already been run. This
change should fix that.

@edx/devops


fredsmith:
I suppose  :+1:

PR #2289, opened 2015-09-18
Test courses are split mongo, and thus no longer use SlashSeparatedCourseKey

maxrothman:
@jzoldak @jibsheet 
@edx/devops

### DIFF
--- a/playbooks/roles/testcourses/tasks/deploy.yml
+++ b/playbooks/roles/testcourses/tasks/deploy.yml
@@ -21,7 +21,7 @@
 
 - name: import the test courses
   shell: >
-    {{ demo_edxapp_venv_bin }}/python ./manage.py cms --settings=aws import {{ demo_edxapp_course_data_dir }} /var/tmp/{{ item.path|basename }}/{{ item.course_id.split('/')[2] }}
+    {{ demo_edxapp_venv_bin }}/python ./manage.py cms --settings=aws import {{ demo_edxapp_course_data_dir }} /var/tmp/{{ item.path|basename }}/{{ item.xml_path }}
     chdir={{ demo_edxapp_code_dir }}
   sudo_user: "{{ common_web_user }}"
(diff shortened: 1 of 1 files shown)


feanil:
:+1:

jibsheet:
:+1:

jzoldak:
I used the ansible-provision job to create a sandbox with max's branch and when I sign in to the lms as the honor user I am still not enrolled in the course. `sudo tail -f /edx/var/log/lms/edx.log` shows the same error messages about the user enrolled in broken or non-existent courses.

Here's a [link to my build](http://jenkins.edx.org:8080/job/ansible-provision/3528/parameters/) if it helps.

jibsheet:
Max tells me he sees a bug in his code, however, you'll need to build a sandbox from scratch (specifying the base 12.04 image which we can find you) since the courses are imported during the base sandbox image build, not during sandbox provisioning.  Happy to talk through that process for testing.

jzoldak:
Whatever's most expedient, either tell me which ami to use for the image or kick off a job. Just want to make sure that someone tests this out before merging so that we don't have to revisit it later.

jibsheet:
I figured out this requires a change to the sandbox secure file.  I'll steal this from Max to do as part of OPS-1019 which will end up also doing the DEVOPS ticket.

jibsheet:
@jzoldak jibsheet.sandbox.edx.org now lacks any messages about being enrolled in a broken course.  Was there any other test you were doing?

There's a secure change needed, but I'll also be doing two OPS tickets related to updating test courses before merging that.

jzoldak:
@jibsheet if when logging in as one of the test users (honor, audit, verified), the dashboard shows that the user is enrolled in the demo course and the other test courses (I think there are 6 total?) then :+1: from me.

jibsheet:
@feanil I didn't change Max's code, but added some, can you sanity check (it's the second commit)

feanil:
:+1:

PR #2294, opened 2015-09-21
We don't need to do multi-line strings for ansible anymore.

feanil:
@clintonb The newer version of ansible supports sub dictionaries for module parameters.


clintonb:
:+1:

PR #2295, opened 2015-09-21
Add ecommerce_worker role

rlucioni:
Allows deployment of the ecommerce worker, a Celery-based IDA for asynchronously handling ecommerce tasks introduced in https://github.com/edx/ecommerce-worker/pull/1. ECOM-2306.

@feanil, I'd appreciate a look from you, when you have a chance. @clintonb and @jimabramson, FYI.

### DIFF
--- a/playbooks/roles/ecommerce/defaults/main.yml
+++ b/playbooks/roles/ecommerce/defaults/main.yml
@@ -99,7 +99,9 @@ ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
 # Celery
 ECOMMERCE_BROKER_USERNAME: 'celery'
 ECOMMERCE_BROKER_PASSWORD: 'celery'
-ECOMMERCE_BROKER_HOST: '127.0.0.1'
+# Used as the default RabbitMQ IP.
+ECOMMERCE_BROKER_HOST: '{{ ansible_default_ipv4.address }}'
+# Used as the default RabbitMQ port.
 ECOMMERCE_BROKER_PORT: 5672
 ECOMMERCE_BROKER_URL: 'amqp://{{ ECOMMERCE_BROKER_USERNAME }}:{{ ECOMMERCE_BROKER_PASSWORD }}@{{ ECOMMERCE_BROKER_HOST }}:{{ ECOMMERCE_BROKER_PORT }}'
 
--- a/playbooks/roles/ecommerce_worker/tasks/main.yml
+++ b/playbooks/roles/ecommerce_worker/tasks/main.yml
@@ -1,0 +1,73 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+# Tasks for role ecommerce_worker.
+# 
+
+- name: install application requirements
+  pip:
+    requirements: '{{ ecommerce_worker_requirements_base }}/{{ item }}'
+    virtualenv: '{{ ecommerce_worker_home }}/venvs/{{ ecommerce_worker_service_name }}'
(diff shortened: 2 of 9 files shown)


playbooks/roles/ecommerce_worker/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+# Tasks for role ecommerce_worker.
+# 
+
+- name: install application requirements
+  pip: >
```
feanil: Use the new style syntax:

```
pip:
  requirements: ...
  virtualenv: ...
  state: ...
```

This new style is valid anywhere you're using the multiline string(strings with > at the beginning) right now.

playbooks/roles/ecommerce_worker/templates/edx/app/ecommerce_worker/ecommerce_worker.sh.j2
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
+
+# {{ ansible_managed }}
+
+source {{ ecommerce_worker_home }}/ecommerce_worker_env
+{{ executable }} celery -A ecommerce_worker worker --app=ecommerce_worker.celery_app:app --loglevel=info
```
feanil: The number of concurrent workers should be configurable.

playbooks/roles/ecommerce_worker/tasks/main.yml
rlucioni: Updated. Thanks for the heads-up.

playbooks/roles/ecommerce_worker/templates/edx/app/ecommerce_worker/ecommerce_worker.sh.j2
rlucioni: Updated.

feanil:
@rlucioni it looks like you need a rebase.

rlucioni:
@feanil thanks for the heads up. Something about GitHub's recent changes makes it seem like branches need to be rebased more often than they actually do; the branch didn't actually require a rebase in this case.

rlucioni:
@feanil I think we're good to go here. I resolved issues with Celery logging and verified that ecommerce and the worker are able to connect to RabbitMQ. Would you mind taking another look when you have a moment?

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -99,7 +99,9 @@ ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
 # Celery
 ECOMMERCE_BROKER_USERNAME: 'celery'
 ECOMMERCE_BROKER_PASSWORD: 'celery'
-ECOMMERCE_BROKER_HOST: '127.0.0.1'
+# Used as the default RabbitMQ IP.
+ECOMMERCE_BROKER_HOST: '{{ ansible_default_ipv4.address }}'
```
feanil, line 103: What does this value resolve to in devstack?  We haven't needed to do this with other services that connect to rabbitmq as far as I know, why for this one?

rlucioni, line 103: This resolves to `10.0.2.15`. It's used as the [default IP for RabbitMQ in the rabbitmq role](https://github.com/edx/configuration/blob/master/playbooks/roles/rabbitmq/defaults/main.yml#L52). During testing, RabbitMQ doesn't appear to listen on localhost and refuses connection attempts on `127.0.0.1`. Output from `rabbitmqctl report` indicates that RabbitMQ has a single listener attached to `10.0.2.15`:

```
$ sudo rabbitmqctl report
...
{tcp_listeners,[{"10.0.2.15",5672}]},
...
```

feanil:
:+1:

PR #2299, opened 2015-09-23
Remove old custom plugins and code that is needed to build them.

feanil:
We don't use the feature that we added to our fork.

@edx/devops @edx/testeng

### DIFF
--- a/playbooks/roles/jenkins_master/defaults/main.yml
+++ b/playbooks/roles/jenkins_master/defaults/main.yml
@@ -4,7 +4,7 @@ jenkins_group: "edx"
 jenkins_server_name: "jenkins.testeng.edx.org"
 jenkins_port: 8080
 
-jenkins_version: 1.574
+jenkins_version: "1.630"
 jenkins_deb_url: "http://pkg.jenkins-ci.org/debian/binary/jenkins_{{ jenkins_version }}_all.deb"
 jenkins_deb: "jenkins_{{ jenkins_version }}_all.deb"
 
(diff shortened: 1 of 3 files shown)


playbooks/roles/jenkins_master/defaults/main.yml
```diff
@@ -4,7 +4,7 @@ jenkins_group: "edx"
 jenkins_server_name: "jenkins.testeng.edx.org"
 jenkins_port: 8080
 
-jenkins_version: 1.574
+jenkins_version: "1.630"
```
benpatterson, line 7: Ooo nice.

playbooks/roles/jenkins_master/meta/main.yml
```diff
@@ -1,3 +1,4 @@
 ---
 dependencies:
   - common
+  - oraclejdk
```
benpatterson, line 4: I just want to call out that you may need to override defaults here if you want Java 8.

feanil, line 4: It currently defaults to 7 and I thought that was fine for now.  Changing it to 8 seemed risky as the default.

benpatterson:
You'll just want to be sure you are good with Java and from my perspective this gets a :+1:

playbooks/roles/jenkins_master/meta/main.yml
benpatterson, line 4: Ok. No objection here. Just the fact that you looked is all I wanted. We are planning to move to 8 in a month or so and we can revisit at that time.

jibsheet:
:+1:

PR #2301, opened 2015-09-24
Updates for ecommerce in devstack

clintonb:
- Added explicit values for variables that must be shared between ecommerce and edxapp
- Added ecommerce virtualenv bin directory to PATH for ecommerce user
- Using local Django settings (instead of production)
- Made static file compilation optional (defaulting to false)
- Allowing ecommerce version to be specified via environment variable

ECOM-2185

@feanil @jibsheet

### DIFF
--- a/playbooks/roles/ecommerce/defaults/main.yml
+++ b/playbooks/roles/ecommerce/defaults/main.yml
@@ -35,6 +35,7 @@ ECOMMERCE_DATABASES:
     CONN_MAX_AGE: 60
 
 ECOMMERCE_VERSION: "master"
+ECOMMERCE_DJANGO_SETTINGS_MODULE: "ecommerce.settings.production"
 
 ECOMMERCE_SECRET_KEY: 'Your secret key here'
 ECOMMERCE_TIME_ZONE: 'UTC'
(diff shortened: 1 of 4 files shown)


playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -83,16 +83,18 @@
     chdir={{ ecommerce_code_dir }}
     . {{ ecommerce_nodeenv_bin }}/activate && {{ ecommerce_node_bin }}/r.js -o build.js
   sudo_user: "{{ ecommerce_user }}"
+  when: ecommerce_compile_static is defined and ecommerce_compile_static|lower == "yes"
```
clintonb: @feanil @jibsheet thoughts? Have I updated the appropriate playbooks to ensure this _is_ run for sandboxes and other non-devstack environments?

feanil:
@clintonb what's the motivation? You don't want to compile static when building devstacks?  Wont django ignore those when running in debug mode?

clintonb:
@feanil no, we don't want to compile static assets for devstack. Devstack isn't setup to serve the compiled assets, and compiling them would require developers working on static assets to delete the compiled directory. Lastly, the compress command fails due to a setting (set for production) not being set for devstack (because it shouldn't be set).

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -186,6 +184,11 @@ ecommerce_requirements:
   - production.txt
   - optional.txt
 
+ecommerce_environment:
+  DJANGO_SETTINGS_MODULE: "{{ ECOMMERCE_DJANGO_SETTINGS_MODULE }}"
+  ECOMMERCE_CFG: "{{ COMMON_CFG_DIR }}/{{ ecommerce_service_name }}.yml"
+  PATH: "{{ ecommerce_venv_dir }}/bin:${PATH}"
```
feanil, line 192: What's the reason for the path update?

playbooks/vagrant-devstack.yml
```diff
@@ -15,6 +15,16 @@
     EDXAPP_LMS_BASE: 127.0.0.1:8000
     EDXAPP_OAUTH_ENFORCE_SECURE: false
     EDXAPP_LMS_BASE_SCHEME: http
+
+    # E-Commerce variables
+    ECOMMERCE_ECOMMERCE_URL_ROOT: "http://localhost:8002"
```
feanil: Isn't this the default for this setting?  Why does it need to be overridden?  Same questions for the urls and keys below, I would prefer for the defaults to just work.

playbooks/roles/ecommerce/defaults/main.yml
clintonb, line 192: I want to avoid having to use the full path to `python` or `pip`. Copied the behavior from edxapp.

clintonb:
@feanil any other feedback?

clintonb:
@feanil @jibsheet please review.

feanil:
:+1:

PR #2303, opened 2015-09-24
Modified local ecommerce database user

clintonb:
Granted the local ecommerce database user full privileges. This will ensure developers can continue to run migrations without any special commands or environment variables.

ECOM-2185

@jibsheet @feanil


jibsheet:
:+1:

PR #2306, opened 2015-09-24
Django IDA Template Updates

clintonb:
- Added quotes where necessary
- Removed quotes around integer values (e.g. timeout, port)
- Updated service user's env to include the service's virtualenv on the PATH
- Created a variable for the service's virtualenv directory

ECOM-2180

@jibsheet @feanil


feanil:
:+1:

PR #2307, opened 2015-09-24
Make sure we check and run the ecommerce role

jibsheet:


jibsheet:
@clintonb I believe this was the issue, going to rebuild my sandbox
@feanil

feanil:
:+1:

clintonb:
:+1:

PR #2308, opened 2015-09-24
New base sandbox AMI

jibsheet:
Cut as a result of the work in OPS-1003 OPS-1019 and adding the new
ecommerce role (work by Clinton).

@edx/devops
@clintonb


feanil:
:+1:

clintonb:
:+1:

PR #2312, opened 2015-09-28
Update config variable names for Segment

bderusha:
@rlucioni is this all I need to do?  How do we actually set the real keys? DevOps?


rlucioni:
@bderusha file an devops support ticket to have stage and production values set for these variables in our secure repo. That aside, this looks good to me, :+1: . You should try to get a thumb from someone in devops (e.g., @feanil, @jibsheet).

bderusha:
@edx/devops This config change goes along with https://github.com/edx/edx-platform/pull/9875

feanil:
@bderusha can you please also add a note in the changelog indicating this is a backwards incompatible change that has renamed two top level variables.

bderusha:
@feanil added a note in the changelog.  Let me know if you'd like any more detail in there.

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+- Role: edxapp
+  - Renamed two top level variables SEGMENT_IO_LMS_KEY and SEGMENT_IO_KEY to SEGMENT_KEY.
```
feanil: Can you also provide what you re-name them to here as well?

bderusha: I did :) `"to SEGMENT_KEY"`

feanil: I meant specifically to EDXAPP_CMS_SEGMENT_KEY and EDXAPP_LMS_SEGMENT_KEY in the global configuration scope.  People who are using ansible to manage their config will use these global names to change the setting rather than overriding the full dictionary so it's good to have them present in the docs.

fredsmith:
are we still able to disable segment on the CMS?  Will a key value of "None" switch off Segment?

bderusha:
@fredsmith yes, a value of `None` will disable Segment

bderusha:
@feanil updated the changelog.  Let me know if there is anything else you would like me to add to it.

fredsmith:
:+1:

feanil:
:+1:

PR #2314, opened 2015-09-29
Fixed OAuth setup bug for sandboxes

clintonb:
ECOM-2180

@feanil

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -143,19 +143,20 @@ ease_version: $ease_version
 certs_version: $certs_version
 discern_version: $discern_version
 configuration_version: $configuration_version
-ECOMMERCE_VERSION: $ecommerce_version
 EDXAPP_STATIC_URL_BASE: $static_url_base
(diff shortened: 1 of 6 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -261,7 +262,7 @@ EOF
 fi
 
 declare -A deploy
-roles="edxapp forum ecommerce notifier xqueue xserver ora discern certs demo testcourses"
+roles="edxapp forum ecommerce notifier xqueue xserver ora discern certs demo testcourses oauth_client_setup"
```
feanil: If you want to do it this way, you also need to create a playbook by the same name.  Talke a look at line 281/282 below to see why.  Another option is to just run that role using the 'run_role.yml' playbook at the same time as we run edx_ansible.yml(line 287/288)

`run ansible ../run_role.yml $extra_var_arg -e "role=oauth_client_setup"`

clintonb: Updated.

feanil:
:sheep:  :it:

PR #2315, opened 2015-09-29
Allow setting up ecommerce_worker in devstack.

peter-fogg:
@rlucioni @feanil This PR adds configuration for the ecommerce_worker in devstack. I'm not too familiar with Ansible, so much of this is lifted from the ecommerce service setup.


playbooks/vagrant-devstack.yml
```diff
@@ -24,6 +24,8 @@
     - elasticsearch
     - forum
     - ecommerce
+    - ecommerce_worker
+    - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
```
feanil, line 28: Why is this explicitly set to 127.0.0.1?

peter-fogg, line 28: It's copied from `vagrant-fullstack.yml`. Should it be unset?

rlucioni:
:+1: Please file a support ticket to have values for these variables set in the secure repo.

playbooks/vagrant-devstack.yml
peter-fogg, line 28: @feanil?

feanil:
:+1:

PR #2318, opened 2015-09-30
Fixed OAuth setup bug for sandboxes

clintonb:
ECOM-2180

Replaces #2314 

@feanil

### DIFF
--- a/playbooks/edx-east/oauth_client_setup.yml
+++ b/playbooks/edx-east/oauth_client_setup.yml
@@ -1,0 +1,11 @@
+- name: Configure OAuth2 clients
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars_files:
+    - "roles/common/defaults/main.yml"
+    - "roles/edxapp/defaults/main.yml"
+    - "roles/insights/defaults/main.yml"
+    - "roles/ecommerce/defaults/main.yml"
+  roles:
+    - oauth_client_setup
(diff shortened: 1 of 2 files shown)


clintonb:
@feanil @jibsheet @jimabramson @rlucioni ready for final review

playbooks/edx-east/oauth_client_setup.yml
```diff
@@ -0,0 +1,11 @@
...
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars_files:
+    - "roles/common/defaults/main.yml"
+    - "roles/edxapp/defaults/main.yml"
+    - "roles/insights/defaults/main.yml"
+    - "roles/ecommerce/defaults/main.yml"
+  roles:
+    - oauth_client_setup
```
rlucioni, line 11: Does the ecommerce playbook need to be updated to use this role?

rlucioni:
I don't understand why a separate playbook is required for OAuth2 client configuration. @clintonb for those unfamiliar, can you provide some context on why this was necessary?

playbooks/edx-east/oauth_client_setup.yml
feanil, line 11: no.

feanil:
@rlucioni this is required to be seperate because it needs to be run on every sandbox as the sandbox is being provisioned, not just when the base box is being built.  This is because of the callback URLs will change with each sandbox.  It needs to be a playbook because it needs to be run in ansible_provision.sh and needs access to vars from the other roles.

@clintonb :+1:

rlucioni:
Thanks @feanil. :+1:

PR #2319, opened 2015-09-30
Add JWT_ISSUERS configuration for ecommerce.

peter-fogg:
@rlucioni @jimabramson @clintonb @feanil


rlucioni:
:+1:

jimabramson:
:+1:

clintonb:
Why are there two issuers?

clintonb:
Never mind. :+1:

peter-fogg:
@clintonb The ecommerce service needs to accept JWTs from both the LMS and the worker process.

clintonb:
Reverting because `ecommerce_worker_service_name` is undefined. See http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/3698/console.

@feanil please advise how we should share variables across roles.

rlucioni:
From the [Ansible docs](http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable):

> Generally speaking, variables set in one role are available to others. This means if you have a `roles/common/vars/main.yml` you can set variables in there and make use of them in other roles and elsewhere in your playbook.

Sounds like what was being done here was legal. @peter-fogg if variables can't be shared across roles in this way, you might try passing the issuer name to the ecommerce_worker role as a parameter; see "parameterized roles" in the link above.

PR #2320, opened 2015-09-30
Added programs service to devstack

clintonb:
ECOM-2184

@feanil @jimabramson @rlucioni

FYI @bderusha @peter-fogg @cpennington

### DIFF
--- a/playbooks/vagrant-devstack.yml
+++ b/playbooks/vagrant-devstack.yml
@@ -26,6 +26,7 @@
     - ecommerce
     - ecommerce_worker
     - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
+    - programs
     - role: notifier
       NOTIFIER_DIGEST_TASK_INTERVAL: "5"
     - role: ora
--- a/vagrant/base/devstack/Vagrantfile
+++ b/vagrant/base/devstack/Vagrantfile
@@ -14,6 +14,7 @@ forum_mount_dir = "cs_comments_service"
 ora_mount_dir = "ora"
 ecommerce_mount_dir = "ecommerce"
 ecommerce_worker_mount_dir = "ecommerce-worker"
+programs_mount_dir = "programs"
 insights_mount_dir = "insights"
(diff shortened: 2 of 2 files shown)


rlucioni:
:+1:

clintonb:
@fredsmith @maxrothman review please?

maxrothman:
Same comment as https://github.com/edx/configuration/pull/2321

clintonb:
Yes. At this point were are waiting on DevOps.

## 

Clinton Blackburn

feanil:
Have you successfully built this?  What about fullstack?

clintonb:
@feanil yes, I have built it. We are not focusing on fullstack at the moment.

jimabramson:
any outstanding issues with this?   Having programs in devstack is about to become very useful.

clintonb:
@jibsheet Review?

feanil:
@singingwolfboy What are your thoughts on having full stack and devstack not matching with respect to which apps are running?  If this is okay from the opensource perspective it is fine by me but I would want those two to match so that we can have go parity for the open source community when they switch between these.

jimabramson:
@singingwolfboy @feanil FYI this will become a blocker soon, as we need for platform and programs to be able to talk to each other inside devstack to get development work done.

jbarciauskas:
How much effort is it to add fullstack?

clintonb:
@jbarciauskas I have never touched fullstack; however, I would estimate 1-2 days. This is work that can be done later, closer to when we are ready to support open source installations of the programs service.

jbarciauskas:
OK, I don't have a particular burning need for these to be done at the same time, but I want to make sure it doesn't get lost. If you're confident of that, then we can push ahead here

clintonb:
ECOM-2597 created to track fullstack development.

feanil:
:+1:

clintonb:
@feanil what are the next steps for getting a new base box, and updating the release `Vagrantfile`?

PR #2323, opened 2015-10-01
Fixed analytics database issue

clintonb:
- Fixed issue with migration user not having access to analytics databases
- Further consolidated creation of analytics databases and users

ECOM-2180

@feanil @mulby This should resolve the nightly build breakage. Please let me know how I can test.

### DIFF
--- a/playbooks/roles/edxlocal/defaults/main.yml
+++ b/playbooks/roles/edxlocal/defaults/main.yml
@@ -15,6 +15,8 @@ edxlocal_databases:
   - "{{ EDXAPP_MYSQL_DB_NAME | default(None) }}"
   - "{{ EDX_NOTES_API_MYSQL_DB_NAME | default(None) }}"
   - "{{ PROGRAMS_DEFAULT_DB_NAME | default(None) }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/edxlocal/defaults/main.yml
```diff
@@ -15,6 +15,8 @@ edxlocal_databases:
   - "{{ EDXAPP_MYSQL_DB_NAME | default(None) }}"
   - "{{ EDX_NOTES_API_MYSQL_DB_NAME | default(None) }}"
   - "{{ PROGRAMS_DEFAULT_DB_NAME | default(None) }}"
+  - "{{ ANALYTICS_API_DEFAULT_DB_NAME | default('None') }}"
```
mulby: Does it matter that None is in quotes here? I only ask because it isn't in the other entries in this list.

clintonb: It definitely matters. This is why we have reviews! Fixed!

mulby:
cool. :+1:

clintonb:
@feanil @jibsheet @maxrothman

feanil:
:+1:

PR #2324, opened 2015-10-01
Increase client upload size limit for jenkins master

jzoldak:
@benpatterson @e0d 

nginx was giving a 413 response code when trying to upload a local copy of the github-sqs-plugin from pluginManager/advanced because the .hpi is just under 16M.


benpatterson:
:+1:

PR #2325, opened 2015-10-01
Theming work was revereted from master.

feanil:
This is related to that so it is also reverted.


feanil:
@jibsheet @fredsmith

jibsheet:
@singingwolfboy  This is what you'll want to go back.

:+1:

singingwolfboy:
Thanks for finding this, guys.

PR #2327, opened 2015-10-03
Added ecommerce host to LMS CORS whitelist for sandboxes

clintonb:
ECOM-2469

@feanil @rlucioni @jimabramson


feanil:
@clintonb why a new variable?

clintonb:
@feanil is there a better way to append to a list?

feanil:
What would set it that you're afraid of overwriting?

clintonb:
I want the changes from configuration_secure in addition to the sandbox host. How would I combine those two values?

clintonb:
@devops Any feedback?

feanil:
@clintonb does the cors origin whitelist take wildcards?  This seems like the sort of thing we can add in sandbox-secure if that is the case.

clintonb:
@feanil that would be awesome, assuming exposing `*.sandbox.edx.org` is not a security risk.

feanil:
For sandboxes I think it's probably fine, actually I think the correct thing is to pas the hostname to ansible and then do it in sandbox secure, I think this is not a code change but a config change to sandbox secure.  Can you put in a ticket for that?

clintonb:
@feanil DEVOPS-3093

PR #2328, opened 2015-10-05
Remove unsupported Vagrant boxes

singingwolfboy:
edX can only support the latest Open edX release, which is Cypress. This commit clears out all the cruft from old Open edX releases that we cannot support anyway.


feanil:
:+1:

PR #2329, opened 2015-10-05
Add JWT_ISSUERS list to ecommerce defaults

rlucioni:
Follow up to https://github.com/edx/configuration/pull/2319, which was reverted.

@peter-fogg @clintonb @feanil


peter-fogg:
:+1:

clintonb:
:+1:

rlucioni:
Could I get a thumb from someone in @edx/devops?

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -121,8 +121,10 @@ ECOMMERCE_SERVICE_CONFIG:
   LMS_DASHBOARD_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/dashboard'
   JWT_AUTH:
     JWT_SECRET_KEY: '{{ ECOMMERCE_JWT_SECRET_KEY }}'
-    JWT_ISSUER: '{{ ECOMMERCE_LMS_URL_ROOT }}/oauth2'
     JWT_VERIFY_EXPIRATION: '{{ ECOMMERCE_JWT_VERIFY_EXPIRATION }}'
+  JWT_ISSUERS:
+    - '{{ ECOMMERCE_LMS_URL_ROOT }}/oauth2'
+    - 'ecommerce_worker'
```
maxrothman: I don't have the background on this, but why isn't this also a URL?

rlucioni: @maxrothman because it doesn't make sense to use a URL to identify the worker. According to the JWT spec, the issuer claim should identify the application issuing the token; processing of the claim is application-specific. In this case, the string `ecommerce_worker` is the clearest way to identify the worker app.

maxrothman:
:+1: after rebase
Pending @jibsheet's comment

feanil:
:+1:

rlucioni:
Thanks for the reviews everyone. It turns out that the correct fix for ecommerce's JWT settings is a bit more nuanced than we first thought. I've updated this PR to include values for JWT_ALGORITHM and the JWT_DECODE_HANDLER settings.

Could you please take another look? Let me know if you have questions.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -121,8 +123,12 @@ ECOMMERCE_SERVICE_CONFIG:
   LMS_DASHBOARD_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/dashboard'
   JWT_AUTH:
     JWT_SECRET_KEY: '{{ ECOMMERCE_JWT_SECRET_KEY }}'
-    JWT_ISSUER: '{{ ECOMMERCE_LMS_URL_ROOT }}/oauth2'
+    JWT_ALGORITHM: '{{ ECOMMERCE_JWT_ALGORITHM }}'
     JWT_VERIFY_EXPIRATION: '{{ ECOMMERCE_JWT_VERIFY_EXPIRATION }}'
+    JWT_DECODE_HANDLER: '{{ ECOMMERCE_JWT_DECODE_HANDLER }}'
+  JWT_ISSUERS:
```
clintonb: Is there any reason this cannot be nested in the `JWT_AUTH` dict?

rlucioni: @clintonb yes and no; it's a judgement call.

`JWT_ISSUERS` isn't one of the default settings provided by DRF-JWT. That being the case, if it's nested in the `JWT_AUTH` dict, it can't be accessed using the nice `api_settings` object - like other DRF-JWT settings - without some manual overrides on our part. If we want to be able to access it via the `api_settings` object, we'll have to override DRF-JWT's [default settings](https://github.com/GetBlimp/django-rest-framework-jwt/blob/master/rest_framework_jwt/settings.py#L9).

The alternatives are pulling the setting directly from the `JWT_AUTH` dictionary (i.e., `settings.JWT_AUTH['JWT_ISSUERS']`), or leaving it out of the `JWT_AUTH` dictionary altogether, making it clear that this setting isn't part of DRF-JWT.

clintonb: I see nothing wrong with `settings.JWT_AUTH['JWT_ISSUERS']`.

rlucioni: @clintonb updated. Please see https://github.com/edx/ecommerce/pull/378 for relevant ecommerce changes.

clintonb:
:+1:

rlucioni:
@edx/devops, could I please get a second look here? Thank you.

jibsheet:
:+1:

PR #2331, opened 2015-10-06
OPS-966 - disable transparent hugepages

fredsmith:


playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -22,7 +22,21 @@
     state=absent purge=yes
     force=yes
   when: mongodb_needs_upgrade.stat.exists and not MONGO_CLUSTERED
-  
+
+- name: disable transparent huge pages on startup
```
feanil: Can you add a link to [this page](http://docs.mongodb.org/manual/tutorial/transparent-huge-pages/) in the comments as well so we can easily see why we're doing this.

feanil:
:+1: once comment is added.

PR #2338, opened 2015-10-12
Allow overriding of the django SESSION_COOKIE_SECURE setting

jibsheet:
https://django.readthedocs.org/en/1.4/ref/settings.html
DEVOPS-3064


jibsheet:
@edx/devops

feanil:
:+1:

e0d:
I think this deserves a mention on the list and maybe the change log.

jibsheet:
@e0d  This change actually leaves it false (which is the default from edx-platform).
Our change is done in the secure override files.
If we decide to default it on here, then I agree, Changelog/list would be required.

I think we should default a number of things to more secure defaults, but that's not this PR.

PR #2339, opened 2015-10-13
Removing AWS dependency here, why was it duped?

e0d:


fredsmith:
git blame says: https://github.com/edx/configuration/commit/41159619e2853de7cbacbf38d2817c82950e336c

"allowing ansible-provision to do bare installations"

I can't see why the code in that commit should need the aws role.  :+1:

feanil:
:+1:

PR #2351, opened 2015-10-14
Updated programs role to support static assets

clintonb:
ECOM-2559

This is copied from ecommerce.

@feanil @jimabramson @rlucioni @jibsheet


rlucioni:
Is static asset configuration part of the Ansible IDA template?

clintonb:
@rlucioni nope. I am going to leave that addition to the next person who needs an IDA with static assets.

rlucioni:
:+1:

jimabramson:
looks reasonable to me :+1:

PR #2353, opened 2015-10-14
Fix insights DB name to use the right variable

shnayder:
INSIGHTS_DEFAULT_DB_NAME isn't actually set anywhere.

We verified that the playbook got past the migrate step for Insights with this change. 

@mulby @clintonb @e0d


clintonb:
:+1:

mtyaka:
:+1:

PR #2360, opened 2015-10-15
Refactoring and cleanups that resulted from DEVOPS-3068

jibsheet:
Add tags so you can run part of the role, and particularly, deploy just
mongo scripts, or just mysql scripts for an environment.

Install 3.0.7 mongo tools so the shell works with our new clusters, as
well as supporting all the features of our 2.6 compose instances.

Clean up ansible syntax.


jibsheet:
@edx/devops

feanil:
:+1:

PR #2362, opened 2015-10-15
starting to turn analytics_single into a single-box sandbox playbook

shnayder:
Added missing hive metastore config. 
Removed unneeded EDXAPP config. 
Turned edx services back on.

@mulby @e0d

### DIFF
--- a/playbooks/vagrant-fullstack.yml
+++ b/playbooks/vagrant-fullstack.yml
@@ -40,4 +40,3 @@
     - certs
     - role: analytics_api
       when: ANALYTICS_API_GIT_IDENTITY
-    - edx_ansible
--- a/playbooks/edx-east/analytics_single.yml
+++ b/playbooks/edx-east/analytics_single.yml
@@ -4,12 +4,15 @@
   gather_facts: True
   vars:
     migrate_db: "yes"
-    disable_edx_services: true
-    EDXAPP_LMS_BASE: 127.0.0.1:8000
-    EDXAPP_OAUTH_ENFORCE_SECURE: false
-    EDXAPP_LMS_BASE_SCHEME: http
+    disable_edx_services: false
+    ENABLE_DATADOG: False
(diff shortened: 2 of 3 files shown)


mulby:
:+1:

playbooks/vagrant-fullstack.yml
```diff
@@ -40,4 +40,4 @@
     - certs
     - role: analytics_api
       when: ANALYTICS_API_GIT_IDENTITY
-    - edx_ansible
+    - edx_ansible    # TODO: why is this line repeated? Remove if not needed, comment otherwise.
```
feanil: Shouldn't need this twice, let's remove it.

feanil:
:+1:

playbooks/vagrant-fullstack.yml
shnayder: removed

jibsheet:
Unfortunately, this appears to be breaking Continuous Integration
http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/18283/console

Relevant bit below (looks like something isn't getting substituted?).

failed: [int.sandbox.edx.org] => (item={'db': u'{# HIVE_METASTORE_DATABASE.name | default(None) #}', 'user': u'{# HIVE_METASTORE_DATABASE.user | default(None) #}', 'pass': u'{# HIVE_METASTORE_DATABASE.password | default(None) #}'}) => {"failed": true, "item": {"db": "{# HIVE_METASTORE_DATABASE.name | default(None) #}", "pass": "{# HIVE_METASTORE_DATABASE.password | default(None) #}", "user": "{# HIVE_METASTORE_DATABASE.user | default(None) #}"}}
01:12:26 msg: (1470, "String '{# HIVE_METASTORE_DATABASE.user | default(None) #}' is too long for user name (should be no longer than 16)")

PR #2363, opened 2015-10-15
Separate installing mysql for configuring apps that use it.

feanil:
edxlocal is too generic and does a lot of things.  This is the beginning
of separating it into roles that are more single purpose and easier to
understand.


fredsmith:
:+1:

PR #2366, opened 2015-10-16
Fix edxlocal when not running analytics stack build

mulby:
@e0d - I'm far from an ansible expert, it looks like there is a consequential difference between {{ FOO | default(None) }} and {{ FOO.bar | default(None) }}. In the former case, it actually falls back to the default if FOO is not defined, in the later case, it uses {# FOO.bar | default(None) #} as the value of the variable. Is that correct? If so, the "| default(None)" clause is not really very meaningful. For future reference - is there a simple way I can run the CI build locally before merging?

@shnayder, @feanil - sanity check me here

### DIFF
--- a/playbooks/roles/analytics_pipeline/defaults/main.yml
+++ b/playbooks/roles/analytics_pipeline/defaults/main.yml
@@ -11,6 +11,7 @@
 # Defaults for role analytics_pipeline
 #
 
+ANALYTICS_PIPELINE_OUTPUT_DATABASE_NAME: "{{ ANALYTICS_API_REPORTS_DB_NAME }}"
 ANALYTICS_PIPELINE_OUTPUT_DATABASE:
   username: pipeline001
   password: password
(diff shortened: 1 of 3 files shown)


jibsheet:
Ah, yes, this jives with my note on #2362 
We can run the CI job with this as the configuration branch, but that just uncovers:
http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/18288/console

You can also try booting a sandbox and using the reconfigure flag with your branch

mulby:
@jibsheet cool, thanks for the pointers.

Is there a known workaround for this {{ dictionary.access | default }} issue? Or is it what we are doing here...

e0d:
@mulby I suspect that ansible will return none for the var reference FOO though it doesn't exist and None will be falsey.  None.bar, however, will through an exception.  Don't think we have a null coalescing operator we can use.

A possible workaround is:

```
- name: dict
  debug: var={{ FOO.bar | default(None) }}
  when: FOO is defined
```

mulby:
Alright, it looks like my latest commit fixed the issue. I applied the same pattern used elsewhere here.

http://jenkins.edx.org:8080/job/ansible-provision-continuous-integration/18309/

e0d:
:+1:

feanil:
:+1:

PR #2379, opened 2015-10-20
Mount a src directory for ad-hoc use.

nedbat:
When developing on devstack, it's useful to have a way to share source trees
between the host and the guest OS. The src directory is for this sharing, so
that working trees can be accessed inside the guest OS.

### DIFF
--- a/vagrant/base/devstack/Vagrantfile
+++ b/vagrant/base/devstack/Vagrantfile
@@ -17,6 +17,7 @@ ecommerce_worker_mount_dir = "ecommerce-worker"
 programs_mount_dir = "programs"
 insights_mount_dir = "insights"
 analytics_api_mount_dir = "analytics_api"
+src_mount_dir = "src"
 
 if ENV['VAGRANT_MOUNT_BASE']
 
@@ -29,6 +30,9 @@ if ENV['VAGRANT_MOUNT_BASE']
   programs_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + programs_mount_dir
   insights_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + insights_mount_dir
   analytics_api_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + analytics_api_mount_dir
+  # This src directory won't have useful permissions. You can set them from the
+  # vagrant user in the guest OS. "sudo chmod 0777 /edx/src" is useful.
+  src_mount_dir = ENV['VAGRANT_MOUNT_BASE'] + "/" + src_mount_dir
 
 end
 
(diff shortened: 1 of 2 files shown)


nedbat:
@singingwolfboy @feanil @maxrothman I'm not sure if I got all of the details right, but this is an important thing to add to devstack so that developers will stop hacking their own solutions.

singingwolfboy:
You've scoped the `src` directory under `/edx/app/edxapp/`, but I'm not sure that fulfills all use cases -- especially as we add more applications and services into devstack. Maybe call this `edxapp_src` on the host OS, instead?

nedbat:
Yeah, I'm not sure where to put it and who should own it.  For devstack, is it important that we have all these different accounts?

maxrothman:
What types of files do you envision being shared with this method?

feanil:
@nedbat why not just mount '/edx/app' from the VM to local?

nedbat:
@maxrothman mostly, it would be repos that developers are working on, so they can install them locally in the devstack, and commit changes from them in the host, though I've found it useful for other things too (dotfiles going in, log files coming out).
@feanil I didn't consider that as a possibility, it changes the existing mapping of edxapp/edx-platform, right?  Also, it means the entire virtualenv is on a shared directory, and I'm not sure what the performance implications of that would be.
If you have other ideas about how developers can share trees from the host to the guest, I'm all ears.

nedbat:
@andy-armstrong this is the one.

andy-armstrong:
@maxrothman as some background, here's what the Lahore team has been doing to develop ORA2 on devstack (from @symbolist):

https://openedx.atlassian.net/wiki/display/TNL/ORA+FAQ?replyToComment=44073390#ORAFAQ-UsingORAwithdevstack

It would be great to have a more standard mechanism for managing developments of all of our IDAs from within devstack. We could add each IDA individually to the Vagrantfile, but having a `src` directory seems like an elegant solution to the problem.

andy-armstrong:
I was thinking more about this last night. A lot of folks have similar workarounds, but what I like about this PR is that the `src` directory makes it clear that you are bringing in other repos for development. This directory won't be there in production, and instead the developer uses it to `pip install` new features that they need. There is a slight risk that a developer would add a dependency on `src` accidentally, but this would be seen quickly on sandboxes, stage etc. :+1: from me.

maxrothman:
I think that perhaps a better location for these sort of files might be in `/edx/var`

nedbat:
If the var in /edx/var is meant to correspond to the Filesystem Hierarchy Standard (https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard), then var makes no sense to me.  The files I'm talking about are developers' work.

maxrothman:
If you're primarily talking about ancillary files (dotfiles, logs, etc.) then `var` seems reasonable. For new repos, you are correct.

maxrothman:
Just to clarify, the use case is for repos that go in edx-platform, not other IDAs?

nedbat:
@maxrothman and I discussed offline, and settled on /edx/src as a location, with liberal permissions.

nedbat:
@maxrothman @feanil I have changed the Vagrantfile to use /edx/src, but cannot figure out how to set the owner, group, and mode.  When I test this, I get a directory owned by used id 502, which is my id on the host OS.
Can anyone help a poor newb? thanks in advance :)

feanil:
@nedbat you can set the owner on that file yourself: https://docs.vagrantup.com/v2/synced-folders/basic_usage.html

My suggestion would be vagrant?

nedbat:
@feanil I'm already using those settings.  When starting a devstack, we are in the NFS branch, not the vagrantfs branch, and the owner is incorrect.  Are you suggesting I make my devstack differently?  How are the existing directories given the proper owners?  I don't see how that happens in the current code.

feanil:
:sheep: :it:

PR #2380, opened 2015-10-20
DEVOPS-3094 - Require HTTPS access to insights.edx.org

arbabnazar:
@feanil can you please review it. Thanks


feanil:
@arbabnazar can you please make the same change for programs as well.

FYI: @mulby @clintonb @jimabramson @rlucioni

arbabnazar:
@feanil please review.

feanil:
:+1:

clintonb:
Can you update the ansible-role-django-ida role as well? It generates the nginx config for future IDAs.

:+1:

clintonb:
FYI @mattdrayer

feanil:
@clintonb it looks like the ansible-role-django-ida role doesn't generate the NGINX template at all right now... this makes sense since it is not part of the role but a new template that is added to the existing NGINX role.

PR #2387, opened 2015-10-21
Allow us to override the boto ebs_optimized flag

jibsheet:
Ansible defaults to False, despite documentation that doesn't state
this.  Boto then turns off EBS Optimization on instance types where this
should be impossible.
https://github.com/ansible/ansible-modules-core/pull/2326

@edx/devops


feanil:
:+1:

PR #2392, opened 2015-10-22
Add splunk indexing from jenkins of sitespeed results

jzoldak:
This needs the props.conf change from DEVOPS-3130 to be done first before applying to jenkins, so that the data will index correctly.

@benpatterson @arbabnazar

### DIFF
--- a/playbooks/edx-east/jenkins_testeng_master.yml
+++ b/playbooks/edx-east/jenkins_testeng_master.yml
@@ -18,18 +18,27 @@
         sourcetype: junit
         followSymlink: false
         crcSalt: '<SOURCE>'
+
       - source: '/var/lib/jenkins/jobs/*/builds/*/build.xml'
         index: 'testeng'
         recursive: true
         sourcetype: build_result
         followSymlink: false
         crcSalt: '<SOURCE>'
+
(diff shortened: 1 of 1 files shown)


playbooks/edx-east/jenkins_testeng_master.yml
```diff
@@ -18,18 +18,27 @@
...
         recursive: true
         sourcetype: build_log
         followSymlink: false
         crcSalt: '<SOURCE>'
+
+      - source: '/var/lib/jenkins/jobs/*/builds/*/archive/sitespeed-result/*/data/result.json'
+        index: 'testeng'
+        recursive: false
+        sourcetype: sitespeed_result
+        followSymlink: false
```
benpatterson, line 40: do we need the crcSalt in this case?

feanil:
:+1:

playbooks/edx-east/jenkins_testeng_master.yml
jzoldak, line 40: No, because this file is written to disk all at once. We need the crcSalt in the other cases so that splunk knows when indexing whether or not a file is the same one it already saw, but has been appended to.

benpatterson:
:+1:

PR #2393, opened 2015-10-22
Only close cache on post_fork() if cache supports it.

ormsbee:
In particular LocMemCache doesn't have this option.

@mulby, @e0d, @adampalay, @jibsheet


mulby:
:+1:

adampalay:
:+1:

jibsheet:
:+1:

PR #2395, opened 2015-10-23
Add mysql docker container

robrap:
Hackathon work for mysql docker container

@e0d Feel free to do what you want with this.  I just wanted to make it available to you.

This includes grabbing Dockerfile from dockerhub as well as loading edx
users into db.  The db VOLUME is not shared at this point.


fredsmith:
:+1:

PR #2401, opened 2015-10-23
one of us one of us

fredsmith:
@robrap @e0d


fredsmith:
actually, it looks like it would be good to set these all to edxops devops@edx.org or somesuch.

PR #2404, opened 2015-10-23
We don't want to use this tag anymore, it's not reliable.

feanil:
Also it breaks expectations when adding new non deploy tasks.

@edx/devops


maxrothman:
According to a quick grep, this is the last reference to `--tags deploy` in the repo!

:tada: :fireworks: :+1:

PR #2408, opened 2015-10-26
Recurse through dirs for sitespeed indexing on splunk. PERF-182

jzoldak:
@benpatterson 

Otherwise the "*" from the source specifier doesn't really do anything, so the files don't get indexed.


benpatterson:
:+1:

PR #2410, opened 2015-10-26
Make tagging work with new tag input format for ec2_asg

feanil:
It looks like the format of this module changed in one of the upgrades on is not compatible with what we're doing right now.  This resolves it.

Module docs: http://docs.ansible.com/ansible/ec2_asg_module.html

### DIFF
--- a/playbooks/edx-east/edx_service.yml
+++ b/playbooks/edx-east/edx_service.yml
@@ -163,19 +163,27 @@
     - debug: msg="{{ created_service_subnets.results|map(attribute='subnet_id')| list | join(',') }}"
       register: service_vpc_zone_identifier_string
 
+    - name: Transform tags into list dict format for the modules that expect it
+      util_map:
+        function: zip_to_listdict
+        input: "{{ asg_instance_tags }}"
+        args: ['key', 'value']
+      register: listdict_asg_instance_tags
+
+    - debug: msg="Instance Tags:{{ listdict_asg_instance_tags }}"
+
     - name: Manage ASG
       ec2_asg:
         profile: "{{ profile }}"
         region: "{{ aws_region }}"
         name: "{{ asg_name }}"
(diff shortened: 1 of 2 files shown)


feanil:
@edx/devops

playbooks/edx-east/edx_service.yml
```diff
@@ -163,19 +163,27 @@
...
+
+    - debug: msg="Instance Tags:{{ listdict_asg_instance_tags }}"
+
     - name: Manage ASG
       ec2_asg:
         profile: "{{ profile }}"
         region: "{{ aws_region }}"
         name: "{{ asg_name }}"
         launch_config_name: "{{ service_config.name }}"
-        load_balancers: "{{ elb_name }}"
```
maxrothman, line 172: Why'd you remove this line?

e0d:
:+1:  we need to write tests for this at some point, can you add a ticket.

maxrothman:
The documentation gives this example:

```
 tags:
    - environment: production
      propagate_at_launch: no
```

As I understand it, you're producing this format:

```
tags:
  - environment: production
  - propegate_at_launch: no
```

Is the documentation incorrect?

playbooks/edx-east/edx_service.yml
feanil, line 172: It's in there twice.

maxrothman, line 172: :+1: I see.

feanil:
Ticket created: https://openedx.atlassian.net/browse/OPS-1117

@maxrothman The input I'm taking is not currently in the format you are using.

It's in the format documented in example of when you would use this functien.

maxrothman:
@feanil I'm not sure I got that. Can you explain further?

feanil:
Explained offline.

PR #2411, opened 2015-10-26
We don't use these plugins.

feanil:
They cost us build time and make the new-relic role more complicated.

@e0d @benpatterson


benpatterson:
SGTM. Didn't realize the logwatcher was not used. 
:+1:

e0d:
:+1:

PR #2414, opened 2015-10-27
Feanil/programs pre supervisor

feanil:
@edx/devops


playbooks/roles/supervisor/templates/etc/init/pre_supervisor.conf.j2
```diff
@@ -5,4 +5,10 @@ task
...
+{% if programs_code_dir is defined %}
```
feanil, line 8: If we like this approach, I'm happy to put in a separate PR to pull out the rest of the services to their own blocks instead of a single un-readable line.

playbooks/roles/supervisor/files/pre_supervisor_checks.py
```diff
@@ -156,8 +166,8 @@ def edp_for_instance(instance_id):
                 play=play,
                 instance_id=instance_id)
             break
-        except:
-            print("Failed to get EDP for {}".format(instance_id))
+        except Exception as e:
+            print("Failed to get EDP for {}: {}".format(instance_id, str(e)))
```
jibsheet, line 170: :fireworks:

playbooks/roles/supervisor/templates/etc/init/pre_supervisor.conf.j2
jibsheet, line 8: I find this somewhat more readable and would review a later PR to change it.

jibsheet:
:+1:

PR #2418, opened 2015-10-28
simple example to introduce terraform

e0d:
Opening as a simple example for @benpatterson and @jzoldak.  This is just basic scaffolding, but you get the idea.

Terraform is more flexible that cloudformation and has more modules than Ansible for infrastructure provisioning.


e0d:
@benpatterson this is ready to go?

benpatterson:
Ready to go!

When the script is executed, we will need the keys that are produced for the new user.

feanil:
:+1:

PR #2419, opened 2015-10-28
update to latest ansible

fredsmith:


fredsmith:
is what I would have said if not for:

```
error: pathspec 'stable-1.9.4-rc1-edx' did not match any file(s) known to git
```

PR #2423, opened 2015-10-29
Django 1.8 support

symbolist:


feanil:
@symbolist Is this PR ready for review? 

Also, `SECRET_KEY` is already set here: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L660

symbolist:
@feanil Not yet, still have to try out a few things. And I'll remove the extra `SECRET_KEY` in this PR.

singingwolfboy:
This does not work properly. Failure is here: https://gist.github.com/singingwolfboy/5cbe13e305528dae7fc9

symbolist:
@singingwolfboy I created a new sandbox with the branch and am able to run `/edx/bin/update` without any failures. The branch seems to be out of sync on your system, we no longer have a task `syncdb and migrate` only `migrate`: https://github.com/edx/configuration/blob/django18-support/playbooks/roles/edxapp/tasks/service_variant_config.yml#L106.

@feanil Needs a re-review. Django 1.8 expects a non-empty key.

singingwolfboy:
@symbolist I just tried again, and got the same error. Is this a problem with the `cypress-devstack` box file that Vagrant uses? Maybe I need to build a devstack from scratch, instead of via our usual method...

symbolist:
@singingwolfboy I doubt. When building the sandbox, where are you specifying the configuration branch?

singingwolfboy:
@symbolist I modified the `CONFIG_VER` variable in the Vagrantfile to refer to `django18-support` rather than `master`. https://github.com/edx/configuration/blob/master/vagrant/release/devstack/Vagrantfile#L35

symbolist:
@singingwolfboy That is strange. Do you get the error on `syncdb` when using master? Master does not have that anymore either: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/tasks/service_variant_config.yml#L106

singingwolfboy:
@symbolist I just finished building a devstack from scratch using Django 1.8, and it worked. I think it is a problem with the `cypress-devstack` box file. :(

symbolist:
Already fixed in master.

PR #2425, opened 2015-10-29
Terraform for SQS and SNS

benpatterson:
Taking @e0d's example and implementing OPS-1064

@jzoldak @feanil feedback welcome. If I could draw your attention specifically to the IAM policy and assignments, a look at that in particular would be appreciated.


benpatterson:
Also note I am removing the tfvars file and replacing it with tfvars.example. A user would need to supply his or her own tfvars.

feanil:
If I'm reading it correctly, it seems like we're giving the topics the ability to publish to themselves?  Shouldn't they have permissions to publish to the queue?  Also does it make sense to have two IAM roles so that the provision queue and sitespeed queue can't be published to from the wrong topic?

benpatterson:
@feanil Yeah that doesn't make sense. In terms of the SQS endpoint, that's being established through the `aws_sns_topic_subscription` task, so we should be good from that end. Updated.

terraform/sitespeed/sitespeed.tf
```diff
@@ -1,37 +1,47 @@
...
+	name = "${var.queue_name_pipeline}"
 	delay_seconds = "${var.queue_delay_seconds}"
 	max_message_size = "${var.queue_max_message_size}"
 	message_retention_seconds = "${var.queue_message_retention_seconds}"
 	receive_wait_time_seconds = "${var.queue_receive_wait_time_seconds}"
 }
 
-resource "aws_sns_topic" "build_requests" {
-	name = "user-updates-topic"
+resource "aws_sns_topic" "edx-pipeline-provision" {
```
feanil, line 19: This queue has the same name as the topic above.  We should change it otherwise we won't be able to refer to both of them.

benpatterson, line 19: I think it would be ok because of how terraform does some variable localization. But, even so, it's confusing. I've updated it.

benpatterson:
@feanil @jzoldak Ok I have refactored the iam/policy stuff. We now have a dedicated user that can publish to either queue. (The heroku app is built with one user in mind to manage filtering publications to the right place.)

Note that the user is build with its own iam policy, rather than a role. I did not see a way to assign a user to a role in terraform. I think you could technically assign a role-policy. I did not try it though, since that would just create its own copy of the policy as opposed to actually assigning the role.

benpatterson:
Sorry just another comment, when the user is created, the keys are not echoed; however, they are accessible by running `terraform show`. (They are saved on disk.) Whenever this is eventually run by devOps, those secrets will need to be shared so we can finish setting up the publication side.

feanil:
Do we not need a user that can publish to the `edx-pipeline-provision-topic` topic?  I only see an IAM policy for the Sitespeed User.

benpatterson:
The user being created has two 'resource' stanzas in the JSON policy. I'm thinking that'll do it.

benpatterson:
@feanil @e0d We were able to validate much of this in our aws account, and it appears to set everything up correctly.

This is a PR against ed's branch...is that still appropriate? Anything else need to get done for this change?

e0d:
:+1: to merge onto the branch, open to suggestions where this ultimately belongs

benpatterson:
For where this ultimately lives, I wonder if another repo is the right way to go?

PR #2427, opened 2015-10-29
Bump version of mongo to 3.0.7

jibsheet:
We can't use the 3.0.7 tools (mongodump/mongorestore) because of bugs in
the restore on gridfs and even draft mongo modulestore.
https://jira.mongodb.org/browse/TOOLS-939

This does not affect the mongo shell version

@edx/devops


feanil:
:+1:

PR #2431, opened 2015-10-30
Fixed programs sandbox configuration issues

clintonb:
Fixed programs sandbox configuration issues
- Creating DNS entry
- Specifying correct ports
- Enabling nginx site for CI builds
- Resolved nginx default site issues
- Fixed issue with OAuth client creation
- Setting LMS URL
- Corrected ecommerce repo

ECOM-2755


clintonb:
FYI @jimabramson @rlucioni 

@feanil @jibsheet How do I resolve the issue of both ecommerce and programs being set as default nginx sites? We override `nginx_default_sites` in `edx_provision.yml`; but, I am guessing the value is being overridden again when we run the programs and ecommerce playbooks. Any suggestions?

playbooks/roles/launch_ec2/tasks/main.yml
```diff
@@ -78,7 +78,7 @@
     value: "{{ item[0].public_dns_name }}"
   with_nested:
     - ec2.instances
-    - ['studio', 'ecommerce', 'preview']
+    - ['studio', 'ecommerce', 'preview', 'programs']
```
rlucioni, line 81: Will this change the DNS name for the service in production, too?

jimabramson:
@clintonb there are no logs in my sandbox for programs (created a few days ago, updated today).   also, there was no trusted oauth2 client set up.   are those things that can/should be addressed in this PR too?

clintonb:
@jimabramson I just added a fix for the OAuth issue.

playbooks/roles/launch_ec2/tasks/main.yml
clintonb, line 81: Nope.

rlucioni:
:+1:

clintonb:
Nginx issues have been resolved. @edx/devops please review.

clintonb:
@fredsmith

playbooks/roles/oauth_client_setup/defaults/main.yml
```diff
@@ -30,7 +30,7 @@ oauth_client_setup_oauth2_clients:
         secret: "{{ INSIGHTS_OAUTH2_SECRET }}" 
       }
     - {
-        name: "{{ PROGRAMS_OAUTH2_APP_CLIENT_NAME | default('None') }}",
+        name: "{{ programs_service_name | default('None') }}",
```
jimabramson, line 33: the name 'programs' is hardcoded in the app; should it be that way here too?  any other value won't work without manual modifications.

jimabramson:
one question but otherwise LGTM

playbooks/roles/oauth_client_setup/defaults/main.yml
clintonb, line 33: The variable is used to determine if the programs service is available. If you make a sandbox without programs, `programs_service_name` will be `None` and the OAuth 2.0 client will not be created.

jimabramson, line 33: got it, thanks.

util/jenkins/ansible-provision.sh
```diff
@@ -172,6 +178,10 @@ rabbitmq_refresh: True
 COMMON_HOSTNAME: $dns_name
 COMMON_DEPLOYMENT: edx
 COMMON_ENVIRONMENT: sandbox
+
+nginx_default_sites:
```
feanil, line 182: Isn't this the default?  Why override it here?

clintonb, line 182: When `run_ansible` is called on line 312 below, there doesn't seem to be a default value set for `nginx_default_sites`. This results in each role's value being used. I suspect this is why both the ecommerce and programs nginx configs were being set with each service as the default site, preventing nginx from starting.

feanil:
:+1:

PR #2436, opened 2015-11-02
Install edxlocal debian packages

jzoldak:
@benpatterson @feanil

### DIFF
--- a/playbooks/roles/edxlocal/tasks/main.yml
+++ b/playbooks/roles/edxlocal/tasks/main.yml
@@ -3,7 +3,8 @@
 ---
 
 - name: install packages needed for single server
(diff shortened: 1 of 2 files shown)


benpatterson:
LGTM :+1:

playbooks/roles/edxlocal/tasks/main.yml
```diff
@@ -2,6 +2,9 @@
 #
 ---
 
+- name: install packages needed for single server
+  apt: pkg={{','.join(edxlocal_debian_pkgs)}} install_recommends=yes state=present
```
e0d: This idiom should go away in favor of with_items, which does the right thing.

benpatterson: @e0d Updated in both places.

feanil:
:+1:

PR #2437, opened 2015-11-03
Killing celery processes shouldn't depend on being interactive

nedbat:


nedbat:
@feanil The old code only killed celery processes if it was run interactively, which seems wrong.  I made it run all the time, was it supposed to only be run if the target is cypress?

feanil:
@nedbat this was written as a workaround to a bug in the version of celery in Birch.  That version specifically got into a state where it was stuck running but not picking up any work and wouldn't restart.  So in cypress we wanted to actually re-start it after we upgraded it.

nedbat:
OK, adjusted to be conditional on cypress.  Look OK @feanil ?

feanil:
:+1:

PR #2441, opened 2015-11-03
normal bash history

fredsmith:


playbooks/roles/user/templates/default.bashrc.j2
```diff
@@ -96,13 +96,17 @@ if [ "$BASH_VERSINFO" -ge 2 ]; then
     shopt -s checkhash
     shopt -s no_empty_cmd_completion
     shopt -s execfail
+    export HISTCONTROL=ignorespace
+    unset HISTFILESIZE
+    export HISTFILESIZE
+    unset HISTSIZE
+    export HISTSIZE
```
feanil: Why unset and then export?

fredsmith: I was under the mistaken belief that unless you exported the variable, it wouldn't unassign it in the user's environment, but after some testing I've shown that belief to be wrong.

feanil:
:+1:

PR #2442, opened 2015-11-04
E0d/munin node

e0d:
@edx/devops role to install and configure munin-node


playbooks/roles/munin_node/meta/main.yml
```diff
@@ -0,0 +1,20 @@
+---
```
feanil: nit: Do we need this meta file considering it's empty?

feanil:
One nit but :+1: whether or not you decide to fix it.

playbooks/roles/munin_node/meta/main.yml
e0d: Was debating add common.

feanil: I think this role has no need for common from what I saw so I don't see a need to add it unless we want to make this fully edx specific. I could go either way but can't personally think of a good reason to include common for this role.

PR #2445, opened 2015-11-04
sitespeed terraform pipeline fix perms

benpatterson:
@jzoldak


benpatterson:
As per discussion, I see the new changes and they LGTM @jzoldak

PR #2448, opened 2015-11-06
Mongo modules for getting cluster status and managing cluster config.

feanil:
@edx/devops

### DIFF
--- a/playbooks/library/mongo_rs_member
+++ b/playbooks/library/mongo_rs_member
@@ -1,0 +1,269 @@
+#!/usr/bin/env python
+
+DOCUMENTATION = """
+---
+module: mongo_rs_member
+short_description: Modify replica set config for a member.
+description:
+  - Modify/Add/Remove members from a replica set.  Member management as done by rs.reconfig().
+version_added: "1.9"
+author: Feanil Patel
+options:
+  rs_host:
+    description:
+      - The hostname or ip of a server already in the mongo cluster.
+    required: false
+    default: 'localhost'
+  rs_port:
+    description:
+      - The port to connect to mongo on.
+    required: false
+    default: 27017
+  host:
+    description:
+      - The hostname of the member we want to modify.
+    required: false
+    default: 'localhost'
+  port:
+    description:
+      - The port of the member we want to modify.
+    required: false
+    default: 27017
+  username:
+    description:
+      - The username of the mongo user to connect as.
+    required: false
+  password:
+    description:
+      - The password to use when authenticating.
+    required: false
+  priority:
+    description:
+      - The priority of the member in the replica set.  Ignored if
+        `hidden` is set to true.
+    required: false
+  hidden:
+    description:
+      - Whether or not the member is hidden.
+    required: false
+  state:
+    description:
+      - Whether or not the member exists in the replica set. The membere
+        will be added or removed to reach this state.
+    required: false
+"""
+
+EXAMPLES = '''
+- name: Get status for the stage cluster
+  mongo_rs_member:
+    rs_host: some.mongo
+    rs_port: 27017
+    host: localhost
+    port: 27017
+    username: root
+    password: password
+  register: mongo_config
+'''
+# Magic import
+from ansible.module_utils.basic import *
+
+try:
+    from pymongo import MongoClient
+    from pymongo.errors import OperationFailure
+    from bson import json_util
+except ImportError:
+    pymongo_found = False
+else:
+    pymongo_found = True
+
+import json
+
+def get_mongo_uri(host, port, username, password):
+    mongo_uri = 'mongodb://'
+    if username and password:
+        mongo_uri += "{}:{}@".format(username,password)
+
+    mongo_uri += "{}:{}".format(host,port)
+    return mongo_uri
+
+def get_replset(module, client):
+    # Get the current config using `replSetGetConfig`
+    rs_config = client.admin.command("replSetGetConfig")
+    if 'config' not in rs_config:
+        module.fail_json(msg="Failed to get replset config from {}".format(primary_host), response=rs_config)
+    rs_config = rs_config['config']
+    return rs_config
+
+def reconfig_replset(module, client, rs_config):
+    # Update the config version
+    try: 
+        client.admin.command("replSetReconfig", rs_config)
+    except OperationFailure as e:
+        raise
+        module.fail_json(msg="Failed to reconfigure replSet: {}".format(e.message))
+
+def primary_client(module, some_host, some_port, username, password):
+    """
+    Given a member of a replica set, find out who the primary is
+    and provide a client that is connected to the primary for running
+    commands.
+    """
+    mongo_uri = get_mongo_uri(some_host, some_port, username, password)
+    client = MongoClient(mongo_uri)
+    try:
+        status = client.admin.command("replSetGetStatus")
+    except OperationFailure as e:
+        module.fail_json(msg="Failed to get replica set status from host({}): {}".format(some_host, e.message))
+
+    # Find out who the primary is.
+    rs_primary = filter(lambda member: member['stateStr']=='PRIMARY', status['members'])[0]
+    primary_host, primary_port = rs_primary['name'].split(':')
+
+    # Connect to the primary if this is not the primary.
+    if primary_host != some_host or primary_port != some_port:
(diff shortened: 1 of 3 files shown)


playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+    description:
+      - The hostname or ip of a server already in the mongo cluster.
+    required: false
+    default: 'localhost'
+  rs_port:
+    description:
+      - The port to connect to mongo on.
+    required: false
+    default: 27017
+  host:
```
e0d, line 22: Should this module allow management of the whole rs rather than be host oriented?  Input would be a configuration block rather than a host.

feanil, line 22: I considered that as well, and it wouldn't be hard to make that module, but this was so that there is an easy way to modify a member easily.  I think it's useful to be able to do this easily without having to know what the current state of the config is.  A module that maps the config from code to the servers would also be useful and wouldn't be heard to write now that I've written some of these utility methods.  Maybe we can put a small ticket on the backlog for it?

maxrothman, line 22: One way to handle it might be to treat membership in a replset as an idempotent state, so you could do things like say "this host is absent from the replset" or "these three hosts are present in the replset". That would give you the ability to manage 0-N members, and seems more idiomatic to me.

You already have most of the basic structure there, you'd just need to take a list of dicts for `host` instead of having `host` and `port`.

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+    description:
+      - The hostname of the member we want to modify.
+    required: false
+    default: 'localhost'
```
maxrothman, line 26: Would having `rs_host` and `host` both be localhost ever work? If not, I don't think it's a good thing to have a default for.

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+      - The priority of the member in the replica set.  Ignored if
+        `hidden` is set to true.
+    required: false
+  hidden:
+    description:
+      - Whether or not the member is hidden.
+    required: false
+  state:
+    description:
+      - Whether or not the member exists in the replica set. The membere
```
maxrothman: Typo: `membere`

maxrothman:
Where is this module intended to be run? In `local_action`?

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+  priority:
+    description:
+      - The priority of the member in the replica set.  Ignored if
+        `hidden` is set to true.
+    required: false
+  hidden:
+    description:
+      - Whether or not the member is hidden.
+    required: false
+  state:
```
maxrothman, line 49: What are the options for `state`? What happens if you don't specify one?

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+    password = module.params.get('password')
+    priority = module.params.get('priority')
+    hidden = module.params.get('hidden')
+    state = module.params.get('state')
+    changed = False
+
+    # Ignore priority when member should be hidden.
+    # Hidden members have to be set to priority 0.
+    if hidden:
+       priority = 0
```
maxrothman, line 180: Is this what `priority` is in mongo when a member is hidden?

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+
+    existing_member_names = [ member['host'] for member in rs_config['members'] ]
+    new_member_name = "{}:{}".format(host,port)
+
+    # See if member is already in the replica set
+    if new_member_name in existing_member_names:
+        # Make sure its config is the same.
+        current_settings = filter(lambda member: member['host'] == new_member_name, rs_config['members'])[0]
```
maxrothman, line 189: A comment here indicating that this is a pointer to a dict inside of `rs_config`, and thus when we modify it, we're modifying `rs_config` would be useful in addition to the comment below.

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+
+def reconfig_replset(module, client, rs_config):
+    # Update the config version
+    try: 
+        client.admin.command("replSetReconfig", rs_config)
+    except OperationFailure as e:
+        raise
+        module.fail_json(msg="Failed to reconfigure replSet: {}".format(e.message))
+
+def primary_client(module, some_host, some_port, username, password):
```
maxrothman, line 109: Nitpick: why pass `module` and the rest of the args? Aren't they all in `module.params`?

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+        module.fail_json(msg="Failed to get replica set status from host({}): {}".format(some_host, e.message))
+
+    # Find out who the primary is.
+    rs_primary = filter(lambda member: member['stateStr']=='PRIMARY', status['members'])[0]
+    primary_host, primary_port = rs_primary['name'].split(':')
+
+    # Connect to the primary if this is not the primary.
+    if primary_host != some_host or primary_port != some_port:
+        new_uri = get_mongo_uri(primary_host, primary_port, username, password)
+        client = MongoClient(new_uri)
```
maxrothman, line 130: You should probably explicitly close the old client.

playbooks/library/mongo_rs_member
```diff
@@ -0,0 +1,269 @@
...
+    rs_config = rs_config['config']
+    return rs_config
+
+def reconfig_replset(module, client, rs_config):
+    # Update the config version
+    try: 
+        client.admin.command("replSetReconfig", rs_config)
+    except OperationFailure as e:
+        raise
+        module.fail_json(msg="Failed to reconfigure replSet: {}".format(e.message))
```
maxrothman, line 107: If you `raise`, will this line ever be run?

maxrothman:
Nitpick: given that the only existing mongo module is named `mongodb_user`, these should probably be named `mongodb_status` and `mongodb_rs_member`.

playbooks/library/mongo_status
```diff
@@ -0,0 +1,92 @@
...
+  - Get the status of a mongo cluster. Provide the same info as rs.status()
+version_added: "1.9"
+author: Feanil Patel
+options:
+  host:
+    description:
+      - The hostname or ip of a server in the mongo cluster. It can include a port number.
+    required: false
+    default: 'localhost'
+  port:
```
maxrothman, line 17: Why support `host: host:port` and `host: host, port: port`? As far as I can tell, only the latter is supported in `mongo_rs_member`.

playbooks/library/mongo_rs_member
feanil, line 26: It would work if you are changing the priority or hiddenness of the local machine and it's already connected to a replica set.

playbooks/library/mongo_rs_member
feanil, line 180: Yes. It has to be 0.

playbooks/library/mongo_rs_member
feanil, line 109: I didn't want to get them out in every case since it's quite verbose.  Gonna stick with this for now even though it is not consistent in all the code, since I think it maximizes readability.

feanil:
@edx/devops I've addressed all comments, can I get a thumb?

playbooks/library/mongo_status
```diff
@@ -0,0 +1,92 @@
...
+  password:
+    description:
+      - The password to use when authenticating.
+    required: false
+"""
+
+EXAMPLES = '''
+- name: Get status for the stage cluster
+  mongo_status:
+    host: localhost:27017
```
maxrothman, line 35: This no longer matches the above spec

maxrothman:
@feanil 1 new comment and looking for a reply on https://github.com/edx/configuration/pull/2448#discussion-diff-44156358R49 (where I was mostly looking for documentation). After that, :+1:

playbooks/library/mongo_rs_member
feanil, line 49: @maxrothman I added those a few lines below here in the comments.

maxrothman, line 49: :+1:

PR #2322, opened 2015-09-30
Allow enabling HTTP basic auth for individual services.

smarnach:
Currently, there is a single flag COMMON_ENABLE_BASIC_AUTH to decide whether to
enable basic auth in the nginx config.  This change allows to enable basic auth
selectively for individual services.

I tested this locally targetting a virtual machine and using variations of this playbook:

```

---
- hosts: all
  sudo: True
  gather_facts: True
  vars:
    COMMON_ENABLE_BASIC_AUTH: True
    EDXAPP_LMS_ENABLE_BASIC_AUTH: False
  roles:
    - role: nginx
      nginx_sites:
      - cms
      - lms
      nginx_default_sites:
      - lms
    - edxapp
```

PR #2450, opened 2015-11-06
Revert ff670ea8c14f0a57bab827d6b3c5f98ce17a1b40

fredsmith:
Reverts edx/configuration#2322

This breaks edx's internal sandbox provisioning.  We at least need to fix internal scripts, but it looks like we may need to put more safeties around the defaults here.


jibsheet:
:+1:

PR #2451, opened 2015-11-06
Add a swapfile role to edx_sandbox playbook

bradenmacdonald:
We (OpenCraft) are running into occasional out-of-memory issues with sandboxes on some VMs.

This is a simple import of [Kamal Nasser's ansible `swapfile` role](https://github.com/kamaln7/ansible-swapfile) (MIT licensed), which will enable a swapfile for the `edx_sandbox` playbook by default. It is enabled by default with `SWAPFILE_SIZE = 2GB` but can easily be disabled by setting `SWAPFILE_SIZE = false`

The role is unmodified from the upstream version except that a couple variables are capitalized differently to fit better with edX ansible conventions.

Test instructions: You can test this on any VM by running:

```
ansible-playbook -i localhost, -c local run_role.yml -e role=swapfile -e "SWAPFILE_SIZE=2GB"
```

**Project / Deadline**: This is an OpenCraft project  with no specific deadline and is just a community contribution.

**Questions for edX reviewer**:
- Should this be enabled or disabled by default?
- Any comments on importing an external ansible galaxy role like this? Should I keep the `galaxy_info` or not?

### DIFF
--- a/playbooks/roles/swapfile/defaults/main.yml
+++ b/playbooks/roles/swapfile/defaults/main.yml
@@ -1,11 +1,22 @@
 ---
-# To enable the swapfile role, set this variable using a MB or GB suffix
-# e.g. SWAPFILE_SIZE: 512MB
-# To disable the role, leave the default value of 'false'
-SWAPFILE_SIZE: false
+# Size of the desired swap file. Use a MB or GB suffix.
+SWAPFILE_SIZE: 2GB
+# SWAPFILE_LOCATION: Location of the swap file managed by this role.
+SWAPFILE_LOCATION: /swapfile
+# SWAPFILE_USE_DD: set this to True if your filesystem does not support 
+# fallocate (e.g. if you use ext3). dd will then be used instead of fallocate.
+SWAPFILE_USE_DD: False
 
+#
+# Advanced, optional settings: 
+#
 
-swapfile_location: /swapfile
-swapfile_swappiness: False
-swapfile_vfs_cache_pressure: False
-swapfile_use_dd: False
+# SWAPFILE_SWAPPINESS: Update sysctl.conf to set the swappiness percentage
+# (vm.swappiness) -- the lower it is, the less your system swaps memory pages.
+# If this is False (default), no change will be made.
+SWAPFILE_SWAPPINESS: False
+# SWAPFILE_VFS_CACHE_PRESSURE: Update sysctl.conf to set the VFS cache pressure.
+# "this percentage value controls the tendency of the kernel to reclaim the
+#  memory which is used for caching of directory and inode objects."
+# If this is False (default), no change will be made.
+SWAPFILE_VFS_CACHE_PRESSURE: False
--- a/playbooks/roles/swapfile/defaults/main.yml
+++ b/playbooks/roles/swapfile/defaults/main.yml
@@ -1,0 +1,11 @@
+---
+# To enable the swapfile role, set this variable using a MB or GB suffix
+# e.g. SWAPFILE_SIZE: 512MB
+# To disable the role, leave the default value of 'false'
+SWAPFILE_SIZE: false
+
+
+swapfile_location: /swapfile
+swapfile_swappiness: False
+swapfile_vfs_cache_pressure: False
+swapfile_use_dd: False
--- a/playbooks/roles/swapfile/meta/main.yml
+++ b/playbooks/roles/swapfile/meta/main.yml
@@ -1,0 +1,10 @@
+---
+galaxy_info:
+  author: "Kamal Nasser"
+  description: swapfile
+  license: MIT
+  min_ansible_version: 1.4
+  version: 0.4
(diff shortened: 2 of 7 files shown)


playbooks/roles/swapfile/meta/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+galaxy_info:
```
mtyaka, line 2: Does `galaxy_info` serve any purpose now that the role has been modified and pulled into this repository or should we just remove it?

mtyaka:
You might want to remove `galaxy_info`, but other than that this looks good and works for me :+1:

playbooks/roles/swapfile/meta/main.yml
bradenmacdonald, line 2: No, I just left `galaxy_info` to preserve the author/credit information in its original form.

e0d:
@nedbat @sarina is there any issue bringing in new licenses for specific components?

playbooks/roles/swapfile/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
+---
+- name: Write swapfile
+  command: |
+    {% if swapfile_use_dd %}
```
e0d: Not a huge fan of jinja conditionals in the task, especially considering when is also in play.

playbooks/roles/swapfile/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+# To enable the swapfile role, set this variable using a MB or GB suffix
+# e.g. SWAPFILE_SIZE: 512MB
+# To disable the role, leave the default value of 'false'
+SWAPFILE_SIZE: false
```
e0d: Why overload this?  A size of 0 could be disabled or there could be a separate var for enabling.  Practically speaking I don't think we'll often include this role in a play without enabling swap.

playbooks/roles/swapfile/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+
+
+swapfile_location: /swapfile
```
e0d: should be upper case I think.

playbooks/roles/swapfile/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+swapfile_swappiness: False
```
e0d: should document acceptable values and consider whether these should also be upper case.

nedbat:
@e0d It's fine from a license point of view to include MIT-licensed code into an AGPL project.

playbooks/roles/swapfile/tasks/main.yml
bradenmacdonald: I just left that as it was in the original/upstream version. I can change it if you want, but then we lose the ability to easily incorporate improvements. Since it's such a small role, I'm on the fence about whether it matters that we retain the ability to update it with upstream improvements or not.

playbooks/roles/swapfile/defaults/main.yml
bradenmacdonald: That's just how the author chose to implement it. If you want, I can just leave the implementation as is (for compatibility with upstream) but change the default and remove the documentation around the `false` value. As you say, people can always enable/disable it via the play.

playbooks/roles/swapfile/defaults/main.yml
bradenmacdonald: Ok, I pushed a commit with more documentation and uppercase vars.

e0d:
@edx/devops second look from someone.

playbooks/roles/swapfile/tasks/main.yml
```diff
@@ -0,0 +1,37 @@
...
+  command: mkswap {{ SWAPFILE_LOCATION }}
+  register: create_swapfile
+  when: SWAPFILE_SIZE != false and write_swapfile.changed
+
+- name: Enable swapfile
+  command: swapon {{ SWAPFILE_LOCATION }}
+  when: SWAPFILE_SIZE != false and create_swapfile.changed
+
+- name: Add swapfile to /etc/fstab
+  lineinfile: dest=/etc/fstab line="{{ SWAPFILE_LOCATION }}   none    swap    sw    0   0" state=present
```
feanil: Use the `mount` module instead of lineinfile to insure the configuration is in `/etc/fstab`

bradenmacdonald: @feanil Ok, changed.

feanil:
:+1:

bradenmacdonald:
Thanks @feanil. @e0d you didn't explicitly give +1 - is this ok to merge?

e0d:
:+1:

bradenmacdonald:
@smarnach @mtyaka You may want to know that this has been merged.

PR #2453, opened 2015-11-09
Add a postfix queue role

bradenmacdonald:
This new optional ansible role will enable edxapp instances to send emails instantly to a postfix server that runs on localhost, and then queues messages and forwards them to a third party SMTP service. This is usually a faster and more robust setup than getting edxapp to connect directly to the third party SMTP server.

We've tested this role on a client instance with https://www.authsmtp.com/ as the external SMTP server.

**Usage Instructions**:
1. Add this ansible role to the playbook that you're using.
2. Set these new variables in the secure configuration:
   
   ```
   POSTFIX_QUEUE_EXTERNAL_SMTP_HOST: mail.authsmtp.com
   POSTFIX_QUEUE_EXTERNAL_SMTP_PORT: 2525
   POSTFIX_QUEUE_EXTERNAL_SMTP_USER: username_here
   POSTFIX_QUEUE_EXTERNAL_SMTP_PASSWORD: password_here
   ```
3. _unset_ all of the `EDXAPP_EMAIL_` variables in the secure configuration.

**Author Notes**
I have not added this to any existing playbooks - I could do so with a `when:` condition if that would be preferable.

### DIFF
--- a/playbooks/roles/postfix_queue/defaults/main.yml
+++ b/playbooks/roles/postfix_queue/defaults/main.yml
@@ -1,0 +1,30 @@
+---
+
+# postfix_queue: Configure a local postfix server to forward mail to an
+# external SMTP server. This way postfix acts as an outgoing mail queue, and
+# web apps can send mail instantly, while still taking advantage of an
+# external SMTP service.
+#
+# The external service is assumed to use TLS.
+#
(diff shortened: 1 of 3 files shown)


mtyaka:
:+1:

feanil:
:+1:   @fredsmith second review

fredsmith:
:+1:

PR #2454, opened 2015-11-09
Update the ad-hoc-reporting mongo script.

feanil:
If the cluster has a hidden secondary prefer that over a regular secondary.

@edx/devops


jibsheet:
:+1:

PR #2455, opened 2015-11-09
Add a key prefix to the config model.

feanil:
Add a prefix from the hostname.  If building images, this will be the hostname
of the machine that the image is based on.  If running in a rolling fashion,
this will give each machine its own cache.

@edx/devops

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -807,6 +807,9 @@ generic_env_config:  &edxapp_generic_env
     staticfiles:
       <<: *default_generic_cache
       KEY_PREFIX: "{{ ansible_hostname|default('staticfiles') }}_general"
+    configuration:
+      <<: *default_generic_cache
+      KEY_PREFIX: "{{ ansible_hostname|default('configuration') }}"
     celery:
       <<: *default_generic_cache
       KEY_PREFIX:  'celery'


jibsheet:
:+1:

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -807,6 +807,9 @@ generic_env_config:  &edxapp_generic_env
     staticfiles:
       <<: *default_generic_cache
       KEY_PREFIX: "{{ ansible_hostname|default('staticfiles') }}_general"
+    configuration:
+      <<: *default_generic_cache
+      KEY_PREFIX: "{{ ansible_hostname|default('configuration') }}"
```
cpennington, line 812: I'm confused. This is using the host-name for the prefix? Won't that change on a per-machine basis in the cluster? Or is it the hostname of the AMI originator?

Honestly, I'd rather this be based on the edx-platform version number, so that it doesn't flush the cache if the code hasn't changed.

feanil, line 812: It's the version of the machine that built the AMI, it gives us a snapshot of all content on the machine.

maxrothman, line 812: :+1: to @cpennington's comment. We have the edx-platform version in a var already, and it seems a lot easier to reason about than the IP of the machine that became the AMI. Also, that would give the prefix semantic value for debugging.

PR #2458, opened 2015-11-10
[Jenkins-master] Use Java 7 with custom ARGS.

benpatterson:


benpatterson:
@jzoldak @edx/devops please take a look.

playbooks/roles/jenkins_master/tasks/main.yml
```diff
@@ -38,6 +38,18 @@
 - name: stop Jenkins
   service: name=jenkins state=stopped
 
+- name: set jvm args
+  lineinfile: >
+    backup=yes
```
feanil: This should use the new style syntax like we do [here](https://github.com/edx/configuration/pull/2458/files#diff-d887fe1442a4be97ae2d3183ac49489eR28).

playbooks/roles/jenkins_master/meta/main.yml
```diff
@@ -1,4 +1,5 @@
 ---
 dependencies:
   - common
-  - oraclejdk
+  - role: oraclejdk
+    tags: java
```
feanil, line 5: Why tag here instead of in the oraclejdk role itself?

benpatterson, line 5: Tagging here means I can run jenkins set-up, but only java stuff. So, in the case of dependencies, we don't run common. (Does that answer your question?)

playbooks/roles/jenkins_master/tasks/main.yml
benpatterson: Done. Pls take another look.

benpatterson:
@feanil ready for another pass

playbooks/roles/jenkins_master/meta/main.yml
feanil, line 5: If you tag in the oraclejdk role that should do the same, but this is a much smaller change to get you to do what you need so I think that's fine.

playbooks/roles/jenkins_master/tasks/main.yml
```diff
@@ -38,6 +38,18 @@
 - name: stop Jenkins
   service: name=jenkins state=stopped
 
+- name: set jvm args
+  lineinfile:
+    backup: yes
+    dest: /etc/default/jenkins
+    regexp: '^JAVA_ARGS'
```
feanil: Is there only one var that starts with `JAVA_ARGS`?  Theres not other vars like `JAVA_ARGS_THREADING`?  I would add an `=` in the regex just to be safe.

feanil:
One nit in the regex if you think it's worth fixing, but :+1: even if you don't want to fix it.

jzoldak:
:+1:

feanil:
:+1:

benpatterson:
Fixed the regex, merging. Thanks guys.

PR #2464, opened 2015-11-11
Arbab/rewrite common

arbabnazar:
@edx/owners, kindly review it, try to rewite the common role with new style and combined the similar tasks into one. Thanks


playbooks/roles/common/tasks/main.yml
```diff
@@ -1,121 +1,99 @@
...
-  apt_repository: repo="{{ common_git_ppa }}"
+  apt_repository: 
+    repo: "{{ common_git_ppa }}"
 
 - name: Install role-independent useful system packages
   # do this before log dir setup; rsyslog package guarantees syslog user present
-  apt: >
-    pkg={{','.join(common_debian_pkgs)}} install_recommends=yes
-    state=present update_cache=yes
+  apt:
```
e0d, line 27: We can and should use with_items instead of join.  It will do the install in a single transaction and that has been the case for a while.

feanil:
:+1:

e0d:
:+1:

PR #2469, opened 2015-11-12
Remove the xml modulestore from the list of modulestore that cat called.

feanil:
@edx/devops


jibsheet:
:+1:

maxrothman:
FYI typo in commit message

PR #2471, opened 2015-11-12
Add SUPPORT_SITE_LINK setting.

peter-fogg:
Public side of https://openedx.atlassian.net/browse/DEVOPS-3238.

@edx/devops


maxrothman:
Is an empty string a sane working default? Also, is that var lms- or cms-specific? If so, then it should go in a more specific env_config.

peter-fogg:
@maxrothman The empty string is the correct default; it will result in the support site link not being shown in the LMS footer. This is currently only displayed in the LMS, but it may be used in Studio in the future, so I think it makes sense for it to remain in the `generic_env_config`.

maxrothman:
:+1:

PR #2477, opened 2015-11-13
Fix for PayPal config.  Add missing error_url setting.

mjfrey:
@clintonb please take a look


clintonb:
:+1: 

FYI @feanil @jibsheet

jibsheet:
:+1:

PR #2478, opened 2015-11-13
Run the version of ansible set by CONFIG_VER

feanil:
First we update ansible then we run the playbook for the devstack.

@edx/devops


e0d:
:+1:

PR #2480, opened 2015-11-14
extending the hack

e0d:
@cpennington works for now.


docker/build/course-discovery/Dockerfile
```diff
@@ -7,6 +7,10 @@ ARG repo_owner=edx
 
 ENV ansible_playbook=/edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook
 
+USER root
+RUN apt-get -y install libmysqlclient-dev
+RUN pip install MySQL-python
```
cpennington, line 12: Should this be part of the `trusty-common` base build which install ansible?

PR #2076, opened 2015-06-03
E0d/common refactor

e0d:
@jibsheet @fredsmith @feanil 

Finally, getting around to what was discussed here:

https://github.com/edx/configuration/issues/1760

PR #2482, opened 2015-11-15
E0d/refactor utils

e0d:
@edx/devops 

I'd like to move the rando apt dependencies that support interactive debugging to  the own role.

Phase two of this is adding this role explicitly to plays rather than making common dependent on it.

This has been re-based on https://github.com/edx/configuration/pull/2076, which should be merged first.

### DIFF
--- a/playbooks/roles/common/meta/main.yml
+++ b/playbooks/roles/common/meta/main.yml
@@ -1,6 +1,7 @@
 ---
 dependencies:
   - common_vars
+  - server_utils
   - role: user
     user_info: "{{ COMMON_USER_INFO }}"
   - role: security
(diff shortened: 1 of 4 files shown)


e0d:
@edx/devops rebased, ready for review.

feanil:
Should this role also remove the install section of common?  https://github.com/edx/configuration/blob/master/playbooks/roles/common/tasks/main.yml#L29-L36

That installs packages defined [here](https://github.com/edx/configuration/blob/master/playbooks/roles/common_vars/defaults/main.yml#L89-L106), which overlap with the packages listed here.

e0d:
@feanil the idea is that the overlap would be removed, could be part of the this PR.  Some packages will remain common, but there's a need for things like tmux and screen, but someone should know what they're getting.

feanil:
:+1:

e0d:
@feanil one more quick look, I've removed the duplicated dependencies.

feanil:
:+1:

PR #2483, opened 2015-11-16
E0d/inventory

e0d:
@edx/devops 

This ends up being way harder than it could or should be. What's more, some things are missing and hard to get, like the the SHA of the config repo.

Still, this is something that's been requested and is useful outside of AWS.  Thoughts on ways to improve?


feanil:
This needs to be rebased from master it looks like.

PR #2485, opened 2015-11-16
quote, fixing, unquote, bad plays

e0d:
@maxrothman @jzoldak 

I feel these plays really should not be doing this, but this fix will unblock JZ.


maxrothman:
ಠ_ಠ :+1:

maxrothman:
@e0d can you make a story to fix this long-term?

jzoldak:
:+1: thx @e0d

PR #2487, opened 2015-11-17
Fix the HTTP-to-HTTPS redirect on sandboxes

arbabnazar:
@edx/devops , can you please check this change, so I can modify the other nginx configs as well. Thanks

P.S: "COMMON_ENVIRONMENT" variable is taken from the [ansible-provision.sh](https://github.com/edx/configuration/blob/master/util/jenkins/ansible-provision.sh#L180).

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -180,6 +180,11 @@ location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
   {% include "robots.j2" %}
   {% include "static-files.j2" %}
 
+ {% if NGINX_SET_X_FORWARDED_HEADERS %}
+   if ($scheme != "https") {
+    rewrite ^ https://$host$uri permanent;
+   }
+ {% else %}
   # Forward to HTTPS if we're an HTTP request...
   if ($http_x_forwarded_proto = "http") {
     set $do_redirect "true";
@@ -189,4 +194,5 @@ location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
   if ($do_redirect = "true") {
     rewrite ^ https://$host$request_uri? permanent;
   }
+ {% endif %}
 }
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
@@ -91,6 +91,11 @@ error_page {{ k }} {{ v }};
   {% include "robots.j2" %}
   {% include "static-files.j2" %}
 
+  {% if NGINX_SET_X_FORWARDED_HEADERS %}
+   if ($scheme != "https") {
+    rewrite ^ https://$host$uri permanent;
+   }
+ {% else %}
   # Forward to HTTPS if we're an HTTP request...
   if ($http_x_forwarded_proto = "http") {
     set $do_redirect "true";
@@ -100,5 +105,5 @@ error_page {{ k }} {{ v }};
   if ($do_redirect = "true") {
     rewrite ^ https://$host$request_uri? permanent;
   }
-
+ {% endif %}
 }
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/ecommerce.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/ecommerce.j2
@@ -53,7 +53,12 @@ location @proxy_to_app {
     proxy_pass http://ecommerce_app_server;
   }
 
-# Forward to HTTPS if we're an HTTP request...
+{% if NGINX_SET_X_FORWARDED_HEADERS %}
+   if ($scheme != "https") {
+    rewrite ^ https://$host$uri permanent;
+   }
+ {% else %}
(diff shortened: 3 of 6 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -180,6 +180,11 @@ location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
   {% include "robots.j2" %}
   {% include "static-files.j2" %}
 
+ {% if COMMON_ENVIRONMENT|lower == 'sandbox' %}
```
feanil: You should key based on the `NGINX_SET_X_FORWARDED_HEADERS` like we do above.  If the X_FORWARDED headers are not seat we should use 'scheme'.

arbabnazar:
@feanil please second look. Thanks

Can I do the same for cms, insights, program etc?

e0d:
:+1:

feanil:
@arbabnazar :+1:   Please do this for the other NGINX profiles as well.

smarnach:
It looks like this patch broke OpenCraft's sandboxes.

With this change applied, it looks like all requests will be redirected to HTTPS, even when `NGINX_ENABLE_SSL` is set to `False` (its default value).  Is this intentional?  If so, what's the meaning of `NGINX_ENABLE_SSL`?

When accessing a sandbox with `NGINX_ENABLE_SSL` set to `False`, we get immediately redirected to HTTPS, but nginx isn't even listening on port 443, so we can't access that sandbox at all.

arbabnazar:
@smarnach can you please make the value of `NGINX_SET_X_FORWARDED_HEADERS` to `false` and re-configure you sandbox for verification. Thanks,

smarnach:
Since the sandbox isn't behind any other reverse proxy, setting `NGINX_SET_X_FORWARDED_HEADERS` to `False` isn't what I want.  It's likely that it will fix the issue with the sandbox being completely inaccessible, since it will essentially recover the old behaviour, but we'd end up with empty X-Forwarded-… headers, which will break other things.

smarnach:
@arbabnazar I tried redeploying with `NGINX_SET_X_FORWARDED_HEADERS: false`, but the sandbox still isn't reachable at all (which is not what I expected).

I still don't understand why redirecting to HTTPS isn't done conditionally based on `NGINX_ENABLE_SSL`.

arbabnazar:
@smarnach  we didn't disable the SSL, in this patch we just fixed one thing by checking that the system is behind the ELB or not and based on this we configure the Nginx redirect.

smarnach:
@arbabnazar I understand that you didn't change the default value of `NGINX_ENABLE_SSL`.  The problem is that the new version redirects to HTTPS even when SSL is disabled, which obviously doesn't work, since nginx won't even be listening for HTTPS connections in that case.  I agree the old version was buggy, too, but at least it was buggy in a way that didn't break things for us. :)

smarnach:
This is still a problem for us.  I'd appreciate further insight. :)

I could also ask on the openedx-ops mailing list, if you think that would be more appropriate.

e0d:
Looking at this move closely I have a couple thoughts.
- I don't like that fact that logic is split across jinja and nginx config conditionals, too much indirection.
- I think we would be better off introducing a new. explicit var, say, `NGINX_REDIRECT_TO_SECURE_TRANSPORT` rather than glean from other headers what the redirect behavior should be.

@edx/devops thoughts here?

feanil:
@e0d I think that makes sense since in the case to separate whether or not we should redirect from whether or not we should enable SSL in NGINX.   Since in the case where we have an ELB, we still want NGINX to redirect to HTTPS but we don't want to enable SSL via NGINX.

smarnach:
@e0d @feanil Thanks, that makes sense to me.  We need two different redirect behaviours depending on whether we are behind ELB or not, so if we make `NGINX_REDIRECT_TO_SECURE_TRANSPORT` a simple Boolean flag, we still need to check other settings to decide which redirect behaviour to apply.

smarnach:
One more question.  Are you going to fix this soon?  We fixed it for us by reverting this PR, but we currently have to cherry-pick the revert commit in various places.

smarnach:
Friendly ping. :)  It would be good to know whether you plan to fix the HTTPS redirect.

e0d:
We'll try to get this into our next sprint.  We do support two different re-direct behaviors today in a way that works well for us.  The key difference seems to be that you can't or don't use SSL on your sandboxes.  Correct me if I'm wrong about that.  We will fix this, because I think it's generally a good idea, but wouldn't either a snake oil cert or an expensive Gandi cert also unblock you?

We'll track the work here:

https://openedx.atlassian.net/browse/OPS-1208

smarnach:
@e0d Thanks for your reply, and for scheduling work on this.  I can see that using HTTPS would fix the problem, and I'm personally fine if you intend to only support HTTPS going forward.  However, my impression was that this PR broke support for plain HTTP accidentally.  At the moment, running the [one-step installation](https://github.com/edx/configuration/wiki/edX-Ubuntu-12.04-64-bit-Installation) with the default configuration results in an unusable server, which I presume isn't what you intended. :)

When I tried the fall-back to the snake oil certificate, I found it to be broken as well, since it doesn't use the correct path to the snake oil certificate that comes with Ubuntu.  I didn't debug this any further, and I know I can easily generate my own snake oil certificate, but since we can live with our current work-around for now, I first wanted to know what your plans are before spending any more time on this issue.

PR #2489, opened 2015-11-17
Close all caches in gunicorn postfork

nedbat:


nedbat:
@feanil @adampalay @doctoryes opinions?

doctoryes:
The code looks fine. :+1: 

But do we know enough about what each of these caches are? Are they all self-building? Or is there an explicit cache building step for any of them?

nedbat:
cache.close isn't discarding any information, it's closing the connection to the cache.

jbarciauskas:
I'd vote for explicitly closing the cache that we know is the new one, and implementing a generic solution later.

playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -140,3 +140,23 @@ common_redhat_variants:
...
+    # Close the cache so that newly forked workers cannot accidentally share
+    # the socket with the processes they were forked from. This prevents a race
+    # condition in which one worker could get a cache response intended for
+    # another worker.
+    # We do this in a way that is safe for 1.4 and 1.8 while we still have some
+    # 1.4 installations.
+    from django.conf import settings
+    from django.core import cache as django_cache
+    if hasattr(django_cache, 'caches'):
+        get_cache = django_cache.caches.__getitem__
```
adampalay: To be clear, this is because django will deprecate `get_cache` at some point?

adampalay:
:+1:

playbooks/roles/common_vars/defaults/main.yml
nedbat: Yes, and in 1.8, there is already a warning that 1.9 will get rid of get_cache.

doctoryes:
@macdiesel  FYI

macdiesel:
Since this is a post fork process and we are only closing the cache connections (new ones will be made in the code later) this looks good to me. :+1:

feanil:
Tested on sandboxes. :+1:

bradenmacdonald:
This PR seems to be breaking devstack provisioning for me - any idea why?

```
TASK: [edxapp | add gunicorn configuration files] ***************************** 
fatal: [localhost] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'common_close_all_caches' is undefined", 'failed': True}
fatal: [localhost] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'common_close_all_caches' is undefined", 'failed': True}]}

```

I was able to work around it by explicitly making `edxapp` depend on `common_vars` but that seems like it shouldn't be necessary.

nedbat:
@bradenmacdonald something like this happened while we were getting this right, and it turned out to be something simple, but now I can't remember what it was.  Or I'm misremembering... @feanil?

e0d:
@bradenmacdonald is this happening on `vagrant up`?

bradenmacdonald:
@e0d yes it was during vagrant up.

PR #2491, opened 2015-11-17
Course Discovery (includes a docker devstack)

cpennington:
This builds on hackathon work to enable docker to build a new IDA that uses docker for its devstack.

This should be rebased after #2499, #2500, and #2501 are reviewed and merged.

### DIFF
--- a/docker/build/course_discovery/Dockerfile
+++ b/docker/build/course_discovery/Dockerfile
@@ -1,0 +1,29 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/course_discovery/Dockerfile .
+#
+# This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
+# with the currently checked-out configuration repo.
+
+FROM edxops/trusty-common
+MAINTAINER edxops
+
+ARG COURSE_DISCOVERY_VERSION=master
+ARG REPO_OWNER=edx
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+USER docker
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+COPY docker/build/course_discovery/ansible_overrides.yml /
+RUN sudo ansible-playbook course_discovery.yml -c local \
+    -t 'install:base,install:code,install:system-requirements,install:app-requirements,install:configuration,install:vhosts,install:devstack' \
+    --extra-vars="@/ansible_overrides.yml" \
+    --extra-vars="COURSE_DISCOVERY_VERSION=$COURSE_DISCOVERY_VERSION" \
+    --extra-vars="COMMON_GIT_PATH=$REPO_OWNER"
+
+USER root 
+CMD ["/edx/app/supervisor/venvs/supervisor/bin/supervisord", "-n", "--configuration", "/edx/app/supervisor/supervisord.conf"]
--- a/docker/plays/course_discovery.yml
+++ b/docker/plays/course_discovery.yml
@@ -1,0 +1,13 @@
+- name: Deploy Course Discovery
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    serial_count: 1
+  serial: "{{ serial_count }}"
+  roles:
+    - docker
+    - nginx
+    - role: course_discovery
+      nginx_default_sites:
+        - course_discovery
--- a/playbooks/roles/course_discovery/defaults/main.yml
+++ b/playbooks/roles/course_discovery/defaults/main.yml
@@ -1,0 +1,143 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role course_discovery
+# 
+COURSE_DISCOVERY_GIT_IDENTITY: !!null
+
+# depends upon Newrelic being enabled via COMMON_ENABLE_NEWRELIC
+# and a key being provided via NEWRELIC_LICENSE_KEY
+COURSE_DISCOVERY_NEWRELIC_APPNAME: "{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-{{ course_discovery_service_name }}"
+COURSE_DISCOVERY_PIP_EXTRA_ARGS: "-i {{ COMMON_PYPI_MIRROR_URL }}"
+COURSE_DISCOVERY_NGINX_PORT: 18381
+COURSE_DISCOVERY_SSL_NGINX_PORT: 48381
+
+COURSE_DISCOVERY_DEFAULT_DB_NAME: 'course_discovery'
+COURSE_DISCOVERY_MYSQL: 'db'
+# MySQL usernames are limited to 16 characters
+COURSE_DISCOVERY_MYSQL_USER: 'course_discov001'
+COURSE_DISCOVERY_MYSQL_PASSWORD: 'password'
+
+COURSE_DISCOVERY_DATABASES:
+  # rw user
+  default:
+    ENGINE: 'django.db.backends.mysql'
+    NAME: '{{ COURSE_DISCOVERY_DEFAULT_DB_NAME }}'
+    USER: '{{ COURSE_DISCOVERY_MYSQL_USER }}'
+    PASSWORD: '{{ COURSE_DISCOVERY_MYSQL_PASSWORD }}'
+    HOST: '{{ COURSE_DISCOVERY_MYSQL }}'
+    PORT: '3306'
+    ATOMIC_REQUESTS: true
+    CONN_MAX_AGE: 60
+
+COURSE_DISCOVERY_DB_ADMIN_USER: 'root'
+COURSE_DISCOVERY_DB_ADMIN_PASSWORD: ''
+
+COURSE_DISCOVERY_MEMCACHE: [ 'memcache' ]
+
+COURSE_DISCOVERY_CACHES:
+  default:
+    BACKEND:  'django.core.cache.backends.memcached.MemcachedCache'
+    KEY_PREFIX: 'default'
+    LOCATION: '{{ COURSE_DISCOVERY_MEMCACHE }}'
+
+COURSE_DISCOVERY_VERSION: "master"
+COURSE_DISCOVERY_DJANGO_SETTINGS_MODULE: "course_discovery.settings.production"
+COURSE_DISCOVERY_URL_ROOT: 'http://course_discovery:18381'
+COURSE_DISCOVERY_OAUTH_URL_ROOT: 'http://127.0.0.1:8000'
+
+COURSE_DISCOVERY_SECRET_KEY: 'Your secret key here'
+COURSE_DISCOVERY_TIME_ZONE: 'UTC'
+COURSE_DISCOVERY_LANGUAGE_CODE: 'en-us'
+
+# Used to automatically configure OAuth2 Client
+COURSE_DISCOVERY_SOCIAL_AUTH_EDX_OIDC_KEY : 'course_discovery-key'
+COURSE_DISCOVERY_SOCIAL_AUTH_EDX_OIDC_SECRET : 'course_discovery-secret'
+COURSE_DISCOVERY_SOCIAL_AUTH_REDIRECT_IS_HTTPS: false
+
+COURSE_DISCOVERY_PLATFORM_NAME: 'Your Platform Name Here'
+
+COURSE_DISCOVERY_SERVICE_CONFIG:
+  SECRET_KEY: '{{ COURSE_DISCOVERY_SECRET_KEY }}'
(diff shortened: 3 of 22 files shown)


cpennington:
@e0d: Review?

cpennington:
Also, it seems like squashing many of these commits would be appropriate.

.dockerignore
```diff
@@ -0,0 +1,3 @@
+*
+!playbooks/
+!docker/
```
clintonb: Add a newline.

docker/build/course-discovery/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/course-discovery/Dockerfile .
+#
+# This allows the dockerfile to updated /edx/app/edx_ansible/edx_ansible
```
clintonb: updated -> update

.dockerignore
cpennington: Done.

docker/build/course-discovery/Dockerfile
cpennington: Done

e0d:
@cpennington planning to review tomorrow in the hack.  Noticed that there's a syntax issue your last commit, see the failed check.

docker/build/mysql/edxlocal.yml
```diff
@@ -0,0 +1,17 @@
+# Configure a Jenkins worker instance
```
e0d: where's this comment block from?

docker/plays/course_discovery.yml
```diff
@@ -0,0 +1,13 @@
...
+  hosts: all
+  sudo: True
+  gather_facts: True
+  vars:
+    serial_count: 1
+  serial: "{{ serial_count }}"
+  roles:
+    - docker
+    - role: course_discovery
+      nginx_default_sites:
```
e0d: Nginx is running on the CD container?

playbooks/filter_plugins/ports_plugin.py
```diff
@@ -0,0 +1,15 @@
+import hashlib
```
e0d: Can you provide a comment for the rationale for pulling this in and what needs to happen to remove it..

Oh wait, this is the proposed hash based generation of ports.  I'm not sure I favour this.  I think a more simplistic /etc/services approach makes more sense.

playbooks/roles/container/tasks/main.yml
```diff
@@ -0,0 +1,26 @@
+---
+#
```
e0d: we should delete before merging.

playbooks/roles/course_discovery/defaults/main.yml
```diff
@@ -0,0 +1,134 @@
...
+
+COURSE_DISCOVERY_CACHES:
+  default:
+    BACKEND:  'django.core.cache.backends.memcached.MemcachedCache'
+    KEY_PREFIX: 'default'
+    LOCATION: memcache
+
+COURSE_DISCOVERY_VERSION: "master"
+COURSE_DISCOVERY_DJANGO_SETTINGS_MODULE: "edx_course_discovery.settings.production"
+COURSE_DISCOVERY_URL_ROOT: 'http://localhost:8381'
```
e0d: I'd like to remove all localhost references to in favour of something like coursediscover.edx.

playbooks/roles/course_discovery/defaults/main.yml
```diff
@@ -0,0 +1,134 @@
...
+course_discovery_code_dir: "{{ course_discovery_home }}/{{ course_discovery_service_name }}"
+
+course_discovery_gunicorn_host: "127.0.0.1"
+course_discovery_gunicorn_port: 8381
+course_discovery_gunicorn_timeout: 300
+
+course_discovery_log_dir: "{{ COMMON_LOG_DIR }}/{{ course_discovery_service_name }}"
+
+course_discovery_requirements_base: "{{ course_discovery_code_dir }}/requirements"
+course_discovery_requirements:
```
e0d: Would love to remove this in favour of `make requrements`

playbooks/roles/course_discovery/tasks/main.yml
```diff
@@ -0,0 +1,236 @@
...
+- name: add gunicorn configuration file
+  template:
+    src: edx/app/course_discovery/course_discovery_gunicorn.py.j2
+    dest: "{{ course_discovery_home }}/course_discovery_gunicorn.py"
+  sudo_user: "{{ course_discovery_user }}"
+  tags:
+    - install
+    - install:configuration
+
+- name: add deadsnakes repository
```
e0d, line 33: @feanil thoughts on whether we should factor this out into  a role, there's not much to it, but we're likely to reuse a lot.

playbooks/roles/course_discovery/tasks/main.yml
```diff
@@ -0,0 +1,236 @@
...
+  apt_repository:
+    repo: "ppa:fkrull/deadsnakes"
+  tags:
+    - install
+    - install:system-requirements
+
+- name: install python3.5
```
e0d, line 40: why hardcode 3.5

docker/build/mysql/edxlocal.yml
cpennington: This was part of the pre-existing docker work. My container doesn't depend on it.

docker/plays/course_discovery.yml
cpennington: My goal was to have the docker devstack running production-style, as far as access to static assets and the like go, so, yes, I put nginx in the same container.

playbooks/filter_plugins/ports_plugin.py
cpennington: The goal here is to have fewer steps from running the IDA generation script to actually having a deployment up and running. Rather than having to search through and specify a bunch of ports, this simply creates them for you.

That said, I'd be ok with having the IDA creation script just ask for a port suffix as an argument.

playbooks/roles/container/tasks/main.yml
cpennington: Happily. No idea where that came from (I assume pre-existing docker work).

playbooks/roles/course_discovery/defaults/main.yml
cpennington: Ah, an excellent point.

playbooks/roles/course_discovery/defaults/main.yml
cpennington: Works for me.

playbooks/roles/course_discovery/tasks/main.yml
cpennington, line 33: My thinking was just that for now, the python3 stuff is a bit experimental (gevent and gunicorn don't yet have released/stable python 3 support). So, until it's more stable, I'd rather have it specific to course-discovery (that's also why it's not in the template code).

playbooks/roles/course_discovery/tasks/main.yml
cpennington, line 40: See my above note. This is experimental.

cpennington:
@e0d: I don't think the error is a syntax error that's breaking the build, it's that the `filter_plugins` directory isn't being found when jenkins runs. But I'm not sure why that is.

docker/plays/course_discovery.yml
e0d: Is there something more complex than proxying to the backend going on here?

cpennington: No, but that's partly since the app doesn't have any assets yet. I expect once we have assets, we'll want nginx to serve them.

playbooks/roles/course_discovery/tasks/main.yml
```diff
@@ -0,0 +1,236 @@
...
+  sudo_user: "{{ course_discovery_user }}"
+  with_items: "{{ course_discovery_requirements }}"
+  tags:
+    - install
+    - install:app-requirements
+
+- name: wait for database
+  wait_for: host=db port={{ COURSE_DISCOVERY_DATABASES.default.PORT }} delay=2
+  tags:
+    - migrate:devstack
```
e0d: maybe this should just be install:db.  There's not reason this could not be used to bootstrap any deployment of this IDA is there?

vagrant/base/devstack/Vagrantfile
```diff
@@ -8,32 +8,25 @@ VAGRANTFILE_API_VERSION = "2"
 MEMORY = 4096
```
e0d: Can you pull these changes into a separate PR?  We're going to need to tease this changeset apart generally.  I'd like to pull in the related pieces as discrete PRs and then have you rebase your PR on the new master.

playbooks/roles/course_discovery/tasks/main.yml
cpennington: Yeah, that would be fine too. I just wasn't sure whether we wanted this type of bootstrapping (and permissions) elsewhere, or preferred a different bootstrapping style.

vagrant/base/devstack/Vagrantfile
cpennington: Yes, I can do that.

docker/build/course_discovery/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+ARG course_discovery_version=master
+ARG repo_owner=edx
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+USER docker
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+RUN sudo ansible-playbook course_discovery.yml -c local -t 'install:base'
```
feanil: Why are you running all of these tags separately?  Is there a tag you're avoiding?  Also why not as a single command?

docker/compose/docker-compose.yml
```diff
@@ -0,0 +1,69 @@
+db:
+#  build: ../build/mysql
+  image: edxops/mysql:v1
+  environment:
+    - MYSQL_ROOT_PASSWORD=''
+
+mongo:
+  image: edxops/mongo:v2
+  volumes: 
+    - /var/devstack/mongo/data:/data
```
mrgnr: We might want to specify volume paths with [variables](https://docs.docker.com/compose/compose-file/#variable-substitution) for more flexibility.

cpennington: I think that's not going to be in this PR going forward.

docker/build/course_discovery/Dockerfile
```diff
@@ -0,0 +1,32 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/course-discovery/Dockerfile .
```
clintonb: course_discovery

clintonb:
Note: Getting reacquainted with Docker. We should work to get this on Docker Hub and add documentation for public consumption, once it's ready.

clintonb:
Docs? How do I run this?

playbooks/roles/course_discovery/defaults/main.yml
cpennington: Done.

docker/build/course_discovery/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ARG COURSE_DISCOVERY_VERSION=master
+ARG REPO_OWNER=edx
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+USER docker
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+RUN sudo ansible-playbook course_discovery.yml -c local \
+    -t 'install:base,install:code,install:system-requirements,install:app-requirements,install:configuration,install:vhosts,install:devstack' \
```
e0d: @cpennington I've started using this ansible_overrides.yml pattern in the build files that I think works pretty well.

https://github.com/edx/configuration/tree/e0d/docker-artifacts/docker/build/forums

The Dockerfile then uses the following:

```
COPY ansible_overrides.yml /
RUN sudo ansible-playbook forum.yml -i '127.0.0.1,' -c local -t "install:base,install:configuration,install:code" -e@/ansible_overrides.yml
```

cpennington: I actually don't like that, because it means that the results vary without busting the cache. If the overrides are coded in, either as ARGs or explicit values, then changing them will cause the docker file to rebuild. That's not the case for an external file.

e0d: Not true, changing the file will cause the copy to be re-run and the ansible command.

cpennington: Oh, I see. Yes, ok, that works then.

cpennington:
@e0d, @clintonb, @feanil: All comments addressed.

cpennington:
@e0d @feanil: Can I get a final review, now that all of the prior commits have been merged?

playbooks/roles/course_discovery/defaults/main.yml
feanil: Is there an update missing here? or are you going to do that separately?

cpennington: Oh, oops. Thought I had fixed that.

feanil:
I had one more comment on the `course_discovery_requirements` but :+1: either way.

PR #2494, opened 2015-11-18
adding the callback_plugin for track the time of each task

arbabnazar:
@edx/devops please review the callback plugin and help me to make it as per Ansible's standard. Thanks


playbooks/callback_plugins/tasks_timing.py
```diff
@@ -0,0 +1,80 @@
...
+        results = sorted(self.stats.items(), key=lambda value: value[1], reverse=True)
+
+        datadog.initialize(api_key='9775a026f1ca7d1c6c5af9d94d9595a4',
+                           app_key='87ce4a24b5553d2e482ea8a8500e71b8ad4554ff')
+
+        # Print the timing of each task
+        for name, elapsed in results:
+            # send the metric to datadog
+            datadog.api.Metric.send(
+                metric="edx.jenkins.{0}".format(name.replace(" | ", ".").replace(" ", "-").lower()),
```
e0d: I would propose edx.ansible.{0}.task_duration

playbooks/callback_plugins/tasks_timing.py
```diff
@@ -0,0 +1,80 @@
...
+                points=elapsed,
+                host='jenkins',
+            )
```
feanil: Datadog also allows us to send tags for a metric, if the EDC is available we should send each value as a tag.

http://docs.datadoghq.com/api/  Take a look at the section on posting 'TIME SERIES POINTS`

playbooks/callback_plugins/tasks_timing.py
feanil: +1 to this idea.  That way we can push more metrics in the future for a task without needing to change the data structure.

feanil: Another Note: Metric.send will by default mark the time the metric was sent to be time.now() but we probably don't want this because it will show all the events happening within a very small time slice.  We probably want to store the time when the job ended and use that as the time when the event occured.

feanil:
What was the reason for posting the metrics to datadog in the `playbook_on_stats` function instead of in the `playbook_on_task_start` function?  It seems like that would be better because then we can have more realtime data as well.

playbooks/callback_plugins/tasks_timing.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+
+        # Record the timing of the very last task, we use it here, because we don't have stop task function by default
+        if self.current_task is not None:
+            self.stats[self.current_task] = time.time() - self.stats[self.current_task]
+
+        # Sort the tasks by their running time
+        results = sorted(self.stats.items(), key=lambda value: value[1], reverse=True)
+
+        datadog.initialize(api_key='9775a026f1ca7d1c6c5af9d94d9595a4',
```
arbabnazar: @e0d @feanil can you please help me here that how we can pass the `api_key` and `app_key` as extra variable from our secure repo? Thanks

feanil: @arbabnazar the only pattern I've seen for this is environment variables.

https://github.com/edx/configuration/blob/master/playbooks/callback_plugins/hipchat_plugin.py#L54

If the environment variables are not set, you should not try to send data to datadog.

arbabnazar:
@feanil I have placed the metrics in the `playbook_on_stats` function because I have to write most the datadog steps in the `playbook_on_stats` function for last task as ansible doesn't have `playbook_on_task_stop`, if you want me, I can move the stuff into the `playbook_on_task_start` function. Thanks

e0d:
I think we should rename the plugin to indicated that this is a datadog specific implementation.

playbooks/callback_plugins/datadog_tasks_timing.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+
+        # Record the timing of the very last task, we use it here, because we don't have stop task function by default
+        if self.current_task is not None:
+            self.stats[self.current_task] = (time.time(), time.time() - self.stats[self.current_task])
+
+        # Sort the tasks by their running time
+        results = sorted(self.stats.items(), key=lambda value: value[1][1], reverse=True)
+
+        datadog_api_key = os.getenv('DATADOG_API_KEY')
```
feanil: We should initialize the datadog connection in **init** and then we can use it in both playbook_on_task_start and on playbook_on_stats as needed.

playbooks/callback_plugins/datadog_tasks_timing.py
```diff
@@ -0,0 +1,88 @@
...
+        for name, elapsed in results:
+            logger.info(
+                "{0:-<80}{1:->8}".format(
+                    '{0} '.format(name),
+                    ' {0:.02f}s'.format(elapsed[1]),
+                )
+            )
+
+        # Total time to run the complete playbook
+        total_seconds = sum([x[1][1] for x in self.stats.items()])
```
e0d: Let's send the total to datadog as well.

edx.ansible.play_duration

This code will need to move up.

e0d:
After the last comment I'm :+1:

feanil:
Looks like this got rebased oddly. It's showing 19 commits between master and this.  Not sure why.  :+1: to @arbabnazar 's actual changes but we should fix the branch before merging.

arbabnazar:
Close this PR in favor of that [PR](https://github.com/edx/configuration/pull/2546), because it has conflicts

e0d:
:+1:  After new rebase.

PR #2497, opened 2015-11-18
A base box with python 2.7.10 and django 1.8

feanil:
@edx/devops  @edx/django-team 

This should be merged once django 1.8 lands in master.


e0d:
:+1:

PR #2498, opened 2015-11-19
Maybe this makes the most sense.

e0d:


playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -161,3 +161,9 @@ common_close_all_caches: |
             cache = get_cache(cache_name)
             if hasattr(cache, 'close'):
                 cache.close()
+
+common_robot_rules: |
+  {% for item in NGINX_ROBOT_RULES %}
```
feanil, line 166: Doesn't this meen that the common_vars role would have to depend on the nginx role?  Do we need an `if NGINX_ROBOT_RULES is defined` check?

e0d, line 166: No, this var would get updated to be something provided by services.

feanil, line 166: Seems like we need to check to see if it's defined then. Since a service might fail to define it.

PR #2501, opened 2015-11-20
Move django role template to docker

cpennington:
This sets up the django IDA template for docker, with no other changes.

### DIFF
--- a/playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
+++ b/playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
@@ -9,27 +9,42 @@
 # and a key being provided via NEWRELIC_LICENSE_KEY
 {{ role_name|upper }}_NEWRELIC_APPNAME: "{{ '{{ COMMON_ENVIRONMENT }}' }}-{{ '{{ COMMON_DEPLOYMENT }}' }}-{{ '{{' }} {{ role_name }}_service_name }}"
 {{ role_name|upper }}_PIP_EXTRA_ARGS: "-i {{ '{{ COMMON_PYPI_MIRROR_URL }}' }}"
-{{ role_name|upper }}_NGINX_PORT: XXX       # TODO Change this!!!
-{{ role_name|upper }}_SSL_NGINX_PORT: XXX   # TODO Change this!!!
+{{ role_name|upper }}_NGINX_PORT: 18{{ port_suffix }}
+{{ role_name|upper }}_SSL_NGINX_PORT: 48{{ port_suffix }}
 
 {{ role_name|upper }}_DEFAULT_DB_NAME: '{{ role_name }}'
+{{ role_name|upper }}_MYSQL: 'db'
+# MySQL usernames are limited to 16 characters
+{{ role_name|upper }}_MYSQL_USER: '{{ role_name[:13] }}001'
+{{ role_name|upper }}_MYSQL_PASSWORD: 'password'
 
 {{ role_name|upper }}_DATABASES:
   # rw user
   default:
     ENGINE: 'django.db.backends.mysql'
     NAME: '{{ '{{' }} {{ role_name|upper }}_DEFAULT_DB_NAME }}'
-    USER: '{{ role_name }}001'
-    PASSWORD: 'password'
-    HOST: 'localhost'
+    USER: '{{ '{{' }} {{ role_name|upper }}_MYSQL_USER }}'
+    PASSWORD: '{{ '{{' }} {{ role_name|upper }}_MYSQL_PASSWORD }}'
+    HOST: '{{ '{{' }} {{ role_name|upper}}_MYSQL }}'
     PORT: '3306'
     ATOMIC_REQUESTS: true
     CONN_MAX_AGE: 60
 
+{{ role_name|upper }}_DB_ADMIN_USER: 'root'
+{{ role_name|upper }}_DB_ADMIN_PASSWORD: ''
+
+{{ role_name|upper }}_MEMCACHE: [ 'memcache' ]
+
+{{ role_name|upper }}_CACHES:
+  default:
+    BACKEND:  'django.core.cache.backends.memcached.MemcachedCache'
+    KEY_PREFIX: 'default'
+    LOCATION: '{{ '{{' }} {{ role_name|upper}}_MEMCACHE }}'
+
 {{ role_name|upper }}_VERSION: "master"
 {{ role_name|upper }}_DJANGO_SETTINGS_MODULE: "{{ role_name }}.settings.production"
-{{ role_name|upper }}_URL_ROOT: 'http://localhost:####'   # TODO Set port number!
-{{ role_name|upper }}_LMS_URL_ROOT: 'http://127.0.0.1:8000'
+{{ role_name|upper }}_URL_ROOT: 'http://{{ role_name }}:18{{ port_suffix }}'
+{{ role_name|upper }}_OAUTH_URL_ROOT: 'http://127.0.0.1:8000'
 
 {{ role_name|upper }}_SECRET_KEY: 'Your secret key here'
 {{ role_name|upper }}_TIME_ZONE: 'UTC'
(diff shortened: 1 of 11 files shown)


cpennington:
@e0d: Review?

playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
```diff
@@ -19,17 +19,23 @@
   default:
     ENGINE: 'django.db.backends.mysql'
     NAME: '{{ '{{' }} {{ role_name|upper }}_DEFAULT_DB_NAME }}'
-    USER: '{{ role_name }}001'
+    USER: '{{ role_name[:13] }}001'
```
feanil: Add a comment for why you're doing this.

playbooks/roles/ansible-role-django-ida/templates/docker/build/ROLE_NAME/Dockerfile.j2
```diff
@@ -0,0 +1,32 @@
...
+#
+# docker build -f docker/build/course-discovery/Dockerfile .
+#
+# This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
+# with the currently checked-out configuration repo.
+
+FROM edxops/trusty-common
+MAINTAINER edxops
+
+ARG {{ role_name }}_version=master
```
feanil: This should be `{{ role_name| upper}}`_VERSION`to match the settings in`defaults/main.yml`

playbooks/roles/ansible-role-django-ida/templates/docker/build/ROLE_NAME/Dockerfile.j2
```diff
@@ -0,0 +1,32 @@
...
+ARG repo_owner=edx
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+USER docker
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+RUN sudo ansible-playbook {{ role_name }}.yml -c local -t 'install:base'
```
feanil: This should be consolidated to one ansible-playbook call.

playbooks/roles/ansible-role-django-ida/templates/docker/plays/ROLE_NAME.yml.j2
```diff
@@ -0,0 +1,13 @@
...
+  gather_facts: True
+  vars:
+    serial_count: 1
+  serial: "{{ '{{' }} serial_count }}"
+  roles:
+    - docker
+    - role: {{ role_name }}
+      nginx_default_sites:
+        - {{ role_name }}
+    - nginx
```
feanil: NGINX sholud be installed before the role since the role relies on nginx folders existing.

playbooks/roles/ansible-role-django-ida/templates/tasks/main.yml.j2
```diff
@@ -34,6 +34,47 @@
...
     - install:app-requirements
 
+- name: wait for database
+  wait_for: host=db port={{ '{{' }} {{ role_name|upper }}_DATABASES.default.PORT }} delay=2
+  tags:
+    - migrate:devstack
+
+- name: create databases
+  mysql_db:
+    login_host: "db"
```
feanil: These should come from vars we can easily override to specify different DB locations.  Var name along the lines of `{{ role_name}}_admin_user`

cpennington:
@feanil: Comments addressed. Any more to add?

playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
```diff
@@ -19,17 +19,29 @@
   default:
     ENGINE: 'django.db.backends.mysql'
     NAME: '{{ '{{' }} {{ role_name|upper }}_DEFAULT_DB_NAME }}'
-    USER: '{{ role_name }}001'
+    # MySQL usernames are limited to 16 characters
+    USER: '{{ role_name[:13] }}001'
     PASSWORD: 'password'
-    HOST: 'localhost'
+    HOST: 'db'
```
feanil: This should be pulled out into a variable.  `{{ role_name|upper }}_MYSQL_HOST`

playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
```diff
@@ -19,17 +19,29 @@
...
+{{ role_name|upper }}_DB_ADMIN:
+  default:
+    USER: 'root'
+    PASSWORD: ''
+
+{{ role_name|upper }}_CACHES:
+  default:
+    BACKEND:  'django.core.cache.backends.memcached.MemcachedCache'
+    KEY_PREFIX: 'default'
+    LOCATION: memcache
```
feanil: memcache url should be a variable with a list of one item in it.

playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
cpennington: Why? This _is_ the variable that everything else uses.

playbooks/roles/ansible-role-django-ida/templates/templates/edx/app/nginx/sites-available/ROLE_NAME.j2
```diff
@@ -0,0 +1,70 @@
+#
+# {{ '{{' }} ansible_managed }}
+#
+
+
+{{ '{%' }} if "{{ role_name }}" in nginx_default_sites {{ '%}' }}
```
feanil: Using `nginx_default_sites` here without defining a default in the role means we can't use the role in a standalone manner.   This var should have a default or a check to see if it's defined here.  I lean towards a check to see if it's defined because the var name should be changed if the var has a default in this role.

playbooks/roles/ansible-role-django-ida/templates/tasks/main.yml.j2
```diff
@@ -24,16 +24,69 @@
...
+    login_user: "{{ '{{' }} {{ role_name|upper }}_DB_ADMIN.default.USER }}"
+    login_password: "{{ '{{' }} {{ role_name|upper }}_DB_ADMIN.default.PASSWORD }}"
+    db: "{{ '{{' }} {{ role_name|upper }}_DEFAULT_DB_NAME }}"
+    state: present
+    encoding: utf8
+  tags:
+    - migrate:devstack
+
+- name: create database users
+  mysql_user:
```
feanil, line 66: Use with_items instead of copying this twice.

playbooks/roles/ansible-role-django-ida/templates/tasks/main.yml.j2
```diff
@@ -24,16 +24,69 @@
...
+    login_host: "{{ '{{' }} {{ role_name|upper }}_DATABASES.default.HOST }}"
+    login_user: "{{ '{{' }} {{ role_name|upper }}_DB_ADMIN.default.USER }}"
+    login_password: "{{ '{{' }} {{ role_name|upper }}_DB_ADMIN.default.PASSWORD }}"
+    name: "{{ '{{' }} {{ role_name|upper }}_DATABASES.default.USER }}"
+    host: "%"
```
feanil: This should be configurable and the default should only refer to the database it needs access to.

playbooks/roles/ansible-role-django-ida/templates/defaults/main.yml.j2
feanil: In production the hostname for the DB will not be `db` and we need to be able to override it, the username and the password.  All three should be extracted to variables.

cpennington: So the goal is that every configurable thing be a single-value variable?

playbooks/roles/ansible-role-django-ida/templates/tasks/main.yml.j2
cpennington: Sorry, you mean that the host that the user is associated with should be configurable? We won't know what hosts should have access until run-time, though, right? So it needs to be `%`.

playbooks/roles/ansible-role-django-ida/templates/tasks/main.yml.j2
cpennington, line 66: Done.

cpennington:
@feanil: All comments addressed.

feanil:
:+1:

PR #2503, opened 2015-11-20
Install a global mongorc.js

jibsheet:
Right now, users need to type rs.slaveOk() after running a read-replica
mongo script.  This avoids that.  It is compatible with anyone who also
has a ~/.mongorc.js

@edx/devops


feanil:
:+1:

PR #2505, opened 2015-11-20
Update various versions to match build-jenkins.

benpatterson:
@jzoldak 

Mainly recordkeeping but with an eye towards making things properly match.

### DIFF
--- a/playbooks/roles/jenkins_master/defaults/main.yml
+++ b/playbooks/roles/jenkins_master/defaults/main.yml
@@ -4,59 +4,79 @@ jenkins_group: "edx"
 jenkins_server_name: "jenkins.testeng.edx.org"
 jenkins_port: 8080
 
-jenkins_version: "1.630"
+jenkins_version: "1.638"
 jenkins_deb_url: "http://pkg.jenkins-ci.org/debian/binary/jenkins_{{ jenkins_version }}_all.deb"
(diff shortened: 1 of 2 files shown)


playbooks/roles/jenkins_master/defaults/main.yml
```diff
@@ -4,59 +4,79 @@ jenkins_group: "edx"
...
-    - { name: "xunit", version: "1.89"}
-    - { name: "git-client", version: "1.19.0"}
-    - { name: "git", version: "2.4.0"}
+    - { name: "windows-slaves", version: "1.0" }
+    - { name: "xunit", version: "1.93"}
 
 jenkins_bundled_plugins:
     - "credentials"
     - "git"
+    - "pam-auth"
```
feanil, line 79: What is this needed for?

benpatterson, line 79: The bundled_plugins are the pinned ones, and I added this once since I noticed it was pinned, too. It had already been in the fuller list above.

jzoldak:
:+1:

playbooks/roles/jenkins_master/defaults/main.yml
benpatterson, line 79: @feanil GTG?

benpatterson:
@feanil I'm going ahead and merging, but can change the pam-auth setting later if you think that's better.

PR #2506, opened 2015-11-21
Adding tags to forum role

e0d:
Cherry picked from hack/2015


playbooks/roles/forum/tasks/deploy.yml
```diff
@@ -45,6 +57,9 @@
   sudo_user: "{{ common_web_user }}"
   environment: "{{ forum_environment }}"
   notify: restart the forum service
+  tags:
+    - install
+    - install:code
```
feanil: this should be `install:app-requirements`

feanil:
:+1:  once the bundle tag is updated.

PR #2510, opened 2015-11-21
E0d/remove aws include

e0d:
@edx/devops 

This refactors role dependencies to the play  so they are very explicit.  I have mixed feelings about common.  I think if it were truly common I'd say it belongs here more.

Thoughts about this as a general pattern?  I'm thinking that the only place we would use meta includes for things that are more likely inheritance and plays for things that are more like composition.

### DIFF
--- a/playbooks/roles/edx_ansible/meta/main.yml
+++ b/playbooks/roles/edx_ansible/meta/main.yml
@@ -9,5 +9,5 @@
 #
 ##
 # Role includes for role edx_ansible
-dependencies:
-  - common
+
+dependencies: []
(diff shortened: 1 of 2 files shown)


playbooks/roles/edx_ansible/meta/main.yml
```diff
@@ -9,6 +9,5 @@
 #
 ##
 # Role includes for role edx_ansible
-dependencies:
-  - common
```
feanil, line 13: edx_ansible still needs common_vars directly so that it can build directories in the right place.

maxrothman:
IIRC, there was a desire to have roles be able to run independently. If `roles/edx_ansible` can run successfully on a bare machine without including `common`, then :+1:, but wouldn't it require the directory structure provided by `common`?

e0d:
@feanil confirm this satisfies your comment.

feanil:
:+1:

PR #2513, opened 2015-11-23
Upgrade ora2 playbook to expect Django 1.8 on platform.

benpatterson:


benpatterson:
@andy-armstrong here is the fix discussed in chat. I posted the job result there as well if you want a look.

andy-armstrong:
Thanks @benpatterson. @dianakhuang FYI

dianakhuang:
@benpatterson Closing this now that it's redundant.

PR #2515, opened 2015-11-23
Playbook to create mongo users

jibsheet:
This runs from the local machine and needs to know an IP of any member
of the cluster as well as the replica set.  We can add both of those to
the shim db/edxapp-mongo.yml file in the secure directories easily (see relevant PRs).

@edx/devops


feanil:
:+1:

maxrothman:
Given that this isn't part of our normal flow for provisioning mongo servers (e.g. run cloudmigrations, run mongo play), I wonder if this fits better in sysadmin.

fredsmith:
:+1:   - this is generally useful to the community and is an ansible play.  it belongs in configuration.

PR #2518, opened 2015-11-24
Django 1.8 migration changes for xqueue

jibsheet:
Now that XQueue's repo is running on 1.8, we need to use the new syntax
https://docs.djangoproject.com/en/1.8/ref/django-admin/#syncdb

@edx/devops


feanil:
:+1:

FremanZhang:
The same applies to migrate lms/cms
image

PR #2522, opened 2015-11-25
Add CSMH DB provisioning automation

maxrothman:
Also includes local version of RDS module (see commit message)

@edx/devops

### DIFF
--- a/playbooks/edx-east/create_rds_secondary.yml
+++ b/playbooks/edx-east/create_rds_secondary.yml
@@ -1,0 +1,34 @@
+# Usage: AWS_PROFILE=myprofile ansible-playbook create_csmh_db.yml -i localhost, -e 'from_db=my-rds-identifier rds_name=env-dep-edxapphistory'
+
+- name: Create new edxapp history RDS instance
+  hosts: all
+  connection: local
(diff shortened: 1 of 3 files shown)


playbooks/edx-east/create_csmh_db.yml
```diff
@@ -0,0 +1,41 @@
...
+
+#NB: to set in promoter script:
+#  multi-zone, username, password, size
+
+- name: Create new StudentModule RDS instance
+  hosts: all
+  connection: local
+  gather_facts: false
+  vars:
+    from_db:
```
feanil: nit: Can we explicitly set these to empty strings?

feanil:
:+1:

playbooks/edx-east/create_csmh_db.yml
maxrothman: They're currently `null`, not empty strings. If you prefer, I could set them to `null` explicitly.

feanil: Yes, please set it to `!!null` explicitly.

feanil:
:+1:

PR #2524, opened 2015-11-27
adding the tanaguru role

arbabnazar:
@edx/devops can you please review this role. Thanks


playbooks/roles/tanaguru/defaults/main.yml
```diff
@@ -0,0 +1,26 @@
+---
+tanaguru_prerequisites:
```
e0d: follow standard in other roles, tanguru_debian_pkgs.

playbooks/roles/tanaguru/defaults/main.yml
```diff
@@ -0,0 +1,26 @@
...
+  - openjdk-7-jre
+  - unzip
+  - libmysql-java
+  - python-mysqldb
+  - tomcat6 
+  - libspring-instrument-java
+  - xvfb
+  - postfix
```
e0d: I'd rather not install a mail server if we can avoid it.  Minimally, I'd like us to be able to use Google SMTP ourselves, but I think that should just work.

playbooks/roles/tanaguru/tasks/deploy.yml
```diff
@@ -0,0 +1,128 @@
+---
+- name: Add the Partner repository
+  apt_repository: 
+    repo: "{{ item }}"
+    state: present
+  with_items:
+    - "deb http://archive.canonical.com/ubuntu {{ ansible_distribution_release }} partner"
+    - "deb-src http://archive.canonical.com/ubuntu {{ ansible_distribution_release }} partner"
+
+- name: Set the Locale
```
e0d: This should probably be pulled up to common.  It would be suprising to include a role and have it override your locale.

playbooks/roles/tanaguru/tasks/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+- include: deploy.yml
```
e0d: I think we should just move all the tasks to main.yml and avoid this indirection.

playbooks/roles/tanaguru/defaults/main.yml
arbabnazar: @e0d that's just a dummy postfix configuration that they have mentioned on the prerequisites page.

playbooks/roles/tanaguru/tasks/deploy.yml
arbabnazar: @e0d [this](https://github.com/edx/configuration/blob/master/playbooks/roles/common/templates/sshd_config.j2#L77) is the same thing or still we need to add the locale to the common role?

cptvitamin:
@arbabnazar i tested this in your sandbox and everything was working great.

playbooks/roles/tanaguru/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
+---
+- name: Add the Partner repository
+  apt_repository: 
+    repo: "{{ item }}"
+    state: present
+  with_items:
+    - "deb http://archive.canonical.com/ubuntu {{ ansible_distribution_release }} partner"
+    - "deb-src http://archive.canonical.com/ubuntu {{ ansible_distribution_release }} partner"
```
e0d, line 8: I want to tag all the tasks with our role lifecycle tags from here: https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle
For example, this should be tagged with install and install-base.

playbooks/roles/tanaguru/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+
+- name: Download the latest TanaGuru tarball
+  get_url: 
+    url: "{{ tanaguru_download_link }}" 
+    dest: "/tmp/{{ tanaguru_download_link | basename }}"
+
+- name: Unzip the downloaded Firfox zipped file
+  unarchive: 
+    src: "/tmp/{{ tanaguru_download_link | basename }}" 
+    dest: "~/"
```
e0d: Why this destination?

playbooks/roles/tanaguru/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
+---
+tanguru_debian_pkgs:
+  - openjdk-7-jre
+  - unzip
+  - libmysql-java
+  - python-mysqldb
+  - tomcat6
```
e0d: Tomcat6 isn't EOL until 31 December 2016, but we should test with TC7

feanil: +1 to moving to TC7

playbooks/roles/tanaguru/tasks/main.yml
feanil: We should have the paths for the tanaguru role defined in defaults like we do in other roles.

feanil:
- There should be a tanaguru user that all of this in installed under.  
- There should be defaults vaules for the home directory in `defaults/main.yml` that is referenced in the tasks.
- Instead of using handlers you should have tasks to do the restart that run only when the relevant changes have been made. So you would register tasks and see if they have changed before doing the restart.

playbooks/roles/tanaguru/tasks/main.yml
arbabnazar: that's just a path of extracted tarball, we can delete it after application installation. I have changed it to the `/tmp` directory.

playbooks/roles/tanaguru/tasks/main.yml
arbabnazar, line 8: @e0d I have added the tags to each task, please review it, if it is correct. Thanks

arbabnazar:
- replaced the `tomcat6` with `tomcat7` and everything run smoothly.
- `tanaguru` is installed using the tomcat7 user and deploy under `/var/lib/tomcat7/webapps`

``` shell
drwx------   2 tomcat7 root    4096 Dec  3 07:52 tanaguru/
```
- we don't need the unzipped directory after installation, so I have changed the unzipped directory path to `tmp` instead of `~/`
- I have removed the handlers and replaced it with the tasks using the same approach that @feanil has mentioned.
- Each task is tag using the [Role-Life-Cycle](https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle) approach.

Thanks

playbooks/roles/tanaguru/tasks/main.yml
```diff
@@ -0,0 +1,175 @@
...
+    --tg-admin-email "{{ tanaguru_parameters.tanaguru_admin_email }}" \
+    --tg-admin-passwd "{{ tanaguru_parameters.tg_admin_passwd }}" \
+    --firefox-esr-path /opt/firefox/firefox \
+    --display-port ":99.1"
+  args:
+    chdir: "/tmp/{{ tanaguru_download_link | basename | regex_replace('.tar.gz$', '') }}"
+  when: "tanaguru_app.stdout.find('/etc/tanaguru/') == -1"
+  register: tanaguru_install
+  tags:
+    - install
```
e0d, line 169: add install:base

playbooks/roles/tanaguru/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
+  - mysql
```
e0d: Let's move this to the play.  It gives us more flexibility in installing, we can choose to install mysql or not, for example.  We are trying to use met dependencies specifically for extension, like the edx_service case, and to do composition in plays.

e0d:
After my last two comments, I'm :+1:

arbabnazar:
@e0d I have modified the playbook as per your comments, please check it. Thanks

playbooks/roles/tanaguru/templates/xvfb.j2
```diff
@@ -0,0 +1,35 @@
+#!/bin/sh
+
+set -e
+
+RUN_AS_USER=tomcat6
```
e0d: Should update this for tomcat7

playbooks/roles/tanaguru/defaults/main.yml
```diff
@@ -0,0 +1,25 @@
...
+  - mailutils
+  - postfix
+
+tanaguru_download_link: "http://download.tanaguru.org/Tanaguru/tanaguru-3.1.0.i386.tar.gz"
+# Go this link to find your desired ESR Firefox
+# http://download-origin.cdn.mozilla.net/pub/firefox/releases/24.0esr/linux-x86_64/
+# Default is en-US in our example
+fixfox_esr_link: "http://download-origin.cdn.mozilla.net/pub/firefox/releases/24.0esr/linux-x86_64/en-US/firefox-24.0esr.tar.bz2"
+
+tanaguru_parameters:
```
e0d, line 26: We should follow our convention of using UPPERCASE vars in cases where we expect them to be overridden.   Similar to what we do, say, here:
https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L82

arbabnazar:
@e0d kindly review it. Thanks

e0d:
:+1:

PR #2526, opened 2015-11-30
Properly close the global default cache for Django 1.4.

nedbat:


nedbat:
@e0d @feanil kind of belt-and-suspenders, but it should work.

e0d:
:+1:

feanil:
:+1:

PR #2481, opened 2015-11-15
Allow enabling HTTP basic auth for individual services.

smarnach:
This is a fixed version of #2322, which was reverted due to a bug.

**Explanation of the bug in the original version**. The problem with the original PR was that there was no default value for `COMMON_ENABLE_BASIC_AUTH`.  So far, this variable was only used in Jinja2 conditionals, like

```
{% if COMMON_ENABLE_BASIC_AUTH %}
```

In this context, the undefined variable was interpreted as falsy.  With the original PR applied, `COMMON_ENABLE_BASIC_AUTH` was also used in Jinja2 variable expansions like

```
EDXAPP_LMS_ENABLE_BASIC_AUTH: "{{ COMMON_ENABLE_BASIC_AUTH }}"
```

An undefined variable in this context results in the string being unchanged, so `EDXAPP_LMS_ENABLE_BASIC_AUTH` becomes the literal string `"{{ COMMON_ENABLE_BASIC_AUTH }}"`, which is considered truthy in a conditional.  This way, the Ansible equivalent of variable assignment turned a falsy value into a truthy one, resulting in undesired defaults.

The fix is to add a default value for `COMMON_ENABLE_BASIC_AUTH`.

**Original PR description**. Currently, there is a single flag COMMON_ENABLE_BASIC_AUTH to decide whether to enable basic auth in the nginx config. This change allows to enable basic auth selectively for individual services.

I tested this locally targetting a virtual machine and using variations of this playbook:

``` yaml

---
- hosts: all
  sudo: True
  gather_facts: True
  vars:
    COMMON_ENABLE_BASIC_AUTH: True
    EDXAPP_LMS_ENABLE_BASIC_AUTH: False
  roles:
    - role: nginx
      nginx_sites:
      - cms
      - lms
      nginx_default_sites:
      - lms
    - edxapp
```

PR #2527, opened 2015-11-30
Refactor installation of Python requirements.

smarnach:
The current code installs the Python requirements files in a different order than "paver install_prereqs".  This has been causing problems for the edx-solutions fork of edx-platform.

The current code is also rather repetitive.  This patch cleans the repetitions up a bit, while at the same time making sure that the order in which the requirements files are processed matches the order paver uses.

Additionally, the patch extends the list of requirements files which are scanned for Github URLs by two items that seemed to be missing.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -595,18 +595,6 @@ edxapp_sandbox_user: 'sandbox'  # I think something about the codejail requires 
 # apparmor command
 edxapp_aa_command: "{% if EDXAPP_SANDBOX_ENFORCE %}aa-enforce{% else %}aa-complain{% endif %}"
 
-# all edxapp requirements files
-edxapp_requirements_with_github_urls:
-  - "{{ pre_requirements_file }}"
-  - "{{ post_requirements_file }}"
-  - "{{ base_requirements_file }}"
-  - "{{ paver_requirements_file }}"
-  - "{{ github_requirements_file }}"
-  - "{{ sandbox_post_requirements }}"
-  - "{{ sandbox_local_requirements }}"
-  - "{{ sandbox_base_requirements }}"
-
-
 # TODO: old style variable syntax is necessary
 # for lists and dictionaries
 edxapp_helper_scripts:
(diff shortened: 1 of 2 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=2527

e0d:
@smarnach thanks for this PR.  Internally we have been talking about enforcing a contract with all apps such that

`make pre_reqs`

will do the same thing everywhere and that all the roles can be dramatically simplified.  Depending upon time, it may make sense to merge this in the interim, but I think that's the direction that we ultimately want to go.

Thoughts?

smarnach:
@e0d That sounds like a great goal.  Making requirement installation more consistent accross apps, and more importantly, implementing it only once per app would be a significant improvement.  Currently, it's implemented in Paver and Ansible for edxapp, so the two implementations need to be kept in sync.  This is just asking for trouble.

After my experience with #2481, I'm also all in favour of moving code out of the Ansible scripts.  Ansible isn't a great language for more complex logic.

smarnach:
@e0d Oh, and yes, getting this PR merged in the meantime might be worthwhile.  Implement `make prereqs` for all apps sounds like a longer-term goal.

openedx-webhooks:
Thanks for the pull request, @smarnach! I've created [OSPR-995](https://openedx.atlassian.net/browse/OSPR-995) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

You haven't added yourself to the [AUTHORS](https://github.com/edx/configuration/blob/master/AUTHORS)      file for this repo. Please create a new pull request to do so!     Please see the [CONTRIBUTING](https://github.com/edx/edx-platform/blob/master/CONTRIBUTING.rst)     file for more information.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -162,105 +162,41 @@
...
-- name : install python custom-requirements
-  pip: >
-    requirements="{{ custom_requirements_file }}"
-    virtualenv="{{ edxapp_venv_dir }}"
-    state=present
-    extra_args="-i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w"
-  sudo_user: "{{ edxapp_user }}"
-  environment: "{{ edxapp_environment }}"
-  when: custom_requirements.stat.exists
+- name: Stat each Python requirements file to ensure it exists
```
bradenmacdonald, line 176: Out of curiosity, when do these not exist?

bradenmacdonald:
@smarnach I don't see any issues with this, but your related sandbox provisioning failed. It wasn't immediately obvious to me from the logs whether the failure is related or not.

playbooks/roles/edxapp/tasks/deploy.yml
e0d, line 176: They don't exist when someone refactors requirements in the platform without updating the ansible code, usually because they don't know about it.  As @smarnach it's a bit of a mess, logic is duplicated for no particularly useful reason.

I think that our projects should be approachable in their own right.  You shouldn't necessarily need to know about ansible to stand up the platform -- though in the aggregate knowing ansible is useful.  A platform developer should be able to refactor requirements into whatever files they want without breaking deployments, but they'll have to make sure that

```
make pre_reqs
```

does the right thing.  We will achieve some curatorial control over this internally by, eventually, installing everything from an internally wheelhouse or pypi proxy so that we can have some control over libraries.

smarnach, line 176: The old version of the code only checked post.txt and custom.txt for existence before installing.  The fiel custom.txt only exists for the solutions fork, so I understand why this file was checked.  I've got no idea whether there was a specific reason to check for post.txt.

Since I'm using a loop now, I'm simply checking all files for existence before installing the requirements from them, since I figured that it will never hurt.

smarnach:
@bradenmacdonald The failed sandbox provisioning is unrelated – it failed during the installation of the Ansible requirements, so it didn't even start executing the playbook.  I'm giving it another shot now.

e0d:
@smarnach mysql related?

smarnach:
@e0d Yes.

smarnach:
@e0d I've seen #2537 and rebased on top of current master.

bradenmacdonald:
@smarnach New failure:

```
TASK: [edxapp | ensure edxapp_workers has started]
...
2015-12-03 14:11:44-0800 | INFO | instance.models.instance  | instance=configuration-pr2527 | failed: [149.202.181.2] => (item={'queue': 'default', 'service_variant': 'lms', 'monitor': True, 'concurrency': 3}) => {"failed": true, "item": {"concurrency": 3, "monitor": true, "queue": "default", "service_variant": "lms"}}
2015-12-03 14:11:44-0800 | INFO | instance.models.instance  | instance=configuration-pr2527 | msg: edxapp_worker:lms_default_3: ERROR (abnormal termination)
...
2015-12-03 14:11:46-0800 | INFO | instance.models.instance  | instance=configuration-pr2527 | FATAL: all hosts have already failed -- aborting
```

smarnach:
@bradenmacdonald Yes, I've seen this.  I'm pretty sure this is caused by a supervisord bug (a deadlock when some buffer fills up or similar).  So far I've only experienced this under memory pressure, but this time it seems to be different.  I'll have to investigate further tomorrow.

feanil:
@smarnach can you rebase this off of the latest master of configuration?  I added a change that will make it easier for us to review this on our sandboxes and I need your branch to have it for it to work correctly.

smarnach:
@feanil Thanks for looking into this.  I've just rebased on top of master.

smarnach:
@bradenmacdonald @feanil I've managed to make the sandbox provisioning complete successfully by adding a hack to circumvent the supervisord bug.  Studio on the sandbox works fine and the LMS is throwing the same 500 as master.sandbox.opencraft.com, which was deployed from current master.  It looks like this patch is working, but we will have to fix the master branch again to be sure.

I'll send an email to openedx-ops about the 500.  Regarding the problem with supervisord, I'm currently testing a less hacky workaround than the one I used to make the provisioning succeed.  If it works, I'll file a PR with additional detail, and if it fails another post to the mailing list.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -995,19 +983,48 @@ edx_platform_repo: "https://{{ COMMON_GIT_MIRROR }}/edx/edx-platform.git"
...
-github_requirements_file: "{{ edxapp_code_dir }}/requirements/edx/github.txt"
 private_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/edx-private.txt"
 
 sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/base.txt"
 sandbox_local_requirements: "{{ edxapp_code_dir }}/requirements/edx-sandbox/local.txt"
 sandbox_post_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/post.txt"
 
+# The Python requirements files in the order they should be installed.  This order should
+# match the order of PYTHON_REQ_FILES in edx-platform/pavelib/prereqs.py.  (Also check the
+# edx-solutions fork of edx-platform when if you consider reordering this list.)
```
feanil, line 1001: nit: `when if you consider` should be `if you are considering`

feanil:
:+1:  once the comment is fixed.  @jibsheet can you do a second review.

e0d:
:+1:

smarnach:
@e0d @feanil Thanks for the reviews.  I just meant to fix the comment, but now it's already merged.  Since the comment can be understood (even if it's broken English), I figure it's not worth the trouble to open yet another PR to fix it.

PR #2529, opened 2015-12-01
Rotate mongodb.log and serverStatus.log independently

jibsheet:
We were ending up with gaps in our logs on disk (everything was in
splunk, but missing in /edx/var/log/mongo).  When this ran previously
and roatated serverStatus.log but mongodb.log was less than 1M and
didn't need to rotate, the killall -USR1 to mongo would cause the log to
be reopened and written over.


jibsheet:
@edx/devops

feanil:
:+1:

PR #2530, opened 2015-12-01
Make request atomic.

feanil:


jibsheet:
:+1:

playbooks/roles/xqueue/defaults/main.yml
```diff
@@ -79,6 +79,7 @@ xqueue_auth_config:
       PASSWORD: "{{ XQUEUE_MYSQL_PASSWORD }}"
       HOST: "{{ XQUEUE_MYSQL_HOST }}"
       PORT: "{{ XQUEUE_MYSQL_PORT }}"
+      ATOMIC_REQUEST: True
```
doctoryes, line 82: @feanil It's actually `ATOMIC_REQUESTS` with an `S` on the end. So we didn't test this fix for real - it might still work. That said, I think it's worth trying to test things locally before giving it another push.

PR #2532, opened 2015-12-01
Django 1.8 doesn't need to syncdb

jibsheet:
https://docs.djangoproject.com/en/1.8/ref/django-admin/#syncdb

@edx/devops
FYI @doctoryes


feanil:
:+1:

doctoryes:
:+1:

PR #2533, opened 2015-12-02
Course Discovery default db is setup for docker.

feanil:
Sandboxes don't have a 'db' host like docker will.  They need to use 'localhost' for mysql.


feanil:
@edx/devops I verified that this fixes CI.

feanil:
@cpennington FYI

docker/build/course_discovery/ansible_overrides.yml
```diff
@@ -1,2 +1,3 @@
 ---
-course_discovery_gunicorn_host: 0.0.0.0
\ No newline at end of file
+course_discovery_gunicorn_host: 0.0.0.0
+COURSE_DISCOVERY_MYSQL: 'localhost'
```
e0d: this should be db

e0d:
:+1:

PR #2534, opened 2015-12-02
Restart based on the flower user, not a hardcoded service name

jibsheet:
d267342a changed supervisor to run as FLOWER_USER not flower, without
this change, ansible errors out asking supervisor to restart when the
tools box has jobs named stage_flower, prod_flower etc.

@edx/devops


feanil:
:+1:

PR #2537, opened 2015-12-02
Need to install Python-MySQL pre-reqs

e0d:
libmysqlclient-dev needs to be installed for the configuration repo pip requirements to install now.

@jibsheet @feanil @clintonb @cpennington


clintonb:
:+1:

PR #2541, opened 2015-12-03
mongodb_user module doesn't work under pymongo 3.1.1

maxrothman:
@edx/devops


feanil:
Have you tested that this won't break the mongo modules that were added a few weeks ago?

maxrothman:
I have not. I'll comment again when that's complete.

maxrothman:
Discovered that the issue wasn't pymongo, it was that we had a partial hostname in an rs member's name instead of an IP address. Closing.

PR #2544, opened 2015-12-03
Remove the ORA role and clean up some unused ansible-provision arguments

jibsheet:


jibsheet:
Still deleting more ORA code, will need a few more commits.

jibsheet:
@edx/devops I removed more things than I expected here, but all seems safe.  I'm going to do some CI builds to make sure.

@mulby I touched a bunch of analytics stuff in here (devstack) so would appreciate a once-over of https://github.com/edx/configuration/commit/7066dc90ba5523409cc4a5e2df4b1213476c0625

@benpatterson FYI - what we spoke about this afternoon.

mulby:
:+1: on the analytics stuff

feanil:
:+1:

benpatterson:
@jibsheet I just would want to verify that there aren't any tests that would fail whenever we provision a new jenkins-worker. I'm currently seeing an issue in provisioning a new one (it is unrelated to this change, but prevents any verification down that pathway). Do you have a cycle this morning? I can show you....it is a catch-22 problem with mysql.

But, ultimately what I'm after, whether it's devstack or jenkins-worker: I just want to be sure there are no old ora tests hanging around that would subsequently fail. Perhaps some help from TNL folks could point us in the right direction.

jibsheet:
@benpatterson Good questions.  I'm happy to look at the jenkins worker errors so we can test this, but I'm WFH, so you'll probably need to hit up on hipchat.  I've tested AMI/sandbox/clean build sandbox and will test devstack before merging.

I wonder who is most ORA1 knowledgable in TNL land, ideas?

benpatterson:
I'd start with @cahrens. I'll also hit you up on hipchat.

cahrens:
WooHoo-- deleting ORA stuff! Note though that we still have old ORA tests running in edx-platform. They should be removed (without that, I'm worried that they may fail when run on devstack or elsewhere).

jibsheet:
@cahrens Ah - good, that was @benpatterson's concern.  Are the tests grouped together in platform or scattered about?  I may need some help disabling/deleting things for devstack and jenkins workers.

cahrens:
@jibsheet I have a feeling that they are somewhat spread out (certainly both unit tests and acceptance tests). Someone is just going to have to start digging in the code, unfortunately. I was considering working on this last hackathon, but ended up doing something else.

We do have a story to delete all the ORA code in edx-platform (which contains some details that may be helpful), but it is not on any immediate backlog. https://openedx.atlassian.net/browse/TNL-3645

benpatterson:
Sorry point of clarification, is ORA1 _removed_ from edx-platform, or just turned off?

cahrens:
Not removed, just turned off. But we have the OK from product and open source to rip it out.

benpatterson:
Well, if that's the case, we shouldn't remove tests. Then we'd have code but no tests. Maybe the right thing to do here is remove it from edx-platform entirely, before it's removed from configuration? I'm guessing that is non-trivial?

:umbrella:

cahrens:
See TNL-3645, which the team has groomed (and assigned story points of roughly one week of work for a single developer).

sarina:
@jibsheet ORA1 removal code has landed

ref https://github.com/edx/edx-platform/pull/10845

jibsheet:
@benpatterson Results from jenkins test-worker testing look good - jobs 67 through 73 here https://build.testeng.edx.org/view/All/job/meta-certain-tests-pr/67/

sarina:
## Ship It

benpatterson:
:+1: sir

jibsheet:
We're fixing CI this morning, once we do, I'll merge this.

PR #2555, opened 2015-12-04
Max/improve mongo status

maxrothman:
@edx/devops


jibsheet:
requirements.txt documents PyMongo as being used by mongo_status, will you update that command (probably to refer to both mongo modules)

Otherwise :+1:

playbooks/library/mongodb_rs_status
```diff
@@ -57,6 +62,7 @@ def main():
         port=dict(required=False, type='int', default=27017),
         username=dict(required=False, type='str'),
         password=dict(required=False, type='str'),
+        auth_database=dict(required=False, type='str')
```
feanil, line 65: why not default this to admin and make it not required?

maxrothman, line 65: AFAIK, mongo will default to using the "admin" database when the admin user is specified. On the other hand, if the specified user was "foo", it'd try to auth against the "foo" database, which might be correct if you'd granted the proper permissions to the "foo" user. I'd rather not obscure that behavior.

PR #2558, opened 2015-12-07
Remove ORA1 configuration

sarina:
This will be needed once ORA1 is removed from the codebase (ref https://github.com/edx/edx-platform/pull/10845)


jibsheet:
Any idea how this relates to my slash and burn last week here https://github.com/edx/configuration/pull/2544

sarina:
Oh, #2544 is way better than this.

jibsheet:
You should probably cherry pick this commit on there, because I don't think it was one of the ORA things I caught (I was looking for ora and ORA_LEGACY but didn't know all the other vars to grep for).

sarina:
good point, I will do that. I deleted the branch and can't restore it, ug.

jibsheet:
You could wget https://github.com/edx/configuration/commit/4902051e2fbb645250bb2a511b5298255bbb6be3.patch and git am it on in a worst case scenario.

sarina:
eh I just opened a PR into your branch. #2560

PR #2559, opened 2015-12-07
Tweak python package versions to remove hardcoded os version and calculate it instead

jibsheet:
Without this - installation of python 2.7.10 fails on 14.04

@edx/devops This was preventing AMI builds of programs on Friday due to the
2.7.10 upgrade and the precise package not existing on 14.04

### DIFF
--- a/playbooks/roles/common_vars/defaults/main.yml
+++ b/playbooks/roles/common_vars/defaults/main.yml
@@ -93,7 +93,7 @@ common_debian_pkgs:
   - rsyslog
   - git
   - unzip
-  - "python2.7=2.7.10-0+precise1"
+  - "python2.7=2.7.10-0+{{ ansible_distribution_release }}1"
   - python-pip
   - python2.7-dev
   


jibsheet:
@jimabramson I've built a stage AMI with this change, just getting signoff on the code change before I deploy it for you

maxrothman:
:+1:

playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -93,7 +93,7 @@ common_debian_pkgs:
   - rsyslog
   - git
   - unzip
-  - "python2.7=2.7.10-0+precise1"
+  - "python2.7=2.7.10-0+{{ansible_distribution_release}}1"
```
feanil: add spaces between the jinja curlies

feanil:
Fix the spacing to match our style guidelines and then :+1:

jimabramson:
:+1: many thanks @jibsheet

PR #2560, opened 2015-12-07
Kill ora1 configuration

sarina:
@jibsheet


jibsheet:
I just removed the defaults too @sarina but we declare these in the secure repos, so we'll want to remove them there too (That's a devops task, but I'm not sure we can do it safely prior to all this landing).

PR #2562, opened 2015-12-07
[ECOM-3111] E-Commerce devstack uses wrong configuration

vkaracic:
Devstack ecommerce app should now use the proper 'ecommerce.settings.devstack' Django module.

https://openedx.atlassian.net/browse/ECOM-3111

@clintonb @feanil , please take a look.

### DIFF
--- a/playbooks/vagrant-devstack.yml
+++ b/playbooks/vagrant-devstack.yml
@@ -15,6 +15,7 @@
     EDXAPP_LMS_BASE: 127.0.0.1:8000
     EDXAPP_OAUTH_ENFORCE_SECURE: false
     EDXAPP_LMS_BASE_SCHEME: http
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @vkaracic! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=2562&repo=edx%2Fconfiguration

clintonb:
Squash your commits

:+1:

feanil:
:+1:

PR #2212, opened 2015-08-13
OPS-631: Upgrade to the latest version of python

feanil:

PR #2563, opened 2015-12-07
Benp/venv provision

benpatterson:
@jibsheet @e0d This gets the ami builds passing again for jenkins_worker.

Somehow PR# 2316 didn't get merged with #2212 ? Don't have time to figure out that headscratcher so here's a new PR.


jibsheet:
That's super weird - all I can think of is some sort of rebase
:+1:

PR #2564, opened 2015-12-07
WIP, review only: E0d/hacking

e0d:
@benpatterson I realize you've got this working, but this is what I'm thinking of for a bootstrap script.  Some other things would need to be refactored to use this as well.

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -1,0 +1,62 @@
+#!/usr/bin/env bash
+
+#
+# Script for installing Ansible and the edX configuration repostory
+# onto a host to enable running ansible to complete configuration.
+# This script can be used by Docker, Packer or any other system
+# for building images that requires having ansible available.
+#
+
+set -e
+
+VIRTUAL_ENV="/tmp/bootstrap"
+PYTHON_BIN="${VIRTUAL_ENV}/bin"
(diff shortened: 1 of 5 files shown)


util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,62 @@
...
+add-apt-repository ppa:fkrull/deadsnakes-python2.7
+apt-get update -y
+apt-get install -y sudo python2.7 python2.7-dev python-pip python-apt python-yaml python-jinja2 libmysqlclient-dev
+
+pip install virtualenv==13.1.2
+/usr/local/bin/virtualenv ${VIRTUAL_ENV}
+
+# ansible bootstrap
+git clone --recursive ${ANSIBLE_REPO} ${ANSIBLE_DIR}
+cd /tmp/ansible
```
benpatterson: replace with `${ANSIBLE_DIR}` ?

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,62 @@
...
+PATH=$PATH:/tmp/ansible/bin
```
benpatterson: same here...use the var instead of `/tmp/ansible`

benpatterson:
I had some comments but overall I like it, and am a big fan of unifying how we set this up.

PR #2566, opened 2015-12-08
OPS-241 Ansible: MySql 5.6 installs should remove the 'test' database

arbabnazar:
@edx/devops kindly review it. Thanks


playbooks/roles/mysql/tasks/main.yml
```diff
@@ -10,18 +10,16 @@
...
+    state: present
 
 - name: install apt repository
   apt_repository:
-    repo='deb http://ppa.launchpad.net/ondrej/mysql-5.6/ubuntu precise main'
-    update_cache=yes
+    repo: 'deb http://ppa.launchpad.net/ondrej/mysql-5.6/ubuntu precise main'
+    update_cache: yes
 
 - name: look for mysql 5.5
```
e0d: While we are updating this I think it would be a good time to remove the logic that checks of MySQL 5.5.  That was required for upgrades, but can now be safely removed.

playbooks/roles/mysql/tasks/main.yml
```diff
@@ -32,14 +30,28 @@
...
+  apt: 
+    name: "{{ item }}" 
+    install_recommends: yes 
+    force: yes 
+    state: present
   with_items: mysql_debian_pkgs
 
+- name: Ensure Anonymous user(s) does not exist
+  mysql_user: 
+    name: ''
```
e0d, line 30: Does this affect root logins without a password challenge?

e0d:
Looks good in general, need to test with the create_dbs_and_users play to ensure that still works as expected.

@edx/devops second review please.

playbooks/roles/mysql/tasks/main.yml
arbabnazar, line 30: Yes, it doesn't need root password.

playbooks/roles/mysql/defaults/main.yml
```diff
@@ -3,4 +3,6 @@ mysql_debian_pkgs:
   # Note that mysql 5.6 is installed from a non standard PPA in tasks/main.yml
```
feanil: This comment doesn't seem relevant anymore.

feanil:
:+1:  once the comment is removed/updated.

playbooks/roles/mysql/defaults/main.yml
jibsheet: It looks like we're still grabbing MySQL from http://ppa.launchpad.net/ondrej/mysql-5.6/ubuntu
The diff is super annoying to read.

playbooks/roles/mysql/defaults/main.yml
```diff
@@ -3,4 +3,6 @@ mysql_debian_pkgs:
...
   # It is not listed here because of non-standard steps required to install
```
feanil: I meant this comment is not relevant since it's in the list now.

e0d:
:+1:

benpatterson:
@edx/devops It looks like this pull request has broken our packer builds. (If I were to build an AMI, it would have mysql 5.5 installed when the job completes.) 

It may make more sense to use the same base AMI as you-all. However, I'd like to resolve this issue quickly so am going to put in a revert PR for now.

(EDIT: packer builds for jenkins_worker machines)

e0d:
What are you using for a base ami?

benpatterson:
I can send you the AMI ID if you'd like. Here is the manually-maintained changelog:

```
Update 2/2/2015:
Based on ubuntu precise 12.04 free image on EC2. 

In addition:
* python-pip 
* python-dev 
* added master and test master key
* patched for GHOST vulnerability (upgrade libc6)
```

e0d:
Are you installing mysql on the base image?  Otherwise I'm missing
something.

feanil:
Agreed, the only way this would be relevant is if you were using the mysql role on your base image.  Are you using the configuration repo on the packer build?

benpatterson:
Well, we do use the mysql role when building out jenkins_worker. I believe that's called out as a dependent role somewhere (common? edxapp?)

benpatterson:
FWIW I spun up a machine based on our base AMI, and it does not have mysql installed.

benpatterson:
Found the problem. The ppa for ubuntu precise is gone. I'm going to enter an ops ticket to resolve the overall problem (e.g., pull down from the mysql apt repository, or whatever you guys decide is right).

benpatterson:
Additional info/commentary can be found on https://openedx.atlassian.net/browse/TE-1118.

PR #2567, opened 2015-12-08
Ansible bootstrap script

e0d:
@benpatterson cleaned  up script and Makefile

@edx/devops Bootstrapping script that ultimately installs the edx_ansible role on an Ubuntu Precise hosts.  Overlaps with the sandbox.sh script, but that script should use this.  This script can be used by both packer and docker to bootstrap image building, regardless of the packing tools

I would like to merge this to master and rebase the docker branch to pull it in.  This will remove a substantial amount of duplication from the Dockerfiles.

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -79,10 +79,15 @@ fi
 # Required for add-apt-repository
 apt-get install -y software-properties-common python-software-properties git
 
-# Install python 2.7.10
-add-apt-repository ppa:fkrull/deadsnakes-python2.7
+# Add git PPA
+add-apt-repository ppa:git-core/ppa
+
+# Add python PPA
+ppa:git-core/ppa
+
+# Install python 2.7.10, git and other common requirements
 apt-get update -y
-apt-get install -y build-essential sudo python2.7 python2.7-dev python-pip python-apt python-yaml python-jinja2 libmysqlclient-dev
+apt-get install -y build-essential sudo git python2.7 python2.7-dev python-pip python-apt python-yaml python-jinja2 libmysqlclient-dev
 
 pip install virtualenv==${VIRTUAL_ENV_VERSION}
 
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -7,8 +7,9 @@
 # for building images that requires having ansible available.
 #
 # Can be run as follows:
-# bash <(curl -s https://raw.githubusercontent.com/edx/configuration/e0d/bootstrap-script/util/install/ansible-bootstrap.sh)
 #
+# UPGRADE_OS=true CONFIGURATION_VERSION="master" \
+# bash <(curl -s https://raw.githubusercontent.com/edx/configuration/master/util/install/ansible-bootstrap.sh)
 
 set -xe
 
@@ -89,7 +90,9 @@ add-apt-repository -y ppa:git-core/ppa
 # Add python PPA
 add-apt-repository -y ppa:fkrull/deadsnakes-python2.7
 
-# Install python 2.7.10, git and other common requirements
+# Install python 2.7 latest, git and other common requirements
+# NOTE: This will install the latest version of python 2.7 and
+# which may differ from what is pinned in virtualenvironments
 apt-get update -y
 apt-get install -y build-essential sudo git-core python2.7 python2.7-dev python-pip python-apt python-yaml python-jinja2 libmysqlclient-dev
 
@@ -123,7 +126,9 @@ cat << EOF
 ******************************************************************************
 
 Done bootstrapping, edx_ansible is now installed in /edx/app/edx_ansible.
-Time to run some plays.
+Time to run some plays.  Activate the virtual env with 
+
+> . /edx/app/edx_ansible/venvs/edx_ansible/bin/activate 
 
 ******************************************************************************
 EOF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -1,0 +1,106 @@
+#!/usr/bin/env bash
+
+#
+# Script for installing Ansible and the edX configuration repostory
+# onto a host to enable running ansible to complete configuration.
+# This script can be used by Docker, Packer or any other system
+# for building images that requires having ansible available.
+#
+
+set -xe
+
+if [[ -z "$ANSIBLE_REPO" ]]; then
+  ANSIBLE_REPO="https://github.com/edx/ansible.git"
+fi
+
+if [[ -z "$ANSIBLE_VERSION" ]]; then
+  ANSIBLE_VERSION="master"
+fi
+
+if [[ -z "$CONFIGURATION_REPO" ]]; then
+  CONFIGURATION_REPO="https://github.com/edx/configuration.git"
+fi
+
+if [[ -z "$CONFIGURATION_VERSION" ]]; then
+  CONFIGURATION_VERSION="e0d/hacking"
+fi
+
+#
+# Bootstrapping constants
+#
+VIRTUAL_ENV="/tmp/bootstrap"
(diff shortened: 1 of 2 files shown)


util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,106 @@
...
+if [[ -z "$ANSIBLE_VERSION" ]]; then
+  ANSIBLE_VERSION="master"
+fi
+
+if [[ -z "$CONFIGURATION_REPO" ]]; then
+  CONFIGURATION_REPO="https://github.com/edx/configuration.git"
+fi
+
+if [[ -z "$CONFIGURATION_VERSION" ]]; then
+  CONFIGURATION_VERSION="e0d/hacking"
```
maxrothman: Shouldn't this be master?

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,106 @@
...
+ANSIBLE_VERSION="${ANSIBLE_VERSION}"
+CONFIGURATION_REPO="${CONFIGURATION_REPO}"
+CONFIGURATION_VERSION="${CONFIGURATION_VERSION}"
+
+******************************************************************************
+EOF
+
+
+if [[ $(id -u) -ne 0 ]] ; then
+    "Please run as root";
```
maxrothman: Did you mean to have an echo here?

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,106 @@
...
+    This script is only known to work on Ubuntu Precise, exiting.
+    If you are interested in helping make installation possible
+    on other platforms, let us know.
+EOF
+   exit 1;
+fi
+
+# Upgrade the OS
+apt-get update -y
+apt-get upgrade -y
```
maxrothman: This seems a bit aggressive.

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,106 @@
...
+
+# Required for add-apt-repository
+apt-get install -y software-properties-common python-software-properties git
+
+# Install python 2.7.10
+add-apt-repository ppa:fkrull/deadsnakes-python2.7
+apt-get update -y
+apt-get install -y build-essential sudo python2.7 python2.7-dev python-pip python-apt python-yaml python-jinja2 libmysqlclient-dev
```
maxrothman: Why libmysqlclient-dev and python2.7-dev?
Also, shouldn't git be in here?

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,106 @@
...
+
+pip install virtualenv==13.1.2
```
maxrothman: Version should be a variable

util/install/ansible-bootstrap.sh
feanil: libmysqlclient-dev is needed for the mysql-python dependency in the configuration requirements.

maxrothman: is -dev specifically required?

feanil:
:+1: from me  once 
- `CONFIGURATION_VERSION` defaults to master
- `echo` added on line 51

util/install/ansible-bootstrap.sh
feanil: Mysql-python compiles against libmysql so it needs the headers.

util/install/ansible-bootstrap.sh
e0d: I tend to agree, but it's in the existing sandbox script, prior art, and I think the benefit probably outweighs the harm, but I could be convinced otherwise.

maxrothman: Someone installing the stack on a machine that also hosted other things wouldn't appreciate this. if we're dead-set on doing it, we should at least warn that it's going to happen and/or ask for confirmation.

e0d: From the docs:

```
WARNING:

    These instructions will potentially destroy the server they are run on, you should
only do them on a freshly installed virtual machine. But if you still want to have a
try to re-install the Open edX stack on the same server, please see this article for 
some issues you may face and how to fix them.
```

How about a force argument and, in its absence, a prompt?

maxrothman: Sounds good.

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env bash
+
+#
+# Script for installing Ansible and the edX configuration repostory
+# onto a host to enable running ansible to complete configuration.
+# This script can be used by Docker, Packer or any other system
+# for building images that requires having ansible available.
```
benpatterson, line 7: The underlying edx_ansible role installs ansible in a venv, right? I think this readme/blurb should be updated slightly to point that out. That everything ansible needs is installed, as opposed to some globally-available ansible-playbook target when the shell script is done.

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,107 @@
...
+#
+
+set -xe
+
+if [[ -z "$ANSIBLE_REPO" ]]; then
```
benpatterson, line 16: FWIW `-z` (in my experience) can give unexpected results if the var is set, but empty. (It's stored as "" IIRC). nothing to change here but worth noting in other scripts you might use in this area.

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,107 @@
...
+EOF
+   exit 1;
+fi
+
+# Upgrade the OS
+apt-get update -y
+apt-get upgrade -y
+
+# Required for add-apt-repository
+apt-get install -y software-properties-common python-software-properties git
```
benpatterson: We're using a ppa for git for jenkins_worker, because the apt default git debian package was installing something too old for Jenkins core. (Can't remember which.)

util/install/ansible-bootstrap.sh
benpatterson: Should build-essential be installed with the first install line above? Just...I'd expect that to be one of the first things installed, but NBD.

util/install/ansible-bootstrap.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# cleanup
+rm -rf ${ANSIBLE_DIR}
+rm -rf ${CONFIGURATION_DIR}
+rm -rf ${VIRTUAL_ENV}
+
+cat << EOF
+******************************************************************************
+
+Done bootstrapping, edx-ansible is now installed in /edx/app/edx-ansible.
```
benpatterson: Ah here's the answer. Sorry if I missed this, but when I call edx-ansible, do I need an active venv?

e0d:
@cpennington the thought is that this will replace most of the meat of the common Dockerfiles as well as be used in packer for jenkins workers and the one-step install script.  Chime in if you object.

benpatterson:
@e0d I have not had a cycle to do this, but I'd like to take that bootstrap script and plug it into our packer jenkins_worker.json and ensure it gets the job done. Hopefully in the next few days unless someone else beats me to it.

e0d:
@benpatterson I've had to make a few changes, can you give me up or down thumb?

util/install/ansible-bootstrap.sh
cpennington: `edx-ansible` or `edx_ansible`? (It's the latter right now. Is this PR changing that?)

benpatterson:
@e0d I've been spending time getting our packer provisioning to the point where we can consume this, but that work is not going smoothly. I did not look at your updates but if you've got enough thumbs, you can feel free to merge. Anything problematic I find can be follow-on work you promise to help on :)

util/install/ansible-bootstrap.sh
e0d: Nope, just a typo in docs in comments, will fix.

jibsheet:
:+1:

PR #2568, opened 2015-12-08
Run devstack-specific install tasks in Dockerfile (those tasks are no…

cpennington:
…t part of the global install tag to avoid pollution)


cpennington:
@e0d @feanil: Can I get a quick review?

cpennington:
Closing in favor of #2571.

PR #2570, opened 2015-12-08
Update the apt-cache if it hasn't been updated within the last hour (…

cpennington:
…when installing system packages in edx_service)

@feanil @e0d: apt-get cache updates, as promised.


e0d:
:+1:

feanil:
:+1:

PR #2569, opened 2015-12-08
Install python-mysqldb to allow ansible to run mysql tasks

cpennington:

PR #2571, opened 2015-12-08
Use devstack as a prefix, rather than a suffix, in tags

cpennington:
@e0d @feanil: Review?


feanil:
@e0d I want to get your thoughts on this.  It makes sense to me since it seems like we need some name spacing for the devstack specific tasks.

cpennington:
I've tested this both against e0d/docker-artifacts and against the currently published trusty-common image, and both work (although w/ the trusty-common image, it currently can't bootstrap the mysql database because an image with #2569 hasn't yet been pushed).

e0d:
:+1:

PR #2511, opened 2015-11-23
E0d/docker artifacts

e0d:
Merging updated docker artefacts from the hack2015/docker branch.

PR #2574, opened 2015-12-09
Add CI tasks for docker images

cpennington:
@e0d @feanil: I'd love some review on this.


cpennington:
Also, note that this will only run once #2511 has been merged.

cpennington:
@mulby: Too many `Makefile` shenanigans, or just the right amount?

.travis.yml
```diff
@@ -1,63 +1,34 @@
...
+
 branches:
   only:
     - master
+
 python:
   - "2.7"
+
+services:
+  - docker
```
e0d, line 12: Can we specify minimum version here?  Is that even an option?

.travis.yml
```diff
@@ -1,63 +1,34 @@
...
+  
+addons:
+  apt:
+    packages:
+      - nodejs
```
e0d, line 17: Why is this here?

Makefile
```diff
@@ -1,14 +1,43 @@
...
+	@echo '    make test                      run all tests'
+	@echo '    make build                     build everything'
+	@echo '    make pkg                       package everything'
+	@echo '    make clean                     remove build by-products'
+	@echo ''
+	@echo '    Docker:'
+	@echo '        $$image: any dockerhub image'
+	@echo '        $$container: any container defined in docker/build/$$container/Dockerfile'
+	@echo ''
+	@echo '        make $(docker_pull)$$image        pull $$image from dockerhub'
```
e0d, line 21: Why is this needed a make provided command?  Are we providing a competing idiom?

docker/build/elasticsearch/xDockerfile
```diff
@@ -1,7 +1,6 @@
 FROM edxops/precise-common:latest
```
e0d, line 1: What's up with all these renames?

e0d:
@cpennington Lot's of good stuff here, but have some questions.  If the Dockerfile were renamed to disable them, I'd like to address whatever issues necessitated that before merging.

cpennington:
@e0d: Those files are broken on master. I'd rather get the CI integration merged, and then make separate PRs to fix the disabled files individually.

e0d:
@cpennington I'm OK with that.  I had some other questions as well, but nothing too big.

docker/build/elasticsearch/xDockerfile
cpennington, line 1: As you surmised, these are disabling dockerfiles.

Makefile
cpennington, line 21: It's used so that we can use it as a pre-requisite task and not pull multiple times.

.travis.yml
cpennington, line 12: Sadly, no.

.travis.yml
cpennington, line 17: Because it was install using `sudo apt-get install` below.

cpennington:
@e0d: Good to merge?

e0d:
:+1:

PR #2575, opened 2015-12-10
Remove openid_workaround

jibsheet:
We now run off a fork of django_openid_auth
The unique constraint was removed in our fork, this sed is a no-op
https://github.com/edx/django-openid-auth/commit/ea49a01d85cb2fccc5fe34f3d0e042dc1deaf8e8

@edx-ops/devops


feanil:
:+1:

PR #2577, opened 2015-12-10
Vagrant playbooks need common now.

feanil:
This is needed now because edx-ansible no longer directely depends on common.

@edx/devops


jibsheet:
:+1:

PR #2583, opened 2015-12-11
Upgrade packer so we can use ansible provisioner instead of shell.

benpatterson:
@jzoldak @e0d 

This will precede some retooling of the packer script so we can ultimately absorb the work Ed is putting together for #2567


jzoldak:
:+1:

PR #2584, opened 2015-12-11
Use ansible-local provisioner.

benpatterson:
Note: We have some redundancy in terms of the virtualenv, and that's because our base image has python 2.7.3, and over the course of the upgrade, the python version will change. So, the subsequent ansible play will need a fresh virtual environment. We'll improve that aspect of things later.

### DIFF
--- a/playbooks/roles/test_build_server/tasks/main.yml
+++ b/playbooks/roles/test_build_server/tasks/main.yml
@@ -24,7 +24,7 @@
   git: >
     repo=https://github.com/edx/edx-platform.git
     dest={{ test_build_server_repo_path }}/edx-platform-clone
-    version=master
+    version={{ test_edx_platform_version }}
   sudo_user: "{{ test_build_server_user }}"
 
 - name: get xargs limit
--- a/playbooks/roles/test_build_server/tasks/main.yml
+++ b/playbooks/roles/test_build_server/tasks/main.yml
@@ -27,12 +27,14 @@
     version=master
   sudo_user: "{{ test_build_server_user }}"
 
+- name: get xargs limit
+  shell: "xargs --show-limits"
+
 - name: Copy test-development-environment.sh to somewhere the jenkins user can access it
   copy: >
     src=test-development-environment.sh
     dest="{{ test_build_server_repo_path }}"
     mode=0755
-  sudo_user: "{{ test_build_server_user }}"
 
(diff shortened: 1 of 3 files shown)


benpatterson:
@jzoldak @e0d 

From here, we can wrap in Ed's provisioning script and get rid of all/most of the various apt-installs.

If we update our base image to already-contain Python 2.7.10, we can likewise get rid of the two virtualenvs. But for now: this is an incremental move towards using the ansible provisioner. Thoughts?

benpatterson:
FYI we are now using an image based on this branch on Jenkins.

benpatterson:
@jibsheet Heads up on this PR as well. It does two things:
1. uses packer's ansible-local target for provisioning, rather than shell scripts
2. it makes the platform branch we use to validate the given AMI configurable (so, if we want to build the AMI, and run smoketests according to a specific branch (e.g., kill-ora1), we can do so).

playbooks/roles/test_build_server/tasks/main.yml
```diff
@@ -24,15 +24,17 @@
   git: >
     repo=https://github.com/edx/edx-platform.git
     dest={{ test_build_server_repo_path }}/edx-platform-clone
-    version=master
+    version={{ test_edx_platform_version }}
   sudo_user: "{{ test_build_server_user }}"
 
+- name: get xargs limit
+  shell: "xargs --show-limits"
+
```
jibsheet, line 32: Was the leftover debugging?

jibsheet:
@benpatterson Looks good, :+1:

jzoldak:
:+1:

playbooks/roles/test_build_server/tasks/main.yml
benpatterson, line 32: I can't explain it, but without that, the next `copy` step would hang. I grepped ansible bug logs and...well, either way i'd like to just leave it in there so i don't continue getting blocked. it will be noise for now.

jzoldak, line 32: @benpatterson I realize this is merged already.  If you remember maybe in the next change to this repo you could tack on a comment here for our future selves and others so nobody ever needs to re-figure that out.

benpatterson, line 32: Sounds good.

PR #2585, opened 2015-12-11
Max/idempotent mongodb cluster config

maxrothman:
OPS-1209

@edx/devops

### DIFF
--- a/playbooks/library/mongodb_replica_set
+++ b/playbooks/library/mongodb_replica_set
@@ -150,7 +150,6 @@ def set_member_ids(members, old_members=None):
                 member['_id'] = match['_id'] if match is not None else available_ids.pop()
             else:
                 member['_id'] = available_ids.pop()
-    return members
 
 def get_matching_member(member, members):
     '''Return the rs_member from `members` that "matches" `member` (currently on host)'''
@@ -158,12 +157,16 @@ def get_matching_member(member, members):
     return match[0] if len(match) > 0 else None
 
 def members_match(new, old):
-    "Compare 2 lists of members, discounting their `_id`s and matching on a custom criterion"
+    "Compare 2 lists of members, discounting their `_id`s and matching on hostname"
     if len(new) != len(old):
         return False
     for old_member in old:
-        new_member = get_matching_member(old_member, new)
-        if old_member.copy().pop('_id') != new_member.copy().pop('_id'):
+        new_member = get_matching_member(old_member, new).copy()
+        #Don't compare on _id
+        new_member.pop('_id', None)
+        old_member = old_member.copy()
+        old_member.pop('_id', None)
+        if old_member != new_member:
             return False
     return True
 
@@ -176,27 +179,33 @@ def fix_host_port(rs_config):
         for member in rs_config['members']:
             if ':' not in member['host']:
                 member['host'] = '{}:{}'.format(member['host'], member.get('port', 27017))
-                del member['port']
+                if 'port' in member:
+                    del member['port']
 
 def update_replset(rs_config):
     changed = False
     old_rs_config = get_replset(client)
-    fix_host_port(module, rs_config)  #fix host, port to host:port
+    fix_host_port(rs_config)  #fix host, port to host:port
 
     #Decide whether we need to initialize
     if old_rs_config is None:
         changed = True
         if '_id' not in rs_config:
             rs_config['_id'] = get_rs_config_id(client)  #Errors if no replSet specified to mongod
-        rs_config['members'] = set_member_ids(rs_config['members'])  #Noop if all _ids are set
+        set_member_ids(rs_config['members'])  #Noop if all _ids are set
         #Don't set the version, it'll auto-set
         initialize_replset(client, rs_config)
 
     else:
-        rs_config_scalars = {k:v for k,v in rs_config.items()
-            if not isinstance(v, list) if not isinstance(v, dict)}
-        old_rs_config_non_collections = {k:v for k,v in old_rs_config.items() 
-            if not isinstance(v, list) if not isinstance(v, dict)}
+        old_rs_config_scalars = {k:v for k,v in old_rs_config.items() if not isinstance(v, (list, dict))}
+
+        rs_config_scalars = {k:v for k,v in rs_config.items() if not isinstance(v, (list, dict))}
+        if '_id' not in rs_config_scalars and '_id' in old_rs_config_scalars:
+            # _id is going to be managed, don't compare on it
+            del old_rs_config_scalars['_id']
+        if 'version' not in rs_config and 'version' in old_rs_config_scalars:
+            # version is going to be managed, don't compare on it
+            del old_rs_config_scalars['version']
 
         # Special comparison to test whether 2 rs_configs are "equivalent"
         # We can't simply use == because of special logic in `members_match()`
@@ -205,8 +214,8 @@ def update_replset(rs_config):
         # 3. Compare the members dicts using `members_match()`
         # Since the only nested structures in the rs_config spec are "members" and "settings",
         # if all of the above 3 match, the structures are equivalent.
-        if rs_config_scalars != old_rs_config_non_collections \
-            or rs_config['settings'] != old_rs_config['settings'] \
+        if rs_config_scalars != old_rs_config_scalars \
+            or rs_config.get('settings') != old_rs_config.get('settings') \
             or not members_match(rs_config['members'], old_rs_config['members']):
 
             changed=True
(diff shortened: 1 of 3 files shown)


maxrothman:
Tests are failing in the mongo_3_0 module because the module was renamed. I won't merge this until the mongo_3_0 role changes land, but I thought it'd be useful to review these changes separately.

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+    required: false
+    default: false
+  rs_config:
+    description: A `replica set configuration document <https://docs.mongodb.org/manual/reference/replica-configuration/>`.
+        This structure can be a valid document, but this module can also manage some details for you:
+
+        - members can have separate ``host`` and ``port`` properties. ``port`` defaults to 27017.
+          To override this, provide a ``host`` like ``somehost:27017``.
+        - ``_id`` is automatically managed if not provided
+        - members' ``_id`` are automatically managed
```
feanil, line 50: You said `_id` in the line above this as well.

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+"""
+
+EXAMPLES = '''
+- name: Basic example
+  mongodb_replica_set:
+    username: root
+    password: password
+    rs_config:
+      members:
+        - host: some.host
```
feanil, line 63: in this case, is there a default port or does it assume that a port is provided as a part of the host name?

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+          host: other.host:27017
+          hidden: true
+'''
+# Magic import
+from ansible.module_utils.basic import *
+
+try:
+    from pymongo import MongoClient
+    from pymongo.errors import OperationFailure
+    from bson import json_util
```
feanil, line 88: Should we do a version check here since 2.x of pymongo and 3.x of pymongo are not compatible?

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+        if changed_rs_config != rs_config:
+            module.fail_json(msg="Failed to validate that the replica set was changed", config=rs_config)
+
+    module.exit_json(changed=changed, config=rs_config)
+
+
+######### Client making stuff #########
+def get_mongo_uri(host, port, username, password, auth_database):
+    mongo_uri = 'mongodb://'
+    if username and password:
```
feanil, line 244: We should error out if only username or only password are provided.

playbooks/library/mongodb_replica_set
maxrothman, line 50: Document ID vs member IDs. They're two separate fields.

feanil:
There are some triple quoated docstrings and some double quoted docstrings.  We should consintently use `'''`(three single quotes).

playbooks/library/mongodb_replica_set
maxrothman, line 63: It defaults to port 27017, as reflected in the "port" field doc.

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+def reconfig_replset(rs_config):
+    try:
+        client.admin.command("replSetReconfig", rs_config, force=module.params['force'])
+    except OperationFailure as e:
+        raise
+        module.fail_json(msg="Failed to reconfigure replSet: {}".format(e.message))
+
+def get_rs_config_id():
+    try:
+        replset_name = client.admin.command('getCmdLineOpts')['parsed']['replication']['replSetName']
```
feanil: Replica set name can also be provided in the config file instead of at the command line.  Will this handle that correctly?  This seems to be specifically getting command-line parsed options.

playbooks/library/mongodb_replica_set
maxrothman, line 244: That check is done in `validate_args()`. Would you prefer it move here? I feel like that error case is specific to the ansible module behavior, not to getting a URI.

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+        module.fail_json(msg="Unable to get replSet name. Was mongod started with --replSet? Error: " + e.message)
+
+
+########### Helper functions ###########
+def set_member_ids(members, old_members=None):
+    '''
+    Set the _id property of members who don't already have one.
+    Prefer the _id of the "matching" member from `old_members`.
+    '''
+    available_ids = set(range(len(members)*2))
```
feanil, line 140: Why times 2? Add a comment?

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,295 @@
...
+    if old_rs_config is None:
+        changed = True
+        if '_id' not in rs_config:
+            rs_config['_id'] = get_rs_config_id(client)  #Errors if no replSet specified to mongod
+        rs_config['members'] = set_member_ids(rs_config['members'])  #Noop if all _ids are set
+        #Don't set the version, it'll auto-set
+        initialize_replset(client, rs_config)
+
+    else:
+        rs_config_non_collections = {k:v for k,v in rs_config.items()
```
feanil: There is too much logic compressed into this comprehension. Pull it out and add some comments here.

feanil:
Overall approach makes sense.  Have you tested actually making changes to a mongo replicaset using this?

playbooks/library/mongodb_replica_set
maxrothman: Nope, the config file puts the property in the same place. The error mentions the command-line arg because that's how all of mongodb's errors look. I'm happy to change it though.

playbooks/library/mongodb_replica_set
maxrothman, line 140: Just so it's impossible to run out of IDs. I'll make a comment.

playbooks/library/mongodb_replica_set
maxrothman: I disagree. The variable name describes the intent of the comprehension (to get all the non-collections) and the comprehension itself is only a couple filters. I'm happy to add a comment though.

maxrothman:
I have not yet done a functional test, but I thought it'd be more time-effective to test it when testing the mongo_3_0 role.

playbooks/library/mongodb_replica_set
feanil: I think we should change it to be more clear even though mongodb isn't.

playbooks/library/mongodb_replica_set
feanil: I don't know what that means.  I think you know what it means because of context but more comments would be good place to start.  Multi-line dict comprehension w/multiple if statements are not easy to grok in my opinion.

maxrothman: I felt it was simple because the _only_ operation the comprehension did was 2 if statements, (i.e. it's the same as `list.filter().filter()`) but I'll add a comment.

playbooks/library/mongodb_replica_set
maxrothman, line 88: Is it incompatible? AFAIK they support the same API.

feanil, line 88: They don't have the same API, there are differences.  From the [PyMongo guide for migrating to PyMongo 3.x](https://api.mongodb.org/python/2.9/migrate-to-pymongo3.html):

> PyMongo 3 is a partial rewrite bringing a large number of improvements. It also brings a number of backward breaking changes.

playbooks/library/mongodb_replica_set
feanil, line 244: No it's fine where it is,  missed it during my first review.

playbooks/library/mongodb_replica_set
```diff
@@ -0,0 +1,304 @@
...
+        if '_id' not in rs_config:
+            rs_config['_id'] = get_rs_config_id(client)  #Errors if no replSet specified to mongod
+        rs_config['members'] = set_member_ids(rs_config['members'])  #Noop if all _ids are set
+        #Don't set the version, it'll auto-set
+        initialize_replset(client, rs_config)
+
+    else:
+        rs_config_scalars = {k:v for k,v in rs_config.items()
+            if not isinstance(v, list) if not isinstance(v, dict)}
+        old_rs_config_non_collections = {k:v for k,v in old_rs_config.items()
```
feanil: scalers is a much better name for this, please update this to `old_rs_config_scalers`.

Also, instead of multiple if statemets, just take advantage of the fact that `isinstance` can have a [tuple as the second argument](https://docs.python.org/2/library/functions.html#isinstance).

playbooks/library/mongodb_replica_set
maxrothman, line 88: We require pymongo 3.1 in configuration's `requirements.txt`. I realize in retrospect that I wrote this reading the pymongo 2.7 docs while using 3.1, so it's most likely ok. Given that the simple APIs I'm using here didn't change, I highly doubt it'll be an issue. Still, it'd be good to test with both.

EDIT: after reviewing the page you linked to, I strongly doubt this module would have any issues running on pymongo 2.x. It only uses `MongoClient` with a raw URI, `find_one` with no arguments, and `command`. All of those behave consistently across both versions.

jibsheet:
At this point, we consider this code reviewed and I've rebased https://github.com/edx/configuration/pull/2587/commits onto it to pick up the fixes.  I don't think we're planning to merge this separately, so closing.  We can reopen if we want to merge later.

PR #2586, opened 2015-12-11
Small fixes

maxrothman:
@edx/devops

### DIFF
--- a/playbooks/library/util_map
+++ b/playbooks/library/util_map
@@ -17,7 +17,7 @@ options:
     description:
       - The input
     required: true
-  args
+  args:
     description:
       - Arguments to the function other than the input, varies by function.
 """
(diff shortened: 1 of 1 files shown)


feanil:
:+1:

playbooks/library/util_map
```diff
@@ -159,7 +159,7 @@ def main():
 
     arg_spec = dict(
         function=dict(required=True, type='str'),
-        input=dict(required=True, type='str'),
+        input=dict(required=True),
```
jibsheet, line 162: With this change, this
https://github.com/edx/configuration/blob/master/playbooks/edx-east/edx_service.yml#L102
Results in a blank input being passed along and us getting Invalid String

jibsheet, line 162: We figured out that removing type='str' allowed things like [] to come through literally, and when we eval them, it dies because it isn't a string.  The type='str' was coercing into something eval'able.

We should remove both the type='str' and on line 169 the ast.literal_eval, but we're likely still passing a string somewhere, so will need to test. @maxrothman  is going to make a story.

PR #2588, opened 2015-12-11
Elasticsearch 1.5

fredsmith:


cpennington:
We're using 2.1? I thought we were using the amazon hosted 1.5?

feanil:
Good point.  Fred will update this to match to the version of ES that AWS
supports.

playbooks/roles/elasticsearch/templates/etc/default/elasticsearch.j2
```diff
@@ -34,8 +34,5 @@ DATA_DIR={{ COMMON_DATA_DIR }}
 # ElasticSearch configuration directory
 CONF_DIR={{ elasticsearch_cfg_dir }}
 
-# ElasticSearch configuration file (elasticsearch.yml)
-CONF_FILE=${CONF_DIR}/elasticsearch.yml
```
feanil, line 38: Why was this removed?

feanil:
Changes look good, but are using the old style module syntax instead of YAML syntax.  Can you update it to use the new YAML based syntax?

playbooks/roles/elasticsearch/templates/etc/default/elasticsearch.j2
fredsmith, line 38: newer versions of elasticsearch no longer support config files with names other than elasticsearch.yml, and so setting this variable causes it to throw an error.  the conf_dir is still configurable, though.

feanil:
:+1:  once the task syntax uses the new style.

fredsmith:
Updated to yaml syntax, but can't fully test the play because of python 2.7

PR #2590, opened 2015-12-14
Benp/break out tests

benpatterson:
@jzoldak @jibsheet 

This will help when reading packer output. Currently, there is just a general 'validate build environment' step that takes a very long time. Here, we'll provide a few more hints to the end-user, to get a sense of which step is being run. It also makes reading failures easier, since it's very clear which particular class of smoketests failed.

### DIFF
--- a/playbooks/roles/test_build_server/files/test-development-environment.sh
+++ b/playbooks/roles/test_build_server/files/test-development-environment.sh
@@ -34,6 +34,7 @@ case "$1" in
         paver test_system -t lms/djangoapps/courseware/tests/tests.py
         paver test_system -t cms/djangoapps/course_creators/tests/test_views.py
         ;;
+
     "js")
 
         # Run some of the javascript unit tests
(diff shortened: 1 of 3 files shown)


jibsheet:
I like it :+1: Although I'd want @jzoldak 's thoughts as well since he's more familiar with the testing than I am (the ansible change makes sense).

playbooks/roles/test_build_server/files/test-development-environment.sh
```diff
@@ -20,28 +20,46 @@ cd edx-platform-clone
...
+        # Generate some coverage reports
+        paver coverage
 
-# Run some of the javascript unit tests
-paver test_js_run -s xmodule
+        # Run some of the djangoapp unit tests
+        paver test_system -t lms/djangoapps/courseware/tests/tests.py
+        paver test_system -t cms/djangoapps/course_creators/tests/test_views.py
+        ;;
+    "js")
```
jzoldak, line 38: nit: this could use an empty line above

jzoldak:
I like it!
:+1:

benpatterson:
Updated with Jay's suggestion. Merging.

PR #2592, opened 2015-12-15
upgrade jdk to version 1.8

fredsmith:


feanil:
@benpatterson will this PR cause any trouble with your jenkins deploys?  I seem to remember some issue with jenkins and java 1.8.  It might be more prudent to leave the default as is and override in the meta/main where we need to jump forward?

benpatterson:
@feanil Ah, yes, after we'd heard java 8 wasn't quite up to par based on your experience, we went with Java 7 for the master node. Indeed this change would cause us issues; however, we haven't used the full ansible script on jenkins master in awhile. If you have the time to update either the jenkins_master role or the edx-east/testeng_jenkins_master.yml playbook with the appropriate overrides (we are using these overrides for jenkins_workers), then that would just help us avoid churn down the road.

benpatterson:
So: would be nice to update our plays at the same time, but not a must.

feanil:
:+1:   @fredsmith up to you if you have time to update the `jenkins_master/meta/main.yml` to explicitly use 1.7 please do.

fredsmith:
@feanil @benpatterson https://github.com/edx/configuration/pull/2599

PR #2593, opened 2015-12-15
Remove unused setting from edxapp configuration

rlucioni:
Corresponds to https://github.com/edx/edx-platform/pull/10963. @ormsbee, since you reviewed https://github.com/edx/edx-platform/pull/10963, would you mind taking a quick look at this, too?


jibsheet:
These are set to true in prod/stage/loadtest/matasano - so we'll want to pull them out of there after this clears.

ormsbee:
I'm surprised it's set to true. @chrisndodge: This config isn't used for microsites too, is it?

chrisndodge:
I'm OOO today so it's hard for me to comment/investigate

jibsheet:
@ormsbee It's probably historical - they date back to mid-2014 and a big config reord.

rlucioni:
The only part of edx-platform which used this setting was deleted when [microsites was introduced](https://github.com/edx/edx-platform/commit/a3211a7405d92d09d59d3e2fe9ef437051324bcb#diff-f94db9e2a89aa376923087ddc5e85515L23) in 2013. @ormsbee @jibsheet good to merge?

ormsbee:
:+1:

PR #2594, opened 2015-12-15
OPS-1208 Refactor Nginx SSL redirect bahavior

arbabnazar:
@edx/devops kindly review it. Thanks

### DIFF
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -11,6 +11,7 @@ NGINX_EDXAPP_EXTRA_CONFIGS: []
 NGINX_EDXAPP_CUSTOM_REDIRECTS: {}
 
 NGINX_ENABLE_SSL: False
+NGINX_REDIRECT_TO_HTTPS: False
 # Set these to real paths on your
 # filesystem, otherwise nginx will
 # use a self-signed snake-oil cert
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -51,21 +51,37 @@ server {
 error_page {{ k }} {{ v }};
   {% endfor %}
 
+  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
 
   {% if NGINX_ENABLE_SSL %}
-
-  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
   listen {{ EDXAPP_LMS_SSL_NGINX_PORT }} {{ default_site }} ssl;
 
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
-
-  {% else %}
-  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
   {% endif %}
 
+  {% if NGINX_REDIRECT_TO_HTTPS %}
+  # Redirect http to https over single instance
+  if ($scheme != "https") 
+  { 
+   set $do_redirect_to_https "true";
+  }
+
+  # Nginx does not support nested conditions 
+  # Forward to HTTPS if we're an HTTP request... and the server is behind ELB 
+  if ($http_x_forwarded_proto = "http") 
+  {
+   set $do_redirect_to_https "true";
+  }
+
+  if ($do_redirect_to_https = "true")
+  {
+  rewrite ^ https://$host$request_uri? permanent;
+  }
+  {% endif %}
+  
   access_log {{ nginx_log_dir }}/access.log {{ NGINX_LOG_FORMAT_NAME }};
   error_log {{ nginx_log_dir }}/error.log error;
 
@@ -185,19 +201,4 @@ location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
   {% include "robots.j2" %}
   {% include "static-files.j2" %}
 
- {% if NGINX_SET_X_FORWARDED_HEADERS %}
-   if ($scheme != "https") {
-    rewrite ^ https://$host$uri permanent;
-   }
- {% else %}
-  # Forward to HTTPS if we're an HTTP request...
-  if ($http_x_forwarded_proto = "http") {
-    set $do_redirect "true";
-  }
-
-  # Run our actual redirect...
-  if ($do_redirect = "true") {
-    rewrite ^ https://$host$request_uri? permanent;
-  }
- {% endif %}
 }
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -174,6 +174,7 @@ COURSE_DISCOVERY_SSL_NGINX_PORT: 443
 COURSE_DISCOVERY_VERSION: $course_discovery_version
 
 NGINX_SET_X_FORWARDED_HEADERS: True
+NGINX_REDIRECT_TO_HTTPS: True
 EDX_ANSIBLE_DUMP_VARS: true
 migrate_db: "yes"
 openid_workaround: True
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
@@ -30,20 +30,38 @@ server {
 error_page {{ k }} {{ v }};
   {% endfor %}
 
+  listen {{ EDXAPP_CMS_NGINX_PORT }} {{ default_site }};
+
   {% if NGINX_ENABLE_SSL %}
 
-  listen {{ EDXAPP_CMS_NGINX_PORT }} {{ default_site }};
   listen {{ EDXAPP_CMS_SSL_NGINX_PORT }} ssl;
 
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
-
-  {% else %}
-  listen {{ EDXAPP_CMS_NGINX_PORT }} {{ default_site }};
   {% endif %}
+  
+  {% if NGINX_REDIRECT_TO_HTTPS %}
+  # Redirect http to https over single instance
+  if ($scheme != "https") 
+  { 
+   set $do_redirect_to_https "true";
+  }
 
+  # Nginx does not support nested conditions 
+  # Forward to HTTPS if we're an HTTP request... and the server is behind ELB 
+  if ($http_x_forwarded_proto = "http") 
+  {
+   set $do_redirect_to_https "true";
+  }
+
+  if ($do_redirect_to_https = "true")
+  {
+  rewrite ^ https://$host$request_uri? permanent;
+  }
+  {% endif %}
+  
   server_name {{ CMS_HOSTNAME }};
 
   access_log {{ nginx_log_dir }}/access.log {{ NGINX_LOG_FORMAT_NAME }};
@@ -91,19 +109,4 @@ error_page {{ k }} {{ v }};
   {% include "robots.j2" %}
   {% include "static-files.j2" %}
 
-  {% if NGINX_SET_X_FORWARDED_HEADERS %}
-   if ($scheme != "https") {
-    rewrite ^ https://$host$uri permanent;
-   }
- {% else %}
-  # Forward to HTTPS if we're an HTTP request...
-  if ($http_x_forwarded_proto = "http") {
-    set $do_redirect "true";
-  }
-
-  # Run our actual redirect...
-  if ($do_redirect = "true") {
-    rewrite ^ https://$host$request_uri? permanent;
-  }
- {% endif %}
 }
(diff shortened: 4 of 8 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -61,9 +61,24 @@ error_page {{ k }} {{ v }};
...
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+  
+  # Redirect http to https over signle instance
+  if ($scheme != "https") {
+    rewrite ^ https://$host$uri permanent;
+  }
 
   {% else %}
   listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
```
arbabnazar: @e0d we need it, because in case of server behild ELB come to `else` block and get this listen value.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -61,9 +61,24 @@ error_page {{ k }} {{ v }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+
```
arbabnazar: logic is really simple, `NGINX_ENABLE_SSL` is `True` for sandbox and will enter to `if` block and set all the required value and we don't need to compare it with the `NGINX_SET_X_FORWARDED_HEADERS`.
but for the standalone server/server behind ELB which don't want to use the SSL will enter to `else` block and get all the value from there and work as expected, before making this [change](https://github.com/edx/configuration/pull/2487).

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -185,19 +200,4 @@ location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
   {% include "robots.j2" %}
   {% include "static-files.j2" %}
 
- {% if NGINX_SET_X_FORWARDED_HEADERS %}
```
fredsmith, line 188: does this mean that this behavior can't be switched off anymore?

fredsmith:
LGTM :+1:

feanil:
I think this is not quite right.  With your change as it is now, even if someone wanted to turn off SSL they would get redirected to https which is not ideal for users that don't want to use SSL.  I think what we need is two flags `NGINX_ENABLE_SSL` which will setup SSL certs on the box running NGINX, and `NGINX_REDIRECT_TO_HTTPS` which will determine whether or not HTTP requests should redirect to HTTPS.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -61,9 +61,23 @@ error_page {{ k }} {{ v }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
-
+  
+  {% if NGINX_REDIRECT_TO_HTTPS %}
```
arbabnazar: @feanil I have introduced the `NGINX_REDIRECT_TO_HTTPS` on which we shall decide whether or not HTTP requests should redirect to HTTPS, while `NGINX_ENABLE_SSL` will check that we want ssl or not. Thanks

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -61,9 +61,23 @@ error_page {{ k }} {{ v }};
...
+    rewrite ^ https://$host$uri permanent;
+  }
+  {% endif %}
+  
   {% else %}
+
   listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
+
+  # Forward to HTTPS if we're an HTTP request... and the server is behind ELB
+  if ($http_x_forwarded_proto = "http") {
```
arbabnazar: @edx/devops, can we combine these two `if` statements into one like this, because first `if` statement is just setting the value to `true` for second `if` statement, so we can directly setting the value of second `if` statement into first.
I think, even we are not passing the value of `do_redirect` as extra variable, so we can safely remove it? Thanks

```
  # Forward to HTTPS if we're an HTTP request...
   if ($http_x_forwarded_proto = "http") {
    set $do_redirect "true";
  }

  # Run our actual redirect...
  if ($do_redirect = "true") {
    rewrite ^ https://$host$request_uri? permanent;
   }
```

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
e0d: @feanil  @arbabnazar I think this simplifies the logic and covers everything we need, what do you think:

```
  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};

  {% if NGINX_ENABLE_SSL %}
  listen {{ EDXAPP_LMS_SSL_NGINX_PORT }} {{ default_site }} ssl;

  ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
  ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
  # request the browser to use SSL for all connections
  add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
  {% endif %}

  {% if NGINX_REDIRECT_TO_HTTPS %}
  # Redirect http to https over signle instance
  if ($scheme != "https") {
    rewrite ^ https://$host$uri permanent;
  }
  {% endif %}
```

feanil: :+1:   Yes, this is easy to read and separates the two pieces of logic cleanly.  @arbabnazar can you update to match the code @e0d provided.

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -11,6 +11,7 @@ NGINX_EDXAPP_EXTRA_CONFIGS: []
 NGINX_EDXAPP_CUSTOM_REDIRECTS: {}
 
 NGINX_ENABLE_SSL: False
+NGINX_REDIRECT_TO_HTTPS: False
```
arbabnazar, line 14: @edx/devops added `NGINX_REDIRECT_TO_HTTPS: False` to nginx role's defaults.

arbabnazar:
@e0d @feanil please review it, so that I'll modify all other templates as well. Thanks

e0d:
@smarnach please chime in as to whether this will meet your usecase.

feanil:
@arbabnazar this is not quite what @e0d previously mentioned.  Because of the if/else you are still coupling the two features(ssl enabled and redirect to https).

```
  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};

  {% if NGINX_ENABLE_SSL %}
  listen {{ EDXAPP_LMS_SSL_NGINX_PORT }} {{ default_site }} ssl;

  ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
  ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
  # request the browser to use SSL for all connections
  add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
  {% endif %}

  {% if NGINX_REDIRECT_TO_HTTPS %}
  # Redirect http to https over single instance
  if ($scheme != "https") {
    rewrite ^ https://$host$uri permanent;
  }
  {% endif %}
```

They should be two seperate `if` statements. If you need both you turn on both independently.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -51,19 +51,27 @@ server {
...
   {% if NGINX_ENABLE_SSL %}
-
-  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
   listen {{ EDXAPP_LMS_SSL_NGINX_PORT }} {{ default_site }} ssl;
 
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+  {% endif %}
```
arbabnazar, line 63: @feanil I did it in two separate `if` statements, the one end here enable the ssl certificate.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -51,19 +51,27 @@ server {
...
 
+  {% if NGINX_REDIRECT_TO_HTTPS %}
```
arbabnazar, line 65: @feanil other `if` statement start from here and it checks whether redirect to https or not.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -51,19 +51,27 @@ server {
...
+  # Redirect http to https over signle instance
+  if ($scheme != "https") {
+    rewrite ^ https://$host$uri permanent;
```
smarnach: This doesn't seem to redirect to the correct port for a single instance.

util/jenkins/ansible-provision.sh
```diff
@@ -177,6 +177,7 @@ COURSE_DISCOVERY_SSL_NGINX_PORT: 443
 COURSE_DISCOVERY_VERSION: $course_discovery_version
 
 NGINX_SET_X_FORWARDED_HEADERS: True
+NGINX_REDIRECT_TO_HTTPS: True
```
smarnach, line 177: If we make the default `True` here, the [one-step installation for the Open edX fullstack](https://github.com/edx/configuration/wiki/edX-Ubuntu-12.04-64-bit-Installation) will still result in an unusable server, since the default for `NGINX_ENABLE_SSL` is `False`.

smarnach:
@e0d Thanks for pinging me.  Unfortunately, I don't really have time to test this (I'm on parental leave for a few weeks), but I added two comments.  Basically, if the one-step installation for the Open edX fullstack works again, I'm happy with the outcome.

smarnach:
@mtyaka @bradenmacdonald Does one of you have time to look into this in more detail?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -51,19 +51,27 @@ server {
...
+  }
   {% else %}
-  listen {{ EDXAPP_LMS_NGINX_PORT }} {{ default_site }};
+  # Forward to HTTPS if we're an HTTP request... and the server is behind ELB
+  if ($http_x_forwarded_proto = "http") {
```
feanil: I think it's this statement that confused me, because it tries to assume that you want to go to SSL if you're behind an ELB.  I think we should remove this `else` section.

util/jenkins/ansible-provision.sh
arbabnazar, line 177: @smarnach we have set it to `NGINX_REDIRECT_TO_HTTPS: False` inside the role defaults at `playbooks/roles/nginx/defaults/main.yml`. The value above that you have pointed is passing as extra variable, not the default.Thanks

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -51,19 +51,27 @@ server {
...
   listen {{ EDXAPP_LMS_SSL_NGINX_PORT }} {{ default_site }} ssl;
 
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   # request the browser to use SSL for all connections
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+  {% endif %}
 
+  {% if NGINX_REDIRECT_TO_HTTPS %}
+  # Redirect http to https over signle instance
```
mtyaka: Typo "signle" -> "single".

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
mtyaka: I agree, we should respect the value of `NGINX_REDIRECT_TO_HTTPS` even when deployed behind a load balancer.

util/jenkins/ansible-provision.sh
smarnach, line 177: @arbabnazar You are right, thanks.  I really looked at this in a hurry, so I didn't read it carefully enough.

smarnach:
There is still an open comment here: https://github.com/edx/configuration/pull/2594#discussion_r47952969.  Basically, I think that this will not redirect to the right port if deployed on a single instance.

util/jenkins/ansible-provision.sh
arbabnazar, line 177: @e0d @edx/devops @smarnach I have verified and it's worked for [one-step installation for the Open edX fullstack](https://github.com/edx/configuration/wiki/edX-Ubuntu-12.04-64-bit-Installation) perfectly. Thanks

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
e0d: We do not support the ability to redirect when https is listening on a port other than 443 and there is no upstream proxy that is. However this wasn't supported in prior versions either.  Open to  it as a separate PR.  Don't think it should block merging this.

e0d:
Please verify config is in place for our deployments.

:+1: when that is the case.

e0d:
Rebase and :+1: to merge.

feanil:
Looks like something weird happened during the rebase.  Did you rebase from master?

arbabnazar:
@feanil all set now. Thanks

PR #2596, opened 2015-12-15
Remove un-necessary directory permission change.

feanil:
The var it points to is actually the github URL and not the location of the source code. Since
the code has continued to work without these permissions being applied to the code directory,
I think this is un-necessary and should just be deleted.

@edx/devops


fredsmith:
:+1:

fredsmith:
```
$ pwd
/home/ubuntu/https:/github.com/edx/notifier.git
```

jibsheet:
:+1:

PR #2600, opened 2015-12-16
Send front-end PageView custom metrics to New Relic Insights.

ormsbee:
@maxrothman, @feanil: May I get a review from someone on your team?

We're emitting custom attributes to record the `course_id` and `org` for courseware views, and those show up in the server side transactions on NR Insights. However, they're not forwarded on through to PageViews. [This doc](https://docs.newrelic.com/docs/agents/python-agent/installation-configuration/python-agent-configuration#browser-settings) leads me to believe it's because `browser_monitoring.attributes.enabled` defaults to false:

>  This setting can be used to turn on or off all attributes for browser monitoring. This is the data which gets sent to page views in Insights.

Though it's not entirely clear to me because [this doc](https://docs.newrelic.com/docs/insights/new-relic-insights/decorating-events/insights-custom-attributes#adding-attributes) says that we need to modify `browser_monitoring.capture_attributes` (maybe just outdated docs?)

FYI: @benpatterson, @tobz

### DIFF
--- a/playbooks/roles/newrelic/templates/etc/newrelic/nrsysmond.cfg.j2
+++ b/playbooks/roles/newrelic/templates/etc/newrelic/nrsysmond.cfg.j2
@@ -231,3 +231,9 @@ labels={{ labels }}
 #          path if, for example, you have multiple CGroup mounts.
 # Default: Distribution dependent, but usually /sys/fs/cgroup or /cgroup.
 #cgroup_root=
+
+# Allow custom attributes to be sent to Insights for browser PageViews
+#
+# https://docs.newrelic.com/docs/agents/python-agent/installation-configuration/python-agent-configuration#browser-settings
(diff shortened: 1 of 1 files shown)


feanil:
:+1:

ormsbee:
So the sad thing is that I don't really know how to test this except to put it in staging... any thoughts?

benpatterson:
We could request a sandbox with new relic enabled. I think that'd work.

sent via Android

ormsbee:
@benpatterson: the problem is that we only bought the browser functionality for the specific apps on staging and prod, so I don't think a sandbox would send that data along anyway, even if properly configured.

@feanil: does it make sense to just try this on staging?

ormsbee:
Got the ok from @maxrothman to merge.

PR #2602, opened 2015-12-16
Compile Programs Sass prior to running r.js and cache static assets

rlucioni:
Allows r.js to minify CSS hosted by the Programs service. This in turn yields minified files with unhashed names we can use to fetch these assets from Studio. ECOM-2862.

Corresponds to https://github.com/edx/programs/pull/69. @jimabramson please review. I'd also appreciate a look from someone in @edx/devops. @AlasdairSwan FYI.

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
@@ -25,7 +25,7 @@ server {
 
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
-  # request the browser to use SSL for all connections
+  # Request that the browser use SSL for all connections.
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
 
   {% else %}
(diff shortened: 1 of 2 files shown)


rlucioni:
I've modified Programs' Nginx configuration to allow static assets to be held in public caches for an hour.

At the moment, the only static assets served by Programs are the JS and CSS required for the Programs authoring app, an internal tool for configuring XSeries. Given that, an hour's lag between deployment and when users may actually see changes to the app is acceptable. Programs CSS and JS are requested at the same time, so I expect these to remain in sync when cached.

@feanil, since we discussed this offline today, I'd appreciate a review from you.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
```diff
@@ -35,6 +35,15 @@ server {
   location ~ ^/static/(?P<file>.*) {
     root {{ COMMON_DATA_DIR }}/{{ programs_service_name }};
     try_files /staticfiles/$file =404;
+
+    # Request that the browser use SSL for these connections. Repeated here
+    # beacuse add_header directives are only inherited from the previous level
```
jimabramson: typo (because)

feanil:
:+1:  once the typo is fixed.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
rlucioni: Fixed.

jimabramson:
:+1:

PR #2606, opened 2015-12-17
Don't mount directories we aren't going to use.

nedbat:


nedbat:
@mulby @feanil devstack doesn't use insights, so why mount the directories just to leave them empty?  I tested the base build with this change, it completed fine.

mulby:
:+1:

clintonb:
:+1:

feanil:
:+1:

PR #2607, opened 2015-12-17
updating refs, use external script

e0d:
@edx/devops @cpennington 

Simplifies the common builds by using a shared external script.  Updates refs to master now that everything has merged.

### DIFF
--- a/docker/build/precise-common/Dockerfile
+++ b/docker/build/precise-common/Dockerfile
@@ -2,53 +2,8 @@ FROM ubuntu:precise
 MAINTAINER edxops 
 ENV ANSIBLE_REPO="https://github.com/edx/ansible"
 ENV CONFIGURATION_REPO="https://github.com/edx/configuration.git"
-ENV CONFIGURATION_VERSION="e0d/docker-artifacts"
+ENV CONFIGURATION_VERSION="master"
 
-# system bootstrap
-RUN apt-get update
-RUN apt-get -y install software-properties-common python-software-properties
-RUN add-apt-repository ppa:fkrull/deadsnakes-python2.7
-RUN apt-get update
-RUN apt-get -y install sudo
-RUN useradd docker && echo "docker:docker" | chpasswd
-RUN echo "docker ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
-RUN mkdir -p /home/docker && chown -R docker:docker /home/docker
-RUN apt-get install -y \
-    git \
-    libmysqlclient-dev \
-    python-apt \
-    python-jinja2 \
-    python-pip \
-    python-yaml \
-    python2.7 \
-    python2.7-dev
-
-# Temporary hacking related to an SELinux bug.  This issue causes, at least,
-# useradd to fail silently when the -m flag is passed in.  The bug affects
-# Ubuntu precise and is tracked here:
-#
-# https://bugs.launchpad.net/ubuntu/+source/libselinux/+bug/1424795
-RUN apt-get install wget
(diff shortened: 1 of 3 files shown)


docker/build/precise-common/Dockerfile
```diff
@@ -2,53 +2,8 @@ FROM ubuntu:precise
...
-RUN sudo git checkout ${CONFIGURATION_VERSION}
-RUN sudo pip install -r pre-requirements.txt
-RUN sudo pip install -r requirements.txt
-WORKDIR /tmp/configuration/playbooks/edx-east
-RUN sudo /tmp/ansible/bin/ansible-playbook edx_ansible.yml -i '127.0.0.1,' -c local -e "configuration_version=${CONFIGURATION_VERSION}"
-WORKDIR /edx/app/edx_ansible
-# cleanup
-RUN sudo rm -rf /tmp/ansible
-RUN sudo rm -rf /tmp/configuration
+ADD https://raw.githubusercontent.com/edx/configuration/master/util/install/ansible-bootstrap.sh /tmp/ansible-bootstrap.sh
```
cpennington: Can we `ADD` this from the local directory, so that it's possible to actually do local testing/debugging?

cpennington: Especially with the changes in the #2574, which would let us do pushes from CI (so the risk of a dirty local repo is significantly less).

e0d: @cpennington this isn't really possible.
`ADD` doesn't all fq or relative references to files outside the the build context and also doesn't support symlinks.  It does support URLs but not file URLs, e.g., `file:///tmp/file.txt`

For debugging, though a pain, I think the best we can do is bash onto a bare server and run the file locally.

It's a pain, but I think the reuse is worth the trade-off.  Obviously providing access outside the build directory is not hard to do, but has significant security implications for folks building from provided Dockerfiles.

feanil:
Seems like a good incremental improvement. :+1:

e0d:
@cpennington updated to use `-f` strategy.

docker/build/precise-common/Dockerfile
```diff
@@ -2,53 +2,8 @@ FROM ubuntu:precise
 MAINTAINER edxops 
 ENV ANSIBLE_REPO="https://github.com/edx/ansible"
 ENV CONFIGURATION_REPO="https://github.com/edx/configuration.git"
-ENV CONFIGURATION_VERSION="e0d/docker-artifacts"
+ENV CONFIGURATION_VERSION="master"
```
cpennington, line 5: Do you need this anymore?

e0d, line 5: For now, yes.  May pull your suggestion in from the other PR later.

cpennington, line 5: Oh, is that environment variable used inside the `ansible-bootstrap.sh` script?

e0d, line 5: Yes.

e0d:
@cpennington OK to merge?

cpennington:
:+1:

docker/build/precise-common/Dockerfile
```diff
@@ -2,53 +2,8 @@ FROM ubuntu:precise
...
 
-# system bootstrap
-RUN apt-get update
-RUN apt-get -y install software-properties-common python-software-properties
-RUN add-apt-repository ppa:fkrull/deadsnakes-python2.7
-RUN apt-get update
-RUN apt-get -y install sudo
-RUN useradd docker && echo "docker:docker" | chpasswd
-RUN echo "docker ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
-RUN mkdir -p /home/docker && chown -R docker:docker /home/docker
```
cpennington, line 15: I just noticed that you removed the stuff that set up the docker user. Was that deliberate? Should our dockerfiles switch to just using the root user?

PR #2608, opened 2015-12-17
Don't need to sync folders when building the base box.

feanil:


feanil:
@edx/devops @nedbat

jibsheet:
:+1:

PR #2610, opened 2015-12-18
Mongo 3.0.8 includes a fix for data loss

jibsheet:
https://groups.google.com/forum/#!topic/mongodb-announce/mBesDEtR5l0

Additionally, it includes the fix for TOOLS-939 so we should not need to
pin to 3.0.6 any more.

@edx/devops


e0d:
:+1:

PR #2613, opened 2015-12-18
Add Elasticsearch overrides to Course Discovery for production and do…

cpennington:
…cker

Pairs with https://github.com/edx/course-discovery/pull/11

@e0d @feanil @clintonb: Review?


playbooks/roles/course_discovery/defaults/main.yml
```diff
@@ -37,6 +37,13 @@ COURSE_DISCOVERY_DATABASES:
     ATOMIC_REQUESTS: true
     CONN_MAX_AGE: 60
 
+COURSE_DISCOVERY_ELASTICSEARCH: 'localhost'
+COURSE_DISCOVERY_ES_INDEX: 'course_discovery'
+
+COURSE_DISCOVERY_ELASTICSEARCH:
```
clintonb: You will also want to define the boolean `connect_on_startup`. I'm on the fence as to whether this should be enabled by default.

clintonb:
:+1: (once `connect_on_startup` added)

playbooks/roles/course_discovery/defaults/main.yml
cpennington: I made a comment on hipchat, but I think we probably shouldn't be connecting on startup, but instead have it on demand, so that things like collecting static assets will work without ES.

cpennington: I also think that index-creation shouldn't happen as part of app startup.

cpennington: I've moved index creation to be part of `make migrate`, so `connect_on_startup` is no longer needed.

e0d:
:+1:

feanil:
:+1:

PR #2614, opened 2015-12-18
Auto confirm for gcc/g++ ppa

arbabnazar:
@edx/devops `sandbox.sh` hang at the step when it add the gcc/g++ ppa for user respone(yes/enter), you make it better to make it auto install without user interaction. Thanks

### DIFF
--- a/util/install/sandbox.sh
+++ b/util/install/sandbox.sh
@@ -20,7 +20,7 @@ fi
 ## Set ppa repository source for gcc/g++ 4.8 in order to install insights properly
 ##
 sudo apt-get install -y python-software-properties
-sudo add-apt-repository ppa:ubuntu-toolchain-r/test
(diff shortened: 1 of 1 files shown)


util/install/sandbox.sh
```diff
@@ -11,7 +11,7 @@
 ##
 ## Sanity check
 ##
-if [[ ! "$(lsb_release -d | cut -f2)" =~ $'Ubuntu 12.04' ]]; then
+if[[ ! "$(lsb_release -d | cut -f2)" =~ $'Ubuntu 12.04' ]]; then
```
arbabnazar: there is an extra space between `if` and `[[` which produce this error:
```shell
sandbox.sh: 14: ./sandbox.sh: [[: not found
```

e0d:
:+1:

PR #2616, opened 2015-12-19
add credentials for devstack

zubair-arbi:
ECOM-3180


e0d:
@zubair-arbi is this PR something we want to move forward with?

zubair-arbi:
@e0d I believe not.
@cpennington can you please confirm if we need to add new services to devstack or just make them work with docker?

cpennington:
No, I don't think we should add to the vagrant-devstack.

PR #2625, opened 2015-12-22
do the same action(ppa adding) using with_items

arbabnazar:
@edx/devops kindly review this change, I have verified it against my sandbox creation job. Thanks


feanil:
:+1:

PR #2626, opened 2015-12-22
These should pull in the vhost virtualization role

e0d:
@nedbat this should address the banner issue.


nedbat:
:+1:, though I have to make another fix to get back the "Open edX" banner.

PR #2628, opened 2015-12-22
We still need to coerce into a string

jibsheet:
Later, zip_to_dict calls ast.literal_eval which requires this be a
string.  We can remove this _and_ the literal_eval call, however that
would require testing all the places we call these util functions.

@maxrothman


maxrothman:
:+1: :sob:

PR #2629, opened 2015-12-23
elasticsearch module

fredsmith:


terraform/elasticsearch/.gitignore
```diff
@@ -0,0 +1,4 @@
+# Do not accidentally check in a file with secret info e.g. AWS credentials
```
feanil: we should move this gitignore up one level so we don't do this anywhere under the terraform directory.

feanil:
:+1: once the .gitignore is moved.

PR #2630, opened 2015-12-23
Move the default location for the locust requiremnts files.

feanil:
There is a corresponding change in the repo so the public
and private repos have the requirements file in the same
location.

@fredsmith 

Requires that this merge: https://github.com/edx/load-tests/pull/88


fredsmith:
:+1: :shipit:

PR #2631, opened 2015-12-23
SEC-79: Install nginx from their ppa

maxrothman:
This change allows us to install nginx from nginx's ppa as per their
instructions at http://nginx.org/en/linux_packages.html

@edx/devops

### DIFF
--- a/playbooks/roles/nginx/tasks/main.yml
+++ b/playbooks/roles/nginx/tasks/main.yml
@@ -52,7 +52,7 @@
 
 - name: Add nginx repository
   apt_repository:
-    repo: deb http://nginx.org/packages/ubuntu/ {{ ansible_distribution_release }} nginx
+    repo: "{{ NGINX_APT_REPO }}"
     state: present
   notify: restart nginx
   tags:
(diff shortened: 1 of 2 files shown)


playbooks/roles/nginx/tasks/main.yml
```diff
@@ -33,8 +33,37 @@
...
+    repo: deb http://nginx.org/packages/ubuntu/ {{ ansible_distribution_release }} nginx
+    state: present
+  notify: restart nginx
+  tags:
+    - install
+    - install:system-requirements
+
+- name: Install the nginx package
+  apt:
+    name: nginx
```
feanil, line 78: This will auto-upgrade NGINX, I think we should probably pin instead.

maxrothman, line 78: Unfortunately, nginx doesn't distribute multiple versions, just a "stable" and a "mainline", as you can see from their [documentation](http://nginx.org/en/linux_packages.html) and their [repo](http://nginx.org/packages/ubuntu/).

e0d, line 78: I think I'd like to repo defined in `defaults/main.yml` and uppercase.  I see a couple cases where we would like to override.
- where we host our own package.
- where we need a customized package like that provided by AppNeta.

fredsmith:
:+1:

jibsheet:
:+1:

PR #2632, opened 2015-12-23
Create config file for Python New Relic Agent.

ormsbee:
@feanil, @maxrothman: Creating a new config file for the python New Relic agent API so that we can capture per-course and per-org browser metrics in New Relic Insights.

Haven't tested this out yet (I'm really not familiar with working with this repo, need to look up how to do that). Please let me know if the basic premise here (what's going where) seems reasonable.

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -6,7 +6,7 @@
   tags:
     - install
     - install:configuration
-    
+
 - name: create edxapp configuration dir
   file: >
     path="{{ EDXAPP_CFG_DIR }}" state=directory
@@ -14,7 +14,16 @@
   tags:
     - install
     - install:configuration
-    
+
+- name: setup the newrelic python agent config
+  template: >
+    src=newrelic.ini.j2 dest={{ edxapp_app_dir }}/newrelic.ini
+    owner={{ edxapp_user }}  group={{ common_web_user }}
+    mode=0644
+  tags:
+    - install
+    - install:configuration
+
 # Optional auth for git
 - name: create ssh script for git (not authenticated)
   template: >
@@ -24,7 +33,7 @@
   tags:
     - install
     - install:base
-  
+
 - name: create ssh script for git (authenticated)
   template: >
     src=git_ssh_auth.sh.j2 dest={{ edxapp_git_ssh }}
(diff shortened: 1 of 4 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -14,7 +14,16 @@
   tags:
     - install
     - install:configuration
-    
+
+- name: setup the newrelic python agent config
+  template: >
```
maxrothman, line 19: Please use new YAML-style tasks (i.e. make each `key=value` pair into an actual YAML `key: value` pair).

playbooks/roles/newrelic/templates/etc/newrelic/nrsysmond.cfg.j2
```diff
@@ -231,12 +231,3 @@ labels={{ labels }}
 #          path if, for example, you have multiple CGroup mounts.
 # Default: Distribution dependent, but usually /sys/fs/cgroup or /cgroup.
 #cgroup_root=
-
```
maxrothman, line 234: Is this essentially a revert of a previous change? If so, can it be done as an actual revert?

maxrothman:
Generally this looks good. Have you reverted all the other ways you tried to get this to work?

:+1: after this has been tested. Let me know if you need help figuring out how to do that.

playbooks/roles/edxapp/tasks/deploy.yml
ormsbee, line 19: So you mean like:

``` YAML
template:
  src: newrelic.ini.j2
  dest: {{ edxapp_app_dir }}/newrelic.ini
  owner: {{ edxapp_user }}
  group: {{ common_web_user }}
  mode: 0644
```

?

maxrothman, line 19: Exactly. You should also quote all of those lines so a) `{{` isn't interpreted as the start of a dictionary and b) so `0644` isn't interpreted as an octal number. We generally just quote everything except numbers and booleans to be safe.

ormsbee, line 19: Got it, thank you.

playbooks/roles/newrelic/templates/etc/newrelic/nrsysmond.cfg.j2
ormsbee, line 234: Will do.

ormsbee:
I tested it out by looking at New Relic's `Settings > Environment > Agent Initialization` for my sandbox, and the config seems to be working. I'll make the changes, redeploy to test, and then merge. Thank you.

ormsbee:
Closing in favor of #2635

PR #2634, opened 2015-12-28
Feanil/locust system pkgs

feanil:
@edx/devops


maxrothman:
We have a fork of locust? :cry: 
:+1: after squash

PR #2636, opened 2015-12-28
Include nodeenv in the PATH used when installing node dependencies

rlucioni:
@feanil, please review.


playbooks/roles/programs/tasks/main.yml
```diff
@@ -64,7 +65,7 @@
 - name: compile sass
   shell: >
     chdir={{ programs_code_dir }}
-    . {{ programs_nodeenv_bin }}/activate && npm rebuild node-sass && {{ programs_node_bin }}/gulp css
```
rlucioni, line 67: This was formerly required because node-sass was being built with a version of node different from that being used to execute it. Setting the environment when installing node dependencies should fix that issue. I'll test this in CI before merging.

feanil:
:+1:

PR #2637, opened 2015-12-28
Install github-sqs-plugin from fork on Jenkins TE-1052

jzoldak:
@benpatterson


benpatterson:
Not a must (since we're not actively using this play right now), but did we want to store an hpi on S3?

benpatterson:
Discussed with Jay...it's non-trivial. This change :+1:

PR #2638, opened 2015-12-29
adding ec2 terraform module

arbabnazar:
@feanil @fredsmith kindly review. Thanks


feanil:
This feels like it's too small a thing to make our own module for.  Why not just use the builtin EC2 module in our secure config?

arbabnazar:
@feanil sorry I didn't get, we have builtin ec2 module in secure config or move this one there? Thanks

feanil:
Since all this module is doing is creating an ec2 instance,  it doesn't make sense to have it.  If we need to build out ec2 instances, we should just use the builtin ec2_instance module directly instead of having this abstraction around it.

PR #2640, opened 2015-12-29
Add SVGs to the list of things gzip can work on.

tobz:
Since SVGs are text-based, they have the potential to actually compress, unlike binary image formats such as PNG or JPEG.  Basic testing shows that something like the Apple App Store SVG badge compresses roughly 66% under gzip.


e0d:
Looks like a good change, :+1:

tobz:
Success!

Production:
screen shot 2015-12-29 at 9 34 21 pm

Sandbox:
screen shot 2015-12-29 at 9 35 03 pm

A difference of 12.4KB to 4.9KB.

tobz:
Just some numbers:
- 3337 assets that are SVG
- average size of 76.7KB
- minimum size of 209 bytes
- maximum size of 5MB (!)

feanil:
:+1:

PR #2641, opened 2015-12-29
E0d/fix forum

e0d:
Fixing the forum role and re-enabling Docker file for CI.


e0d:
@cpennington

docker/build/forums/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM edxops/precise-common:latest
+MAINTAINER edxops
+
+RUN apt-get update
```
cpennington: No longer needed.

cpennington:
:+1:, after my comment is addressed.

PR #2642, opened 2015-12-29
Add script & playbook for reporting on package upgrades

maxrothman:
@edx/devops


util/jenkins/apt_check_upgrades.py
```diff
@@ -0,0 +1,64 @@
...
+  return args
+
+
+def check_upgrade(pkg):
+  if pkg.installed != pkg.candidate:
+    return {
+      'current_version': pkg.installed.version,
+      'new_version': pkg.candidate.version,
+      'summary': pkg.candidate.summary,
+      'current_md5': pkg.installed.md5,
```
feanil, line 31: What will happen if the current package doesn't have one of these properties?

util/jenkins/apt_check_upgrades.py
```diff
@@ -0,0 +1,64 @@
...
+      'homepage': pkg.candidate.homepage,
+    }
+
+
+def main(args):
+  need_upgrade = {}
+  cache = apt.Cache()
+  
+  if args.all:
+    for pkg in cache:
```
feanil, line 42: You're only checking the cache, is the cache list going to be just those packages that have upgrades?  A comment to that effect would be helpful.

util/jenkins/check_package_upgrades.yml
```diff
@@ -0,0 +1,116 @@
...
+  gather_facts: false
+  vars:
+    ami: !!null
+    profile: !!null
+    region: us-east-1
+    key_name: deployment-201407
+    instance_type: t2.large
+
+    #Admin security group and subnet
+    security_group_id: sg-06be2e63
```
feanil: Let's not check these into the public repo...

feanil:
Why put the playbook under util/jenkins?  It seems like we should put it somewhere under `playbooks/`  For one thing, there is ansible.cfg settings that we should try to have be the same for all playbooks we run.

util/jenkins/apt_check_upgrades.py
maxrothman, line 31: Then they're empty. I've seen that in testing. It's safe because packages are all wrapped in the same class that implements all of those methods.

util/jenkins/apt_check_upgrades.py
maxrothman, line 42: I'm not sure I understand. `cache` is the apt cache. `check_upgrade()` checks if any upgrades are available for the package.

util/jenkins/check_package_upgrades.yml
maxrothman: Whoops, you're right. Do you think I need to go and purge this commit?

maxrothman:
Since this playbook is very single-use and depends on the python script, I thought they should live in the same place, but I'm open to suggestions.

util/jenkins/apt_check_upgrades.py
feanil, line 42: I see so this is just iterating through all installed packages?

util/jenkins/check_package_upgrades.yml
feanil: Yes please.

util/jenkins/apt_check_upgrades.py
maxrothman, line 42: All packages in the cache. That includes packages that apt knows about but aren't installed.

util/jenkins/check_package_upgrades.yml
maxrothman: Request sent.

util/jenkins/check_package_upgrades.yml
```diff
@@ -0,0 +1,125 @@
...
+#
+# Relevant optional arguments:
+#  -e 'script_path=PATH': path to the apt_check_upgrades.py script
+#  -e 'report_dest_path=PATH': path to which the resulting report will be written.
+#  --private-key=PATH_TO_PRIVATE_KEY_FILE: ssh key to use when connecting to the new host
+#  -e 'key_name': AWS key to use for the new instance. This key must be available locally
+#                 either as an ssh profile or as specified with the above option.
+#  -e 'profile=PROFILE': AWS profile to use for AWS API calls
+#  -e 'region=REGION': AWS region to make the instance in
+#  -e 'security_group_id=sg-xxxxxxxx': security group to attack to the new instance
```
fredsmith: s/ack/ach/

fredsmith:
LGTM :+1:

util/jenkins/apt_check_upgrades.py
```diff
@@ -0,0 +1,65 @@
...
+      if pkg.is_installed:
+        result = check_upgrade(pkg)
+        if result:
+          need_upgrade[pkg.name] = result
+  else:
+    for pkg_name in args.packages:
+      if pkg_name not in cache:
+        raise Exception('no package named "{}" exists in the cache!'.format(pkg_name))
```
e0d, line 50: why not return data that reflects this?

maxrothman, line 50: My thought was that in this case, the most likely cause is that the user mistyped the package name, therefore it was a good place for an error.

fredsmith:
:+1:

feanil:
@maxrothman Per my earlier comment, I think the playbook should be moved to the playbooks directory.  We have other on-off playbook around but they are usually under playbook somewhere and I'd like for that to remain true.  It makes sense to have the python script under `util/jenkins` but putting playbooks under there is surprising since we generally want to make use of the same callback plugins and the same ansible.cfg even if we don't explicitly need them right now.

:+1:  once that's done.

PR #2643, opened 2015-12-29
E0d/revert push to master

e0d:
@edx/devops accident push to master, thought this was disabled.


maxrothman:
:+1: Protected branches only protects against force-pushing.

PR #2644, opened 2015-12-29
E0d/fix nginx

e0d:
@cpennington fix the nginx role.  Contains the docker/container role commits until that gets thumbs.


cpennington:
:+1:

PR #2646, opened 2015-12-30
E0d/fix edxapp

e0d:
@cpennington working edxapp builds.


e0d:
Fixes the timeout but exhausts travis memory :facepunch:

e0d:
@cpennington until I understand what our options are for making more memory available to Travis or constraining memory usage during CI, we will likely need to disable CI for edxapp.  I don't want to do so using by renaming the file, since this container will build outside of Travis.  Thoughts on the best way to achieve that?  Perhaps a CI `.excludes` file?

cpennington:
By my reading of the logs, it's still timing out. Why do you think it's memory exhaustion?

Also, as far as excluding it from CI: yeah, adding a .ci-excludes would be fine.

e0d:
I was misinterpreting the sig 137 I think, I agree that it is still probably a timeout issue.  20 minutes is going to be the max we can wait without output

cpennington:
We should be able to make it log incrementally, I would think.

e0d:
Yes, agree, will track down and see what can be done.

Ed.

e0d:
@cpennington see that last commit.

cpennington:
:+1:

PR #2649, opened 2015-12-30
Dogwood upgrade automation

nedbat:


nedbat:
@e0d @fredsmith @feanil @singingwolfboy This now works to update Cypress fullstack and devstack properly.

nedbat:
No one has any comments/thumbs?

singingwolfboy:
Seems reasonable to me. :+1:

nedbat:
@e0d @fredsmith any comments?

util/vagrant/migrate.sh
```diff
@@ -1,19 +1,22 @@
 #!/usr/bin/env bash
```
feanil, line 2: Add a `set -e` up here so we stop on errors that we can't think of right now.

util/vagrant/migrate.sh
```diff
@@ -45,6 +50,53 @@ while getopts "hc:t:y" opt; do
...
+  if [ $exitcode != 0 ]; then
+    exit $exitcode;
+  fi
+}
+
+# Helper to ask to proceed.
+confirm_proceed () {
+  echo "Do you wish to proceed?"
+  read input
+  if [ "$input" != "yes" -a "$input" != "y" ]; then
```
feanil: Why single brackets here and double brackets everywhere else?  Does the double-bracket syntax not support this kind of test?

feanil: I see that this was pulled up from elsewhere, let's go to double-brackets if that will work for this so that it's consintent.  @fredsmith your bash is better than mine, do you think that will cause any issues?

util/vagrant/migrate.sh
```diff
@@ -113,30 +158,96 @@ if [ -f /edx/app/edx_ansible/server-vars.yml ]; then
...
+    vagrant-$CONFIGURATION-delta.yml
+  bail_if_fail
+  cd ../../..
+
+  # Remake our own venv because of the Python 2.7.10 upgrade.
+  rm -rf venv
+  make_config_venv
+
+  # Need to get rid of South from edx-platform, or things won't work.
+  sudo -u edxapp /edx/app/edxapp/venvs/edxapp/bin/pip uninstall -y South
```
feanil, line 197: If we re-made the venvs, does this still need to happen?

feanil, line 197: NM, I see why.  Because we rebuild the venvs as cypress venvs.  Why not re-build them as `dogwood-first-18` ?

util/vagrant/migrate.sh
fredsmith: [[ is less portable than [, but for this comparison and where this script is going to be used, I think it will be fine.

util/vagrant/migrate.sh
nedbat, line 197: We need to build them as last-14 so that we can do the database migration.

PR #2654, opened 2015-12-31
DEVOPS-3420 MTU not being set to 1500

arbabnazar:
@edx/devops,  forums suggest that adding `default interface-mtu 1500;` will solve the problem. Kindly review. Thanks


e0d:
@arbabnazar I'm going to close this in favor of the merged alternative.

PR #2660, opened 2016-01-04
modified the user role

arbabnazar:
@edx/devops kindly review. Main change is to above the `curl` command and creating the `.ssh` directory manually. Ansible `authorized_key` moudle can fetch the `ssh key(s)` from github. Thanks


e0d:
@fredsmith the key change here is to use authorized_key module rather than a template.  It looks good to me, any reason that you remember, why we didn't do this in the first place?

playbooks/roles/user/tasks/main.yml
```diff
@@ -69,90 +71,69 @@
...
-# get_url was failing due to certificate verification errors
-
-- name: get github key[s]
-  shell: >
-    curl https://github.com/{{ item.name }}.keys
-  sudo_user: "{{ item.name }}"
+- name: get github key(s) and update the authorized_keys file
+  authorized_key:
+    user: "{{ item.name }}"
+    key: "https://github.com/{{ item.name }}.keys"
```
fredsmith, line 118: does this work for users that have multiple keys defined in github?  eg. https://github.com/e0d.keys

fredsmith, line 118: tested, it does.

fredsmith:
:+1:   - e0d I can't think of a good reason why we were using the template.

PR #2663, opened 2016-01-04
Fixing MTU setting

e0d:
@edx/devops testing in sandbox indicates this works.

### DIFF
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -34,12 +34,12 @@
   when: ansible_distribution in common_debian_variants
 
 - lineinfile: >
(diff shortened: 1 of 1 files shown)


arbabnazar:
@e0d  this looks nice :+1: 
Can you please test [this one](http://askubuntu.com/questions/350992/how-do-i-override-an-mtu-provided-by-dhcp) 

```
Remove the interface-mtu option from /etc/dhcp/dhclient.conf and it will set 1500 as standard.
```

e0d:
@arbabnazar that did not seem to work, though it did in the past.  I ended up using the AWS specific recommended option from their docs.  A key to testing this is that the sanbox had to be an instance type with an MTU > 1500 by default, which requires choosing a non-default instance type.

PR #2666, opened 2016-01-05
Installing forums bundle with --deployment flag

clintonb:
This will prevent Gemfile.lock from being modified. See http://bundler.io/deploying.html for additional information.

The `BUNDLED WITH` section of Gemfile.lock is there on purpose. See https://github.com/bundler/bundler/issues/3697 for more info.

OPS-1243

PR #2667, opened 2016-01-05
Not needed with --deployment passed to bundle

e0d:
@clintonb  This was added to work-around a perms issues experienced when bundle was updating the Gemfile.lock.  While it fixed the immediate problem, it uncovered the more fundamental issue your PR fixes.

@fredsmith FYI


clintonb:
:+1: @e0d let me know if you want to merge both changes now or hold off until tomorrow. I haven't actually tested #2666 on a sandbox.

PR #2670, opened 2016-01-05
VMware is not supported

singingwolfboy:
Nobody tests on it, and it's probably broken. @edx/devops, can we merge this?


fredsmith:
@e0d are we building vmware images on our virtual machine build server, or just virtualbox?

singingwolfboy:
I confirmed with @e0d that we're not building vmware images, and there have been no other objections, so I'm merging this pull request.

PR #2674, opened 2016-01-06
Tag jenkins_worker AMIs for deletion

benpatterson:
This will give us the ability to run packer builds (which will
actually register an AMI) for the strict purpose of testing
the packer buildout itself. Currently, we only build an AMI
as-needed, and when we want to test the buildout, we must manually
delete the registered AMI if the build is successful. The ability
to tag according to testing or real-use will also allow us to
automatically deregister AMIs that were only being used for test
purposes, rather than actual Jenkins builds.


benpatterson:
@jzoldak

jzoldak:
oic, you set up a new parameter on the build-packer-ami job. cool. :+1:

PR #2679, opened 2016-01-06
Fix value of ecomworker JWT_ISSUER setting

rlucioni:
This fixes a bug accidentally introduced by https://github.com/edx/configuration/commit/b5d8bfa14034221badbea5038569e922831d8f7e. Formerly, the ecommerce worker service name was recycled as the value of JWT_ISSUER. When the name of the service was changed to "ecomworker," the value used for that setting changed to something which the ecommerce service was not configured to expect. ECOM-3341.

@fredsmith please review. @jimabramson FYI.


fredsmith:
:+1:   Thanks for catching this.

PR #2681, opened 2016-01-06
Moving heartbeat

e0d:
@jibsheet @arbabnazar


e0d:
@edx/devops @arbabnazar 

Currently with redirect to ssl enabled, the heartbeat isn't working properly as it too is being redirected.  The needs to merge before we release from master.

jibsheet:
:+1:

PR #2656, opened 2016-01-01
Arbab/change name coursediscovery

arbabnazar:
@edx/devops kindly review, this change is linked with [this](https://github.com/edx-ops/edx-secure/pull/700).
@cpennington @clintonb kindly have a look, if I have missed something to rename.

Thanks

PR #2683, opened 2016-01-07
Added course_discovery to update script help text

clintonb:
This ensures sandbox users are aware that course_discovery can be updated by /edx/bin/update.


clintonb:
FYI @arbabnazar @maxrothman this will need to be changed again by #2656.

PR #2686, opened 2016-01-09
added the nginx redirected logic

arbabnazar:
@edx/devops kindly review. Thanks

PR #2689, opened 2016-01-12
use the elif instead of second if statement

arbabnazar:

For reference: 
use the same logic that we have implemented here https://github.com/edx/configuration/pull/2686


feanil:
:+1:

PR #2692, opened 2016-01-12
Stop looking for docker in our production boxes

jibsheet:
This quiets a bunch of logging like
error: Docker GET from 'http://localhost/info' failed

@edx/devops


e0d:
:+1:

PR #2697, opened 2016-01-13
Add some docs, fix some bugs

maxrothman:
@edx/devops

### DIFF
--- a/playbooks/roles/alton/meta/main.yml
+++ b/playbooks/roles/alton/meta/main.yml
@@ -22,4 +22,6 @@
 dependencies:
   - common
   - supervisor
-  - redis
+  - role: redis
+    redis_debian_pkgs:
+      - "redis-server"
--- a/playbooks/roles/alton/meta/main.yml
+++ b/playbooks/roles/alton/meta/main.yml
@@ -22,6 +22,5 @@
 dependencies:
   - common
   - supervisor
-  - role: redis
-    redis_debian_pkgs:
-      - "redis-server"
+  - redis
+
--- a/playbooks/edx-east/alton.yml
+++ b/playbooks/edx-east/alton.yml
@@ -1,4 +1,5 @@
 # Configure an admin instance with jenkins and asgard.
+# Usage: ansible-playbook alton.yml -i <admin-host>, -e <secure-repo>/admin/edx_admin.yml -e <secure-repo>/admin/admin.yml
 - name: Configure instance(s)
   hosts: all
   sudo: True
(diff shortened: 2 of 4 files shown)


fredsmith:
:+1:

maxrothman:
@edx/devops rereview

playbooks/roles/alton/meta/main.yml
```diff
@@ -22,4 +22,6 @@
 dependencies:
   - common
   - supervisor
-  - redis
+  - role: redis
+    redis_debian_pkgs:
+      - "redis-server"
```
feanil: What's up with this change?

maxrothman: The redis role gets used in other places where a specific version is necessary, but the version specified in the play is too old for Will. This simply un-pins the version.

feanil: Which other roles?

```
feanil@edx ~/src/configuration/playbooks/roles$ find . -name main.yml | grep meta | xargs grep redis
./alton/meta/main.yml:  - redis
./redis/meta/main.yml:# Role includes for role redis
```

I was unable to find them.

maxrothman: @fredsmith, didn't you say it was on sandboxes or something?

fredsmith: i checked, and it seems like nothing else uses the redis role.

feanil: Let's move the pin in the redis role to whatever new version is needed to work with our current version of Will.

maxrothman: Why pin at all? We're using the Ubuntu repo which won't change except for bug/security fixes.

feanil: It sounded like there are versions that might break alton?  So it seems good to not accidentally get that version.

maxrothman: I don't think so. The issues I ran into were with the pinned version being so old that the repo for it had disappeared.

fredsmith:
:shipit:

playbooks/roles/alton/meta/main.yml
maxrothman: @feanil If there aren't any objections, I'll simply un-pin the version in the redis role.

feanil: Sounds good.

feanil:
:+1:

PR #2700, opened 2016-01-13
Sandboxes for ecommerce don't pass along the scheme properly

jibsheet:
The nginx config is only configured for prod (when it is behind an ELB
and has http_x_forwarded headers).  We want to pass along the NGINX
headers too (mainly scheme) whcih is what the LMS/CMS does on sandboxes.

@clintonb Let me know how this goes

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
@@ -78,9 +78,15 @@ server {
   {% include "robots.j2" %}
 
 location @proxy_to_app {
+    {% if NGINX_SET_X_FORWARDED_HEADERS %}
+    proxy_set_header X-Forwarded-Proto $scheme;
+    proxy_set_header X-Forwarded-Port $server_port;
+    proxy_set_header X-Forwarded-For $remote_addr;
+    {% else %}
     proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
     proxy_set_header X-Forwarded-Port $http_x_forwarded_port;
     proxy_set_header X-Forwarded-For $http_x_forwarded_for;
+    {% endif %}
     proxy_set_header Host $http_host;
 
     proxy_redirect off;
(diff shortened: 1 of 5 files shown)


clintonb:
I have confirmed this fix works for course-discovery: https://course-discovery-course-discovery-demo.sandbox.edx.org/api-docs/.

I added two commits to update existing IDAs—ecommerce, course-discovery, programs, credentials—and the IDA template.

FYI @jimabramson @rlucioni @zubair-arbi 

Please review @jibsheet @feanil

jibsheet:
The non-ecom changes look good.  Since I made the ecom changes, I'll ask another of @edx/devops to also thumb.  :+1:

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
```diff
@@ -78,9 +78,15 @@ server {
   {% include "robots.j2" %}
 
 location @proxy_to_app {
+    {% if NGINX_SET_X_FORWARDED_HEADERS %}
```
rlucioni, line 81: I'd consider clarifying the name of this setting to reflect its use. X-Forwarded headers are set regardless of the setting's value; the values used to populate the headers is what's changing.

rlucioni:
:+1: , though clarifying the setting name would be nice.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
clintonb, line 81: If I had time to deal with the potential fallout from breaking other IDAs, I would rename the variable. However, given that this is an injection, I cannot justify taking the time to test the variable change. I've learned that, "it should just work," usually leads to me coding at 4 AM. #neveragain

jibsheet, line 81: I agree it could be better, but is a larger change - looking at the original intent I understand why it was named as it was 28f0e0298c9239e48c94c208b49ccb831b265722  The intent was "set your own because we don't get any" vs "pass through what you get".

feanil:
:+1:

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/programs.j2
jibsheet, line 81: @rlucioni We have a ticket to collapse these templates into one to avoid all the duplication, I added your request to clarify the variable name to that ticket.

rlucioni, line 81: Thanks @jibsheet!

PR #2701, opened 2016-01-13
Jibsheet/idempotent mongo small fixes

jibsheet:
@maxrothman These are the tiny pre-fixes before I get into a real PR, but you should review since it'll help cut down on clutter.


maxrothman:
:+1:

PR #2702, opened 2016-01-13
Added E-Commerce API URL to course-disovery Django settings

clintonb:
DEVOPS-3518 filed to update configuration-secure.


clintonb:
FYI @arbabnazar @jibsheet @cpennington

PR #2708, opened 2016-01-15
Jibsheet/member comparison

jibsheet:
I suggest reading both commits to see the before version we wrote and the after.

### DIFF
--- a/playbooks/library/mongodb_replica_set
+++ b/playbooks/library/mongodb_replica_set
@@ -226,11 +226,12 @@ def update_replset(rs_config):
 
     #Validate it worked
     if changed:
-        changed_rs_config = get_replset(client)
-        if changed_rs_config != rs_config:
-            module.fail_json(msg="Failed to validate that the replica set was changed", config=rs_config)
+        changed_rs_config = get_replset()
+        if not ( all(changed_rs_config[k] == v for k,v  in rs_config.items() if k != 'members') \
+                 or all([m for m in changed_rs_config['members'] if m['_id'] == member['_id']][0][k] == v for member in rs_config['members'] for k,v in member.items())):
+          module.fail_json(msg="Failed to validate that the replica set was changed", changed_rs_config=changed_rs_config, config=rs_config)
 
-    module.exit_json(changed=changed, config=rs_config)
+    module.exit_json(changed=changed, config=rs_config, new_config=changed_rs_config)
 
 
 ######### Client making stuff #########
--- a/playbooks/library/mongodb_replica_set
+++ b/playbooks/library/mongodb_replica_set
@@ -177,6 +177,38 @@ def fix_host_port(rs_config):
                 if 'port' in member:
                     del member['port']
 
+def check_config_subset(old_config, new_config):
+    '''
+    Compares the old config (what we pass in to Mongo) to the new config (returned from Mongo)
+    It is assumed that old_config will be a subset of new_config because Mongo tracks many more
+    details about the replica set and the members in a replica set that we don't track in our
+    secure repo.
+    '''
+
+    for k in old_config:
+        if k == 'members':
+            matches = is_members_subset(old_config['members'],new_config['members'])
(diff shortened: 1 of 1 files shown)


playbooks/library/mongodb_replica_set
```diff
@@ -177,6 +177,38 @@ def fix_host_port(rs_config):
                 if 'port' in member:
                     del member['port']
 
+def compare_rs_config(old_config, new_config):
```
feanil: I would maybe reword this to be something like `is_subset` instead of compare since it's not actually checking for equality.

playbooks/library/mongodb_replica_set
```diff
@@ -177,6 +177,38 @@ def fix_host_port(rs_config):
...
+    '''
+    Compares the member list of a replica set configuration as specified (old_members)
+    to what Mongo has returned (new_members).  If it finds anything in old_members that
+    does not match new_members, it will return False.  new_members is allowed to contain
+    extra information that is not reflected in old_members because we do not necesarily
+    track all of mongo's internal data in the config.
+    '''
+    for member in old_members:
+        for k in member:
+            if member[k] != new_members[member['_id']][k]: return False
```
feanil, line 208: nit: Put the `return False` on its own line.

feanil:
Logic makes sense and is a lot easier to read than before.  The naming was misleading since it's not an equality check.

playbooks/library/mongodb_replica_set
```diff
@@ -177,6 +177,38 @@ def fix_host_port(rs_config):
...
+
+    return True
+
```
maxrothman, line 211: This entire function:

```
return any(v != new_members[member['_id'][k] for member in old_members for k,v in member.items())
```

playbooks/library/mongodb_replica_set
```diff
@@ -226,11 +258,11 @@ def update_replset(rs_config):
 
     #Validate it worked
     if changed:
-        changed_rs_config = get_replset(client)
-        if changed_rs_config != rs_config:
-            module.fail_json(msg="Failed to validate that the replica set was changed", config=rs_config)
+        changed_rs_config = get_replset()
+        if not compare_rs_config(rs_config, changed_rs_config):
+            module.fail_json(msg="Failed to validate that the replica set was changed", changed_rs_config=changed_rs_config, config=rs_config)
```
maxrothman: Since the existing results key is `config`, should `changed_rs_config` be `changed_config` instead?

jibsheet: The existing results key is rs_config (The first argument to the function) so I don't understand the comment?

Ah - I see what you're commenting on again.  I didn't pick the internal variable names, and I don't see a ton of benefit in renaming them to strictly match what is returned from fail_json or exit_json a few lines later with the same data.

maxrothman: No I'm suggesting changing this line to `module.fail_json(msg="...", changed_config=changed_rs_config, config=rs_config)` so that the user-facing names are consistent.

playbooks/library/mongodb_replica_set
```diff
@@ -226,11 +258,11 @@ def update_replset(rs_config):
 
     #Validate it worked
     if changed:
-        changed_rs_config = get_replset(client)
-        if changed_rs_config != rs_config:
-            module.fail_json(msg="Failed to validate that the replica set was changed", config=rs_config)
+        changed_rs_config = get_replset()
+        if not check_config_subset(rs_config, changed_rs_config):
+            module.fail_json(msg="Failed to validate that the replica set was changed", changed_rs_config=changed_rs_config, config=rs_config)
```
maxrothman: Porting old comment over:
I feel this line should be changed to `module.fail_json(msg="...", changed_config=changed_rs_config, config=rs_config)` to maintain consistency in user-facing names.

maxrothman:
After comment addressed, :+1:

PR #2712, opened 2016-01-19
Jibsheet/play reordering

jibsheet:
This is my rewrite/reordering of the mongo_3_0 role.

It has been tested standing up a new cluster, reconfiguring an existing cluster, and adding a new member to a replica set where that new member needed the mongo installs.
I've also added more tags than the original role had, although I'm not fully sold on all the choices there.

@edx/devops

### DIFF
--- a/playbooks/edx-east/mongo_3_0.yml
+++ b/playbooks/edx-east/mongo_3_0.yml
@@ -1,3 +1,13 @@
+# Manages a mongo cluster.
+# To set up a new mongo cluster, make sure you've configured MONGO_RS_CONFIG
+# as used by mongo_replica_set in the mongo_3_0 role.
+#
+# If you are initializing a cluster, your command might look like:
+# ansible-playbook mongo_3_0.yml -i 10.1.1.1,10.2.2.2,10.3.3.3 -e@/path/to/edx.yml -e@/path/to/ed.yml
+# If you are adding a member to a cluster, you must be sure that the new machine is not first in your inventory
+# ansible-playbook mongo_3_0.yml -i 10.1.1.1,10.2.2.2,new-machine-ip -e@/path/to/edx.yml -e@/path/to/ed.yml
+# If you just want to deploy an updated replica set config, you can run
+# ansible-playbook mongo_3_0.yml -i any-cluster-ip -e@/path/to/edx.yml -e@/path/to/ed.yml --tags configure_replica_set
 - name: Deploy MongoDB
   hosts: all
   sudo: True
--- a/playbooks/edx-east/mongo_3_0.yml
+++ b/playbooks/edx-east/mongo_3_0.yml
@@ -4,10 +4,12 @@
 #
 # If you are initializing a cluster, your command might look like:
 # ansible-playbook mongo_3_0.yml -i 10.1.1.1,10.2.2.2,10.3.3.3 -e@/path/to/edx.yml -e@/path/to/ed.yml
-# If you are adding a member to a cluster, you must be sure that the new machine is not first in your inventory
-# ansible-playbook mongo_3_0.yml -i 10.1.1.1,10.2.2.2,new-machine-ip -e@/path/to/edx.yml -e@/path/to/ed.yml
 # If you just want to deploy an updated replica set config, you can run
 # ansible-playbook mongo_3_0.yml -i any-cluster-ip -e@/path/to/edx.yml -e@/path/to/ed.yml --tags configure_replica_set
+#
+# ADDING A NEW CLUSTER MEMBER
+# If you are adding a member to a cluster, you must be sure that the new machine is not first in your inventory
+# ansible-playbook mongo_3_0.yml -i 10.1.1.1,10.2.2.2,new-machine-ip -e@/path/to/edx.yml -e@/path/to/ed.yml
 - name: Deploy MongoDB
   hosts: all
   sudo: True
--- a/playbooks/roles/mongo_3_0/tasks/main.yml
+++ b/playbooks/roles/mongo_3_0/tasks/main.yml
@@ -121,11 +121,6 @@
   debug: msg="You do not appear to have a Replica Set configured, deploying one for you"
   when: MONGO_CLUSTERED and initialize_replica_set
 
-- name: stop mongod service
-  service:
-    name: mongod
-    state: stopped
-
 - name: copy mongodb key file
   copy:
     content: "{{ MONGO_CLUSTER_KEY }}"
@@ -148,6 +143,7 @@
     dest: /etc/mongod.conf
     backup: yes
   notify: restart mongo
+  register: update_mongod_conf
 
 - name: install logrotate configuration
   template:
@@ -158,10 +154,11 @@
     - "install:configuration"
     - "logrotate"
 
-- name: start mongo service
+- name: restart mongo service if we changed our configuration
   service:
     name: mongod
-    state: started
+    state: restarted
+  when: update_mongod_conf.changed
 
 - name: wait for mongo server to start
   wait_for:
--- a/playbooks/roles/mongo_3_0/tasks/main.yml
+++ b/playbooks/roles/mongo_3_0/tasks/main.yml
@@ -179,6 +179,7 @@
   tags:
     - "install"
     - "install:configuration"
+    - "update_mongod_conf"
 
 - name: install logrotate configuration
   template:
@@ -256,7 +257,6 @@
     src: mongod.conf.j2
     dest: /etc/mongod.conf
     backup: yes
-  notify: restart mongo
   when: MONGO_CLUSTERED and initialize_replica_set
   tags:
     - "install"
--- a/playbooks/roles/mongo_3_0/tasks/main.yml
+++ b/playbooks/roles/mongo_3_0/tasks/main.yml
@@ -1,14 +1,4 @@
 ---
-# - name: check to see if MongoDB is already installed
-#   stat:
-#     path: /etc/init.d/mongod
-#   register: mongodb_already_installed
-#   when: MONGO_CLUSTERED
-
-# - name: verify MongoDB not installed (clustered upgrades)
-#   fail: msg="MongoDB is currently installed and cannot be safely upgraded in a clustered configuration.  FIXME"
-#   when: mongodb_already_installed.stat.exists and MONGO_CLUSTERED
-
 - name: disable transparent huge pages on startup (http://docs.mongodb.org/manual/tutorial/transparent-huge-pages/)
   copy:
     src: disable-transparent-hugepages.conf
@@ -49,7 +39,6 @@
     repo: "{{ MONGODB_REPO }}"
     state: present
 
-#Will this break if it actually upgrades?
 - name: install mongo server and recommends
   apt:
     pkg: "{{ item }}"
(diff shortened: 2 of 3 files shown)


playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -90,12 +94,65 @@
...
-#### DANGER WILL ROBINSON ####
-- name: stop mongod service
-  service:
-    name: mongod
-    state: stopped
+# This will error when run on a new replica set, so we ignore_errors
+# and connect anonymously next.
+- name: determine if there is a replica set already
+  mongodb_rs_status:
+    host: "{{ ansible_default_ipv4['address'] }}"
```
maxrothman, line 105: Why not `localhost`?

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -90,12 +94,65 @@
...
+    username: "{{ MONGO_ADMIN_USER }}"
+    password: "{{ MONGO_ADMIN_PASSWORD }}"
+  run_once: true
```
maxrothman, line 108: What if the arbitrarily-selected host is the one that you're adding?

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -90,12 +94,65 @@
...
+  register: authed_replica_set_already_configured
+  when: MONGO_CLUSTERED
+  ignore_errors: true
+  tags:
+    - "install"
+    - "install:configuration"
+
+- name: Try checking the replica set with no user/pass in case this is a new box
+  mongodb_rs_status:
+    host: "{{ ansible_default_ipv4['address'] }}"
```
maxrothman, line 118: as above

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -90,12 +94,65 @@
...
+    - "install:configuration"
+
+- name: warn about unconfigured replica sets
+  debug: msg="You do not appear to have a Replica Set configured, deploying one for you"
+  when: MONGO_CLUSTERED and initialize_replica_set
+  tags:
+    - "install"
+    - "install:configuration"
 
 - name: copy mongodb key file
```
maxrothman, line 158: What are the implications of this happening when `initialize_replica_set == false` and `skip_replica_set == false`?

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -105,13 +162,30 @@
     owner: mongodb
     group: mongodb
   when: MONGO_CLUSTERED
+  tags:
+    - "install"
+    - "install:configuration"
 
+- name: Set an initialize replica set fact for mongod.conf to skip defining a replica set
```
maxrothman: This sets a skip replica set fact

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -105,13 +162,30 @@
...
+  set_fact:
+    skip_replica_set: true
+  when: MONGO_CLUSTERED and initialize_replica_set
```
maxrothman: Doesn't this mean that `skip_replica_set` will always be `true` when `initialize_replica_set` is `true` here? Why not set them both at the same time?

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -122,16 +196,91 @@
...
-- name: start mongo service
+- name: restart mongo service if we changed our configuration
   service:
     name: mongod
-    state: started
+    state: restarted
+  when: update_mongod_conf.changed
+  tags:
+    - "install"
+    - "install:configuration"
```
maxrothman, line 202: I think we might want a new tag for replset initialization

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -122,16 +196,91 @@
...
+# the localhost exemption to create a user who will be
+# able to initialize the replica set.
+- name: create super user
+  mongodb_user:
+    name: "{{ MONGO_ADMIN_USER }}"
+    password: "{{ MONGO_ADMIN_PASSWORD }}"
+    database: admin
+    roles: root
+  when: initialize_replica_set
+  run_once: true
```
maxrothman, line 227: Not sure this is desirable. I don't think `run_once` is guaranteed to choose the same host every time. Why not just do all of them?

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -122,16 +196,91 @@
...
+  tags:
+    - "manage"
+    - "manage:db"
+
+- name: create super user
+  mongodb_user:
```
maxrothman, line 233: On initialize, you'll get `mongodb_user` running twice, which is kind of odd.

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -122,16 +196,91 @@
...
+    - "manage"
+    - "manage:db"
+
+# Now that the localhost exemption has been used to create the superuser, we need
+# to add replica set to our configuration.  This will never happen if we detected
+# a replica set in the 'determine if there is a replica set already' task.
+- name: Unset our initialize replica set fact so that mongod.conf gets a replica set
+  set_fact:
+    skip_replica_set: false
+  when: MONGO_CLUSTERED and skip_replica_set and initialize_replica_set
```
maxrothman: The last two are paired, why check both?

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -122,16 +196,91 @@
...
+  tags:
+    - "install"
+    - "install:configuration"
+
+- name: re-copy configuration template with replica set enabled
+  template:
+    src: mongod.conf.j2
+    dest: /etc/mongod.conf
+    backup: yes
+  notify: restart mongo
```
maxrothman: Don't we restart mongo anyways below?

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 105: The play previously used the non-local address for connections, so I maintained that.
If there is a compelling reason to always use localhost, I can make sweeping changes, but trying to make sure we could connect to it on the external IP seemed sound to me.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 108: My plan was to update our "adding a new member" directions to state not to put your new host first in the inventory.  Ansible takes inventory[0] when you say run_once (in the docs and in practice) so testing with the new host last in the inventory was successful.

Additionally, if the new host tried to initiate a new replica set on itself, it would fail to join the existing replica set for several reasons.  We need to let it not try to create users or the replica set, and having it not search for the replica set seemed simplest.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 118: Also as above.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 158: skip_replica_set is only used in the template, and is gated by initialize_replica_set being true

playbooks/roles/mongo_3_0/tasks/main.yml
maxrothman, line 108: Right, but why have it run-once at all? Is there a disadvantage to running this on all hosts?

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet: Pushed a fixed name:

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet: Good point, this dates from a much more complicated logic flow.  I also had tried to put it closer to the template where it is used, but that doesn't seem that important.  I've pushed a collapsing commit.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 202: Initialization should be safe to run on installed machines.  I had added a tag for "just update the replset config" later in the role.  However, when I started adding initialize_replica_set tags in, I felt like I tagged 75% of the role and took them back out.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 227: run_once is documented by ansible to be inventory[0] and that bears out in testing.
You cannot run it across all hosts because then you can't initialized the replica set (host B will have data and fail to come into the replset).

I'll add a clarifying note.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 233: You do, it won't do anything, but I have pushed a commit to skip it.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet: Overly paranoid checking, I'll back it back to a single var check.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet: We do, and I don't think we'll run this task directly (I can foresee having a tag on the other mongod.conf templating task so you could roll out a template update and have the notify fire) but this was copy/pasta from that task so carried along the notify.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 108: Because of the confusing case where one machine will not have a replica set and will try to initialize and will later fail to come into the replica set.

playbooks/roles/mongo_3_0/tasks/main.yml
maxrothman, line 158: Right, but it'll still copy the key regardless of `initialize_replica_set`

playbooks/roles/mongo_3_0/tasks/main.yml
maxrothman, line 108: Ok, then we should have copious comments somewhere telling people not to put new hosts at the top

jibsheet, line 108: I believe our docs currently live in the wiki - I was planning to update them appropriately.

playbooks/roles/mongo_3_0/tasks/main.yml
jibsheet, line 158: Oh, I didn't realize you were talking about the key.  Copying that is a no-op in most cases, and you should be permitted to deploy a new cluster key without needing to intitialize it.

playbooks/roles/mongo_3_0/tasks/main.yml
maxrothman, line 108: I meant lots of in-line comments, probably around the variable that sets the replica set config

playbooks/roles/mongo_3_0/tasks/main.yml
maxrothman, line 158: Yeah, but wouldn't that be surprising since it'd pick it up the next time you restarted for any reason?

jibsheet, line 158: Changing the mongo cluster key now notifiers, and has a separate tag in case you want to roll that out.

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -60,7 +58,7 @@
   with_items: mongodb_debian_pkgs
   tags:
     - install
-    - install:system-requirements
+    - install:app-requirements
     - mongo_packages
```
maxrothman, line 62: Why the separate tag?

jibsheet, line 62: We have app vs system requirements in our list of tags and this these are mongo itself, thus are app packages, not system packages

playbooks/roles/mongo_3_0/tasks/main.yml
```diff
@@ -122,16 +192,95 @@
...
+  run_once: true
+  when: not initialize_replica_set
+  tags:
+    - "manage"
+    - "manage:db"
+
+# Now that the localhost exemption has been used to create the superuser, we need
+# to add replica set to our configuration.  This will never happen if we detected
+# a replica set in the 'determine if there is a replica set already' task.
+- name: Unset our initialize replica set fact so that mongod.conf gets a replica set
```
maxrothman: It's a skip replica set fact

playbooks/edx-east/mongo_3_0.yml
```diff
@@ -1,3 +1,13 @@
+# Manages a mongo cluster.
+# To set up a new mongo cluster, make sure you've configured MONGO_RS_CONFIG
+# as used by mongo_replica_set in the mongo_3_0 role.
+#
+# If you are initializing a cluster, your command might look like:
+# ansible-playbook mongo_3_0.yml -i 10.1.1.1,10.2.2.2,10.3.3.3 -e@/path/to/edx.yml -e@/path/to/ed.yml
+# If you are adding a member to a cluster, you must be sure that the new machine is not first in your inventory
```
maxrothman: moar caps. This warning should jump out at me.

maxrothman:
:+1: after 2 comments addressed

PR #2718, opened 2016-01-22
MA-1955: Add OAuth expiration config.

jcdyer:
https://openedx.atlassian.net/browse/MA-1955

Cf: https://github.com/edx/edx-platform/pull/11237

Reviewers: @jibsheet @nasthagiri


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -442,6 +442,8 @@ EDXAPP_CELERY_WORKERS:
 
 EDXAPP_DEFAULT_CACHE_VERSION: "1"
 EDXAPP_OAUTH_ENFORCE_SECURE: True
+EDXAPP_OAUTH_EXPIRE_DELTA_DAYS: 365
+EDXAPP_OAUTH_EXPIRE_DELTA_PUBLIC_DAYS: 30
```
nasthagiri: Looking at this outside of context, I had to double take when I read "Public Days".

It may be worth renaming this to be more clear: `EDXAPP_OAUTH_EXPIRE_PUBLIC_CLIENT_DELTA_DAYS`.

nasthagiri:
:+1: with suggestion to rename one of the settings.

playbooks/roles/edxapp/defaults/main.yml
jibsheet: Oh - good.  My request was going to be comments explaining it or better names since I'm truly out of context and know I was confused by it.

jcdyer: I have mixed feelings about this.  On the one hand, it does read better, as you say, but on the other hand, the current name matches the original setting name with a prefix (EDXAPP_) and suffix (_DAYS) added to it, which means that if you grep for the original setting name, you will find this one as well.

The change you suggest breaks this use case.  You would have to grep the original setting name, find the overridden setting name, and then search for that.

I'd like to defer to @jibsheet and the devops team, as to whether readability or greppability is more important for their normal use cases.

nasthagiri: @jcdyer My working assumption was that we would also update the name of the corresponding setting in the lms/envs/aws.py files.  So the edX naming convention would all be the same.  Only the underlying OAuth-library name may be different (which is fine since it's crossing application boundaries).

jibsheet: ==nimisha - pick a better internal name for this var and wrap it around their names where it crosses the boundary.

jcdyer: OK.  I'll use `OAUTH_EXPIRE_{CONFIDENTIAL,PUBLIC}_CLIENT_DAYS`.

jcdyer:
@jibsheet Variable renamed.  Can I get a thumb?

feanil:
:+1:

jibsheet:
:+1:

PR #2719, opened 2016-01-25
WIP, not for review: E0d/targeted tests

e0d:


e0d:
@cpennington WIP to target tests based on changes, still needs integration into the Makefile, but you can see the idea in the travis hackery.

Have a first look at the python-deps script if you have a moment.

e0d:
@benpatterson @cpennington Any value in pushing on this approach for solving the Travis utilization issue?

cpennington:
@e0d I think this could help a lot, if we could go down to a single targeted shard.

e0d:
@bilalahmad99

util/ansible-deps.py
```diff
@@ -0,0 +1,180 @@
...
+
+    def __init__(self):
+        self.ansible_config = self._load_ansible_config()
+
+    def get_roles_by_playbook(self, playbook_path):
+        """
+        Starts from a playbook and finds all roles included in all plays as well as roles
+        included via meta/main.yml includes of directly included roles.  Returns the set of roles.
+
+        TODO: make analysis recursive
```
benpatterson, line 38: I think this must be sorted out before making this real. (You know this too, just calling it out on my end.)

util/ansible-deps.py
```diff
@@ -0,0 +1,180 @@
...
+        for role in roles:
+            resolved_name = self._get_role_name(role)
+            _roles.add(resolved_name)
+            _roles |= set(self.get_roles_by_role(resolved_name))
+
+        return _roles
+
+    def get_roles_by_role(self, role_name):
+        """
+        For a given role_name, finds it's included dependent roles in meta/main.yml. Roles are resolved
```
benpatterson, line 65: nit: its

util/ansible-deps.py
```diff
@@ -0,0 +1,180 @@
...
+        using ansible.cfg and a path relative roles directory, if available.
+        :param role_name:
+        :return:
+        """
+        role_dir = self._locate_role_dir(role_name)
+        meta_file = role_dir + "/meta/main.yml"
```
benpatterson, line 71: Do we have any other non-main.yml files under meta? (I can't think of any.)

benpatterson:
Are specific playbooks going to be involved (e.g., under `edx-east`)? How about the vars that are set in `meta.yml` when calling dependent roles?

I like the idea overall, but think you shouldn't spend too many cycles on this until we start running low on travis cluster overhead/available nodes or the builds are taking too long.

cpennington:
I think it's worth keeping at this just to cut down the turnaround time on PRs.

feanil:
We landed some variant of this.

PR #2728, opened 2016-01-26
Move to using a more recent build for the default devstack.

feanil:
The dogwood devstack was not compatible with some changes made after dogwood was cut.

@edx/devops @nedbat


nedbat:
:+1:

jibsheet:
Testing this - I just got a 403 on the box, confirmed with curl

Ah - I see - the S3 URL is ok, but files.edx.org is 403'ing it.

jibsheet:
I fiddled the filename and am testing an up, assuming this fixes the forums issue, I'll merge

PR #2732, opened 2016-01-27
CI needs to run

e0d:
@edx/devops @zubair-arbi I'm proposing  that we disable credentials in CI until we know how to fix the memory issues with collecting static resources.  CI for other roles is too important to let this remain broken and credentials has Docker coverage.

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -319,7 +319,7 @@ EOF
 fi
 
 declare -A deploy
-roles="edxapp forum ecommerce programs credentials course_discovery notifier xqueue xserver certs demo testcourses"
+roles="edxapp forum ecommerce programs course_discovery notifier xqueue xserver certs demo testcourses"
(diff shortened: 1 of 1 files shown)


e0d:
@cpennington @clintonb FYI.

clintonb:
:+1: @zubair-arbi let me know if you need help debugging.

feanil:
:+1:  -> changing this to a :-1:   we should close the PR.

feanil:
Actually, thinking about it some more, we don't need this PR. The original line make it so jenkins _could_ run the credentials playbook, but we can just update the CI jenkins job to not run against credentials.  I've done so in it's configuration as of right now.

jibsheet:
@feanil while you turned off the checkbox, it continues to run credentials in CI - it's still failing on the same error.  I suspect we do need this PR or something like it.

feanil:
Closing this.  #2733 accomplishes what we actually want here.

PR #2733, opened 2016-01-27
Don't run credentials in CI

jibsheet:
It runs out of memory on the CI box every time compressing assets and at
least it is tested in Travis as part of that CI job.

@edx/devops 

This may replace https://github.com/edx/configuration/pull/2732

### DIFF
--- a/playbooks/edx-east/edx_continuous_integration.yml
+++ b/playbooks/edx-east/edx_continuous_integration.yml
@@ -18,7 +18,6 @@
       - analytics_api
       - ecommerce
       - programs
-      - credentials
       nginx_default_sites:
       - lms
     - mysql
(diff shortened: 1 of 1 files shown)


jibsheet:
This may not be needed if pruning the gunicorn workers trims memory hard enough.

feanil:
pruning gunicorn workers does not seem to be sufficient.  let's merge this for now :+1:

zubair-arbi:
@feanil Does the CI still breaking due to credentials?

jibsheet:
Closing for now

feanil:
@zubair-arbi looks like it is okay for now, we found an issues with the notifier that was exacerbating the problem.

PR #2736, opened 2016-01-27
Print only the 10 slowest tasks instead of all of them.

feanil:
Print only the top 10 slowest tasks to the console after an ansible run.  All data will still go to datadog but we don't print to the console.

@edx/devops


jibsheet:
:+1:

nedbat:
:+1: Now if we could figure out why it prints in red... (btw, it's not just me, a number of people have commented on it!)

PR #2738, opened 2016-01-28
remove the demo role

arbabnazar:
@edx/devops kindly review this, we are not using the `demo` role, instead of this we are using `testcourses` role to import all the courses even the demo course that `demo` role import. Thanks

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -319,7 +319,7 @@ EOF
 fi
 
 declare -A deploy
-roles="edxapp forum ecommerce programs credentials course_discovery notifier xqueue xserver certs demo testcourses"
+roles="edxapp forum ecommerce programs credentials course_discovery notifier xqueue xserver certs testcourses"
 for role in $roles; do
     deploy[$role]=${!role}
 done


feanil:
@arbabnazar the demo role is used in a lot of other places, should it be deleted all together?

arbabnazar:
@feanil I have just propose to remove it from the ansible provision script and update the jenkins job, because we are using `testcourses` role which is doing this job for us. It's just sit there for nothing.

feanil:
It sounds like these days, `testcourses` can do everything that `demo` used to do, should we just delete the role all together?

arbabnazar:
+1 from my side but @e0d raise one point that how the `demo` user has been created even the task has been skipped, do you have any idea that where else the it has been created or it has already been created on the ami that we are using for our sandbox? Thanks

feanil:
I think if we move user-creation into either a `testuser` role or into the existing `testcourses` role, then we wouldn't need the `demo` role for anything.

arbabnazar:
@feanil I am working on this [PR](https://github.com/edx/configuration/pull/2739) to make all the changes that you have suggested. Thanks

PR #2741, opened 2016-01-29
OPS-1278 Create role to install MySQL locally that is not experimental

arbabnazar:

### DIFF
--- a/playbooks/roles/mysql/defaults/main.yml
+++ b/playbooks/roles/mysql/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
 mysql_debian_pkgs:
-  - mysql-server
+  - "mysql-server-5.6"
   - python-mysqldb
   
--- a/playbooks/roles/mysql/defaults/main.yml
+++ b/playbooks/roles/mysql/defaults/main.yml
@@ -1,5 +1,5 @@
 ---
 mysql_debian_pkgs:
-  - software-properties-common
+  - mysql-server
   - python-mysqldb
-  - "mysql-server=5.6.*"
+  
--- a/playbooks/roles/mysql/defaults/main.yml
+++ b/playbooks/roles/mysql/defaults/main.yml
@@ -1,4 +1,6 @@
 ---
+remove_experimental_mysql: false
+
 mysql_debian_pkgs:
   - "mysql-server-5.6"
   - python-mysqldb
--- a/playbooks/roles/mysql/meta/main.yml
+++ b/playbooks/roles/mysql/meta/main.yml
@@ -1,3 +1,3 @@
 ---
 dependencies:
-  - common
+  - common_vars
--- a/playbooks/roles/mysql/meta/main.yml
+++ b/playbooks/roles/mysql/meta/main.yml
@@ -1,3 +1,3 @@
 ---
 dependencies:
-  - common_vars
+  - common
--- a/playbooks/roles/mysql/tasks/main.yml
+++ b/playbooks/roles/mysql/tasks/main.yml
@@ -1,8 +1,13 @@
 ---
 - name: Look for mysql 5.6
-  shell: dpkg -L mysql-server-5.6
+  shell: "dpkg -s mysql-server"
   ignore_errors: yes
   register: mysql_56_installed
- 
+  changed_when: no
+
+- fail:
+    msg: "You have already installed the experimental MySQL 5.6 repo, Please first remove that"
+  when: "'5.6.14' in  mysql_56_installed.stdout"
+
 - include: mysql.yml
   when: mysql_56_installed.rc == 1
--- a/playbooks/roles/mysql/tasks/main.yml
+++ b/playbooks/roles/mysql/tasks/main.yml
@@ -9,6 +9,7 @@
     msg: "MySQL experimental is already installed, make remove_experimental_mysql: true in defaults/main.yml, if you want to remove it"
   when: "'5.6.14' in  mysql_56_installed.stdout and not remove_experimental_mysql"
 
+# remove this, once the new devstack is out
 - include: remove_mysql_experimental.yml
   when: remove_experimental_mysql
 
--- a/playbooks/roles/mysql/tasks/main.yml
+++ b/playbooks/roles/mysql/tasks/main.yml
@@ -1,36 +1,8 @@
 ---
-- name: Add MySQL community apt key
-  apt_key:
-    id: 8C718D3B5072E1F5
-    keyserver: "{{ COMMON_EDX_PPA_KEY_SERVER }}"
-    state: present
-
-- name: Install MySQL community apt repositories
-  apt_repository:
-    repo: "{{ item }}"
-    update_cache: yes
-  with_items:
-    - "deb http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6"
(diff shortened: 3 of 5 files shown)


playbooks/roles/mysql/tasks/mysql.yml
```diff
@@ -0,0 +1,36 @@
...
+    state: present
+
+- name: Install MySQL community apt repositories
+  apt_repository:
+    repo: "{{ item }}"
+    update_cache: yes
+  with_items:
+    - "deb http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6"
+    - "deb-src http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6"
+
```
e0d, line 46: Will this work without providing input?

playbooks/roles/mysql/meta/main.yml
```diff
@@ -1,3 +1,3 @@
 ---
 dependencies:
-  - common
+  - common_vars
```
e0d: We want to keep this as common ensures the role can stand on its own.  Went back and forth on my point, but landed on using common.

playbooks/roles/mysql/tasks/mysql.yml
arbabnazar, line 46: yes, I have installed their `.deb` package and found out that this deb package only add these ppa inside the source.list file, so I have removed all those steps and make it simple and test on the sandbox. Thanks

playbooks/roles/mysql/tasks/mysql.yml
```diff
@@ -0,0 +1,36 @@
...
+    id: 8C718D3B5072E1F5
+    keyserver: "{{ COMMON_EDX_PPA_KEY_SERVER }}"
+    state: present
+
+- name: Install MySQL community apt repositories
+  apt_repository:
+    repo: "{{ item }}"
+    update_cache: yes
+  with_items:
+    - "deb http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6"
```
e0d: Mark vars, use HTTPS.  Should source install be optional?

playbooks/roles/mysql/defaults/main.yml
```diff
@@ -1,5 +1,5 @@
 ---
 mysql_debian_pkgs:
-  - software-properties-common
+  - mysql-server
```
e0d: Can we pin version now?

playbooks/roles/mysql/tasks/main.yml
```diff
@@ -1,33 +1,8 @@
...
-    name: '' 
-    host: "{{ item }}" 
-    state: absent
-  with_items:
-    - localhost
-    - "{{ ansible_hostname }}"
+- name: Look for mysql 5.6
+  shell: dpkg -L mysql-server-5.6
+  ignore_errors: yes
+  register: mysql_56_installed
```
e0d, line 5: How does this work with other version of mysql.  For example, if you are installing on top of the experimental version?

e0d, line 5: @benpatterson @nedbat we want to respond appropriately when we detect that MySQL 5.6 is already installed.  My initial thought is that we should not automatically upgrade -- slightly complicated by the fact we are changing PPAs.  I wanted to make sure we have something that is tenable for Jenkins workers and Devstack.

e0d:
@benpatterson  @nedbat General review would be appreciated.

playbooks/roles/mysql/tasks/main.yml
benpatterson, line 5: FWIW jenkins-workers are installing mysql for the first time during the run IIRC, so that shouldn't be too much of an issue.

benpatterson, line 5: at this point the only complication would be  `libmysqlclient-dev` which is installed to get ansible running. AFAIK that does not actually install mysql, so that should be fine.

arbabnazar:
Here are the few steps that I have performed to get the `debconf` setting information:
- downloaded the mysql-apt-repo from the mysql official site(named: mysql-apt-config_0.6.0-1_all.deb)
- Find the `debconf` setting information after installing this `deb` package

```
# debconf-show mysql-apt-config
* mysql-apt-config/select-server: mysql-5.6
  mysql-apt-config/unsupported-platform: abort
  mysql-apt-config/repo-codename: precise
* mysql-apt-config/select-product: Apply
  mysql-apt-config/select-tools:
  mysql-apt-config/repo-distro: ubuntu
```
- Even to get more information about the `debconf` options of the package:

```
# debconf-get-selections|grep mysql-apt-config
mysql-apt-config    mysql-apt-config/select-server  select  mysql-5.6
mysql-apt-config    mysql-apt-config/unsupported-platform   select  abort
mysql-apt-config    mysql-apt-config/repo-codename  select  precise
mysql-apt-config    mysql-apt-config/select-product select  Apply
mysql-apt-config    mysql-apt-config/select-tools   select
mysql-apt-config    mysql-apt-config/repo-distro    select  ubuntu
```
- After the installation, I have checked the `/etc/apt/sources.list.d` directory and found one file `mysql.list` with following contents:

```
deb http://repo.mysql.com/apt/ubuntu/ precise mysql-apt-config
deb http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6
deb-src http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6
```

Thought that instead of performing all those steps and get the repo, why not directly use this repo 
`deb http://repo.mysql.com/apt/ubuntu/ precise mysql-5.6`, I just picked this line and directly used it and it worked for us. 

Thanks,

e0d:
:+1: 
@edx/devops  second review please
@arbabnazar please add the details above into the role as a comment.

playbooks/roles/mysql/tasks/main.yml
```diff
@@ -1,33 +1,13 @@
...
-- name: install mysql 56 and dependencies
-  apt: 
-    name: "{{ item }}" 
-    install_recommends: yes 
-    force: yes 
-    state: present
-  with_items: mysql_debian_pkgs
+- fail:
+    msg: "You have already installed the experimental MySQL 5.6 repo, Please first remove that"
+  when: "'5.6.14' in  mysql_56_installed.stdout"
```
feanil: This seems like a bit of a time-bomb because eventually 5.6.14 won't be the experimental version and will have been released.  Is the expectation that we will remove this task when that happens?

feanil:
Generally looks good but I am concerned about the failure task for `MySQL 5.6.14`.

playbooks/roles/mysql/tasks/main.yml
arbabnazar: `5.6.14` is old experimental repo for precise and they have launched it stable release, even remove it on 30 Dec for precise(no longer supported PPA of 5.6 for precise). We have found it and it's still available,so we used it on urgent basis to unblock some work.

We use this check for long running old sandboxes and devstack.

arbabnazar:
@e0d @feanil I have address the remove of `5.6.14`, kindly second review. Thanks

e0d:
@feanil what alternative do you propose?

feanil:
@e0d should have put that as a NIT, I don't have a good idea for how to resolve it, so it's fine if we leave it, we just need to remember to remove that task once 5.6.14 is public and not in the experimental repo.

PR #2743, opened 2016-01-29
Add PARTNER_SUPPORT_EMAIL in default FEATURES

caesar2164:
PARTNER_SUPPORT_EMAIL gets used to display contact email for Studio support


openedx-webhooks:
Thanks for the pull request, @caesar2164! I've created [OSPR-1104](https://openedx.atlassian.net/browse/OSPR-1104) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -115,6 +115,8 @@ EDXAPP_VIDEO_CDN_URLS:
 
 EDXAPP_CREDIT_HELP_LINK_URL: ""
 
+EDXAPP_PARTNER_SUPPORT_EMAIL: 'your_support_email@your_platform.com'
```
stvstnfrd: Don't use a real domain here; how about `partner@example.com`?

stvstnfrd:
Context: https://github.com/edx/edx-platform/blob/master/cms/templates/widgets/sock.html#L26

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -115,6 +115,8 @@ EDXAPP_VIDEO_CDN_URLS:
 
 EDXAPP_CREDIT_HELP_LINK_URL: ""
 
+EDXAPP_PARTNER_SUPPORT_EMAIL: 'partner@example.com'
```
stvstnfrd: Actually, looking at how this is used in platform, I don't think this is what we want. Platform code enables this if the key is present, disables it otherwise.

Perhaps a better approach would be to default to `None` in configuration and then update platform to check for a Truthy value, instead of presence of the key?

nedbat:
@andy-armstrong you may want to weigh in on this also.

caesar2164:
Updated PR to match the platform one: https://github.com/edx/edx-platform/pull/11398

andy-armstrong:
@nedbat This looks fine but it probably makes more sense to have someone from DevOps look at configuration changes.

feanil:
Why is the e-mail address in the features dictionary. While we have some historic versions of that patterns I think it is generally a bad idea.  Ideally all new flags going into the features dictionary should be  booleans and other config like this should live at the top level of envs.json

stvstnfrd:
> Why is the e-mail address in the features dictionary. 

@feanil I couldn't say. We aren't adding a new flag here; we're just enabling support for it in configuration.

This was introduced by @peter-fogg [1]; perhaps he can shed some light.
From the looks of it, this work was done out-of-step with DevOps:
- `PARTNER_SUPPORT_EMAIL` was not added to configuration
- `TENDER_DOMAIN` and `TENDER_SUBDOMAIN` were not removed from configuration [2]

> Ideally all new flags going into the features dictionary should be booleans and other config like this should live at the top level of envs.json

I don't disagree. As this flag already exists in platform, I'd consider the change outside the scope of this PR, though we'll gladly review the PRs should anyone at edX decide to update it.

BTW: I have created a PR to remove the `TENDER_*` references from configuration [3]

[1]  https://github.com/edx/edx-platform/pull/9738
[2] https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L145-L146
[3] https://github.com/edx/configuration/pull/2749

peter-fogg:
@feanil @stvstnfrd As a part of my work to introduce the `PARTNER_SUPPORT_EMAIL` I filed [DEVOPS-2981](https://openedx.atlassian.net/browse/DEVOPS-2891) to add the setting to configuration-secure. As Ed mentioned on that ticket, `FEATURES` is intended only for boolean values, which I hadn't realized at the time. I filed [TNL-3363](https://openedx.atlassian.net/browse/TNL-3363) to clean it up.

stvstnfrd:
Thanks for the context @peter-fogg !

feanil:
@stvstnfrd I'd prefer not to merge this and rather get the code fixed up before the feature goes into wider use since it's much harder to not break people once the config goes public.

@peter-fogg can you please work with your team to prioritize this work a bit higher as it now effects people in our community as well.

peter-fogg:
@feanil I've opened https://github.com/edx/edx-platform/pull/11409 to change the feature to a setting. @stvstnfrd, would you mind moving `PARTNER_SUPPORT_EMAIL` in this PR? I'll also open a devops ticket to update configuration-secure to this to a setting.

caesar2164:
thanks @peter-fogg, done.

peter-fogg:
Quick update: opened https://openedx.atlassian.net/browse/DEVOPS-3611.

Also just FYI that the commit message is inaccurate after the change.

feanil:
change looks good to me @edx/devops can I get a second review?

jibsheet:
:+1:

peter-fogg:
@caesar2164 Just FYI that I've merged this. The edx-platform and secure configuration parts were merged last week, so this needed to go in for our release this week.

caesar2164:
awesome, thanks @peter-fogg!

PR #2744, opened 2016-02-01
OPS-1240 Datadog role erroring and causing builds to be marked failed

arbabnazar:


e0d:
:+1: Test failure is unrelated to this code, and being looked at separately.

PR #2745, opened 2016-02-01
OPS-1251 Create an ansible role to install Asqatasun (formerly Tanaguru)

arbabnazar:


playbooks/roles/asqatasun/defaults/main.yml
```diff
@@ -0,0 +1,55 @@
...
+# MySQL variables for Asqatasun
+default_character_set: "utf8"
+collation_server: "utf8_general_ci"
+init_connect: "SET NAMES utf8"
+character_set_server: "utf8"
+mysql_max_allowed_packet: "64M"
+
+ASQATASUN_DATABASE_NAME: 'asqatasun'
+ASQATASUN_DATABASE_USER: 'asqatasun'
+ASQATASUN_DATABASE_PASSWORD: 'asqaP4sswd'
```
feanil: This should be set to `password` or `changeme` so it is obvious that this should be updated in production environments.

playbooks/roles/asqatasun/defaults/main.yml
```diff
@@ -0,0 +1,55 @@
...
+ASQATASUN_DATABASE_HOST: 'localhost'
+ASQATASUN_URL: 'http://localhost:8080/asqatasun/'
+ASQATASUN_ADMIN_EMAIL: 'admin@example.com'
+ASQATASUN_ADMIN_PASSWORD: 'asqatasun15'
```
feanil: This should be update to `password` or `changeme` or something along those lines.

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,175 @@
...
+- name: Restart xvfb
+  service: 
+    name: xvfb 
+    pattern: /etc/init.d/xvfb 
+    state: restarted
+  when: xvfb.changed
+
+- name: Configure xvfb to run at startup
+  command: update-rc.d xvfb defaults
+  ignore_errors: yes
```
feanil: Why are we ignoring errors here?

feanil:
This role assumes that we are always going to run mysql locally.  Is that a reasonable thing to assume?  How big can that database get?

arbabnazar:
@feanil initial size of the database is about `2.5 MB`. Thanks

arbabnazar:
@feanil kindly review. Thanks

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,171 @@
...
+    - install:base
+
+- name: Create MySQL database for Asqatasun
+  mysql_db: 
+    name: "{{ asqatasun_parameters.db_name }}"
+    state: present 
+    encoding: utf8
+    collation: utf8_general_ci
+  tags:
+    - install
```
e0d: should be migrate and migrate:db

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,171 @@
...
+- name: Create MySQL user for Asqatasun
+  mysql_user: 
+    name: "{{ asqatasun_parameters.db_user }}"
+    password: "{{ asqatasun_parameters.db_password }}"
+    host: localhost  
+    priv: "{{ asqatasun_parameters.db_name }}.*:ALL"
+    state: present
+  tags:
+    - install
+    - install:base
```
e0d: same as above.

playbooks/roles/asqatasun/templates/asqatasun.cnf.j2
```diff
@@ -0,0 +1,11 @@
+[client]
```
e0d: Under templates should have the path where the file will end up.  For example

```
templates/etc/mysql
```

playbooks/edx-east/asqatasun.yml
```diff
@@ -0,0 +1,8 @@
+- name: Deploy Asqatasun
+  hosts: all
+  sudo: True
+  gather_facts: True
+  roles:
+    - aws
+    - mysql
```
e0d: Agree with @feanil 's comment.  The install of mysql should be in the play, rather than a dependency of the role.  We are likely to run MySQL in RDS once we go to production with this.  So the DB related tasks will need to be optionally runnable and supported for remote configurations.

playbooks/edx-east/asqatasun.yml
```diff
@@ -0,0 +1,8 @@
+- name: Deploy Asqatasun
+  hosts: all
+  sudo: True
+  gather_facts: True
+  roles:
+    - aws
```
e0d: Is there anything AWS specific in this role?  Seems like it should run on Ubuntu generally.

playbooks/roles/asqatasun/defaults/main.yml
```diff
@@ -0,0 +1,55 @@
...
+  - python-mysqldb
+  - libtomcat7-java
+  - tomcat7 
+  - libspring-instrument-java
+  - xvfb
+  - libdbus-glib-1-2
+  - mailutils
+  - postfix
+
+locale: 'en_US.UTF-8'
```
e0d: This should be up-case and configurable.

playbooks/roles/asqatasun/defaults/main.yml
```diff
@@ -0,0 +1,55 @@
+---
+asqatasun_debian_pkgs:
```
e0d, line 24: Organize so the UPPERCASE vars are at the top, and the things folks should not need to change, lowercase, are at the bottom of the file.

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,171 @@
...
+    copy: no
+  tags:
+    - install
+    - install:base
+
+- name: Create MySQL database for Asqatasun
+  mysql_db: 
+    name: "{{ asqatasun_parameters.db_name }}"
+    state: present 
+    encoding: utf8
```
e0d: Let's make encoding and collation uppercase vars in defaults/main.yml

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,171 @@
...
+    collation: utf8_general_ci
+  tags:
+    - install
+    - install:base
+
+- name: Create MySQL user for Asqatasun
+  mysql_user: 
+    name: "{{ asqatasun_parameters.db_user }}"
+    password: "{{ asqatasun_parameters.db_password }}"
+    host: localhost
```
e0d: host should be configurable.

feanil:
I agree with @e0d's comments.  :+1: once they are addressed.

cptvitamin:
The installation works as expected.  :+1: 
I should mention that the admin user cannot create any "contracts".  I had to create an additional user, then login as that user to create "contracts" or audit jobs.

arbabnazar:
@e0d @feanil kindly second review. Thanks

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+    - install:configuration
+
+- name: Copy the xvfb template to /etc/init.d
+  template:
+    dest: /etc/init.d/xvfb
+    src: etc/init.d/xvfb.j2
+    owner: root
+    group: root
+    mode: 755
+  register: xvfb
```
e0d, line 92: All the tasks should be tagged.  This should have install and install:config

playbooks/roles/asqatasun/tasks/main.yml
```diff
@@ -0,0 +1,172 @@
...
+
+- name: Restart xvfb
+  service: 
+    name: xvfb 
+    pattern: /etc/init.d/xvfb 
+    state: restarted
+    enabled: yes
+  when: xvfb.changed
```
e0d, line 103: Here too

e0d:
After the last two comments, I'm :+1:

e0d:
@feanil 2nd pass please.

arbabnazar:
addressed the last comments. Thanks

feanil:
:+1:

PR #2746, opened 2016-02-01
Ned/unhandle certs xqueue

nedbat:
Converted two handlers into normal tasks.  Notice carefully that I removed the `certs_installed` and `xqueue_installed` facts also.

@feanil ?


nedbat:
@feanil @e0d opinions?

e0d:
:+1: thanks for doing this @nedbat

feanil:
:+1:

PR #2751, opened 2016-02-03
Write to stdout so that vagrant doesn't color it red

nedbat:


nedbat:
@edx/devops Mystery solved.  A Boston Python member last night (hi, @TaymonB!) at project night told me that Vagrant colors stderr red, and it is true.  Changing to log to stdout instead of stderr makes the timings at the end not be red any more.

feanil:
:+1:

PR #2749, opened 2016-02-02
Remove Tender references

stvstnfrd:
The relevant code was removed from edx-platform six months ago [1][2].

[1] https://github.com/edx/edx-platform/pull/9738
[2] fec5129a493e53b10a1b63c2766dc347d9f17790

PR #2755, opened 2016-02-05
Update to not receive input on apt-get tasks.

feanil:
Without the '-y' the tests stop to ask for input.

@edx/devops I'm not sure why we haven't seen issues with this before but https://github.com/edx/configuration/pull/2749 ran into this.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,8 +18,8 @@ addons:
       - python-demjson
 
 before_install:
-  - sudo apt-get update
-  - sudo apt-get install docker-engine
+  - sudo apt-get -y update
+  - sudo apt-get -y install -o Dpkg::Options::="--force-confold" docker-engine
 
 install:
   - "pip install --allow-all-external -r requirements.txt"


feanil:
@cpennington looks like the upgrade of docker is causing issues. I believe the reason for the update/install of docker engine was to get a newer version than what is installed by default is that correct?

feanil:
@cpennington what is the minimum version of docker-engine that you need?

cpennington:
@feanil yes, it was to get a newer version. We need docker new enough to handle build args (1.9)

cpennington:
:+1:

feanil:
The docker tasks are  still failing with this because it is installing Docker 1.10 which was released a few days ago.  It looks like it's having trouble pulling down the bace precise images.  I'm going to merge this now but I think those failures should become tickets on the ecom team to look at.  @cpennington @e0d does that seem reasonable?

feanil:
Merging this change for now since both non-docker tests are passing now so it's an improvement over the current state of master.

PR #2757, opened 2016-02-06
Update sandbox base AMIs.

feanil:
Update the base fully bulit image as well as the 12.04 and 14.04 clean images.

@edx/devops


e0d:
:+1:

PR #2758, opened 2016-02-08
add s3 backend configuration settings

zubair-arbi:
ECOM-3605
@awais786 @ahsan-ul-haq @tasawernawaz 
- Add credentials S3 backend configuration settings with the name `CREDENTIALS_S3_BACKEND`

Related credentials service settings: https://github.com/edx/credentials/pull/54


zubair-arbi:
@maxrothman please review

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,6 +83,20 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#To use AWS S3 as your backend, you need different kwargs:
+# CREDENTIALS_S3_BACKEND:
+#   class: storages.backends.s3boto.S3BotoStorage
+#   options:
+#     bucket: mybucket
+#     custom_domain: mybucket.s3.amazonaws.com
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+#     querystring_auth: false
+#     querystring_expire: false
```
maxrothman: You're gonna need another option to set CORS headers

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,6 +83,20 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+
+# 'access_key' and 'secret_key' are unneccessary if you use IAM roles
+CREDENTIALS_S3_BACKEND: ''
```
maxrothman: Is there really no configuration necessary to use the file storage backend? I thought you'd at least need `class: storages.backends.FileStorage` or something.

Also, if you're making no assumptions about using S3, this variable should not have S3 in the name.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -103,9 +117,12 @@ CREDENTIALS_SERVICE_CONFIG:
 
   PLATFORM_NAME: '{{ CREDENTIALS_PLATFORM_NAME }}'
 
-  AWS_STORAGE_BUCKET_NAME: ''
-  AWS_ACCESS_KEY_ID: ''
-  AWS_SECRET_ACCESS_KEY: ''
+  AWS_STORAGE_BUCKET_NAME: '{{ CREDENTIALS_S3_BACKEND.bucket }}'
```
maxrothman: You have this backwards. You should pass `CREDENTIALS_STORAGE_BACKEND` into `CREDENTIALS_SERVICE_CONFIG`. I should be able to change all details of the storage backend from configuration.

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman can you please link to any existing example?

zubair-arbi:
@maxrothman please review

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,6 +83,32 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+#     querystring_auth: False
+#     querystring_expire: False
+#     headers:
+#       Cache-Control: max-age-{{ CREDENTIALS_CACHE_MAX_AGE }}
+
+# 'access_key' and 'secret_key' are unneccessary if you use IAM roles
+CREDENTIALS_STORAGE_BACKEND:
+  class: storages.backends.s3boto.S3BotoStorage
```
maxrothman: The default storage should be on-disk files. You can do the S3 one as an example in a comment, like there is above. You also haven't set the CORS headers you need. Did you intend to add the cache-control header, or was that from copy/paste?

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -103,9 +129,15 @@ CREDENTIALS_SERVICE_CONFIG:
 
   PLATFORM_NAME: '{{ CREDENTIALS_PLATFORM_NAME }}'
 
-  AWS_STORAGE_BUCKET_NAME: ''
-  AWS_ACCESS_KEY_ID: ''
-  AWS_SECRET_ACCESS_KEY: ''
+  File_STORAGE_BACKEND: '{{ CREDENTIALS_STORAGE_BACKEND.class }}'
```
maxrothman: typo: `File` should be in all caps

Instead of pulling in all of the options by reference individually, simply have a single variable for the whole dict, e.g. `FILE_STORAGE_BACKEND: '{{ CREDENTIALS_STORAGE_BACKEND }}'`. It's also preferable if the names match, so either `CREDENTIALS_FILE_STORAGE_BACKEND` and `FILE_STORAGE_BACKEND` or `CREDENTIALS_STORAGE_BACKEND` and `STORAGE_BACKEND`.

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman In django settings we will use this backend class only if there are valid values for storage backend. django will automatically use default file storage.
I am not sure about the `cache-control` header. I expect that the credentials `scss` and other static assets wont change much. Also there is no CORS headers settings for django storage and I think that CORS settings are added on AWS S3 bucket configuration.

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman I have added these settings so that the `credentials.yml` can have same settings as described in [django-storages](https://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html#settings).

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman please refer the production settings here: https://github.com/edx/credentials/pull/54

zubair-arbi:
@maxrothman please review

zubair-arbi:
@maxrothman please review

zubair-arbi:
@e0d It would be great if you can also review these changes.

playbooks/roles/credentials/defaults/main.yml
maxrothman: Right, I'm asking you to also change the way these are ingested by the django settings so that you basically just instantiate the storage object from these settings. No intelligence, no defaults, just do what the settings say.

RE headers: In S3 all headers are set per-object, not per-bucket. If you want a header to be served for all objects, you need to tell each object to serve it. My understanding is that anything you put in `headers` will be served, so there's nothing special about the cache-control one here. You could put your CORS ones in and they'd work fine, assuming you've wired it up correctly in django settings.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -103,9 +133,15 @@ CREDENTIALS_SERVICE_CONFIG:
...
-  AWS_SECRET_ACCESS_KEY: ''
+  FILE_STORAGE_BACKEND: '{{ CREDENTIALS_STORAGE_BACKEND.class }}'
+
+  AWS_STORAGE_BUCKET_NAME: '{{ CREDENTIALS_STORAGE_BACKEND.options.bucket }}'
+  AWS_CUSTOM_DOMAIN: '{{ CREDENTIALS_STORAGE_BACKEND.options.custom_domain }}'
+  AWS_ACCESS_KEY_ID: '{{ CREDENTIALS_STORAGE_BACKEND.options.access_key }}'
+  AWS_SECRET_ACCESS_KEY: '{{ CREDENTIALS_STORAGE_BACKEND.options.secret_key }}'
+  AWS_QUERYSTRING_AUTH: '{{ CREDENTIALS_STORAGE_BACKEND.options.querystring_auth }}'
+  AWS_QUERYSTRING_EXPIRE: '{{ CREDENTIALS_STORAGE_BACKEND.options.querystring_expire }}'
+  AWS_HEADERS: '{{ CREDENTIALS_STORAGE_BACKEND.options.headers }}'
```
maxrothman: As before, do not disassemble the dictionary here. Simply do `STORAGE_BACKEND: '{{ CREDENTIALS_STORAGE_BACKEND }}'` and make the appropriate changes in django settings to make that work.

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @rlucioni can you please give some example for making CORS work with the credentials django settings?

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -94,7 +123,8 @@ CREDENTIALS_SERVICE_CONFIG:
   SOCIAL_AUTH_EDX_OIDC_URL_ROOT: '{{ CREDENTIALS_OAUTH_URL_ROOT }}/oauth2'
   SOCIAL_AUTH_REDIRECT_IS_HTTPS: '{{ CREDENTIALS_SOCIAL_AUTH_REDIRECT_IS_HTTPS }}'
 
-  STATIC_ROOT: "{{ COMMON_DATA_DIR }}/{{ credentials_service_name }}/staticfiles"
+  MEDIA_ROOT: "{{ CREDENTIALS_DATA_DIR }}/media"
```
zubair-arbi: @cpennington we were getting exception for uploading signatory image with django default file storage.

``` bash
Feb  9 11:17:02 ip-192-168-0-96 [service_variant=credentials][django.request][env:no_env] ERROR [tasawer-credential  27747] [base.py:256] - Internal Server Error: /admin/credentials/signatory/1/
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, callback_args, *callback_kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 618, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, args, *kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/views/decorators/cache.py", line 57, in _wrapped_view_func
    response = view_func(request, args, *kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/contrib/admin/sites.py", line 233, in inner
    return view(request, args, *kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1521, in change_view
    return self.changeform_view(request, object_id, form_url, extra_context)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, args, *kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/utils/decorators.py", line 145, in inner
    return func(*args, **kwargs)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1469, in changeform_view
    self.save_model(request, new_object, form, not add)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/contrib/admin/options.py", line 1080, in save_model
    obj.save()
  File "/edx/app/credentials/credentials/credentials/apps/credentials/models.py", line 123, in save
    super(Signatory, self).save(force_update=True)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/db/models/base.py", line 734, in save
    force_update=force_update, update_fields=update_fields)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/db/models/base.py", line 762, in save_base
    updated = self._save_table(raw, cls, force_insert, force_update, using, update_fields)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/db/models/base.py", line 824, in _save_table
    for f in non_pks]
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/db/models/fields/files.py", line 314, in pre_save
    file.save(file.name, file, save=False)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/db/models/fields/files.py", line 93, in save
    self.name = self.storage.save(name, content, max_length=self.field.max_length)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/core/files/storage.py", line 63, in save
    name = self._save(name, content)
  File "/edx/app/credentials/venvs/credentials/local/lib/python2.7/site-packages/django/core/files/storage.py", line 222, in _save
    os.makedirs(directory)
  File "/edx/app/credentials/venvs/credentials/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 13] Permission denied: '/edx/app/credentials/credentials/credentials/media/signatories/1'
```

My question is that does the credentials role will get permission to upload on `/edx/var/credentials/` and create directories in it?

playbooks/roles/credentials/defaults/main.yml
rlucioni: @zubair-arbi Programs uses https://github.com/ottoyiu/django-cors-headers/ to attach CORS headers to responses. Base settings are [here](https://github.com/edx/programs/blob/master/programs/settings/base.py#L261-L280).

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman any reason for not doing this? This way its easy to override the settings by editing `/edx/etc/credentials.yml`.
@cpennington your thoughts on this?

zubair-arbi: @rlucioni can you please also give your input here?

playbooks/roles/credentials/defaults/main.yml
cpennington: https://github.com/edx/configuration/blob/master/playbooks/roles/edx_service/tasks/main.yml#L82-L94 should be creating `/edx/var/credentials` and giving the service user permissions to write to it. However, your error message states that it's trying to upload to `edx/app/credentials`, which is the wrong place.

playbooks/roles/credentials/defaults/main.yml
cpennington: @maxrothman is correct. By specifying all of the details for the storage backend, it means that we have complete control when it comes to changing the configuration in production w/o needing a code change.

@maxrothman @e0d: This is another place, however, where allowing developers to look at non-secret edx.org-specific configuration would be a) very useful and b) likely take some work off your back.

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: Ok. So If I change the MEDIA_ROOT = '/edx/var/credentials/media/' it should be all well, right?

cpennington: Looks like it, yes.

zubair-arbi:
@maxrothman I have updated the PR, please review

maxrothman:
@zubair-arbi That looks better, except you're making the S3 storage default. Instead, leave a comment on how to set up S3 storage above a config dict for file storage.

zubair-arbi:
@maxrothman I have removed the default storage value so that django's default file storage can be used. Please review

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @cpennington I have tested on my new sandbox with setting `MEDIA_ROOT = '/edx/var/credentials/media/'` and I am still getting exception:

``` python
OSError: [Errno 13] Permission denied: '/edx/var/credentials/media'
```

I checked the `credentials` dir there is no folder with name `media`:

``` bash
credentials@newcert:~/credentials$ ls -al /edx/var/credentials/
total 12
drwxr-xr-x  3 credentials www-data    4096 Feb  8 19:17 .
drwxr-xr-x 19 root        root        4096 Feb  8 19:18 ..
drwxr-xr-x 10 credentials credentials 4096 Feb 10 10:26 staticfiles
```

cpennington: You'll need to add something to `credentials/tasks/default.yml` (in the edx/configuration repo) for creating that folder.

playbooks/roles/credentials/tasks/main.yml
```diff
@@ -119,6 +119,17 @@
     - migrate
     - migrate:db
 
+# var should have more permissive permissions than the rest
+- name: create credentials var dirs
```
zubair-arbi, line 123: @cpennington I have added this task and created a new sandbox with it and it seems to work well for uploading files.

``` yml
MEDIA_URL: /media/
MEDIA_ROOT: /edx/var/credentials/media
```

Now there is only one issue that the media urls are giving `404` and urls are like this:

```
https://credentials-zubair-arbi.sandbox.edx.org/media/signatories/1/abc-signatory_sORZkhg.png
```

any idea how to resolve this?

cpennington, line 123: I think perhaps that we'll need an nginx route to serve up uploaded files.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,18 +83,44 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#   options:
+#     bucket: mybucket
+#     custom_domain: mybucket.s3.amazonaws.com
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+#     querystring_auth: False
+#     querystring_expire: False
+#     bucket_acl: ''
+#     headers:
+#       Cache-Control: max-age-{{ CREDENTIALS_CACHE_MAX_AGE }}
```
cpennington: We need add CORS headers here so that the css served up by S3 will be allowed on pages hosted by credentials.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,18 +83,44 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+
+# 'access_key' and 'secret_key' are unneccessary if you use IAM roles
+CREDENTIALS_STORAGE_BACKEND_SETTINGS: ''
+
```
maxrothman, line 116: Just discussed this with @cpennington, the solution we reached was to store all of the settings required (`AWS_ACCESS_KEY`, etc.) in `CREDENTIALS_FILE_STORAGE_BACKEND` so they're overrideable, but to merge them into `CREDENTIALS_SERVICE_CONFIG` using a [yaml hash merge](http://yaml.org/type/merge.html).

A few of notes:
- The default value for `CREDENTIALS_FILE_STORAGE_BACKEND` should explicitly contain all of the necessary settings to set up file storage
- The commented example above it should be what's needed to set up S3 file storage (with values inserted for access/secret keys)
- Using this approach, no settings relating to storages will be necessary in django settings files

zubair-arbi:
@cpennington @maxrothman Please review and let me know if this is what you suggested.

cpennington:
Looks good on this side. You still need to remove the block [here](https://github.com/edx/credentials/pull/54/files#diff-d2380620490b882124966d3d97429112R35) that is parsing out and explicitly setting AWS configuration.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,69 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#     bucket: mybucket
+#     custom_domain: mybucket.s3.amazonaws.com
+#     access_key: XXXAWS_ACCESS_KEYXXX
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+#     querystring_auth: False
+#     querystring_expire: False
+#     bucket_acl: ''
+#     headers:
+#       Cache-Control: max-age-{{ CREDENTIALS_CACHE_MAX_AGE }}
+#       CORS: mybucket.s3.amazonaws.com
```
cpennington: This should be something like `Access-Control-Allow-Origin: http://mozilla.com`. (See https://developer.mozilla.org/en-US/docs/Web/HTTP/Access_control_CORS for an explanation)

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,69 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+  AWS_QUERYSTRING_EXPIRE: {{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.options.querystring_expire }}
+  AWS_DEFAULT_ACL: {{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.options.bucket_acl }}
+  AWS_HEADERS: {{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.options.headers }}
+
+  COMPRESS_URL: {{ CREDENTIALS_COMPRESS_URL }}
+  STATIC_URL: {{ CREDENTIALS_COMPRESS_URL }}
+  COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
+  COMPRESS_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.class }}
+  STATICFILES_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.class }}
+  DEFAULT_FILE_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.class }}
```
cpennington: Please remove all of these lines. These should all be set by the import from the yaml file based on the values being set by configuration in the `CREDENTIALS_SERVICE_CONFIG` block.

cpennington: Oops. I take it back. I got confused about which PR I was looking at.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,69 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#     secret_key: XXXAWS_SECRETY_KEYXXX
+#     querystring_auth: False
+#     querystring_expire: False
+#     bucket_acl: ''
+#     headers:
+#       Cache-Control: max-age-{{ CREDENTIALS_CACHE_MAX_AGE }}
+#       CORS: mybucket.s3.amazonaws.com
+
+# 'access_key' and 'secret_key' are unneccessary if you use IAM roles
+CREDENTIALS_STORAGE_BACKEND_SETTINGS:
```
maxrothman: Since we're using `CREDENTIALS_FILE_STORAGE_BACKEND`, this dict is unnecessary. Any relevant details should be moved to the new dict.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,69 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+    access_key: ''
+    secret_key: ''
+    querystring_auth: False
+    querystring_expire: False
+    bucket_acl: ''
+    headers: ''
+
+# Settings for django storages to use Amazon S3
+# https://django-storages.readthedocs.org/en/latest/backends/amazon-S3.html#amazon-s3
+CREDENTIALS_COMPRESS_URL: 'https://{{ CREDENTIALS_STORAGE_BACKEND_SETTINGS.options.custom_domain }}/'
```
maxrothman: This should be moved inside of `CREDENTIALS_FILE_STORAGE_BACKEND`

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,69 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+CREDENTIALS_FILE_STORAGE_BACKEND:
```
maxrothman, line 114: - Move all of the values inside here. We'll override this structure in production.
- Change the keys in this dict to be only the ones required for file storage
- Leave a commented version of this structure with the required variables for S3

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -103,9 +161,10 @@ CREDENTIALS_SERVICE_CONFIG:
 
   PLATFORM_NAME: '{{ CREDENTIALS_PLATFORM_NAME }}'
 
-  AWS_STORAGE_BUCKET_NAME: ''
-  AWS_ACCESS_KEY_ID: ''
-  AWS_SECRET_ACCESS_KEY: ''
+  # Set credentials files storage backend
+  << : '{{ CREDENTIALS_FILE_STORAGE_BACKEND }}'
```
maxrothman: This will not work as expected. A subtlety of Ansible is that there are 2 different syntaxes for referencing variables: the `{{ var }}` one which works across files (provided by the Ansible software) and the other ([node anchors and alias nodes](http://yaml.org/spec/1.2/spec.html#& anchor//)) is part of the YAML spec. Here, you want to use the YAML syntax since the hash merge is done in YAML. That means you have to make an anchor at `CREDENTIALS_FILE_STORAGE_BACKEND` and use an alias node in the hash merge.

zubair-arbi:
@cpennington @maxrothman I have refined the configuration. Please review.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,47 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
 CREDENTIALS_USER_API_JWT_SECRET_KEY: 'lms-secret'
 
+CREDENTIALS_STATIC_ROOT: '{{ COMMON_DATA_DIR }}/{{ credentials_service_name }}/staticfiles'
+
+# CORS
+# See: https://github.com/ottoyiu/django-cors-headers/.
+CREDENTIALS_CORS_ORIGIN_WHITELIST: []
+
+CREDENTIALS_CACHE_MAX_AGE: 31536000
+CREDENTIALS_STORAGE_BACKEND_CLASS: 'django.core.files.storage.DefaultStorage'
```
maxrothman: Move both of these into `CREDENTIALS_FILE_STORAGE_BACKEND`. We probably will not need to override them separately.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,47 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#    Access-Control-Allow-Origin: https://127.0.0.1:8150/
+#
+#  COMPRESS_URL: 'https://mybucket.s3.amazonaws.com/'
+#  STATIC_URL: 'https://mybucket.s3.amazonaws.com/'
+#  COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
+#  COMPRESS_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
+#  STATICFILES_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
+#  DEFAULT_FILE_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
+
+credentials_file_storage_backend: &CREDENTIALS_FILE_STORAGE_BACKEND
```
maxrothman: Make this uppercase. Our convention is that uppercase vars are intended to be overridden, whereas lowercase vars are "private".

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,47 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#  AWS_HEADERS:
+#    Cache-Control: max-age-{{ CREDENTIALS_CACHE_MAX_AGE }}
+#    Access-Control-Allow-Origin: https://127.0.0.1:8150/
+#
+#  COMPRESS_URL: 'https://mybucket.s3.amazonaws.com/'
+#  STATIC_URL: 'https://mybucket.s3.amazonaws.com/'
+#  COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
+#  COMPRESS_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
+#  STATICFILES_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
+#  DEFAULT_FILE_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
```
maxrothman: Do the above 6 only get overridden when using S3?

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,47 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#  AWS_STORAGE_BUCKET_NAME: mybucket
+#  AWS_CUSTOM_DOMAIN: mybucket.s3.amazonaws.com
+#  AWS_ACCESS_KEY_ID: XXXAWS_ACCESS_KEYXXX
+#  AWS_SECRET_ACCESS_KEY: XXXAWS_SECRETY_KEYXXX
+#  AWS_QUERYSTRING_AUTH: False
+#  AWS_QUERYSTRING_EXPIRE: False
+#  AWS_DEFAULT_ACL: ''
+#  AWS_HEADERS:
+#    Cache-Control: max-age-{{ CREDENTIALS_CACHE_MAX_AGE }}
+#    Access-Control-Allow-Origin: https://127.0.0.1:8150/
```
maxrothman: This will not work in production, since to a browser, that'll be the user's machine. If you intended it as a placeholder, then just replace it with `PUT YOUR HOSTNAME HERE` or something.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,47 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
 CREDENTIALS_USER_API_JWT_AUDIENCE: 'lms-key'
 CREDENTIALS_USER_API_JWT_SECRET_KEY: 'lms-secret'
 
+CREDENTIALS_STATIC_ROOT: '{{ COMMON_DATA_DIR }}/{{ credentials_service_name }}/staticfiles'
+
+# CORS
+# See: https://github.com/ottoyiu/django-cors-headers/.
+CREDENTIALS_CORS_ORIGIN_WHITELIST: []
```
maxrothman: Out of curiosity, what do you need this for? Is a site from a different domain that will be accessing resources from this app?

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman yes

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: @maxrothman Nope. I will remove CORS settings since these are unnecessary.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -83,11 +83,37 @@ CREDENTIALS_USER_API_URL: 'https://127.0.0.1:8000/api/user/v1/'
...
+#
+#  COMPRESS_URL: 'https://mybucket.s3.amazonaws.com/'
+#  STATIC_URL: 'https://mybucket.s3.amazonaws.com/'
+#  COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
+#  COMPRESS_STORAGE: storages.backends.s3boto.S3BotoStorage
+#  STATICFILES_STORAGE: storages.backends.s3boto.S3BotoStorage
+#  DEFAULT_FILE_STORAGE: storages.backends.s3boto.S3BotoStorage
+
+CREDENTIALS_FILE_STORAGE_BACKEND: &CREDENTIALS_FILE_STORAGE_BACKEND
+  DEFAULT_FILE_STORAGE: {{ CREDENTIALS_STORAGE_BACKEND_CLASS }}
```
maxrothman: You need quotes around this

EDIT: actually, this variable is no longer defined. Did you intend to put its value here?

maxrothman:
Having thought about this further, I think the current strategy for hash merging will not work because the YAML merge will get processed before Ansible resolves variables, meaning the override will never get used. I think the way forward is to unroll that dict inside `production.py` by the same mechanism it unrolls the settings from the YAML file (i.e. `vars().update()`.

maxrothman:
:+1: after tests

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -113,9 +139,8 @@ CREDENTIALS_SERVICE_CONFIG:
 
   PLATFORM_NAME: '{{ CREDENTIALS_PLATFORM_NAME }}'
 
-  AWS_STORAGE_BUCKET_NAME: ''
-  AWS_ACCESS_KEY_ID: ''
-  AWS_SECRET_ACCESS_KEY: ''
+  # Set credentials files storage backend
+  FILE_STORAGE_BACKEND: '{{ CREDENTIALS_FILE_STORAGE_BACKEND }}'
```
zubair-arbi, line 143: @cpennington @feanil @maxrothman wouldn't this fix the issue for `FILE_STORAGE_BACKEND`?

zubair-arbi:
@maxrothman @cpennington @feanil I am going to merge this PR now.
@maxrothman If we will need some other changes for credentials configuration then we will have a separate PR.

PR #2761, opened 2016-02-08
Split CSMH configuration

fredsmith:


feanil:
There needs to be an update to edxlocal to make sure this DB exists.

e0d:
@fredsmith should this be closed?

PR #2763, opened 2016-02-08
Remove ruby from jenkins-workers.

benpatterson:
@feanil @ormsbee @jibsheet @cpennington 

Also planning to run a packer build once I get stuff wired into edx-platform.

Note this change includes an improvement to allow us to build the venv based on a specific edx-platform branch. That'll help with testing this and similar kinds of changes.

PR #2767, opened 2016-02-09
Benp/remove ruby jenkins worker

benpatterson:
See #2763


benpatterson:
Also verified on packer builds against the related edx-platform branch.

feanil:
:+1:

PR #2780, opened 2016-02-16
OPS-1327 Allow User to specify the lifetime of the sandbox when creating

arbabnazar:

Reference:

http://xxxxx.edx.org:8080/job/ansible-provision-2/2/console

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -37,7 +37,13 @@ run_ansible() {
 }
 
 # This DATE_TIME will be used as instance launch time tag
-TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-7 days ago"`
+if [[ ! -n ${sandbox_life//[0-9]/} ]]  && [[ ${sandbox_life} -le 30 ]]; then
+    TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date "${sandbox_life=7} days"`
+else
+   echo "Please enter the valid value for the sandbox_life(between 1 to 30)"
+   exit 1
+fi
+
 
 if [[ -z $BUILD_USER ]]; then
     BUILD_USER=jenkins


util/jenkins/ansible-provision.sh
```diff
@@ -37,7 +37,7 @@ run_ansible() {
 }
 
 # This DATE_TIME will be used as instance launch time tag
-TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-7 days ago"`
+TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-${sandbox_life} days ago"`
```
feanil: We should set a default for this if it is not set.  `if [[ -z $sandbox_life ]]`

maxrothman: Actually, that will also fail if `sandbox_life` isn't defined. You could quote it, e.g. `[[ -z "$sandbox_life" ]]`, but I think the better option would be to use `${sandbox_life-default}` ([ref](http://www.tldp.org/LDP/abs/html/parameter-substitution.html#PARAMSUBREF)). If `sandbox_life` is used elsewhere in the script, you might instead want to use `${sandbox_life=default}`.

arbabnazar:
We have default value of it( 8 days) from jenkins. Thanks

arbabnazar:
@e0d kindly review it, I have added the validation check for the number of days. Thanks

arbabnazar:
@feanil can you please review this, the default value of this variable will come from here:
screen shot 2016-02-17 at 5 58 14 pm
Thanks,

util/jenkins/ansible-provision.sh
```diff
@@ -37,7 +37,13 @@ run_ansible() {
 }
 
 # This DATE_TIME will be used as instance launch time tag
-TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-7 days ago"`
+if [[ ! -n ${sandbox_life//[0-9]/} ]]  && [[ ${sandbox_life} -le 30 ]]; then
+    TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-${sandbox_life} days ago"`
```
fredsmith: this wording is confusing to read - why not just say "${sandbox_life} days"?

```
$ date +"%m-%d-%Y %T" --date "8 days"
02-25-2016 08:57:55
```

arbabnazar:
@fredsmith fix the message, kindly review. Thanks

maxrothman:
As per @feanil's previous comment, there should be a default if `sandbox_life` is not set.

arbabnazar:

maxrothman:
@arbabnazar as per my previous comment, you can get rid of the if-statement if you use `${sandbox_life-default}` or `${sandbox_life=default}`.

arbabnazar:
fixed, thanks for the tips @maxrothman

maxrothman:
:+1:

util/jenkins/ansible-provision.sh
```diff
@@ -37,7 +37,13 @@ run_ansible() {
 }
 
 # This DATE_TIME will be used as instance launch time tag
-TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-7 days ago"`
+if [[ ! -n ${sandbox_life//[0-9]/} ]]  && [[ ${sandbox_life} -le 30 ]]; then
+    TERMINATION_DATE_TIME=`date +"%m-%d-%Y %T" --date="-${sandbox_life:-7} days"`
```
fredsmith: you took out the 'ago', but you left the negative.  This won't do what you want.

Also, your default has a negative in it, and you're prefixing your string with a negative, so make sure you get rid of both.

arbabnazar:
@fredsmith thanks, I have fixed it as per your suggestion.

PR #2781, opened 2016-02-16
update s3 configuration to use custom credentials s3 backends

zubair-arbi:
@awais786 @ahsan-ul-haq @tasawernawaz @waheedahmed 
Update the configuration for credentials to use custom S3 backends for creating subfolders in the bucket.

Related credentials changes: https://github.com/edx/credentials/pull/60


zubair-arbi:
@cpennington @maxrothman please review. With these settings the S3 bucket will have to base folders 'media' and 'staticfiles' for storing media and static files/folders respectively.

maxrothman:
- Please use a consistent quoting style
- Have you verified that including a path in the bucket URL works as expected?
- If so, I think it would be preferable to define the bucket and path separately and then combine the variables.

zubair-arbi:
@maxrothman 
1. I will do that
2. Yes
3. I don't think we can do that. This prefixing is consistent with the folder structure for default storage.

maxrothman:
3. You can do it in Ansible. Just define 2 different vars and combine them like `{{ BUCKET }}/{{ PATH }}`.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -104,12 +104,13 @@ CREDENTIALS_USER_API_JWT_SECRET_KEY: 'lms-secret'
...
-#  COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
-#  COMPRESS_STORAGE: storages.backends.s3boto.S3BotoStorage
-#  STATICFILES_STORAGE: storages.backends.s3boto.S3BotoStorage
-#  DEFAULT_FILE_STORAGE: storages.backends.s3boto.S3BotoStorage
+#  COMPRESS_URL: 'https://mybucket.s3.amazonaws.com/staticfiles/'
+#  STATIC_URL: 'https://mybucket.s3.amazonaws.com/staticfiles/'
+#  COMPRESS_STORAGE: credentials.apps.core.s3utils.StaticS3BotoStorage
+#  STATICFILES_STORAGE: credentials.apps.core.s3utils.StaticS3BotoStorage
+#
+#  AWS_LOCATION: 'media'
```
zubair-arbi: @cpennington @maxrothman I have verified that we can use this parameter `AWS_LOCATION` to add a default prefix for all the files.
https://github.com/jschneier/django-storages/blob/master/storages/backends/s3boto.py#L217

I have kept our custom backend for the staticfiles and using the parameter `AWS_LOCATION` we can add prefix for media files.
With about config we will have S3 bucket structure like:

```
mybucket
├───media (cutomizable with parameter 'AWS_LOCATION')
├───staticfiles
```

Please let me know your thoughts.

zubair-arbi:
@maxrothman Please let me know if you are ok with these changes.

maxrothman:
@zubair-arbi you haven't addressed any of my comments

zubair-arbi:
@maxrothman, I have talked to @cpennington and we agreed to keep the current settings.

maxrothman:
My comments are about the current quoting style consistency and the way the variables are put together in ansible. Neither of them require any changes on the app side.

zubair-arbi:
@maxrothman please review the changes.

zubair-arbi:
@jimabramson @maxrothman after this PR is merged then we don't need to override `STATIC_URL` and `MEDIA_URL` for AWS S3.

maxrothman:
@zubair-arbi why not?

zubair-arbi:
Since after these changes we won't need to change the the paths as both the static files and media files will be in different folders.
Also according to my understanding we will have separate buckets for multiple environment.

zubair-arbi:
@maxrothman I would also like to know what kind of issues we will face with these settings and your suggestions to resolve them.

maxrothman:
Won't we need to override `STATIC_URL` to something like `https://bucket.s3.amazonaws.com/static` and the same for media?

zubair-arbi:
@maxrothman I have verified that we will only need to override the `STATIC_URL` like:

```
STATIC_URL: 'https://mybucket.s3.amazonaws.com/staticfiles/'
```

but we don't need to update the `MEDIA_URL`.

zubair-arbi:
@rlucioni @jimabramson @maxrothman I have updated this PR according to the change by @rlucioni on the [credential PR](https://github.com/edx/credentials/pull/60). I have also move the files settings inside the dict `CREDENTIALS_FILE_STORAGE_BACKEND`.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -93,25 +93,31 @@ CREDENTIALS_USER_API_JWT_SECRET_KEY: 'lms-secret'
...
+#  MEDIA_URL: 'https://mybucket.s3.amazonaws.com/media/'
+#  COMPRESS_STORAGE: 'credentials.apps.core.s3utils.StaticS3BotoStorage'
+#  STATICFILES_STORAGE: 'credentials.apps.core.s3utils.StaticS3BotoStorage'
+#  DEFAULT_FILE_STORAGE: 'credentials.apps.core.s3utils.MediaS3BotoStorage'
 
 CREDENTIALS_FILE_STORAGE_BACKEND:
+  MEDIA_ROOT: '{{ CREDENTIALS_MEDIA_ROOT }}'
+  MEDIA_URL: '{{ CREDENTIALS_MEDIA_URL }}'
+  STATIC_ROOT: '{{ CREDENTIALS_STATIC_ROOT }}'
+  STATIC_URL: '{{ CREDENTIALS_STATIC_URL }}'
```
rlucioni, line 119: Why are these here? These are settings for `django.contrib.staticfiles`. They're not particular to the storage backend in use.

zubair-arbi, line 119: I tried to group all the file settings here. As for AWS these will change.

zubair-arbi, line 119: @maxrothman is this change you were suggesting?

rlucioni, line 119: @zubair-arbi I'm confused. These are settings typically found at the top level of a Django settings module, not nested within another setting. This change is also incompatible with changes to https://github.com/edx/credentials/pull/60.

zubair-arbi, line 119: @rlucioni we pass the `CREDENTIALS_FILE_STORAGE_BACKEND` as `FILE_STORAGE_BACKEND` to django settings and the dict key are imported as django setting vars.
https://github.com/edx/credentials/blob/master/credentials/settings/production.py#L24

rlucioni, line 119: @zubair-arbi I see. So then these do end up as top-level settings.

rlucioni:
:+1:

maxrothman:
I'll review this shortly.

zubair-arbi:
@maxrothman Please review

maxrothman:
I will review this as soon as I'm done dealing with the security hole fallout.

zubair-arbi:
@maxrothman I am still waiting for your review.

maxrothman:
@zubair-arbi I've pushed a change that cleans up and organizes the vars. Take a look at it, but after tests pass and a squash, I'm :+1:

jimabramson:
@maxrothman had a couple side questions but it looks sane to me.  Though I'm having trouble deciphering what went wrong with the build.

zubair-arbi:
@jimabramson @maxrothman the build failed with following error:

```
fatal: [127.0.0.1] => {'msg': "AnsibleUndefinedVariable: One or more undefined variables: 'CREDENTIALS_MEDIA_ROOT' is undefined", 'failed': True}
```

I will update the PR.

zubair-arbi:
@cpennington @jimabramson @maxrothman with recent changes the [credentials sandbox is failing](http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/5791/console) on `create databases` step:

```
11:35:09 TASK: [credentials | create databases] **************************************** 
11:35:10 failed: [zubair-arbi.sandbox.edx.org] => {"failed": true}
11:35:10 msg: unable to connect, check login credentials (login_user, and login_password, 
which can be defined in ~/.my.cnf), check that mysql socket exists and mysql server is running
```

zubair-arbi:
@maxrothman @e0d @jimabramson Please review.

maxrothman:
@zubair-arbi can you also set `CREDENTIALS_DB_ADMIN_PASSWORD` to a hyphenated string rather than an empty one?

zubair-arbi:
@maxrothman I have add new update value for `CREDENTIALS_DB_ADMIN_PASSWORD`. Now waiting for sandbox build to pass.

zubair-arbi:
@maxrothman @cpennington the build fails if we provide any value for `CREDENTIALS_DB_ADMIN_PASSWORD`.

```
15:18:05 TASK: [credentials | create databases] **************************************** 
15:18:06 failed: [zubair-arbi.sandbox.edx.org] => {"failed": true}
15:18:06 msg: unable to connect, check login credentials (login_user, and login_password, 
which can be defined in ~/.my.cnf), check that mysql socket exists and mysql server is running
```

maxrothman:
:+1:

zubair-arbi:
@maxrothman @jimabramson I have fixed the static/media urls and root settings, also added `COMPRESS_STORAGE` settings for AWS S3 settings.
Now only issue remaining is OperationalError on accessing database:

```
OperationalError: (1045, "Access denied for user 'credentials001'@'localhost' (using password: YES)")
```

The full traceback can be found here: https://gist.github.com/zubair-arbi/dcebaa88d8f1e2d28bfe

zubair-arbi:
The database connection works when I provide `PASSWORD` with value `password` instead of `SET-ME-TO-A-UNIQUE-LONG-RANDOM-STRING`.

zubair-arbi:
Ok. I just build another sandbox with `server_type` set to `full_edx_installation_from_scratch` and now the mysql connection is working. I think previously the values were not properly being updated.

I am now going to squash the commits. @jimabramson @maxrothman are you comfortable with these final changes?

maxrothman:
- Why did you capitalize `credentials_static_root` and `credentials_media_root`? 
- Is the gist of the fix to use `MediaRootS3BotoStorage` so you can set your own path, rather than use the default one?

zubair-arbi:
@maxrothman since you have to give separate values for these parameters for AWS S3.

zubair-arbi:
@maxrothman for your second question 'Yes, this way we will use our own path inside the AWS S3 bucket' and also its `MediaS3BotoStorage` not `MediaRootS3BotoStorage`.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -76,43 +65,80 @@ CREDENTIALS_SERVICE_USER: 'credentials_service_user'
...
-#  COMPRESS_ROOT: {{ CREDENTIALS_STATIC_ROOT }}
-#  COMPRESS_STORAGE: storages.backends.s3boto.S3BotoStorage
-#  STATICFILES_STORAGE: storages.backends.s3boto.S3BotoStorage
-#  DEFAULT_FILE_STORAGE: storages.backends.s3boto.S3BotoStorage
+#  MEDIA_ROOT: '{{ CREDENTIALS_MEDIA_ROOT }}'
+#  STATIC_ROOT: '{{ CREDENTIALS_STATIC_ROOT }}'
+#  MEDIA_URL:    'https://{{ CREDENTIALS_BUCKET }}.{{ credentials_s3_domain }}{{ CREDENTIALS_MEDIA_URL }}'
+#  STATIC_URL:   'https://{{ CREDENTIALS_BUCKET }}.{{ credentials_s3_domain }}{{ CREDENTIALS_STATIC_URL }}'
+#
+#  COMPRESS_STORAGE: 'credentials.apps.core.s3utils.StaticS3BotoStorage'
```
maxrothman: Remove this line (and set them the same in app settings)

maxrothman:
After (small) comment addressed, :+1: . I'll let you know when you can merge after AMIs are built.

playbooks/roles/credentials/defaults/main.yml
zubair-arbi: You mean in django? My suggestion is to keep this setting here since its related to AWS S3 configuration.

zubair-arbi:
@max I have squashed the commits. Please review

playbooks/roles/credentials/defaults/main.yml
maxrothman: Yeah, but there's no use-case for making the values different. The fact that django compressor is involved is an implementation detail, so there's no reason to expose it in configuration.

maxrothman:
:+1:

PR #2783, opened 2016-02-16
Max/small fixes

maxrothman:
More details are in the commit messages.


fredsmith:
:+1: - looking back through the history of this part of the play, I can't see why it should need to run locally.

PR #2784, opened 2016-02-17
Add config for SUPPORT_URL

mjfrey:
Add SUPORT_URL placeholder in ecommerce config.


mjfrey:
@clintonb

mjfrey:
@fredsmith FYI, this is the config setting I mentioned earlier.

clintonb:
:+1: (once commits squashed)

PR #2788, opened 2016-02-18
DEVOPS-3685 ensure tools gets the same version the mongo play installs

jibsheet:
While fixing a downgrade from 3 -> 2 on tools, ensure this role uses the
same version the mongo_3 play uses.


jibsheet:
@edx/devops

feanil:
:+1:

PR #2793, opened 2016-02-19
Remove almost all references to syncdb 🔥🔥🔥

jibsheet:
Django 1.8 removed the syncdb command so we can no longer run it, don't
imply that we are running it in documentation/messages, also don't
provide the scripts used on devstack/sandboxes to run it since they'll
never work anyway.

@edx/devops

### DIFF
--- a/playbooks/edx-east/ora2.yml
+++ b/playbooks/edx-east/ora2.yml
@@ -32,7 +32,7 @@
       tags:
         - deploy
 
-    - name: syncdb and migrate
+    - name: migrate
       shell: >
         {{ edxapp_venv_dir }}/bin/python manage.py lms migrate --settings=aws
         chdir={{ edxapp_code_dir }}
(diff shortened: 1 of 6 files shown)


jibsheet:
image

macdiesel:
Burn it all!!!

playbooks/edx-east/test_output/env-dep/cms.auth.json
```diff
@@ -0,0 +1,160 @@
+
```
maxrothman: Was this file added in error?

playbooks/edx-east/test_output/env-dep/cms.auth.yaml
```diff
@@ -0,0 +1,124 @@
+ANALYTICS_API_KEY: ''
```
maxrothman: Ditto

jibsheet:
@maxrothman no idea, looks like test detritus, but how that would be in my config repo is beyond me.  I killed off the cruft.

feanil:
:+1:

maxrothman:
:+1:

PR #2795, opened 2016-02-22
add pull request template

fredsmith:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,7 @@
+Configuration Pull Request
+---
+
+Make sure that the following steps are done before merging
+
+  - [ ] @devops team member has commented with :+1
+  - [ ] are you adding any new default values that need to be overriden when this goes live?  Make sure that they have been added to secure vars.
```
jibsheet: Should this be "Open a ticket to have these set in the secure configuration" to make it clearer how to do that?

feanil: We should also add a `If you are adding new defaults, also add an entry to the changelog.`

PR #2801, opened 2016-02-23
Jibsheet/migrate multiple databases

jibsheet:
Change the edxapp role, edxapp_migrate play and pre_supervisor to support multiple databases.

pre_supervisor leverages changes to edxapp to do so.  If we don't like the pre_supervisor changes, we'll need to go the route of https://github.com/edx/configuration/commit/8c893e24d6b46c3b26e3a70eda51633e91dc6a5e possibly having it autogen the databases from lms.auth.json.  This seemed better.

### DIFF
--- a/playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
+++ b/playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
@@ -1,7 +1,11 @@
 {% include "edxapp_common.j2" %}
 
+if [[ -z "$NO_EDXAPP_SUDO" ]]; then
+    SUDO='sudo -E -u {{ edxapp_user }} env "PATH=$PATH"'
+fi
+
 {% for db in cms_auth_config.DATABASES.keys() %}
   {%- if db != 'read_replica' %}
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py cms migrate --database {{ db }} --noinput --settings $EDX_PLATFORM_SETTINGS $@
+$SUDO {{ edxapp_venv_bin}}/python manage.py cms migrate --database {{ db }} --noinput --settings $EDX_PLATFORM_SETTINGS $@
   {% endif %}
 {% endfor %}
--- a/playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
+++ b/playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
@@ -2,6 +2,6 @@
 
 {% for db in cms_auth_config.DATABASES.keys() %}
   {%- if db != 'read_replica' %}
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py cms migrate --database {{ db }} --noinput --settings $EDX_PLATFORM_SETTINGS
+sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py cms migrate --database {{ db }} --noinput --settings $EDX_PLATFORM_SETTINGS $@
   {% endif %}
 {% endfor %}
--- a/playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
+++ b/playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
@@ -1,3 +1,7 @@
 {% include "edxapp_common.j2" %}
 
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py cms migrate --noinput --settings $EDX_PLATFORM_SETTINGS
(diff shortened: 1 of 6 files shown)


jibsheet:
@edx/devops I've run check-migrations-stage-edx, built a sandbox and booted an AMI using the newer version of pre_supervisor.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -136,6 +136,7 @@
   tags:
     - install
     - install:configuration
+    - edxapp_helper_scripts
```
maxrothman: Why the new tag?

playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
```diff
@@ -1,3 +1,11 @@
 {% include "edxapp_common.j2" %}
 
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py cms migrate --noinput --settings $EDX_PLATFORM_SETTINGS
+if [[ -z "$NO_EDXAPP_SUDO" ]]; then
+    SUDO='sudo -E -u {{ edxapp_user }} env "PATH=$PATH"'
+fi
+
+{% for db in edxapp_generic_auth_config.DATABASES.keys() %}
+  {%- if db != 'read_replica' %}
```
maxrothman, line 8: I'm a little nervous that the new significance of database names could be considered surprising. Could you add a comment near the database structure definition?

playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-lms.j2
```diff
@@ -1,3 +1,11 @@
 {% include "edxapp_common.j2" %}
 
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py lms migrate --noinput --settings $EDX_PLATFORM_SETTINGS
+if [[ -z "$NO_EDXAPP_SUDO" ]]; then
+    SUDO='sudo -E -u {{ edxapp_user }} env "PATH=$PATH"'
+fi
+
+{% for db in edxapp_generic_auth_config.DATABASES.keys() %}
+  {%- if db != 'read_replica' %}
```
maxrothman, line 8: Same as above

playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-cms.j2
```diff
@@ -1,3 +1,11 @@
 {% include "edxapp_common.j2" %}
 
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py cms migrate --noinput --settings $EDX_PLATFORM_SETTINGS
+if [[ -z "$NO_EDXAPP_SUDO" ]]; then
+    SUDO='sudo -E -u {{ edxapp_user }} env "PATH=$PATH"'
+fi
+
+{% for db in edxapp_generic_auth_config.DATABASES.keys() %}
```
feanil: Use `cms_auth_config` instead since users could override that var and that is what is actually used to generate the config files.

playbooks/roles/edxapp/templates/edx/bin/edxapp-migrate-lms.j2
```diff
@@ -1,3 +1,11 @@
 {% include "edxapp_common.j2" %}
 
-sudo -E -u {{ edxapp_user }} env "PATH=$PATH" {{ edxapp_venv_bin}}/python manage.py lms migrate --noinput --settings $EDX_PLATFORM_SETTINGS
+if [[ -z "$NO_EDXAPP_SUDO" ]]; then
+    SUDO='sudo -E -u {{ edxapp_user }} env "PATH=$PATH"'
+fi
+
+{% for db in edxapp_generic_auth_config.DATABASES.keys() %}
```
feanil: Use `lms_auth_config` instead since users could override that var and that is what is actually used to generate the config files.

util/jenkins/check-migrations.sh
```diff
@@ -64,6 +64,8 @@ ansible-playbook -c local $extra_var_args --tags edxapp_cfg -i localhost, -s -U
 # Run migrations and replace literal '\n' with actual newlines to make the output
 # easier to read
 
+extra_var_args+=" -e@${WORKSPACE}/lms.auth.yaml"
```
feanil: Please add a comment for why this is needed.

feanil:
Instead of adding the `extra_var_args` for the yaml file.  A good approach might be to change the `edxapp_migrate.yml` playbook so that it runs the edxapp role first with the tags and overrides needed to generate the yaml and then runs the management command.  This would allow the `edxapp_migrate.yml` playbook to get access to all the correct overrides in the same way as we do for other ansible runs.

jibsheet:
I believe notes have been addressed now.
I was able to eventually make the technique Feanil and I discussed about combining the two plays into one to share variables actually work.

feanil:
:+1:

maxrothman:
:+1:

PR #2803, opened 2016-02-23
New playbook and role for go-server

macdiesel:
@feanil @doctoryes PTAL.


doctoryes:
LGTM - so at the end of this install, GitHub OAuth login will be functional? Or is there further configuration to do?

macdiesel:
The OAuth agent will be installed. 
you can override GO_SERVER_ADMIN_USERS to set the admin users who will then be able to log in to the system.

The unfortunate part is that the plugin does not currently allow you to limit login to only users with an edx account or via group membership. So for the time being we will need to manually add users to the system.

doctoryes:
Unfortunate indeed. @feanil Should we capture items like this inability to limit login to users in a GitHub group as tickets representing work on an external open-source package? Guess it depends on the situation.

:+1:

feanil:
:+1:   Story is on the un-groomed backlog.

PR #2804, opened 2016-02-23
New playbook and role for go-agent

macdiesel:
@feanil @doctoryes PTAL.


doctoryes:
What additional configuration would remain after configuring an agent with this script?
How would multiple GoCD agents be configured on the same machine? Is that something we'd ever want to do - say for the sandbox GoCD instance?

macdiesel:
@doctoryes:

> What additional configuration would remain after configuring an agent with this script?

None assuming that either go-server and go-agent were installed on the same machine or these values: 
- GO_AGENT_SERVER_IP
- GO_AGENT_SERVER_PORT

were overridden to point to the correct go-server for that environment.

> How would multiple GoCD agents be configured on the same machine?

By default only a single agent installation is set for the install. I will take a look to see what options are available if we want to run more than a single agent on a server. When you were evaluating GoCD did you see any options in regard to parallelism for the agents? I had assume that since this was java the parallelism would be driven via threading and JVM tuning.

> Is that something we'd ever want to do - say for the sandbox GoCD instance?

Possibly but lets cross that bridge when we come to it.

feanil:
:+1:

doctoryes:
:+1:

PR #2802, opened 2016-02-23
Move credentials and discovery to edxlocal for database setup

cpennington:
N.B. This may break docker-based devstacks. Fix TBD.

@maxrothman, @jimabramson, @zubair-arbi: I think this will fix the current credentials issues.

PR #2805, opened 2016-02-23
credentials config

jimabramson:
@maxrothman @zubair-arbi @cpennington 

Here's a branch with Cale's fix from #2802 and the meaningful commit from #2781 (removing all the follow-on troubleshooting commits)


jimabramson:
we are sticking with #2781, and the changes here have already been added to that branch by other means

PR #2807, opened 2016-02-23
Include ORGANIZATIONS_API_URL_ROOT in programs defaults

rlucioni:
This setting is used when requesting data from the LMS Organizations API. This shouldn't require any additional overrides if PROGRAMS_LMS_URL_ROOT is already overridden.

@maxrothman you appear to be on-call this week. Would you mind taking a quick look at this?


fredsmith:
@rlucioni  - since this variable is part of a hash, typically we would reference a globally scoped variable in the hash (eg. PROGRAMS_ORGANIZATIONS_API_URL_ROOT) and then set the default in the global variable.  Ansible's hash merging is problematic, and there is otherwise no other way to override a single component of a default hash.

rlucioni:
@fredsmith does that mean that if we were to merge this as-is, we'd have no way to override ORGANIZATIONS_API_URL_ROOT?

fredsmith:
That's correct.

fredsmith:
:+1:

rlucioni:
@fredsmith understood, thanks for the explanation. I've updated this PR to use globally-scoped variables in the hash.

PR #2809, opened 2016-02-25
WL-286 Added new ecommerce payment processor settings

douglashall:
Added new payment processor settings for multitenant support changes in ecommerce

Related PR:

https://github.com/edx/ecommerce/pull/543

### DIFF
--- a/playbooks/roles/ecommerce/defaults/main.yml
+++ b/playbooks/roles/ecommerce/defaults/main.yml
@@ -68,6 +68,8 @@ ECOMMERCE_CYBERSOURCE_TRANSACTION_KEY: 'SET-ME-PLEASE'
 ECOMMERCE_CYBERSOURCE_PAYMENT_PAGE_URL: 'https://set-me-please'
 ECOMMERCE_CYBERSOURCE_RECEIPT_PAGE_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/receipt/'
 ECOMMERCE_CYBERSOURCE_CANCEL_PAGE_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/cancel/'
+ECOMMERCE_CYBERSOURCE_RECEIPT_PAGE_PATH: '/commerce/checkout/receipt/'
+ECOMMERCE_CYBERSOURCE_CANCEL_PAGE_PATH: '/commerce/checkout/cancel/'
 ECOMMERCE_CYBERSOURCE_SOAP_API_URL: 'https://set-me-please'
 ECOMMERCE_OSCAR_FROM_EMAIL: 'oscar@example.com'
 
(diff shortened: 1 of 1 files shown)


douglashall:
@clintonb

clintonb:
Only new settings should be added. Deletions should wait until after a successful release to avoid the need to maintain a feature branch.

douglashall:
Not sure I follow "avoid the need to maintain a feature branch".

clintonb:
If you merge this at any point before you merge edx/ecommerce#543, we cannot deploy Otto. Your changes need to be backwards-compatible to avoid bottlenecking other deployments.

douglashall:
Ha! But that was the whole point of the original solution using `rewrite_lms_url_for_current_site` which would have required no changes to this repo in the interim. In any case, here we are, path settings have been added.

clintonb:
:+1: (pending approval from @edx/devops)

jibsheet:
@douglashall do you mind rewording the commit, since you haven't removed anything in this branch now (I assume you rebased following the conversation with Clinton).

Also, I assume the point is that the variables above the new ones will get dropped later?

douglashall:
@jibsheet Sure thing.

maxrothman:
Out of curiosity, what's the motivation for making this configurable?

douglashall:
@maxrothman I believe the thought is to not have these paths hard-coded in the ecommerce repo. I could see much of this payment processor configuration moved to the database at some point, but I think that is out of scope for the current work.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -78,6 +80,9 @@ ECOMMERCE_PAYPAL_CLIENT_SECRET: 'SET-ME-PLEASE'
 ECOMMERCE_PAYPAL_RECEIPT_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/receipt/'
 ECOMMERCE_PAYPAL_CANCEL_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/cancel/'
 ECOMMERCE_PAYPAL_ERROR_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/error/'
+ECOMMERCE_PAYPAL_RECEIPT_PATH: '/commerce/checkout/receipt/'
```
feanil, line 83: Are the values of these settings expected to change?  Is paypal ever going to have a different error page path from cybersource?  If not, I think we should consolidate the top-level vars along the lines of `ECOMMERCE_RECEIPT_PAGE_PATH` etc.  Also how likely are these configurations to change at all?

douglashall, line 83: Agreed that we should revisit these settings at a later date. We'll need things to remain status quo to ease the rollout of the changes we are making to ecommerce. I have created https://openedx.atlassian.net/browse/WL-342 to track this follow on work.

maxrothman:
My reason for asking is that generally things that the open-source community wouldn't want to configure are kept in the app's settings files. It doesn't seem to make a difference either way from a configurability perspective, since changing the path in either place requires a deploy.

playbooks/roles/ecommerce/defaults/main.yml
feanil, line 83: @douglashall thought on the other question in my comment?  Why not have the top level settings be shared instead of having the same setting repeated with a different name per payment provider?

douglashall, line 83: @feanil Yes, I think that makes sense.

douglashall:
@maxrothman Good point. I don't actually need to make this change here. I can make it in one of the ecommerce settings files. Thanks!

douglashall:
No longer needed.

PR #2811, opened 2016-02-25
Install optional credentials requirements

rlucioni:
@maxrothman please review. The Make target used here is removed in https://github.com/edx/credentials/pull/69.

@zubair-arbi @tasawernawaz @jimabramson FYI.

PR #2812, opened 2016-02-25
Install optional credentials requirements

rlucioni:
Supersedes https://github.com/edx/configuration/pull/2811.

@maxrothman please review. The Make target used here is removed in https://github.com/edx/credentials/pull/69.

@zubair-arbi @tasawernawaz @jimabramson FYI.


playbooks/roles/credentials/tasks/main.yml
```diff
@@ -30,25 +30,13 @@
     - install
     - install:configuration
 
-- name: build virtualenv
```
zubair-arbi, line 33: @rlucioni @maxrothman will this task run from edx local role?

PR #2814, opened 2016-02-26
Install optional credentials requirements

rlucioni:
@maxrothman @zubair-arbi

### DIFF
--- a/playbooks/roles/credentials/tasks/main.yml
+++ b/playbooks/roles/credentials/tasks/main.yml
@@ -30,25 +30,14 @@
     - install
     - install:configuration
 
-- name: build virtualenv
-  command: "virtualenv {{ credentials_venv_dir }}"
-  args:
-    creates: "{{ credentials_venv_dir }}/bin/pip"
-  sudo_user: "{{ credentials_user }}"
-  environment: "{{ credentials_environment }}"
-  tags:
-    - install
-    - install:app-requirements
-
 - name: install application requirements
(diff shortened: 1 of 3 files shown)


playbooks/roles/credentials/tasks/main.yml
```diff
@@ -30,25 +30,13 @@
...
-    - install
-    - install:app-requirements
-
 - name: install application requirements
-  command: make prod-requirements
-  args:
-    chdir: "{{ credentials_code_dir }}"
+  pip:
+    requirements: "{{ credentials_requirements_base }}/{{ item }}"
+    virtualenv: "{{ credentials_venv_dir }}"
```
maxrothman, line 37: I think this requires the virtualenv already exist, which means you shouldn't remove the above task.

rlucioni, line 37: Ansible should initialize the virtualenv if necessary. See http://docs.ansible.com/ansible/pip_module.html and how we do this for [programs](https://github.com/edx/configuration/blob/master/playbooks/roles/programs/tasks/main.yml#L30-L36).

zubair-arbi:
Changes look good :+1:

playbooks/roles/credentials/tasks/main.yml
rlucioni, line 37: We verified this offline. The virtualenv is indeed created if it's missing.

maxrothman:
:+1:

PR #2816, opened 2016-02-27
Improvements for s3cmd in the "aws" role

smarnach:
Two small improvements for the `aws` role:
1. Install s3cmd via pip instead of downloading and installing a tarball.  The tarball was outdated and didn't support the newer AWS4-HMAC-SHA256 authentication mechanism, so it doesn't work at all for S3 buckets in newer regions like Frankfurt.  Moreover, installing via pip is much simpler.
2. Allow to provide explicit IAM credentials for S3 log syncing.  Using EC2 instance role is cumbersome and error-prone in our (rather manual) workflow.

### DIFF
--- a/playbooks/roles/aws/defaults/main.yml
+++ b/playbooks/roles/aws/defaults/main.yml
@@ -50,7 +50,7 @@ aws_s3_sync_script: "{{ aws_dirs.home.path }}/send-logs-to-s3"
 aws_s3_logfile: "{{ aws_dirs.logs.path }}/s3-log-sync.log"
 aws_region: "us-east-1"
 # default path to the aws binary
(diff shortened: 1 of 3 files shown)


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=2816

feanil:
Generally seems reasonable to me.  :+1:   @edx/devops second review.

smarnach:
@feanil Thanks for the quick review!

The test failure seems to be unrelated to this change.  The actual error is in [this log line](https://travis-ci.org/edx/configuration/jobs/112315786#L2019).

maxrothman:
:+1:

mtyaka:
:+1:

PR #2819, opened 2016-02-29
Be able to run jailed code in the CMS as well.

feanil:
https://openedx.atlassian.net/projects/DEVOPS/queues/16/DEVOPS-3753

@edx/devops 

This will make sure that both the LMS and CMS have the same codejail settings.


e0d:
:+1:

PR #2821, opened 2016-03-01
Improved paver autocompletion.

jcdyer:
Hackathon project to improve auto-completion in paver

@aleffert @benpatterson @edx/devops 

To try out the changes manually, simply source the changed file (. path/to/paver_autocomplete) and then play around with paver commands.  It should
- autocomplete subcommand names
- autocomplete subcommand options after a -
- autocomplete path names after subcommands (This is the big thing I've been missing when running tests)

### DIFF
--- a/playbooks/roles/local_dev/templates/paver_autocomplete
+++ b/playbooks/roles/local_dev/templates/paver_autocomplete
@@ -1,20 +1,139 @@
 # Courtesy of Gregory Nicholas
 
+_subcommand_opts()
+{
+    local awkfile command cur usage
+    command=$1
+    cur=${COMP_WORDS[COMP_CWORD]}
+    awkfile=/tmp/paver-option-awkscript-$$.awk
+    echo '
+BEGIN {
+    opts = "";
+}
+
+{
+    for (i = 1; i <= NF; i = i + 1) {
+        # Match short options (-a, -S, -3) 
+        # or long options (--long-option, --another_option)
+        # in output from paver help [subcommand] 
+        if ($i ~ /^(-[A-Za-z0-9]|--[A-Za-z][A-Za-z0-9_-]*)/) {
+            opt = $i;
+            # remove trailing , and = characters.
+            match(opt, "[,=]");  
+            if (RSTART > 0) {
+                opt = substr(opt, 0, RSTART);
+            }
+            opts = opts " " opt;
+        }
+    }
+}
+
+END {
+    print opts
+}' > $awkfile
+
+    usage=`paver help $command`
+    options=`echo "$usage"|awk -f $awkfile`
+
+    COMPREPLY=( $(compgen -W "$options" -- "$cur") )
+}
+
+
 _paver()
 {
-    local cur
+    local cur prev
     COMPREPLY=()
     # Variable to hold the current word
     cur="${COMP_WORDS[COMP_CWORD]}"
+    prev="${COMP_WORDS[COMP_CWORD - 1]}"
 
     # Build a list of the available tasks from: `paver --help --quiet`
     local cmds=$(paver -hq | awk '/^  ([a-zA-Z][a-zA-Z0-9_]+)/ {print $1}')
 
+    subcmd="${COMP_WORDS[1]}"
     # Generate possible matches and store them in the
     # array variable COMPREPLY
-    COMPREPLY=($(compgen -W "${cmds}" $cur))
+    
+    if [[ -n $subcmd ]]
+    then
+        case $subcmd in
+            test_system)
+               
+                _test_system_args
+                if [[ -n $COMPREPLY ]]
+                then
+                    return 0
+                fi
+                ;;
+            test_bokchoy)
+                _test_bokchoy_args
+                if [[ -n $COMPREPLY ]]
+                then
+                    return 0
+                fi
+                ;;
+            *)
+                ;;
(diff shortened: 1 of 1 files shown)


playbooks/roles/local_dev/templates/paver_autocomplete
```diff
@@ -1,20 +1,110 @@
...
+            opts = opts " " opt;
+        }
+    }
+}
+
+END {
+    print opts
+}' > /tmp/paver_parse_opts.awk
+    usage=`paver help $command`
+    options=`echo "$usage"| awk -f /tmp/paver_parse_opts.awk`
```
jcdyer: I tried to do this using a here document (<< EOF), but bash kept finding reasons to complain.  Also, if I didn't put `paver help $command` into a temporary variable, I got broken pipe issues.  Would love some ideas on how to do this more cleanly.

playbooks/roles/local_dev/templates/paver_autocomplete
```diff
@@ -1,20 +1,110 @@
...
+    local awkscript command cur usage
+    command=$1
+    cur=${COMP_WORDS[COMP_CWORD]}
+    echo 'BEGIN {
+    opts = "";
+}
+
+{
+    for (i = 1; i <= NF; i = i + 1) {
+        if ($i ~ /^(-[A-Za-z0-9]|--[A-Za-z][A-Za-z0-9_-]*)/) {
```
benpatterson, line 19: Could you comment an example here?

If I'm following this right, the regex might miss things like `paver test_bokchoy` because of the underscore in the first regex phrase? Or perhaps I'm confusing which subcommand (1st or 2nd) we're parsing at this point.

jcdyer, line 19: This is options, not subcommands.  Don't miss the initial - after the left paren.  This should match -s or --long_opt-i-on.  I will add comments.

benpatterson, line 19: Gotcha. Yeah I sorted it out later when I got to `_paver`. But comments are appreciated...solid.

playbooks/roles/local_dev/templates/paver_autocomplete
```diff
@@ -1,20 +1,106 @@
...
+                opt = substr(opt, 0, RSTART);
+            }
+            opts = opts " " opt;
+        }
+    }
+}
+
+END {
+    print opts
+}' > /tmp/paver_parse_opts.awk
```
aleffert: Should this generate a safe/unique temp path? Seems like a possible race condition?

aleffert: Alternately, keep it in a variable?

benpatterson:
@jcdyer it looks pretty good to me. I love the path completion!  A trick here is going to be `paver test_bokchoy`. When that uses the `-t` argument, it expects the user to put in a path under `common/test/acceptance`. 

So, for example, if my test is in `common/test/acceptance/studio/test_studio.py`, paver would need: `paver test_bokchoy -t studio/test_studio.py`. If paver gets the full path, it's going to blindly append the arg into something like this for notestests: `nosetests common/test/acceptance/common/test/acceptance/studio/test_studio.py`.

Any ideas?

jcdyer:
@benpatterson We might get some traction building up a list of possible filenames using `find common/test/acceptance -type f` and then sedding away the common prefix.  I'll prototype something.

jcdyer:
@benpatterson I think I've got a working but not perfect solution for your issue.  From the docs it looks like the option you are looking for is -d/--test_dir, not -t/--test_spec, but let me know if that's wrong.  Try it out.  It seems to expand the right way, but provides too much output.  Cleaning that up might be painful.  I'm not going to try to do it in this PR.  More cleanup can come later.

benpatterson:
Ah nice. I don't have time at the moment to give it another dry run, but the changes get a :+1: from me

aleffert:
I don't know if my bash is really up to snuff, but this looks plausible :+1:

mushtaqak:
@jcdyer @benpatterson I was building a new installation of devstack yesterday, after this PR was merged, and I kept getting errors due to this code change, I found the same error on other workstations too. You may want to take a look at it.

`==> default: TASK: [local_dev | add paver autocomplete] ************************************ 
==> default: fatal: [localhost] => {'msg': 'AnsibleError: file: /edx/app/edx_ansible/edx_ansible/playbooks/roles/local_dev/templates/paver_autocomplete, line number: 79, error: Missing end of comment tag', 'failed': True}
==> default: fatal: [localhost] => {'msg': 'One or more items failed.', 'failed': True, 'changed': False, 'results': [{'msg': 'AnsibleError: file: /edx/app/edx_ansible/edx_ansible/playbooks/roles/local_dev/templates/paver_autocomplete, line number: 79, error: Missing end of comment tag', 'failed': True}]}`

I fixed the issue by replacing the `playbooks/roles/local_dev/templates/paver_autocomplete` file with previous version.

jcdyer:
Thanks.  I'll take a look at it.

jibsheet:
Please note, I've opened DEVOPS-3836 since it breaks devstack provisioning, and you can see my referenced proposed fix.

mushtaqak:
@jibsheet that is great, thanks for the quick proposal :)

PR #2822, opened 2016-03-01
Previously, this was hammered on in aws.py for all DBs

jibsheet:
Instead, turn it on for the databases where it makes sense.

https://github.com/edx/edx-platform/pull/11722


jibsheet:
@edx/devops See also the referenced platform PR for the why.

feanil:
What is the reason for not having atomic requests on the `read_replica`?

jibsheet:
@feanil Is there a win there?  I figured it's read only, so unlike the default db where someone might take a transaction and update while someone is reading, all the read-replica access should be reads and are outside the transaction boundaries anyway.

feanil:
:+1:

jibsheet:
This was cherry picked into https://github.com/edx/configuration/pull/2815 and merged

PR #2823, opened 2016-03-02
Use utf8mb4 encoding for mysql databases

omarkhan:
Allows full 4-byte UTF-8 characters (e.g. 😜) to be written to mysql.

### DIFF
--- a/playbooks/roles/edxlocal/tasks/main.yml
+++ b/playbooks/roles/edxlocal/tasks/main.yml
@@ -6,13 +6,23 @@
   apt: pkg={{ item }} install_recommends=yes state=present
   with_items: edxlocal_debian_pkgs
 
-# TODO: Add a test to make sure mysql is running.
+- name: start mysql
+  service:
+    name: mysql
+    state: started
+
+# Needed to create djcelery tables with utf8mb4 encoding:
+# https://github.com/celery/django-celery/issues/259
+- name: set innodb_large_prefix
+  mysql_variables:
+    variable: innodb_large_prefix
+    value: 1
 
 - name: create databases
   mysql_db:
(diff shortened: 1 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @omarkhan! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=2823&repo=edx%2Fconfiguration

jibsheet:
@omarkhan Thanks for the utf8mb4 change, it looks great, however mixing it in with a mysql 5.7 upgrade  is a much larger change.

Can you split these apart?  We're still running RDS 5.6 so we will want to continue to do so, so will need to figure out a 5.7 change.

brianhw:
Also just wondering about the effect (if any) on indexing.  It used to be that we had varchar(255) values because these could be indexed within the 768-byte limit that Mysql supported.  If Mysql has the same limit, and we move from 3-byte to 4-byte characters, then not all 255-char values will be indexed.

smarnach:
@jibsheet @brianhw We are still reviewing and testing this change on our side.  Currently, it isn't easily possible to make utf8mb4 work without upgrading to version 5.7 due to the indexing issue Brian mentions, so we might drop the plan.

I'll close this PR for now.  We might reopen it in case we figure out a better way to make this work.

PR #2826, opened 2016-03-02
credentials: fix tags on requirements install task

jimabramson:
@maxrothman @rlucioni look correct?

### DIFF
--- a/playbooks/roles/credentials/tasks/main.yml
+++ b/playbooks/roles/credentials/tasks/main.yml
@@ -38,6 +38,9 @@
     state: present
   sudo_user: "{{ credentials_user }}"
   with_items: "{{ credentials_requirements }}"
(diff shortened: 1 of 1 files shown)


rlucioni:
:+1: 

Thanks for catching this @jimabramson. I didn't realize those tags were important here. For those wondering, we invoke `ansible-playbook` with the `install:app-requirements` tag [here](https://github.com/edx/configuration/blob/master/playbooks/roles/credentials/templates/edx/app/credentials/devstack.sh.j2#L15). Older IDAs like programs and ecommerce don't use these tags.

jibsheet:
:+1:  once Travis completes (Since that's where we're seeing the error).  Thanks Jim.

feanil:
:+1:

PR #2832, opened 2016-03-03
Define the CSMH database so it is created / migrate users defined

jibsheet:
This is safe to merge since it creates databases on sandboxes built from scratch and devstacks, but they're unused until we also merge code to enable the second database / define them in lms.auth.json:DATABASES


feanil:
Looks like you need to rebase. :+1:

jibsheet:
Yeah, I cherry picked yours in so I could build a sandbox and work around rabbit, will rebase after builds finish.

jibsheet:
@fredsmith This is the change needed to create DBs on sandboxes / devstacks.
We'll need an additional override in each env to turn on the feature flag for the new base images.
I'll be cutting a sandbox image tomorrow.

PR #2833, opened 2016-03-03
Workaround for debian squeeze deprecation and Chrome.

benpatterson:
The older chrome repositories have been taken down; however, we
use a saved copy of the debian package. After installing that package,
ensure it is not referenced by the target machine after Chrome
is installed.

### DIFF
--- a/playbooks/roles/browsers/tasks/main.yml
+++ b/playbooks/roles/browsers/tasks/main.yml
@@ -15,6 +15,12 @@
   when: download_deb.changed
   with_items: browser_s3_deb_pkgs
 
+# Because the source location has been deprecated, we need to
+# ensure it does not interfere with subsequent apt commands
+# TODO: use the file module when using ansible 2.x
+- name: archive google chrome debian source list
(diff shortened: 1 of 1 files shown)


benpatterson:
@edx/devops @jzoldak @nedbat

e0d:
So the issue here is that the chrome packages provides `/etc/apt/sources.list.d/google-chrome.list`  which seems like a strange side-effect to me of manually installing a package.

Removing or moving the file seems like a decent work-around.

Another option would be to update `/etc/apt/sources.list.d/google-chrome.list` to:

```
deb [arch=amd64] http://dl.google.com/linux/chrome/deb/ stable main
```

benpatterson:
@e0d FWIW my thinking was that the package is old enough such that we'd want any update or re-install to be more genuine. That said, I don't have strong feelings about it. Am happy to take your suggested route if you think that's the right move.

I'm going to merge this in order to fix the urgency of the issue. But can submit a follow-up PR, no problem.

PR #2753, opened 2016-02-03
Fix analyticstack

itsjeyd:
cf. [OLIVE-5](https://openedx.atlassian.net/browse/OLIVE-5)

**Test instructions**

Testing this PR involves two steps:
1. Creating a base box using `vagrant/base/analyticstack/Vagrantfile`.
2. Setting up an instance of the Analytics Devstack based on the [official instructions](https://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/devstack/analytics_devstack.html) using the newly created box file.

Detailed steps:
- Create the base box by doing:
  
  ``` sh
  mkdir analytics
  cd analytics
  git clone https://github.com/open-craft/configuration.git
  cd configuration
  git checkout tim/fix-analyticstack
  virtualenv -p /usr/bin/python2 venv
  source venv/bin/activate
  pip install -r pre-requirements.txt
  pip install -r requirements.txt
  cd vagrant/base/analyticstack
  export OPENEDX_RELEASE=named-release/dogwood
  vagrant up
  ```
- Package the base box via:
  
  ``` sh
  vagrant package
  ```
- Go back to parent directory of `configuration` repo and perform steps from "Install the Software Prerequisites" section of [Installing the Open edX Analytics Developer Stack](https://edx.readthedocs.org/projects/edx-installing-configuring-and-running/en/latest/devstack/analytics_devstack.html):
  
  ``` sh
  deactivate
  cd ../../../..
  git clone https://github.com/edx/edx-analytics-pipeline
  cd edx-analytics-pipeline
  virtualenv -p /usr/bin/python2 venv
  source venv/bin/activate
  make bootstrap
  ```
- Create directory for Analytics Devstack and `cd` into it:
  
  ``` sh
  deactivate
  cd ..
  mkdir analyticstack
  cd analyticstack
  ```
- Copy `vagrant/release/analyticstack/Vagrantfile` into this directory:
  
  ``` sh
  cp ../configuration/vagrant/release/analyticstack/Vagrantfile ./
  ```
- Edit Vagrantfile to point to local box file:
  
  ``` ruby
      Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
        # ...
        config.vm.box     = "package"
        config.vm.box_url = "file:///path/to/analytics/configuration/vagrant/base/analyticstack/package.box"
        # ...
  ```
- Use Vagrantfile to bring up a new instance of the Analytics Devstack:
  
  ``` sh
  vagrant up
  ```
[...]

PR #2834, opened 2016-03-03
Ensure login shell is set correctly for Analytics users

itsjeyd:
This PR follows up on https://github.com/edx/configuration/pull/2753 and makes sure that the default login shell for the `analytics_api` and `insights` users is set to `/bin/bash` during provisioning.

**Test instructions**
1. Create a new base box for analyticstack by doing:
   
   ``` sh
   mkdir analytics
   cd analytics
   git clone https://github.com/open-craft/configuration.git
   cd configuration
   git checkout master-b3a7dc0
   virtualenv -p /usr/bin/python2 venv
   source venv/bin/activate
   pip install -r pre-requirements.txt
   pip install -r requirements.txt
   cd vagrant/base/analyticstack
   export OPENEDX_RELEASE=named-release/dogwood
   vagrant up
   ```
2. SSH into the VM and check the default shell for `analytics_api` and `insights` users:
   
   ``` sh
   vagrant ssh
   getent passwd analytics_api | cut -d: -f7
   getent passwd insights | cut -d: -f7
   ```
   
   The last two commands should output `/bin/bash`.

### DIFF
--- a/playbooks/vagrant-analytics.yml
+++ b/playbooks/vagrant-analytics.yml
@@ -26,9 +26,9 @@
     - elasticsearch
     - forum
     - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
+    - analytics_api
+    - insights
     - local_dev
     - demo
-    - analytics_api
     - analytics_pipeline
-    - insights
     - oauth_client_setup


openedx-webhooks:
Thanks for the pull request, @itsjeyd! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=2834

playbooks/roles/analytics_api/meta/main.yml
```diff
@@ -31,3 +31,11 @@ dependencies:
     edx_service_packages:
       debian: "{{ analytics_api_debian_pkgs }}"
       redhat: "{{ analytics_api_redhat_pkgs }}"
+  - role: local_dev
+    localdev_accounts:
```
feanil: Local Dev should never be a dependency of a role.  If we add it here it will also be added in production as well and this would be incorrect.  It should instead be added to the relevant playbooks.

playbooks/roles/insights/meta/main.yml
```diff
@@ -9,7 +9,7 @@
 #
 ##
 # Role includes for role insights
-# 
+#
```
mulby: I would revert the changes to this file

mulby:
one nit

:+1:

playbooks/roles/analytics_api/meta/main.yml
itsjeyd: @feanil Thanks, I changed the approach based on your comment (and also addressed Gabe's nit below). Could you have another look please?

feanil:
:+1:

e-kolpakov:
@itsjeyd :+1:

PR #2836, opened 2016-03-04
Allow a single user access to multiple databases

jibsheet:
This avoids a class of confusing problems where a user might be given
access to two DBs and then when starting the app, fail because they can
only access one DB.


feanil:
:+1:

PR #2837, opened 2016-03-07
Adding go-agent Docker file

macdiesel:
Docker file for go-agents. (Includes python 2.7.10 and PIP)

### DIFF
--- a/docker/build/go-agent/Dockerfile
+++ b/docker/build/go-agent/Dockerfile
@@ -1,0 +1,18 @@
+# Build using: docker build -f Dockerfile.gocd-agent -t gocd-agent .
+# FROM edxops/precise-common:latest
+FROM gocd/gocd-agent:16.2.1
+
+LABEL version="0.01" \
+      description="This custom go-agent docker file installs additional requirements for the edx pipeline"
+
+
+RUN apt-get update && apt-get install -y -q \
+    python \
+    python-dev \
+    python-distribute \
+    python-pip
+
+# TODO: repalce this with a pip install command so we can version this properly
+RUN git clone -b bbeggs/cli https://github.com/edx/tubular.git /opt/tubular
+RUN pip install -r /opt/tubular/requirements.txt
+RUN cd /opt/tubular;python setup.py install
--- a/docker.mk
+++ b/docker.mk
@@ -52,7 +52,7 @@ $(docker_push)%: $(docker_pkg)%
 
 .build/%/Dockerfile.d: docker/build/%/Dockerfile Makefile
 	@mkdir -p .build/$*
-	$(eval FROM=$(shell grep "FROM" $< | sed --regexp-extended "s/FROM //" | sed --regexp-extended "s/:/@/g"))
+	$(eval FROM=$(shell grep "^\s*FROM" $< | sed --regexp-extended "s/FROM //" | sed --regexp-extended "s/:/@/g"))
 	$(eval EDXOPS_FROM=$(shell echo "$(FROM)" | sed --regexp-extended "s#edxops/([^@]+)(@.*)?#\1#"))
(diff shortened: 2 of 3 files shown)


macdiesel:
@feanil updated the from and added some labels. PTAL.

macdiesel:
@doctoryes FYI

docker/build/go-agent/gocd-agent/go-agent-start.sh
```diff
@@ -0,0 +1,19 @@
...
+echo -e "${COLOR_START}Starting Go Agent to connect to server $GO_SERVER ...${COLOR_END}"
+sed -i -e 's/GO_SERVER=.*/GO_SERVER='$GO_SERVER'/' /etc/default/go-agent
+
+mkdir -p /var/lib/go-agent/config
+/bin/rm -f /var/lib/go-agent/config/autoregister.properties
+
+AGENT_KEY="${AGENT_KEY:-123456789abcdef}"
+echo "agent.auto.register.key=$AGENT_KEY" >/var/lib/go-agent/config/autoregister.properties
+if [ -n "$AGENT_RESOURCES" ]; then echo "agent.auto.register.resources=$AGENT_RESOURCES" >>/var/lib/go-agent/config/autoregister.properties; fi
+if [ -n "$AGENT_ENVIRONMENTS" ]; then echo "agent.auto.register.environments=$AGENT_ENVIRONMENTS" >>/var/lib/go-agent/config/autoregister.properties; fi
```
doctoryes: Good ole bash. How about splitting these two lines? Like:

```
if [ -n "$AGENT_RESOURCES" ]; then 
  echo "agent.auto.register.resources=$AGENT_RESOURCES" >>/var/lib/go-agent/config/autoregister.properties;
fi
if [ -n "$AGENT_ENVIRONMENTS" ]; then
  echo "agent.auto.register.environments=$AGENT_ENVIRONMENTS" >>/var/lib/go-agent/config/autoregister.properties;
fi
```

?

macdiesel:
jenkins run all

docker/build/go-agent/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN apt-get update && apt-get install -y -q \
+    python \
+    python-dev \
+    python-distribute \
+    python-pip
+
+# TODO: repalce this with a pip install command so we can version this properly
+RUN git clone -b bbeggs/cli https://github.com/edx/tubular.git /opt/tubular
+RUN pip install -r /opt/tubular/requirements.txt
+RUN cd /opt/tubular;python setup.py install
```
doctoryes, line 18: Shouldn't this happen inside a created virtualenv?
Also, couldn't you create a one line `requirements.txt` pointing to tubular in git - and then `pip install -r` it?

macdiesel, line 18: After some research and talking to @feanil it seems that most people are not using virtualenv inside of docker containers. The configuration is very minimal for the linux machine and we could tag these in such a way so that only these agents run the deploy scripts to avoid possible conflicts with python libraries. 

> Also, couldn't you create a one line requirements.txt pointing to tubular in git - and then pip install -r it?

I could but where would that file live? My hope here is that we would replace this with a `pip install edx/tubular` one liner once we aren't iterating quite so rapidly.

doctoryes:
:+1:

docker/build/go-agent/Dockerfile
feanil, line 18: Is there a reason not to push tubular to PyPI and just pip install tubular?  I'm fine with that being a second order change if you think it will take too long.

macdiesel, line 18: That is the plan, I was going to hold until tubular was more stable as I was working off a branch and didn't want to pollute up the repository while we were testing.

feanil:
:+1:

feanil:
:+1:

PR #2838, opened 2016-03-08
Increasing No of gunicorn workers

muzaffaryousaf:


muzaffaryousaf:
@fredsmith Do we need to merge this PR or should i close ?

muzaffaryousaf:
I'm also not sure if we need to add hashlib or not.

feanil:
This setting should be managed in secure config as it is specific to the sizing of the hardware we use in our deployment.

muzaffaryousaf:
@feanil Thanks for looking into. I suppose that's something Ops can do and maybe @fredsmith already made the changes in secure repo ?

e0d:
The value is 6 in secure config, I think this PR should be closed.

muzaffaryousaf:
Thanks @e0d for looking into.

PR #2840, opened 2016-03-08
credentials: correct AWS_S3_CUSTOM_DOMAIN setting

jimabramson:
@maxrothman @zubair-arbi @rlucioni @clintonb FYI.


clintonb:
:+1:

zubair-arbi:
Great find. @jimabramson I now recall your warnings about custom domain.

> we did see this before with profile images, but it went away when we switched to custom domain
> 
> @max lessons learned from last time about aws custom domain https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L487-L500

:+1:

maxrothman:
:+1: :shipit:

rlucioni:
Nice. :+1:

PR #2841, opened 2016-03-08
Sandbox base AMI built with CSMHE enabled

jibsheet:
OPS-1171

@edx/devops
https://github.com/edx-ops/sandbox-secure/pull/36


arbabnazar:
:+1:

PR #2844, opened 2016-03-08
Adding go-agent docker server role

macdiesel:
@doctoryes @feanil PTAL.

### DIFF
--- a/playbooks/roles/go-agent-docker-server/README.md
+++ b/playbooks/roles/go-agent-docker-server/README.md
@@ -1,0 +1,2 @@
+##In order to use this role you must use a specific set of AMIs
+[This role is for use with the AWS ECS AMIs listed here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html)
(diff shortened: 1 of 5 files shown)


playbooks/roles/go-agent-docker-server/README.md
```diff
@@ -0,0 +1,2 @@
+##In order to use this role you must use a specific set of AMIs
+[This role is for use with the AWS ECS AMIs listed here](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/launch_container_instance.html)
```
doctoryes, line 2: So each agent instance would be launched according to those AWS instructions? And then this playbook is run against that instance? Where does the go-agent get installed? Does it live already inside the Docker container that gets spun up?

macdiesel, line 2: @doctoryes some additional documentation can be found here as well: https://openedx.atlassian.net/wiki/display/ENG/go-agent+Docker+Container+Server+Setup

The workflow would look something like this:
- Use Jenkins to build and deploy this machine (this could possibly be done later by the pipeline or Terraform).
- This server would then be registered as a docker host for a particular "ECS Cluster"
- Terraform would have a configuration that would setup the ECS clusters and run the correct docker containers and their configuration on this host.

>  Where does the go-agent get installed?
> Does it live already inside the Docker container that gets spun up?

The go-agent would run on the docker container(s) that this docker host runs. You can find that work in this pr: #2837

doctoryes:
:+1:

feanil:
:+1:

PR #2845, opened 2016-03-09
New base box for devstacks that includes fixes for the chrome depreca…

jibsheet:
…tion

### DIFF
--- a/vagrant/release/devstack/Vagrantfile
+++ b/vagrant/release/devstack/Vagrantfile
@@ -76,7 +76,7 @@ openedx_releases = {
   # },
 }
 openedx_releases.default = {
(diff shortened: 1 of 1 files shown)


arbabnazar:
:+1:

nedbat:
Any chance we can standardize on either "2016-03-09" or "20160309"?  I see both forms used for devops-built boxes.

jibsheet:
@nedbat I'll check what the job does and we should just standardize to that, I thought I did that, but maybe not (since I had to build this locally).

jibsheet:
Script uses dashes, I like dashes, I've renamed my file and am running a provision locally then will merge.

PR #2849, opened 2016-03-10
credentials: add CORS header for static in nginx

jimabramson:
@maxrothman I think it can be this simple.  this is roughly how it works for edxapp, but not as selective.  

cf. https://github.com/edx/configuration/blob/master/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/static-files.j2#L18-L21

cc: @edx/devops


maxrothman:
A few notes:
- We try to have defaults "just work" if you don't pass any overrides. Since `CREDENTIALS_DOMAIN` and `CREDENTIALS_URL_ROOT` have the same info in them, I think the domain part of `CREDENTIALS_URL_ROOT` should be moved inside of `CREDENTIALS_DOMAIN` and `CREDENTIALS_URL_ROOT` should reference `CREDENTIALS_DOMAIN`.
- If the above is done, then it seems like `CREDENTIALS_URL_ROOT` should be switched to a lowercase var since it is no longer intended for modification. It's only referenced in a small number of places, so this should be fairly painless.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -46,7 +46,7 @@ CREDENTIALS_CACHES:
     LOCATION: '{{ CREDENTIALS_MEMCACHE }}'
 
 CREDENTIALS_DJANGO_SETTINGS_MODULE: "credentials.settings.production"
-CREDENTIALS_URL_ROOT: 'http://credentials:18150'
+CREDENTIALS_DOMAIN: 'credentials.example.com'
```
e0d: I think this will break docker builds, what the purpose of it?

e0d:
Generally looks OK, given the constraints.  :+1:   Assuming you have tested in a sandbox.

jibsheet:
:+1: once tests pass

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -220,6 +223,8 @@ credentials_requirements:
   - production.txt
   - optional.txt
 
+credentials_url_root: 'http://{{ CREDENTIALS_DOMAIN }}:18150'
```
jimabramson: is it correct that this var will always have :18150 at the end?  Definitely seems incorrect for prod.

maxrothman:
Having spoken to @jimabramson it seems that the best approach is to re-uppercase `CREDENTIALS_URL_ROOT` and override it for sandboxes in `ansible_provision.sh`. Basically we need one version for devstack and one version for everything else.

fredsmith:
:+1:

PR #2851, opened 2016-03-11
Adding new role to deploy AppNeta (Work in Progresss)

bilalahmad99:
DO NOT MERGE

### DIFF
--- a/playbooks/roles/appneta/defaults/main.yml
+++ b/playbooks/roles/appneta/defaults/main.yml
@@ -1,0 +1,31 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role appneta
+# 
+
+#
+# vars are namespaced with the module name.
+#
+appneta_role_name: appneta
+USE_STATIC_CONFIG: false
+APT_SERVER: "apt.appneta.com"
+CONFIG_SERVER: "config.tv.appneta.com"
+APT_SERVER_STATIC: "apt-static.appneta.com"
+CONFIG_SERVER_STATIC: "config-static.appneta.com"
+SKIP_CERTS: true
+
+#
+# OS packages
+#
+
+appneta_debian_pkgs: []
+
+appneta_redhat_pkgs: []
--- a/playbooks/roles/appneta/defaults/main.yml
+++ b/playbooks/roles/appneta/defaults/main.yml
@@ -25,6 +25,13 @@ SKIP_CERTS: true
 #
 # OS packages
 #
+nginx_extra_pkgs:
+  - nginx-extras
+  - nginx-light
+
+appneta_lib_pkgs:
+  - liboboe0
+  - liboboe-dev
 
 appneta_debian_pkgs: []
 
--- a/playbooks/roles/appneta/tasks/main.yml
+++ b/playbooks/roles/appneta/tasks/main.yml
@@ -1,0 +1,140 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Tasks for role appneta
+# 
+# Overview:
+# 
+#
+# Dependencies:
+#
+# 
+# Example play:
+#
+
+# write all configs related to static IP mechanism
+- name: Write static config in case of static IPs
+  lineinfile:
+    dest: /etc/default/tracelyzer
+    regexp: ^STATIC_IP=
+    line: "STATIC_IP=YES"
+    state: present
+    create: yes
+  when: USE_STATIC_CONFIG
+
+- name: Change config in case of non-static IPs
+  lineinfile:
+    dest: /etc/default/tracelyzer
+    regexp: ^STATIC_IP=
+    line: "STATIC_IP=NO"
+    state: present
+    create: yes
+  when: not USE_STATIC_CONFIG
+
+# install ca certificates if required
+- name: Install certificates
+  apt:
+    name: ca-certificates
+    state: latest
+    force: yes
+  when: not SKIP_CERTS
+
+# Remove sources file if exists
+- name: Removing pre-exiting sources.list
+  file: path=/etc/apt/sources.list.d/appneta.list state=absent
+
+# Create appneta sources.list
+- name: Create appneta sources.list
+  lineinfile:
+    dest: /etc/apt/sources.list.d/appneta.list
+    line: "deb http://{{ APT_SERVER }}/{{ ACCESS_KEY }} {{ ansible_distribution_release }} main"
+    state: present
+    create: yes
+
+## Downloading AppNeta package signature public key
+- name: Adding AppNeta package signature public key to system list
+  file: path=/etc/apt/sources.list.d/appneta-apt-key.pub* state=absent force=yes
+  #get_url: url=https://{{ APT_SERVER }}/appneta-apt-key.pub dest=/etc/apt/sources.list.d/appneta-apt-key.pub
+  apt_key: id=appneta-apt-key.pub url=https://{{ APT_SERVER }}/appneta-apt-key.pub state=present
+
+# Update repositories
+- name: Running apt-update
+  apt:
+    update_cache: yes
+
+# Installing common library and development headers (liboboe)
+- name: Install common library and headers
+  apt: name={{item}} state=latest
+  with_items:
+       - liboboe0
+       - liboboe-dev
+
+# Install tracelyzer
+# TODO: need to add the logic for disabling error reporting hook if necessary
+- name: Install tracelyzer
+  apt:
+    name: tracelyzer
+    state: latest
+
+# Install appneta-sequencer
+- name: Install appneta-sequencer
+  apt:
+    name: appneta-sequencer
+    state: latest
+
+# Starting the sequencer.
+- name: Start appneta-sequencer
+  service:
+    name: appneta-sequencer
+    state: started
+
+# Install nginx agent from Appneta repo
+- name: Install nginx agent
+  apt:
+    name: nginx
+    state: latest
+
+# Installing nginx light and nginx extra
+- name: Install nginx light and nginx extras
+  apt: name={{item}} state=latest
+  with_items:
+       - nginx-extras
+       - nginx-light
+
+# Starting the sequencer.
+- name: Restart nginx
+  service:
+    name: nginx
+    state: restarted
+
+# Installing python agents
+- name: Install oboe
+  pip:
+    name: oboe
+    state: latest
+
+# Installing greenlet instrumentaion
+- name: Add greenlet instrumentation
+  pip:
+    name: https://pypi.appneta.com greenlet==0.4.5-oboe1
+    state: latest
+
+# Installing tornado instrumentaion
+- name: Add tornado instrumentation
+  pip:
+    name: http://pypi.appneta.com tornado==2.3-oboe3
+    state: latest
+
+# Install Ruby agents
+- name: Add ruby instrumentaion
+  gem:
+    name: traceview
+    state: latest
--- a/playbooks/roles/appneta/tasks/main.yml
+++ b/playbooks/roles/appneta/tasks/main.yml
@@ -20,7 +20,6 @@
 # Example play:
 #
 
-# write all configs related to static IP mechanism
 - name: Write static config in case of static IPs
   lineinfile:
     dest: /etc/default/tracelyzer
@@ -39,7 +38,6 @@
     create: yes
   when: not USE_STATIC_CONFIG
 
-# install ca certificates if required
 - name: Install certificates
   apt:
     name: ca-certificates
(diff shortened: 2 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @bilalahmad99! I've created [OSPR-1186](https://openedx.atlassian.net/browse/OSPR-1186) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

arbabnazar:
You are mixing all the YAML formatting, please stick with one

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+# install ca certificates if required
+- name: Install certificates
+  apt:
+    name: ca-certificates
+    state: latest
+    force: yes
+  when: not SKIP_CERTS
+
+# Remove sources file if exists
+- name: Removing pre-exiting sources.list
```
arbabnazar: why we need to remove it? is it default in Ubuntu?

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+  file: path=/etc/apt/sources.list.d/appneta.list state=absent
+
+# Create appneta sources.list
+- name: Create appneta sources.list
+  lineinfile:
```
arbabnazar: Please use the `apt_repository` module instead of `lineinfile`

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+    dest: /etc/apt/sources.list.d/appneta.list
+    line: "deb http://{{ APT_SERVER }}/{{ ACCESS_KEY }} {{ ansible_distribution_release }} main"
+    state: present
+    create: yes
+
+## Downloading AppNeta package signature public key
+- name: Adding AppNeta package signature public key to system list
+  file: path=/etc/apt/sources.list.d/appneta-apt-key.pub* state=absent force=yes
```
arbabnazar: use `apt_key` module instead of `file`.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+  #get_url: url=https://{{ APT_SERVER }}/appneta-apt-key.pub dest=/etc/apt/sources.list.d/appneta-apt-key.pub
+  apt_key: id=appneta-apt-key.pub url=https://{{ APT_SERVER }}/appneta-apt-key.pub state=present
```
arbabnazar: This is wrong, role will fail on this task due to wrong formatting

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+- name: Install common library and headers
+  apt: name={{item}} state=latest
+  with_items:
+       - liboboe0
+       - liboboe-dev
+
+# Install tracelyzer
+# TODO: need to add the logic for disabling error reporting hook if necessary
+- name: Install tracelyzer
+  apt:
```
arbabnazar, line 67: combine all `apt` into one for simplicity and readability, instead of using separate for each package(s).

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+
+# Install appneta-sequencer
+- name: Install appneta-sequencer
+  apt:
+    name: appneta-sequencer
+    state: latest
+
+# Starting the sequencer.
+- name: Start appneta-sequencer
+  service:
```
arbabnazar, line 77: use condition to start the service like when the package installs then only start the service

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+# Installing nginx light and nginx extra
+- name: Install nginx light and nginx extras
+  apt: name={{item}} state=latest
+  with_items:
+       - nginx-extras
+       - nginx-light
+
+# Starting the sequencer.
+- name: Restart nginx
+  service:
```
arbabnazar: same here, please put the condition

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+# Installing python agents
+- name: Install oboe
+  pip:
+    name: oboe
+    state: latest
+
+# Installing greenlet instrumentaion
+- name: Add greenlet instrumentation
+  pip:
+    name: https://pypi.appneta.com greenlet==0.4.5-oboe1
```
arbabnazar: did you try this, I think it will not work.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,140 @@
...
+    state: latest
+
+# Installing tornado instrumentaion
+- name: Add tornado instrumentation
+  pip:
+    name: http://pypi.appneta.com tornado==2.3-oboe3
```
arbabnazar: same as above

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: removed. not needed.

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99, line 67: Doing this to stay in line with the bash installer script of appneta. They are installing packages one by one and ensuring a few things which I will add once I get the access_key and do final validation.

playbooks/roles/appneta/tasks/main.yml
bilalahmad99, line 77: This will not cause service lock when playbook is run again as it is doing start only in case it is not started already.

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+    filename: 'appneta'
+
+# Downloading AppNeta package signature public key
+- name: Adding AppNeta package signature public key to system list
+  apt_key:
+    url: https://{{ APT_SERVER }}/appneta-apt-key.pub
+    state: present
+
+# Update repositories
+- name: Running apt-update
```
e0d: Should be able to combine this task with the one below.

nedbat:
I've cleared up the OSPR confusion, and will delete the ticket.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+- name: Change config in case of non-static IPs
+  lineinfile:
+    dest: /etc/default/tracelyzer
+    regexp: ^STATIC_IP=
+    line: "STATIC_IP=NO"
+    state: present
+    create: yes
+  when: not USE_STATIC_CONFIG
+
+# install ca certificates if required
```
arbabnazar: all these comments are unnecessary, because you have addressed it in the name description.

playbooks/roles/appneta/tasks/main.yml
arbabnazar: actually we don't need it, because `apt_repository` has by default `update_cache` enabled.

playbooks/roles/appneta/tasks/main.yml
arbabnazar, line 67: I don't think that we need to follow the bash methodology here, we can use it as reference and convert it to meet ansible standards.

playbooks/roles/appneta/tasks/main.yml
arbabnazar, line 77: still we are avoiding it as much as we can in all other places

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+
+# Install nginx agent from Appneta repo
+- name: Install nginx agent
+  apt:
+    name: nginx
+    state: latest
+
+# Installing nginx light and nginx extra
+- name: Install nginx light and nginx extras
+  apt: pkg={{ item }} state=present
```
arbabnazar, line 87: formatting mixing

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+    state: present
+
+# Update repositories
+- name: Running apt-update
+  apt:
+    update_cache: yes
+
+# Installing common library and development headers (liboboe)
+- name: Install common library and headers
+  apt: pkg={{ item }} state=present
```
arbabnazar, line 61: formatting fixing

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+  with_items:
+       - liboboe0
+       - liboboe-dev
```
arbabnazar: you can add all the packages inside the `appneta_debian_pkgs` list and pass that list instead of running it separately.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+    name: nginx
+    state: latest
+
+# Installing nginx light and nginx extra
+- name: Install nginx light and nginx extras
+  apt: pkg={{ item }} state=present
+  with_items:
+       - nginx-extras
+       - nginx-light
+  notify: restart nginx
```
dkuebric, line 88: Because nginx doesn't allow dynamic module loading, AppNeta ships a replacement package for nginx.  In case of multiple nginx packages (nginx-\* have different sets of modules compiled in), AppNeta provides a replacement package for each one.

By replacement package I mean a package built off the latest OS-native package with a slightly higher version number (version incremented with +tracelytics20 or similar).

So with regard to lines 94-106: I'm not super-familiar with Ansible but it looks like this may end up installing multiple versions of nginx.  Only one, the version being used by the application, needs to be re-installed after the appneta apt repository is added to apt sources list.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+
+# Installing python agents
+- name: Install oboe
```
e0d, line 91: These next options, and nginx above too, should be conditional.  It makes this role complex. but we need to be able to specify which agents to install based on environment.  Most typically, we'll be installing the python agent and nginx, but in some cases nginx will not be on the same server.  We may want to consider folding the nginx installation into a separate role that could be included instead of the existing nginx role.  Another possibility would be to override the PPA and use the existing nginx role.  Would need some experimentation to see whether that would be a tractable option.

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99, line 91: will discuss this with you in the meeting to get more clarity

playbooks/roles/appneta/tasks/main.yml
bilalahmad99, line 88: Agree. I have removed the latest tag and added the state=present. Which will not change the version in case playbook is run again. But still I will check the required versions are present at the right time once I do the final validation. And make amends accordingly.

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99, line 61: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99, line 87: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: removed them

playbooks/roles/appneta/defaults/main.yml
```diff
@@ -0,0 +1,38 @@
...
+#
+nginx_extra_pkgs:
+  - nginx-extras
+  - nginx-light
+
+appneta_lib_pkgs:
+  - liboboe0
+  - liboboe-dev
+
+appneta_debian_pkgs: []
```
arbabnazar, line 36: @bilalahmad99 the best way is to do like this:

``` yaml
appneta_debian_pkgs:
   - nginx-extras
   - nginx-light
   - liboboe0
   - liboboe-dev
```

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+- name: Install appneta-sequencer
+  apt:
+    name: appneta-sequencer
+    state: present
+
+- name: Start appneta-sequencer
+  service:
+    name: appneta-sequencer
+    state: started
+
```
arbabnazar, line 80: Then all in one, you can install like this:

``` yaml
- name: Install the desired packages
  apt:
      name: "{{ item }}"
      state: present
  with_items: appneta_debian_pkgs
```

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,113 @@
...
+    name: nginx
+    state: present
+
+- name: Install nginx light and nginx extras
+  apt: pkg={{ item }} state=present
+  notify: restart nginx
+  with_items: nginx_extra_pkgs
+
+- name: Install oboe
+  pip:
```
arbabnazar, line 92: @e0d I think we need to create virtual environment to install the python packages or is it ok to install the python to the system python?

e0d:
Unfortunately, I think we need to close this for now as appneta has been unresponsive about gives us a trial to verify that the role works.

PR #2854, opened 2016-03-11
copy rather than template this file

jibsheet:
It contains things that look like jinja, but are not.
DEVOPS-3836

Fixes https://github.com/edx/configuration/pull/2821


jibsheet:
@edx/devops @nedbat @jcdyer @mushtaqak

jcdyer:
Cool.  I thought this would be the solution.  Just hadn't had time yet to try it out. :+1:

feanil:
:+1:

PR #2676, opened 2016-01-06
Support externally hosted comprehensive themes

fghaas:
The old Stanford-style themes can be pulled in from an external Git repo; no such provision was previously available for comprehensive themes.
- Introduce two new configuration variables:
  - EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO: Git repo for the   comprehensive theme
  - EDXAPP_COMPREHENSIVE_THEME_VERSION: Git branch, tag, or revision to   check out from the repo
- Introduce a new task, "checkout comprehensive theme", doing the equivalent of the previously existing "checkout theme" task. Check out the theme repo to EDXAPP_COMPREHENSIVE_THEME_DIR. If 
  EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO is unset, continue to assume that EDXAPP_COMPREHENSIVE_THEME_DIR already exists and is populated with a comprehensive theme.
- Rename the "checkout theme" task to "checkout Stanford-style theme".

PR #2856, opened 2016-03-13
This registers over the microsites theming

jibsheet:
This clobbers edxapp_theme_checkout set by the previous task, even when
the task is skipped because EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO is
unset.

edxapp_theme_checkout is used to tag the instance when the edxapp play
is run.

That tag is later used by our build infrastructure to ensure that
theme is consistent between builds.

If we want to also tag the comprehensive theme value, that's fine, but
it should not clobber the value needed by the other type of themes, and
should declare a different tag.

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -118,7 +118,7 @@
   sudo_user: "{{ edxapp_user }}"
   environment:
     GIT_SSH: "{{ edxapp_git_ssh }}"
-  register: edxapp_theme_checkout
+  register: edxapp_comprehensive_theme_checkout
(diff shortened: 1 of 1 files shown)


feanil:
:+1:

jibsheet:
FYI @fghaas since this was introduced by https://github.com/edx/configuration/pull/2676
Also, @mattdrayer @douglashall - this is what broke theming

fghaas:
Got it, sorry about that.

Please also see the discussion in edx/edx-platform#11138 — I tried to get information on microsites from several sources in the process, including Slack, and there seemed to be no-one available to explain how microsites were actually expected to work. Apologies again.

PR #2858, opened 2016-03-14
update credentials cors whitelist values

zubair-arbi:
ECOM-3572

@jimabramson @maxrothman @tobz

### DIFF
--- a/playbooks/roles/credentials/defaults/main.yml
+++ b/playbooks/roles/credentials/defaults/main.yml
@@ -124,8 +124,10 @@ CREDENTIALS_FILE_STORAGE_BACKEND:
   STATICFILES_STORAGE: 'django.contrib.staticfiles.storage.StaticFilesStorage'
   DEFAULT_FILE_STORAGE: 'django.core.files.storage.FileSystemStorage'
 
+# Note: the protocol for CORS whitelist values is necessary for matching the correct origin by nginx
 CREDENTIALS_CORS_WHITELIST:
-  - "{{ CREDENTIALS_DOMAIN }}"
+  - "http://{{ CREDENTIALS_DOMAIN }}"
+  - "https://{{ CREDENTIALS_DOMAIN }}"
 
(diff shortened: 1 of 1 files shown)


jimabramson:
:+1:   @max whaddya think

jimabramson:
@zubair-arbi actually does this need to be adjusted to escape the `/` char?   https://github.com/edx/configuration/blob/master/playbooks/roles/credentials/templates/edx/app/nginx/sites-available/credentials.j2#L20

maxrothman:
:+1:, though I think a nearby comment explaining that the protocol is necessary would be nice.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -125,7 +125,8 @@ CREDENTIALS_FILE_STORAGE_BACKEND:
   DEFAULT_FILE_STORAGE: 'django.core.files.storage.FileSystemStorage'
 
 CREDENTIALS_CORS_WHITELIST:
-  - "{{ CREDENTIALS_DOMAIN }}"
+  - "http://{{ CREDENTIALS_DOMAIN }}"
```
feanil, line 129: Why allow `http`?

jimabramson:
I've heard anecdotally that learners in some regions (CN?) access the site via http.  If that is not supported by actual evidence then I guess we can drop it.

jimabramson:
@feanil ^ that was in response to your http/https question

zubair-arbi:
@maxrothman I think its self explanatory since we have explicitly added the protocol.

maxrothman:
@zubair-arbi the comment is to explain _why_ you added the protocol.

maxrothman:
@jimabramson if you want to support both http and https, just add more entries to the map

zubair-arbi:
@maxrothman I have added a comment. Please review

maxrothman:
:+1:

PR #2864, opened 2016-03-15
use dogwood box for release analyticstack Vagrantfile

mulby:
FYI: @bradenmacdonald @itsjeyd @e-kolpakov 

Note I haven't actually uploaded the file with this name yet. Hoping to use this PR as a forum for discussing the name - open to suggestions.

Reviewers: @jibsheet @brianhw

### DIFF
--- a/vagrant/base/analyticstack/Vagrantfile
+++ b/vagrant/base/analyticstack/Vagrantfile
@@ -75,6 +75,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         certs_version: ENV['OPENEDX_RELEASE'],
         forum_version: ENV['OPENEDX_RELEASE'],
         xqueue_version: ENV['OPENEDX_RELEASE'],
+        ANALYTICS_API_VERSION: ENV['OPENEDX_RELEASE'],
+        INSIGHTS_VERSION: ENV['OPENEDX_RELEASE'],
       }
     end
     if ENV['CONFIGURATION_VERSION']
--- a/vagrant/release/analyticstack/Vagrantfile
+++ b/vagrant/release/analyticstack/Vagrantfile
@@ -24,6 +24,8 @@ if [ -n "$OPENEDX_RELEASE" ]; then
     -e certs_version=$OPENEDX_RELEASE \
     -e forum_version=$OPENEDX_RELEASE \
     -e xqueue_version=$OPENEDX_RELEASE \
+    -e ANALYTICS_API_VERSION=$OPENEDX_RELEASE \
+    -e INSIGHTS_VERSION=$OPENEDX_RELEASE \
   "
   CONFIG_VER=$OPENEDX_RELEASE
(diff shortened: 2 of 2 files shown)


e-kolpakov:
@mulby Well, name is fine by me - it should be unique so that Vagrant would know to update it when something have changed. 

On the other hand, why are named releases uses the same base box? Wouldn't it be a bit more flexible/robust if at least two boxes were provided: one for latest master (updated as we go, when needed) and one for latest named release (updated in release cycle)?

bradenmacdonald:
Is the base box image going to be Dogwood? My impression was that it was a post-Dogwood build, so I second Eugeny's comment.

mulby:
Yeah, I've been thinking about this a bit and agree. I think it's valuable to have the base box pinned to a particular named release since that is assumed to be fairly stable. I'll build with "named-release/dogwood.1" and name the box accordingly. I will copy that box up as both "analyticstack-dogwood-1.box" and "analyticstack-latest.box".

In the future we can upload new copies of builds off of master and overwrite the "analyticstack-latest.box" file. That would ensure people can pull the bleeding edge or the stable version depending on their needs.

How does that sound?

@nedbat - we should probably coordinate on build procedures and naming.

bradenmacdonald:
@mulby That makes sense to me. If we have time, it would be nice to do an updated "latest" box in the near future too, but that's less of a priority, as long as the configuration can migrate from dogwood to master cleanly.

bradenmacdonald:
@mulby Here's an example of configuration updates not applying cleanly, FYI: I tried to re-provision my analyticsstack that's about a month old, and got this message:

```
==> default: TASK: [insights | install bower dependencies] ********************************* 
==> default: failed: [localhost] => {"changed": true, "cmd": ". /edx/app/insights/nodeenvs/insights/bin/activate && /edx/app/insights/edx_analytics_dashboard/node_modules/.bin/bower install --production --config.interactive=false", "delta": "0:00:20.804529", "end": "2016-03-16 05:26:04.708470", "rc": 1, "start": "2016-03-16 05:25:43.903941", "warnings": []}
==> default: stderr: bower  ECONFLICT Unable to find suitable version for backbone
```

I think the problem was that some bower dependency has been removed from insights but it was still present on my devstack, requiring a different version of backbone than other current bower dependencies. I nuked the bower_components folder then provisioned again and it worked fine.

mulby:
@bradenmacdonald: I've updated this PR with revised names, files are uploaded to S3, I just need to ensure they are publicly accessible.

bradenmacdonald:
@mulby Yeah, I'm getting "Access Denied" if I try to download them.

mulby:
@bradenmacdonald can you try now?

bradenmacdonald:
@mulby Seems to work! I haven't tested the boxes, just confirmed that I can download them.

mulby:
@jibsheet and @brianhw is this ready to merge?

e0d:
LGTM :+1:

PR #2868, opened 2016-03-16
Added OAUTH2_ACCESS_TOKEN_URL setting to discovery service

clintonb:
ECOM-3895


clintonb:
@edx/devops please review

feanil:
:+1:

clintonb:
@jibsheet I updated this per our conversation earlier today. Would you review, please? I will trim this to 2 commits after approval.

jibsheet:
@clintonb at first glance, looks fine, I want to grep in the secure repos when I'm on a machine with those checked out to make sure I know if we need a matching PR over there, or just deletions.  Will look in the morning.

jibsheet:
:+1: 
@clintonb I need to get thumbs on the two private PRs, otherwise if we build out of sync, credentials or discovery will fail to log in (we should probably do test builds of them once everything merges).
I'll ping when mine are ready to merge.

jibsheet:
@clintonb unfortunately, the travis failures are real, the CI jobs are failing because the docker containers don't have EDXAPP_LMS_ISSUER - you'll need to take a look at that code.

clintonb:
@jibsheet ready for final review; I ended up adding a default to work with Docker.
@cpennington please review the changes to `docker.mk`.

docker.mk
```diff
@@ -26,7 +26,7 @@ pkg: docker.pkg
 clean:
 	rm -rf .build
 
-docker.test.shard: $(foreach image,$(shell echo $(images) | tr ' ' '\n' | sed -n '$(SHARD)~$(SHARDS)p'),$(docker_test)$(image))
+docker.test.shard: $(foreach image,$(shell echo $(images) | tr ' ' '\n' | awk 'NR%$(SHARDS)==$(SHARD)'),$(docker_test)$(image))
```
cpennington, line 29: Ah, that's where that was coming from.

cpennington:
:+1:

jibsheet:
:+1: Assuming our builds check out in stage

clintonb:
Validated on stage!

PR #2872, opened 2016-03-17
Removing HOSTNAME_MODULESTORE_DEFAULT_MAPPINGS to make preview domain…

muzaffaryousaf:
Use `PREVIEW_LMS_BASE` in `HOSTNAME_MODULESTORE_DEFAULT_MAPPINGS` instead of the hardcoded value from configuration. This way we can set values other than `preveiw`.

[TNL-4073](https://openedx.atlassian.net/browse/TNL-4073)


feanil:
Is there a relevant edx-platform PR?

muzaffaryousaf:
@feanil  https://github.com/edx/edx-platform/pull/11661 We decided to hold on this PR.
Thanks.

muzaffaryousaf:
@symbolist @feanil This is ready to review, PTAL. Thanks

symbolist:
:+1:

feanil:
@muzaffaryousaf The goal of the `HOSTNAME_MODULESTORE_DEFAULT_MAPPINGS` list is to provide a general list of hostname mappings to modulestore versions.  Right now we only have draft and live but technically the modulestore allows for more kinds of versions and we may want to have different versions for different domain names.  We can have the default value be the same as the value of `PREVIEW_LMS_BASE` but I think removing the list altogether is not the correct decision.

symbolist:
@feanil The changes in LMS are doing that. If you look at the following code:
https://github.com/edx/edx-platform/blob/master/lms/envs/aws.py#L407-L414
1. It reads `HOSTNAME_MODULESTORE_DEFAULT_MAPPINGS` if it is defined in aws.py.
2. And adds a mapping to it of `PREVIEW_LMS_BASE` to the draft store.

Since 2 must always exist for preview to work, the entry in the configuration is redundant and requires making the same change in two places in the configuration if the preview domain needs to be changed. However, if we need to add more mappings in the future, we can define `HOSTNAME_MODULESTORE_DEFAULT_MAPPINGS` in the configuration again. Though I think in most cases, we will be using that hostname in LMS in other places and the pattern for 2 will make more sense.

symbolist:
Though we can leave in an empty dict in the configuration if that will be useful (and possible?).

feanil:
@symbolist that makes sense thanks.  :+1:   Can you please file a devops ticket pointing to this PR to remove the override from our secrets files?

muzaffaryousaf:
Ops ticket has been created https://openedx.atlassian.net/browse/DEVOPS-3898

PR #2873, opened 2016-03-17
Adding support for multiple auth credentials for nginx (OPS-952)

bilalahmad99:


playbooks/roles/nginx/defaults/main.yml
```diff
@@ -10,6 +10,15 @@ NGINX_EDXAPP_EXTRA_SITES: []
 NGINX_EDXAPP_EXTRA_CONFIGS: []
 NGINX_EDXAPP_CUSTOM_REDIRECTS: {}
 
+# Override these vars for adding user to nginx.htpasswd
+nginxusers:
+  - name: "{{ COMMON_HTPASSWD_USER }}"
+    password: "{{ COMMON_HTPASSWD_PASS }}"
+    state: present
+  - name: bilal
```
feanil: You should remove your self from the default settings.  Having a list of one as the default should be sufficient.

bilalahmad99: Done

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -10,6 +10,12 @@ NGINX_EDXAPP_EXTRA_SITES: []
 NGINX_EDXAPP_EXTRA_CONFIGS: []
 NGINX_EDXAPP_CUSTOM_REDIRECTS: {}
 
+# Override these vars for adding user to nginx.htpasswd
+nginxusers:
```
e0d: NGINXUSERS should be uppercase as we expect it to be overridden

bilalahmad99: Done

e0d:
:+1:

PR #2871, opened 2016-03-17
Install rabbitmq before celery workers in sandboxes

omarkhan:
#### Problem

We have been seeing this error when provisioning sandboxes:

```
TASK: [edxapp | ensure edxapp_workers has started] ****************************
failed: [149.202.174.205] => {"failed": true}
msg: edxapp_worker:cms_default_1: ERROR (abnormal termination)

FATAL: all hosts have already failed -- aborting
```

In `/edx/var/log/supervisor/cms_default_1-stderr.log` on that sandbox we can see:

```
[2016-03-14 16:15:43,947: ERROR/MainProcess] consumer: Cannot connect to amqp://celery:**@127.0.0.1:5672//: [Errno 111] Connection refused.
Trying again in 4.00 seconds...

[2016-03-14 16:15:47,955: ERROR/MainProcess] consumer: Cannot connect to amqp://celery:**@127.0.0.1:5672//: [Errno 111] Connection refused.
Trying again in 6.00 seconds...

[2016-03-14 16:15:53,965: ERROR/MainProcess] consumer: Cannot connect to amqp://celery:**@127.0.0.1:5672//: [Errno 111] Connection refused.
Trying again in 8.00 seconds...

[2016-03-14 16:16:01,980: ERROR/MainProcess] consumer: Cannot connect to amqp://celery:**@127.0.0.1:5672//: [Errno 111] Connection refused.
Trying again in 10.00 seconds...
```

It turns out that rabbitmq is not running, in fact it is not even installed. This is because the `edx_sandbox.yml` playbook runs the `edxapp` role **before** the `rabbitmq` role, and the `edxapp` role errors out because rabbitmq is missing, so rabbitmq never gets installed.
#### Solution

Run the `rabbitmq` role **before** the `edxapp` role in `edx_sandbox.yml`.

PR #2875, opened 2016-03-17
Get rabbit ready before we start workers

nedbat:
This is the same change as in #2871, but also applies it to the devstack
and fullstack vagrant plays.

### DIFF
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -34,6 +34,7 @@
     - edxlocal
     - role: mongo
       when: "'localhost' in EDXAPP_MONGO_HOSTS"
+    - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
     - { role: 'edxapp', celery_worker: True }
     - edxapp
     - notifier
@@ -42,7 +43,6 @@
     - edx_notes_api
     - demo
     - oauth_client_setup
-    - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
     - oraclejdk
     - role: elasticsearch
       when: "'localhost' in EDXAPP_ELASTIC_SEARCH_CONFIG|map(attribute='host')"
--- a/playbooks/vagrant-devstack.yml
+++ b/playbooks/vagrant-devstack.yml
@@ -22,13 +22,13 @@
     - mysql
     - edxlocal
     - mongo
+    - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
     - edxapp
     - oraclejdk
     - elasticsearch
     - forum
     - ecommerce
     - ecomworker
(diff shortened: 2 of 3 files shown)


openedx-webhooks:
## PR Cover Letter
- [ ] If available, a link to the relevant JIRA ticket in the body of the PR
  - `[PROJ-XXXX](https://openedx.atlassian.net/browse/PROJ-XXXX)`

## Sandbox
- [ ] Build a sandbox for your branch and add a link here

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Unit, integration, acceptance tests as appropriate
- [ ] Analytics
- [ ] Performance
- [ ] Database migrations are backwards-compatible

## Reviewers

- [ ] Code review #1
- [ ] Code review #2
- [ ] Docs review (required for UI strings/error messages)
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## Documentation considerations
- [ ] Are there user visible changes? If so, list who is affected, and include a pointer to the relevant code.
- [ ] Should this be mentioned in the release notes? If so, make sure that you included a description.

## DevOps assistance

If it's needed, or even if you suspect it might be needed, do any/all these things as early as possible. Don't surprise devops right before merging! For example, touching envs/aws.py or or envs/common.py are dead giveaways that you'll need assistance when pushing your changes out. Things that require devops attention include, but are not limited to:
- configuration changes
- feature flags
- migrations
  Ways to get devops attention, pick-and-choose whichever is most appropriate for your specific PR:
- [ ] Tag devops on your PR (`@edx/devops`), either for review (more specific to devops = better) or just a general FYI
- [ ] File a support ticket

FYI: Tag anyone who might be interested in this PR here.

## Post-review
- [ ] Squash commits

nedbat:
@feanil @jibsheet This is the same as #2871, with the change made in more places.  Do I need rabbit in devstack if edxapp doesn't have worker=True?

Also, bonus points for anyone who can explain why this is happening now.  It feels like we are leaking the future into the Dogwood past, and we don't know where.

I'll be cherry-picking this commit onto dogwood.rc when approved.

feanil:
The error looks to be un-related to this change. :+1:

jibsheet:
It looks like in devstacks, you can leave the order alone, since we only install rabbit for ecom_worker
caebe2d
Otherwise :+1:

feanil:
@jibsheet there is also an error with the edxapp workers that is being reported on the mailing list and in slack that should be adressed by moving the rabbit role up.  It seems like there is a chance for occassional deadlock when the workers come up and can't connect to rabbit.

jibsheet:
@feanil on devstack?  Devstack workers are in process I thought.

nedbat:
@jibsheet I wondered about the need for the re-ordering on devstack, but it still seems like a good idea to install lower-level services (like rabbit) before upper-level, and to have devstack and fullstack be similar where possible.

PR #2876, opened 2016-03-17
Prevent invalid display courseware in IE 10+ with high privacy settings

dmitry-viskov:
The problem reproduced only in IE browser with high privacy settings.
JS couldn't get Cookie for the CSRF protection and sends always `null`.
So because of this users can't login, get the course problems and make other actions connected with CSRF protection.
high_privaci_issue_in_ie
high_privaci_issue_in_ie2
So as you can see from the screenshots for every request server returns HTTP 403.
As we found it is connected with the cookie-restricting privacy feature called P3P. More details can be found here: http://blogs.msdn.com/b/ieinternals/archive/2013/09/17/simple-introduction-to-p3p-cookie-blocking-frame.aspx
and here:
http://stackoverflow.com/questions/8048306/what-is-the-most-broad-p3p-header-that-will-work-with-ie
As we found out adding P3P HTTP header solve this issue:

```
P3P: CP="Open EdX does not have a P3P policy."
```

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
@@ -42,6 +42,9 @@ error_page {{ k }} {{ v }};
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
   {% endif %}
 
+  # Prevent invalid display courseware in IE 10+ with high privacy settings
+  add_header P3P 'CP="Open EdX does not have a P3P policy."';
+
   # Nginx does not support nested condition or or conditions so
   # there is an unfortunate mix of conditonals here.
   {% if NGINX_REDIRECT_TO_HTTPS %}
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -62,6 +62,10 @@ error_page {{ k }} {{ v }};
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
   {% endif %}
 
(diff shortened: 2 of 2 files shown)


openedx-webhooks:
Thanks for the pull request, @StraNNiKK! I've created [OSPR-1197](https://openedx.atlassian.net/browse/OSPR-1197) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

nedbat:
@andy-armstrong do you know something about this? I'm assigning to devops but thought you might have an opinion also.

e0d:
@fredsmith let's pull this into next sprint.

fredsmith:
This is the first I remember hearing about P3P - according to [Wikipedia](https://en.wikipedia.org/wiki/P3P) it's only supported by MSIE, and no other browsers will notice this header.   I think this PR is OK to merge.  
:+1: 

@jibsheet second review?

jibsheet:
I wonder which of the other IDAs want this on also (things that the user interacts with and would have a CSRF cookie).  Probably at least insights, ecommerce and some others.

I'm also concerned by the persistently failing phantomjs installation in the travis tests.  It seems to insist on installing phantomjs 1.9.8 instead of 1.9.17/1.9.18, but I haven't been able to look harder.

e0d:
@nedbat noticing this has a DevOps thumb, but is still in awaiting prioritization, what the right way to get it's status updated?

nedbat:
@e0d you can change the status of the JIRA issue to update the labels (which I just did).

jibsheet:
I'll merge this, but it wants to be on for the other IDAs and wants to be overridable (Edx.org shouldn't serve a note about Open Edx, especially since we have a privacy policy).

PR #2878, opened 2016-03-17
Restore ecommerce roles to analytics devstack

bradenmacdonald:
In 48f79b673c1fd358232c5d68b06399ef0dc4100a we removed `ecommerce` from a default analytics stack installation, at my suggestion, but I think that was a mistake. Lately everyone on my team has had to locally revert that change since we've been working on eCommerce+analytics.

This is a simple proposed change to add Otto/ecommerce to future analytics devstacks.

### DIFF
--- a/playbooks/vagrant-analytics.yml
+++ b/playbooks/vagrant-analytics.yml
@@ -14,6 +14,7 @@
     EDXAPP_LMS_BASE: 127.0.0.1:8000
     EDXAPP_OAUTH_ENFORCE_SECURE: false
     EDXAPP_LMS_BASE_SCHEME: http
+    ECOMMERCE_DJANGO_SETTINGS_MODULE: "ecommerce.settings.devstack"
   roles:
     - common
     - vhost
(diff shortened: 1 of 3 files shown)


bradenmacdonald:
@itsjeyd FYI
@e-kolpakov Please review as part of OC-1469.

e-kolpakov:
@bradenmacdonald :+1: 

I believe it would look better if it was a revert of that commit, but it is fine as is.

e0d:
:+1:  Seems right.

bradenmacdonald:
@mulby Can you please review this too? (lower priority)

mulby:
@bradenmacdonald I just merged my PR that also adds in versions for insights and the api, can you rebase on master again?

We should then be able to grab dogwood versions for most of the repos.

Once that's all set :+1:

bradenmacdonald:
@mulby There are no conflicts between this PR and master after PR 2864 was merged - what do I need to change when I rebase?

mulby:
oh, huh, I had thought there would be conflicts since I also added versions to the Vagrantfile.

No changes necessary!

PR #2879, opened 2016-03-18
Increasing file descriptor limit (OPS-728)

bilalahmad99:


playbooks/roles/locust/tasks/main.yml
```diff
@@ -72,3 +72,9 @@
     name={{ locust_service_name }}
   when: not disable_edx_services
   sudo_user: "{{ supervisor_service_user }}"
+
+- name: increase file descriptor limit of the system (Session Logout and Login would be required)
+  lineinfile:
+    dest: "{{ ulimit_conf_file }}"
```
arbabnazar, line 78: what's the value of `ulimit_con_file`?

bilalahmad99: missed it

arbabnazar, line 78: Can you please share the verification method? I think we need to reboot to system after configuring these value to take affect.

bilalahmad99, line 78: No, just a logout of the session and log back in would be required. I verified it on my sandbox using command "ulimit -n".

e0d:
:+1:

PR #2881, opened 2016-03-20
Fix a bug in the git module: branches might not check out properly

nedbat:
If a git version is specified as a branch, and the commit referred to is
already on the local working tree, but the branch is not, then the
checkout will fail.

This is because the Ansible 1.9.3 module would check to see if the
commit was already present, and if so, did not fetch. This meant the
branch wasn't defined in the local working tree.

This can happen if you make a branch but there aren't any changes on the
branch.

### DIFF
--- a/playbooks/edx-east/legacy_certificates.yml
+++ b/playbooks/edx-east/legacy_certificates.yml
@@ -14,7 +14,7 @@
     - name: stop certs service
       service: name="certificates" state="stopped"
     - name: checkout code
-      git: >
+      git_2_0_1: >
         repo="{{ repo_url }}"
         dest="{{ repo_path }}"
         version="{{ certificates_version }}"
--- a/playbooks/library/git_2_0_1
+++ b/playbooks/library/git_2_0_1
@@ -1,0 +1,820 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# (c) 2012, Michael DeHaan <michael.dehaan@gmail.com>
+#
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# Ansible is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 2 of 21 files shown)


edx-webhook:
## PR Cover Letter
- [ ] If available, a link to the relevant JIRA ticket in the body of the PR
  - `[PROJ-XXXX](https://openedx.atlassian.net/browse/PROJ-XXXX)`

## Sandbox
- [ ] Build a sandbox for your branch and add a link here

## Testing
- [ ] i18n
- [ ] RTL
- [ ] Unit, integration, acceptance tests as appropriate
- [ ] Analytics
- [ ] Performance
- [ ] Database migrations are backwards-compatible

## Reviewers
- [ ] Code review #1
- [ ] Code review #2
- [ ] Docs review (required for UI strings/error messages)
- [ ] UX review
- [ ] Accessibility review
- [ ] Product review

## Documentation considerations
- [ ] Are there user visible changes? If so, list who is affected, and include a pointer to the relevant code.
- [ ] Should this be mentioned in the release notes? If so, make sure that you included a description.

## DevOps assistance
If it's needed, or even if you suspect it might be needed, do any/all these things as early as possible. Don't surprise devops right before merging! For example, touching envs/aws.py or or envs/common.py are dead giveaways that you'll need assistance when pushing your changes out. Things that require devops attention include, but are not limited to:
- configuration changes
- feature flags
- migrations
Ways to get devops attention, pick-and-choose whichever is most appropriate for your specific PR:
- [ ] Tag devops on your PR (`@edx/devops`), either for review (more specific to devops = better) or just a general FYI
- [ ] File a support ticket

FYI: Tag anyone who might be interested in this PR here.

## Post-review
- [ ] Squash commits

nedbat:
@feanil @e0d I changed all of the mentions of the git module that used a version= parameter.  Not sure if this is the way you would go, but I like that it will keep this mysterious problem from happening to others in the future.

feanil:
Instead of local, can you name it based on the version you copied from?  ie. `git_2_0` so we know when we can get rid of it.  Also if it's just 1.9.4 that is something we can move our fork to so I assume it's at least 2.0 that the module needs to come from?

nedbat:
The fix is not in 1.9.4, so we are stuck with a local version.  The fix is in 2.0.1 and not in 2.0, so I guess "git_2_0_1" is what you want?

feanil:
Yes, please.

nedbat:
OK: I've added "git_2_0_1", which is the git.py from 2.0.1.  It has this fix, plus a new option, verify_commit, which is the bulk of the difference.  I'm trying to find a way to show the diff between the 1.9.3 version and the 2.0.1 version, but GitHub doesn't have a way?

nedbat:
@e0d @feanil ping ping

feanil:
:+1:

e0d:
:+1:

PR #2882, opened 2016-03-21
Add scripting to restore a GoCD server backup.

doctoryes:
The current implementation assumes that a file named `go-server-latest-backup.zip` will be in the `playbooks` directory. This location assumption will change once the work to perform a backup and save it in external storage is performed. The backup will be downloaded from external storage instead.

Also, the backup ZIP file is assumed to have this structure:

```
backup_20160317-151706/
├── config-dir.zip
├── config-repo.zip
├── db.zip
└── version.txt
```

@macdiesel @feanil PTAL.

### DIFF
--- a/playbooks/roles/go-server/tasks/download_backup.yml
+++ b/playbooks/roles/go-server/tasks/download_backup.yml
@@ -1,0 +1,84 @@
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Task to download go-server backup
+#
+# Overview:
+#
+# Downloads a TGZ file containing a go-server backup.
+#
+# Example play:
+#
+# - name: Configure instance(s)
+#   hosts: go-server
+#   sudo: True
+#   vars_files:
+#     - "{{ secure_dir }}/admin/sandbox.yml"
+#   gather_facts: True
+#   roles:
+#     - common
+#
+
+- name: stop go-server
+  service:
+    name: "{{ GO_SERVER_SERVICE_NAME }}"
+    state: stopped
+
+- name: install system packages needed for downloading backup
+  apt:
+    pkg: "{{ item }}"
+    state: present
+    update_cache: true
+    cache_valid_time: 3600
+  with_items: GO_SERVER_BACKUP_APT_PKGS
+
+- name: install required python packages
+  pip:
+    name: "{{ item }}"
+    state: present
+  with_items: GO_SERVER_BACKUP_PIP_PKGS
+
+- name: create the temp directory
+  file:
+    path: "{{ GO_SERVER_TEMP_RESTORE_DIRECTORY }}"
+    state: directory
+    mode: 0755
+
+- name: get s3 one time url
+  s3: >
+    bucket="{{ GO_SERVER_BACKUP_S3_BUCKET }}"
+    object="{{ GO_SERVER_BACKUP_S3_OBJECT }}"
+    mode="geturl"
+    expiration=30
+  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT
+  register: s3_one_time_url
+
+- name: download from one time url
+  get_url:
+    url="{{ s3_one_time_url.url }}"
+    dest="{{ GO_SERVER_TEMP_RESTORE_DIRECTORY }}/{{ GO_SERVER_BACKUP_FILENAME }}"
+    mode=0600
+  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT
+  register: download_backup_s3
+
+- name: copy go-server backup from backup storage
+  unarchive:
+    src: "{{ GO_SERVER_TEMP_RESTORE_DIRECTORY }}/{{ GO_SERVER_BACKUP_FILENAME }}"
+    dest: "{{ GO_SERVER_TEMP_RESTORE_DIRECTORY }}/"
+    copy: no
+  when: download_backup_s3.changed
+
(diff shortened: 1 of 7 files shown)


macdiesel:
@doctoryes You can put the zip file under a directory named `playbooks/roles/go-server/files/<destination path>/<filename>.zip`

all else :+1:

feanil:
I think I would rather not check the file in, since this is an open repo and the backup is specific to us.  It would be better to be able to pass in a URL for the location of the backup and do the restore if the Backup URL is set.

doctoryes:
@feanil @macdiesel This PR is ready for another look. It creates a bash script that's called by cron to achieve the backups. I've tested this locally - and works on my machine! :fried_shrimp: 
There's a section in the `defaults` that is temporary. And in a couple of YML files, AWS credentials are saved - those sections are temporary. Once the go-server machine is authed to S3, those sections can be removed.

doctoryes:
Also @feanil , is the test failure a valid one?

playbooks/roles/go-server/tasks/download_backup.yml
```diff
@@ -0,0 +1,98 @@
...
+#   roles:
+#     - common
+#
+
+- name: stop go-server
+  service:
+    name: "{{ GO_SERVER_SERVICE_NAME }}"
+    state: stopped
+
+- name: install python-pip using apt-get
```
feanil: We have a general pattern for doing apt based installs.  It's good to have it take a list so that it's easy to add new dependencies without having to write new tasks.

https://github.com/edx/configuration/blob/master/playbooks/roles/edx_service/tasks/main.yml#L118-L128

playbooks/roles/go-server/templates/gocd_backup.j2
```diff
@@ -0,0 +1,47 @@
...
+gocd_backup_api_call="{{ GO_SERVER_BACKUP_API_URL }}"
+gocd_username="{{ GO_SERVER_USERNAME }}"
+gocd_password="{{ GO_SERVER_PASSWORD }}"
+gocd_backup_location="{{ GO_SERVER_BACKUP_TMP_LOCATION }}"
+s3_backup_bucket="{{ GO_SERVER_BACKUP_S3_BUCKET }}"
+snitch_url="{{ GO_SERVER_BACKUP_SNITCH_URL }}"
+
+# Trigger the backup and capture the backup path.
+backup_path=\
+$(curl -sS "$gocd_backup_api_call" \
```
feanil, line 39: What is the output path that comes out of this curl command?  Can you add a comment with an example?

feanil:
Just a few question but generally looks good.  The error is related to credential sand to related to work that you did here.

playbooks/roles/go-server/tasks/download_backup.yml
doctoryes: Will do.

playbooks/roles/go-server/templates/gocd_backup.j2
doctoryes, line 39: I added comments.

playbooks/roles/go-server/templates/gocd_backup.j2
```diff
@@ -0,0 +1,73 @@
...
+  ;;
+esac
+
+# Make backup into a tarball.
+tar zcvf "$gocd_backup_location" -C "$(dirname "$backup_path")" "$(basename "$backup_path")"
+
+# Transfer tarball to S3.
+aws_output=$(aws s3 cp "$gocd_backup_location" "s3://$s3_backup_bucket" 2>&1)
+case $aws_output in
+*failed*)
```
macdiesel, line 61: very clever.

doctoryes, line 61: Compliment - or insult? :smiley_cat:

macdiesel, line 61: Compliment!

playbooks/roles/go-server/tasks/setup_regular_backup.yml
```diff
@@ -0,0 +1,69 @@
...
+    src: credentials.j2
+    dest: /root/.aws/credentials
+    mode: 0600
+# TEMPORARY!!!!!!
+
+- name: create backup shell script
+  template:
+    src: gocd_backup.j2
+    dest: "{{ GO_SERVER_BACKUP_CRON_SCRIPT_LOCATION }}"
+    mode: 0770
```
macdiesel: Since this file will have a username and password perhaps we should set this to 700 and ensure that the owner and group is root:root.

macdiesel:
@doctoryes one question, since go-cd will need a username and password to call the api to generate the backup perhaps we should generate this combination during the initial server setup and keep it unique to the machine. Thoughts?

playbooks/roles/go-server/tasks/setup_regular_backup.yml
doctoryes: Couldn't hurt - will do.

doctoryes:
I agree that'd be a security boost as opposed to using a pre-baked u/p that's the same on all GoCD servers. I'll attempt it.

doctoryes:
@macdiesel @feanil PTAL. This PR is ready for another review. It requires the changes in this PR:
https://github.com/edx-ops/sandbox-secure/pull/46

macdiesel:
:+1:

macdiesel:
I had thought that the backup user would be not the admin user. That said I'm not sure that a user could call the backup API without being an admin user so it's probably a moot point.

doctoryes:
Indeed - I just experimented and only admin users are allowed to trigger a backup. GoCD has roles around authorizing users to perform pipeline tasks - but admin tasks like backup require an admin.
https://docs.go.cd/current/configuration/dev_authorization.html#authorization

feanil:
:+1:  once temp sections are deleted.

doctoryes:
Temp sections are now removed - removing the AWS credentials creation. Re-jiggering the IAM to auth the go-server machine to the S3 bucket will be covered in this story:
https://openedx.atlassian.net/browse/TE-1259

macdiesel:
:shipit:

PR #2884, opened 2016-03-24
complete and operationalize harstorage role [OPS-1372]

bilalahmad99:


playbooks/roles/harstorage/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+
+- name: build virtualenv
+  command: "virtualenv {{ harstorage_venv_dir }}"
+  args:
+    creates: "{{ harstorage_venv_dir }}/bin/pip"
+  sudo_user: "{{ harstorage_user }}"
+
+- name: install python packages
+  pip:
+    name: "{{ item.name }}"
```
arbabnazar, line 44: `virtualenv` is missing. it has been created in the previous step but didn't use here.

playbooks/roles/harstorage/meta/main.yml
```diff
@@ -0,0 +1,32 @@
...
+# Example:
+#
+# dependencies:
+#   - {
+#   role: my_role 
+#   my_role_var0: "foo"
+#   my_role_var1: "bar"
+# }
+
+dependencies:
```
arbabnazar, line 22: @e0d I think we have changed this, instead of using `meta` dependencies here, we addressed these in playbook.

playbooks/roles/harstorage/defaults/main.yml
```diff
@@ -0,0 +1,51 @@
...
+#
+
+harstorage_debian_pkgs:
+  -  lib32stdc++6
+  -  python
+  -  python-cairo
+  -  python-rsvg
+  -  python-setuptools
+  -  python-pastescript
+
```
arbabnazar, line 40: @bilalahmad99 where did you install the `virtualenv` package?

playbooks/roles/harstorage/tasks/main.yml
```diff
@@ -21,13 +21,50 @@
...
+  file:
+    path: "/{{ harstorage_etc }}"
+    state: directory
+    mode: 0755
 
 - name: setup the harstorage production.ini file
   template:
     src: './{{ harstorage_etc }}/production.ini.j2'
-    dest: '{{ harstorage_etc }}/production.ini'
+    dest: '/{{ harstorage_etc }}/production.ini'
```
arbabnazar: I think, we don't need the `/` here because the value of `COMMON_APP_DIR: "/edx/app"` already containing the leading `/`. Did you test this?

playbooks/roles/harstorage/tasks/main.yml
```diff
@@ -21,13 +21,50 @@
...
     owner: '{{ harstorage_user }}'
     group: '{{ harstorage_user }}'
     mode: 0644
 
+- name: install harstorage
+  easy_install:
+     name: "{{ harstorage_code_dir }}/harstorage-1.0-py2.7.egg"
+
+- name: apply config
+  command: "/usr/bin/paster setup-app /{{ harstorage_etc }}/production.ini"
```
arbabnazar: we can make it idempotent by using `creates` or with `when`

playbooks/roles/harstorage/tasks/main.yml
```diff
@@ -21,13 +21,50 @@
...
 
+- name: run harstorage
+  command: "/usr/bin/paster serve /{{ harstorage_etc }}/production.ini"
```
arbabnazar: same here

playbooks/roles/harstorage/tasks/main.yml
arbabnazar, line 44: `virtualenv: "{{ harstorage_venv_dir  }}"` something like this

playbooks/roles/harstorage/defaults/main.yml
```diff
@@ -0,0 +1,51 @@
...
+harstorage_python_pkgs:
```
e0d, line 41: We should get virtualenv and python from the common role, it should be included in the meta file for this role.  Compare with the ecommerce role.

playbooks/roles/harstorage/meta/main.yml
e0d, line 22: We want deployment target specific roles in the play and keep things like common and edx_service here. Common must be valid everywhere, but we decide on the deployment target in the play, target might be AWS or docker, for example.

playbooks/roles/harstorage/tasks/main.yml
```diff
@@ -0,0 +1,70 @@
...
+#
+#
+
+- name: install debian dependecies
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items: harstorage_debian_pkgs
+
+- name: download harstorage egg file]
```
e0d: @cpennington any idea why this is the recommended installation path rather than installing via pip from github?  Is setup.py fully bjorked?

playbooks/roles/harstorage/templates/edx/etc/harstorage/production.ini.j2
```diff
@@ -0,0 +1,58 @@
...
+[app:main]
+use = egg:harstorage
+full_stack = true
+static_files = true
+temp_store = %(here)s/data
+bin_store  = %(here)s
+ps_enabled = true
+static_version = 1.0
+
+mongo_replicate = false
```
e0d, line 23: We'll want to make much of this vars with defaults.

playbooks/roles/harstorage/defaults/main.yml
```diff
@@ -0,0 +1,51 @@
...
+   VERSION: master
+   DESTINATION: '{{ harstorage_code_dir }}'
+
+#
+# OS packages
+#
+
+harstorage_debian_pkgs:
+  -  lib32stdc++6
+  -  python
```
e0d: Python will come from common.

playbooks/roles/harstorage/defaults/main.yml
```diff
@@ -0,0 +1,51 @@
...
+   REPO: harstorage
+   VERSION: master
+   DESTINATION: '{{ harstorage_code_dir }}'
+
+#
+# OS packages
+#
+
+harstorage_debian_pkgs:
+  -  lib32stdc++6
```
e0d, line 39: I'm a little concerned about this dependency, but assume it's coming from the harstorage installation guide.

playbooks/roles/harstorage/tasks/main.yml
cpennington: Nope, no idea.

playbooks/roles/harstorage/tasks/main.yml
bilalahmad99, line 44: done

playbooks/roles/harstorage/defaults/main.yml
bilalahmad99, line 40: comming from common role

playbooks/roles/harstorage/defaults/main.yml
bilalahmad99, line 39: yes It is coming from installation guide https://code.google.com/archive/p/harstorage/wikis/Installation.wiki

PR #2885, opened 2016-03-24
OPS-1383: Add role for turning on enhanced networking

maxrothman:
@edx/devops

### DIFF
--- a/playbooks/roles/enhanced_networking/defaults/main.yml
+++ b/playbooks/roles/enhanced_networking/defaults/main.yml
@@ -1,0 +1,2 @@
+profile: edx
+compatible_instance_types: ['c3', 'c4', 'd2', 'i2', 'm4', 'r3']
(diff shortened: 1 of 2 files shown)


feanil:
You use a profile when directly calling the aws cli but not when you are using the `ec2` module?  What's the reason for the inconsistency?

smaeda-ks:
Hi,

Anyway it looks like need few more tasks for this role if we use Ubuntu, right?

http://docs.aws.amazon.com/AWSEC2/latest/UserGuide/enhanced-networking.html#enhanced-networking-ubuntu

maxrothman:
@smaeda-ks we only ever run this role on Ubuntu 14.04 which already has the kernel module installed, so those steps are not necessary. However, I will add a check to the beginning of this role that makes sure you've got >14.04.

smaeda-ks:
@maxrothman 

Thank you for the information.
I understood.

Thanks

jibsheet:
@smaeda-ks I added some kernel module checking (Rather than explicit OS checking) but we're running this on 14.04 for our mongo hosts.
@feanil I don't know why profile was missing from the ec2 module call, I've added it while I was touching up the kernel module checking.

PR #2886, opened 2016-03-25
New payment processor configuration format

e-kolpakov:
**Description:** This PR updates default ecommerce PAYMENT_PROCESSOR_CONFIG to use new format introduced in https://github.com/edx/ecommerce/pull/630

### DIFF
--- a/playbooks/roles/ecommerce/defaults/main.yml
+++ b/playbooks/roles/ecommerce/defaults/main.yml
@@ -80,6 +80,7 @@ ECOMMERCE_PAYPAL_CANCEL_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/can
 ECOMMERCE_PAYPAL_ERROR_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/error/'
 
 ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
+  edx:
     cybersource:
       profile_id: '{{ ECOMMERCE_CYBERSOURCE_PROFILE_ID }}'
       merchant_id: '{{ ECOMMERCE_CYBERSOURCE_MERCHANT_ID }}'


openedx-webhooks:
Thanks for the pull request, @e-kolpakov! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=2886&repo=edx%2Fconfiguration

e-kolpakov:
@douglashall, @clintonb / @mjfrey configuration PR for https://github.com/edx/ecommerce/pull/630 - please review/approve

jbzdak:
This is incomplete, as in recent ecommerce changes settings processor configuration has undergone major change, and this yet to be reflected in the provisioning scripts. PR that was supposed to bring the two in line was closed recently: https://github.com/edx/configuration/pull/2809 .

clintonb:
@edx/devops this will require an update to the secure configuration.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -80,6 +80,7 @@ ECOMMERCE_PAYPAL_CANCEL_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/can
 ECOMMERCE_PAYPAL_ERROR_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/commerce/checkout/error/'
 
 ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
+  edX:
```
clintonb: Lowercase.

clintonb:
:+1:

clintonb:
P.S. This should _not_ be merged until after https://github.com/edx/ecommerce/pull/630 is merged.

douglashall:
:+1: @jbzdak https://github.com/edx/configuration/pull/2809 was closed because it was no longer required by the refactoring done in ecommerce.

maxrothman:
@clintonb we don't seem to override `ECOMMERCE_PAYMENT_PROCESSOR_CONFIG` in secure_config. Is there a particular reason why you think secure config changes will be necessary?

clintonb:
@maxrothman we have to override the payment settings in order to keep our CyberSource and PayPal credentials.

maxrothman:
@clintonb right, but that's being done by overriding `ECOMMERCE_CYBERSOURCE_MERCHANT_ID`, `ECOMMERCE_CYBERSOURCE_PROFILE_ID`, etc. which get pulled into `ECOMMERCE_PAYMENT_PROCESSOR_CONFIG`. As far as I can tell, this change doesn't require any override changes.

e-kolpakov:
@clintonb @maxrothman I've just merged https://github.com/edx/ecommerce/pull/630 - merging this one as well.

@maxrothman the idea behind this PR is that there might be multiple Sites/Partners, each having dedicated Cybersource and Paypal config. I don't have access to secure config of course, but if it is similar in structure to `/edx/etc/ecommerce.yml` (produced from this file), it should be updated the same way it is done in this PR (i.e. wrap existing values into `edx`). As far as I can tell, settings read from yaml [replace](https://github.com/edx/ecommerce/blob/master/ecommerce/settings/production.py#L48) `PAYMENT_PROCESSOR_CONFIG` values set in _oscar.py and base.py, so if secure config should be updated if it is consumed in a similar way.

e0d:
@e-kolpakov @douglashall @clintonb @maxrothman why is edx the exemplar in the public defaults rather than something generic?  Is that key name depended on elsewhere?

maxrothman:
re secure config changes: we don't have a fully separate config file for secure config, we just have a set of overrides. If we don't override something, the default is used. Since we don't override `ECOMMERCE_PAYMENT_PROCESSOR_CONFIG`, the default is used, which you changed already.

e-kolpakov:
@e0d no, it's not; but essentially it is edx-ecommerce, so having `edx` as a default is more or less sensible. If you have strong arguments for changing it to something else - I'll be happy to change that.

PR #2889, opened 2016-03-25
Added settings for edx-drf-extensions to discovery role

clintonb:
ECOM-3895


clintonb:
@jibsheet can you confirm that `DISCOVERY_OAUTH2_ACCESS_TOKEN_URL` is NOT set in the secure repo? Note that I am not asking you to add stuff to secure. Progress!! :clap:

jibsheet:
@clintonb can confirm that is not overridden.  I also checked the DISCOVERY_SERVICE_CONFIG and it is not overridden.

:+1:

PR #2892, opened 2016-03-28
Set ecommerce JWT_LEEWAY

rlucioni:
@jimabramson @fredsmith, in case restarting the crashed NTP daemon isn't enough, this change should make ecommerce more tolerant of slight clock skew. We've used this approach before to address clock skew issues with [programs](https://github.com/edx/programs/commit/afe3b221743ed42dc0e70d1e410e9a6470f94f29).


jimabramson:
:+1:  - assuming this nested value gets passed through in overrides as well.

jimabramson:
@rlucioni if not already I also suggest making this the application default in ecommerce (base) settings.

rlucioni:
@jimabramson the JWT_AUTH hash here will overwrite the base JWT_AUTH dictionary in ecommerce base settings. Maybe a better solution would be to update the JWT_AUTH dictionary within production.py after loading YAML overrides, like we do for newer IDAs?

fredsmith:
:+1:

jimabramson:
@rlucioni that does sound like a nicer approach - your call as to whether it's necessary / justified in the scope of this bugfix.  if config defaults/overrides fix the problem we have today then the ecommerce refactor feels optional.

jimabramson:
@rlucioni might want to rerun the build - one of the fails looks like a transient issue but not sure about the js-related one.

rlucioni:
Looks like the remaining build failure is due to difficulties [installing phantomjs for credentials](https://travis-ci.org/edx/configuration/jobs/119072198#L2023), unrelated to the changes in this PR. I'm going to try re-running to see if the error is transient.

jimabramson:
:+1:

PR #2894, opened 2016-03-28
Jibsheet/mongo 3֊ combined updates

jibsheet:
Pulling together a number of new ansible roles and a version bump and some other testing for our new mongo cluster.  Only part of this will ever be merged, most will be rebased away as other branches merge.

PR #2895, opened 2016-03-28
mount ebs role

jibsheet:
Adds a new mount_ebs role that you can pass a group of volumes to and it will ensure those volumes have the correct filesystem and are found at the correct mountpoint.

@maxrothman You probably have the most context to look at my updates.

I need to squash this all down before merge, but wanted people to be able to look.

### DIFF
--- a/playbooks/roles/mount_ebs/tasks/main.yml
+++ b/playbooks/roles/mount_ebs/tasks/main.yml
@@ -54,8 +54,16 @@
     volumes | selectattr('device', 'equalto', item.device) | list | length != 0 and
     (volumes | selectattr('device', 'equalto', item.device) | first).mount != item.mount
   with_items: ansible_mounts
-# fail here if you attempted to unmount something instead of unmounting it
-# Meaning UNMOUNT disks is false, but the volumes indicates we should be fiddling mounts
+
+# If there are disks we want to be unmounting, but we can't because UNMOUNT_DISKS is false
+# that is an errorable condition, since it can easily allow us to double mount a disk.
+- name: Check that we don't want to unmount disks when UNMOUNT_DISKS is false
+  fail: msg="Found disks mounted in the wrong place, but can't unmount them.  This role will need to be re-run with -e 'UNMOUNT_DISKS=True' if you believe that is safe."
+  when: >
+    not UNMOUNT_DISKS and
+    volumes | selectattr('device', 'equalto', item.device) | list | length != 0 and
+    (volumes | selectattr('device', 'equalto', item.device) | first).mount != item.mount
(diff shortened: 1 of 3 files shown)


playbooks/roles/mount_ebs/tasks/main.yml
```diff
@@ -0,0 +1,72 @@
...
+    dev: "{{ item.device }}"
+    fstype: "{{ item.fstype }}"
+    # Necessary because AWS gives some ephemeral disks the wrong fstype by default
+    force: true
+  with_items: volumes
+
+# This can fail if one volume is mounted on a child directory as another volume
+# and it attempts to unmount the parent first. You can fix it by manually
+# unmounting the child before running this role or hoping that the volumes dict
+# is randomly ordered correctly.
```
feanil: Why not make it a list of dictionaries?

jibsheet: We wrote this comment before I fully understood the role.  It has to be an ordering issue, but it's also a "your server is getting upgraded/is configured wrong" case.

I'll see if I can retest tomorrow.

jibsheet: @feanil turns out it is order dependent, but we're not iterating the list we control.  Review the updated comment please.  If you see a clever solution, let me know.  I only see over complicated solutions that will be hard to implement in jinja.  If we want to fix the case, I suspect this needs to become a module.

feanil:
:+1:

jibsheet:
Since this includes a jinja upgrade, I'm building sandboxes and AMIs to test it.

jibsheet:
Looking at the last commit here - https://github.com/edx/configuration/commit/45ae2295f3d9b11dba30d5467800bc700ef28b78
Is that something we want?  This isn't overriden, it's passed as an argument (see https://github.com/edx/configuration/pull/2894/commits/a3b875a89a0e59b4ca816f47f61518ae382f708c for an example of how we want to do this - although that may need some additional work)

feanil:
Were we not defining mongo volumes before? I thought we had work in the existing mongo playbook that dealt with volumes, were those hard coded?

jibsheet:
They were hard coded in cloud migrations before, this is replacing that part of it.

feanil:
Can we have the mongo role have a meta dependency on this role and pass things that way instead of in playbooks?

maxrothman:
@jibsheet FYI the second commit (a3b875a) is a dead link.

We've talked about whether dependencies should be meta-dependencies or listed in playbooks before. Anyone remember what conclusion we reached?

PR #2896, opened 2016-03-29
Add playbook for managing edxapp users

maxrothman:
@edx/devops

### DIFF
--- a/playbooks/edx-east/manage_edxapp_users_and_groups.yml
+++ b/playbooks/edx-east/manage_edxapp_users_and_groups.yml
@@ -72,7 +72,7 @@
         manage_group {{ item.name | quote }}
         {% if item.get('permissions', []) | length %}--permissions {{ item.permissions | default([]) | map('quote') | join(' ') }}{% endif %}
         {% if item.get('remove') %}--remove{% endif %}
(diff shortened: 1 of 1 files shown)


playbooks/edx-east/manage_edxapp_users_and_groups.yml
```diff
@@ -0,0 +1,77 @@
...
+- hosts: all
+  vars:
+    python_path: /edx/bin/python.edxapp
+    manage_path: /edx/bin/manage.edxapp
+  tasks:
+    - name: Manage groups
+      shell: >
+        {{ python_path }} {{ manage_path }} lms manage_group {{ item.name }}
+        --permissions {{ item.permissions | default([]) | join(' ') }}
+        {% if item.get('remove') %}--remove{% endif %}
```
feanil, line 74: Why `item.get()` instead of `item.remove`?

maxrothman, line 74: Because `remove` is optional. If it's not defined, we don't want an error.

feanil:
:+1:

jibsheet:
:+1:

PR #2897, opened 2016-03-30
pass the read replica hosts list as external variables from secure repo

arbabnazar:

Reference: https://github.com/edx-ops/edx-secure/pull/815/files


e0d:
:+1:

PR #2901, opened 2016-03-30
cleaning up my mistake

bilalahmad99:
please review @e0d @arbabnazar

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -219,6 +219,7 @@ if [[ $basic_auth == "true" ]]; then
 COMMON_ENABLE_BASIC_AUTH: True
 COMMON_HTPASSWD_USER: $auth_user
 COMMON_HTPASSWD_PASS: $auth_pass
+NGINX_USERS: $nginx_users
 XQUEUE_BASIC_AUTH_USER: $auth_user
 XQUEUE_BASIC_AUTH_PASSWORD: $auth_pass
 EOF_AUTH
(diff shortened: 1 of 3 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -219,6 +219,7 @@ if [[ $basic_auth == "true" ]]; then
 COMMON_ENABLE_BASIC_AUTH: True
 COMMON_HTPASSWD_USER: $auth_user
 COMMON_HTPASSWD_PASS: $auth_pass
+NGINX_USERS: $nginx_users
```
e0d: I think we need to check that $nginx_users is not null before overriding NGINX_USERS.  Otherwise we need to pass in the default value everywhere.

Something like this, though untested.

```
if [[ -n $nginx_users ]]; then
    NGINX_USERS=$nginx_users
fi
```

bilalahmad99: done

arbabnazar: Here you have just checked the string for non-zero, what if it will be zero? like @e0d said we need to pass the default value if it will be zero.

bilalahmad99: Either it will be null in case of a job where we dont give text param or it will be a json where we have manually added the param. Dont see any case where it will be zero.
speaking of which, is there any job which uses the anible_provision.sh script except for the ones where we have added the param i.e. ansible-provision and ansible-provision-continuous-integration?

arbabnazar: I mean zero(empty string) not the digit(0), how you will or are addressing if it is empty or null here?

e0d:
@edx/devops second set of eyes, the idea of the complexity in ansible_provision.py is to only override the defaults in defaults/main.yaml when a value is provided in the environment.  In this was Jenkins jobs can, but needn't always, provide overrides.

e0d:
@arbabnazar second set of eyes.

From me: :+1:

arbabnazar:
:+1:

PR #2902, opened 2016-03-30
Update wiki links in file headers

maxrothman:
The headers were pointing to the old github wiki. Since we've officially
moved to Confluence, these links should be updated.


feanil:
:+1:

PR #2904, opened 2016-03-31
Add API management settings.

peter-fogg:
@bderusha @clintonb this is the configuration PR to add the new settings. I'll open a ticket to update the secure configuration once you've finished with review of this and https://github.com/edx/edx-platform/pull/11982.

ECOM-3940


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -137,6 +137,11 @@ EDXAPP_ENABLE_OAUTH2_PROVIDER: false
 
 EDXAPP_ENABLE_MOBILE_REST_API: false
 
+# Settings for API access management
+EDXAPP_ENABLE_API_ACCESS_MANAGEMENT: false
```
clintonb: Any reason we aren't using configuration models for this?

peter-fogg: Yes. I originally wanted to write it with a config model, but we're using the setting to enable/disable the associated URLs. Since the urlconf is only loaded once, a (dynamic) configuration model doesn't work.

bderusha:
Makes sense. :+1:

playbooks/roles/edxapp/defaults/main.yml
clintonb: How abut we add the check at the API request view level? If the feature is disabled, the view returns 404. My preference for a DB-based feature flag is to allow for quickly enabling/disabling this functionality if things go wrong, and to be able to test in stage/production without a redeploy.

The flag is only needed for the non-staff-facing views. It's fine if the administrative portions are enabled as that code should be isolated enough to avoid breaking other admin panels.

peter-fogg: Yeah, I certainly agree with using database-backed configuration when possible. That idea seems like a good way of going about it; I'll update.

peter-fogg: I've updated this and https://github.com/edx/edx-platform/pull/11982.

clintonb:
:+1:

PR #2906, opened 2016-03-31
Ansible playbook to launch an instance for the release-pipeline work.

macdiesel:
@feanil PTAL. Would love to hear your recommendations around where to store the artifacts and where these templates should be placed to appease the Dev-ops gods.

Jira: https://openedx.atlassian.net/browse/TE-1249


playbooks/continuous_delivery/launch_instance.yml
```diff
@@ -0,0 +1,89 @@
+# This playbook will launch an ec2 instance in a VPC.
+# This instance will have an autogenerated key.
+#
+# required variables for this playbook:
+#   - ec2_region              - The region the server should be brought up in
+#   - ec2_subnet_id           - Subnet to bring up the ec2 instance
+#   - base_ami_id             - The base base AMI-ID
+#   - ec2_vpc_subnet_id       - The Subnet ID to bring up the instance
+#   - ec2_instance_type       - The instance type to use
```
feanil: This looks like it's not required since you set a default. Let's seperate the list.  A set of required things, and a set of optional things.  I also notice that `actifact_path` is not in this list.

playbooks/continuous_delivery/launch_instance.yml
```diff
@@ -0,0 +1,89 @@
...
+    ebs_volume_size: 8
+    ec2_timeout: 500
+    ec2_assign_public_ip: no
+  gather_facts: False
+  connection: local
+  tasks:
+
+  - name: Generate UUID for keypair
+    command: cat /proc/sys/kernel/random/uuid
+    delegate_to: "127.0.0.1"
```
feanil: Why do you have all these `delegate_to` lines?  If you are already on a local connection, I would expect not to need this.

playbooks/continuous_delivery/launch_instance.yml
```diff
@@ -0,0 +1,89 @@
...
+
+  - name: Generate key material artifact for jobs down stream
+    template:
+      src: templates/local/key.pem.j2
+      dest: "{{ artifact_path }}/{{ unique_key_name.stdout }}/key.pem"
+      mode: 0600
+    delegate_to: "127.0.0.1"
+
+
```
feanil, line 93: Maybe add a `wait_for` to make sure the instance is up and you can hit port 22?  Or some other check to make sure the instance came up and is healthy.

feanil:
A few questions but general design makes sense.

playbooks/continuous_delivery/launch_instance.yml
macdiesel: done

playbooks/continuous_delivery/launch_instance.yml
macdiesel: removed.

playbooks/continuous_delivery/launch_instance.yml
macdiesel, line 93: I have added a task to wait for ssh to be available on the instance.

macdiesel:
@feanil I have updated the playbook. The last outstanding question I have is what should we set for the location of the artifacts that this job will generate? It's currently set to /tmp/ansible-runtime though that can be overridden at runtime.

feanil:
I think that is a fine default if it can be overridden later.  We probably want it to be keyed specific to any give run of a pipeline but we can pass that in later.

feanil:
:+1:

macdiesel:
@feanil the base directory that is set by the defaults holds a directory addressed by the identifier generated during the key making process. Information specific to each run is stored there.

macdiesel:
@doctoryes if you want to get in on the review train. :steam_locomotive:

doctoryes:
:mountain_cableway: 

LGTM :+1:

PR #2908, opened 2016-04-01
Use the newest version of the agents

jibsheet:
@edx/devops


jibsheet:
Now being used happily in both environments.

maxrothman:
:+1:

PR #2910, opened 2016-04-01
The downloads-distro.mongodb.org URL only has 2.6 packages?

jibsheet:
This worked at one time, but stopped.  I've updated the URL to be
consistent with their current docs (and what we do in the mongo_3 role).


feanil:
:+1:

PR #2911, opened 2016-04-01
Update mongo 3.0 play to use our two new roles

jibsheet:
This moves functionality from cloud_migrations into the ansible
role so that we can build machines from their scratch 'terraformed' state.

@edx/devops

### DIFF
--- a/playbooks/roles/mongo_3_0/meta/main.yml
+++ b/playbooks/roles/mongo_3_0/meta/main.yml
@@ -1,3 +1,5 @@
 ---
 dependencies:
   - common
+  - role: mount_ebs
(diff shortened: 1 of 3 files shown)


playbooks/roles/mongo_3_0/meta/main.yml
```diff
@@ -1,3 +1,5 @@
 ---
 dependencies:
   - common
+  - role: mount_ebs
```
feanil, line 4: Nit: We should rename this to mount_volumes or something since it's not really EBS specific but that doesn't have to be a part of this PR.

feanil:
:+1:

maxrothman:
:+1:

PR #2913, opened 2016-04-01
Add dependency on "common" role to the elasticsearch role.

smarnach:
The role makes use of some variables from common_vars (notably COMMON_*_DIR), so it should depend on the "common" role.  With this change applied, it's possible to run the elasticsearch role using the run_role.yml playbook, which is sometimes convenient.

### DIFF
--- a/playbooks/roles/elasticsearch/meta/main.yml
+++ b/playbooks/roles/elasticsearch/meta/main.yml
@@ -1,0 +1,3 @@
+---
+dependencies:
+  - common


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=2913

feanil:
:+1: 

@edx/devops second review?

jibsheet:
:+1:

smarnach:
@feanil @jibsheet Thanks for the quick response!

feanil:
Will merge once the tests have passed.

PR #2914, opened 2016-04-01
Update nltk URLs

maxrothman:
The old ones were returning 301s pointing to the updated URL

This unbreaks CI.

This was the response from the old URLs:

```
HTTP/1.1 301 Moved Permanently
Server: GitHub.com
Content-Type: text/html
Location: http://nltk.github.io/nltk_data/packages/corpora/wordnet.zip
```


jibsheet:
:+1:

PR #2919, opened 2016-04-04
Continuous delivery playbook to create an AMI.

macdiesel:
@doctoryes @feanil PTAL.

Please note that playbooks/library/ec2_ami_local.py has been pulled in from ansible 2.0.0.1 to be able to support ami tagging. See #2881 for reference.

Jira: https://openedx.atlassian.net/browse/TE-1250

### DIFF
--- a/playbooks/continuous_delivery/create_ami.yml
+++ b/playbooks/continuous_delivery/create_ami.yml
@@ -1,0 +1,100 @@
+# This playbook will create an AMI from an EC2 instance launched with the
+# continuous_deliver/launche_instance playbook.
+#
+# Required variables for this playbook:
+#
+#   - instance_id                     - the ec2 instance ID use to create the AMI
+#   - edx_environment                 - value to use for the environment tag
+#   - deployment                      - value to use for the deploy tag
+#   - cluster_repo                    - the url of the github repo for this cluster
+#   - cluster_version                 - git hash of the cluster (play, service, IDA) being deployed
+#   - play                            - the play that was run
+#   - configuration_repo              - The github url for the configuration repo
+#   - configuration_version           - The version (git hash) of configuration
+#   - configuration_secure_repo       - the github url for the configuration secure repository
+#   - configuration_secure_version    - the version (git hash) of configuration secure
+#   - cache_id                        - the cache_id version
+#
+# Other variables
+#   - ec2_region                      - The region used to create the AMI
+#   - ami_creation_timeout            - how long to wait before giving up on AMI creation in seconds
+#   - ami_wait                        - (yes/no) should ansible pause while
+#   - no_reboot                       - (yes/no) should the instance not be rebooted during AMI creation
+#   - artifact_path                   - the path to where this ansible run stores the artifacts for the pipeline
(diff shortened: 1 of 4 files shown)


feanil:
The AMI also needs to be tagged with all the tags that are on the instance as well the version of configuration and configuration_secure that were used to make it: https://github.com/edx/configuration/blob/master/util/vpc-tools/abbey.py#L543-L566

playbooks/library/ec2_ami_local.py
```diff
@@ -0,0 +1,405 @@
+#!/usr/bin/python
```
feanil: This should be named something like `ec2_ami_2_x_x` to indicate the version of ansible it was pulled from.  This makes it easier for us to know which up-stream version it came from and when we can get rid of it.

macdiesel: would you prefer the version of ansible this was pulled from or the version of ansible that we needed to be on for the missing feature?

macdiesel:
@feanil @doctoryes comments addressed PTAL. 

I named the ec2_ami module ec2_ami_2_0_0_1 as that was the version I pulled this from and is also the latest current ansible release. This release looks like a bug fix release so the version is both the version it was pulled from as well as the version that if we surpass should remove this module.

playbooks/continuous_delivery/create_ami.yml
```diff
@@ -0,0 +1,70 @@
...
+
+  - name: Create AMI
+    ec2_ami_2_0_0_1:
+      instance_id: "{{ instance_id }}"
+      name: "{{ cluster }} -- {{ cluster_version }}"
+      region: "{{ ec2_region }}"
+      wait: "{{ ami_wait }}"
+      wait_timeout: "{{ ami_creation_timeout }}"
+      no_reboot: "{{ no_reboot }}"
+      description: "AMI built via edX continious delivery pipeline - Ansible version: {{ ansible_version }}"
```
doctoryes: Nit: typo for `continuous`.

doctoryes:
:+1:

playbooks/continuous_delivery/create_ami.yml
macdiesel: fixed.

macdiesel:
@feanil this is ready for another look.

playbooks/continuous_delivery/create_ami.yml
```diff
@@ -0,0 +1,107 @@
...
+      mode: 0600
+
+  - name: Send Hipchat notification AMI is finished baking
+    hipchat:
+      api: "{{ hipchat_url }}"
+      token: "{{ hipchat_token }}"
+      room: "{{ hipchat_room }}"
+      msg: "Finished baking AMI for: {{ play }} \n
+            AMI-ID: {{ ami_register.image_id }} \n
+            {{ play }} repo: {{ cluster_repo }} \n
```
feanil: We only need to know the AMI ID. Alton can show you all tags that are currently on an AMI.

feanil:
:sheep: :it:

PR #2920, opened 2016-04-04
If user has no ssh keys, user role should fail when trying to add keys.

benpatterson:


benpatterson:
@mulby I was working with @jzoldak yesterday when we noticed that, if a given github user has no keys in his/her account, the add-ssh-keys job would pass, even though things were loaded (i.e., the /home/<user>/.ssh directory didn't even exist). I'm tagging you in case you have a good understanding of the portion of the statement I'm removing.

I'm also wondering if @fredsmith has an opinion, based on some git history.

Thanks for any feedback.

playbooks/roles/user/tasks/main.yml
```diff
@@ -116,7 +116,8 @@
   authorized_key:
     user: "{{ item.name }}"
     key: "https://github.com/{{ item.name }}.keys"
-  when: item.github is defined and item.get('state', 'present') == 'present'
+    state: present
```
arbabnazar, line 119: @benpatterson `state:present` is the default behaviour for `authorized_key` module. did this change fail the provisioning?

benpatterson, line 119: It was not present in our case yesterday, and it did not fail in the ansible play. Status was `OK`.

benpatterson, line 119: I'm assuming that `item.github` is defined, even when it is blank, and I'm not sure if the next `item.get` statement is a self-fulfilling prophecy. I'd need to dig more. But thought I'd start the discussion in case anyone has already run into this.

benpatterson:
I don't have the cycles to finish this or validate it. I'm closing it. Will log a bug.

PR #2928, opened 2016-04-05
Make jenkins_analytics install its required dependency

maxrothman:
@edx/devops


feanil:
:+1:   Nit: I'd rather this be a list of pip packages rather than specific to one package so it's easier to make changes in the future.  Not a deal breaker.

maxrothman:
@feanil Normally I'd agree, but I think that in this case, since this package isn't for the software set up by ansible, I prefer the self-documenting nature of a single task.

PR #2937, opened 2016-04-07
Fixed Docker issues for discovery service

clintonb:
- Corrected WSGI package path
- Using a single database user


clintonb:
@cpennington @fredsmith please review

cpennington:
Ah, nice. So, when running docker, this just uses the db user setup in the mysql container for all of the migration steps?

clintonb:
Yep. One user for (allthethings).

cpennington:
Awesome. :+1:

PR #2940, opened 2016-04-07
Add libffi to jenkins_worker and devstack

clytwynec:
When adding a python dependency to Jenkins builds for edxapp, I found that libffi was installed in devstack (via the ecommerce role) but not on jenkins_workers.  It is a requirement of the new python package (which is only needed for testing purposes, not in prod).  This PR adds libffi to the jenkins_worker role,  and explicitly adds it to devstack in the local_dev role to avoid it depending on the ecommerce role.

I've created an AMI and tested that this fixes the pip install issues with my platform changes.

### DIFF
--- a/playbooks/roles/jenkins_worker/defaults/main.yml
+++ b/playbooks/roles/jenkins_worker/defaults/main.yml
@@ -12,6 +12,7 @@ jenkins_edx_platform_version: master
 jenkins_debian_pkgs:
     - nodejs
     - pkg-config
+    - libffi-dev
 
 # packer direct download URL
 packer_url: "https://releases.hashicorp.com/packer/0.8.6/packer_0.8.6_linux_amd64.zip"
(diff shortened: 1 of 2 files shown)


clytwynec:
@benpatterson @jzoldak  Please review.

benpatterson:
@clytwynec for context, what would it be needed for in testing on the worker? (er, whats' the python dependency and what's it for?)

clytwynec:
@benpatterson scrapy requires cffi, which requires libffi (http://cffi.readthedocs.org/en/latest/installation.html#platform-specific-instructions)

benpatterson:
Gotcha.
:+1:

jzoldak:
👍

feanil:
@clytwynec @benpatterson @jzoldak this should have gone into the edxapp_common/defaults/main.yml file instead as I thought that was the whole point of that file.

feanil:
It would have prevented this: https://github.com/edx/configuration/pull/2981

benpatterson:
In our defense, it was not put into edxapp_common because it was needed for our testing framework, not running the app itself. That's why it was jenkins_worker and local_dev.

IIRC

PR #2941, opened 2016-04-07
Get studio domain into programs' cors whitelist

jimabramson:
@edx/devops review requested.  The intent is to have the correct CORS origin whitelist for Programs IDA in sandboxes, without manual editing.  It needs to have the public url for Studio in it, instead of being empty.

CORS is already fine in edx.org prod/stage so this is just for sandboxes.

@schenedx FYI

### DIFF
--- a/playbooks/roles/programs/defaults/main.yml
+++ b/playbooks/roles/programs/defaults/main.yml
@@ -53,7 +53,9 @@ PROGRAMS_PLATFORM_NAME: 'Your Platform Name Here'
 
 # CORS
 # See: https://github.com/ottoyiu/django-cors-headers/.
-PROGRAMS_CORS_ORIGIN_WHITELIST: []
+# the whitelist should always contain the public hostname for Studio in the given environment.
+PROGRAMS_CORS_ORIGIN_WHITELIST:
+  - '127.0.0.1:8001'
 
 PROGRAMS_DATA_DIR: '{{ COMMON_DATA_DIR }}/{{ programs_service_name }}'
 PROGRAMS_MEDIA_ROOT: '{{ PROGRAMS_DATA_DIR }}/media'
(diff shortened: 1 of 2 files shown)


playbooks/roles/programs/defaults/main.yml
```diff
@@ -53,7 +53,8 @@ PROGRAMS_PLATFORM_NAME: 'Your Platform Name Here'
 
 # CORS
 # See: https://github.com/ottoyiu/django-cors-headers/.
-PROGRAMS_CORS_ORIGIN_WHITELIST: []
+PROGRAMS_CORS_ORIGIN_WHITELIST:
+  - '{{ EDXAPP_CMS_BASE }}'
```
feanil: Because EDXAPP and PROGRAMS are separate independent roles, it is not guaranteed that this var is defined.  In the cases when it isn't it should have a default and that default should match the default value of that var.

jimabramson:
@feanil does this pass make more sense?

jimabramson:
@feanil I'm treading lightly here because I don't want to potentially mess up overrides for stage/prod that are presently working fine.  it's just for sandboxes.

when the approach here is validated I'm going to add an additional piece to update edxapp lms' cors whitelist with the same value - this is also an issue (also only on sandboxes).

feanil:
Yes, makes sense. In the discovery app Clinton called this `DISCOVERY_LMS_ROOT_URL`  I would like to follow that pattern for this so can we change the name to `PROGRAMS_CMS_ROOT_URL`?  If that sounds good: :+1: 

Note that this will need to be overridden in secure config for our environments so there should be a devops ticket to make this change.

feanil:
Actually, thinking about it a bit more, unless you need the URL_ROOT for other reasons, I don't think that var should exist.  The default whitelist should default to the localhost url for Studio and we should override the whole thing in secure config.

jimabramson:
@feanil updated.   does what i have here avoid the need for the secure config override?

feanil:
@jimabramson that looks good but I think you still need overrides for stage and production is secure.

You'll have to put a ticket in for that.

jimabramson:
@feanil OK thanks, filed DEVOPS-3991

feanil:
:+1:   I think I did not fully understand what it was that you were trying to achieve until you summarized it in that ticket.  This should be sufficient to fix the sandboxes, and I've verified it won't break the stage and production config.

feanil:
@jimabramson any reason you didn't delete the branch?

jimabramson:
@feanil no reason, thanks for the heads up

PR #2944, opened 2016-04-07
Disabling Boto/S3 querystring auth for edxapp

clintonb:
This removes the querystring at the end of the URLs for our content uploaded to S3.

ECOM-4081


peter-fogg:
:+1:

clintonb:
@edx/devops please review

PR #2945, opened 2016-04-08
Add cluster tag as a part of abbey.py

feanil:
The new code in the pipeline looks for the 'cluster' tag not the 'play' tag.  Add both for now,
we are working on a full replacement for abbey.py so this is just a shim until we can get to that.

@edx/devops


jibsheet:
:+1:

PR #2946, opened 2016-04-08
[OPS-1096] adding appneta role 

bilalahmad99:
still need to figure out the ruby/python instrumentation.


playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+- name: change config in case of non-static IPs
+  lineinfile:
+    dest: /etc/default/tracelyzer
+    regexp: ^STATIC_IP=
+    line: "STATIC_IP=NO"
+    state: present
+    create: yes
+  when: not USE_STATIC_CONFIG
+
+- name: set facts according to static config
```
e0d: why set a fact if the var USE_STATIC_CONFIG is available?

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+  set_fact:
+    apt_server: "apt-static.appneta.com"
+    config_server: "config-static.appneta.com"
+  when: USE_STATIC_CONFIG
+
+- name: install certificates
```
e0d: This also happens in common, does it need to happen here too?

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+- name: adding key to tracelyzer conf
+  lineinfile:
+    dest: /etc/tracelytics.conf
+    line: "tracelyzer.access_key={{ ACCESS_KEY }}"
+    state: present
+    create: yes
+
+- name: install common library and headers
+  apt: >
+    pkg={{ item }}
```
e0d: I think this should be name=, I believe that's now preferred to pkg.  @arbabnazar can you confirm?

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,121 @@
...
+#  apt:
+#    name: nginx
+#    state: present
+#
+#- name: Install nginx light and nginx extras
+#  apt: pkg={{ item }} state=present
+#  notify: restart nginx
+#  with_items: nginx_extra_pkgs
+#
+#- name: Install oboe
```
e0d: @edx/devops Thoughts on the best mechanism for installing the proper instrumentor?  

Firstly, I think the pattern used for new relic is no good.  We should be setting a python executable var well before we are in the template.  Currently each template repeats the logic to set the executable path.  Pulling this up would allow us to specify whichever instrumenting runtime wrapper we wanted easily.

Question two, should we use any of these approaches to install the correct wrappers?
- separate roles for python, ruby, etc.
- tags
- vars like APPNETA_PYTHON, APPNETA_RUBY?
- something else?

feanil: We've talked previously about maybe wrapping the base ex-service role in a
'django-ida' role that in turn is used by most of our IDAs.  This would
give us a central place to control things like which APM we use.

e0d: Actually, looking at the appneta mechanism for runtime configuration it's completely different than Newrelic.

They require that you import a library in your gunicorn entry point like so:

```
import oboeware.djangoware
```

Let's write the code for doing this for the [lms|cms]_gunicorn.py files in this PR.

For IDAs we really need to make the guncorn files generic and treat them like a function as you mention above.  Having a nearly identical template for each IDA is yuck, cookie cutter helps us make it bad faster in this case.

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: Done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: Done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: Done

playbooks/roles/edxapp/templates/cms_gunicorn.py.j2
```diff
@@ -4,6 +4,7 @@ gunicorn configuration file: http://docs.gunicorn.org/en/develop/configure.html
 {{ ansible_managed }}
 """
 import multiprocessing
+import oboeware.djangoware
```
e0d, line 8: This needs to be conditional up a variable being set, say, `EDXAPP_USE_APPNETA` or something.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+- name: start appneta-sequencer
+  service:
+    name: appneta-sequencer
+    state: started
+  when: install_sequencer
+
+- name: install oboe
+  pip:
+    name: oboe
+    virtualenv: "{{ appneta_venv_dir }}r"
```
e0d: This needs to be installed in the virtualenv of the application that we will be instrumenting.  So, in the case of edxapp, this needs to be in edxapp virtualenv so the import statement below will work.

playbooks/roles/edxapp/templates/lms_gunicorn.py.j2
```diff
@@ -4,6 +4,7 @@ gunicorn configuration file: http://docs.gunicorn.org/en/develop/configure.html
 {{ ansible_managed }}
 """
 import multiprocessing
+import oboeware.djangoware
```
e0d, line 8: Same as above, make conditional.

playbooks/roles/appneta/tasks/main.yml
```diff
@@ -0,0 +1,82 @@
...
+  apt:
+    name: appneta-sequencer
+    state: present
+    force: yes
+  when: install_sequencer
+
+- name: start appneta-sequencer
+  service:
+    name: appneta-sequencer
+    state: started
```
e0d, line 73: We need to ensure that it will be configured to start on restart of the machine too.

bilalahmad99, line 73: Done

playbooks/roles/edxapp/templates/lms_gunicorn.py.j2
bilalahmad99, line 8: done

playbooks/roles/appneta/tasks/main.yml
bilalahmad99: done

playbooks/roles/edxapp/templates/cms_gunicorn.py.j2
bilalahmad99, line 8: done

playbooks/roles/appneta/tasks/main.yml
arbabnazar: @bilalahmad99 I think typo, unnecessary  "r".

playbooks/edx-east/appneta.yml
```diff
@@ -0,0 +1,13 @@
+# When running, do pass ACCESS_KEY for appneta account as an argument to playbook e.g.
```
e0d: We should renames this playbook `edxapp-appneta.yml`

We also need to pull in the aws role and the nginx site relate code from the edxapp.yml play.

playbooks/roles/appneta/defaults/main.yml
```diff
@@ -0,0 +1,40 @@
...
+apt_server: "{{ (USE_STATIC_CONFIG == true) | ternary('apt-static.appneta.com','apt.appneta.com') }}"
+config_server: "{{ (USE_STATIC_CONFIG == true) | ternary('config-static.appneta.com','config.tv.appneta.com') }}"
+skip_certs: true
+ACCESS_KEY: ""
+install_sequencer: true
+appneta_apt_repo: "deb http://{{ apt_server }}/{{ ACCESS_KEY }} {{ ansible_distribution_release }} main"
+#
+# OS packages
+#
+nginx_extra_pkgs:
```
e0d: Do these packages get installed via the nginx role or this role?

e0d:
@feanil can you give another pass

playbooks/edx-east/appneta.yml
bilalahmad99: Done

playbooks/roles/appneta/defaults/main.yml
bilalahmad99: Actually they are optional in installation, either you install full nginx package or the lighter one according to your need. So removing this, as we are already installing full nginx package in its role.

edx-webhook:
Thanks for the pull request, @bilalahmad99! I've created [OSPR-2098](https://openedx.atlassian.net/browse/OSPR-2098) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

PR #2951, opened 2016-04-11
Change packages installed in go-agent.

doctoryes:
@feanil @macdiesel Removed tubular. Added Docker and AWS CLI.

For your review.

### DIFF
--- a/docker/build/go-agent/Dockerfile
+++ b/docker/build/go-agent/Dockerfile
@@ -32,6 +32,9 @@ RUN /bin/bash /tmp/docker/docker_install.sh
 # Install AWS command-line interface - for AWS operations in a go-agent task.
 RUN pip install awscli
 
+# Install Python virtualenv.
+RUN pip install virtualenv
+
 ADD docker/build/go-agent/files/go-agent-start.sh /etc/service/go-agent/run
 ADD docker/build/go-agent/files/go-agent-env-vars /etc/default/go-agent
 RUN update-java-alternatives -s java-7-oracle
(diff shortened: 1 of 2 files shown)


docker/build/go-agent/Dockerfile
```diff
@@ -18,17 +18,20 @@ WORKDIR /data
...
     python-pip
 
-# TODO: repalce this with a pip install command so we can version this properly
-RUN git clone https://github.com/edx/tubular.git /opt/tubular
-RUN pip install -r /opt/tubular/requirements.txt
-RUN cd /opt/tubular;python setup.py install
+# Install Docker - for Docker container building by a go-agent.
+RUN \
+  apt-get install -y curl && \
+  curl -fsSL https://get.docker.com/ | sh
```
feanil: Curling to bash makes me a bit wary?  Can we just do the exact things we need instead?

```
            $sh_c "apt-key adv --keyserver hkp://p80.pool.sks-keyservers.net:80 --recv-keys 58118E89F3A912897C070ADBF76221572C52609D"
            $sh_c "mkdir -p /etc/apt/sources.list.d"
            $sh_c "echo deb [arch=$(dpkg --print-architecture)] https://apt.dockerproject.org/repo ${lsb_dist}-${dist_version} ${repo} > /etc/apt/sources.list.d/docker.list"
            $sh_c 'sleep 3; apt-get update; apt-get install -y -q docker-engine'
```

I think this is all we need:
- Add the repo
- Install docker-engine

doctoryes: `curl -L http://mysketchysite.com/weird_script | sh`? What could go wrong? 😄 

How about instead I commit the Docker install script and use it during installation? That way, we don't have to figure out the important parts - and updating it is as simple as downloading a new version in the future.

feanil: That sounds good to me.

doctoryes:
@macdiesel @feanil This PR removes tubular from the go-agent Docker container and adds a couple of other things. Please review.

macdiesel:
👍

PR #2952, opened 2016-04-11
Added edxapp setting for AWS_S3_CUSTOM_DOMAIN

clintonb:
ECOM-4081


clintonb:
@feanil @rlucioni I'd love to get this into this release, if possible. Willing to do a patch, if necessary. Let me know which is preferred.

clintonb:
Related: https://openedx.atlassian.net/browse/DEVOPS-3982

rlucioni:
Following up here, too. The RC is already a monster and hard to manage as it is. I think a patch release would be the most appropriate thing at this point.

clintonb:
Note that this setting only applies to `django-storages`: https://github.com/jschneier/django-storages/blob/12ceed1dc822ce960e02176b6c7dce9f6e08a0e0/storages/backends/s3boto.py#L219

feanil:
:+1:

PR #2953, opened 2016-04-12
Adds rebuild to the default plugins

jibsheet:
I always miss this when we don't have it.

We turn it on manually in tools/admin jenkins' roles.  It isn't used in build jenkins, but that role passes an override of the plugin list anyway.  This would impact the jenkins_worker builds, which may be a reason not to do it (@benpatterson thoughts?  I'm running a packer rebuild and won't merge if it causes any issues https://build.testeng.edx.org/view/packer/job/build-packer-ami/1263/console)

### DIFF
--- a/playbooks/roles/jenkins_master/defaults/main.yml
+++ b/playbooks/roles/jenkins_master/defaults/main.yml
@@ -62,6 +62,7 @@ jenkins_plugins:
     - { name: "postbuild-task", version: "1.8" }
     - { name: "plain-credentials", version: "1.1" }
     - { name: "PrioritySorter", version: "2.9" }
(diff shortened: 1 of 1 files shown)


benpatterson:
Actually we've been talking about adding this to the build jenkins anyway. And, it shouldn't affect jenkins_worker builds, because that's a separate role anyway.

:+1: from me.

jibsheet:
@edx/devops Any objections?
@mulby this guarantees the new analytics jenkins will have rebuild by default also.

mulby:
:+1:

maxrothman:
:shipit:

PR #2954, opened 2016-04-12
Updated Elasticsearch settings for discovery service

clintonb:


clintonb:
@feanil this is the cause of the `refresh_course_metadata` failure.

clintonb:
Ticket to update secure settings: https://openedx.atlassian.net/browse/DEVOPS-3990

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -37,12 +37,8 @@ DISCOVERY_DATABASES:
     ATOMIC_REQUESTS: true
     CONN_MAX_AGE: 60
 
-DISCOVERY_ELASTICSEARCH_HOST: 'localhost'
-DISCOVERY_ES_INDEX: 'discovery'
-
-DISCOVERY_ELASTICSEARCH:
-  host: '{{ DISCOVERY_ELASTICSEARCH_HOST }}'
-  index: '{{ DISCOVERY_ES_INDEX }}'
+DISCOVERY_ELASTICSEARCH_URL: 'http://127.0.0.1:9200/'
```
feanil, line 43: You're specifying both a protocol and a port, how does this work with https?  Please put a comment with an example for that use case.

clintonb, line 43: ES + SSL is unfamiliar territory for me. I can link to docs (https://www.elastic.co/guide/en/shield/current/ssl-tls.html).

clintonb:
@feanil ultimately, I'd like to use [client nodes](https://www.elastic.co/guide/en/elasticsearch/reference/current/modules-node.html) and eliminate the need for configuring this altogether. I'm unsure if/how this works with AWS, and haven't had the time to investigate.

clintonb:
Clarification, we'd probably need to configure a cluster name at a minimum.

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -37,12 +37,11 @@ DISCOVERY_DATABASES:
...
 
-DISCOVERY_ELASTICSEARCH_HOST: 'localhost'
-DISCOVERY_ES_INDEX: 'discovery'
-
-DISCOVERY_ELASTICSEARCH:
-  host: '{{ DISCOVERY_ELASTICSEARCH_HOST }}'
-  index: '{{ DISCOVERY_ES_INDEX }}'
+# Using SSL? See https://www.elastic.co/guide/en/shield/current/ssl-tls.html.
+# Using AWS? Use the AWS-provided host (e.g. https://search-test-abc123.us-east-1.es.amazonaws.com/).
+# Assuing the port corresponds that of the protocol (e.g. http=80, https=443), there is no need to specify a port.
```
feanil: s/Assuing/Assuming/

clintonb: Fixed.

feanil:
:+1:

PR #2956, opened 2016-04-13
Add P3P headers to the IDAs

jibsheet:
make it a variable so that you can override it for your
installation.

Extends the OSPR work from https://github.com/edx/configuration/pull/2876 so that the header is set in other user-facing IDAs, and is overridable in our secure vars.


jibsheet:
@edx/devops

feanil:
:+1:

PR #2958, opened 2016-04-14
Remove boto connection default for analytics API elasticsearch config

mulby:
@dsjen @dan-f @fredsmith - I believe this will resolve the issues people are having with installing the analytics API.

I have not yet tested this with a run on devstack or anything.


playbooks/roles/analytics_api/defaults/main.yml
```diff
@@ -58,7 +58,7 @@ ANALYTICS_API_AUTH_TOKEN: 'put-your-api-token-here'
 ANALYTICS_API_ELASTICSEARCH_LEARNERS_HOST: 'localhost'
```
dsjen, line 58: It might make sense to set this to `null` too.  There's a certain code path that is his when the host is set, and it may make sense to now set this by default.

dsjen:
This change makes sense to me, although I honestly don't know much about how a playbook "plays" into the builds.  Does it fill in the settings with these default values basically on a build?  Do we also need to update edX specific builds with the removed settings?

dan-f:
LGTM. 👍 on @dsjen's question. We should sanity check on a devstack, if possible.

dsjen:
I kicked up devstack with this branch and it ran into an error:

```
django.db.utils.OperationalError: (1045, "Access denied for user 'api001'@'localhost' (using password: YES)")
```

I think we should try removing that localhost line too!

dsjen:
Oops, I changed `ANALYTICS_API_ELASTICSEARCH_LEARNERS_HOST` to null, updated config, and ran into the same error, which I now realize is probably unrelated.  I've reverted the change, but am unsure what's going on...

dan-f:
@dsjen I hit that issue as well, and I think @mulby helped me work it out. It had to do with changing the database user in the yaml file...

dsjen:
:+1:

dan-f:
👍

fredsmith:
👍

PR #2961, opened 2016-04-14
[WIP] OPS-1379 As an operator I want automation to deploy API gateways

arbabnazar:
@e0d can you please review the progress. Thanks

### DIFF
--- a/terraform/aws-api-gateway/main.tf
+++ b/terraform/aws-api-gateway/main.tf
@@ -1,0 +1,55 @@
+provider "aws" {
+    region = "${var.aws_region}"
+}
+
+resource "aws_api_gateway_rest_api" "api" {
+  name = "${var.api_name}"
+  description = "${var.api_description}"
+}
+
+resource "aws_api_gateway_resource" "res" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  parent_id = "${aws_api_gateway_rest_api.api.root_resource_id}"
+  path_part = "${var.resource_name}"
+}
+
+resource "aws_api_gateway_method" "resource_method" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  resource_id = "${aws_api_gateway_resource.res.id}"
+  http_method = "${var.http_method}"
+  authorization = "${var.authorization}"
+}
+
+resource "aws_api_gateway_integration" "method_request" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  resource_id = "${aws_api_gateway_resource.res.id}"
+  http_method = "${aws_api_gateway_method.resource_method.http_method}"
+  type = "HTTP"
+  uri = "${var.backend_api_url}"
+  integration_http_method = "${var.integration_http_method}"
+}
+
+resource "aws_api_gateway_method_response" "method_response" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  resource_id = "${aws_api_gateway_resource.res.id}"
+  http_method = "${aws_api_gateway_method.resource_method.http_method}"
+  status_code = "200"
+   response_models = {
+    "application/json" = "Empty"
+  }
+}
+
+resource "aws_api_gateway_integration_response" "integration_request" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  resource_id = "${aws_api_gateway_resource.res.id}"
+  http_method = "${aws_api_gateway_method.resource_method.http_method}"
+  status_code = "${aws_api_gateway_method_response.method_response.status_code}"
+  selection_pattern = "${var.api_selection_pattern}"
+}
+
+resource "aws_api_gateway_deployment" "api_gateway_deployment" {
+  depends_on = ["aws_api_gateway_integration.method_request"]
+
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  stage_name  = "${var.api_deployment_stage_name}"
+}
--- a/terraform/aws-api-gateway/variables.tf
+++ b/terraform/aws-api-gateway/variables.tf
@@ -1,0 +1,38 @@
+variable aws_region {
+	default = "us-east-1"
+}
+
+variable api_name {
+	description = "Name that you want to assign to your api"
+}
+
+variable api_description {
+	description = "Description that you want to give to your api"
+}
+
+variable resource_name {
+	description = "Name that you want to assign to your resource"
+}
+
+variable http_method {
+    default = "GET"
+}
+
+variable authorization {
+	default = "NONE"
+}
+
+variable backend_api_url {
+	description = "URL of your backend api"
+}
+
+variable integration_http_method {
+	default = "GET"
+}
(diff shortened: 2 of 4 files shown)


terraform/aws-api-gateway/variables.tf
```diff
@@ -0,0 +1,35 @@
+variable rds_password {
```
e0d: cut and paste?

terraform/aws-api-gateway/main.tf
```diff
@@ -0,0 +1,55 @@
...
+
+resource "aws_api_gateway_integration_response" "IntegrationRequest" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  resource_id = "${aws_api_gateway_resource.res.id}"
+  http_method = "${aws_api_gateway_method.resource_method.http_method}"
+  status_code = "${aws_api_gateway_method_response.MethodResponse.status_code}"
+  selection_pattern = "${var.api_selection_pattern}"
+}
+
+resource "aws_api_gateway_deployment" "test" {
```
e0d: what is "test" here?

terraform/aws-api-gateway/main.tf
```diff
@@ -0,0 +1,55 @@
...
+}
+
+resource "aws_api_gateway_method" "resource_method" {
+  rest_api_id = "${aws_api_gateway_rest_api.api.id}"
+  resource_id = "${aws_api_gateway_resource.res.id}"
+  http_method = "${var.http_method}"
+  authorization = "${var.authorization}"
+}
+
+resource "aws_api_gateway_integration" "MethodRequest" {
```
e0d: there's a mixture of snake case foo_bar and standing camel case, FooBar in the resource name, let's standardize and follow the common pattern.

arbabnazar:
@e0d please check the progress on cloudfront and route53. Thanks

terraform/cloudfront/main.tf
```diff
@@ -0,0 +1,74 @@
+provider "aws" {
+    region = "${var.aws_region}"
+}
+
+// Setup your S3 Bucket
+resource "aws_s3_bucket" "cdn_bucket" {
```
e0d, line 6: Unless we need to specify some origin when creating the distribution, this should not be necessary.  The origin in our case will be the API gateway itself.

terraform/cloudfront/main.tf
```diff
@@ -0,0 +1,74 @@
...
+
+// Setup the CloudFront Distribution
+resource "aws_cloudfront_distribution" "cloudfront_distribution" {
+  origin {
+    domain_name = "${var.bucket_name}.s3.amazonaws.com"
+    origin_id = "S3-${var.bucket_name}"
+    s3_origin_config {}
+  }
+  enabled = true
+  //aliases = ["mytestapi.com", "www.mytestapi.com"]
```
e0d, line 34: remove commented code.

terraform/cloudfront/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  price_class = "${var.price_class}"
+  default_cache_behavior {
+    allowed_methods = [ "DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT" ]
+    cached_methods = [ "GET", "HEAD" ]
+    target_origin_id = "S3-${var.bucket_name}"
```
e0d, line 39: This should by the API gateway.

terraform/cloudfront/main.tf
```diff
@@ -0,0 +1,74 @@
...
+    forwarded_values {
+      query_string = true
+      cookies {
+        forward = "none"
+      }
+    }
+    viewer_protocol_policy = "allow-all"
```
e0d, line 46: @efagin do we want to terminate SSL at CF?  Whose responsibility should wire-encryption be?

terraform/cloudfront/main.tf
```diff
@@ -0,0 +1,74 @@
...
+    min_ttl = 0
```
e0d, line 47: The TTLs should be vars and we'll want to set the min TTL to something other than 0.  For now, let's have it match the default_ttl.

terraform/cloudfront/main.tf
```diff
@@ -0,0 +1,74 @@
...
+  //aliases = ["mytestapi.com", "www.mytestapi.com"]
+  price_class = "${var.price_class}"
+  default_cache_behavior {
+    allowed_methods = [ "DELETE", "GET", "HEAD", "OPTIONS", "PATCH", "POST", "PUT" ]
+    cached_methods = [ "GET", "HEAD" ]
+    target_origin_id = "S3-${var.bucket_name}"
+    forwarded_values {
+      query_string = true
+      cookies {
+        forward = "none"
```
e0d, line 43: @efagin what's the appropriate cookie forwarding policy?

terraform/cloudfront/main.tf
efagin, line 46: @e0d no, SSL termination is still the responsibility of the origin. Also - why all the CloudFront code? API Gateway manages an internal CF distribution but that isn't exposed to us and we certainly don't have any control over it. You shouldn't have to touch CF to manage an API gateway - the only non-APIG service you should be messing with is Route53 for the one-time CNAME to the gateway-generated distribution ID. Let me know if you guys want a walkthrough.

terraform/aws-api-gateway/main.tf
```diff
@@ -0,0 +1,55 @@
+provider "aws" {
+    region = "${var.aws_region}"
+}
+
+resource "aws_api_gateway_rest_api" "api" {
+  name = "${var.api_name}"
+  description = "${var.api_description}"
+}
+
+resource "aws_api_gateway_resource" "res" {
```
efagin, line 10: I don't understand why we care about resources, methods, or integrations at all? As far as we're concerned these are API Gateway internals, provided we're using the ImportRestAPI.

terraform/cloudfront/main.tf
e0d, line 46: When we groomed this we added an AC to create a CF distribution.  Was not clear this was automatic, it was clear the gateway controlled some internal details.  We can remove this code.

terraform/cloudfront/main.tf
efagin, line 43: Again, not something CF should be managing at all. Individual origins can specify in their Swagger fragments if they want the Cookie header passed through (in full) or not. Individual cookie filters aren't supported, at least at the moment, it's the entire header value.

terraform/cloudfront/main.tf
efagin, line 46: BTW "origin" generally means some service-specific ELB, so what I mean to say is SSL termination is probably happening at the ELB layer (or lower) in general, and this gateway shouldn't change that policy. Especially if we have multiple gateways or access pathways.

efagin:
I honestly don't understand why the majority of this pull request is necessary - to summarize some of my inline comments, API gateway automatically manages its private CloudFront distribution and low-level objects like methods and integrations, and exposes higher-level APIs for us (ImportRestApi and CreateDomainName specifically) to manage those resources.

PR #2969, opened 2016-04-19
Updated DISCOVERY_OAUTH_URL_ROOT

clintonb:
Using EDXAPP_LMS_ROOT_URL instead of EDXAPP_LMS_ISSUER. This removes the need to define another EDXAPP_\* override in our secure variables.


clintonb:
@e0d please review
FYI @bderusha @peter-fogg 
@PaulWattenberger you are seeing 403 errors on stage because we are attempting to validate the access token against localhost instead of the actual auth server.

clintonb:
Fix confirmed on stage.

PR #2970, opened 2016-04-20
(WL-435): Add configuration variable DEFAULT_SITE_THEME for setting default site theme

saleem-latif:
Hi @arbabnazar , @mattdrayer 

This PR contains addition of a new variable `DEFAULT_SITE_THEME` that can be used to set a default site theme for openedx instance, Kindly, review the changes and let me know if changes look good.

Note: default value for `COMPREHENSIVE_THEME_DIR` also needs to be updated, but I think those default values are set somewhere other that 'edx/configuration'. @arbabnazar can you share your thoughts on this ?

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -555,6 +555,10 @@ EDXAPP_PROCTORING_BACKEND_PROVIDER:
 # Comprehensive Theming
 # Full path to the comprehensive theme directory
 EDXAPP_COMPREHENSIVE_THEME_DIR: ""
+
+# Name of the default site theme
+EDXAPP_DEFAULT_SITE_THEME: ""
+
 # Git repo for the comprehensive theme (if using a comprehensive theme
 # other than the ones bundled with edx/platform)
 EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO: ""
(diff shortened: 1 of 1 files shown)


saleem-latif:
@edx/devops -- FYI on this PR

arbabnazar:
what if it will be empty?

e0d:
@saleem-latif what is the behavior in the default case from the code perspective?

saleem-latif:
@e0d Currently, `DEFAULT_SITE_THEME` is not present in configurations, so [`DEFAULT_SITE_THEME`](https://github.com/edx/edx-platform/pull/12184/files#diff-82bda515879c96ab0641bfcbe58631c3R2890) defaults to `None` and no theme is applied.

And the value for `COMPREHENSIVE_THEME_DIR` is `/edx/app/edxapp/edx-platform/themes/edx.org` and after theming PR (#12184) is merged, this would cause errors and unexpected behavior.

When this PR is merged, `DEFAULT_SITE_THEME` will have "edx.org" as default value, so "edx.org" theme will be applied to new instances.

arbabnazar:
if we assign `unset` to `COMPREHENSIVE_THEME_DIR` in the jenkins then we didn't get any error and default theme has been applied. So what will be happened if we still apply the `unset` to `COMPREHENSIVE_THEME_DIR`?

saleem-latif:
if `unset` means that `COMPREHENSIVE_THEME_DIR` will not be added to lms.env.json and cms.env.json then default themes dir will be used and we will not get any errors. 
And if `unset` means empty string for `COMPREHENSIVE_THEME_DIR` then theming would be disabled.

saleem-latif:
@arbabnazar empty value of `DEFAULT_SITE_THEME` means no default theme and openedx theme would be used.

arbabnazar:
Can you please build sandbox with this branch and use the value of `COMPREHENSIVE_THEME_DIR` as `unset` so that we'll check the result.

saleem-latif:
@arbabnazar setting `COMPREHENSIVE_THEME_DIR` to `unset` causes an error.

The reason is the in `lms.env.json` following value is set `"COMPREHENSIVE_THEME_DIR": "unset"` shouldn't it be `"COMPREHENSIVE_THEME_DIR": null` or  `"COMPREHENSIVE_THEME_DIR": ""` ?

mattdrayer:
:+1: on adding this new settings parameter -- COMPREHENSIVE_THEME_DIR is not part of the change set, so we should be clear to merge the branch

saleem-latif:
@e0d Can you take a look at this PR, it needs to go before today's patch release regarding theming.

e0d:
:+1:

mattdrayer:
@e0d -- is it okay to revert this PR -- I'm not sure about the impact of doing so.

mattdrayer:
actually, i'm sorry, it's not this PR that needs to be reverted -- it's the change to COMPREHENSIVE_THEME_DIR -- sorry, getting my change sets mixed up

PR #2974, opened 2016-04-21
Neo4j community edition role

benpatterson:
This will install the community edition of neo4j and configure it
to run on particular ports, but listen to all IPs. The debian package
installs it as a service under 'neo4j-service'. See vars for
additional info on where the properties files are located.

### DIFF
--- a/playbooks/roles/neo4j/defaults/main.yml
+++ b/playbooks/roles/neo4j/defaults/main.yml
@@ -1,0 +1,26 @@
+---
+#
+#  Adds the latest stable package of neo4j community edition
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role neo4j
+#
+
+#
+# vars are namespaced with the module name.
+#
+neo4j_role_name: neo4j
+
+neo4j_gpg_key_url: https://debian.neo4j.org/neotechnology.gpg.key
+neo4j_apt_repository: "deb http://debian.neo4j.org/repo stable/"
+neo4j_server_config_file: "/etc/neo4j/neo4j-server.properties"
+neo4j_https_port: 8081  # default is 7473
+neo4j_http_port: 8080  # default is 7474
+
--- a/playbooks/roles/neo4j/defaults/main.yml
+++ b/playbooks/roles/neo4j/defaults/main.yml
@@ -17,10 +17,15 @@
 # vars are namespaced with the module name.
 #
 neo4j_role_name: neo4j
+neo4j_server_name: "coursegraph.edx.org"
 
 neo4j_gpg_key_url: https://debian.neo4j.org/neotechnology.gpg.key
 neo4j_apt_repository: "deb http://debian.neo4j.org/repo stable/"
 neo4j_server_config_file: "/etc/neo4j/neo4j-server.properties"
-neo4j_https_port: 8081  # default is 7473
-neo4j_http_port: 8080  # default is 7474
+neo4j_https_port: 7473  # default in package is 7473
+neo4j_http_port: 7474  # default in package is 7474
 
+# Properties file settings
+neo4j_address_settings_key: "org.neo4j.server.webserver.address"
+neo4j_https_settings_key: "org.neo4j.server.webserver.https.port"
+neo4j_http_settings_key: "org.neo4j.server.webserver.port"
--- a/playbooks/roles/neo4j/defaults/main.yml
+++ b/playbooks/roles/neo4j/defaults/main.yml
@@ -16,8 +16,7 @@
 #
 # vars are namespaced with the module name.
 #
-neo4j_role_name: neo4j
-neo4j_server_name: "coursegraph.edx.org"
+NEO4J_SERVER_NAME: "localhost"
 
 neo4j_gpg_key_url: https://debian.neo4j.org/neotechnology.gpg.key
 neo4j_apt_repository: "deb http://debian.neo4j.org/repo stable/"
--- a/playbooks/roles/neo4j/defaults/main.yml
+++ b/playbooks/roles/neo4j/defaults/main.yml
@@ -21,6 +21,7 @@ neo4j_server_name: "coursegraph.edx.org"
 
 neo4j_gpg_key_url: https://debian.neo4j.org/neotechnology.gpg.key
 neo4j_apt_repository: "deb http://debian.neo4j.org/repo stable/"
+neo4j_version: "3.0.3"
 neo4j_server_config_file: "/etc/neo4j/neo4j-server.properties"
 neo4j_https_port: 7473  # default in package is 7473
 neo4j_http_port: 7474  # default in package is 7474
--- a/playbooks/roles/neo4j/handlers/main.yml
+++ b/playbooks/roles/neo4j/handlers/main.yml
@@ -1,0 +1,4 @@
+---
+- name: restart neo4j
+  service: name=neo4j-service state=restarted
+
--- a/playbooks/roles/neo4j/handlers/main.yml
+++ b/playbooks/roles/neo4j/handlers/main.yml
@@ -1,4 +1,0 @@
----
-- name: restart neo4j
-  service: name=neo4j-service state=restarted
-
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -33,7 +33,7 @@
 
 - name: install neo4j
   apt:
-    name=neo4j
+    name=neo4j=2.3.4
     state=present
 
 - name: set to listen on all IPs
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -39,21 +39,21 @@
 - name: set to listen on all IPs
   lineinfile:
     dest="{{ neo4j_server_config_file }}"
-    regexp="org.neo4j.server.webserver.address="
-    line="org.neo4j.server.webserver.address=0.0.0.0"
+    regexp="{{ neo4j_address_settings_key }}="
+    line="{{ neo4j_address_settings_key }}=0.0.0.0"
     state=present
   notify: restart neo4j
 
 - name: set to listen on specific port for https
   lineinfile:
     dest="{{ neo4j_server_config_file }}"
-    regexp="org.neo4j.server.webserver.https.port="
-    line="org.neo4j.server.webserver.https.port={{ neo4j_https_port }}"
+    regexp="{{ neo4j_https_settings_key }}="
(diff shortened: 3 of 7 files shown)


benpatterson:
@adampalay @e0d

playbooks/roles/neo4j/defaults/main.yml
```diff
@@ -0,0 +1,26 @@
...
+#
+# vars are namespaced with the module name.
+#
+neo4j_role_name: neo4j
+
+neo4j_gpg_key_url: https://debian.neo4j.org/neotechnology.gpg.key
+neo4j_apt_repository: "deb http://debian.neo4j.org/repo stable/"
+neo4j_server_config_file: "/etc/neo4j/neo4j-server.properties"
+neo4j_https_port: 8081  # default is 7473
+neo4j_http_port: 8080  # default is 7474
```
adampalay: @benpatterson , any reason why not to use the default?

benpatterson: @adampalay I assume those ports will be firewalled on sandboxes. Though @e0d can probably recommend ports to us (whether or not it's the defaults).

benpatterson:
BTW this build is red because one of the docker shards hit the Travis timeout.

playbooks/roles/neo4j/defaults/main.yml
e0d: We should adjust the firewall rather than changing the ports I think.  I'd argue we should deploy to an actual IDA-lite cluster rather than a sandbox if we are operationalizing.

playbooks/roles/neo4j/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: restart neo4j
+  service: name=neo4j-service state=restarted
+
```
e0d: We're avoiding handlers entirely these days in favor of managing via tasks with tags.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+- name: install neo4j
+  apt:
+    name=neo4j
+    state=present
+
+- name: set to listen on all IPs
+  lineinfile:
+    dest="{{ neo4j_server_config_file }}"
+    regexp="org.neo4j.server.webserver.address="
+    line="org.neo4j.server.webserver.address=0.0.0.0"
```
e0d: should be a var

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+    state=present
+  notify: restart neo4j
+
+- name: set to listen on specific port for https
+  lineinfile:
+    dest="{{ neo4j_server_config_file }}"
+    regexp="org.neo4j.server.webserver.https.port="
+    line="org.neo4j.server.webserver.https.port={{ neo4j_https_port }}"
+  notify: restart neo4j
```
e0d: Should we put behind ngxinx so we can support SSL without needing to handle that for neo4j?  Would be easier than add keystore management to this role.

e0d:
Is there any story/need for authz/authn?

playbooks/roles/neo4j/defaults/main.yml
benpatterson: Ok. I am going to leave this config in there, but set it to the actual defaults if that is ok with everyone.

playbooks/roles/neo4j/handlers/main.yml
benpatterson: @e0d do you have an example for me handy? In this case, it's important for the service to be restarted after the setting adjustments are made.

benpatterson:
@e0d The auth side is up to you. I am seeing this living in our internal network (wherein someone needs to be in the office or VPN'd in). So, I don't have a story for auth, but would look to you for any sort of security requirements.

benpatterson:
Also, with auth, this plugin surfaces: https://github.com/neo4j-contrib/authentication-extension

benpatterson:
@adampalay The role installs now. I'm working on the nginx configuration. Might need assistance with certs.

@e0d is there a way I could test certs locally? If so, can you point out an example?

benpatterson:
@adampalay @e0d I set up certs manually/locally to verify that the nginx config works. So I'm feeling pretty good about it right now. 

Note that I had to override the nginx template directory when testing locally. When looking at roles that do similar things with nginx, I concluded that this must be happening when those are deployed in real life.

Ready for another look.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+    state=present
+
+- name: add neo4j apt repository
+  apt_repository:
+    repo="{{ neo4j_apt_repository }}"
+    state=present
+
+- name: install neo4j
+  apt:
+    name=neo4j=2.3.4
```
benpatterson: author's note, the latest neo4j (3.0) is a jump from where our coursegraph scripts are right now (and, also seems to have an issue recognizing when java8 is installed). So we are pinning it here.

benpatterson:
@edx/devops

Broadening the feedback pool.

Short story is we are operationalizing neo4j installs so we can put coursegraph onto a machine within our internal network. Because it'll be internal, we are descoping auth n down to the native neo4j auth strategy (i.e., username/password), and I don't think there is an auth z in neo4j, but haven't looked because, well, it's not in scope.

Feedback appreciated.

maxrothman:
@benpatterson your syntax is not correct for a bunch of your tasks. There are 2 syntaxes:

String syntax:

```
- name: foo
  bar: >
    one=two
    three=four
```

YAML syntax:

```
- name: foo
  bar:
    one: two
    three: four
```

Notice that in the former, the `>` character indicates a multiline string, so the resulting structure is `{name: 'foo', bar: 'one=two three=four'}`, whereas the latter uses a nested dict (`{name: 'foo', bar: {one: 'two', three: 'four'}}`

We usually prefer the YAML syntax.

benpatterson:
Thanks @maxrothman . Would you mind pointing out a specific one in my PR?

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+    state=present
+
+- name: set to listen on all IPs
+  lineinfile:
```
maxrothman, line 49: @benpatterson [My comment](https://github.com/edx/configuration/pull/2974#issuecomment-223979532) references all of these tasks

benpatterson:
och nvm I see what you mean...

benpatterson:
@maxrothman updated

benpatterson:
@adampalay @edx/devops ready for another pass.

playbooks/roles/neo4j/handlers/main.yml
e0d: @benpatterson it's the same code, just include it in the tasks file.  Add tags to allow skipping and running explicitly

```
tags:
    - manage
    - manage:start
```

e0d:
I'm :+1: after responding to my clarification of the handler comment.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
...
+    state: present
+
+- name: add neo4j apt repository
+  apt_repository:
+    repo: "{{ neo4j_apt_repository }}"
+    state: present
+
+- name: install neo4j
+  apt:
+    name: neo4j=2.3.4
```
adampalay: we'll want to use neo4j 3.0.3

http://debian.neo4j.org/?_ga=1.105760331.912593541.1467745138

benpatterson:
@fredsmith I'm going to assume you take this from here (or throw it away and construct something new).

If that assumption is not correct, let me know so we can figure out next steps.

playbooks/roles/neo4j/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
...
+#
+##
+# Defaults for role neo4j
+#
+
+#
+# vars are namespaced with the module name.
+#
+neo4j_role_name: neo4j
+neo4j_server_name: "coursegraph.edx.org"
```
maxrothman: This should probably not default to coursegraph. Instead, it should be required to pass it in from a playbook. That way, we don't make the initial assumption that this is the only way this role will be used.

maxrothman:
Why should neo4j be listening on all ports?

playbooks/roles/neo4j/templates/edx/app/nginx/sites-available/coursegraph.j2
```diff
@@ -0,0 +1,53 @@
...
+  {% set default_site = "" %}
+{% endif %}
+
+server {
+#   server_name {{ neo4j_server_name }};
+  server_name localhost;
+
+  {% if NGINX_ENABLE_SSL %}
+
+  listen 443 ssl;
```
maxrothman, line 17: Shouldn't this be `neo4j_https_port`?

playbooks/roles/neo4j/templates/edx/app/nginx/sites-available/coursegraph.j2
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  location @proxy_to_app {
+    proxy_set_header X-Forwarded-Proto $http_x_forwarded_proto;
+    proxy_set_header X-Forwarded-Port $http_x_forwarded_port;
+    proxy_set_header X-Forwarded-For $http_x_forwarded_for;
+    proxy_set_header Host $http_host;
+
+    proxy_redirect off;
+    proxy_pass http://127.0.0.1:7473;
```
maxrothman: This should also use the appropriate port variable

playbooks/roles/neo4j/defaults/main.yml
```diff
@@ -0,0 +1,32 @@
...
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role neo4j
+#
+
+#
+# vars are namespaced with the module name.
+#
+neo4j_role_name: neo4j
```
maxrothman: This variable does not appear to be used anywhere

playbooks/roles/neo4j/templates/edx/app/nginx/sites-available/coursegraph.j2
adampalay: @fredsmith , I should add that the TCP port is different from the HTTPS one. I think that one defaults to 7687: http://py2neo.org/v3/database.html#the-graph. So we'd want to expose that port too.

playbooks/roles/neo4j/templates/edx/app/nginx/sites-available/coursegraph.j2
fredsmith, line 17: no, nginx forwards to neo4j, which is listening on neo4j_https_port

playbooks/roles/neo4j/templates/edx/app/nginx/sites-available/coursegraph.j2
fredsmith: @adampalay expose how?  directly, or via http/https?

adampalay: I'm guessing directly. I'm not really sure though. How would I figure that out

PR #2975, opened 2016-04-21
Pin things installed by devstack and fullstack

nedbat:
Trying to prevent future version slippage from breaking installations.


nedbat:
@edx/devops are we ok with specific version pinning of things like supervisor?

maxrothman:
Are those all the latest versions? 👍

nedbat:
Those are the latest versions, as pypi reports them.

maxrothman:
:shipit:

PR #2979, opened 2016-04-22
Add asymmetric key JWT variables for course-discovery and LMS

bderusha:
Please review @clintonb


playbooks/roles/discovery/defaults/main.yml
```diff
@@ -68,6 +68,24 @@ DISCOVERY_SOCIAL_AUTH_EDX_OIDC_KEY : 'discovery-key'
 DISCOVERY_SOCIAL_AUTH_EDX_OIDC_SECRET : 'discovery-secret'
 DISCOVERY_SOCIAL_AUTH_REDIRECT_IS_HTTPS: false
 
+# This should be the same key as the EDXAPP_JWT_PUBLIC_KEY
+DISCOVERY_JWT_SECRET_KEY: 'replace-with-lms-rsa-public-key'
```
clintonb: You can reference the LMS value. You'll want to make sure said value is defined for all plays in the secure configuration, similar to `EDXAPP_LMS_ROOT_URL`.

`{{ EDXAPP_JWT_PUBLIC_KEY | default("replace-me") }}`

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -570,6 +570,10 @@ EDXAPP_COMPREHENSIVE_THEME_VERSION: ""
 EDXAPP_SOCIAL_AUTH_SAML_SP_PRIVATE_KEY: ''
 EDXAPP_SOCIAL_AUTH_SAML_SP_PUBLIC_CERT: ''
 
+# RSA KEY PAIR for JWT signing/validation
+EDXAPP_JWT_PUBLIC_KEY: ''
```
clintonb, line 574: Are these sane defaults? If someone runs this without an override, will their installation be insecure?

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -68,6 +68,24 @@ DISCOVERY_SOCIAL_AUTH_EDX_OIDC_KEY : 'discovery-key'
...
+DISCOVERY_JWT_ALGORITHM: 'RS512'
+DISCOVERY_JWT_VERIFY_EXPIRATION: true
+DISCOVERY_JWT_DECODE_HANDLER: 'edx_rest_framework_extensions.utils.jwt_decode_handler'
+# This should be the same as the EDXAPP_LMS_ISSUER
+DISCOVERY_JWT_ISSUER: DISCOVERY_OAUTH_URL_ROOT
+DISCOVERY_JWT_AUDIENCE: 'course-discovery'
```
clintonb: This needs to be more global value that applies to all IDAs.

clintonb:
These changes need to be made for all of the IDAs; otherwise, they will break. We will also need to coordinate a rollout of every IDA, with the expectation of downtime (since we have not implemented OIDC discovery).

playbooks/roles/edxapp/defaults/main.yml
bderusha, line 574: No, it just wont work since `''` is not a valid public-private pair

bderusha, line 574: Not insecure, just broken.  Cannot use asymmetric signing without this set and `''` is not a valid public-private pair.  I used the Public/Private pair above for `SOCIAL_AUTH_SAML_SP_PRIVATE_KEY` as my reference point for how to default this.

clintonb, line 574: Please add a comment explaining this. We'll want to figure out a way to either hardcode or (preferably) generate this information for sandboxes in `ansible-provision.sh`.

bderusha:
Closing until actively in progress

PR #2981, opened 2016-04-22
Builds are failling because ffi.h is not installed

jibsheet:
We're not yet sure what changed, but something between Friday morning's
patchset release and this afternoon's rc cut.

@ormsbee
@edx/devops

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -1069,6 +1069,8 @@ edxapp_debian_pkgs:
   - ntp
   # matplotlib needs libfreetype6-dev
   - libfreetype6-dev
+  # cffi needs libffi-dev
+  - libffi-dev
 
 # Deploy Specific Vars
 edxapp_lms_variant: lms


maxrothman:
👍

feanil:
This should go in this file: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp_common/defaults/main.yml  so that jenkins workers can also pick it up.  @benpatterson  how did tests pass without the jenkins workers having this requirement?

jibsheet:
@feanil jenkins_workers/defaults/main.yml declares libffi-dev
Oh, I wonder when that was added....

feanil:
yea, whoever did that, did it wrong...

jibsheet:
@feanil so, long story short, it's for a new set of accessibility tests, so it was added to workers and devstacks, but not platform.  However, obviously we need it to build the platform.  We're going to rejigger this to be in edxapp_common and then devstack/jenkins_worker and edxapp will be covered.

just talked to @clytwynec

benpatterson:
What needs libffi on platform?

We put it into jenkins_worker and devstack because it was used for validation only. AFAIK it was not needed for running the edxapp itself. That's why we took that approach FWIW.

clytwynec:
Yea, sorry for this!  The thinking was that the dependency was actually only needed for testing, so I was avoiding adding a new dependency to production that wasn't needed.

jibsheet:
@benpatterson platform now installs scrapy, which needs libffi-dev and there's no difference between what's installed for testing and what is installed during platform builds.  This was causing the platform to fail to build.

benpatterson:
ahhh and perhaps this is chalked up to: the need to separate test requirements from prod?

benpatterson:
I'm sorry I didn't think of that either :(

jibsheet:
@benpatterson I feel like that's a long conversation, but yes.

@ormsbee Here's the root cause of the build failures.  It's clear I need to move some deps around to correct final destination on Monday.

jibsheet:
@benpatterson @clytwynec Feanil pointed out requirements/edx/development.txt exists and is not installed everywhere.  It may be time for testing.txt that only gets pulled in on devstacks and jenkins workers.  I'll be making a new PR for now to collect the 3 libffi-dev requirements into the one common place for them.

benpatterson:
Ok @jibsheet . Yeah we have put in cycles to figure out what could be pulled out of base.txt and put into a new file, and it always ends up having a scope balloon. Getting started with a known quantity and at least putting that into a separate file sounds great to me. We'll want to be sure that we run development.txt as part of builds though, before we merge the change in ansible, or else we'll have the opposite problem (i.e., not in production, and not in builds, either).

jibsheet:
Yeah @benpatterson - we'll need edx-local or other similar scaffolding for devstack and jenkins workers.  I think development.txt is explicitly not installed, it's only if you want extreme development techniques provided by John Eskew.

I think a specific testing.txt that starts with known quantities that we can add to over time is the medium term goal.

PR #2982, opened 2016-04-22
Add libffi-dev to discovery

bderusha:
FYI @clintonb @jibsheet

### DIFF
--- a/playbooks/roles/discovery/defaults/main.yml
+++ b/playbooks/roles/discovery/defaults/main.yml
@@ -9,7 +9,7 @@
 #
 ##
 # Defaults for role discovery
-# 
+#
 DISCOVERY_GIT_IDENTITY: !!null
 
 # depends upon Newrelic being enabled via COMMON_ENABLE_NEWRELIC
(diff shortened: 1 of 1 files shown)


jibsheet:
What changed to add the dep?

clintonb:
👍

clintonb:
@bderusha this is needed for all IDAs, and the template.

bderusha:
@jibsheet required by the `cryptography` package, which is required by `pyJWT` to use asymmetric key signing for JWTs

bderusha:
@clintonb it will be for them to start using asymmetric keys, but I think we should wait until we validate this with course discovery before we add the dependency to all IDAs

clintonb:
LMS is being updated to issue JWTs as access tokens. Those tokens won't be able to be validated without updating the IDAs.

jibsheet:
I'll let you folks decide which IDAs get it, but I'd like to request that the commit be amended to explain why libffi-dev is installed (either in the commit message or in a comment).
After that I'll thumb, I'd like to get at least this merged since it's causing Travis to fail for all Pull Requests against configuration right now.

clintonb:
@bderusha is sick today. I'll add the comment and get this out of the way. We'll follow-up with changes to the other IDAs later.

clintonb:
@jibsheet comment added.

jibsheet:
👍

PR #2985, opened 2016-04-26
go user may now execute the pip command as root.

macdiesel:
@doctoryes @feanil PTAL. This will allow the go user on the go-agent containers to run pip as root.


doctoryes:
👍

PR #2990, opened 2016-04-28
MAYN-198: Enhance ansible provisioning script to support edx/edx-themes

saleem-latif:
Hi @mattdrayer 

Kindly take a look at the change set, More info about the change set can be found at [MAYN-198](https://openedx.atlassian.net/browse/MAYN-198).


playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -102,6 +102,8 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
 # Theming
 ECOMMERCE_PLATFORM_NAME: 'Your Platform Name Here'
 ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
+THEME_VERSION: 'master'
+ECOMMERCE_GIT_IDENTITY: !!null
```
feanil: This variable is already defined in this file at the top.

playbooks/roles/ecommerce/meta/main.yml
```diff
@@ -17,6 +17,7 @@ dependencies:
     edx_service_name: "{{ ecommerce_service_name }}"
     edx_service_config: "{{ ECOMMERCE_SERVICE_CONFIG }}"
     edx_service_repos: "{{ ECOMMERCE_REPOS }}"
+    ed_themes_repo: "{{ EDX_THEMES_REPOS }}"
```
feanil: This is spelled wrong.

Also the service role knows nothing about theme repos and most services don't have theme repos.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -102,6 +102,8 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
 # Theming
 ECOMMERCE_PLATFORM_NAME: 'Your Platform Name Here'
 ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
+THEME_VERSION: 'master'
```
mattdrayer: For edx/edx-microsite we deploy from the 'release' branch -- I am wondering if we should follow-suit with edx/edx-themes, or if we want to switch to a deployable-master model.  I prefer deploying from master, but I would like to see how @douglashall feels about it.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -202,6 +217,7 @@ ecommerce_environment:
   DJANGO_SETTINGS_MODULE: "{{ ECOMMERCE_DJANGO_SETTINGS_MODULE }}"
   ECOMMERCE_CFG: "{{ COMMON_CFG_DIR }}/{{ ecommerce_service_name }}.yml"
   PATH: "{{ ecommerce_venv_dir }}/bin:{{ ansible_env.PATH }}"
+  COMPREHENSIVE_THEME_DIR: "{{ ecommerce_themes_dir }}"
```
mattdrayer: Need to consider the multi-directory aspect that has been discussed recently -- we are going to have public (open source) themes as well as private themes and need to be able to support both.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -168,6 +170,16 @@ ECOMMERCE_REPOS:
     SSH_KEY: "{{ ECOMMERCE_GIT_IDENTITY }}"
 
 
+EDX_THEMES_REPOS:
+  - PROTOCOL: "{{ COMMON_GIT_PROTOCOL }}"
+    DOMAIN: "{{ COMMON_GIT_MIRROR }}"
+    PATH: "{{ COMMON_GIT_PATH }}"
+    REPO: edx-themes.git
+    VERSION: "{{ THEME_VERSION }}"
+    DESTINATION: "{{ themes_code_dir }}"
```
mattdrayer: Is it okay to reference themes_code_dir here when it is not defined until line 197?

playbooks/roles/ecommerce/meta/main.yml
mattdrayer: @saleem-latif can you provide some info on what the intent is/was for adding this line here?  Maybe there's a better way to accomplish the desired outcome?

mattdrayer:
WIP is moving along -- thanks much for sharing your work so early -- @feanil we need to identify a good spot on the server for the edx-themes clone to live.  We are thinking in "/edx/etc" but that is sort of an arbitrary decision.  If you have a better location we might use, please let us know.

playbooks/roles/ecommerce/defaults/main.yml
douglashall: I prefer having an integration branch that won't get deployed by default. This allows us to have a more manageable manual QA process which is important for edx-microsite and this repo since we don't have much automated testing around it.

mattdrayer: Good point -- so then we'll plan to deploy edx/edx-themes from a 'release' branch, much like edx/edx-microsite.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -182,6 +195,9 @@ ecommerce_home: "{{ COMMON_APP_DIR }}/{{ ecommerce_service_name }}"
 ecommerce_code_dir: "{{ ecommerce_home }}/{{ ecommerce_service_name }}"
 ecommerce_venv_dir: "{{ ecommerce_home }}/venvs/{{ ecommerce_service_name }}"
 
+themes_code_dir: "{{ COMMON_CFG_DIR }}/edx-themes"
+ecommerce_themes_dir: "{{ themes_code_dir }}/ecommerce"
```
mattdrayer: I would rename this to "ecommerce_themes_code_dir" to maintain consistency with the parent variable.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -103,6 +103,14 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
 ECOMMERCE_PLATFORM_NAME: 'Your Platform Name Here'
 ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
 
+ECOMMERCE_COMPREHENSIVE_THEME_DIR: "{{ ecommerce_themes_dir }}"
+ECOMMERCE_ENABLE_COMPREHENSIVE_THEMING: true
+ECOMMERCE_DEFAULT_SITE_THEME: "edx.org"
```
mattdrayer: This value should come from the secure configuration -- we need to avoid putting "edx.org" into our open source code.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -103,6 +103,14 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
...
+
+ECOMMERCE_COMPREHENSIVE_THEME_DESTINATION_DIR: "{{ themes_code_dir }}"
+ECOMMERCE_COMPREHENSIVE_THEME_SOURCE_REPO: "https://github.com/edx/edx-themes.git"
```
mattdrayer: See my comment @ https://openedx.atlassian.net/browse/DEVOPS-4114 -- i think what we want here is something more along the lines of the ECOMMERCE_REPOS block further down.  Also we should be clear about which parameters are for general management of edx-themes code, and which parameters are used to control the comprehensive theming feature.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -103,6 +103,14 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
...
+ECOMMERCE_COMPREHENSIVE_THEME_VERSION: "master"
```
mattdrayer: I believe our intention is to work off of a release branch, so we should not default to master here -- also this value should probably come from the secure configuration, along with the source repo value above.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -182,6 +195,9 @@ ecommerce_home: "{{ COMMON_APP_DIR }}/{{ ecommerce_service_name }}"
 ecommerce_code_dir: "{{ ecommerce_home }}/{{ ecommerce_service_name }}"
 ecommerce_venv_dir: "{{ ecommerce_home }}/venvs/{{ ecommerce_service_name }}"
 
+themes_code_dir: "{{ COMMON_CFG_DIR }}/edx-themes"
```
mattdrayer: The name of this subdirectory should probably be a variable as well

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -35,6 +35,22 @@
   sudo_user: "{{ ecommerce_user }}"
   with_items: ecommerce_requirements
 
+# To enable Comprehensive theming, point
+# ECOMMERCE_COMPREHENSIVE_THEME_SOURCE_REPO to your theme repo, and set
+# ECOMMERCE_COMPREHENSIVE_THEME_DIR to the directory you want to check
+# out to.
+- name: checkout comprehensive theme
```
mattdrayer: We should be more clear about what is being checked out here -- we are not checking out a single comprehensive theme, we are checking out the ecommerce service's themes repository

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -35,6 +35,22 @@
...
+  git_2_0_1: >
+    dest={{ ECOMMERCE_COMPREHENSIVE_THEME_DESTINATION_DIR }}
```
mattdrayer: I would use THEMES here instead of COMPREHENSIVE_THEME

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -35,6 +35,22 @@
   sudo_user: "{{ ecommerce_user }}"
   with_items: ecommerce_requirements
 
+# To enable Comprehensive theming, point
+# ECOMMERCE_COMPREHENSIVE_THEME_SOURCE_REPO to your theme repo, and set
```
mattdrayer: Per some discussion with @fredsmith, we only need to namespace variables to ECOMMERCE which have to do with the actual ecommerce service -- for general themes-related variables we could instead prefix with THEMES

playbooks/roles/ecommerce/defaults/main.yml
saleem-latif: @mattdrayer I kept it "master" as other repos in configuration default to "master" ([here](https://github.com/edx/configuration/blob/master/playbooks/roles/ecommerce/defaults/main.yml#L37)) and are overridden via secure configuration.

saleem-latif:
@mattdrayer I have incorporated your feedback, kindly take a look.

mattdrayer:
@saleem-latif, this PR is looking good to me now -- the CI failure appears to be related to an insights issue and not ecommerce, so I think we're okay on that front.  So it's a 👍 from me -- @fredsmith are you the right person to handle the second review?

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -103,6 +105,14 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
 ECOMMERCE_PLATFORM_NAME: 'Your Platform Name Here'
 ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
 
+ECOMMERCE_COMPREHENSIVE_THEME_DIR: "{{ ecommerce_themes_code_dir }}"
+ECOMMERCE_ENABLE_COMPREHENSIVE_THEMING: true
```
feanil: Default should be false so we don't break the open-source community.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -103,6 +105,14 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
...
+ECOMMERCE_DEFAULT_SITE_THEME: ""
+
+ECOMMERCE_COMPREHENSIVE_THEME_DESTINATION_DIR: "{{ themes_code_dir }}"
+ECOMMERCE_COMPREHENSIVE_THEME_SOURCE_REPO: "https://github.com/edx/edx-themes.git"
```
feanil: This is a private repo, we should not have it in this config.

playbooks/roles/ecommerce/meta/main.yml
```diff
@@ -22,4 +22,13 @@ dependencies:
     edx_service_packages:
       debian: "{{ ecommerce_debian_pkgs }}"
       redhat: "{{ ecommerce_redhat_pkgs }}"
+  - role: edx_service
```
feanil: Is themes a separate django app?  If not then this should not be how we pull it in.   If you need to pull in multiple repos, you can do so with the existing ecommerce repos list by adding a second repo there.  However consider that this second repo is optional so you need a way to not provide this info for open source contributors.

mattdrayer: edx/edx-themes is a private collection of Django templates and static assets that we would like use to override the default look and feel of the ecommerce service.  So it's not really a Django app, so to speak.  The ecommerce service will check the local edx-themes directory (ie, wherever the repo is cloned) to see if a particular override exists for the request site/page.

Is there an example of conditional logic for playbooks that we can review?

playbooks/roles/ecommerce/defaults/main.yml
mattdrayer: We will change this value to be a parameter instead

playbooks/roles/ecommerce/defaults/main.yml
mattdrayer: Agreed -- is there something we need to do in order to ensure the "true" value is set for our own sandboxes and stage/prod installations?

playbooks/roles/ecommerce/meta/main.yml
saleem-latif: @feanil , @mattdrayer How about creating a separate role for edx-themes, with configurable parameters like repo, protocol etc. And later this role can be a conditional dependency for other roles (like edxapp, ecommerce etc.) ?
e.g.

```
- dependencies:
  - role: theming
    when: do_setup_theming == yes
```

This would keep theming related config separate from other roles, and give other openedx users the opportunity to enable/disable theming or use other private or public repo for themes.

mattdrayer: This seems like a good plan to me, and looks like a pretty simple solution -- but I'll need to defer to @feanil on the actual implementation

playbooks/roles/themes/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+themes_service_name: "edx-themes"
+themes_user: "{{ themes_service_name }}"
+themes_home: "{{ COMMON_CFG_DIR }}/{{ themes_service_name }}"
+themes_code_dir: "{{ themes_home }}/{{ themes_service_name }}"
+
+THEMES_GIT_IDENTITY: !!null
+THEMES_GIT_PROTOCOL: 'ssh'
+THEMES_GIT_MIRROR: "{{ COMMON_GIT_MIRROR }}"
+THEMES_GIT_PATH: "{{ COMMON_GIT_PATH }}"
+THEMES_REPO: "edx-themes.git"
```
mattdrayer: @feanil -- should this value also come from a secure setting?

mattdrayer: I'm not an expert with ansible but this latest version seems much more flexible for open source operation and on-line with @feanil's earlier request -- I had one question about the edx-themes.git reference but otherwise this LGTM

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -103,6 +103,12 @@ ECOMMERCE_PAYMENT_PROCESSOR_CONFIG:
 ECOMMERCE_PLATFORM_NAME: 'Your Platform Name Here'
 ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
 
+ECOMMERCE_COMPREHENSIVE_THEME_DIR: !!null
+ECOMMERCE_ENABLE_COMPREHENSIVE_THEMING: false
+ECOMMERCE_DEFAULT_SITE_THEME: !!null
+
+ECOMMERCE_DO_SETUP_THEMES: false
```
feanil: This should map to the the key in the app. Change this to 'ECOMMERCE_ENABLE_COMPREHENSIVE_THEMING'

feanil: Actually wait, this is a completely different var, is there any need for this?  Why not use the same var?

playbooks/roles/themes/defaults/main.yml
feanil: Is there a sample open-source repo we can put here instead?  So people can see how it's done and the role works by default.

playbooks/roles/themes/defaults/main.yml
```diff
@@ -0,0 +1,33 @@
...
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role themes
+#
+themes_service_name: "edx-themes"
```
feanil: The service name should match the role name.

playbooks/roles/themes/defaults/main.yml
mattdrayer: @feanil how should we go about adding a new public sample themes repo?  DevOps ticket?

feanil:
`playbooks/roles/themes/tasks/main.yml` seems like it was copied from the edx-service playbook, and is more generic than just themes.  Can we make a generic role that is used by both edx-service and here?

playbooks/roles/themes/defaults/main.yml
mattdrayer: OKAY!  We now have a new public repo located at https://github.com/edx/sample-themes which we can use in this PR and eliminate some of the conditional complexity.

saleem-latif:
@feanil I have incorporated your feedback changes, kindly take a look. And related your final comment, I think, given the current time frame, it would be better to create a generic task to be used by both edx-service and here in a separate PR.

fredsmith:
LGTM - it looks like we'll need to set the following variables in order to deploy this:  THEMES_REPO, THEMES_GIT_IDENTITY.

Will the django apps that use theming be able to pull things out of /edx/app/themes/, or do we need to teach nginx about any additional paths?

feanil:
It looks like there isn't a generic role yet, that both themes and edx-service would use.

feanil:
@saleem-latif is there a ticket for the follow on work somewhere?  Can you link it here?

jibsheet:
@saleem-latif @mattdrayer Can we make sure there is a sol ticket for further refactoring other theme paths to use this role (and possible to refactor away the "check out a git repo" code into yet another small role).  We don't want the theming implementations to drift in a weird hard to debug way.

Please mention that ticket here before merging.

mattdrayer:
@feanil @jibsheet @saleem-latif I have logged Fahrenheit WL-451 (https://openedx.atlassian.net/browse/WL-451) to track the tech debt cleanup -- we will address it in our new sprint which kicks off today.

PR #2991, opened 2016-04-28
mattdrayer/add-default-theme-setting: Update ecommerce playbook

mattdrayer:
@clintonb @jibsheet -- I think this is where the plumbing needs to be, actually, versus in ecommerce/settings/base.py


jibsheet:
You need to set ECOMMERCE_DEFAULT_SITE_THEME to something as a sane default 
Further up in the same file with the other initial defaults.

clintonb:
If the application defaults to `None`, why do we need this at all?

jibsheet:
@clintonb He asked us to override it in the secure config, which needs pluming in configuration to copy the value

clintonb:
Got it.

clintonb:
👍

mattdrayer:
Added default value for the setting

jibsheet:
👍

mattdrayer:
Failing checks are due to timeouts and an insights npm failure -- unrelated to the change introduced by this PR -- going ahead with the merge now.

PR #2997, opened 2016-05-01
Restoring this role for possible use with tools automation

e0d:
@edx/devops The PR restores a role I wrote ages ago and that was removed as we were not actively using it.  I think it's directionally a good pattern for allowing jobs on a shared Jenkins box to run remote automation, but apply least privilege and give control over application to the teams.

Even if we decide it's not a a good pattern, we can kick off the conversation because we'll need something like this for opening tools automation up.

@rlucioni FYI

### DIFF
--- a/playbooks/roles/automated/defaults/main.yml
+++ b/playbooks/roles/automated/defaults/main.yml
@@ -1,0 +1,24 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+# Vars for role automated
+#
+
+automated_role_name: automated
+AUTOMATED_USER: "changeme"
+automated_sudoers_template: "roles/automated/templates/99-automated.j2"
+
+#
+# OS packages
+#
+
+automated_debian_pkgs: []
+
+automated_redhat_pkgs: []
--- a/playbooks/roles/automated/tasks/main.yml
+++ b/playbooks/roles/automated/tasks/main.yml
@@ -1,0 +1,92 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://github.com/edx/configuration/wiki
+# code style: https://github.com/edx/configuration/wiki/Ansible-Coding-Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+# Tasks for role automated
+#
+# Overview:
+#
+# This role is included as a dependency by other roles which provide
+# automated jobs.  Automation occurs over ssh.  The automator user is
+# is allowed to run explicitly listed commands via sudo.
+#
+# Dependencies:
+#
+# This role depends upon variables provided by an including role
+# via the my_role/meta/main.yml file.  Includes take the following forms:
+#
+# For example, the edxapp role might designate that ecom and analytics users
+# are allowed to run specific management commands on edxapp instancs.
+#
+# EDXAPP_AUTOMATED_USERS:
+#   ecom:
+#     sudo_commands:
+#       - command: "/edx/app/edxapp/venvs/edxapp/bin/python /edx/app/edxapp/edx-platform/manage.py lms migrate --list --settings=aws"
+#         sudo_user: "edxapp"
+#       - command: "/edx/app/edxapp/venvs/edxapp/bin/python /edx/app/edxapp/edx-platform/manage.py cms migrate --list --settings=aws"
+#         sudo_user: "edxapp"
+#     authorized_keys:
+#       - 'ssh-rsa <REDACTED> ecom+admin@example.com'
+#       - 'ssh-rsa <REDACTED> ecom+devel@example.com'
+#   analytics:
+#     sudo_commands:
+#       - command: "/edx/app/edxapp/venvs/edxapp/bin/python /edx/app/edxapp/edx-platform/manage.py lms help --settings=aws"
+#         sudo_user: "edxapp"
+#     authorized_keys: ['ssh-rsa <REDACTED> analytics@example.com']
+#
+# The play for the role enabling automation should include the role like so
+#
+#  - role: automated
+#    AUTOMATED_USERS: "{{ EDXAPP_AUTOMATED_USERS | default({}) }}"
+#
+
+- name: Create automated user
+  user:
+    name: "{{ item.key }}"
+    state: present
+    shell: "/bin/bash"
+    createhome: yes
+  with_dict: "{{ AUTOMATED_USERS }}"
+
+- name: Create sudoers file from template
+  template:
+    dest: "/etc/sudoers.d/99-{{ item.key }}"
+    src: "{{ automated_sudoers_template }}"
+    owner: "root"
+    group: "root"
+    mode: "0440"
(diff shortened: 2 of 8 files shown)


rlucioni:
Thanks for the heads-up @e0d. Please let me know if we'll need to make changes to the management command job DSL to take advantage of this role.

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+    path={{ automated_home }}/bin state=directory mode=0750
+    owner="root" group={{ automated_user }}
+
+- name: re-write .profile
+  copy:
+    src=home/automator/.profile
+    dest={{ automated_home }}/.profile
+    owner="root"
+    group={{ automated_user }}
+    mode="0744"
```
jibsheet: Why a different mode?  Earlier we made this file 0640

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+- name: re-write .bashrc
+  copy:
+    src=home/automator/.bashrc
+    dest={{ automated_home }}/.bashrc
+    owner="root"
+    group={{ automated_user }}
+    mode="0744"
```
jibsheet: Same, was 0640

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+- name: create .ssh directory
+  file:
+    path={{ automated_home }}/.ssh state=directory mode=0700
```
jibsheet: We probably shouldn't let their user put things in .ssh

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+    owner={{ automated_user }} group={{ automated_user }}
+
+- name: build authorized_keys file
+  template:
+    src=home/automator/.ssh/authorized_keys.j2
+    dest={{ automated_home }}/.ssh/authorized_keys mode=0600
```
jibsheet: Do we want this to just be readonly?

jibsheet:
Couple of minor questions about file permissions, but those are not super important.

How does someone run a mangement command?  Are you going to symlink manage.py in?  But you will also need to symlink in their env file (or otherwise enable loading of the env file to setup virtualenvs and other django settings).  My usual scenario:

  sudo to programs
  source programs_env
  cd programs
  ./manage.py yada yada

I assume the restricted user has manage.py symlinked into their bin, but we'll also need to bring in the config file and allow sourcing (which is fine, rbash prevents them from sourc'ing /edx/app/programs/program_env but would allow it in their own directory).

Do we think the restricted user will be in the www-data group or do we not expect commands to generate output on disk?  There are edxapp commands that drop files and if you don't run them as the right user, it causes issues.  I'm not sure if anything else could encounter this (something something asset generation like we did for programs once).

playbooks/roles/automated/files/home/automator/.bashrc
```diff
@@ -0,0 +1 @@
+PATH=${HOME}/bin
```
maxrothman: Don't you need to export this?

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+#
+# That list should be provided via my_role's defaults
+#
+# role_automated_rbash_links:
+#   - /usr/bin/sudo
+#   - /usr/bin/scp
+#
+
+- fail: automated_rbash_links required for role
+  when: automated_rbash_links is not defined
```
maxrothman: I think this will always pass because `automated_rbash_links` is set to `!!null` in `defaults/main.yml`. You probably want to test for `is None` instead.

playbooks/roles/automated/defaults/main.yml
```diff
@@ -0,0 +1,29 @@
...
+# Vars for role automated
+#
+
+#
+# vars are namespace with the module name.
+#
+automated_role_name: automated
+automated_user: "changeme"
+automated_home: "/home/{{ automated_user }}"
+automated_rbash_links: !!null
```
maxrothman: Should `automated_authorized_keys` be defined here too?

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+- fail: automated_authorized_keys required for role
+  when: automated_authorized_keys is not defined
```
maxrothman: Potentially same as above

maxrothman:
Nitpick: use new-style yaml ansible syntax rather than old string syntax

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+- name: create automated user
+  user:
+    name={{ automated_user }} state=present shell=/bin/rbash
+    home={{ automated_home }} createhome=yes
+
+- name: create sudoers file from template
+  copy:
```
maxrothman: did you mean to use `template` here?

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+#
+# Dependencies:
+#
+# This role depends upon variables provided by an including role
+# via the my_role/meta/main.yml file.  Includes take the following forms:
+#
+#  dependencies:
+#   - role: automated
+#     automated_rbash_links: "{{ edxapp_automated_rbash_links }}"
+#     automated_sudoers_template: 'roles/edxapp/templates/etc/sudoers.d/99-automator-edxapp-server.j2'
```
maxrothman: Is there an example of what such a template would look like somewhere?

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+  # environment.
+  #
+- name: update shell file mode
+  file:
+    path={{ automated_home }}/{{ item }} mode=0640
+    state=file owner="root" group={{ automated_user }}
+  with_items:
+    - .bashrc
+    - .profile
+    - .bash_logout
```
maxrothman: also `.bash_profile`?

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+    path={{ automated_home }} mode=0750 state=directory
+    owner="root" group={{ automated_user }}
+
+- name: create ~automated/bin directory
+  file:
+    path={{ automated_home }}/bin state=directory mode=0750
+    owner="root" group={{ automated_user }}
+
+- name: re-write .profile
+  copy:
```
maxrothman: template?

playbooks/roles/automated/tasks/main.yml
```diff
@@ -0,0 +1,114 @@
...
+
+- name: build authorized_keys file
+  template:
+    src=home/automator/.ssh/authorized_keys.j2
+    dest={{ automated_home }}/.ssh/authorized_keys mode=0600
+    owner={{ automated_user }} group={{ automated_user }}
+
+- name: create allowed command links
+  file:
+    src={{ item }} dest={{ automated_home }}/bin/{{ item.split('/').pop() }}
```
maxrothman: Rather than `.pop()`, it'd probably be better to make explicit that you want the (first | last) index

maxrothman:
How much hardening do we want around this? Do we want to chroot this user?

e0d:
@maxrothman I don't think we want to chroot at this time, but maybe in the future.

fredsmith:
👍

PR #2999, opened 2016-05-02
Try to nudge folks towards slack/mailing list for installation issues

jibsheet:
@edx/devops @nedbat Encourage folks towards slack/mailing list for install issues.
https://github.com/blog/2111-issue-and-pull-request-templates

Wording thoughts appreciated, I just wanted to get it written quickly.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,18 @@
+GitHub issues are not meant for support questions. Please use the 
+[mailing list](https://groups.google.com/forum/#!forum/openedx-ops)
+or [Slack channels](https://open.edx.org/blog/open-edx-slack) to get help.
+
+Please open issues here to report bugs in the ansible scripts themselves.
+
+When reporting an issue, please include the following information.
+
+If `/edx/bin/show-repo-heads` is available on your system, it can provide much of this information:
+
+- Configuration ref:
+- edx-platform ref:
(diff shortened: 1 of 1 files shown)


maxrothman:
Why not just turn off issues?

nedbat:
Worth a shot! :)  :+1:

feanil:
👍   since it is easy for developers outside edx to create issues there for actual bugs.

maxrothman:
Make the second section a template rather than instructions, e.g.

```
Configuration ref:
edx-platform ref:
other refs:

#Steps to replicate:
- one
- two
- three
```

feanil:
that seems like a separate PR @maxrothman

maxrothman:
How so? @jibsheet asked for wording thoughts, I provided them. I'm happy to make the commit myself.

feanil:
This PR seems to be about the wording of the nudge, not about wording in the template generally but perhaps I misunderstood.

jibsheet:
I'm fine with blocking the merge until I can type in some better words for the prompt, there's no real rush, I just didn't want to forget to push the PR when it was so simple to turn on.

feanil:
I think we should merge this since it's already an improvement.  We can make more improvements later but don't need to block on them to make this available.

jibsheet:
Updated with wording changes and merging.  We can iterate further as needed.  I will test in the Issue creation screen after merging.

PR #3000, opened 2016-05-02
Updating go-agent container to a modern git client

macdiesel:
@feanil @doctoryes PTAL. The version of git included in this image is 1.9.1. I am updating this to the latest version to help cloning repositories using a ssh key that I do not want to write out to a file.


doctoryes:
👍 👍

PR #3002, opened 2016-05-02
We use logrotate, not this absent cronjob

jibsheet:
This was removed in 2014-08, don't run the task anymore.
aa780887

@edx-devops


feanil:
👍

jibsheet:
BTW - there is a much larger refactor we want here between this script and the logrotate job we install in common which depends on this script.  So, logrotate will have an hourly job to sync things, but no script to run unless you have multiple feature flags turned on.

PR #3004, opened 2016-05-03
Django 1.8 compatibility

jibsheet:
Also wire in the settings properly for this weird IDA

@edx/devops


e0d:
@jibsheet we need

> make migrate to just work.
> :+1:

jibsheet:
@e0d xqueue has a rakefile, we could make `rake migrate` work

PR #3005, opened 2016-05-03
Derf/split csmh configuration

jibsheet:
This branch enables the split CSMH configuration by default and drops in the csmh database into DATABASES by default.  edxlocal was previously updated to create the csmh database.

@edx/devops


feanil:
👍

PR #3010, opened 2016-05-04
add symmetric key JWT_AUTH config for course discovery

bderusha:
FYI @clintonb @peter-fogg


peter-fogg:
👍

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -77,6 +77,19 @@ DISCOVERY_ORGANIZATIONS_API_URL: '{{ DISCOVERY_LMS_ROOT_URL }}/api/organizations
 DISCOVERY_MARKETING_API_URL: 'https://example.org/api/catalog/v2/'
 DISCOVERY_MARKETING_URL_ROOT: 'https://example.org/'
 
+DISCOVERY_JWT_ALGORITHM: 'HS256'
+DISCOVERY_JWT_AUDIENCE: '{{ DISCOVERY_SOCIAL_AUTH_EDX_OIDC_KEY }}'
```
feanil, line 81: This makes it backwards compatible but is there a plan on simplifying the vars eventually as well?  If so, putting a deprecation warning in the CHANGLOG now would be a good idea.

clintonb:
https://github.com/edx/configuration/pull/3014

PR #3012, opened 2016-05-04
We routinely have more than 10 messages when we GC a bunch of clusters

jibsheet:
We should probably even just make this a bigger loop of 10.

@edx/devops


maxrothman:
👍

PR #3013, opened 2016-05-04
E0d/unbreak npm

e0d:
@jibsheet @clintonb @feanil 

This essentially follows what was happening below for bower.  I'm assuming this issue was introduced by the fact that we are not pinning the version of the prebuilt package we are installing, sigh.

Older version of the npm wrapper start with

```
#!/bin/sh
// 2>/dev/null; exec "`dirname "$0"`/node" "$0" "$@"
```

New versions start with 

```
#!/usr/bin/env node
;(function () { // wrapper in case we're in module_context mode

  // windows: running "npm blah" in this folder will invoke WSH, not node.
  /*global WScript*/
```

### DIFF
--- a/playbooks/roles/ecommerce/defaults/main.yml
+++ b/playbooks/roles/ecommerce/defaults/main.yml
@@ -205,7 +205,7 @@ ecommerce_requirements:
 ecommerce_environment:
   DJANGO_SETTINGS_MODULE: "{{ ECOMMERCE_DJANGO_SETTINGS_MODULE }}"
   ECOMMERCE_CFG: "{{ COMMON_CFG_DIR }}/{{ ecommerce_service_name }}.yml"
-  PATH: "{{ ecommerce_venv_dir }}/bin:{{ ansible_env.PATH }}"
+  PATH: "{{ ecommerce_nodeenv_bin }}:{{ ecommerce_venv_dir }}/bin:{{ ansible_env.PATH }}"
 
 #
 # OS packages
--- a/playbooks/roles/insights/defaults/main.yml
+++ b/playbooks/roles/insights/defaults/main.yml
@@ -150,8 +150,11 @@ INSIGHTS_REPOS:
 insights_environment:
   DJANGO_SETTINGS_MODULE: "analytics_dashboard.settings.production"
   ANALYTICS_DASHBOARD_CFG: "{{ COMMON_CFG_DIR  }}/{{ insights_service_name }}.yml"
+  PATH: "{{ insights_nodeenv_bin }}:{{ insights_venv_dir }}/bin:{{ ansible_env.PATH }}"
 
+  
 insights_service_name: insights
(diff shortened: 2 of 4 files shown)


clintonb:
👍 (once commits squashed)

Do we need to do this for the other IDAs?

jibsheet:
That's kind of terrible that the npm module (even in ansible 2) doesn't allow setting up the nodeenv.
👍

e0d:
The other IDAs will need this I believe.  I can do that in a separate PR.  I can confirm that the Insights AMI build successfully.

clintonb:
Thanks.

e0d:
@dan-f @dsjen FYI

e0d:
@clintonb one more look, I've moved the ecom changes here, programs doesn't require any and I've updated insights and ecom to source the environment before running npm.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -205,7 +205,7 @@ ecommerce_requirements:
 ecommerce_environment:
   DJANGO_SETTINGS_MODULE: "{{ ECOMMERCE_DJANGO_SETTINGS_MODULE }}"
   ECOMMERCE_CFG: "{{ COMMON_CFG_DIR }}/{{ ecommerce_service_name }}.yml"
-  PATH: "{{ ecommerce_venv_dir }}/bin:{{ ansible_env.PATH }}"
+  PATH: "{{ ecommerce_nodenv_bin }}:{{ ecommerce_venv_dir }}/bin:{{ ansible_env.PATH }}"
```
feanil: Missing an 'e'.

feanil:
👍  once the missing `e` in nodeenv are put back in insights and ecommerce.

playbooks/roles/insights/defaults/main.yml
```diff
@@ -150,8 +150,11 @@ INSIGHTS_REPOS:
 insights_environment:
   DJANGO_SETTINGS_MODULE: "analytics_dashboard.settings.production"
   ANALYTICS_DASHBOARD_CFG: "{{ COMMON_CFG_DIR  }}/{{ insights_service_name }}.yml"
+  PATH: "{{ insights_nodenv_bin }}:{{ insights_venv_dir }}/bin:{{ ansible_env.PATH }}"
```
clintonb: insights_nodeenv_bin

playbooks/roles/ecommerce/defaults/main.yml
clintonb: ecommerce_nodeenv_bin

clintonb:
What about credentials and the IDA template?

clintonb:
👍 (once variable names fixed)

e0d:
@clintonb credentials is not using npm, though it installs nodeenv, the role template doesn't include it.

clintonb:
@e0d can we squash the commits, and merge? We have many broken builds.

PR #3014, opened 2016-05-05
Added JWT_AUTH settings for discovery

clintonb:


clintonb:
FYI @maxrothman 

This should be grouped with DEVOPS-4137 and https://github.com/edx/course-discovery/pull/92.

clintonb:
FYI @peter-fogg @bderusha

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -59,6 +59,10 @@ DISCOVERY_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:80
 DISCOVERY_EDX_DRF_EXTENSIONS:
   OAUTH2_USER_INFO_URL: '{{ DISCOVERY_OAUTH_URL_ROOT }}/user_info'
 
+DISCOVERY_JWT_ISSUER: '{{ DISCOVERY_OAUTH_URL_ROOT }}'
+DISCOVERY_JWT_SECRET_KEY: '{{ EDXAPP_JWT_SECRET_KEY | default("SET-ME-PLEASE") }}'
```
maxrothman, line 64: Why the default? Shouldn't it break it if it can't get a value for `EDXAPP_JWT_SECRET_KEY`?

clintonb, line 64: Without a default CI won't pass, because we are not exposing `EDXAPP_JWT_SECRET_KEY` in this repo. That only happens in the secure configuration.

maxrothman:
👍

clintonb:
This relies on DEVOPS-4144.

clintonb:
I have confirmed this resolves the authentication issues on stage.

PR #3019, opened 2016-05-05
Instead of hardcoding a bucket name, use the common sync bucket

jibsheet:
This var is already properly used in things like TrackingLogVoter.yml.j2

@edx-ops/devops


fredsmith:
👍

PR #3022, opened 2016-05-06
Jibsheet/region aware lifecycle watcher

jibsheet:
Add region awareness to the two dynamic inventory scripts used in the minos lifecycles

@edx/devops


e0d:
:+1:

PR #3023, opened 2016-05-06
Allow specification of an aws_region

jibsheet:
We need to be able to override this to run outside us-east-1
@edx/devops


e0d:
:+1:

PR #3024, opened 2016-05-06
Preserve older versions of go-server backups.

doctoryes:
https://openedx.atlassian.net/browse/TE-1259

@edx/pipeline-team Please review.

### DIFF
--- a/playbooks/roles/go-server/tasks/main.yml
+++ b/playbooks/roles/go-server/tasks/main.yml
@@ -103,9 +103,8 @@
     group: "{{ GO_SERVER_GROUP }}"
     force: no
 
-# If a GoCD restore file is specified, attempt to download and restore it.
 - include: download_backup.yml
-  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT
+  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT and GO_SERVER_RESTORE_BACKUP == "true"
 
 - name: replace the admin line in the password file post-restore
   lineinfile:
@@ -127,4 +126,4 @@
     state: restarted
 
 - include: setup_regular_backup.yml
-  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT
+  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT and GO_SERVER_CONFIGURE_BACKUPS == "true"
--- a/playbooks/roles/go-server/templates/gocd_backup.j2
+++ b/playbooks/roles/go-server/templates/gocd_backup.j2
@@ -7,6 +7,7 @@ gocd_password="{{ GO_SERVER_BACKUP_PASSWORD }}"
 gocd_backup_location="{{ GO_SERVER_BACKUP_TMP_LOCATION }}"
 s3_backup_bucket="{{ GO_SERVER_BACKUP_S3_BUCKET }}"
 snitch_url="{{ GO_SERVER_BACKUP_SNITCH_URL }}"
+gocd_backup_base=gocd-"{{ GO_SERVER_BACKUP_ENVIRONMENT }}"
 
 # Trigger the backup and capture the backup path.
 # The output of the api call is documented here:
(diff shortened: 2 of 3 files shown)


playbooks/roles/go-server/tasks/download_backup.yml
```diff
@@ -59,13 +59,14 @@
     expiration=30
   when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT
   register: s3_one_time_url
+  ignore_errors: yes
```
feanil: Why ignore errors on this?

playbooks/roles/go-server/templates/gocd_backup.j2
```diff
@@ -52,22 +53,37 @@ null)
...
 
-# Make backup into a tarball.
-tar zcvf "$gocd_backup_location" -C "$(dirname "$backup_path")" "$(basename "$backup_path")"
+# Make latest backup into a tarball.
+backup_dir_name="$(basename "$backup_path")"
+tar zcvf "$gocd_backup_location" -C "$(dirname "$backup_path")" "${backup_dir_name}"
 
-# Transfer tarball to S3.
-aws_output=$(aws s3 cp "$gocd_backup_location" "s3://$s3_backup_bucket" 2>&1)
+dated_backup_filename="${gocd_backup_base}-${backup_dir_name}.tgz"
```
feanil, line 60: How does the date become a part of this filename?

playbooks/roles/go-server/tasks/download_backup.yml
doctoryes: Because if the file named `GO_SERVER_BACKUP_FILENAME` does not exist in the backup S3 bucket, an error will occur when generating the one-time URL. That error causes no `changed` attribute to be defined on `s3_one_time_url` and the next download step will be skipped, along with the subsequent restore.

playbooks/roles/go-server/templates/gocd_backup.j2
doctoryes, line 60: GoCD backups are dumped into top-level directories named like this:
`backup_20160421-153453`
So the basename of the path will contain the actual backup datetime.

playbooks/roles/go-server/tasks/download_backup.yml
feanil: It seems like that error is legitimate.  If we are trying to download a backup and it doesn't exist, it should fail.

doctoryes: That's fine - but we'd need a way to run the go-server playbook against an instance without having a backup present, since I imagine we'd want to support that scenario. What's the best practice for doing this in Ansible? Should I use a variable? Add a command-line `ansible-playbook` switch?

doctoryes: As we agreed, I will add two different variables that can be added to the command-line. One will enable the restore of a backup from S3 (`restore_backup`) and one will enable the setup of regular, cron-based backups to S3 (`configure_backups`).

doctoryes:
Flags are now added and I've tested them on the sandbox instance. To invoke both a backup restore and a regular backup configuration, you must issue this type of command:
`ansible-playbook go-server.yml -i gocd, --extra-vars "restore_backup=true configure_backups=true"`

playbooks/roles/go-server/tasks/main.yml
```diff
@@ -103,9 +103,8 @@
     group: "{{ GO_SERVER_GROUP }}"
     force: no
 
-# If a GoCD restore file is specified, attempt to download and restore it.
 - include: download_backup.yml
-  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT
+  when: GO_SERVER_BACKUP_S3_BUCKET and GO_SERVER_BACKUP_S3_OBJECT and restore_backup is defined and restore_backup == "true"
```
feanil: Let's make these ALL_CAPS vars since we expect users to override them.  I think we should also set a default of 'false' in the defaults/main.yml so that end users can just look in there to see what is overridable.

doctoryes: Yep - that'd be better. Will do.

doctoryes: And - done. Ready for another review.

feanil:
👍

PR #3028, opened 2016-05-09
Support configuration of tracking log rotation frequency

mulby:
@e0d - as discussed in passing, this change can enable us to change the behavior of logrotate for our different environments for tracking log files.

I would like to set this variable to different values in different environments. Environments that don't get as much traffic might want to set this variable to "hourly" to ensure that logs are rotated once per hour regardless of the size.

Reviewers:
- [ ] @e0d 
- [x] @feanil 
- [x] @brianhw

### DIFF
--- a/playbooks/roles/common_vars/defaults/main.yml
+++ b/playbooks/roles/common_vars/defaults/main.yml
@@ -171,3 +171,6 @@ common_close_all_caches: |
         cache = django_cache.cache
         if hasattr(cache, 'close'):
             cache.close()
+
+COMMON_TRACKING_LOG_ROTATION:
+  - size 1M
--- a/playbooks/roles/common/templates/etc/logrotate.d/hourly/edx_logrotate_tracking_log.j2
+++ b/playbooks/roles/common/templates/etc/logrotate.d/hourly/edx_logrotate_tracking_log.j2
@@ -1,4 +1,7 @@
 {{ COMMON_LOG_DIR }}/tracking/tracking.log {
+  {% for config in COMMON_TRACKING_LOG_ROTATION %}
+  {{ config }}
+  {% endfor %}
   compress
   create
   dateext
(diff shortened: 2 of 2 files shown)


brianhw:
In particular, having this option would be useful for rolling worker logs for edx prod, so that they're not days or weeks behind.  Even daily would be adequate.  :+1:

feanil:
I like having vars like `COMMON_TRACKING_LOG_ROTATION_CONDITION` defined in a default/main.yml so that all the default values are in one place as opposed to in the template files that uses the variables.

mulby:
@jibsheet I modified this to roll if the file gets larger than 10M _or_ once per day. The condition that is satisfied first will trigger a rotation.

We had set it to 1M so that it would upload to S3 more frequently since we used to run answer distribution every hour or so. Now we run it once per day, so there is no need to produce so many small files.

@brianhw FYI

mulby:
@feanil I updated this to set the common var

playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -171,3 +171,7 @@ common_close_all_caches: |
         cache = django_cache.cache
         if hasattr(cache, 'close'):
             cache.close()
+
+COMMON_TRACKING_LOG_ROTATION:
+  - maxsize 10M
```
feanil: I don't see `maxsize` as a valid directive for the version of logrotate that is running on ubuntu.  Have you tested this somewhere to validate it? Is the directive just not well documented?

mulby: Good catch, it looks like this was added in a later version than the one we run with. Hrm... looks like it was added in 3.8.1

We can use this directive when we upgrade to trusty. I'll revert to the old setting for now and we can override it in other environments.

feanil:
👍

mulby:
The failure here appears to be unrelated to this change, so I'm going to squash and merge.

PR #3029, opened 2016-05-09
Revert "MAYN-198: Enhance ansible provisioning script to support edx/edx-themes"

mattdrayer:
Reverts edx/configuration#2990

@jibsheet -- FYI, reverting this commit, as discussed

@saleem-latif -- we ran into a permissions issue with the ecommerce user attempting to compile assets in the "edx-themes" user-owned directory -- it seems like a better approach to copy the private themes templates+assets to the local ecommerce/themes folder and then compile the assets within ecommerce.


jibsheet:
👍 to the revert.  Glad you now have a recipe for testing on the sandbox.

PR #3031, opened 2016-05-10
Add env variable for dead man's snitch url

sanfordstudent:
Reviewers: @jibsheet @feanil 

For https://openedx.atlassian.net/browse/TNL-4493

### DIFF
--- a/playbooks/roles/notifier/defaults/main.yml
+++ b/playbooks/roles/notifier/defaults/main.yml
@@ -58,6 +58,7 @@ NOTIFER_REQUESTS_CA_BUNDLE: "/etc/ssl/certs/ca-certificates.crt"
 
 NOTIFIER_DD_API_KEY: "NOT_USED"    # data dog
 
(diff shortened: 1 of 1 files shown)


playbooks/roles/notifier/defaults/main.yml
```diff
@@ -58,6 +58,7 @@ NOTIFER_REQUESTS_CA_BUNDLE: "/etc/ssl/certs/ca-certificates.crt"
 
 NOTIFIER_DD_API_KEY: "NOT_USED"    # data dog
 
+NOTIFIER_DEAD_MANS_SNITCH_URL: "---"
```
feanil: Make the default something like `PUT_DEAD_MANS_SNITCH_URL_HERE` and add a comment indicating that it's optional with a link to https://deadmanssnitch.com/

feanil:
One small change and then it looks good to me. 👍

PR #3033, opened 2016-05-10
Installing libmemcached-dev for the discovery service

clintonb:
This is a requirement for installing the pylibmc pip package, necessary for interfacing with memcached.

ECOM-4370

### DIFF
--- a/playbooks/roles/discovery/defaults/main.yml
+++ b/playbooks/roles/discovery/defaults/main.yml
@@ -47,7 +47,7 @@ DISCOVERY_MEMCACHE: [ 'memcache' ]
 
 DISCOVERY_CACHES:
   default:
-    BACKEND:  'django.core.cache.backends.memcached.MemcachedCache'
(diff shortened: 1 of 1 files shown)


clintonb:
@jibsheet please review
FYI @edx/ecommerce this is necessary to use `pylibmc`. Stage consistently fails with `python-memcached` despite that package claiming support for Python 3.5. Locally, I get inconsistent behavior with Docker—sometimes caching works, other times it fails. `pylibmc` works consistently.

clintonb:
Related to https://github.com/edx/course-discovery/pull/98.

feanil:
👍

jibsheet:
👍

clintonb:
While validating this fix on stage, we discovered the security group was not properly set. This is most likely the cause of issues with `python-memcached` there. However, given that I cannot explain the issues with Docker/devstack, I am proceeding with the switch to `pylibmc`.

PR #3035, opened 2016-05-10
Use firefox 42 by default.

benpatterson:


jibsheet:
👍 Let me know if you want to cut a new devstack base image with this installed - just file a DEVOPS ticket.

benpatterson:
@jibsheet I'm going to wait for https://github.com/edx/edx-platform/pull/12434 to land before merging this one. Prod deployments would not be affected without the change, but building jenkins-workers would be.

PR #3036, opened 2016-05-11
Use a different smoke test, since xmodule js is flaky. TNL-4492.

benpatterson:
@jzoldak @estute this should significantly reduce the build-packer-ami failures.

It will also help me make better progress on [TE-1290](https://openedx.atlassian.net/browse/TE-1290)

PTAL


jzoldak:
👍

PR #3040, opened 2016-05-12
DEVOPS-4181 provide a URL for user accounts to the Analytics API

jibsheet:
@dsjen If you want to test this, you can build an analytics-api AMI with configuration=jibsheet/analytics-api-user-base-url while you wait for this to be reviewed.

I believe this is the dictionary where you want this setting set.

@edx/devops review?


dsjen:
Ack!  We actually are still iterating on the setting.  I'm so sorry to have wasted your time.  I left a message on the ticket a bit ago to that effect, but wasn't able to close it (permissions I think).

We do eventually need this setting to be set at some point (and the setting does exist on the API settings file), but just don't have the exact URL yet if that makes sense.

Please let me know if I can clarify anything.

jibsheet:
@dsjen Once you know the url, please take this branch as-is, edit it to change the /u/ part and trust that the other part will be set correctly.  You can then test with it and eventually PR it in for the API.
I'm closing the PR, but will leave the branch.

dsjen:
Thanks!

PR #3041, opened 2016-05-14
Disabled basic authentication for the /api paths of ecommerce and programs

clintonb:
ECOM-4414


clintonb:
@jibsheet please review

clintonb:
FYI @edx/ecommerce

rlucioni:
I'd like to understand why this hasn't affected us on stage (or prod) before.

peter-fogg:
Code looks good to me, but I'd also be interested in knowing the answer to @rlucioni's question.

clintonb:
We don't use basic auth for stage-edx-ecommerce.

rlucioni:
@clintonb or prod-edx-ecommerce, I'm guessing? 👍, thanks for looking into this.

feanil:
👍

PR #3042, opened 2016-05-16
Add affiliate cookie setting in all environments.

peter-fogg:
FYI @clintonb @bderusha I should have caught this earlier, but the setting override should be in both LMS and CMS environments since the code is being used in common/djangoapps and could run under either settings.


bderusha:
👍

PR #3057, opened 2016-05-20
add splunk searches via ansible

fredsmith:


fredsmith:
👍  Thumbing my own PR, but the changes @maxrothman made look good.

playbooks/roles/splunk-server/defaults/main.yml
```diff
@@ -27,8 +27,66 @@ splunk_frozen_dir: "{{ splunk_root }}/frozen"
 SPLUNK_INDEXES: 
   - "default"
 
+SPLUNK_ALERTS:
```
feanil: This isn't valid YAML right now, you should set it to `[]` so it has a default.

playbooks/roles/splunk-server/defaults/main.yml
```diff
@@ -27,8 +27,66 @@ splunk_frozen_dir: "{{ splunk_root }}/frozen"
...
+  #
+  #   severity: ([1-6], {{splunk_alert_default_severity}})
+  #     The severity of the alert. 1-debug, 2-info, 3-warn, 4-error, 5-severe, 6-fatal
+  #
+  #   digest_delay: (string, optional)
+  #     Whether to send email digests at most every "digest_delay" rather than for every alert. e.g. 15m
+  #
+  # NB: None of the string values can contain newlines except "message"
+
+SPLUNK_FIELD_EXTRACTIONS:
```
feanil: Same here.  Set a default of `[]`

playbooks/roles/splunk-server/defaults/main.yml
```diff
@@ -27,8 +27,66 @@ splunk_frozen_dir: "{{ splunk_root }}/frozen"
...
+  #   name:
+  #   regex:
+
+splunk_alert_default_severity: "3"
+
+
 SPLUNK_VOLUMES: []
 
+# Should include protocol & a trailing slash, e.g. http://splunk.mydomain.com/
+SPLUNK_HOSTNAME:
```
feanil: We should set default values even if they are empty strings.

playbooks/roles/splunk-server/defaults/main.yml
feanil: Capitalized vars should be at the top of the file and non-capital vars should be at the bottom per our convention.

feanil:
Had a few comments, 👍  once those are addressed.

playbooks/roles/splunk-server/defaults/main.yml
maxrothman: FWIW, that is valid YAML. It's equivalent to `SPLUNK_ALERTS: !!null`. But you're right, it should be a list.

PR #3058, opened 2016-05-20
This code has moved to an es module in the terraform repo

jibsheet:
Also update the README to be more accurate.

@edx/devops


maxrothman:
👍

PR #3059, opened 2016-05-22
Add role to create initial LMS users.

smarnach:
This role can be used to create LMS users on an Open edX instance.  It can be run using the `run_role.yml` playbook, and can be used on any readily provisioned instance.

For testing, create a file `vars.yml` with these contents:

```
role: create_lms_users

CREATE_LMS_USERS:
  - email: john@example.com
    name: John Smith
    username: john
    password: edx
    password_hash: false
    is_staff: false
```

and run it against any sandbox with edxapp installed

```
ansible-playbook ./run_role.yml -i "sandbox.domain," -u username_on_sandbox -e @vars.yml
```

**JIRA Ticket**: [OSPR-1282](https://openedx.atlassian.net/browse/OSPR-1282)
**Dependencies**: edx/edx-platform#12519 (if you want to use `password_hash: true` and to ensure idempotency)

### DIFF
--- a/playbooks/roles/create_lms_users/tasks/deploy.yml
+++ b/playbooks/roles/create_lms_users/tasks/deploy.yml
@@ -1,0 +1,10 @@
+- name: create LMS users
+  # It's important to use "command" here instead of "shell", since otherwise we would have to
+  # escape the special characters in the password (or other fields) for the shell.
+  command: >
+    {{ COMMON_APP_DIR }}/edxapp/venvs/edxapp/bin/python ./manage.py lms --settings=aws
+    create_user -e {{ item.email }} -n {{ item.name }} -u {{ item.username }} -p {{ item.password }}
+    {{ "-s" if item.is_staff else "" }} {{ "--password-hash" if item.password_hash else "" }}
+    chdir={{ COMMON_APP_DIR }}/edxapp/edx-platform
+  sudo_user: "{{ common_web_user }}"
+  with_items: CREATE_LMS_USERS
--- a/playbooks/roles/create_lms_users/defaults/main.yml
+++ b/playbooks/roles/create_lms_users/defaults/main.yml
@@ -1,0 +1,41 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role create_lms_users
+# 
+
+#
+# vars are namespaced with the module name.
+#
+create_lms_users_role_name: create_lms_users
+
+#
+# OS packages
+#
+
+create_lms_users_debian_pkgs: []
+
+create_lms_users_redhat_pkgs: []
+
+# The list of users to create.  Example:
+# CREATE_LMS_USERS:
+#   - email: martha@example.com
+#     name: Martha Jones
+#     username: martha
+#     password: 'pbkdf2_sha256$20000$TCWPsU7v6LEl$B+Rk8O1Sw4/lxAsnpYpNN2vtL/RAoL/G/+OFBNlolBI='
+#     password_hash: true
+#     is_staff: true
+#   - email: john@example.com
+#     name: John Smith
+#     username: john
+#     password: edx
+#     password_hash: false
+#     is_staff: false
+CREATE_LMS_USERS: []
--- a/playbooks/roles/create_lms_users/meta/main.yml
+++ b/playbooks/roles/create_lms_users/meta/main.yml
@@ -1,0 +1,14 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Role includes for role create_lms_users
+
+dependencies:
+  - common_vars
--- a/playbooks/roles/create_lms_users/tasks/main.yml
+++ b/playbooks/roles/create_lms_users/tasks/main.yml
@@ -1,0 +1,24 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Tasks for role create_lms_users
+# 
+# Overview:
+# 
+#
+# Dependencies:
+#
+# 
+# Example play:
+#
+#
+
+- include: deploy.yml tags=deploy


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=3059

playbooks/roles/create_lms_users/tasks/deploy.yml
```diff
@@ -0,0 +1,10 @@
+- name: create LMS users
+  # It's important to use "command" here instead of "shell", since otherwise we would have to
+  # escape the special characters in the password (or other fields) for the shell.
+  command: >
+    {{ COMMON_APP_DIR }}/edxapp/venvs/edxapp/bin/python ./manage.py lms --settings=aws
+    create_user -e {{ item.email }} -n {{ item.name }} -u {{ item.username }} -p {{ item.password }}
+    {{ "-s" if item.is_staff else "" }} {{ "--password-hash" if item.password_hash else "" }}
+    chdir={{ COMMON_APP_DIR }}/edxapp/edx-platform
+  sudo_user: "{{ common_web_user }}"
```
jbzdak, line 9: NIT: Ansible prefers `become_user`. `sudo_*` directives are deprecated and will be retired.

jbzdak:
Technically this role is not idempotent. To be idempotent it should also update password for the user. Right now if password is changed in the configuration between runs --- it will not be updated.

jbzdak:
@smarnach Looks good, tested. Will give :+1:  once we decide what do to with the fact that password will not be updated if changed in the configuration.

smarnach:
@jbzdak The role deliberately does not update the password, because otherwise it would overwrite any password changes the user might have done on the instance, because the role will be rerun everytime we provision a new AppServer.  (Note that this doesn't mean the role isn't idempotent.  Idempotency only means that applying the role twice in direct succession is the same as applying it once.  It doesn't guarantee _convergence_, but most Ansible module fail to provide that property.)

playbooks/roles/create_lms_users/tasks/deploy.yml
smarnach, line 9: I know.  However, there are currently 220 occurrences of `sudo_user` in the configuration code base, and 0 occurrences of `become_user`.  I decided that consistency is more important in this situation, and that a single place where `become_user` is used doesn't provide any value.

jbzdak, line 9: OK. You convinced me.

On Mon, May 23, 2016 at 5:20 PM, Sven Marnach notifications@github.com
wrote:

> In playbooks/roles/create_lms_users/tasks/deploy.yml
> https://github.com/edx/configuration/pull/3059#discussion_r64239755:
> 
> > @@ -0,0 +1,10 @@
> > +- name: create LMS users
> > -  # It's important to use "command" here instead of "shell", since otherwise we would have to
> > -  # escape the special characters in the password (or other fields) for the shell.
> > -  command: >
> > -    {{ COMMON_APP_DIR }}/edxapp/venvs/edxapp/bin/python ./manage.py lms --settings=aws
> > -    create_user -e {{ item.email }} -n {{ item.name }} -u {{ item.username }} -p {{ item.password }}
> > -    {{ "-s" if item.is_staff else "" }} {{ "--password-hash" if item.password_hash else "" }}
> > -    chdir={{ COMMON_APP_DIR }}/edxapp/edx-platform
> > -  sudo_user: "{{ common_web_user }}"
> 
> I know. However, there are currently 220 occurrences of sudo_user in the
> configuration code base, and 0 occurrences of become_user. I decided that
> consistency is more important in this situation, and that a single place
> where become_user is used doesn't provide any value.
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> https://github.com/edx/configuration/pull/3059/files/e57ed7024e67c647f2f9858ae94fd2b95a5aa1e0#r64239755

## 

Jacek
@OpenCraft

jbzdak:
OK. If we assume that after initial provisioning passwords will be managed
by the user itself this makes sense.

:+1: On the PR.

On Mon, May 23, 2016 at 5:18 PM, Sven Marnach notifications@github.com
wrote:

> @jbzdak https://github.com/jbzdak The role deliberately does not update
> the password, because otherwise it would overwrite any password changes the
> user might have done on the instance, because the role will be rerun
> everytime we provision a new AppServer. (Note that this doesn't mean the
> role isn't idempotent. Idempotency only means that applying the role twice
> in direct succession is the same as applying it once. It doesn't guarantee
> _convergence_, but most Ansible module fail to provide that property.)
> 
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> https://github.com/edx/configuration/pull/3059#issuecomment-221008295

## 

Jacek
@OpenCraft

e0d:
@smarnach I'd like this to support, and the typical usage to be, setting unusable passwords when creating the accounts.  The assumption being that real user would just use the password reset work-flow and passwords would not need to be stored in config or transferred.

smarnach:
@e0d While it doesn't match our use case to set unusable passwords, I can see that this would be the most common use case.  Currently, there are two problems with implementing this:
1. The `create_user` management command does not support creating users with an unusable password.  I've implemented support for this in [this commit](https://github.com/open-craft/edx-platform/commit/4366e62833e89573c7ac5d5a1556362f5be90797).
2. The password reset form [rejects users with an unusable password](https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/forms.py#L49-L51).  The two easiest ways to fix this are to either remove the validation code that rejects users with unusable passwords, or to use a long random password in `create_user` rather than an unusable password.  Do you know why users with unusable passwords can't reset their passwords?  Is it acceptable to change this?  Who could give me advice on how to proceed with this?

openedx-webhooks:
Thanks for the pull request, @smarnach! I've created [OSPR-1282](https://openedx.atlassian.net/browse/OSPR-1282) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

nedbat:
@smarnach another bot hiccup.  I'll clean it up.

smarnach:
@nedbat Thanks.  Is there anything I should do differently?

bradenmacdonald:
@nedbat @smarnach Unless I'm mistaken, this PR and https://github.com/edx/edx-platform/pull/12519 are both supposed to be OSPRs. They are contributions from OpenCraft that are related to our new hosting infrastructure plans. So what was the bot hiccup?

smarnach:
@bradenmacdonald Yes, they definitely are supposed to be OSPRs.  I thought @nedbat  was referring to the fact that I had to trigger the bot twice (the first time resulted in a 500), though I checked that only one JIRA ticket was created.  The one that got created was definitely intentional, though.

playbooks/roles/create_lms_users/tasks/deploy.yml
feanil, line 9: @smarnach there is a PR to fix those that should merge soon, please use the non-deprecated format.

feanil:
@smarnach why not use https://github.com/edx/configuration/blob/master/playbooks/edx-east/manage_edxapp_users_and_groups.yml ?  We've started using that playbook and it's been working well for our use-cases.  If it's missing features I think we should add them or move it into the role(though I think this is a bit heavy handed for how little it does.)

smarnach:
@feanil Yes, indeed, these look very similar to what we need.  I searched all _roles_ for something that's usable for us, but I didn't search the playbooks in the edx-east directory, so thanks for the pointer.

I'm a bit confused about the [`manage_user`](https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/management/commands/manage_user.py) and [`create_user`](https://github.com/edx/edx-platform/blob/master/common/djangoapps/student/management/commands/create_user.py) management commands.  They are almost identical, with each providing a few extra features over the other one.  The "demo" role uses the `create_user` command to create users.  That command used to be idempotent as well, but currently isn't anymore due to a bug fixed in edx/edx-platform#12519.  I added the additional features we need to the `create_user` command, but could just as well add them to the `manage_user` command.  Shouldn't these two commands be merged?

`manage_user` creates users with an unusable password.  How do they get enabled?  Users with an unusable password can't use the password reset feature (see also [my comment further up](https://github.com/edx/configuration/pull/3059#issuecomment-223349894)).

feanil:
@smarnach `manage_user` is new and was written by the ecom team to be able to do user management things.  Probably, the two should be merged and all new functionality should be added to `manage_user`.   `manage_user` was originally part of a separate django app that we thought could be shared between various edx IDAs but I think we are not doing that anymore.  @clintonb @rlucioni can you provide some context on the two very similar commands that I might have missed?

clintonb:
Please use `manage_user`. As @feanil stated, this used to live in https://github.com/edx/edx-django-extensions, but it made no sense to have this outside of edx-platform since we need to also create a profile. `create_user` is an older command that does a bit too much—enrolling users—in a non-ideal manner—forms. I'd like to see its usage replaced entirely by `manage_user` and a separate command to handle enrollments.

PR #3060, opened 2016-05-23
mattdrayer/ecommerce-worker-multisite: Add site-specific overrides

mattdrayer:
@fredsmith @rlucioni -- FYI, this is the third part of the solution to enable site-specific configuration overrides for the ecommerce-worker service.  Pretty sure I'm missing something here.  If you can take a look and let me know what else needs to be changed that would be great.

Ref: 
- edx/ecommerce-worker#16
- edx/ecommerce#751

### DIFF
--- a/playbooks/roles/ecomworker/defaults/main.yml
+++ b/playbooks/roles/ecomworker/defaults/main.yml
@@ -52,9 +52,10 @@ ECOMMERCE_WORKER_WORKER_ACCESS_TOKEN: 'your-secret-here'
 ECOMMERCE_WORKER_MAX_FULFILLMENT_RETRIES: 11
 # END ORDER FULFILLMENT
 
-# Ecommerce Worker settings 
+# Ecommerce Worker settings
 ECOMMERCE_WORKER_JWT_SECRET_KEY: 'insecure-secret-key'
 ECOMMERCE_WORKER_JWT_ISSUER: 'ecommerce_worker'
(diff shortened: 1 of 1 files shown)


playbooks/roles/ecomworker/defaults/main.yml
```diff
@@ -63,6 +63,21 @@ ECOMMERCE_WORKER_SERVICE_CONFIG:
...
+  # SITE_OVERRIDES: {
+  #   "site1": {
+  #      "ECOMMERCE_API_ROOT": "http://ecommerce-subdomain.domain.com"
+  #   },
+  #   "site2": {
+  #      "JWT_SECRET_KEY": "site2-secret-key",
+  #      "JWT_ISSUER": "site2-worker"
+  #   }
+  # }
+  ECOMMERCE_WORKER_SITE_OVERRIDES: !!null
```
rlucioni: This key must match the name of the ecomworker setting you're trying to override. In your case, `SITE_OVERRIDES`. You should still set `ECOMMERCE_WORKER_SITE_OVERRIDES`, but do it outside the `ECOMMERCE_WORKER_SERVICE_CONFIG` hash. Then set `SITE_OVERRIDES` equal to it.

mattdrayer:
Thanks for the advice, @rlucioni -- I've updated as you indicated

rlucioni:
👍

PR #3061, opened 2016-05-23
use first pip in path

fredsmith:


jibsheet:
:+1:

jzoldak:
👍 
I confirmed that this gets the build-packer-ami job that creates jenkins workers past the point where they were failing. DEVOPS-4251

PR #3063, opened 2016-05-23
Seeing fail: in ansible output is confusing to folks

jibsheet:
@edx/devops I've been asked this question a few times


maxrothman:
👍

PR #3050, opened 2016-05-17
WL-451: Clean up ansible services/roles for theming

saleem-latif:
Hi @mattdrayer , @feanil , @fredsmith 

Kindly review this PR it contains changes for WL-451, I have refactored out git-clone and add-user roles that are common to both edx_service and edx-themes roles.

PR #3068, opened 2016-05-24
Revert "WL-451: Clean up ansible services/roles for theming"

feanil:
Reverts edx/configuration#3050

@saleem-latif @mattdrayer @fredsmith FYI

This broke when building AMIs:  Seems like we could have caught it with earlier testing...

TASK: [git-clone | remove read-only ssh key] **********************************
fatal: [localhost] => One or more undefined variables: 'unicode object' has no attribute 'REPO'


jibsheet:
👍

PR #3071, opened 2016-05-25
Revert "Revert "WL-451: Clean up ansible services/roles for theming""

saleem-latif:
Hi @mattdrayer , @fredsmith , @feanil 

Kindly review this PR, I have fixed and tested the bug related to git-clone

**Error Trace from previous build:**
TASK: [git-clone | remove read-only ssh key] **********************************
fatal: [localhost] => One or more undefined variables: 'unicode object' has no attribute 'REPO'


saleem-latif:
@mattdrayer , @fredsmith , @feanil , @jibsheet I have verified the changes included in this PR by following instructions on https://openedx.atlassian.net/wiki/pages/viewpage.action?spaceKey=EdxOps&title=Ops+Ansible+Testing+Checklist, and all tests are passing. Kindly, take a look into this PR.

fredsmith:
👍

mattdrayer:
Cool, if we can get one more +1 from either @feanil or @jibsheet we can give this another shot.

e0d:
Roles should use _ rather than -
https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions#AnsibleCodeConventions-RoleNamingConventions
This isn't a quibble, python treats - as an infix minus, which causes occasional surprising behavior.  We have some legacy roles with - in the name, but they'll be refactored as touched.

maxrothman:
2 things:
- This changes the order some of the tasks run in, i.e. git clone happens earlier than before. Are you confident it'll work as intended?
- Please use the 2 new roles' `defaults/main.yml`s to document the expected include interface and set defaults for appropriate vars.

saleem-latif:
@e0d I have updated names for new roles to use underscore `_` instead of minus `-`, 
@maxrothman git clone does not depend upon the roles that earlier were being executed before it, that is why we can move it before those tasks, It was also discussed on [WL-451](https://openedx.atlassian.net/browse/WL-451?focusedCommentId=189422&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-189422). I have also update the code according to your suggestion in second point.
Kindly take a look now.

maxrothman:
@saleem-latif I meant to suggest that you actually encode the defaults in `defaults/main.yml`, not just leave comments there.

saleem-latif:
@maxrothman Oh sorry, I forgot to leave that in comment, actually there are no defaults as of now for these roles. that is why `defaults/main.yml` is empty

mattdrayer:
It would be really great if we can wrap up this PR today -- it seems like functionally-speaking everything is in order so is there any reason we can't merge the change set at this point?  It's now blocking other work that we're trying to finish by the end of the week.

mattdrayer:
Okay, been an hour and no further comments -- we are going to merge this PR at 11am Eastern unless we hear otherwise

PR #3073, opened 2016-05-26
Specify path to aws command.

doctoryes:
@edx/pipeline-team Please review. This change will fix go-server backups to S3.


macdiesel:
👍

feanil:
We should also add a check for the return code of the command if it's possible.  👍

doctoryes:
I've added a check of the return code and tested it on a go-server.

PR #3074, opened 2016-05-30
delete the user home directory as well if state is absent

arbabnazar:
@edx/devops kindly review this change. Thanks

**Reason to this change:**  We delete the users from tool-gp server but it leave the user's home directory . We regularly cleanup our user file in secure repo but when new team member work on it and compare it with the server home directory, then he thought that we forgot to remove the ex-employee, so this change will reflect that our user file and server state are both on same line.

One example:
- [File Cleanup Operation](https://github.com/edx-ops/edx-secure/commit/fa99ec7cf2d01fc33150710da2103d60b4968e7d)
- [Re-added by someone else because file doesn't match the system state](https://github.com/edx-ops/edx-secure/commit/a7c9e063693e5faeb39022b08ba184d495439c86)


playbooks/roles/user/tasks/main.yml
```diff
@@ -105,6 +105,13 @@
     state: "{{ item.state | default('present') }}"
   with_items: user_info
 
+- name: delete user home directory if state is absent
```
e0d: Why not use the remove argument to the user module for this?
http://docs.ansible.com/ansible/user_module.html
http://linux.die.net/man/8/userdel

arbabnazar:
@e0d please review. Thanks for the tips, I thought that we need to repeat the user task for removing the users with this option but after testing, it turn out the we can use it even with the `state: present` and it will not affect it and will only affect with the `state: absent`. Once again, thanks for this tips.

CC: @edx/devops

jibsheet:
👍

PR #3075, opened 2016-05-31
rewrite the edxapp role

arbabnazar:

Jenkins build to verify this branch (7501)

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -1,54 +1,36 @@
-- name: setup the edxapp env
-  template: >
-    src=edxapp_env.j2 dest={{ edxapp_app_dir }}/edxapp_env
-    owner={{ edxapp_user }}  group={{ common_web_user }}
-    mode=0644
-  tags:
-    - install
-    - install:configuration
-
+---
 - name: create edxapp configuration dir
-  file: >
-    path="{{ EDXAPP_CFG_DIR }}" state=directory
-    owner="{{ edxapp_user }}" group="{{ common_web_group }}"
-  tags:
-    - install
-    - install:configuration
-
-- name: setup the newrelic python agent config
-  template:
-    src: "newrelic.ini.j2"
-    dest: "{{ edxapp_app_dir }}/newrelic.ini"
+  file:
+    path: "{{ EDXAPP_CFG_DIR }}" 
+    state: directory
     owner: "{{ edxapp_user }}"
-    group: "{{ common_web_user }}"
-    mode: "0644"
-  tags:
-    - install
-    - install:configuration
-
-# Optional auth for git
-- name: create ssh script for git (not authenticated)
-  template: >
-    src=git_ssh_noauth.sh.j2 dest={{ edxapp_git_ssh }}
-    owner={{ edxapp_user }} mode=750
-  when: not EDXAPP_USE_GIT_IDENTITY
+    group: "{{ common_web_group }}"
   tags:
     - install
     - install:base
 
-- name: create ssh script for git (authenticated)
-  template: >
-    src=git_ssh_auth.sh.j2 dest={{ edxapp_git_ssh }}
-    owner={{ edxapp_user }} mode=750
-  when: EDXAPP_USE_GIT_IDENTITY
+- name: copy the template to the desired location
+  template:
+    src: "{{ item.src }}"
+    dest: "{{ item.dest }}"
+    owner: "{{ item.owner }}"  
+    group: "{{ item.group }}"
+    mode: "0644"
+  with_items:
+    - { src: 'edxapp_env.j2', dest: '{{ edxapp_app_dir }}/edxapp_env', owner: '{{ edxapp_user }}', group: '{{ common_web_user }}' }
+    - { src: 'newrelic.ini.j2', dest: '{{ edxapp_app_dir }}/newrelic.ini', owner: '{{ edxapp_user }}', group: '{{ common_web_user }}' }
+    - { src: 'git_ssh.sh.j2', dest: '{{ edxapp_git_ssh }}', owner: '{{ edxapp_user }}', group: '{{ edxapp_user }}' }
   tags:
     - install
     - install:base
 
 - name: install read-only ssh key
-  copy: >
-    content="{{ EDXAPP_GIT_IDENTITY }}" dest={{ edxapp_git_identity }}
-    force=yes owner={{ edxapp_user }} mode=0600
+  copy:
+    content: "{{ EDXAPP_GIT_IDENTITY }}"
+    dest: "{{ edxapp_git_identity }}"
+    force: yes
+    owner: "{{ edxapp_user }}"
+    mode: "0600"
   when: EDXAPP_USE_GIT_IDENTITY
   tags:
     - install
@@ -63,11 +45,11 @@
 
 # Do A Checkout
 - name: checkout edx-platform repo into {{ edxapp_code_dir }}
-  git_2_0_1: >
-    dest={{ edxapp_code_dir }}
-    repo={{ edx_platform_repo }}
-    version={{ edx_platform_version }}
-    accept_hostkey=yes
+  git_2_0_1:
+    dest: "{{ edxapp_code_dir }}"
+    repo: "{{ edx_platform_repo }}"
+    version: "{{ edx_platform_version }}"
+    accept_hostkey: yes
   sudo_user: "{{ edxapp_user }}"
   environment:
     GIT_SSH: "{{ edxapp_git_ssh }}"
@@ -90,11 +72,11 @@
 # (yes, lowercase) to a Stanford-style theme and set
 # edxapp_theme_name (again, lowercase) to its name.
 - name: checkout Stanford-style theme
-  git_2_0_1: >
-    dest={{ edxapp_app_dir }}/themes/{{ edxapp_theme_name }}
-    repo={{ edxapp_theme_source_repo }}
-    version={{ edxapp_theme_version }}
-    accept_hostkey=yes
+  git_2_0_1:
+    dest: "{{ edxapp_app_dir }}/themes/{{ edxapp_theme_name }}"
+    repo: "{{ edxapp_theme_source_repo }}"
+    version: "{{ edxapp_theme_version }}"
+    accept_hostkey: yes
   when: edxapp_theme_name != ''
   sudo_user: "{{ edxapp_user }}"
   environment:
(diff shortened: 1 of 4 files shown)


e0d:
@smarnach can yourself or a delegate also review?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -1,54 +1,37 @@
...
-  tags:
-    - install
-    - install:base
-
-- name: create ssh script for git (authenticated)
-  template: >
-    src=git_ssh_auth.sh.j2 dest={{ edxapp_git_ssh }}
-    owner={{ edxapp_user }} mode=750
-  when: EDXAPP_USE_GIT_IDENTITY
+  with_items:
```
smarnach, line 19: I don't see too much benefit in using a loop here, since the only thing the three tasks have in common is that they use the template module, so it could just as well be written as three separate tasks.  (It's a matter of personal preference of course.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -1,54 +1,37 @@
...
+    - { src: 'edxapp_env.j2', dest: '{{ edxapp_app_dir }}/edxapp_env', owner: '{{ edxapp_user }}', group: '{{ common_web_user }}' }
+    - { src: 'newrelic.ini.j2', dest: '{{ edxapp_app_dir }}/newrelic.ini', owner: '{{ edxapp_user }}', group: '{{ common_web_user }}' }
+    - { src: 'git_ssh.sh.j2', dest: '{{ edxapp_git_ssh }}', owner: '{{ edxapp_user }}', group: '{{ edxapp_user }}' }
   tags:
     - install
     - install:base
+    - install:configuration
```
smarnach: I'm not familiar with the meaning of the individual tags.  Are they documented anywhere?

This changes the tags on the template tasks.  The original version had `install:base` missing on some of the tasks, and `install:configuration` on others, while the new version includes all three tags for all three tasks.  This is probably intended, but without knowing what the tags mean, I can't tell whether it's correct.

playbooks/roles/edxapp/tasks/deploy.yml
arbabnazar, line 19: @smarnach as you say that the `template` module is the common point between three tasks, then there is no point to write/repeat three tasks to do same thing(transfer the template), even I have eliminate the two ssh template tasks into one.

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -160,8 +144,9 @@
 # Set the npm registry permissions
 - name: Set the npm registry permissions
   file:
-    path="{{ edxapp_app_dir }}/.npmrc"
-    owner=edxapp group=edxapp
+    path: "{{ edxapp_app_dir }}/.npmrc"
+    owner: edxapp
+    group: edxapp
```
smarnach: Shouldn't these two be `"{{ edxapp_user }}"` instead, to be consistent with the rest of the file?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -363,53 +349,45 @@
...
-
-- name: create symlinks from the repo dir
-  file: >
-    src="{{ edxapp_code_dir }}/{{ item }}"
-    dest={{ COMMON_BIN_DIR }}/{{ item.split('.')[0] }}.edxapp
-    state=link
+- name: create symlinks from the venv bin dir and repo dir
+  file:
+    src: "{{ item }}"
+    dest: "{{ COMMON_BIN_DIR }}/{{ item.split('.')[0] | basename }}.edxapp"
```
smarnach: Since the items now contain `edxapp_venv_bin`, this code only works if there is no dot in the directory name, which seems a bit brittle.  The following might work, though I have not tested it at all:

```
{{ item | basename | splitext | first }}
```

smarnach:
@arbabnazar Thanks for doing this, the new syntax is a lot more readable!  I've added a few minor comments, and I'll test the new version tomorrow morning (though, from what I can tell, it doesn't look like this should change the behaviour in any way).

playbooks/roles/edxapp/tasks/deploy.yml
arbabnazar: I have tested it by provisioning the sandbox from scratch, can you please also provision the sandbox to verify the end to end change. Thanks

playbooks/roles/edxapp/tasks/deploy.yml
smarnach, line 19: @arbabnazar I like the change of merging the two different git ssh templates into one, and there is nothing wrong with using a loop here.  I'm just wondering whether it would be simpler as three tasks, which is a purely stylistic question.

playbooks/roles/edxapp/tasks/deploy.yml
smarnach: @arbabnazar I'm planning to test your new code tomorrow.  Or are you asking me to test the code I provided?

arbabnazar: @smarnach no, I am just asking to test this branch at your environment to spin up the fresh sandbox, meanwhile, I'll modify the code to implement your suggestions. Thanks

arbabnazar:
Thanks @smarnach for review. Really appreciated.

playbooks/roles/edxapp/tasks/deploy.yml
arbabnazar: @smarnach unfortunately we cannot use the `splitext` it's introduced in 2.0

smarnach: @arbabnazar Right, I didn't notice this when I looked at the documentation.  Could we use `item.rsplit('.', 1)` instead of `item.split('.')`?  What I'm worried about is that splitting at the first dot leads to strange results when the directory name contains a dot.  It's unlikely that it does (and it certainly doesn't in the default configuration), but I'd prefer to code defensively here.

arbabnazar: @smarnach thanks for detail review all the PR. I have tested the `item.rsplit('.', 1)` but even it didn't work if we have dot between the directory name and I have verified that we are not in a practice to use the dot between the directories name.

arbabnazar:
@e0d can I get second review from you please before merge. Thanks

playbooks/roles/edxapp/tasks/deploy.yml
smarnach: @arbabnazar I tested the code using `rsplit()` myself, and I think it works fine.  To avoid miscommunication, here an example I tried:

``` yaml
---
- hosts: localhost
  become: false
  gather_facts: false
  vars:
    dir_a: /a/b.c/d
    dir_b: /e/f/g
    file_a: "{{ dir_a }}/file.name"
    file_b1: "{{ dir_b }}/{{ file_a.split('.') | first | basename }}"
    file_b2: "{{ dir_b }}/{{ file_a.rsplit('.', 1) | first | basename }}"
  tasks:
    - debug:
        var: file_b1
    - debug:
        var: file_b2
```

When run with `ansible-playbook test.yml -i localhost,`, the output shows that `file_b1` has the incorrect value `/e/f/g/b`, while `file_b2` has the expected value of `/e/f/g/file`.

arbabnazar: @smarnach thanks for the example. `rsplit()` works fine if all the `basename` contain `.` but in our case, all the `basename` doesn't contain `.` which mislead to the the wrong value. Here is the full example with the output.

``` yaml
---
- hosts: all
  gather_facts: no
  tasks:
    - debug:
        msg: "{{ item.rsplit('.', 1) | first | basename }}"
      with_items:
        - /edx/app/edxapp.test/python
        - /edx/app/edxapp.test/pip
        - /edx/app/edxapp.test/django-admin.py
        - /edx/app/edxapp.test/manage.py
```

Output of this code is:

``` shell
% ansible-playbook -i "localhost," test.yml -c local

PLAY [all] *********************************************************************

TASK [debug] *******************************************************************
ok: [localhost] => (item=/edx/app/edxapp.test/python) => {
    "item": "/edx/app/edxapp.test/python",
    "msg": "edxapp"
}
ok: [localhost] => (item=/edx/app/edxapp.test/pip) => {
    "item": "/edx/app/edxapp.test/pip",
    "msg": "edxapp"
}
ok: [localhost] => (item=/edx/app/edxapp.test/django-admin.py) => {
    "item": "/edx/app/edxapp.test/django-admin.py",
    "msg": "django-admin"
}
ok: [localhost] => (item=/edx/app/edxapp.test/manage.py') => {
    "item": "/edx/app/edxapp.test/manage.py'",
    "msg": "manage"
}

PLAY RECAP *********************************************************************
localhost                  : ok=1    changed=0    unreachable=0    failed=0
```

I think we need to stick with the `split()` because if work for us. 

Thanks,

smarnach: @arbabnazar Thanks for the example, I see you are right.  This is a minor point, and I didn't expect this discussion to drag on for this long, but now that we have spent the time to understand this, we should find a proper implementation.  My next suggestion is

```
{{ (item | basename).split('.', 1) | first }}
```

which seems to get all test cases right, and does things in the logical order (first extract the basename, then split off the externsion).  If this one doesn't work as well, I'm giving up, I promise. :)

playbooks/roles/edxapp/tasks/deploy.yml
e0d: Because of the loop the union of tags has to be applied to this task, or could.  Tag isolation should be preferred to terseness.  However, in this case I think `install:configuration` might not be correct for these templates, though it was originally applied to them.  I think it should apply to the tasks related to yaml config, nginx vhosts, `/etc/defaults` only.

Docs are WIP, but here https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle

arbabnazar: @e0d can you please review. Thanks

playbooks/roles/edxapp/tasks/deploy.yml
arbabnazar: @smarnach Thanks for suggestion, I have updated the task as per you suggestion, please verify and let me know if you want me to make any change. Thanks

smarnach:
@arbabnazar Thanks, all good now! :+1:

arbabnazar:
@edx/devops second review please. Thanks

arbabnazar:
@feanil can you please review. Thanks

feanil:
:+1:

PR #3078, opened 2016-05-31
Add parens to fix backup script.

doctoryes:
A silly bash logic bug is now fixed - GoCD backups will now successfully upload and report to the snitch.

@edx/pipeline-team Please review.


macdiesel:
Bash is the best! 👍 🚢 🇮🇹

PR #3080, opened 2016-06-01
use the YAML style parameter passing

arbabnazar:

Jenkins build to verify this branch (7542)

### DIFF
--- a/playbooks/roles/edxapp/tasks/main.yml
+++ b/playbooks/roles/edxapp/tasks/main.yml
@@ -50,7 +50,7 @@
   tags:
     - install
     - install:base
-    
+
 - name: create edxapp log dir
   file:
     path: "{{ edxapp_log_dir }}"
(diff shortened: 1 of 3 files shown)


feanil:
The 14.04 style blas and lapack linking should work with newer versions of ubuntu as well including xenial.  I think we should remove the conditionals that check the distribution and just think of it as two possible places where blas and lapack could be installed.  Everything else looks good so if you want to make that change, let's pull it out into a separate PR.

arbabnazar:
@feanil kindly review, is it ok I have removed the condition to check the distribution. Thanks

feanil:
👍   @arbabnazar please build an AMI off of this branch to make sure that the tagging of AMIs still works since that part of the code is being refactored and it's pretty crucial to our workflow.

arbabnazar:
@feanil I have verified the tagging part and it's working. I am merging this branch. Thanks for the review and suggestions.

PR #3081, opened 2016-06-01
Add the edx ppa and upgrade to python 2.7.10.

feanil:
@edx/pipeline-team


macdiesel:
FWIW I've had difficulty when adding multiple repositories when building docker containers without running update between the adds. Assuming CI checks out 👍

PR #3082, opened 2016-06-01
Feanil/upgrade gocd

feanil:
Supersedes https://github.com/edx/configuration/pull/3081

@edx/pipeline-team


doctoryes:
So - GoCD 16.5 is Java 8-compatible. Which means we don't have to upgrade Java immediately - but if we don't upgrade Java along with the GoCD 16.5 upgrade, our hand will be forced at some future GoCD upgrade. Details: https://www.go.cd/releases/#16.5.0

doctoryes:
And 👍

macdiesel:
👍

PR #3052, opened 2016-05-18
edx_sandbox.yml: Run edxlocal role only if edxapp db is on localhost

pomegranited:
This change updates the `edx_sandbox.yml` playbook to run the `edxlocal` role only if the edxapp database is hosted on localhost, and makes the `mysql` role a dependency of `edxlocal`.

It also moves the `memcached` package installation out of `edxlocal` and into the `edxapp` role, and installs it if `EDXAPP_MEMCACHE` contains `localhost`.

Previously, the `mysql` role was run only if the edxapp database was hosted on localhost, and `edxlocal` had no such check.  So if the edxapp DB was not local, the subsequent `edxlocal` role would fail due to missing dependencies.

**Sandboxes**

`EDXAPP_MYSQL_HOST == 'localhost'`:
- LMS: http://pr12539.sandbox.opencraft.com
- Studio: http://studio.pr12539.sandbox.opencraft.com

`EDXAPP_MYSQL_HOST == 'pr12539.sandbox.opencraft.com`:
- LMS: http://pr12539-remote.sandbox.opencraft.com
- Studio: http://studio.pr12539-remote.sandbox.opencraft.com

**Reviewers**
- [x] @mtyaka 
- [x] e0d
- [x] fredsmith
- [x] feanil

PR #3086, opened 2016-06-03
Revert "edx_sandbox.yml: Run edxlocal role only if edxapp db is on localhost"

benpatterson:
Reverts edx/configuration#3052

See the original PR for discussion on why it should be reverted.


benpatterson:
@pomegranited took care of the problem in a separate PR. Closing this one.

PR #3088, opened 2016-06-03
psmisc isn't on precise docker images and the jenkins roles want it.

jibsheet:
It is part of the default AWS AMI though, which is why we haven't noticed.

@edx/devops


feanil:
👍

PR #3089, opened 2016-06-03
Dockerfile for the jenkins_analytics role

jibsheet:
As noted in the second commit, specifically excluded from Travis, but is "good enough"
to get you a local docker testing image, especially if you configure a secret key for github locally.

@edx/devops


feanil:
👍

PR #3095, opened 2016-06-07
Add another region to ec2.py

jibsheet:
This appears to be the only way to ask it to look in multiple regions
since it doesn't support AWS_REGION or --region.
Alternately, we could start having independent .ini files for deploys.

@edx/devops


feanil:
:+1:

PR #3097, opened 2016-06-07
Improvements to django_admin.sh

maxrothman:
- Echo the command to run before running (for post-mortem debugging)
- Stop adding "--settings aws" to the end of every command, even when it
  was added to $manage earlier


feanil:
:+1:

PR #3098, opened 2016-06-08
Updated configuration for multiple IDAs to support Docker-based devstack

clintonb:
ECOM-6560


docker/build/credentials/Dockerfile
```diff
@@ -10,7 +10,7 @@
 FROM edxops/trusty-common:latest
 MAINTAINER edxops
 
-ARG CREDENTIALS_VERSION=master
+ARG CREDENTIALS_VERSION=clintonb/devstack
```
clintonb: These branch changes will be reverted to master once the respective PRs are merged.

clintonb:
@rlucioni @jmbowman @davec-edx @edx/devops please review

davec-edx (approved):

docker/build/ecommerce/Dockerfile
```diff
@@ -1,14 +1,34 @@
...
+ENV ECOMMERCE_VERSION=clintonb/devstack
+ENV REPO_OWNER=edx
+
 COPY docker/build/ecommerce/ansible_overrides.yml /
-RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook ecommerce.yml -i '127.0.0.1,' -c local -t "install:base,install:system-requirements,install:configuration,install:app-requirements,install:code" -e@/ansible_overrides.yml
-COPY docker/build/ecommerce/docker-run.sh /
 
-CMD ["/docker-run.sh"]
-EXPOSE 8130
+# NOTE: We are NOT compiling static assets due to the fact that database access is required to do so,
```
rlucioni, line 23: I know this isn't new, but it's really unfortunate.

playbooks/roles/credentials/tasks/main.yml
```diff
@@ -30,23 +30,40 @@
...
+- name: build virtualenv
+  command: "virtualenv {{ credentials_venv_dir }}"
+  args:
+    creates: "{{ credentials_venv_dir }}/bin/pip"
+  become_user: "{{ credentials_user }}"
+  tags:
+    - install
+    - install:system-requirements
+
+- name: install nodenv
```
rlucioni, line 42: What's the reasoning behind having nodeenv installed by Ansible instead of service requirements files (e.g., https://github.com/edx/credentials/pull/132/files#diff-6f1f964e63cc3c4b164496bc75516b41L9)?

rlucioni (approved):

playbooks/roles/credentials/tasks/main.yml
clintonb: Having `nodeenv` be installed by `pip` leads to an order of operations issue. Ideally, we want Ansible to call Make targets, with a single target, `production-requirements`, being responsible for installing production requirements. That is impossible with the old implementation. Also, nodeenv (to me) sits on the same level as virtualenv. It's infrastructure. The IDAs shouldn't be building infrastructure.

docker/build/ecommerce/Dockerfile
rlucioni: @clintonb for my own edification, when are ecommerce static assets collected if not here?

playbooks/roles/credentials/tasks/main.yml
rlucioni: Makes sense, thanks.

rlucioni: Minor: might want to correct the spelling of nodeenv here.

docker/build/ecommerce/Dockerfile
clintonb: Good question. They should be compiled/collected when you start devstack...assuming you have a site setup. I will express my frustrations about this in-person. *sigh*

rlucioni:
@clintonb is IDA Ansible consolidation going to happen here, or will that be a new PR?

clintonb:
@rlucioni new PR.

PR #3102, opened 2016-06-08
This only worked by coincidence before

jibsheet:
We were implicitly relying on mongo to return the list of members in
their _id order, which is absurd and I have no idea how we didn't catch
it previously.

We could do a little list comprehension to pull out the member with the
same _id,  but it seemed easier to just sort both lists and retain the
original intent.

@edx/devops


maxrothman:
👍  :shipit:  🙀

feanil:
:+1:

PR #3104, opened 2016-06-09
WIP - rolling mongodb upgrades

jibsheet:
Currently, a hidden secondaries upgrade script

This will connect to a cluster, find the hidden secondaries and upgrade
them to the current version specified in the mongo_3_0 play.

This required the creation of a new library module since our two current
ones return rs.status and call replSetReconfig, but neither return
rs.config, and rs.config is what tells you if a host is hidden and the
priority, which are needed for this.

@edx/devops - still need to write the full rolling playbook, but this was an
intermediate AC we liked and plan to run in prod environments in advance of the
full upgrade.

### DIFF
--- a/playbooks/library/mongodb_rs_config
+++ b/playbooks/library/mongodb_rs_config
@@ -108,11 +108,7 @@ def main():
 
     if repl_set:
         status = client.admin.command("replSetGetStatus")
-    else:
-        module.exit_json(changed=False)
-
-    # Not using `replSetGetConfig` because it's not supported in MongoDB 2.x.
-    if repl_set:
+        # Not using `replSetGetConfig` because it's not supported in MongoDB 2.x.
         rs_config = client.local.system.replset.find_one()
     else:
         module.exit_json(changed=False)
@@ -123,9 +119,13 @@ def main():
     status    = json.loads(json_util.dumps(status))
     rs_config = json.loads(json_util.dumps(rs_config))
 
+    # Status contains information about Primary/Secondary, so we iterate that list
+    # But we want to return config for that host, not status (since this is the config module),
+    # this is the inner loop of the comprehension, where we match on the hostname (could also
+    # match on _id).
     primary   = [ c for m in status['members'] if m['stateStr'] == 'PRIMARY'   for c in rs_config['members'] if m['name'] == c['host'] ]
     secondary = [ c for m in status['members'] if m['stateStr'] == 'SECONDARY' for c in rs_config['members'] if m['name'] == c['host'] ]
-    # we're parsing the config here, much simpler
+    # we're parsing the config directly here, much simpler
     hidden    = [ m for m in rs_config['members'] if m['hidden'] ]
 
     module.exit_json(changed=False, primary=primary, secondary=secondary, hidden=hidden, config=rs_config)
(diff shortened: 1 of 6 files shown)


feanil:
I assume the install will restart mongo as a part of the post-install but I think we should add a check afterwards to ensure that mongo has started.

jibsheet:
Yeah, apt-get install the mongo server does an explicit restart (hence no handler) but you're right, I was meaning to add in the task I wrote that waits until the cluster is healthy.  Opinions on cluster health vs server started?  Both?

feanil:
First ensure it's started, then make sure it's healthy?  So it's easier to pinpoint what kind of failure we have.

jibsheet:
Third commit contains the two tasks we use in the mongo_3_0 play.
I wish there was a better mechanism than include: to share this code across the plays, but I may break this up for my next playbook.

playbooks/library/mongodb_rs_config
```diff
@@ -0,0 +1,135 @@
...
+    else:
+        module.exit_json(changed=False)
+
+    # This converts the bson into a python dictionary that ansible's standard
+    # jsonify function can process and output without throwing errors on bson
+    # types that don't exist in JSON
+    status    = json.loads(json_util.dumps(status))
+    rs_config = json.loads(json_util.dumps(rs_config))
+
+    primary   = [ c for m in status['members'] if m['stateStr'] == 'PRIMARY'   for c in rs_config['members'] if m['name'] == c['host'] ]
```
feanil, line 126: Can you add a sentence in english for what this comprehension is doing, it's pretty hard to read because of the double for loop.

feanil:
One request for some clarity then on a list comprehension then  :+1:

playbooks/library/mongodb_rs_config
```diff
@@ -0,0 +1,135 @@
...
+
+    # This checks to see if you have a replSetName configured
+    # This generally means that /etc/mongod.conf has been changed
+    # from the default to use a replica set and mongo has been
+    # restarted to use it.
+
+    try:
+        repl_set = client.admin.command('getCmdLineOpts')['parsed']['replication']['replSetName']
+    except (OperationFailure, KeyError):
+        module.exit_json(changed=False)
```
maxrothman, line 107: Shoudn't this be `fail_json`?

playbooks/library/mongodb_rs_config
```diff
@@ -0,0 +1,135 @@
...
+    if repl_set:
+        status = client.admin.command("replSetGetStatus")
+    else:
+        module.exit_json(changed=False)
+
+    # Not using `replSetGetConfig` because it's not supported in MongoDB 2.x.
+    if repl_set:
+        rs_config = client.local.system.replset.find_one()
+    else:
+        module.exit_json(changed=False)
```
maxrothman, line 114: Group these into the same block

playbooks/library/mongodb_rs_config
jibsheet, line 107: I modeled this module on mongo_rs_status which doesn't bail out of the play if it fails, since we use it to detect if the replset is configured (something the update replica set module doesn't support).  I found that useful when writing other plays, but could be convinced to error if folks prefer it.

jibsheet:
Addressed doc notes / squashing code notes and some commentary on failing vs exiting in the module.

maxrothman:
👍

PR #3106, opened 2016-06-09
Full rolling updates of a cluster

jibsheet:


jibsheet:
@edx/devops Here's the current version of the rolling upgrade code.  This has been run in an environment.  My only real concern is that it is super repetitive.  I don't see a fix for that other than a tiny role with a 2 task file that I include over and over.  Suggestions welcome.

Note that this merges into my current branch which I will then merge to master.

jibsheet:
@fredsmith @e0d It would be great to have comments on these before I use them in prod/edge tomorrow (stage/matasano already done).

fredsmith:
👍

PR #3107, opened 2016-06-09
Correct credentials URL root protocol

rlucioni:
OIDC doesn't work out of the box if the redirect URI uses http.

@clintonb @maxrothman


clintonb:
:+1:

maxrothman:
👍

PR #3111, opened 2016-06-13
Define ALTERNATE_WORKER_QUEUES

efischer19:
@jibsheet Here's my first pass at this.

I'm going to try a sandbox deploy to test this shortly.

### DIFF
--- a/playbooks/roles/edxapp/templates/cms.conf.j2
+++ b/playbooks/roles/edxapp/templates/cms.conf.j2
@@ -10,7 +10,7 @@ command={{ executable }} -c {{ edxapp_app_dir }}/cms_gunicorn.py {{ EDXAPP_CMS_G
 
 user={{ common_web_user }}
 directory={{ edxapp_code_dir }}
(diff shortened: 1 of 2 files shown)


efischer19:
Needs to land before https://github.com/edx/edx-platform/pull/12689

playbooks/roles/apache/templates/lms.j2
```diff
@@ -14,6 +14,7 @@ WSGIRestrictEmbedded On
     </Directory>
 
     SetEnv SERVICE_VARIANT lms
+    # Do I need to define ALTERNATE_WORKER_QUEUES here as well?
```
jibsheet: Shouldn't need this.  I forgot we even have an Apache role.

jibsheet:
Check my logic.  We don't need to touch workers.conf.j2 because workers only care about their queues.  We're ensuring the gunicorn workers can see all queues.  

Need to review the platform PR but this looks fine, and we can probably merge early once we're sure of the names on that PR.

efischer19:
That's the same logic I had - the workers don't need to know about alternate workers, the gunicorn workers do.

I'll get the comment change backed out and merge this after tests pass again.

jibsheet:
retroactive 👍

PR #3112, opened 2016-06-13
SSH needs this script to be executable

jibsheet:
During the edxapp refactor, three tasks were combined, but the modes
were all squished to 0644

@e0d 
FYI @arbabnazar 
FYI @estute this is what caused the AMI build failure, we'll need to rebuild with this after it merges.


e0d:
:+1:

PR #3113, opened 2016-06-13
Quote mode strings

jibsheet:
Ansible turned 0640 into 420 and 0750 into 488, only one of which are
valid file modes.  Quotes prevent it from doing so too early.
Fixes d2f1595c.

@e0d round 2


e0d:
:+1:

PR #3115, opened 2016-06-14

arbabnazar:

Jenkins build to verify this branch (7768)


playbooks/edx-east/test-rolling.yml
```diff
@@ -1,5 +1,5 @@
 - hosts: all
-  sudo: true
+  become: true
```
smaeda-ks: @arbabnazar 
true -> True?

Sorry from outsider...

arbabnazar: yes just follow the convention everywhere.

feanil:
:+1:

PR #3122, opened 2016-06-15
Rename folders directory to folder to match docker/plays/forum.yml

michaelroytman:


jibsheet:
I suspect you need to update docker/compose/edx/docker-compose.yml also to s/forums/forum
but after that (and after the CI runs) 👍

PR #3123, opened 2016-06-15
Single Sign-out updates for IDAs

clintonb:
ECOM-4609 and ECOM-4678


clintonb:
@e0d please review

clintonb:
Confirmed on sandbox.

PR #3108, opened 2016-06-09
Update to ansible 2.x

feanil:
Just trying to figure out what's still failing.

PR #3128, opened 2016-06-16
Fix all failing ansible syntax checks in roles when we move to Ansible 2.0

feanil:
There are still a lot of warnings that need to be fixed but these
were the only failures.

@edx/devops this is a small change that we should merge so that when we get to 2.0 this isn't broken.

Pulling this out of the more controversial PR https://github.com/edx/configuration/pull/3108


jibsheet:
👍

PR #3133, opened 2016-06-17
Added OpenID Connect logout URL to all IDAs and the IDA template

clintonb:
ECOM-4609


clintonb:
@edx/devops please review

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -55,6 +55,7 @@ DISCOVERY_VERSION: "master"
 DISCOVERY_DJANGO_SETTINGS_MODULE: "course_discovery.settings.production"
 DISCOVERY_URL_ROOT: 'http://discovery:18381'
 DISCOVERY_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
+DISCOVERY_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
```
e0d, line 58: This port should be a var ideally.

clintonb, line 58: Is that a blocker for merging this, or can we improve that later?

PR #3138, opened 2016-06-17
Sigh

e0d:
@jibsheet @douglashall @mattdrayer @arbabnazar @fredsmith 

Working theory here is that converting a task to use with_items caused the modes to be interpolated in a strange way.  Note that @arbabnazar 's PR did not change how we were expressing the mode, but dryed out the tasks file by using with_items.

On the server the mode is clearly not correct

```
sudo stat -c "%a %n" /opt/splunkforwarder/etc/system/local/inputs.conf 
1204 /opt/splunkforwarder/etc/system/local/inputs.conf
```

With my change:

```
sudo stat -c "%a %n" /opt/splunkforwarder/etc/system/local/inputs.conf
644 /opt/splunkforwarder/etc/system/local/inputs.conf
```

### DIFF
--- a/playbooks/roles/splunkforwarder/tasks/main.yml
+++ b/playbooks/roles/splunkforwarder/tasks/main.yml
@@ -104,7 +104,7 @@
     dest: "/opt/splunkforwarder/etc/system/local/{{ item }}.conf"
     owner: splunk
     group: splunk
-    mode: 644
+    mode: "0644"
   with_items:
     - inputs
     - outputs


e0d:
@mdinino we'll need to pull this change in once it has merged.

douglashall:
@e0d Nice work! This looks like it was a fun one.

jibsheet:
:+1:
Same fix I had to make for the git ssh identity problem. (quoted octal string).

e0d:
@jibsheet what we didn't realize at the time is that quoting is made necessary by the with_items construct.  In the case that the bare octal string is used in a singly applied task, quoting is not required.

Meant to call this PR to the attention of @feanil and @maxrothman as well.

arbabnazar:
@e0d I didn't know about this and even it was not documenting anywhere that `with_items` behave strangely with mode. Thanks for this finding and share.

👍

smaeda-ks:
@e0d 
FYI
I shared you before about similar things.
https://openedx.slack.com/archives/ops/p1448054575000027

```
playbooks/roles/asqatasun/tasks/main.yml:    mode: 755
playbooks/roles/edx_ansible/tasks/deploy.yml:    mode: 755
playbooks/roles/splunkforwarder/tasks/main.yml:    mode: 644
playbooks/roles/tanaguru/tasks/main.yml:    mode: 755
```

PR #3140, opened 2016-06-20

arbabnazar:


feanil:
There are some places where we use both `become` and `become_user` in the same task and others where we only use `become_user`.  It seems like using just `become_user` should be sufficient but I don't think those changes need to be a part of this PR which updates syntax.

:+1:

arbabnazar:
@feanil yes you are right, we can open separate PR for that change. Thanks

PR #3143, opened 2016-06-20
Michael/dependencies

michaelroytman:
WIP

### DIFF
--- a/util/parsefiles.py
+++ b/util/parsefiles.py
@@ -1,0 +1,280 @@
+import os
+import pathlib2
+import logging
+import yaml
+import sys
+import networkx as nx
+from collections import namedtuple
+
+class FileParser:
+
+    def __init__(self):
+        self.paths = self._load_repo_path()
+
+    def _load_repo_path(self):
+        """Loads the path for the configuration repository from TRAVIS_BUILD_DIR environment variable."""
+
+        if os.environ.get("TRAVIS_BUILD_DIR"):
+            self.repo_path = os.environ.get("TRAVIS_BUILD_DIR")
+        else:
+            raise EnvironmentError("TRAVIS_BUILD_DIR environment variable is not set.")
+
+    def build_graph(self, git_dir, roles_dirs, aws_play_dirs, docker_play_dirs):
+
+        """
+        Builds a dependency graph that shows relationships between roles and playbooks.
+        An edge [A, B], where A and B are roles, signifies that A depends on B. An edge
+        [C, D], where C is a playbook and D is a role, signifies that C uses D.
+        """
+
+        graph = nx.DiGraph()
+
+        self._map_roles_to_roles(graph, roles_dirs, git_dir, "dependencies", "role", "role")
+        self._map_plays_to_roles(graph, aws_play_dirs, git_dir, "roles", "aws_playbook", "role")
+        self._map_plays_to_roles(graph, docker_play_dirs, git_dir, "roles", "docker_playbook", "role")
+
+        return graph
+
+    def _map_roles_to_roles(self, graph, dirs, git_dir, key, type_1, type_2):
+        """Maps roles to the roles that they depend on."""
+
+        Node = namedtuple('Node', ['name', 'type'])
+
+        # for each role directory
+        for d in dirs:
+            d = pathlib2.Path(git_dir, d)
+
+            if d.is_dir():
+                # for all files/sub-directories in directory
+                for directory in d.iterdir():
+
+                    # attempts to find meta/*.yml file in directory
+                    role = [file for file in directory.glob("meta/*.yml")]
+
+                    # if role exists
+                    if role:
+                        with (open(str(role[0]), "r")) as file:
+                            yaml_file = yaml.load(file)
+
+                        # if a yaml file and key in file
+                        if yaml_file is not None and key in yaml_file:
+                            # for each dependent role
+                            for dependent in yaml_file[key]:
+                                # get role name
+                                name = self._get_role_name(dependent)
+
+                                # add node for role
+                                node_1 = Node(directory.name, type_1)
+                                # add node for dependent role
+                                node_2 = Node(name, type_2)
+                                # add edge role - dependent role
+                                graph.add_edge(node_1, node_2)
+
+    def _map_plays_to_roles(self, graph, dirs, git_dir, key, type_1, type_2):
+        """Maps plays to the roles they use."""
+
+        Node = namedtuple('Node', ['name', 'type'])
+
+        # for each play directory
+        for d in dirs:
+            d = pathlib2.Path(git_dir, d)
+
+            if d.is_dir():
+                # for all files/sub-directories in directory
+                for directory in d.iterdir():
+                    # if a yaml file
+                    if directory.is_file() and directory.suffix == ".yml":
+                        with (open(str(directory), "r")) as file:
+                            yaml_file = yaml.load(file)
+
+                        if yaml_file is not None:
+                            # for each play in yaml file
+                            for play in yaml_file:
+                                # if specified key in yaml file (e.g. "roles")
+                                if key in play:
+                                    # for each role
+                                    for role in play[key]:
+                                        # get role name
+                                        name = self._get_role_name(role)
+
+                                        # add node for playbook
+                                        node_1 = Node(directory.stem, type_1)
+                                        # add node for role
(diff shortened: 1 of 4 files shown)


michaelroytman:
@edx/devops Please review.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+import logging
+import yaml
+import sys
+import networkx as nx
+from collections import namedtuple
+
+class FileParser:
+
+    def __init__(self):
+        self.paths = self._load_repo_path()
```
michaelroytman: Remove "self.paths =" initialization; unnecessary.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+        # for each role directory
+        for d in dirs:
+            d = pathlib2.Path(git_dir, d)
+
+            if d.is_dir():
+                # for all files/sub-directories in directory
+                for directory in d.iterdir():
+
+                    # attempts to find meta/*.yml file in directory
+                    role = [file for file in directory.glob("meta/*.yml")]
```
feanil: NIT: `file` is a [built-in function](https://docs.python.org/2/library/functions.html#file) you should avoid using it as the name of your variable.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+                    # if role exists
+                    if role:
+                        with (open(str(role[0]), "r")) as file:
+                            yaml_file = yaml.load(file)
```
feanil: If the file is not valid YAML, this can throw a parsing exception, you should handle that case.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+    # read from standard in
+    for line in sys.stdin:
+        change_set.add(line.rstrip())
+
+    # read config file
+    config_file_path = pathlib2.Path(parser.repo_path, "util", "parsefiles_config.yml")
+
+    with config_file_path.open() as config_file:
+            config = yaml.load(config_file)
```
maxrothman: Nitpick: inconsistent indent

maxrothman:
Please document the expected format of the config file

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+class FileParser:
+
+    def __init__(self):
+        self._load_repo_path()
+
+    def _load_repo_path(self):
+        """Loads the path for the configuration repository from TRAVIS_BUILD_DIR environment variable."""
+
+        if os.environ.get("TRAVIS_BUILD_DIR"):
+            self.repo_path = os.environ.get("TRAVIS_BUILD_DIR")
```
maxrothman: Nitpick: no need to use `.get()` here since you already checked that it was defined.

I'd probably write this like so:

```
repo_path = os.environ.get(TRAVIS_BUILD_DIR_VAR)
if repo_path:
  self.repo_path = repo_path
else:
  raise EnvironmentError(...)
```

where TRAVIS_BUILD_DIR_VAR is a constant defined at the top of the file.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+        Builds a dependency graph that shows relationships between roles and playbooks.
+        An edge [A, B], where A and B are roles, signifies that A depends on B. An edge
+        [C, D], where C is a playbook and D is a role, signifies that C uses D.
+        """
+
+        graph = nx.DiGraph()
+
+        self._map_roles_to_roles(graph, roles_dirs, git_dir, "dependencies", "role", "role")
+        self._map_plays_to_roles(graph, aws_play_dirs, git_dir, "roles", "aws_playbook", "role")
+        self._map_plays_to_roles(graph, docker_play_dirs, git_dir, "roles", "docker_playbook", "role")
```
maxrothman: Nitpick: factor this out into a for-loop with the list of dirs to check in a list as a class constant.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
+import os
+import pathlib2
+import logging
+import yaml
+import sys
+import networkx as nx
+from collections import namedtuple
+
+class FileParser:
```
maxrothman: Inherit from object for python 2.x compatibility

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+              rbenv_ruby_version: "{{ forum_ruby_version }}"
+
+            :param role:
+            :return:
+            """
+            if isinstance(role, dict):
+                return role['role']
+            elif isinstance(role, basestring):
+                return role
+            return None
```
maxrothman: Should this print a warning message or throw an exception rather than failing silently?

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+    def _map_plays_to_roles(self, graph, dirs, git_dir, key, type_1, type_2):
+        """Maps plays to the roles they use."""
+
+        Node = namedtuple('Node', ['name', 'type'])
+
+        # for each play directory
+        for d in dirs:
+            d = pathlib2.Path(git_dir, d)
+
+            if d.is_dir():
```
maxrothman: If a playbooks dir doesn't exist, this tool should fail noisily, so I don't there's any need for this check.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+                # for all files/sub-directories in directory
+                for directory in d.iterdir():
+                    # if a yaml file
+                    if directory.is_file() and directory.suffix == ".yml":
```
maxrothman: `directory` is misnamed. It could be a file or a directory.

You could probably just replace this with `glob('*.yml')`.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+        # set of roles
+        items = set()
+
+        # for all directories containing roles
+        for role_dir in roles_dirs:
+            role_dir_path = pathlib2.Path(git_dir, role_dir)
+
+            # all files in role directory
+            candidate_files = [file for file in role_dir_path.glob("**/*")]
```
maxrothman: Make this a set, not a list

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+                    items.add(name)
+
+        # for all directories containing playbooks
+        for play_dir in playbooks_dirs:
+            play_dir_path = pathlib2.Path(git_dir, play_dir)
+
+            # all files in role directory that end with yml extension
+            candidate_files = [file for file in play_dir_path.glob("*.yml")]
+
+            for file in files:
```
maxrothman: As before, don't use `file` as a var name

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+        items = set()
+
+        for role in roles:
+            items.add(role)
+
+            dependents = nx.descendants(graph, (role, "role"))
+
+            names = {dep.name for dep in dependents}
+
+            items |= names
```
maxrothman: Nitpick: no reason to assign `names`, you can just do `items |= {...}`

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+        self.check_coverage(coverage)
+
+        return items
+
+    def filter_docker_plays(self, plays, repo_path):
+        """Filters out docker plays that do not have a Dockerfile."""
+
+        items = set()
+        logger = logging.getLogger(__name__)
```
maxrothman: If you're using a logger, you should use it everywhere and use no print statements.

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+
+        for play in plays:
+            dockerfile = pathlib2.Path(self.repo_path, "docker", "build", play, "Dockerfile")
```
maxrothman: The path to docker plays should be configurable (at least as a constant at the top of the file)

util/parsefiles.py
```diff
@@ -0,0 +1,280 @@
...
+                items.add(play)
+            else:
+                logger.warning(" covered playbook '%s' does not have Dockerfile." % play)
+
+        return items
+
+    def check_coverage(self, coverage):
+        """Checks which aws roles are not covered by docker plays."""
+
+        logging.basicConfig(level=logging.WARNING)
```
maxrothman: Set the loglevel as a script arg, not in class methods

fredsmith:
LGTM 👍

PR #3144, opened 2016-06-21
OPS-1433 grant-ssh access job should fail if user has no keys

arbabnazar:

[Job Reference](ansible-grant-ssh-access/1244/console)

### DIFF
--- a/playbooks/roles/user/tasks/main.yml
+++ b/playbooks/roles/user/tasks/main.yml
@@ -74,13 +74,13 @@
 - debug: 
     var: user_info
 
-- name: create the edxadmin group
+- name: Create the edxadmin group
   group: 
     name: edxadmin 
     state: present
 
 # some AMIs (such as EMR master nodes) don't read the config files out of /etc/sudoers.d by default
-- name: ensure sudoers.d is read
+- name: Ensure sudoers.d is read
   lineinfile:
     dest: /etc/sudoers 
     state: present
@@ -89,7 +89,7 @@
     validate: 'visudo -cf %s'
 
 # give full sudo admin access to the edxadmin group
-- name: grant full sudo access to the edxadmin group
+- name: Grant full sudo access to the edxadmin group
   copy: 
     content: "%edxadmin ALL=(ALL) NOPASSWD:ALL"
     dest: /etc/sudoers.d/edxadmin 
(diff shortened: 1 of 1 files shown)


playbooks/roles/user/tasks/main.yml
```diff
@@ -98,29 +98,40 @@
...
+- name: Check the ssh key(s) for user(s) over github
+  uri:
+    url: "https://github.com/{{ item.name }}.keys"
+  with_items: "{{ user_info }}"
+  register: github_users_return
+
+- fail:
+    msg: "User {{ item.item.name }} didn't added ssh key to his account over github"
+  with_items: "{{ github_users_return.results }}"
+  when: item.content_length == "0"
```
e0d, line 125: Is this a reliable test?  Out of curiosity do we get a 404 response?

e0d:
:+1:

playbooks/roles/user/tasks/main.yml
arbabnazar, line 125: yes, it's reliable test. didn't get 404 error. Thanks

playbooks/roles/user/tasks/main.yml
```diff
@@ -98,29 +98,40 @@
...
 
-- name: get github key(s) and update the authorized_keys file
+- name: Check the ssh key(s) for user(s) over github
+  uri:
+    url: "https://github.com/{{ item.name }}.keys"
+  with_items: "{{ user_info }}"
+  register: github_users_return
+
+- fail:
+    msg: "User {{ item.item.name }} didn't added ssh key to his account over github"
```
feanil, line 123: A more clear way of saying this might be: `User {{ item.item.name }} doesn't have an SSH key associated with their account`

arbabnazar, line 123: Thanks @feanil, I'll modified it in new PR

jibsheet:
@arbabnazar this breaks AMI builds because there are old absent users in our config without public ssh keys.  Ideally it wouldn't die on absent users without ssh keys.

arbabnazar:
@jibsheet Thanks for pointing and addressing it into separate PR.

jibsheet:
@arbabnazar this additionally causes failures documented in https://openedx.atlassian.net/browse/DEVOPS-4400

TL;DR - original version of this used fetch_url which uses urllib which is installed in 12.04 and 14.04
Updated version using uri: uses httplib2 which is installed on 12.04 but is not on 14.04.

Fixing in abbey.py so that we can build IDAs deployed on 14.04

jibsheet:
Further fallout https://github.com/edx/configuration/pull/3177

PR #3146, opened 2016-06-21
Added unsuable_password option to manage_edxapp_users_and_groups.yml

kylecrawshaw:
This PR is necessary to reflect changes in https://github.com/edx/edx-platform/pull/12804 that will allow for an unusable password to be set for a user. @edx/devops when the next release is completed is it safe to assume that those changes will be pushed to prod? This option shouldn't be used until the changes are live.

### DIFF
--- a/playbooks/edx-east/manage_edxapp_users_and_groups.yml
+++ b/playbooks/edx-east/manage_edxapp_users_and_groups.yml
@@ -19,6 +19,7 @@
 #   - superuser (optional, bool)
 #   - staff (optional, bool)
 #   - remove (optional, bool): ensures the user does not exist
+#   - unusable_password (optional, bool): ensures the password is unusable
 #
 # Groups can have the following properties:
 #   - name (required, str)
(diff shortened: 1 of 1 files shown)


e0d:
:+1:

playbooks/edx-east/manage_edxapp_users_and_groups.yml
```diff
@@ -82,4 +90,5 @@
         {% if item.get('remove') %}--remove{% endif %}
         {% if item.get('superuser') %}--superuser{% endif %}
         {% if item.get('staff') %}--staff{% endif %}
+        {% if item.get('unusable_password') %}--staff{% endif %}
```
maxrothman: Do you want to be adding the `--staff` flag here?

kylecrawshaw: no. good catch. copy and paste error :(

maxrothman:
👍 
Nitpick: squash the commits in this branch

PR #3148, opened 2016-06-21
rewrite the tasks using YAML syntax

arbabnazar:

### DIFF
--- a/playbooks/roles/xqwatcher/tasks/deploy.yml
+++ b/playbooks/roles/xqwatcher/tasks/deploy.yml
@@ -1,18 +1,22 @@
-- name: install courses ssh key
-  copy: >
-    content="{{ XQWATCHER_GIT_IDENTITY }}"
-    dest={{ xqwatcher_app_dir }}/.ssh/{{ xqwatcher_service_name }}-courses
-    owner={{ xqwatcher_user }} group={{ xqwatcher_user }} mode=0600
+- name: Install courses ssh key
+  copy:
+    content: "{{ XQWATCHER_GIT_IDENTITY }}"
+    dest: "{{ xqwatcher_app_dir }}/.ssh/{{ xqwatcher_service_name }}-courses"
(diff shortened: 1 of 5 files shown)


playbooks/roles/xqwatcher/tasks/deploy.yml
```diff
@@ -1,18 +1,22 @@
-- name: install courses ssh key
-  copy: >
-    content="{{ XQWATCHER_GIT_IDENTITY }}"
-    dest={{ xqwatcher_app_dir }}/.ssh/{{ xqwatcher_service_name }}-courses
-    owner={{ xqwatcher_user }} group={{ xqwatcher_user }} mode=0600
+- name: Install courses ssh key
+  copy:
+    content: "{{ XQWATCHER_GIT_IDENTITY }}"
```
arbabnazar, line 3: build fail due to the value of `XQWATCHER_GIT_IDENTITY` which is `!!null`, @e0d @feanil is it normal or we can set it to `None`? Thanks

feanil, line 3: @arbabnazar I would expect that this was what was happening before.  Did the old code work when you tested it?

arbabnazar, line 3: @feanil I didn't make any change to the code, just use the YAML syntax, I am sure that it will even break with old code.

feanil, line 3: Right, so I think that you need to set this in order to test the role.  The easiest way to do this might be to build our stage AMI which has all those settings set.

arbabnazar, line 3: Thanks @feanil, will do that

arbabnazar, line 3: @feanil I have tested on the my sandbox and it worked, please review. Thanks

feanil:
:+1:

PR #3149, opened 2016-06-21
analyticstack enable yarn log retention

mulby:
Description:

When running Hadoop jobs on the analytics devstack (and single-instance install), YARN does not retain the raw logs produced by the python processes that are actually running the map and reduce function in the yarn containers. This makes debugging very difficult since you cannot see the stack traces produced by your code.

Solution:

By enabling yarn log aggregation, you can run `yarn logs -applicationId <your application ID>` after the job completes and it will dump all of the logs to stdout. This allows you to grep through the logs to find the relevant errors. See [the hortonworks guide](http://hortonworks.com/blog/simplifying-user-logs-management-and-access-in-yarn/) for more information.

Reviewers:
- [x] @feanil 
- [x] @brianhw 

FYI:
- [ ] @renevatium - since I suspect you are overriding this in your setup
- [ ] @sanfordstudent 
- [ ] @HassanJaveed84

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+- Role: hadoop_common
+  - Enable log retention by default to assist with debugging. Now YARN will retain stdout and stderr logs produced by map reduce tasks for 24 hours. They can be retrieved by running "yarn logs -applicationId YOUR_APPLICATION_ID".
+
 - Role: rabbitmq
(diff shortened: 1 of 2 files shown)


sanfordstudent:
!!!!

brianhw:
LGTM.  👍   

Should the instructions of how to get the logs be put somewhere?  I was thinking that maybe a comment with the settings would be useful, but really the information needs to be somewhere else.

feanil:
:+1:   once you add a CHANGELOG entry.

renevatium:
@mulby thank you for including me in this! Just hit this exact issue trying to debug ModuleEngagement.  Appreciate it.

PR #3154, opened 2016-06-22
fix the Failed to set permissions on the temporary files in Ansible

arbabnazar:

getting this error while setting permission on the tmp file:

`fatal: [x.x.x.x]: FAILED! => {"failed": true, "msg": "Failed to set permissions on the temporary files Ansible needs to create when becoming an unprivileged user. For information on working around this, see https://docs.ansible.com/ansible/become.html#becoming-an-unprivileged-user"}`

I was testing the roles with Ansible 2.1, after installing the `acl` package everything work and behind the scenes this package make sure those temporary files are handled securely and silently.

### DIFF
--- a/playbooks/roles/common_vars/defaults/main.yml
+++ b/playbooks/roles/common_vars/defaults/main.yml
@@ -102,8 +102,10 @@ common_redhat_pkgs:
   - rsyslog
   - git
   - unzip
+  - acl
 common_debian_pkgs:
   - ntp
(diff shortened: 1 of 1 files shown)


e0d:
@edx/devops This seems like the right thing to do, but we should consider whether there are any issues that might be introduced by relying on facls.

@smarnach FYI.

feanil:
:+1:

PR #3156, opened 2016-06-22
using YAML syntax to rewrite the ansible tasks

arbabnazar:


playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -64,73 +64,74 @@
...
     - install:configuration
   
-- name: install a bunch of system packages on which edx_service relies
+- name: Install a bunch of system packages on which edx_service relies
   apt:
-    pkg: "{{ item }}"
+    name: "{{ item }}"
     state: present
     update_cache: true
-    cache_valid_time: 3600
```
arbabnazar, line 108: remove the `cache_valid_time` may be in some of the role(s), which use this role as dependency add the repo and need to update the cache and this parameter conflict.

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -64,73 +64,74 @@
...
-  with_items: edx_service_packages.debian
+  with_items: "{{ edx_service_packages.debian }}"
   when: ansible_distribution in common_debian_variants
   tags:
     - install
     - install:system-requirements
 
-- name: install a bunch of system packages on which edx_service relies
-  yum: pkg={{ item }} state=present
-  with_items: edx_service_name.redhat
```
arbabnazar, line 117: From `defaults`:

```
 edx_service_packages:
      debian: []
      redhat: []
```

I think it's `edx_service_packages.redhat` instead of `edx_service_name.redhat` didn't notice because we are not using redhat often.

playbooks/roles/edx_service/tasks/main.yml
feanil, line 108: @arbabnazar can you clarify a bit for me.  Is the issue that other parts of the code are adding PPAs and this prevents them from updating the cache correctly?

arbabnazar, line 108: @feanil let assume we use this role as depenedency in `common` role and it update the cache and make it valid for one hour then we add the PPA for python and try to update the cache to get latest repo info but due to the fact that we make it valid for one hour the cache will not updated and will not install the latest package(s). This is the reason came into my mind and I have removed this parameter, on the other hand, this parameter is only use here but not anywhere else where are updating the cache. Thanks

feanil, line 108: Yea, this was put into the service play specifically as an optimization. I think the right thing to do is leave it here.  Any time we add a new repository, we should be running an update_cache as a part of that addition.  That update cache does not look at this cache valid time which is what you would want.

arbabnazar:
@feanil kindly check. Thanks

feanil:
:+1:

PR #3158, opened 2016-06-22
OPS-1601 As an operator, I want my Ansible callback plugins upgraded

arbabnazar:
@edx/devops Kindly review. Thanks

By remove `playbook_on_play_start`, it will work for both Ansible 1.9 and 2.x, once we'll migrate to Ansible 2.x, we'll modify it further using Ansible 2.x API.

### DIFF
--- a/playbooks/callback_plugins/datadog_tasks_timing.py
+++ b/playbooks/callback_plugins/datadog_tasks_timing.py
@@ -38,11 +38,6 @@ class CallbackModule(object):
     def clean_tag_value(self, value):
         return value.replace(" | ", ".").replace(" ", "-").lower()
 
-    def playbook_on_play_start(self, pattern):
-            self.playbook_name, _ = os.path.splitext(
-                os.path.basename(self.play.playbook.filename)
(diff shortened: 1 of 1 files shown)


playbooks/callback_plugins/datadog_tasks_timing.py
```diff
@@ -38,11 +38,6 @@ def __init__(self):
     def clean_tag_value(self, value):
         return value.replace(" | ", ".").replace(" ", "-").lower()
 
-    def playbook_on_play_start(self, pattern):
```
e0d, line 41: https://github.com/edx/configuration/blob/e6908618cacff088e76146a7b4c8e0d050c3c44d/playbooks/callback_plugins/datadog_tasks_timing.py#L78

Using here

feanil:
@arbabnazar has the interface for the function changed?  If not and it's just that the function is no longer supported, I would expect ansible 2.x to ignore any functions it doesn't care about.  We should add a TODO to remove this once we are in the 2.x world.

PR #3159, opened 2016-06-22
fix error - found a duplicate dict key (when).  Using last defined value only

arbabnazar:

Fix those error with this change in Ansible 2.x:

```
[WARNING]: While constructing a mapping from git-repos/edx/configuration/playbooks/roles/aws/tasks/main.yml,
line 151, column 3, found a duplicate dict key (when).  Using last defined value only.

 [WARNING]: While constructing a mapping from git-repos/edx/configuration/playbooks/roles/aws/tasks/main.yml,
line 159, column 3, found a duplicate dict key (when).  Using last defined value only.
```

### DIFF
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -153,13 +153,11 @@
     name: ssh
     state: restarted
   become: True
-  when: sshd_config.changed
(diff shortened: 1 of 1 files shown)


fredsmith:
👍

PR #3160, opened 2016-06-22
We have users who are absent and keyless

jibsheet:
We don't want to remove them from the users lists yet until we've used those config
files to remove them from all machines.


arbabnazar:
👍

maxrothman:
👍

PR #3162, opened 2016-06-22
Initial play to run Django migrations.

doctoryes:
Adding play that will run Django migrations in a release pipeline.

@edx/pipeline-team FYI

### DIFF
--- a/playbooks/continuous_delivery/run_migrations.yml
+++ b/playbooks/continuous_delivery/run_migrations.yml
@@ -1,0 +1,71 @@
+# This playbook will check for migrations that need to be run for Django applications within a larger
+# Django application. If migrations exist, it will run the migrations while saving the output as an artifact.
+#
+# The playbook uses the Django management commands found in this Django app repo:
+# https://github.com/edx/edx-django-release-util
+# So the Django app above needs to be installed in the Django app being checked for migrations.
+#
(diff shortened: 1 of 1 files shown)


doctoryes:
@edx/pipeline-team Ready for review.

doctoryes:
@edx/pipeline-team Please review - this is the Ansible script to run migrations.

playbooks/continuous_delivery/run_migrations.yml
```diff
@@ -0,0 +1,71 @@
...
+      - "{{ migration_output }}"
+
+  - name: Send Hipchat notification cleanup has finished
+    hipchat_2_0_0_1:
+      api: "{{ HIPCHAT_URL }}"
+      token: "{{ HIPCHAT_TOKEN }}"
+      room: "{{ HIPCHAT_ROOM }}"
+      msg: "Migrations have completed."
+    ignore_errors: yes
+    when: hipchat_token is defined
```
feanil: This should be `HIPCHAT_TOKEN is defined`.  I forgot to update it when I capatilazed the rest.

feanil:
One small change otherwise :+1: 

Reminder that this will work fine with IDAs but we will need to do some work to this to make it work for the LMS/CMS.  I think that can be looked at as separate work.

PR #3166, opened 2016-06-23
14.04 doesn't install httplib2 by default (12.04 does)

jibsheet:
Our user play now uses uri which requires httplib2. 
Fixes https://github.com/edx/configuration/pull/3144


jibsheet:
@clintonb FYI
@edx/devops review please, fixes AMI builds on 14.04

clintonb:
👍

arbabnazar:
👍

PR #3168, opened 2016-06-23
Handle the non-build-from-scratch case

jibsheet:
We only run the previous stanza if git and pip are missing.
So if you have an existing 14.04 image and cut a new one, you'll
still run into problems where ansible dies missing httplib2 and we
never install it without a scratch build.

@edx/devops 
@clintonb FYI
@mjfrey FYI


jibsheet:
@fredsmith Please give this a look

fredsmith:
👍 😞

PR #3171, opened 2016-06-24
fix the duplicate entry of rabbitmq_refresh

arbabnazar:

error track:

```
[WARNING]: While constructing a mapping from /var/tmp/extra-vars-18389.yml,
line 1, column 1, found a duplicate dict key (rabbitmq_refresh).  Using last
defined value only.
```

There were two entries in `util/jenkins/ansible-provision.sh`:

```
203:rabbitmq_refresh: True
321:rabbitmq_refresh: True
```


feanil:
:+1:

PR #3176, opened 2016-06-24
This needs to be installed on the servers not locally.

feanil:
This change is needed because we recently moved from the get_url module to the uri module.

@edx/devops


arbabnazar:
👍 @feanil can we add the `python-httplib2` package inside `common_vars` as well. we are using `uri` in many roles. Thanks

feanil:
@arbabnazar I'm going to merge this from now to fix this issue.  We talked about this at stand up today and the long term fix is going to be to put this in common and to make it so that the dependency between common and user is reversed.  Right now common depends on user so actually user is always going to get installed before common.

PR #3177, opened 2016-06-24
This breaks when you add a new person, but a previous person removed a key.

jibsheet:
Other than removing that user you have no recourse to add the new user.

We need a better warning mechanism for this, but we need to stop
failing.

@edx/devops


maxrothman:
👍

PR #3179, opened 2016-06-24

arbabnazar:


e0d:
:+1: 

@nedbat @mulby @macdiesel affects you, but should be backward compatible.  This is preparatory work toward Ansible 2.x.

PR #3181, opened 2016-06-25
rewrite ansible tasks using YAML style

arbabnazar:

Jenkins Build# 7972

### DIFF
--- a/playbooks/roles/xserver/tasks/deploy.yml
+++ b/playbooks/roles/xserver/tasks/deploy.yml
@@ -1,106 +1,140 @@
-- name: "writing supervisor script"
-  template: >
-    src=xserver.conf.j2 dest={{ supervisor_available_dir }}/xserver.conf
-    owner={{ supervisor_user }} group={{ common_web_user }} mode=0644
+- name: Writing supervisor script
+  template:
+    src: xserver.conf.j2
+    dest: "{{ supervisor_available_dir }}/xserver.conf"
+    owner: "{{ supervisor_user }}"
+    group: "{{ common_web_user }}"
+    mode: "0644"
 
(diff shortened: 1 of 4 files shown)


playbooks/roles/xserver/tasks/deploy.yml
```diff
@@ -1,106 +1,140 @@
...
-  template: >
-    src=git_ssh.sh.j2 dest=/tmp/git_ssh.sh
-    owner={{ xserver_user }} mode=750
-  notify: restart xserver
+- name: Upload ssh script
+  template:
+    src: git_ssh.sh.j2
+    dest: "/tmp/git_ssh.sh"
+    owner: "{{ xserver_user }}"
+    mode: "750"
```
e0d: make 0750

playbooks/roles/xserver/tasks/main.yml
```diff
@@ -2,37 +2,39 @@
 # As of right now this role requires
 # access to the edX 6.00x repo which is not public
 ---
-
-- name: checking for grader info
-  fail: msg="You must define XSERVER_GRADER_DIR"
+- name: Checking for grader info
+  fail:
+    msg: "You must define XSERVER_GRADER_DIR"
   when: XSERVER_GRADER_DIR|length == 0
```
e0d: I think that these checks are no longer necessary, can you confirm that undefined vars will result in errors?

e0d:
:+1:  @edx/devops second review?

playbooks/roles/xserver/tasks/main.yml
arbabnazar: @e0d yes it is giving an error if these variable(s) is/are not defined. Thanks

PR #3182, opened 2016-06-26
Sometimes abbey builds fail waiting on the status checks

jibsheet:
They fail with

  Waiting for system status:               An error occurred building
  AMI for loadtest edx edxapp.  The Exception was IndexError('list index
  out of range',)

I'm guessing that it is this list index lookup, seems likely.  Hard
to test as it requires AWS slowness.  Normal builds continue unimpeded, but I've seen this particular list index error 3 or 4 times this week, so this seems like an ok guard to add.

@edx/devops


feanil:
:+1:

PR #3187, opened 2016-06-28
use YAML syntax for ansible tasks

arbabnazar:

### DIFF
--- a/playbooks/roles/rabbitmq/tasks/main.yml
+++ b/playbooks/roles/rabbitmq/tasks/main.yml
@@ -3,136 +3,169 @@
 # There is a bug with initializing multiple nodes in the HA cluster at once
 # http://rabbitmq.1065348.n5.nabble.com/Rabbitmq-boot-failure-with-quot-tables-not-present-quot-td24494.html
 
-- name: install python-software-properties if debian
-  apt: pkg={{ ",".join(rabbitmq_debian_pkgs) }} state=present
+- name: Install python-software-properties if debian
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ rabbitmq_debian_pkgs }}"
 
-- name: fetch the rabbitmq server deb
-  get_url: >
-    url={{ rabbitmq_pkg_url }}
-    dest=/var/tmp/{{ rabbitmq_pkg_url|basename }}
+- name: Fetch the rabbitmq server deb
+  get_url:
+    url: "{{ rabbitmq_pkg_url }}"
+    dest: "/var/tmp/{{ rabbitmq_pkg_url | basename }}"
 
-- name: check if rabbit is installed
-  shell: >
-    dpkg -s rabbitmq-server >/dev/null 2>&1 || echo "not installed"
+- name: Check if rabbit is installed
+  shell: 'dpkg -s rabbitmq-server >/dev/null 2>&1 || echo "not installed"'
   register: is_installed
 
-- name: install rabbit package using gdebi
-  shell: >
-    gdebi --n {{ rabbitmq_pkg_url|basename }}
-    chdir=/var/tmp
+- name: Install rabbit package using gdebi
+  shell: "gdebi --n {{ rabbitmq_pkg_url|basename }}"
+  args:
+    chdir: /var/tmp
   when: is_installed.stdout is defined and is_installed.stdout == "not installed"
 
-- name: stop rabbit cluster
-  service: name=rabbitmq-server state=stopped
+- name: Stop rabbit cluster
+  service:
+    name: rabbitmq-server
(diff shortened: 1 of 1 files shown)


playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -3,149 +3,183 @@
...
+    src: "edx/app/rabbitmq/log-rabbitmq-queues.sh.j2"
+    dest: "{{ rabbitmq_app_dir }}/log-rabbitmq-queues.sh"
+    owner: "{{ rabbitmq_user }}"
+    group: "{{ rabbitmq_group }}"
+    mode: "0755"
 
-- name: set up a cron job to run the script
+- name: Set up a cron job to run the script
   cron:
     name: "log-queue-lenghts"
```
e0d, line 57: Please fix type while you are here.

playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -3,149 +3,183 @@
...
+    mode: "0644"
   register: cluster_configuration
 
-- name: install plugins
+- name: Install plugins
   rabbitmq_plugin:
-    names={{",".join(rabbitmq_plugins)}} state=enabled
+    names: "{{ item }}"
+    state: enabled
+  with_items: "{{ rabbitmq_plugins }}"
```
e0d, line 108: Please verify that this works as expected and also doesn't make the play much slower to run.

playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -3,149 +3,183 @@
...
+    path: "{{ rabbitmq_mnesia_folder }}"
+    state: absent
   when: erlang_cookie.changed or cluster_configuration.changed or rabbitmq_refresh
 
-- name: start rabbit nodes
-  service: name=rabbitmq-server state=restarted
+- name: Start rabbit nodes
+  service:
+    name: rabbitmq-server
+    state: restarted
```
e0d: Should this just be "started" as we stop above unconditionally?  I assume it throws a warning as it cannot stop rabbit.

playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -3,149 +3,183 @@
...
-    dest=/usr/local/bin/rabbitmqadmin
-
-- name: ensure rabbitmqadmin attributes
-  file: >
-    path=/usr/local/bin/rabbitmqadmin owner=root
-    group=root mode=0655
+- name: Install admin tools
+  get_url:
+    url: "http://localhost:{{ rabbitmq_management_port }}/cli/rabbitmqadmin"
+    dest: "/usr/local/bin/rabbitmqadmin"
```
e0d, line 179: Can we add a creates declaration here?

e0d:
A few comments, but after I'm :+1: @edx/devops second review?

playbooks/roles/rabbitmq/tasks/main.yml
arbabnazar, line 179: we don't need `creates` here becuase `get_url` is already idempotent

playbooks/roles/rabbitmq/tasks/main.yml
arbabnazar, line 108: @e0d it doesn't make it slow, I have verified the time from the datadog plugin and no noticeable time difference

e0d:
My comments have been addressed, would still like a second review on this one.  @edx/devops

playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -3,149 +3,184 @@
...
+    dest: "{{ rabbitmq_app_dir }}/log-rabbitmq-queues.sh"
+    owner: "{{ rabbitmq_user }}"
+    group: "{{ rabbitmq_group }}"
+    mode: "0755"
 
-- name: set up a cron job to run the script
+- name: Set up a cron job to run the script
   cron:
-    name: "log-queue-lenghts"
+    name: "log-queue-lengths"
```
jibsheet: Are we going to end up with 2 cron jobs on our existing rabbit servers if we rerun the play (which we do sometimes to update configuration)

jibsheet:
My one question about the cron jobs, but everything else looks 👍

playbooks/roles/rabbitmq/tasks/main.yml
arbabnazar: @e0d , @jibsheet raise a very good question, I think we need to leave it as misspelled now? or we can do it like this that check the wrong job name and delete it and add the new one. Please let me know about your thoughts. Thanks

PR #3188, opened 2016-06-28
Only write out when there is a key

e0d:
@arbabnazar


feanil:
:+1:

playbooks/roles/xqwatcher/tasks/deploy.yml
```diff
@@ -5,6 +5,7 @@
     owner: "{{ xqwatcher_user }}"
     group: "{{ xqwatcher_user }}"
     mode: "0600"
+  when: XQWATCHER_GIT_IDENTITY is defined
```
feanil: Because there is a default value of `!!null` we should set this to `XQWATCHER_GIT_IDENTITY`  so if it has a non falsy value it will do stuff but the `!!null` should convert to a python `None` which should be false.

feanil:
@arbabnazar makes a good point in PR 3189, I think the best solution though is what I suggested in the diff of checking to see if the var has any content.

arbabnazar:
@e0d to make it work, we need to change the `XQWATCHER_GIT_IDENTITY: !!null` to `XQWATCHER_GIT_IDENTITY: none`

arbabnazar:
👍

feanil:
:+1:

PR #3189, opened 2016-06-28
fix the task fail condition

arbabnazar:
@edx/devops using this approach because this [PR](https://github.com/edx/configuration/pull/3188) is not working. THanks


arbabnazar:
@e0d can you please check this. Thanks

PR #3118, opened 2016-06-15
Add playbook to deploy stateless dogwood instances

omarkhan:
We currently use the `edx_sandbox.yml` playbook to deploy all our sandboxes and production instances. While this kind of works, the playbook was never meant to be used to install stateless production instances, and we've already had upstream contributions rejected because they weren't in line with the intended use of the playbook. Moreover, the playbook contains more services than we need on a typical production instance.

This pull request adds a new `edx_production.yml` playbook aimed specifically at our use case: Deploy a production dogwood instance that uses existing MySQL and MongoDB services.
#### Differences with the `edx_sandbox.yml` playbook
- Services are split into separate plays with different group names, so that service deployment can be defined in inventory.
- The `mysql`, `edxlocal` and `mongo` roles are excluded. It is expected that production deployments will use external databases that have been configured separately.
- The `analytics_api`, `insights`, `edx_notes_api`, `demo`, `edx_ansible` roles are excluded.

PR #3191, opened 2016-06-28
WIP: Feanil/small deployments

feanil:


feanil:
@bradenmacdonald FYI, I'm not going to have time to work on this in the near future.  The current state is that the two machines come up and ansible runs cleanly but some services can't reach from the stateless machine to the stateful machine.

bradenmacdonald:
@feanil Too bad. Thanks for the update. Between this and #3118, I'll see what we can do to get a working playbook.

danielmcquillen:
@bradenmacdonald @feanil I was able to use these playbooks with good success on aws. A quick manual change to /etc/memcached.conf on datastores resolved the inaccessibility issue there. I found that by enabling the following ports between the security groups for stateless and datastore things seemed to work ok:  11211 memcached, 5672 for rabbitmq, 27017 for mongo db and 9200 for elasticsearch. @feanil was there something else that you thought needed resolving in terms of access between layers?
(I hope these playbooks get pushed forward as they're really useful for small sites like the one I'm working on...I hope I can contribute at some point. I added some other observations/questions [on the wiki](https://openedx.atlassian.net/wiki/display/OpenOPS/Hosting+edX+in+Production).)

danielmcquillen:
@bradenmacdonald just curious, do you plan on expanding on Feanil's work?

bradenmacdonald:
We have a need for this, so we will eventually help out in some way, whether that's finishing this, testing it, or converting our similar WIP PR.

@feanil are you planning to revisit this any time soon?

feanil:
Unforutnately, I have not been able to get back to this and I think that will stay the case for a little while.  I'm about to head of on paternity leave so I don't think I'll be able to make progress on this.

bradenmacdonald:
@feanil Ok, we will try to finish it up then. Thanks for the update.

danielmcquillen:
Congrats @feanil !

feanil:
Gonna close this as it is out of date and most of the useful things have landed I believe.

PR #3192, opened 2016-06-28
Go back to the default method of stopping workers

nedbat:
We had a persistent problem of workers getting stuck when trying to stop
during 12.04 native installation.  This fixes the problem, while still
allowing active tasks to complete.

This reverts pull request #2133, and removes the stopasgroup setting that
was causing the trouble.


nedbat:
@feanil @cpennington

nedbat:
I don't understand the implications of this comment on the Celery bug: https://github.com/celery/celery/issues/2700#issuecomment-229210623  Should we be doing something different?

jmbowman:
The configuration before this PR basically says "wait for up to `stopwaitsecs` for already-running tasks to finish up, then resort to killing them if necessary".  The referenced celery issue comment says that a TERM signal was failing to do the graceful shutdown because of the [stopasgroup=true](https://groups.google.com/forum/#!topic/celery-users/t8g5KvIvQZ8) entry in the supervisor configuration (which passed the TERM signal along to child processes before the main process could try to shut them down gracefully).  Switching to INT instead was a means of preserving that setting while getting the desired shutdown behavior.  Without [killasgroup=true](http://supervisord.org/configuration.html#program-x-section-values) (which is implied by `stopasgroup=true`), a kill signal sent to the main process fails to also kill the child processes.

It seems like the real problem is that we run some really long task during installation?  If we really want to both do that and then shut the workers down even if the task is still in progress, we probably need to just send the KILL signal in the first place.

nedbat:
During installation, there are no tasks in the worker queues (at least, I don't know of any, and can't imagine what long-running tasks would be in the queue during a fresh installation.)

jmbowman:
I think we'd be ok without `stopsignal=INT` as long as we set `killasgroup=true` instead of `stopasgroup=true`.  Would be nice to get a second opinion from devops, though.

nedbat:
More discussion with the celery developers (https://github.com/celery/celery/issues/2700#issuecomment-229492218 and following) found the right solution, which I think @jmbowman was trying to point us to: stopasgroup=true caused the TERM signal to be too aggressive.  We can use TERM as long as we don't use stopasgroup.

feanil:
:+1:

PR #3193, opened 2016-06-29
[WIP] update nginx conf [OPS-1499]

bilalahmad99:
WIP for OPS-1499


playbooks/roles/credentials/defaults/main.yml
```diff
@@ -147,7 +147,10 @@ CREDENTIALS_GUNICORN_EXTRA: ""
 CREDENTIALS_GUNICORN_EXTRA_CONF: ""
 CREDENTIALS_GUNICORN_WORKER_CLASS: "gevent"
 
-CREDENTIALS_HOSTNAME: '~^((stage|prod)-)?credentials.*'
+CREDENTIALS_HOSTNAME:
+  - '.credentials.edx.org'
```
e0d: I think for the default case we should just convert this to a list of one item with the current value.

For our case, we'll override the values in our secrets for each environment and deployment.

e0d:
@edx/devops second review please?

@smarnach can yourself or a delegate have a look at this proposed change?

e0d:
@clintonb @feanil notice that the IDA template and our other IDAs have diverged in terms of where the HOSTNAME is defined.  For most IDAs it's defined in the nginx role, but for templated IDAs it's moved to their defaults.  

From a general perspective, I prefer that, but we've only done it half way in terms of making IDAs own their nginx config.

Thoughts on whether we should move all config to nginx defaults or live with the differences for now since directionally that's probably where we want to go?

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -80,10 +80,28 @@ NGINX_EDXAPP_ERROR_PAGES:
...
-CREDENTIALS_HOSTNAME: '~^((stage|prod)-)?credentials.*'
+LMS_HOSTNAMES:
+  - '.courses.example.com'
+CMS_HOSTNAMES:
+  - '.studio.example.com'
+ECOMMERCE_HOSTNAMES:
+  - '.ecommerce.example.com'
+PROGRAMS_HOSTNAMES:
+  - '.programs.example.com'
+CREDENTIALS_HOSTNAMES:
```
e0d: This var is duplicated in the role.

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -80,10 +80,28 @@ NGINX_EDXAPP_ERROR_PAGES:
   "502": "{{ nginx_default_error_page }}"
   "504": "{{ nginx_default_error_page }}"
 
-CMS_HOSTNAME: '~^((stage|prod)-)?studio.*'
-ECOMMERCE_HOSTNAME: '~^((stage|prod)-)?ecommerce.*'
-PROGRAMS_HOSTNAME: '~^((stage|prod)-)?programs.*'
-CREDENTIALS_HOSTNAME: '~^((stage|prod)-)?credentials.*'
+LMS_HOSTNAMES:
+  - '.courses.example.com'
```
e0d: I think for the default case, we should remove the leading dot `.`

feanil:
I think we should live with the difference now and plan some time to clean-up the ansible so that older IDAs move towards matching the cookiecutter.  Are there tickets for the work to clean-up the NGINX work for older IDAs?

playbooks/roles/nginx/defaults/main.yml
bilalahmad99: done

playbooks/roles/nginx/defaults/main.yml
bilalahmad99: done

smarnach:
@e0d I've been on leave for the last month, so I missed your ping.  If I can find time to do so, I'll take a look at this PR in the next couple of days.

clintonb:
Oops. I was also on vacation when I was pinged. I second @feanil's suggestion to leave the existing IDAs untouched. I really want to get everything aligned; but, we should make the time to plan it in a sprint.

e0d:
@smarnach let me know when you think you'll be able to have a look at this.

PR #3195, opened 2016-06-29
Create an ansible script for running populate_model.

cahrens:
Initially used for xblock configurations, but can be used for any
ConfigurationModel.

Please review @maxrothman

### DIFF
--- a/playbooks/edx-east/populate_configuration_model.yml
+++ b/playbooks/edx-east/populate_configuration_model.yml
@@ -1,0 +1,46 @@
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+# Usage: ansible-playbook -i lms-host-1, -e "file=/path/to/json/file" -e "user=username"
+#
+# Overview:
+# This executes the "populate_model" management command to populate a ConfigurationModel with
+# data from the supplied JSON file.
+#
+# The username of an existing user must be specified to indicate who is performing the operation.
+#
+# JSON file format should be:
+#
+#    { "model": "config_models.ExampleConfigurationModel",
+#      "data":
+#        [
+#          { "enabled": True,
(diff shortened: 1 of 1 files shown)


playbooks/edx-east/populate_configuration_model.yml
```diff
@@ -0,0 +1,45 @@
...
+  vars:
+    python_path: /edx/bin/python.edxapp
+    manage_path: /edx/bin/manage.edxapp
+  tasks:
+    - name: Manage xblock configurations
+      shell: >
+        {{ python_path }} {{ manage_path }} lms --settings=aws
+        populate_model -f {{ file | quote }} -u {{ user }}
+      register: command_result
+      failed_when: "'Import complete' not in command_result.stdout"
```
cahrens: I tested that this is working, but is there a way to print out command_result.stdout when it succeeds? It would be nice to know how many entries were created.

The management command prints "Import complete, {0} new entries created").format(created_entries))

cahrens:
I was able to run this on a sandbox and populate 2 different configuration models, using JSON files from https://github.com/edx/xblock-configuration.

playbooks/edx-east/populate_configuration_model.yml
maxrothman: You can use the debug module to print arbitrary text.

I think that rather than determining failure from checking output, it'd be better to rely on the exit code of the command (which is ansible's default behavior). If the management command doesn't exit with non-zero exit code on failure, that should be fixed.

Also, it'd be good to use `changed_when` to let ansible show an "OK - unchanged" status when "new entries created" is 0.

cahrens: Ah, I got confused about changed_when vs. failed_when. I will delete failed_when and add changed_when.

cahrens:
@maxrothman Updates are made. This PR failed due to timeout the last time it ran; I assume this has nothing to do with my new file, but let me know if I'm wrong!

maxrothman:
Double check that on failures it doesn't report "changed" because of the `changed_when`, then I'm 👍 

You can ignore the timing-out tests. They don't test this playbook.

cahrens:
@maxrothman It does not report "changed" when it fails:

image

maxrothman:
👍

cahrens:
Merging per @maxrothman comment about timeout failure not being a concern.

PR #3197, opened 2016-06-30
Arbab/user role fix

arbabnazar:
@feanil @e0d please review. `ansible_distribution in common_debian_variants` broken the `ansible-grant-ssh-access` job.

Added the `common_vars` role as meta dependencies because it contain the `common_debian_variants` and enable the `gather_facts` in the playbook to find the `ansible_distribution` so that condition will evaluate successfully.

This change fixed that. Thanks

@edx/devops


e0d:
:+1:  after tests finish

@feanil FYI

PR #3198, opened 2016-06-30
DEVOPS-4424 make Discovery JWT overridable

fredsmith:


mattdrayer:
Will this enable support for multiple issuers?  We have three sites for which we'll need to be able to create dynamic coupons -- and thus, interact with the discovery service.

@clintonb FYI

mattdrayer:
👍

feanil:
:+1:  after you make a changelog entry

PR #3205, opened 2016-07-01
Specify a socket timeout for pymongo and lower the connect timeout.

jibsheet:
Based on the work in https://openedx.atlassian.net/browse/PLAT-1014

@ormsbee Do you buy my argument in the comments, you get 3 seconds for a query to complete, 2 seconds to connect?  We don't get many slow queries from the modulestore, and most of them are from the read replica.  This 3 second timeout is mongo-is-silent from me reading the doc, so even if it takes time to stream us a 5mb document, we should be fine as long as querying the document and BSONing it doesn't take 3 seconds.

Willing to go lower, but wanted to find somewhere to start.

@doctoryes FYI since this would cause us to throw a new exception pymongo.errors.NetworkTimeout which we may want to handle explicitly in the platform, we don't seem to do so right now.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -681,6 +681,11 @@ edxapp_generic_doc_store_config: &edxapp_generic_default_docstore
   user: "{{ EDXAPP_MONGO_USER }}"
   collection:  'modulestore'
   ssl: "{{ EDXAPP_MONGO_USE_SSL }}"
+  # https://api.mongodb.com/python/2.9.1/api/pymongo/mongo_client.html#module-pymongo.mongo_client
+  socketTimeoutMS: 3000 # default is never timeout while the connection is open, this means it needs to explicitly close raising pymongo.errors.NetworkTimeout
+  connectTimeoutMS: 2000 # default is 20000, I believe raises pymongo.errors.ConnectionFailure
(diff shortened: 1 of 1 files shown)


ormsbee:
@jibsheet: Yeah, that seems reasonable. 👍

doctoryes:
@jibsheet Will these timeouts show up in the MongoDB logs? I'd like to keep an eye on how often they happen - whether it's only during a primary stepdown -or- more regularly. If this is an error that happens on a regular basis, the error should be handled and tests should be added to define/enforce how the platform reacts to the timeouts.

doctoryes:
👍 contingent on some sort on monitoring of how often this timeout occurs.

jibsheet:
@doctoryes I would expect them to show up in lms/edx.log, but yes, we should be able to splunk for ConnectionFailure and then prune for pymongo.

There may also be mongo errors (tried to write to a closed socket) in the case of network blips, but I think we're going to have to monitor for those.

jibsheet:
@edx/devops I'm going to cut an AMI with this and look at it, but comments on the new defaults appreciated.

jibsheet:
@doctoryes We have no hits in the last month on NetworkTimeout, so should be easy to monitor.  We do have hits on ConnectionFailure, but those are explainable (all during the master step down).

I'll save a search and we can check it during the deploy and then after.

feanil:
:+1:

PR #3209, opened 2016-07-04
proper YAML syntax for rewriting ansible tasks

arbabnazar:

### DIFF
--- a/playbooks/roles/minos/tasks/main.yml
+++ b/playbooks/roles/minos/tasks/main.yml
@@ -66,19 +66,11 @@
     - "RolledTrackingLogVoter"
 
 # Optional auth for git
-- name: Create ssh script for git (not authenticated)
+- name: Create ssh script for git
   template:
     src: "tmp/git-identity.sh.j2"
     dest: "{{ minos_git_ssh }}"
     mode: "0750"
-  when: MINOS_GIT_IDENTITY is not defined
-
-- name: Create ssh script for git (authenticated)
-  template:
-    src: "tmp/git-identity.sh.j2"
-    dest: "{{ minos_git_ssh }}"
-    mode: "0750"
-  when: MINOS_GIT_IDENTITY is defined
 
 - name: Install read-only ssh key
   copy:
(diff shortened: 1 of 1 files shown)


playbooks/roles/minos/tasks/main.yml
```diff
@@ -35,61 +35,65 @@
...
-    mode=750
-  when:
-    - MINOS_GIT_IDENTITY is not defined
+# Optional auth for git
+- name: Create ssh script for git (not authenticated)
+  template:
+    src: "tmp/git-identity.sh.j2"
+    dest: "{{ minos_git_ssh }}"
+    mode: "0750"
+  when: MINOS_GIT_IDENTITY is not defined
```
jibsheet: Not part of your cleanup, but these tasks appear identical (and the template already checks for MINOS_GIT_IDENTITY) maybe collapse these without the whens?

the not authenticated and authenticated cases for clarity

jibsheet:
👍 with my note about squishing two tasks

playbooks/roles/minos/tasks/main.yml
arbabnazar: @jibsheet when statement is there, sorry If I misunderstood. Can you please explain a bit, may be I got it wrong. Thanks

jibsheet: Sorry, I meant the two tasks look identical.  One run when defined, one when not defined.  But either way, they do the same thing.  I can't tell if I'm misreading the code, or if they're identical and we should have 1 task without the when.  Does that make sense?

arbabnazar: Thanks @jibsheet, we can ask from @e0d and @feanil, they can tell us.

arbabnazar: @e0d @feanil can you please guide here. Thanks

arbabnazar:
@jibsheet I have addressed your comment and removed the identical tasks and when, please review. Thanks

CC: @edx/devops

jibsheet:
👍

PR #3213, opened 2016-07-04
hide the log for ssl key

arbabnazar:

The idea is to hide the ssl private key to be print on the jenkins console output.


jibsheet:
This is needed because in verbose mode it shows the whole key, right?  Not in normal operation.
👍

arbabnazar:
@jibsheet it shows in both cases I think.

jibsheet:
Ah - I checked a sandbox build and didn't see it, so assumed it must be on -v 
http://jenkins.edx.org:8080/job/ansible-provision/8101/consoleFull for example

PR #3214, opened 2016-07-05
Refactor previous script for readability and commenting

michaelroytman:


util/balancecontainers.py
```diff
@@ -3,66 +3,92 @@
...
         line = line.strip("[]")
 
         items = line.split()
         containers.extend(items)
 
-    shards = balancer.pack_containers(containers)
+    # find optimal packing of the containers amongst shards
+    shards = pack_containers(containers, args.num_shards)
 
+    # print space separate list of containers for each shard
```
jibsheet: nit: separated

util/balancecontainers.py
```diff
@@ -3,66 +3,92 @@
...
 if __name__ == '__main__':
 
-    balancer = ContainerBalancer()
+    args = arg_parse()
 
     containers = []
 
+    # get containers from standard in
     for line in sys.stdin:
+        # strip whitespace and brackets
```
jibsheet: I don't think we need this comment, unless you change it to explain why you're stripping them (ie. travis provides them)

util/balancecontainers.py
```diff
@@ -3,66 +3,92 @@
...
-    def pack_containers(self, containers):
+    # open config file containing container weights
+    config_file_path = pathlib2.Path(CONFIG_FILE_PATH)
 
-        num_shards = 3
+    with (config_file_path.open(mode='r')) as file:
+        try:
+            config = yaml.load(file)
+        except yaml.YAMLError, exc:
+            LOGGER.warning("error in configuration file: %s" % str(exc))
```
jibsheet: Should this be error rather than warning?

jibsheet:
Some really minor notes to review, but then 👍

PR #3216, opened 2016-07-05
Renaming the index alias

e0d:
@clintonb

### DIFF
--- a/playbooks/roles/discovery/defaults/main.yml
+++ b/playbooks/roles/discovery/defaults/main.yml
@@ -41,7 +41,7 @@ DISCOVERY_DATABASES:
 # Using AWS? Use the AWS-provided host (e.g. https://search-test-abc123.us-east-1.es.amazonaws.com/).
 # Assuming the port corresponds that of the protocol (e.g. http=80, https=443), there is no need to specify a port.
 DISCOVERY_ELASTICSEARCH_URL: 'http://127.0.0.1:9200/'
-DISCOVERY_ELASTICSEARCH_INDEX_NAME: 'discovery'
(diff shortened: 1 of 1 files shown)


jibsheet:
👍

clintonb:
👍

e0d:
@MichaelRoytman this built the edxapp container, which surprised me, can you take a look?

michaelroytman:
@e0d 

The discovery role was modified (file path _playbooks/roles/discovery/defaults/main.yml_ was classified as being relevant to the discovery role). In the _playbooks/roles/discovery/meta/main.yml_ file, the discovery role has the following dependencies: common, supervisor, and edx_service.

The following expansion takes place from the set (discovery, common, supervisor, edx_service) based on their dependencies:
- common expands to common_vars, server_utils, user, security
- supervisor expands to common
- edx_service expands to add_user, git_clone

The set is now (discovery, common, supervisor, edx_service, common_vars, server_utils, security, add_user, git_clone).

The new additions (common_vars, server_utils, user, security, add_user, git_clone) to the set do not expand further, so the final expanded set of roles is (discovery, common, supervisor, edx_service, common_vars, server_utils, security, add_user, git_clone).

The script then looks through all of the Docker plays that contain at least one reference to one or more of the roles in this set. In the case of edxapp, common_vars is a role in _docker/plays/edxapp.yml_, so its corresponding edxapp Dockerfile was built.

**TL;DR:** 

The discovery role has a dependency on the common role. The common role has a dependency on common_vars. The edxapp Docker playbook contains a reference to the common_vars role, so the edxapp Dockerfile was built.

Please let me know if this contains incorrect logic.

PR #3221, opened 2016-07-05
Updated programs JWT configuration

clintonb:
Similar to the catalog/discovery role, programs now sends an array of sets an array of JWT issuers.
ECOM-4900


clintonb:
@bderusha please review

clintonb:
FYI @edx/devops 

I have confirmed this works on stage.

PR #3223, opened 2016-07-06
Use logrotate to clean up from the log-queue-lenghts (sic) job

jibsheet:
Right now these files never go away, and some are a million lines long.
This means splunk eats all the disk IO if we ever restart.


fredsmith:
👍

PR #3224, opened 2016-07-06
adding the leading zero in mode leaving off will likely have unexpect…

arbabnazar:

Ref: https://github.com/edx/configuration/pull/2464


jibsheet:
👍

PR #3225, opened 2016-07-06
Added additional JWT issuer for programs

clintonb:
ECOM-4900


clintonb:

This is necessary to ensure backward-compatibility with the existing JWT issuer that uses the OAuth secret for signing. This **MUST** be merged before the next production deployment.

rlucioni:
👍, glad you caught this before a deployment.

rlucioni:
@clintonb btw, once this is deployed and https://github.com/edx/edx-platform/pull/12954 is completed, it should be easy to switch programs over to our now-standard JWT so that we can remove this.

PR #3230, opened 2016-07-07
Change logger warning to error when error parsing configuration file …

michaelroytman:
…and update logger disablement level

### DIFF
--- a/util/parsefiles.py
+++ b/util/parsefiles.py
@@ -146,7 +146,7 @@ def _open_yaml_file(file_str):
             yaml_file = yaml.load(file)
             return yaml_file
         except yaml.YAMLError, exc:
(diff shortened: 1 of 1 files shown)


util/parsefiles.py
```diff
@@ -331,7 +331,7 @@ def arg_parse():
     logging.basicConfig()
 
     if not args.verbose:
-        logging.disable(logging.WARNING)
+        logging.disable(logging.ERROR)
```
jibsheet, line 335: I don't think we want to do this.
We always want to know when we get the config file wrong, and this means the rest of the LOGGER.warnings will always emit, which I don't think we want except in verbose mode.

jibsheet:
👍

PR #3232, opened 2016-07-08
Update shell prompt to show database name

jibsheet:
We use the same replica set for forums and edxapp, so showing that is
confusing.  Instead, show the DB.

@edx/devops


maxrothman:
👍

PR #2723, opened 2016-01-25
Add openstack role

omarkhan:
**Background:** many playbooks use S3, for:
- Logs
- Mongo backups
- Antivirus scans of S3 buckets
- Nginx maintenance page

This pull request adds an `openstack` role that will replace S3 with OpenStack Swift:
- The logrotate sync has been adapted from the `aws` role
- Adds swift ansible vars

**Testing:** deploy https://github.com/edx/edx-platform/pull/11286

**See also:**
- https://github.com/edx/edx-platform/pull/11286
- https://github.com/edx/xqueue/pull/103

PR #3233, opened 2016-07-08
Minos role also needs updates for the storage refactors.

jibsheet:
@edx/devops 

FYI @jcdyer 

Updates for https://github.com/edx/configuration/pull/2723


e0d:
:+1:

PR #3234, opened 2016-07-08
Themes need THEMES_CODE_DIR defined to something that exists

jibsheet:
It doesn't matter what it is or what is in it, it just needs to exist
on disk so that platform code can touch it.

@fredsmith Look at this amazing thing.


fredsmith:
👍

PR #3235, opened 2016-07-08
E0d/revert removal of dir creation

e0d:
@jibsheet @bradenmacdonald @feanil 

I have no idea why this was removed, but it's definitely breaking things.


bradenmacdonald:
I think the intention with this was to extract this to the `vhost` role, and add `vhost` as a dependency for aws and openstack roles.

Each role defines `VHOST_NAME` to be either `aws` or `openstack` then the `vhost` role creates those dirs: https://github.com/edx/configuration/blob/924a7a0/playbooks/roles/vhost/defaults/main.yml and https://github.com/edx/configuration/blob/924a7a0/playbooks/roles/vhost/tasks/main.yml#L25

Not sure why that's not working? Perhaps `VHOST_NAME` is only set in `defaults` and that doesn't override the default in the other role, so it's creating `vhost` folders instead of `aws` folders?

PR #3244, opened 2016-07-12
Missed rename of a file

jibsheet:
https://github.com/edx/configuration/pull/2723

@bradenmacdonald FYI - more grep fail, this was renamed, but was easily greppable.  

@edx/devops Please review

### DIFF
--- a/playbooks/edx-east/retire_host.yml
+++ b/playbooks/edx-east/retire_host.yml
@@ -38,7 +38,7 @@
         - "edx-services"
     # This catches the case where tracking.log is 0b
     - name: Sync again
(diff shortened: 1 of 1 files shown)


jibsheet:
@bradenmacdonald And the second commit in the branch is more breakage, grep + change to the script where it wasn't checked that the script was called the same way everywhere.

This BTW is now breaking stage and blocking the release.

FYI @jcdyer

fredsmith:
👍

bradenmacdonald:
Argh :-/ Thanks for fixing this and #3245, @jibsheet.

PR #3246, opened 2016-07-12
Added in install PHP and drush to go-agent

Dillon-Dumesnil:
@edx/pipeline-team please review.
Will be adding in a stage to pipeline to get access to a tar file so drush has access to the different aliases (edx.test, edx.prod, etc.). Out of scope for Dockerfile.

### DIFF
--- a/docker/build/go-agent/Dockerfile
+++ b/docker/build/go-agent/Dockerfile
@@ -37,6 +37,16 @@ RUN apt-get update && apt-get install -y -q \
     python-pip \
     libmysqlclient-dev
 
+# Install php
+RUN apt-get update && apt-get install -y \
+    php5-common \
+    php5-cli
+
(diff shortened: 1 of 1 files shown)


doctoryes:
@Dillon-Dumesnil I assume that the version of PHP provided with the go-agent's OS is fine?

@edx/pipeline-team A rhetorical question: At what point do we have more than one flavor of go-agent and tag them with different resource tags? Perhaps we'll just know when we reach that point - different OS required or the like...

Anyway, 👍

feanil:
:+1:

Dillon-Dumesnil:
Feanil said he does not believe that go-agent is initialized with PHP which is why I am installing it. Is there a way I can test and verify whether it does start with it or not?

feanil:
I think @doctoryes 's question is more around version compatibilty between the version that the image installs by default and the one you want.  You should be able to build the container locally to test it.

PR #3248, opened 2016-07-14
(WIP) -- WL-300: Remove all usages of USE_CUSTOM_THEME and THEME_NAME

saleem-latif:
@mattdrayer , @fredsmith -- FYI

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -221,7 +221,6 @@ EDXAPP_FEATURES:
   ENABLE_INSTRUCTOR_ANALYTICS: false
   PREVIEW_LMS_BASE: "{{ EDXAPP_PREVIEW_LMS_BASE }}"
   ENABLE_GRADE_DOWNLOADS: true
-  USE_CUSTOM_THEME: "{{ edxapp_use_custom_theme }}"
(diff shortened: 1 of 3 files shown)


mattdrayer:
@edx/devops -- need a hand with this one

fredsmith:
There should be a changelog entry for this, as it's a significant change.  

As a note, we're using this code block to check out the edx-microsite repo.  Is that completely replaced by comprehensive theming?

saleem-latif:
@fredsmith I was under the impression that this code is used only by Stanford Theming (related to `USE_CUSTOM_THEME` setting) that is why I am removing this. But if it is being used by `edx-microsites` than I suppose we would not have to remove this.
Can you take a look at https://github.com/edx/edx-platform/pull/12995 and let us know if they require configuration changes as done in this PR if not than I think we can close this PR?
@mattdrayer -- FYI

fredsmith:
@saleem-latif I don't know how assets are built from microsites, but that's a concern i have with your other PR.  We certainly shouldn't merge this PR until we have a new way of deploying the microsites repo.

I would suggest we close this PR.

saleem-latif:
It seems we do need these changes, so closing this PR for now. 
@fredsmith , @mattdrayer -- FYI

PR #3251, opened 2016-07-15
Added PROGRAMS_API_URL for the catalog/discovery service

clintonb:
ECOM-4786


clintonb:
@MichaelRoytman @e0d can you explain why the discovery container was not used during the build: https://travis-ci.org/edx/configuration/builds/145100739?

michaelroytman:
@clintonb @e0d It was related to #3143 (see line 172 of parsefiles.py). Apparently, testing for membership of a string in a set of strings returns false even when the string is in the set (I'm not sure why?). I changed it to a list, which returns true. When I run the fixed script locally, I get the following list of container to-be-built: forum rabbitmq analytics_api nginx xqueue edxapp ecommerce discovery. I'll make a PR to fix this. 

Fix here: #3252

PR #3252, opened 2016-07-15
Fix bug where testing for membership of a file path in a set of file …

michaelroytman:
…paths repeatedly returned false; changed to list


maxrothman:
👍

PR #3254, opened 2016-07-19
[EV-63] Add the ability to log timing information to local logs for c…

cpennington:
…ollection


cpennington:
@e0d: This is adding the start of phone-home capabilities so that we can start measuring how long ansible takes inside of devstack.

cpennington:
@estute: Can you review this to validate that it matches up with https://github.com/edx/vagrant-timer/pull/4 and https://github.com/edx/edx-platform/pull/13021

arbabnazar:
@cpennington can we modify [this plugin](https://github.com/edx/configuration/blob/master/playbooks/callback_plugins/datadog_tasks_timing.py) to get our work done. Thanks

cpennington:
@arbabnazar: Yeah, that was the plugin I based mine on. I'm not sure if we're better served by adding more logging to a single plugin, or having multiple plugins with distinct logging strategies, though (I was leaning towards the latter, which is why I created a clone)

playbooks/callback_plugins/local_task_timing.py
```diff
@@ -0,0 +1,121 @@
...
+        """
+        # Record the timing of the very last task, we use it here, because we
+        # don't have stop task function by default
+        if self.current_task is not None:
+            self.stats[self.current_task].stop()
+
+        self.playbook_end = datetime.utcnow()
+
+        # Sort the tasks by their running time
+        results = sorted(
```
jzoldak: For our purposes, does it make sense to sort the tasks by duration? e.g. the paver timings are written out sequentially.

cpennington: I don't think it matters, really, since each log line is independent. I'm actually in the middle of refactoring the timing classes to make less duplication, at which point this line may go away.

playbooks/callback_plugins/local_task_timing.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+class CallbackModule(object):
+
+    """
+    Ansible plugin get the time of each task and total time
+    to run the complete playbook
+    """
+    def __init__(self):
+        self.stats = {}
```
jzoldak: Since we only output something if the ANSIBLE_TIMER_LOG envar is set, could we avoid the overhead of this plugin calculating timings etc. if not needed?

playbooks/callback_plugins/local_task_timing.py
```diff
@@ -0,0 +1,121 @@
...
+                    json.dump(
+                        log_message,
+                        outfile,
+                        separators=(',', ':'),
+                        sort_keys=True,
+                    )
+                    outfile.write('\n')
+
+                log_message = {
+                    'playbook': self.playbook_name,
```
jzoldak: The overall timing will look like the other timing outputs, except that it will have no signifier that it's the overall timing. Maybe we could repurpose the 'task' key with a value of 'total' or 'overall'? or maybe you can think of something better...

playbooks/callback_plugins/local_task_timing.py
jzoldak: Ah, like the hipchat callback. https://github.com/edx/configuration/blob/5ef742db15ddab60681fe3bca37d9c0eafd12ffe/playbooks/callback_plugins/hipchat_plugin.py#L32

cpennington: Honestly, the overhead should be minimal, as it's just twiddling a few timestamps. I'm not sure it's worth the complexity, as far as making every callback method check whether it should do something or not.

cpennington:
@jzoldak: I've updated this and it should be ready for final review.

@arbabnazar: I refactored the code to merge it with the Datadog timer.

cpennington:
@edx/devops: Review?

e0d:
@cpennington will try to review today, have you verified this is compatible with Ansible 2 as well?  If it can't be, I'd at least like to understand the scope of changes that we'll need to make.

cpennington:
@e0d: Ok, I've added Ansible 2.0 support (per the instructions here: https://docs.ansible.com/ansible/porting_guide_2.0.html#callback-plugins). Is there a good way for me to test this w/ Ansible 2.0 and our current set of plays?

playbooks/callback_plugins/task_timing.py
```diff
@@ -0,0 +1,269 @@
...
+    """
+    def log_play(self, playbook_name, playbook_timestamp, results):
+        # N.B. This is intended to provide a consistent interface and message
+        # format across all of Open edX tooling, so it deliberately eschews
+        # standard python logging infrastructure.
+        if ANSIBLE_TIMER_LOG is None:
+            return
+
+        messages = []
+        for name, timestamp in results:
```
jzoldak: When using this to create a new jenkins worker, I'm getting this error: ValueError: too many values to unpack
See 00:32:27.148 from https://build.testeng.edx.org/view/test-jobs/job/jz-build-packer-ami-test/15/consoleFull
Note the line number on that error is off because I added some debugging statements to figure out what was going on.

cpennington: Oops. Yeah, I dropped a `.items` during the refactoring process.

cpennington: I just pushed up a fix.

arbabnazar:
@cpennington I can test this for Ansible 2.x, is it ready for test?

cpennington:
@arbabnazar: Yeah, it should be ready now (I just fixed something that would have made it fail in 1.9.x)

jzoldak:
👍

playbooks/callback_plugins/task_timing.py
```diff
@@ -0,0 +1,274 @@
...
+logging.getLogger("dd").setLevel(logging.WARNING)
+LOGGER = logging.getLogger(__name__)
+
+"""
+Originally written by 'Jharrod LaFon'
+#https://github.com/jlafon/ansible-profile/blob/master/callback_plugins/profile_tasks.py
+
+"""
+
+ANSIBLE_TIMER_LOG = os.environ.get('ANSIBLE_TIMER_LOG')
```
e0d, line 28: Why not log to STDOUT?

playbooks/callback_plugins/task_timing.py
```diff
@@ -0,0 +1,274 @@
...
+            return
+
+        datadog_tasks_metrics = []
+        for name, timestamp in results.items():
+            datadog_tasks_metrics.append({
+                'metric': 'edx.ansible.task_duration',
+                'date_happened': time.mktime(timestamp.start.timetuple()),
+                'points': timestamp.duration.total_seconds(),
+                'tags': [
+                    'task:{0}'.format(self.clean_tag_value(name)),
```
e0d, line 114: There's an assumption built in here that it may make sense to address.  Ansible automagically pre-pends `"{{ role_name }} | "` to output.  Do we want to parse the role out and treat it as first-class?

e0d:
@cpennington a couple of questions, but generally LGTM.

playbooks/callback_plugins/task_timing.py
cpennington, line 28: I want to eventually ship these to a central collection facility (along with paver and vagrant logs formatted the same way), so I wanted to keep them separate from stdout (rather than having to extract them after the fact).

playbooks/callback_plugins/task_timing.py
cpennington, line 114: Ansible prepends that for datadog logging? `name` in this case is coming from https://github.com/edx/configuration/pull/3254/files#diff-89faaf5bbfb881708fb31fadb34e6fe5R244, which is passed in by ansible.

e0d, line 114: Ansible does it generally for logging, but the general question I guess is is there any loss of semantics having the role concatenated with the task everywhere.  Frankly, not sure.

e0d:
:+1:

PR #3255, opened 2016-07-19
OPS-1536: Add python3 support to xqwatcher

maxrothman:
@edx/devops review

### DIFF
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -87,13 +87,27 @@
  # and add a custom motd.  These do not require an
  # ssh restart
  # Only needed for EC2 instances.
-- name: Update the ssh motd on Ubuntu
-  file:
+- name: Check if files exist so the next task doesn't fail
+  stat:
     path: "{{ item }}"
-    mode: "0644"
-  when: (vagrant_home_dir.stat.exists == false) and (ansible_distribution in common_debian_variants)
+  register: motd_files_exist
   with_items:
     - "/etc/update-motd.d/10-help-text"
     - "/usr/share/landscape/landscape-sysinfo.wrapper"
     - "/etc/update-motd.d/51-cloudguest"
     - "/etc/update-motd.d/91-release-upgrade"
+
+- name: Update the ssh motd on Ubuntu
+  file:
+    path: "{{ item.0 }}"
+    mode: "0644"
+  when: >
+    vagrant_home_dir.stat.exists == false and
+    ansible_distribution in common_debian_variants and
+    item.1.stat.exists
+  with_together:
+    - - "/etc/update-motd.d/10-help-text"
+      - "/usr/share/landscape/landscape-sysinfo.wrapper"
+      - "/etc/update-motd.d/51-cloudguest"
+      - "/etc/update-motd.d/91-release-upgrade"
+    - motd_files_exist.results
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -99,15 +99,10 @@
 
 - name: Update the ssh motd on Ubuntu
   file:
-    path: "{{ item.0 }}"
+    path: "{{ item.item }}"
     mode: "0644"
   when: >
     vagrant_home_dir.stat.exists == false and
     ansible_distribution in common_debian_variants and
-    item.1.stat.exists
-  with_together:
-    - - "/etc/update-motd.d/10-help-text"
-      - "/usr/share/landscape/landscape-sysinfo.wrapper"
-      - "/etc/update-motd.d/51-cloudguest"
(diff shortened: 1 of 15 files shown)


maxrothman:
@edx/devops looking for directional review and input on testing plan on 4dbc15c

maxrothman:
@edx/devops this is ready for review
@jcdyer I can now build AMIs. How should we go about testing?

playbooks/roles/aws/tasks/main.yml
```diff
@@ -87,13 +87,27 @@
...
+- name: Update the ssh motd on Ubuntu
+  file:
+    path: "{{ item.0 }}"
+    mode: "0644"
+  when: >
+    vagrant_home_dir.stat.exists == false and
+    ansible_distribution in common_debian_variants and
+    item.1.stat.exists
+  with_together:
+    - - "/etc/update-motd.d/10-help-text"
```
feanil: Instead of repeating this twice, let's move this list to a var in `defaults/main.yml`

playbooks/roles/common/tasks/main.yml
```diff
@@ -34,24 +34,32 @@
     path: /home/vagrant
   register: vagrant_home_dir
 
+# Ensure that we get a current version of Git
```
feanil: Why was this moved above adding the apt repository?

feanil: NM, misread this.

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -121,6 +121,19 @@
     dest: "/etc/init/{{ supervisor_service }}.conf"
     owner: root
     group: root
+  when: ansible_distribution_major_version < 16
+  tags:
+    - install
+    - install:base
+
+# NB: with systemd, pre_supervisor is a pre-task for supervisor, not a separate service
+- name: Create supervisor systemd job
```
feanil: NIT: Can you move this task above or below the two installs of 'init.d' scripts.  That way visually all the old tasks and the new tasks are separated and not intermingled as much.

jcdyer:
@maxrothman Let's set up a box with the new branch of XQueueWatcher and a python3 queue and point an edx-platform sandbox at it.  I'll create a course on the sandbox that has a python3 coderesponse problem.

maxrothman:
@jcdyer I think the easiest thing would be to grab stage for an hour or two.

maxrothman:
@feanil addressed comments

maxrothman:
To help out with review, here's what a new templated systemd file looks like:

```
[Unit]
Description=supervisord - Supervisor process control system
Documentation=http://supervisord.org
After=network.target


[Service]
# Run pre_supervisor
ExecStartPre=/edx/app/supervisor/venvs/supervisor/bin/python /edx/app/supervisor/pre_supervisor_checks.py \
  --hipchat-api-key <key> --hipchat-room release \
  --available=/edx/app/supervisor/conf.available.d --enabled=/edx/app/supervisor/conf.d

Type=forking
User=www-data
TimeoutStartSec=432000

ExecStart=/edx/app/supervisor/venvs/supervisor/bin/supervisord -n --configuration /edx/app/supervisor/supervisord.conf
ExecReload=/edx/app/supervisor/venvs/supervisor/bin/supervisorctl reload
ExecStop=/edx/app/supervisor/venvs/supervisor/bin/supervisorctl shutdown


[Install]
WantedBy=multi-user.target
```

maxrothman:
@edx/devops this is finalized. Rereview?

playbooks/roles/common/tasks/main.yml
```diff
@@ -34,24 +34,32 @@
...
+# GitHub requires version 1.7.10 or later
+# https://help.github.com/articles/https-cloning-errors
```
fredsmith, line 39: I assume this is no longer necessary for newer versions of ubuntu.  Should we have some logic to exclude this PPA on the latest releases?

maxrothman, line 39: I think in the future we'll still want to keep git current and Ubuntu's default package won't do that. Either way, I think that's out-of-scope for this change (I just moved the task around).

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -121,6 +121,7 @@
     dest: "/etc/init/{{ supervisor_service }}.conf"
     owner: root
     group: root
+  when: ansible_distribution_major_version < 16
```
fredsmith: we should check for ansible_distribution = "Ubuntu"

fredsmith:
I have concerns about how you're limiting tasks to Ubuntu >= 16 and how the git ppa is being installed even where not necessary but other than those, 👍

maxrothman:
@fredsmith fixed. Rereview?

fredsmith:
👍

PR #3256, opened 2016-07-19
Add LMS_ROOT_URL to django settings.

jcdyer:
## Description

LMS_BASE hides information about the URL protocol that then gets reconstructed with unfortunate assumptions throughout the CMS.  We should instead surface an LMS_ROOT_URL setting that explicitly communicates the URL scheme.

For instance:

```
 LMS_BASE = 'courses.edx.org'
```

should be accompanied by an LMS_ROOT_URL like:

```
 LMS_BASE = 'courses.edx.org'
 LMS_ROOT_URL = 'https://courses.edx.org'
```

Adding the setting to ENV_TOKENS here will allow us to surface that information in the CMS and LMS settings.
## Reviewers
- [x] @feanil 
- [x] @nasthagiri

FYI: @wajeeha-khalid 
## Related to

[MA-2536: Server - Retrieve the URL for the course about page](https://openedx.atlassian.net/browse/MA-2536)

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -819,6 +819,7 @@ generic_env_config:  &edxapp_generic_env
   STATIC_ROOT_BASE: "{{ edxapp_staticfile_dir }}"
   LMS_BASE: "{{ EDXAPP_LMS_BASE }}"
   CMS_BASE: "{{ EDXAPP_CMS_BASE }}"
+  LMS_ROOT_URL: "{{ EDXAPP_LMS_ROOT_URL }}"
   BOOK_URL:  "{{ EDXAPP_BOOK_URL }}"
   PARTNER_SUPPORT_EMAIL: "{{ EDXAPP_PARTNER_SUPPORT_EMAIL }}"
   PLATFORM_NAME: "{{ EDXAPP_PLATFORM_NAME }}"


nasthagiri:
A few notes/questions:
1. Is it intentional that this is being added only for LMS and not for CMS?
2. Did we find out where the http scheme is set?  If anyone changes the http scheme they'll need to remember to update this setting as well. 
3. If this setting becomes out of sync with the http scheme setting, what is the error/behavior that the operator would expect?

jcdyer:
I updated your question to use a numbered list to make point by point responses clearer:
1. It is intentional, to get this in quickly for MA-2536.  There will be a follow-up ticket to spread the use of this throughout the code base, and remove all assumptions about URL schemes.  That ticket will also include a comparable setting for CMS_ROOT_URL.  It's not included here, because the infrastructure isn't in place in ansible to surface the setting easily.
2. The scheme is set in `EDXAPP_LMS_BASE_SCHEME`.  [`EDXAPP_LMS_ROOT_URL` is built from that setting](https://github.com/edx/configuration/blob/3994f7d83927de3d31282affe475a40662a7ca53/playbooks/roles/edxapp/defaults/main.yml#L289), so they should not get out of sync.  Note `EDXAPP_CMS_BASE_SCHEME` does not yet exist, which is part of the reason for this PR not surfacing `CMS_ROOT_URL`, though I think we can agree that such a setting should exist.
3.  If somehow the settings do get out of sync (`LMS_ROOT_URL` is manually overridden, for instance), the site will trust the setting, and provide the specified version as the root URL to the LMS.  The user will see broken links in the CMS.  The CMS has no way of knowing what URL is correct for the LMS short of making a web request to that URL to see if it gets a sane response, which I don't think the CMS should be doing.

nasthagiri:
1. Sounds good.  Please post the ticket on this PR once it's created.
2. Ah.  So it's a pre-existing setting and already based on the scheme.  Awesome.
3. Ok.

nasthagiri:
👍   Assuming devops approval.

jcdyer:
@nasthagiri https://openedx.atlassian.net/browse/TNL-5035

jcdyer:
@feanil Ping!

feanil:
:+1:

PR #3263, opened 2016-07-20
Add node dependencies to discovery role

bderusha:
@clintonb please review

### DIFF
--- a/playbooks/roles/discovery/tasks/main.yml
+++ b/playbooks/roles/discovery/tasks/main.yml
@@ -10,13 +10,13 @@
 #
 #
 # Tasks for role discovery
-# 
+#
 # Overview:
-# 
+#
 #
 # Dependencies:
 #
-# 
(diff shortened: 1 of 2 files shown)


playbooks/roles/discovery/tasks/main.yml
```diff
@@ -76,6 +76,35 @@
     - devstack
     - devstack:install
 
+- name: Create nodeenv
```
clintonb: There is another issue afoot. `nodeenv` needs to be installed by the call to `make production-requirements`; however, the call to `make requirements` will attempt to use npm. You will need to modify the Makefile in course-discovery.

See #3261 and https://github.com/edx/course-discovery/pull/172.

bderusha:
FYI @feanil

bderusha:
@clintonb

clintonb:
👍

PR #3265, opened 2016-07-21
Add install_stack.sh script for simplifying fullstack/devstack installation

zrrobbins:
@nedbat

### DIFF
--- a/playbooks/roles/sitespeedio/tasks/main.yml
+++ b/playbooks/roles/sitespeedio/tasks/main.yml
@@ -25,13 +25,22 @@
 #
 
 # adding chris-lea nodejs repo
-- name: add ppas for current versions of nodejs
-  apt_repository: "repo={{ sitespeedio_chrislea_ppa }}"
+- name: Add ppas for current versions of nodejs
+  apt_repository:
+    repo: "{{ sitespeedio_chrislea_ppa }}"
+    state: present
 
 # Install Sitespeed Dependencies
-- name: install sitespeed.io specific system packages
-  apt: pkg={{','.join(sitespeedio_debian_pkgs)}} state=present
+- name: Anstall sitespeed.io specific system packages
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ sitespeedio_debian_pkgs }}"
 
 # Install released version of sitespeedio
-- name: install sitespeed.io
-  npm: name=sitespeed.io global=yes state=latest version=3.10.0
+- name: Install sitespeed.io
+  npm:
+    name: "sitespeed.io"
+    global: yes
+    state: latest
+    version: "3.10.0"
--- a/playbooks/roles/sitespeedio/tasks/main.yml
+++ b/playbooks/roles/sitespeedio/tasks/main.yml
@@ -31,7 +31,7 @@
     state: present
 
 # Install Sitespeed Dependencies
-- name: Anstall sitespeed.io specific system packages
+- name: Install sitespeed.io specific system packages
   apt:
     name: "{{ item }}"
     state: present
--- a/util/install/install_stack.sh
+++ b/util/install/install_stack.sh
@@ -7,9 +7,9 @@ function usage
 {
     cat << EOM
 
-    --- install_devstack.sh ---
+    --- install_stack.sh ---
 
-    Usage: $ bash install_devstack.sh stack release [-p] [-b vagrant_mount_base] [-l] [-v] [-h]
+    Usage: $ bash install_stack.sh stack release [-p] [-b vagrant_mount_base] [-l] [-v] [-h]
 
     Installs the Open edX devstack or fullstack. Reach out on the Open edX community Slack 
     on #ops (https://open.edx.org/blog/open-edx-slack) or the Open edX Ops Google Group
--- a/util/install/install_stack.sh
+++ b/util/install/install_stack.sh
@@ -9,11 +9,11 @@ function usage
 
     --- install_stack.sh ---
 
-    Usage: $ bash install_stack.sh stack release [-p] [-b vagrant_mount_base] [-l] [-v] [-h]
+    Usage: $ bash install_stack.sh stack release [-b vagrant_mount_base] [-v] [-h]
 
-    Installs the Open edX devstack or fullstack. Reach out on the Open edX community Slack 
-    on #ops (https://open.edx.org/blog/open-edx-slack) or the Open edX Ops Google Group
-    (https://groups.google.com/forum/#!forum/openedx-ops) to get support questions answered.
+    Installs the Open edX devstack or fullstack. If you encounter any trouble or have 
+    questions regarding installation of devstack/fullstack, head over to 
+    https://open.edx.org/getting-help.
 
     This script captures a log of all output produced during runtime, and saves it in a .log
     file within the current directory. If you encounter an error during installation, this is
@@ -28,26 +28,21 @@ function usage
         Either 'fullstack' or 'devstack' (no quotes). Full stack mimics a production 
         environment, whereas devstack is useful if you plan on modifying the Open edX 
         code. You must specify this. If you choose fullstack, 'release' should be the
-        latest named-release. If you choose devstack, 'release' should be the latest
-        named-release or master.
+        latest open-release. If you choose devstack, 'release' should be the latest
+        open-release or master.
 
     release
         The release of Open edX you wish to run. Install the given git ref 'release'.
-        You must specify this. Named releases are called "named-release/dogwood",
-        "named-release/dogwood.2", and so on. We recommend the latest stable named 
+        You must specify this. Named releases are called "open-release/eucalyptus",
+        "open-release/eucalyptus.2", and so on. We recommend the latest stable open 
         release for general members of the open source community. Named releases can
         be found at: https://openedx.atlassian.net/wiki/display/DOC/Open+edX+Releases.
-        If you plan on modifying the code, we recommend the "master" branch.
-
-    -p
-        Enable use of "preview" from within Studio. 
+        If you plan on modifying the code, we recommend the "master" branch. 
 
     -b vagrant_mount_base
         Customize the location of the source code that gets cloned during the 
-        devstack provisioning.
-
-    -l
-        Disable logging. Enabled by default.
+        devstack provisioning. The default is the current directory. This option is
+        not valid if installing fullstack.
 
     -v 
         Verbose output from ansible playbooks.
@@ -60,14 +55,16 @@ function usage
 EOM
 }
 
-# Logging
-logging=1
+
+ERROR='\033[0;31m' # Red
+WARN='\033[1;33m' # Yellow
+SUCCESS='\033[0;32m' # Green
+NC='\033[0m' # No Color
+
 # Output verbosity
 verbosity=0
 # OPENEDX_RELEASE
 release=""
-# Enable preview in Studio
-enable_preview=0
 # Vagrant source code provision location
 vagrant_mount_location=""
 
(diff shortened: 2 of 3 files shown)


playbooks/roles/sitespeedio/tasks/main.yml
```diff
@@ -25,22 +25,13 @@
 #
 
 # adding chris-lea nodejs repo
-- name: Add ppas for current versions of nodejs
-  apt_repository:
-    repo: "{{ sitespeedio_chrislea_ppa }}"
-    state: present
+- name: add ppas for current versions of nodejs
+  apt_repository: "repo={{ sitespeedio_chrislea_ppa }}"
```
nedbat: How did these commits get here? This doesn't look like something you would have done.

jibsheet: These look like reverts of Arbab's work to clean up tasks across the codebase.
I suspect an incorrect rebase / old branch point.

(In this case, specifically 97cbfda52f0b3d4e4bc2725e1108a1ca2cea77dc)

zrrobbins: Yes, I did not touch any files in the existing code, I imagine I must've incorrectly pulled in changes for main.yml.

stvstnfrd:
@zrrobbins What's the motivation/objective here?

This looks like a lot of complication (read: maintenance burden) for
what is, effectively, copy/pasting the wiki instructions into a shell
script.

Scripts like this almost always fall out-of-sync with their source
documents; who's going to be in charge of ensuring this script always
gets updated in tandem with the wiki doc? Even if you "assign" someone
and even if you add a note in the wiki doc ("hey, update
`install_stack.sh`!"), eventually it will be overlooked. The only thing worse
than no documentation is conflicting documentation.

For better or for worse, the Devstack requires at least a minimal level
of techincal expertise in order to operate, so I'm not sure what value
is added by abstracting away this logic. You need to know how `vagrant`
works, you need to `vagrant ssh` immediately afterward and you need to
know about `vagrant halt && vagrant up` in order to start using this.
Before anyone could use this script, they would need to be made
explicitly aware that this is creating a vagrant instance for you and
they would need to know how to interact with said instance.

Further, users would _still_ need to be instructed where to grab this
file and how to execute it, so this script would need its own
documentation somewhere [1]. Now _that_ script needs uploaded and
maintained, as this script changes.. and now we're back to the same
synchronisation problem again..

At that point, if they already need to know what this script does in
order to use it, why aren't we just pointing them to the canonical
source of devstack documentation, the wiki [2]?

What's really different between [1] and [2]?

[1] 

``` sh
mkdir -p /tmp/scripts
cd /tmp/scripts
curl -L https://raw.githubusercontent.com/edx/configuration/master/util/install/install_stack.sh > install_stack.sh
bash ./install_stack.sh devstack master
```

[2]

``` sh
mkdir -p devstack
cd devstack
curl -L https://raw.githubusercontent.com/edx/configuration/${OPENEDX_RELEASE}/vagrant/release/devstack/Vagrantfile > Vagrantfile
vagrant plugin install vagrant-vbguest
vagrant up
```

zrrobbins:
The motivation behind this script was primarily a way to guarantee logging of a consistent install process. So if everyone uses this script, a log will always be produced that follows a known order of commands. Then, the user having trouble simply sends us the log file rather than starting a conversation that can often take time to narrow down to the root cause. I do agree that the current location of the script is by no means the best location for it to reside. I think @jbarciauskas and @nedbat may have helpful input on this as well.

nedbat:
@stvstnfrd I hear what you are saying about how the instructions for the install script are nearly as complex as the install instructions themselves.  As @zrrobbins mentions, the original motivation for this script was to ensure a few things:
1. That required choices were made, and were not silently skipped. (BTW: ironically, your demonstration [2] above forgot to set the OPENEDX_RELEASE variable.)
2. That the installation choices were recorded some place for later forensics.
3. That the installation was logged for later forensics.

What I've found while trying to support people who've had difficulties with installation is that these are common meta-problems: they are problems that create more problems, or prevent the debugging of the installation.

I appreciate the feedback.  More eyes on this process are better.  Maybe you have other ideas about how to make the process more likely to succeed.

stvstnfrd:
Thanks @zrrobbins @nedbat 

@nedbat : In that case, what's the discovery mode?
How will people know to find it?

Should the wiki instructions instead be instructions on using this script instead of the Vagrantfile directly?

stvstnfrd:
> BTW: ironically, your demonstration [2] above forgot to set the OPENEDX_RELEASE variable.

In my defense, I've never used that variable in my life! We maintain our own forked devstacks ;)

But seriously, thanks for the follow-up and tolerating me jumping in out of the blue.

nedbat:
We will definitely update the installation instructions, yes.  As we grow the open source team, we will also be looking to garden the wiki/docs back a bit so that people can more easily find the best instructions, also to help with the support burden.

Your feedback is welcome, not tolerated! :)

feanil:
I do like the idea of only having one set of instructions, the script seems like an improvement, can we have the wiki point to it and add in line commenting to cover any non-scripted info that is in the wiki?

nedbat:
@feanil Not sure what you mean by "in line commenting to cover any non-scripted info"?  Do you have an example?

feanil:
This page(https://openedx.atlassian.net/wiki/display/OpenOPS/Running+Devstack) has a bunch the same instructions as can be found in this script.  But there is a lot more human readable explaination in the wiki page than there is in the shell script.  I think we should make the script the canonical location for the info but that means adding any comments that are on the wiki that belong with the code being run for people who are trying to understand what the code is doing.

nedbat:
As much as I don't want to duplicate information, having yet another place for instructions (inside a .sh file), seems like a step backward.  And expecting people to read through a .sh file for the comments, to do things themselves, seems especially esoteric.

zrrobbins:
@feanil @nedbat The documentation within the script was mean to be limited, covering only usage and common use-cases. I imagined that this script would be available to download as an alternative install method on the page you linked Feanil, and comprehensive documentation would be present for it on that page.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
+if [[ $logging > 0 ]]; then
+    exec > >(tee install-$(date +%Y%m%d-%H%M%S).log) 2>&1
+    echo "Logging enabled."
```
nedbat: Please add the name of the log file here.  "Logging to blah-blah.log"

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+        release for general members of the open source community. Named releases can
+        be found at: https://openedx.atlassian.net/wiki/display/DOC/Open+edX+Releases.
+        If you plan on modifying the code, we recommend the "master" branch.
+
+    -p
+        Enable use of "preview" from within Studio. 
+
+    -b vagrant_mount_base
+        Customize the location of the source code that gets cloned during the 
+        devstack provisioning.
```
nedbat: Mention the default here (current directory).

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+            exit
+            ;;
+        *)
+            usage
+            exit 1
+            ;;
+    esac
+done
+
+if [[ $logging > 0 ]]; then
```
nedbat: I'm not sure why we're even allowing the option to disable logging.  Let's make it fool-proof.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+
+# Check for verbosity level
+if [[ $verbosity == 1 ]]; then
+    sed -i '' 's/-e xqueue_version=\$OPENEDX_RELEASE/-e xqueue_version=\$OPENEDX_RELEASE \\\'$'\n    -vvv/' Vagrantfile
+fi
+
+vagrant up --provider virtualbox
+
+# Check if preview mode was chosen
+if [[ $enable_preview != 1 ]] || grep -q '192.168.33.10  preview.localhost' /etc/hosts; then
```
nedbat: Why shouldn't we always enable preview?

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+    curl -L https://raw.githubusercontent.com/edx/configuration/${OPENEDX_RELEASE}/vagrant/release/fullstack/Vagrantfile > Vagrantfile
+    vagrant plugin install vagrant-hostsupdater
+else # Throw error
+    echo -e "${ERROR}Unrecognized stack name, must be either devstack or fullstack!${NC}"
+    exit 1
+fi
+
+# Check for verbosity level
+if [[ $verbosity == 1 ]]; then
+    sed -i '' 's/-e xqueue_version=\$OPENEDX_RELEASE/-e xqueue_version=\$OPENEDX_RELEASE \\\'$'\n    -vvv/' Vagrantfile
```
nedbat, line 150: This seems kind of fragile...

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+    fi
+
+    wiki_link="https://openedx.atlassian.net/wiki/display/OpenOPS/Running+Devstack"
+    mkdir -p devstack
+    cd devstack
+    curl -L https://raw.githubusercontent.com/edx/configuration/${OPENEDX_RELEASE}/vagrant/release/devstack/Vagrantfile > Vagrantfile
+    vagrant plugin install vagrant-vbguest
+elif [[ $stack == "fullstack" ]]; then # Install fullstack
+    # Warn if release chosen is not named-release
+    if [[ $release != *"named-release"* ]]; then
```
nedbat: For Eucalyptus, we're switching naming schemes to open-release/eucalyptus.something .  Not sure if we should only accept that style, or also still accept "named-release" also.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
+#!/usr/bin/env bash
```
nedbat, line 1: If I run this a second time in the same directory, (to install a different version, for example), it says, "Machine already provisioned", and doesn't do anything.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+release=$1
+shift
+
+while getopts "pb:lvh" opt; do
+    case "$opt" in
+        p)
+            enable_preview=1
+            ;;
+        b)
+            vagrant_mount_location=$OPTARG
```
nedbat: This option is only valid for devstack. Need some way to error out if "fullstack -b blah" is specified.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+else
+    echo "Logging disabled."
+fi
+
+ERROR='\033[0;31m' # Red
+WARN='\033[1;33m' # Yellow
+SUCCESS='\033[0;32m' # Green
+NC='\033[0m' # No Color
+
+export OPENEDX_RELEASE=$release
```
nedbat, line 108: You need to remove "export" here, or the variables aren't defined properly.  No, I don't understand why.

nedbat, line 108: Hmm, or maybe I am wrong.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+
+# Check if preview mode was chosen
+if [[ $enable_preview != 1 ]] || grep -q '192.168.33.10  preview.localhost' /etc/hosts; then
+    echo "Studio preview already enabled, skipping..."
+else
+    echo "Enabling use of preview within Studio..."
+    sudo bash -c "echo '192.168.33.10  preview.localhost' >> /etc/hosts"
+fi
+
+echo -e "${SUCCESS}Finished installing! You may now login using 'vagrant ssh'"
```
nedbat: This is a little confusing, because you have to "cd devstack" before "vagrant ssh" will work.

util/install/install_stack.sh
```diff
@@ -0,0 +1,172 @@
...
+{
+    cat << EOM
+
+    --- install_stack.sh ---
+
+    Usage: $ bash install_stack.sh stack release [-p] [-b vagrant_mount_base] [-l] [-v] [-h]
+
+    Installs the Open edX devstack or fullstack. Reach out on the Open edX community Slack 
+    on #ops (https://open.edx.org/blog/open-edx-slack) or the Open edX Ops Google Group
+    (https://groups.google.com/forum/#!forum/openedx-ops) to get support questions answered.
```
nedbat: We should instead point them to https://open.edx.org/getting-help

util/install/install_stack.sh
zrrobbins: In that case, we should add a disclaimer to the script or the wiki letting people know that this script will only support installations of Eucalyptus and up.

nedbat:
:+1:, bring this train into merge-town!

nedbat:
Oops, my bad, I forgot to insist on squashed commits.

PR #3266, opened 2016-07-22
ECOM-4904 fix the broken configuration book where nodeenv was not deleted cleanly

schenedx:
@rlucioni @nedbat Please review this bug fix.


rlucioni:
👍

PR #3268, opened 2016-07-22
Update fullstack Vagrantfile for Euc 1rc1

nedbat:


nedbat:
@feanil These are the boxes we just uploaded.

PR #3270, opened 2016-07-27
Remove ORA1 references

stvstnfrd:
I noticed these were still floating around..


openedx-webhooks:
Thanks for the pull request, @stvstnfrd! I've created [OSPR-1358](https://openedx.atlassian.net/browse/OSPR-1358) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

shaunagm:
@adampalay, any thoughts about this?

adampalay:
@shaunagm , I think it makes more sense for devops to look at this, but since we don't have ORA1 anymore, I don't see why this wouldn't be an issue! (thanks @stvstnfrd ) @jibsheet ?

jibsheet:
I'm luckily blissfully unaware of ORA1, but I threw this in as one of our OSPRs for the sprint planning happening tomorrow.

feanil:
@stvstnfrd this looks great,  it looks like there is still a dangling reference [here](https://github.com/edx/configuration/blob/edc67b6875ad0f2e6e34b845442d208e2b6972a6/playbooks/roles/nginx/defaults/main.yml#L94-L95) in the nginx defaults. Mind grabbing that as well?  Then I think we can merge this.

feanil:
I noticed that that change was only on your branch and is gone from master.

PR #3275, opened 2016-07-28
Add email_marketing queue to ecommerce_worker celery startup

PaulWattenberger:

PR #3280, opened 2016-08-02
Sailthru support ecommerce_worker default values

PaulWattenberger:
ecommerce_worker default values for Sailthru support.  PR https://github.com/edx/configuration/pull/3275 also required to add queue to celery startup

### DIFF
--- a/playbooks/roles/ecomworker/defaults/main.yml
+++ b/playbooks/roles/ecomworker/defaults/main.yml
@@ -52,6 +52,47 @@ ECOMMERCE_WORKER_WORKER_ACCESS_TOKEN: 'your-secret-here'
 ECOMMERCE_WORKER_MAX_FULFILLMENT_RETRIES: 11
 # END ORDER FULFILLMENT
 
+# SAILTHRU INTEGRATION
+# Set to false to ignore Sailthru events
+#  Sailthru support in ecommerce_worker sends purchase/enroll events to the email
+#  marketing system Sailthru for tracking the efficacy of email marketing campaigns.  It does not
+#  have to be enabled for normal ecommerce operation.  If it is enabled, the waffle switch
+#  sailthru_enable should be set to on in ecommerce as well or ecommerce won't send
+#  purchase/enroll events to ecommerce_worker.
+ECOMMERCE_WORKER_SAILTHRU_ENABLE: false
+
+# Template used when user upgrades to verified
+ECOMMERCE_WORKER_SAILTHRU_UPGRADE_TEMPLATE: !!null
+
+# Template used with user purchases a course
+ECOMMERCE_WORKER_SAILTHRU_PURCHASE_TEMPLATE: !!null
+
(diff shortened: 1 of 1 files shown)


PaulWattenberger:
@clintonb 
I'll add devops once you approve

playbooks/roles/ecomworker/defaults/main.yml
```diff
@@ -52,6 +52,42 @@ ECOMMERCE_WORKER_WORKER_ACCESS_TOKEN: 'your-secret-here'
 ECOMMERCE_WORKER_MAX_FULFILLMENT_RETRIES: 11
 # END ORDER FULFILLMENT
 
+# SAILTHRU INTEGRATION
+# Set to false to ignore Sailthru events
+ECOMMERCE_WORKER_SAILTHRU_ENABLE: false
+
+# Template used when user upgrades to verified
+ECOMMERCE_WORKER_SAILTHRU_UPGRADE_TEMPLATE: !!null
```
clintonb, line 65: Do any of these template settings have different values for stage versus production?

PaulWattenberger, line 65: Stage will point to some test templates.  Prod will probably have templates with different names, once marketing develops them...

clintonb:
👍

PaulWattenberger:
@e0d

e0d:
@PaulWattenberger @clintonb this seems to me like configuration that is likely to not be useful to anyone outside of edX.  Given that, I probably would have approached this is a less vendor specific way.

I assume that ecom can run just fine without enable sailthru, is that true?

clintonb:
@e0d neither ecommerce nor ecommerce-worker needs SailThru to function properly.

PaulWattenberger:
@e0d Sailthru not required and disabled by default with setting in ecommerce_worker and waffle switch in ecommerce

playbooks/roles/ecomworker/defaults/main.yml
```diff
@@ -63,6 +99,19 @@ ECOMMERCE_WORKER_SERVICE_CONFIG:
   JWT_SECRET_KEY: '{{ ECOMMERCE_WORKER_JWT_SECRET_KEY }}'
   JWT_ISSUER: '{{ ECOMMERCE_WORKER_JWT_ISSUER }}'
   MAX_FULFILLMENT_RETRIES: '{{ ECOMMERCE_WORKER_MAX_FULFILLMENT_RETRIES }}'
+  SAILTHRU:
```
e0d, line 107: Can you put some comments here giving context and indicating that it's entirely optional, but mentioning what functionality gaps there will be if a deployer isn't using sailthru?

PaulWattenberger, line 107: Okay, comments added.

clintonb:
Remember to squash your commits.

e0d:
Thanks :+1: after squash.

PR #3281, opened 2016-08-03
Removed broken/unnecessary call to make target

clintonb:
There is no need to call requirments.js since we are already installing the node/bower dependencies.

ECOM-5129

### DIFF
--- a/playbooks/roles/discovery/tasks/main.yml
+++ b/playbooks/roles/discovery/tasks/main.yml
@@ -102,16 +102,6 @@
     - install
     - install:app-requirements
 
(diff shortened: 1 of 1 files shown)


jibsheet:
:+1: I assume this will fix the ci failure from calling this step.

clintonb:
Yes, @jibsheet. More importantly, we restore our ability to _build AMIs_. I will first need to merge https://github.com/edx/course-discovery/pull/200.

jibsheet:
Pfft, AMIs.  @MichaelRoytman this is the PR I mentioned in standup

PR #3283, opened 2016-08-03
symlink pip

fredsmith:


fredsmith:
turns out, /usr/local/bin is part of the path, but bash caching is caching the old version.

PR #3289, opened 2016-08-04
The apt syntax which handles urls in deb: wasn't added until 2.1

jibsheet:
This reverts and then cleans up a bit the jenkins_master role to run on 1.9

@edx/devops

### DIFF
--- a/playbooks/roles/jenkins_master/tasks/main.yml
+++ b/playbooks/roles/jenkins_master/tasks/main.yml
@@ -37,9 +37,13 @@
     owner: "{{ jenkins_user }}"
     group: "{{ jenkins_group }}"
 
-- name: Install Jenkins package from the internet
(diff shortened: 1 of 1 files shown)


playbooks/roles/jenkins_master/tasks/main.yml
```diff
@@ -37,9 +37,12 @@
     owner: "{{ jenkins_user }}"
     group: "{{ jenkins_group }}"
 
-- name: Install Jenkins package from the internet
+- name: download Jenkins package
```
feanil, line 41: NIT: can we add a todo here to clean this up when we get to 2.1?

Something like `TODO: Ansible 2.1 - We can use urls in debs now.`?

feanil:
:+1:

PR #3291, opened 2016-08-04
Jibsheet/tools jenkins cleanups

jibsheet:
Fix a bug which prevented this playbook from running.
Add a basic Dockerfile so you can build an image.
Role still needs some tagging around nginx/jenkins.


feanil:
:+1:

PR #3294, opened 2016-08-05
Upgrade tools jenkins and plugins

jibsheet:
Remove plugins I don't believe are in use, or planned to be in use.
We can add back as needed.

### DIFF
--- a/playbooks/edx-east/jenkins_tools.yml
+++ b/playbooks/edx-east/jenkins_tools.yml
@@ -8,6 +8,10 @@
   serial: "{{ serial_count }}"
   roles:
     - aws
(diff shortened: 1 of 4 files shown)


feanil:
Generally looks good, One question, Is the jobdsl plugin not installed via this mechanism?

jibsheet:
Fascinating question - I wonder what else was set up manually.  I'll investigate.

feanil:
:+1:

jibsheet:
@fredsmith @feanil Updates based on running (and re-running) the job.

fredsmith:
👍

PR #3295, opened 2016-08-09
prefix keys and instances with the same text

macdiesel:
@feanil you're welcome ;)


macdiesel:
@feanil updated per request to add the timestamp.

feanil:
:+1:

PR #3301, opened 2016-08-10
Platform/update forums to ruby23 coordinated

tobz:


robrap:
@tobz: When you return from vacation, do you want to close this?

PR #3302, opened 2016-08-10
Correct dependency analyzer

michaelroytman:


jibsheet:
@MichaelRoytman Can we make a quick additional commit on this branch to test what happens if you add a few lines to the common role and then delete that and add some lines to just the discovery role?

util/parsefiles.py
```diff
@@ -192,17 +205,21 @@ def change_set_to_roles(files, git_dir, roles_dirs, playbooks_dirs, graph):
...
-
-                # adds all the roles that a playbook uses to set of roles of the change set
-                items |= {desc.name for desc in descendants}
-    return items
-
+                # if looking in playbooks directory
+                if play_dir == "playbooks":
+                    play = _get_resource_name(file_path, "playbooks")
+                    # remove ".yml" file ending
+                    play = play[:-4]
```
maxrothman: Use [str.rstrip](https://docs.python.org/2/library/stdtypes.html#str.rstrip)

PR #3307, opened 2016-08-11
Adding media storage to config

e0d:
## Configuration Pull Request

@clintonb @schenedx


playbooks/roles/discovery/defaults/main.yml
```diff
@@ -91,6 +91,15 @@ DISCOVERY_MARKETING_API_URL: 'https://example.org/api/catalog/v2/'
 DISCOVERY_MARKETING_URL_ROOT: 'https://example.org/'
 DISCOVERY_PROGRAMS_API_URL: 'https://replace-me/api/v1/'
 
+DISCOVERY_DATA_DIR: '{{ COMMON_DATA_DIR }}/{{ discovery_service_name }}'
+DISCOVERY_MEDIA_ROOT: '{{ DISCOVERY_DATA_DIR }}/media'
+DISCOVERY_MEDIA_URL: '/media/'
+
+DISCOVERY_MEDIA_STORAGE_BACKEND:
```
clintonb, line 98: This is the pattern replicated from Programs. This will require changes to `production.py` in edx/course-discovery. FYI @rlucioni @schenedx

clintonb, line 98: Alternatively, we can explicitly define all of the fields necessary for S3 storage: https://django-storages.readthedocs.io/en/latest/backends/amazon-S3.html. I prefer this approach to avoid the confusion recently encountered with programs.

clintonb:
👍, regardless of if we switch to explicit settings.

playbooks/roles/discovery/defaults/main.yml
rlucioni, line 98: Seconding @clintonb on this. RCA on this issue is [on the wiki](https://openedx.atlassian.net/wiki/display/ENG/RCA%3A+ECOM-4965+-+Images+not+loading+on+LMS+program+listing+page). Though I can understand the appeal of the media settings hash.

schenedx:
@e0d where does the value of `DEFAULT_FILE_STORAGE` in production come from? I want to confirm with you the value of this key is correct for course-discovery IDA

schenedx:
@e0d Can you please sure in the configuration-secure repo, the values below are added?
STATICFILES_STORAGE = 'storages.backends.s3boto.S3BotoStorage'
DEFAULT_FILE_STORAGE = 'storages.backends.s3boto.S3BotoStorage'

playbooks/roles/discovery/defaults/main.yml
schenedx, line 98: @e0d Do you have a strong preference whether we should leave the changes here as it is, or we can have the tuples flat?

schenedx:
👍 (Please make sure the `DEFAULT_FILE_STORAGE` and `STATICFILES_STORAGE` values are correct for S3)

clintonb:
👍

PR #3308, opened 2016-08-11
use Open edX in the motd, and delete unneeded variables

nedbat:
The Open Stack PR from January got rid of different motd for different installations.  We should use "Open edX" as the broadest branding.  And we no longer need the COMMON_MOTD_TEMPLATE variables.

### DIFF
--- a/playbooks/roles/vhost/templates/etc/motd.tail.j2
+++ b/playbooks/roles/vhost/templates/etc/motd.tail.j2
@@ -1,12 +1,18 @@
 *******************************************************************
-*         _ __  __                                                *
-*   _   _| |\ \/ /  This system is for the use of authorized      *
-* / -_) _` | >  <   users only.  Usage of this system may be      *
-* \___\__,_|/_/\_\  monitored and recorded by system personnel.   *
+*     ___                         _ __  __                        *
+*    / _ \ _ __ ___ _ _    ___ __| |\ \/ / (R)                    *
+*   | |_| | '_ \ -_) ' \  / -_) _` | >  <                         *
+*    \___/| .__/___|_|_|  \___\__,_|/_/\_\                        *
+*         |_|                                                     *
+*                                                                 *
+* This system is for the use of authorized users only.  Usage of  *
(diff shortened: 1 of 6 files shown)


nedbat:
@edx/devops ?

playbooks/roles/vhost/templates/etc/motd.tail.j2
```diff
@@ -1,12 +1,18 @@
...
+*    \___/| .__/___|_|_|  \___\__,_|/_/\_\                        *
+*         |_|                                                     *
+*                                                                 *
+* This system is for the use of authorized users only.  Usage of  *
+* this system may be monitored and recorded by system personnel.  *
 *                                                                 *
 * Anyone using this system expressly consents to such monitoring  *
 * and is advised that if such monitoring reveals possible         *
 * evidence of criminal activity, system personnel may provide the *
 * evidence from such monitoring to law enforcement officials.     *
```
nedbat, line 14: Do we still need this threatening language?

nedbat:
@edx/devops what's the secret to getting thumbs?

playbooks/roles/vhost/templates/etc/motd.tail.j2
feanil, line 14: I think we want that on the edx servers where this header is also used.

feanil:
:+1:  assuming this doesn't break AMI or sandbox builds.

nedbat:
@feanil is there something I should do to check AMI or sandbox builds? Or is this good to merge?

feanil:
@nedbat if you haven't already, please build a sandbox from scratch and make sure it doesn't cause any issues.

nedbat:
@feanil made a sandbox from scratch: http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/8758/console  For good measure, I selected all the optional things to install, and "discovery" seems to have failed because of a pgp problem? Can we say that motd is not the cause? :)

nedbat:
@edx/devops sandbox completed, PTAL.

feanil:
:+1:

PR #3310, opened 2016-08-11
Added continous delivery playbook to discovery service refresh metadata

clintonb:
ECOM-5134


clintonb:
@edx/pipeline-team please review

e0d:
@clintonb does this obviate the Jenkins job?  Seems like a better place for this particular job anyway.

clintonb:
I wish it did @e0d, but we need to update data every hour still.

doctoryes:
@clintonb Although our current policy is to keep timed, non-AWS-specific jobs like this on Jenkins, GoCD is capable of having pipelines which run on a timed basis, like cron:
https://docs.go.cd/16.5.0/configuration/admin_timer.html

macdiesel:
👍

PR #3312, opened 2016-08-12
Use force instead of setting fetch.prune

maxrothman:

We originally started setting fetch.prune because it would cause the git
module to fail if there were refs locally that were not on the remote
(because they got deleted). Based on the Ansible devs' response on
https://github.com/ansible/ansible-modules-core/issues/637,
we can safely replace this task with setting the "force: true" flag
wherever we use the git module.

### DIFF
--- a/playbooks/roles/git_clone/tasks/main.yml
+++ b/playbooks/roles/git_clone/tasks/main.yml
@@ -17,17 +17,6 @@
 # require in our default configuration.
 #
 #
-# Rewrite this task using the ansible git-config module once we'll migrate to Ansible 2.x
-# https://docs.ansible.com/ansible/git_config_module.html#git-config
-#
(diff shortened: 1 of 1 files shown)


maxrothman:
Note: `force` will also steamroll any local changes in the working directory. This could potentially silence issues where someone is writing to the wrong place.

openedx-webhooks:
Thanks for the pull request, @maxrothman! I've created [OSPR-2095](https://openedx.atlassian.net/browse/OSPR-2095) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

nedbat:
@maxrothman seems like the old ones can definitely be closed?

maxrothman:
@nedbat I'm just unsubscribing myself from all of them, so whatever works best for y'all.

PR #3313, opened 2016-08-14
Let's Encrypt automatic SSL certificate installation for nginx

mfitzp:
This implements automatic installation of SSL certificates for nginx using the Let's Encrypt `certbot` system (using `certbot-auto`).

Certificates are automatically generated for the domains listed in `LETS_ENCRYPT_DOMAINS`. Generating the certificates requires validation with the Let's Encrypt server, which is here done using the `http-01` ACME scheme. Basically, this involves putting a file with a known hash-name under `./well-known/acme` on a running server. For this reason, we require a running nginx server.

The resulting certificate/key are symlinked to the nginx certificate folder, and the nginx configuration updated to match. To simplify this step, the nginx ssl-certificate configuration is moved to a separate include for all sites that use it. This has a side benefit of removing duplication from the config files. If the nginx playbook is re-run the names of the letsencrypt certs (which are always`fullchain.pem` / `privkey.pem`) need to be passed to prevent reverting to self-signed.

Finally a cronjob is installed to auto-update the certs in place on the recommended twice-daily schedule (they expire in 90 days, this check is for revocation). 

This has been tested on a devstack VM and works fine, running, validating, downloading and installing the certificate. For testing I created a SSL nginx using the self-signed cert, then requested the new certificate. You can also request the cert from a non-SSL nginx, but you will then need to re-run the nginx install (passing the certificate names) to activate it.
### Creating an SSL-enabled nginx install

```
/edx/bin/ansible-playbook ./run_role.yml -i "localhost," -e '{"role":"nginx", "nginx_sites":"cms", "NGINX_ENABLE_SSL":true, "edxapp_cms_gunicorn_port":8001, "EDXAPP_CMS_NGINX_PORT":8000, "NGINX_REDIRECT_TO_HTTPS":false, "EDXAPP_CMS_SSL_NGINX_PORT":8443, "edxapp_data_dir":"/edx/app/edxapp/data"}' -c local
```
### Working with self signed certificate

screen shot 2016-08-14 at 23 52 09

screen shot 2016-08-14 at 23 52 17
### Running lets-encrypt playbook

Running lets-encrypt to download and install certificate. `certbot-testing.mfitzp.io` is pointing to my home network (and forwarded to the VM).

```
/edx/bin/ansible-playbook ./run_role.yml -i "localhost," -e '{"role":"lets_encrypt","LETS_ENCRYPT_DOMAINS":["certbot-testing.mfitzp.io"]}' -c local
```
### Setup complete

Certificate is downloaded and installed. Subsequent requests use the let's encrypt certificate:
screen shot 2016-08-14 at 23 50 29

screen shot 2016-08-14 at 23 50 49
### Load balancing

On the original [JIRA ticket](https://openedx.atlassian.net/browse/OSPR-1382) there was discussion of how to handle sites behind load balancers. The `certbot/http-01` approach will fail intermittently as there is no guarantee which server will receive the request (and whether it will have the correct `./well-known` secret). The suggested solution is to have a dedicated machine handling the cert validation/etc. check — running certbot and handling `./well-known` requests, then transferring updated certificates from that machine to the others. 

This isn't yet implemented here —  it could be implemented as a "push" to a series of hosts/paths (e.g. via `scp`) on the update cron after confirming a change. A simpler approach would be to schedule downloads from a known (single) host to each machine — e.g. `rysnc` the cert folder. This wouldn't be lets-encrypt specific, and probably better as a separate playbook. Thoughts? 

This PR is covered by the OpenCraft contributor agreement @antoviaque @bradenmacdonald 

---


openedx-webhooks:
Thanks for the pull request, @mfitzp! I've created [OSPR-1394](https://openedx.atlassian.net/browse/OSPR-1394) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

playbooks/roles/lets_encrypt/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+  tags:
+    - install
+    - update
+
+# Create symlinks to the letsencrypt certificates into
+# the main cert folder. When requesting multi-domain
+# certs they are shared: symlink only required to the first
+
+- name: symlink certificates
+  file: >
```
arbabnazar: YAML syntax mixup

playbooks/roles/lets_encrypt/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+#
+#
+# Tasks for role lets_encrupt
```
haikuginger: Typo

playbooks/roles/lets_encrypt/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+    src: ssl-certs.conf.j2
+    dest: /etc/nginx/ssl-certs.conf
+    owner: root
+  tags:
+    - install
+    - install:configuration
+
+# Renewal will auto-update the certs in place at the linked-to locations
+# checks twice daily @ 6hr±60mins
+- name: create monthly auto-renew cron job
```
haikuginger: The name should describe the action being taken.

playbooks/roles/lets_encrypt/tasks/main.yml
```diff
@@ -0,0 +1,108 @@
...
+  cron: 
+        name      : "Renew Let's Encrypt certificates via certbot"
+        hour      : 0,12
+        cron_file : "lets-encrypt-renew"
+        user      : "root"
+        job       : sleep $[RANDOM\%60]m; /opt/certbot/certbot-auto renew
```
haikuginger: We can't necessarily assume we get a `bash` shell with `$RANDOM` available inside `cron`; let's either move the random call to within our script, or use Ansible's random template to set a random `minute` value on this play.

haikuginger:
A couple minor nits; once you've addressed them, I'll go ahead and test.

playbooks/roles/lets_encrypt/tasks/main.yml
e0d: Can you document the pre-reqs and process for running this role here.

gsong:
@mfitzp Thank you for the PR, we'll review once you've addressed the issues from @haikuginger.

@antoviaque @bradenmacdonald Can you confirm that @mfitzp is covered under the OpenCraft contributor agreement? If so, is there an expiration?

mfitzp:
Thanks for the review @haikuginger. I've pushed an update with the requested fixes.

The random cron is now implemented using template vars, which has the added benefit of randomising the two hours when it occurs (12h apart). I've also added a `--post-hook` to the renewal to restart the nginx server should the certificates update.

Q: Since this role needs an nginx server running to work (serve the ACME challenge), should it include a task to start the server?

bradenmacdonald:
@gsong Yes, this is a recruitment test task, so he is covered under our agreement for this particular PR. There's no expiration time, but that only applies to this PR.

haikuginger:
This is looking really good to me @mfitzp! I especially like the use of random to pick the hour the renewal task runs.

One final thing, in answer to your question re: nginx. Ideally, the behavior we'd want would be for the playbook to ensure that nginx is started before attempting the certificate creation, but the preferred post-creation behavior would be to replicate nginx's state prior to when the role started.

mfitzp:
Thanks @haikuginger, appreciate the feedback! I've pushed an update that tests/resets the nginx status to what it was before the role is run.

The [initial plan](https://openedx.atlassian.net/browse/OSPR-1382) was to have this used for auto-deployment. To achieve this would require adding a conditional run of `lets_encrypt` in the `edx_sandbox` (or elsewhere?) e.g.

```
<nginx stuff here>
- role: lets_encrypt
    when: LETS_ENCRYPT_DOMAINS
```

...then if/when that variable were set the cert would automatically be deployed. Is this an acceptable addition to this PR? (Together with a var setting in [open-craft/opencraft](https://github.com/open-craft/opencraft) this should wrap it up).

haikuginger:
@mfitzp, that sounds like a good plan. I'm happy with the existing changes, so if you add that conditional role to the nginx role, then I'm 👍. @gsong, over to you, I think.

mfitzp:
@haikuginger @gsong Those changes have gone in, apologies for the notification/PR spam. I committed the changes then noticed I was missing the internal/external URLs.

playbooks/roles/lets_encrypt/tasks/main.yml
```diff
@@ -0,0 +1,174 @@
...
+
+# Create symlinks to the letsencrypt certificates into
+# the main cert folder. When requesting multi-domain
+# certs they are shared: symlink only required to the first
+
+- name: symlink certificates
+  file: >
+    src=/etc/letsencrypt/live/{{ LETS_ENCRYPT_DOMAINS|first }}/{{ item.src }}
+    dest=/etc/ssl/{{ item.dest }} 
+    state=link
```
haikuginger: Late note here - can you change this to the new Ansible 2.0 style, with the keys as variables rather than as a long string?

playbooks/roles/lets_encrypt/handlers/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: reload nginx
+  service:
+    name: nginx
+    state: reloaded
```
haikuginger: I don't believe this handler is used anywhere, so let's go ahead and remove the folder.

playbooks/roles/lets_encrypt/meta/main.yml
```diff
@@ -0,0 +1,20 @@
...
+# Role includes for role lets_encrupt
+# 
+# Example:
+#
+# dependencies:
+#   - {
+#   role: my_role 
+#   my_role_var0: "foo"
+#   my_role_var1: "bar"
+# }
```
haikuginger: Should the nginx role be included as a dependency? If not, remove meta. If so, add it here, being sure to fix the typo on line 11.

haikuginger:
@mfitzp, three final comments. Once those are done, I'm :+1:, and @gsong will handle the rest of the review.

playbooks/roles/lets_encrypt/meta/main.yml
mfitzp: @haikuginger I'm not sure on this one — the lets_encrypt role requires a running nginx process to work (and with SSL enabled, for it the installation of certificates to be "automatic"). But it doesn't need the nginx role to run _every_ time before it (it will bring up the nginx process if it's not active) which seems to be how dependencies are handled.

What's the usual approach here?

haikuginger: I would go ahead and remove the `meta` folder. Like you said, nginx needs to be there, but we don't need to run the entire role every time, and you're already checking for the nginx service.

mfitzp:
@haikuginger @gsong Thanks, have removed the meta file. Think this is good to go now.

mfitzp:
@haikuginger I've pushed the changes [as discussed](https://github.com/open-craft/opencraft/pull/129#issuecomment-244751967).

If there are any further tweaks I'll push to this branch directly.

haikuginger:
@gsong, we're revisiting this to work out some kinks - specifically, we're revamping the PR to create a cron task that while continue attempting to get a certificate even if it fails during the initial provision - for example, in a case where DNS provisioning hadn't propagated yet.

@mfitzp, I ran into an issue while testing - it looks like you're attempting to copy the nginx conf file directly from the configuration repository. However, this repository can be in different locations on different instances. Let's add a play to template that file into a known location, and then also template that known location into the script. That way, we have a working configuration file regardless of where the repo sits.

gsong:
@haikuginger and @mfitzp Let me know when you think it's ready to be reviewed.

haikuginger:
@gsong - will do. Appreciate the patience!

mfitzp:
@haikuginger That's done in the latest commit. At the moment it's templating the file to /opt (alongside the run_certbot.sh script). Would it be worth putting all this (and the certbot repo) under a subdir (e.g. `/opt/lets_encrypt`) to keep this tidy?

haikuginger:
> Would it be worth putting all this (and the certbot repo) under a subdir (e.g. `/opt/lets_encrypt`) to keep this tidy?

That makes sense to me.

haikuginger:
@mfitzp: One other thing. Let's make the role conditional on both the domain list being present and the LETS_ENCRYPT_EMAIL being present. Both are mandatory.

haikuginger:
@mfitzp, this is almost ready to go. As I had mentioned on open-craft/opencraft#129, the `service nginx restart` needs to happen _after_ the symlinks to the new certificates are made and the configuration file is copied; right now, nginx doesn't use the new certs because it hasn't been restart/reloaded since they were put into place.

Thank you for the templating fix, though. That now appears to work properly.

Please let me know when you've made all of the above changes, and I'll do (hopefully) a final review.

mfitzp:
@haikuginger the `LETS_ENCRYPT_EMAIL` check is in the latest commit (should fail if the email is not set), but just realised you meant also adding the conditional in `edx_sandbox.yml`... will do that.

Good catch on the `nginx restart` / symlink ordering, that was a bit daft.

Should be able to wrap this PR up tonight, will notify you when it's done.

haikuginger:
@mfitzp, on `LETS_ENCRYPT_EMAIL`, I meant a conditional on the role in `edx_sandbox.yml`; not just a "fail if missing" inside the role.

mfitzp:
@haikuginger The suggested changes are all in the latest commit, think this is it!

haikuginger:
@mfitzp, sorry to keep doing this to you. Based on our own integration testing needs, let's add variables `LETS_ENCRYPT_CRON_TASK_HOUR`, `LETS_ENCRYPT_CRON_TASK_DAY`, and `LETS_ENCRYPT_CRON_TASK_MINUTE`. If none of these variables are present, then we should set up the twice-daily crontask like we do right now; if any of them are present, then use them in setting up the crontask, defaulting any that aren't set to `*`.

mfitzp:
@haikuginger It's done. Apologies for the delay, I've been busy the past week with PhD thesis submission. I've set it up as suggested, with two cron-blocks in the playbook (couldn't find a simple way to do it in one).  Since it uses `*` for the missing values, setting only the hour value will run it every minute of that hour (this is normal for cron, but worth mentioning).

It's worth considering the [rate limiting](https://letsencrypt.org/docs/rate-limits/) by Let's Encrypt. It doesn't apply to renewal, and I'm unsure if repeatedly attempting to request the same cert will hit it, but it might be a bit antisocial! One option that occurred to me was to attempt to `curl/wget` the target site e.g. 

```
curl -sSf http://{{ LETS_ENCRYPT_DOMAINS|first }} > /dev/null
```

...prior to the certbot check. This would make sure we only hit the Let's Encrypt server once the site is up and accessible.

haikuginger:
@mfitzp, thanks for this change! The code looks really good, and I appreciate you coming back to it repeatedly. I'll give it a manual test tonight, and I don't anticipate any issues.

I think that we should pass on a "sanity check" regarding the LE ratelimit; if the deployer sets a specific cron interval, we should trust them to know what they're doing. Additionally, there are plenty of valid cases where we'll get a server error but still want to provision a certificate - for example, during the initial provisioning of an OpenEdX box, we'll get a generic nginx 500 error on the homepage, but would still want to set up a cert if possible.

haikuginger:
Okay, with the most recent changes, I'm happy with what we've got. 👍; @gsong, this should be ready for review again.

gsong:
@mfitzp and @haikuginger I've added it to the prioritization queue.

haikuginger:
@mfitzp,

I just got word that, unfortunately, OpenCraft isn't going to be using this work, so we won't be able to help you push it towards release any further. One primary reason was the fact that we wouldn't be able to use this to manage load-balanced instances - quoting one of our engineers below:

> If there are multiple VMs for a single domain, the approach taken in this PR will issue separate certificates for each VM, which is fine in principle; there can be multiple valid certificates for the same domain. However, it is very difficult for the load balancer to route the ACME challenges to the right VM during domain validation, since any of the multiple VMs could have requested a new certificate.

You're welcome to continue with the submission process, although EdX may want an individual contributor agreement from you for any further work.

And, of course, we'd welcome any other contributors who want to help drive this role to completion; we still think it's a good implementation, and addresses the needs of a large number of OpenEdX hosts.

mfitzp:
@haikuginger That issue (multiple VMs) was already identified before developing this PR (see [here](https://openedx.atlassian.net/browse/OSPR-1382?focusedCommentId=208973&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-208973)). The proposed solution was to have certificate request handled by a dedicated server (either a specific VM, or a server in front of it) and then copy the certificates to the VMs after they are retrieved. It was decided that the VMs would still need to run this role (specifically `render nginx cert configuration file` with a tag switch) to configure themselves correctly.

This was a recruitment task, so I won't be following it up for now.

gsong:
@mfitzp Thank you for your PR. We appreciate you taking the time to make the contribution.

Please submit a [signed contributor agreement](http://open.edx.org/sites/default/files/wysiwyg/individual-contributor-agreement.pdf) to us so we can review this pull request. We realize it's additional work in this case, but you only have to do this _once_ and we're hoping you would make future contributions as well.

mfitzp:
@gsong I've signed and submitted an individual contributor agreement, so you should be able to merge this PR as-is if everything is OK.

If it needs further changes it will have to wait a month or so as I'm currently busy looking for work. Alternatively, if someone else can finish it off, feel free to fork my fork and open a new PR.

antoviaque:
@mfitzp @gsong The work up to now is covered by our entity agreement (OpenCraft), since @mfitzp was being paid for that work, and mentioned above that he wouldn't follow-up on this from now on, so that should be enough to either close this, or have someone else picking it up if there are volunteers? (We posted on the ML to ask if there was volunteers). If more work is done though, yes, @mfitzp  you would need to send a copy of the contributor agreement to edX for the rest of the work.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
```diff
@@ -36,8 +36,8 @@ error_page {{ k }} {{ v }};
 
   listen {{ EDXAPP_CMS_SSL_NGINX_PORT }} ssl;
 
-  ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
-  ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
+  include /etc/nginx/ssl-certs.conf;
```
fredsmith, line 39: This file should be in {{ nginx_sites_available_dir }}, not /etc/nginx/

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -118,6 +118,8 @@ error_page {{ k }} {{ v }};
     proxy_pass http://lms-backend;
   }
 
+  {% include "lets_encrypt.j2" %}
```
fredsmith, line 121: It seems like by only including the /.well-known path in the LMS and the CMS, that other sites won't be able to get SSL certificates, and they will fail to be validated by the letsencrypt robot.     I'm guessing that in the default case, eg. a sandbox where the LMS and the CMS listen on 80/443 and the rest of the IDAs listen on other ports, that this code will handle the necessary domain validation.  However, in the case where other IDAs listen on other hostnames, this will fail.

Also, I'd prefer to see this behind a flag so that the .well-known location isn't included on hosts that aren't using letsencrypt.

Thoughts?

fredsmith (changes requested):

fredsmith:
@gsong @antoviaque @mfitzp  - I'd like to pull this branch into edx/configuration to get it across the line.   Do we need any more paperwork to accomplish that, or is @mfitzp 's work covered by the Opencraft agreement?

mfitzp:
@fredsmith I was covered by the Opencraft agreement while working up to the current state of this PR — but have also submitted an independent agreement for any further work if required.

I'll take another look through your earlier comments tomorrow and we can take it from there?

gsong:
@mfitzp When did you submit the paperwork? I haven't seen it come across from legal yet.

mfitzp:
@gsong sent to legal@edx.org on 26 September

gsong:
@mfitzp I'll follow up with legal.

gsong:
@mfitzp Your CLA is all set!

/cc @fredsmith

gsong:
@mfitzp Thanks for all your work on this PR, let us know if you'll have time to continue working on it.

jbarciauskas:
Moving this to waiting on author - @mfitzp let us know if you're ready to move forward

gsong:
@fredsmith Are you still interested in moving this forward? If not, I'm going to close it out.

mfitzp:
@gsong @fredsmith I have time in the coming week to wrap this up, but am not clear on what is needed to get it finished off. As far as I remember it —

This implementation works, using a setup-script/cron hook to successfully install certificates when the server/domain becomes available publicly. The support for this delayed setup was added in response to a need on the opencraft side, although ultimately this didn't meet that (because of load balancing). The result is this implementation is 'more complicated' than it needs to be for non-load balanced, single-server implementations, and won't work for load balanced 

So, the option is either to go ahead with this current implementation (if there is an identifiable purpose use) or roll-back to the simple implementation (earlier in this PR). I guess we could also add support for both(!)

I basically don't know how to proceed, since this was written in response to a specific use-case, which was impossible to meet, I'm not sure what "done" would look like. Any suggestions?

jbarciauskas:
Is there a way to package this such that someone could optionally pip install this role?

gsong:
@edx/devops Any suggestions on how we can move forward on this PR?

gsong:
Closing due to lack of activity.

PR #3316, opened 2016-08-16
Make ansible unbuffered for better experience while logging

nedbat:


util/vagrant/upgrade.sh
```diff
@@ -296,11 +293,8 @@ echo "ora2_version: $TARGET" >> vars.yml
 echo "certs_version: $TARGET" >> vars.yml
 echo "forum_version: $TARGET" >> vars.yml
 echo "xqueue_version: $TARGET" >> vars.yml
-sudo ansible-playbook \
-    --inventory-file=localhost, \
-    --connection=local \
+$ANSIBLE_PLAYBOOK \
```
nedbat, line 299: This use of the macro will swap the `--extra-vars` switch and the `$SERVER_VARS` variable: used to be that `--extra-vars` was first, now `$SERVER_VARS` is first.  Does the ordering of those two matter?

nedbat:
@edx/devops This keeps ansible output flowing with the progress of the playbook, and simplifies the handful of playbook invocations that we have.

util/vagrant/upgrade.sh
feanil, line 299: The ordering does matter as ansible will override earlier vars with those added from later files.  So this likely breaks people that are using values in server vars to override the defaults we provide.

nedbat, line 299: OK, i've moved `$SERVER_VARS` outside of the `$ANSIBLE_PLAYBOOK` variable so as not to disrupt the ordering.

nedbat:
@edx/devops now ok?

feanil:
:+1:

PR #3317, opened 2016-08-16
go-agent Dockerfile now installs the RSA key used to clone private re…

macdiesel:
@edx/pipeline-team PTAL.

FYI the key material is junk, intended to be replaced with the real material when building the container.

### DIFF
--- a/docker/build/go-agent/Dockerfile
+++ b/docker/build/go-agent/Dockerfile
@@ -64,3 +64,17 @@ RUN pip install awscli
 ADD docker/build/go-agent/files/go-agent-start.sh /etc/service/go-agent/run
 ADD docker/build/go-agent/files/go-agent-env-vars /etc/default/go-agent
 RUN update-java-alternatives -s java-7-oracle
+
+# !!!!NOTICE!!!! ---- Runner of this pipeline take heed!! You must replace go_github_key.pem with the REAL key material
+# that can checkout private github repositories used as pipeline materials. The key material here is faked and is only
+# used to pass CI!
+# setup the github identity
+ADD docker/build/go-agent/files/go_github_key.pem /var/go/.ssh/id_rsa
+RUN chmod 600 /var/go/.ssh/id_rsa && \
+    chown go:go /var/go/.ssh/id_rsa
+
+# setup the known_hosts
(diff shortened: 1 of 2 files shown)


docker/build/go-agent/Dockerfile
```diff
@@ -64,3 +64,17 @@ RUN pip install awscli
...
+# setup the github identity
+ADD docker/build/go-agent/files/go_github_key.pem /var/go/.ssh/id_rsa
+RUN chmod 600 /var/go/.ssh/id_rsa && \
+    chown go:go /var/go/.ssh/id_rsa
+
+# setup the known_hosts
+RUN touch /var/go/.ssh/known_hosts && \
+    chmod 600 /var/go/.ssh/known_hosts && \
+    chown go:go /var/go/.ssh/known_hosts && \
+    ssh-keyscan -t rsa,dsa github.com 2>&1 | sort -u - /var/go/.ssh/known_hosts > /var/go/.ssh/tmp_hosts && mv /var/go/.ssh/tmp_hosts /var/go/.ssh/known_hosts
```
feanil: Can you explain why you're doing the temp_hosts dance here?  Why not just output directly from ssh-keyscan to the known hosts file overwriting it? Are you concerned that there might be other hosts in there we care about?

macdiesel: done

feanil:
:+1:

docker/build/go-agent/Dockerfile
```diff
@@ -64,3 +64,17 @@ RUN pip install awscli
 ADD docker/build/go-agent/files/go-agent-start.sh /etc/service/go-agent/run
 ADD docker/build/go-agent/files/go-agent-env-vars /etc/default/go-agent
 RUN update-java-alternatives -s java-7-oracle
+
+# !!!!NOTICE!!!! ---- Runner of this pipeline take heed!! You must replace go_github_key.pem with the REAL key material
+# That can checkout github repositories. The key material here is faked and is only used to pass CI!
```
doctoryes: Nit: Just to be explicit:
`that can checkout private github repositories used as pipeline materials.`
?

doctoryes:
Just a nit 👍

PR #3318, opened 2016-08-16
fix CD scripts so instances launch with the correct key pair

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/playbooks/continuous_delivery/launch_instance.yml
+++ b/playbooks/continuous_delivery/launch_instance.yml
@@ -30,7 +30,7 @@
     ebs_volume_size: 8
     ec2_timeout: 500
     ec2_assign_public_ip: no
-    automtion_prefix: "gocd automation run -- {{ ansible_date_time.iso8601 }} -- "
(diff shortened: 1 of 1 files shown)


playbooks/continuous_delivery/launch_instance.yml
```diff
@@ -55,7 +55,7 @@
     ec2:
       instance_tags: {"Name" : "{{ automtion_prefix }} {{ unique_key_name.stdout }}"}
       region: "{{ ec2_region }}"
-      key_name: "{{ unique_key_name.stdout }}"
+      key_name: "{{ automtion_prefix }} {{ unique_key_name.stdout }}"
```
doctoryes: Nit: It's totally anal - but could `automtion` be globally replaced to `automation`?

doctoryes:
👍  with a small nit

playbooks/continuous_delivery/launch_instance.yml
macdiesel: yes we can!

doctoryes:
🚀 🚢 🇮🇹

PR #3319, opened 2016-08-17
add role for configuring android worker

estute:
## Configuration Pull Request

### DIFF
--- a/playbooks/roles/jenkins_worker/meta/main.yml
+++ b/playbooks/roles/jenkins_worker/meta/main.yml
@@ -19,3 +19,15 @@ dependencies:
   # dependencies for android worker
   - role: android_sdk
     when: android_worker is defined
+    # User/group to manage Android SDK
+    android_user: "jenkins"
+    android_group: "jenkins"
+    # Tarball to download
+    android_download: "android-sdk_r24.4.1-linux.tgz"
+    # Checksum of Android SDK (from: https://developer.android.com/studio/index.html#downloads)
+    android_checksum: "725bb360f0f7d04eaccff5a2d57abdd49061326d"
+    # path to installed android sdk
+    android_home: "/opt/android-sdk-linux"
+    # individual android build targets to be downloaded via the androind sdk manager
+    android_build_targets:
+        - android-23
--- a/playbooks/roles/jenkins_worker/meta/main.yml
+++ b/playbooks/roles/jenkins_worker/meta/main.yml
@@ -15,3 +15,7 @@ dependencies:
   # dependencies for sitespeed worker
   - role: sitespeedio
     when: sitespeed_worker is defined
+
+  # dependencies for android worker
+  - role: android_sdk
+    when: android_worker is defined
--- a/util/packer/jenkins_worker_android.json
+++ b/util/packer/jenkins_worker_android.json
@@ -1,0 +1,59 @@
+{
+  "variables": {
+    "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
+    "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
+    "playbook_remote_dir": "/tmp/packer-edx-playbooks",
+    "ami": "{{env `JENKINS_WORKER_AMI`}}",
+    "delete_or_keep": "{{env `DELETE_OR_KEEP_AMI`}}",
+    "remote_branch": "{{env `REMOTE_BRANCH`}}"
+  },
+  "builders": [{
+    "type": "amazon-ebs",
+    "access_key": "{{user `aws_access_key`}}",
+    "secret_key": "{{user `aws_secret_key`}}",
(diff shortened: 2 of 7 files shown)


estute:
@jzoldak @jibsheet Could you guys take a peek at this. After speaking with ed/kevin/feanil, we decided that ubuntu-make was not reliable enough right now, so I deferred to just curling/tarring the sdk directly from google. Maybe at a later date, we can use a PPA.

playbooks/roles/jenkins_worker/meta/main.yml
```diff
@@ -15,3 +15,7 @@ dependencies:
   # dependencies for sitespeed worker
   - role: sitespeedio
     when: sitespeed_worker is defined
+
+  # dependencies for android worker
+  - role: android
```
jzoldak: The role name "android" seems vague or potentially misleading to me. How about android_sdk?

util/packer/jenkins_worker_android.json
```diff
@@ -0,0 +1,67 @@
...
+    "inline": ["sudo add-apt-repository ppa:git-core/ppa -y"]
+  }, {
+    "type": "shell",
+    "inline": ["sudo apt-get update"]
+  }, {
+    "type": "shell",
+    "inline": ["sudo apt-get -y install git-core"]
+  }, {
+    "type": "shell",
+    "inline": ["sudo apt-get -y install libmysqlclient-dev"]
```
jzoldak: Why do we need to install MySQL on the machine?

util/packer/jenkins_worker_android.json
```diff
@@ -0,0 +1,67 @@
...
+  }, {
+      "type": "file",
+      "source": "../../playbooks/edx-east",
+      "destination": "{{user `playbook_remote_dir`}}/edx-east"
+  }, {
+    "type": "file",
+    "source": "../../requirements.txt",
+    "destination": "{{user `playbook_remote_dir`}}/requirements.txt"
+  }, {
+    "type": "shell",
```
jzoldak: Rather than do these steps, let's see if we can copy down util/install/ansible-bootstrap.sh to the system and then execute it.

Or maybe you were going to put that in the next PR? I'm not sure.

playbooks/roles/android_sdk/meta/main.yml
```diff
@@ -0,0 +1,21 @@
+---
```
jzoldak: Unless there is a compelling reason we shouldn't check in this commented-out file, but instead add one later if we need one.

estute: Fair point. Taking it out.

jzoldak:
LGTM. 👍  after squashing.

util/packer/jenkins_worker_android.json
```diff
@@ -0,0 +1,60 @@
...
+  }, {
+    "type": "file",
+    "source": "../../util/install/ansible-bootstrap.sh",
+    "destination": "{{user `playbook_remote_dir`}}/ansible-bootstrap.sh"
+  }, {
+    "type": "shell",
+    "inline": ["cd {{user `playbook_remote_dir`}}",
+      "export CONFIGURATION_VERSION='{{user `remote_branch`}}'",
+      "sudo bash ./ansible-bootstrap.sh", 
+      "sudo rm -Rf $HOME/.ansible" ]
```
jzoldak: @estute is there a reason for the capital R here? I saw that you had a little r in the PR that changes the bootstrap script itself.

estute: I use big R out of habit. The other was copy+paste.

jzoldak: oic. if little r works then I'd say change it to that here (so that we know it works), and also put in a TODO comment for removing the line when the upstream changes are merged in.

feanil:
So it looks like the android_sdk role is actually very specific to jenkins right now.  Can we change the name to indicate that or make the role more generic?  This seems similar to the oraclejdk role which does a better job of being a reusable role and generic and expects various settings to be overridden when called via `meta/main.yml`

eg. https://github.com/edx/configuration/blob/5ef742db15ddab60681fe3bca37d9c0eafd12ffe/playbooks/roles/jenkins_master/meta/main.yml

estute:
@feanil whenever you get a chance

playbooks/roles/jenkins_worker/meta/main.yml
```diff
@@ -15,3 +15,19 @@ dependencies:
...
     when: sitespeed_worker is defined
+
+  # dependencies for android worker
+  - role: android_sdk
+    when: android_worker is defined
+    # User/group to manage Android SDK
+    android_user: "jenkins"
+    android_group: "jenkins"
+    # Tarball to download
+    android_download: "android-sdk_r24.4.1-linux.tgz"
```
feanil, line 26: NIT: You don't necessarily need to override defaults if you are happy with the one set in the role but I also understand pinning them here so that they don't change out from underneath you.

feanil:
:+1:

PR #3320, opened 2016-08-17
SOL-2032: Theming broken for "From Scratch" AMI builds

saleem-latif:
## Configuration Pull Request

**Description of the issue:**
There appears to be an issue with the theming framework when constructing edxapp / edx-platform AMI builds in the "from scratch" configuration. 
The log from Ansible is: https://gist.github.com/maxrothman/5f457577517f7d6f12b3c78d2562f1ef
Here's what Max Rothman determined (from the release HipChat room):

> In secure config, EDXAPP_COMPREHENSIVE_THEME_DIRS is set to ["{{ edxapp_code_dir }}/themes", "{{ THEMES_CODE_DIR }}/edx-platform"]
> thus when then "assign theme's group to all theme's directories" task in edx_themes runs, {{ edxapp_code_dir }}/themes is created
> this takes place before the edx_platform role runs because it depends on the edx_themes role
> then when edx_platform attempts to check out the edx_platform repo to {{ edxapp_code_dir }} it fails because that themes dir is already there.

**Problem:**
edx-themes role assigns 'edx-themes' group to all the directories related to theming, if a directory does not exist it is created (default behavior of ansible). 'edx-themes' group is then used grant read/write access to theme's users like edxapp, ecommerce etc. Since, edx_themes is a dependency of edxapp, it runs before edxapp. The Problem is that one theme directory is present inside edx-platform repo and when the task runs it creates that directory. This happens before edxapp role could clone edx-platform repo.

**Solution:**
We can leave the assignment  of 'edx-themes' group to the directory present inside edx-platform repo because edxapp will already have read/write permissions for this directory. That is why we are excluding this directory from `ECOMMERCE_COMPREHENSIVE_THEME_DIRS` before passing to the edx_themes role.

### DIFF
--- a/playbooks/roles/edxapp/meta/main.yml
+++ b/playbooks/roles/edxapp/meta/main.yml
@@ -7,6 +7,4 @@ dependencies:
   - role: edx_themes
     theme_users:
       - "{{ edxapp_user }}"
-    additional_theme_dirs:
-      - "{{ EDXAPP_COMPREHENSIVE_THEME_DIRS }}"
     when: "{{ EDXAPP_ENABLE_COMPREHENSIVE_THEMING }}"
--- a/playbooks/roles/ecommerce/meta/main.yml
+++ b/playbooks/roles/ecommerce/meta/main.yml
@@ -25,8 +25,6 @@ dependencies:
   - role: edx_themes
     theme_users:
       - "{{ ecommerce_user }}"
-    additional_theme_dirs:
-      - "{{ ECOMMERCE_COMPREHENSIVE_THEME_DIRS }}"
     when: "{{ ECOMMERCE_ENABLE_COMPREHENSIVE_THEMING }}"
 
   - oraclejdk
--- a/playbooks/roles/edx_themes/defaults/main.yml
+++ b/playbooks/roles/edx_themes/defaults/main.yml
@@ -13,7 +13,7 @@
 
 themes_service_name: "edx-themes"
 themes_user: "{{ themes_service_name }}"
-themes_group: "{{ themes_user }}"
+themes_group: "www-data"
 themes_home: "{{ COMMON_DATA_DIR }}/{{ themes_service_name }}"
 
 THEMES_CODE_DIR: "{{ themes_home }}/{{ themes_service_name }}"
--- a/playbooks/roles/edx_themes/tasks/main.yml
+++ b/playbooks/roles/edx_themes/tasks/main.yml
@@ -28,8 +28,6 @@
 #     theme_users:
 #       - ecommerce
 #       - edxapp
-#     additional_theme_dirs:
-#      - /edx/app/ecommerce/ecommerce/ecommerce/themes
 #     when do_setup_themes
 #
 
@@ -42,19 +40,6 @@
     owner: "{{ themes_user }}"
     group: "{{ themes_group }}"
     mode: "g+rw"
-  tags:
-    - install
-    - install:base
-
-- name: assign theme's group to all theme's directories
-  file:
-    path: "{{ item }}"
-    state: directory
-    recurse: yes
-    group: "{{ themes_group }}"
-    mode: "g+rw"
-  with_items: additional_theme_dirs
-  when: additional_theme_dirs is defined
   tags:
     - install
     - install:base


saleem-latif:
@mattdrayer , @fredsmith , @e0d -- FYI

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -115,6 +115,7 @@ ECOMMERCE_THEME_SCSS: 'sass/themes/default.scss'
 
 # Directory name inside edx-themes repo that contain ecommerce themes
 ECOMMERCE_THEMES_DIR_NAME: 'ecommerce'
+ecommerce_themes_dir: '{{ ecommerce_code_dir }}/ecommerce/themes'
```
mattdrayer: should this be `{{ ecommerce_code_dir }}/{{ ECOMMERCE_THEMES_DIR_NAME }}/themes` ?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1043,7 +1043,7 @@ cms_env_config:
 
 # install dir for the edx-platform repo
 edxapp_code_dir: "{{ edxapp_app_dir }}/edx-platform"
-
+edxapp_edx_platform_themes_dir: "{{edxapp_code_dir}}/themes"
```
mattdrayer: Please add a comment describing the purpose/intent for this new parameter

mattdrayer:
@saleem-latif, for future reference, can you please update the description for this PR to include information about how/why these changes resolve the issue?

mattdrayer:
I've read up on Jinja2 filters for Ansible -- I can see now what is being done:  http://docs.ansible.com/ansible/playbooks_filters.html#set-theory-filters

playbooks/roles/edxapp/meta/main.yml
```diff
@@ -15,5 +15,5 @@ dependencies:
     theme_users:
       - "{{ edxapp_user }}"
     additional_theme_dirs:
-      - "{{ EDXAPP_COMPREHENSIVE_THEME_DIRS }}"
+      - "{{ EDXAPP_COMPREHENSIVE_THEME_DIRS|difference(edxapp_edx_platform_themes_dir) }}"
```
mattdrayer: I think it is important to include some information here describing why we are doing this filtering -- it is confusing to have "comprehensive themes" and "themes", and not clear why we need to remove list items from `EDXAPP_COMPREHENSIVE_THEME_DIRS` which match `edxapp_edx_platform_themes_dir`

mattdrayer:
@maxrothman FYI

saleem-latif:
@mattdrayer , @maxrothman , @fredsmith I have completed the  checklist on https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist, tested changes on sandbox and also successfully built from scratch ami for edxapp. 
Kindly take a look at the changes and share your thoughts.

maxrothman:
If the themes directory is world-readable (so www-data can read it) and you've verified that themes work in sandboxes/stage/etc., then 👍

saleem-latif:
@maxrothman I have updated group for theme's dir to be "www-data" so that they are world readable, @mattdrayer can you take a look at the final change set and share your thoughts ?

mattdrayer:
I am okay making the change to "www-data" -- it might be good to have @feanil weigh in as well, because he was originally involved in the themes group permissions work.

feanil:
generally looks good. Have you built an AMI from scratch to verify the changes?

jibsheet:
@saleem-latif @mattdrayer after @feanil merged sandbox-secure this morning, setting a git ssh key for the themes role, it began leaking an ssh private key into the jenkins job logs.  We're going to have to rotate this key now.

the git_clone role contains the ssh key as a parameter, rather than the way edxapp does git which passes it as an environment variable (which is not logged).

You can either solve this by switching to the edxapp pattern
https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/tasks/deploy.yml#L55
or possibly by setting no_log on the task.

This has been wrong since the initial merge in May, but back then there was no configured ssh private key to leak.

saleem-latif:
@feanil Yes I have tested this change set on sandbox and also built a from scratch AMI.
@jibsheet I have opened another PR https://github.com/edx/configuration/pull/3334, to resolve ssh private key leakage.

saleem-latif:
I think all concerns are addressed at this point, so I am going ahead with the merge.

PR #3324, opened 2016-08-18
Add edx-east explanation.

doctoryes:
@nedbat @cpennington Could you check this explanation and suggest edits _or_ more information? Thanks.

@feanil Please review as well.
## Configuration Pull Request


cpennington:
:+1:

Would it be crazy to just symlink edx-east to some more clearly named directory?

doctoryes:
@feanil Thoughts on a symlink?

nedbat:
:+1: Definitely better (we can quibble about line-wrapping....)

doctoryes:
This is a documentation change only. Merging.

PR #3325, opened 2016-08-19
neo4j_3_fixes

fredsmith:


e0d:
Let's make the address a default, then :+1:

PR #3326, opened 2016-08-19
Update sandbox ami

maxrothman:


e0d:
:+1: Make sure the ruby version corresponds to expectation before the release ceremony.

PR #3329, opened 2016-08-22
Add an OEP-2 compliant openedx.yaml file

cpennington:
This adds an `openedx.yaml` file, as described by OEP-2:
http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html

The data in this file was transformed from the contents of
edx/repo-tools-data:repos.yaml


cpennington:
@edx/devops: Labeling this repo as yours.

openedx.yaml
```diff
@@ -0,0 +1,8 @@
+# This file describes this Open edX repo, as described in OEP-2:
+# http://open-edx-proposals.readthedocs.io/en/latest/oeps/oep-0002.html#specification
+
+nick: conf
```
jibsheet, line 4: Can we have config, not conf?

openedx.yaml
```diff
@@ -0,0 +1,8 @@
...
+oeps: {}
+openedx-release: {ref: master}
```
jibsheet, line 6: Having trouble finding the doc for this in the spec, but the openedx-release doesn't actually come from master unless you mean edx-platform's release branch.

openedx.yaml
cpennington: This is for some open-source team tooling that uses very short textual labels.

openedx.yaml
nedbat: This means, when it is time to tag an Open edX release, we are going to tag the master branch in edx/configuration.

openedx.yaml
nedbat: @cpennington seems like we should add some mention of these keys to OEP-2.

jibsheet: That would be grand.

cpennington: Ok. I'd almost rather have them as part of an OEP describing how we do named releases, though.

cpennington: Ok, I now have a PR open to add docs of those keys to OEP-2.

cpennington:
@jibsheet: ok to merge?

openedx.yaml
feanil: Any reason why it can't just be 'configuration'?  Seems like a small enough word.

cpennington: See the definition here: https://github.com/edx/open-edx-proposals/pull/18/files#diff-13c9b7b8133d4bcf6ab907c7f0065187R53

jibsheet:
I would love Open Source to fix their tools that requires conf as a nickname, is there a ticket for that?
But this is fine to merge 👍

cpennington:
Not sure why everyone is getting so hung up on the short nickname...

PR #3332, opened 2016-08-22
Switch back to Ruby 1.9.3

tobz:
## Configuration Pull Request


feanil:
@tobz looks like this is not the exact version we had before, any particular reason for that?

tobz:
Fixed.

feanil:
:+1:

PR #3333, opened 2016-08-22
Revert "Update sandbox ami"

jibsheet:
Reverts edx/configuration#3326 because newer ruby is showing worse memory usage/perf in production.

@feanil


feanil:
:+1:

PR #3335, opened 2016-08-23
SOL-1994: Automate full Otto provisioning with multitenant support

saleem-latif:
## Configuration Pull Request

Hi @mattdrayer, @feanil , @asadiqbal08 
Kindly take a look at the changes and share your thoughts.

**Description:**
We need to create an ansible role that adds `Site`, `Partner` and `SiteConfiguration` records with appropriate values.
Make sure that `edx` is used as the default partner and there is entry for `edx` in payment processor config. This Role will be executed during sandbox and devstack provisioning.

- [x] Have completed [Ops ansible testing checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)


playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -42,6 +42,7 @@ ECOMMERCE_TIME_ZONE: 'UTC'
 ECOMMERCE_LANGUAGE_CODE: 'en-us'
 ECOMMERCE_EDX_API_KEY: 'PUT_YOUR_API_KEY_HERE'  # This should match the value set for edxapp
 ECOMMERCE_ECOMMERCE_URL_ROOT: 'http://localhost:8002'
+ECOMMERCE_ECOMMERCE_DOMAIN: 'localhost:8002'
```
mattdrayer, line 44: Can we move this above ECOMMERCE_ECOMMERCE_URL_ROOT and then reference it within that parameter?

playbooks/roles/otto_multitenant_setup/defaults/main.yml
```diff
@@ -0,0 +1,20 @@
...
+##
+# Defaults for role otto_multitenant_setup
+#
+
+#
+# vars are namespaced with the module name.
+#
+otto_multitenant_setup_role_name: otto_multitenant_setup
+
+ECOMMERCE_PARTNER_CODE: edx
```
mattdrayer: Wondering if, as a default, the partner code should be `openedx` and then we override for our own installation

playbooks/roles/otto_multitenant_setup/tasks/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    --site-domain='{{ ECOMMERCE_ECOMMERCE_DOMAIN }}'
+    --site-name='{{ ECOMMERCE_ECOMMERCE_DOMAIN }}'
+    --partner-code='{{ ECOMMERCE_PARTNER_CODE }}'
+    --partner-name='{{ ECOMMERCE_PARTNER_NAME }}'
+    --lms-url-root='{{ ECOMMERCE_LMS_URL_ROOT }}'
+    --theme-scss-path='sass'
+    --payment-processors='cybersource,paypal'
+    --client-id='{{ ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_KEY }}'
+    --client-secret='{{ ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_SECRET }}'
+    --from-email='sandbox-notifications@edx.org'
```
mattdrayer: This value should come from settings as well -- I believe it already exists

mattdrayer:
Just a couple initial comments -- looks good so far

playbooks/roles/otto_multitenant_setup/tasks/main.yml
saleem-latif: @mattdrayer I am not sure what you mean by settings, However, I have moved this variable to defaults to make it overridable by calling playbooks.

util/jenkins/ansible-provision.sh
```diff
@@ -275,9 +275,11 @@ EDXAPP_ECOMMERCE_API_URL: "https://ecommerce-${deploy_host}/api/v2"
 EDXAPP_COURSE_CATALOG_API_URL: "https://catalog-${deploy_host}/api/v1"
 
 ECOMMERCE_ECOMMERCE_URL_ROOT: "https://ecommerce-${deploy_host}"
+ECOMMERCE_ECOMMERCE_DOMAIN: "ecommerce-${deploy_host}"
```
asadiqbal08, line 289: is there any specific reason to use `ECOMMERCE` two times with underscore? same for `ECOMMERCE_ECOMMERCE_URL_ROOT:` key

playbooks/vagrant-devstack.yml
```diff
@@ -10,6 +10,9 @@
     EDXAPP_NO_PREREQ_INSTALL: 0
     COMMON_SSH_PASSWORD_AUTH: "yes"
     EDXAPP_LMS_BASE: 127.0.0.1:8000
+    ECOMMERCE_ECOMMERCE_DOMAIN: "localhost:8002"
```
asadiqbal08, line 13: why not just `ECOMMERCE_DOMAIN` ? any reason please ?

util/jenkins/ansible-provision.sh
mattdrayer: @asadiqbal08 -- the first "ECOMMERCE" is actually a trigger for ansible to know which settings file to send each parameter.  It is removed during the provisioning process.  

So `ECOMMERCE_ECOMMERCE_DOMAIN` --> `ECOMMERCE_DOMAIN` in the configuration file and `ECOMMERCE_LMS_URL_ROOT` --> `LMS_URL_ROOT`

asadiqbal08: 👍

mattdrayer:
👍

saleem-latif:
@feanil, @jibsheet  please take a look at the PR when you have time.

jibsheet:
@saleem-latif Feanil is on paternity leave for a few months.  I will attempt to look at this, but I'm tagging @edx/devops as I'll be traveling next week.

playbooks/edx-east/otto_multitenant_setup.yml
```diff
@@ -0,0 +1,10 @@
+- name: Setup Site and Site Configurations for Otto
+  hosts: all
+  become: True
+  gather_facts: True
+  vars_files:
+    - "roles/common_vars/defaults/main.yml"
+    - "roles/edxapp/defaults/main.yml"
+    - "roles/ecommerce/defaults/main.yml"
```
maxrothman, line 7: Referencing individual vars files should be avoided. For example, if you want access to common_vars, you should make it a dependency of your role.

maxrothman:
What are the actual steps this is supposed to perform? Is it just running that management command? A playbook and a role seems overkill for that.

Also is running this management command a pre-requisite requirement of provisioning the ecommerce service?

playbooks/edx-east/otto_multitenant_setup.yml
saleem-latif: @maxrothman I have removed "common_vars" from here and added it as a dependancy

saleem-latif:
@maxrothman Currently we are running just this one management command but we will be adding more tasks in the future, e.g. once payment processor refactor work (https://openedx.atlassian.net/browse/SOL-1993) is complete, we will add more tasks into this role.

Management command is not a pre-requisite, it should run once ecommerce and edxapp roles are complete.

maxrothman:
@saleem-latif Why did you make a separate role? When/how do you envision the role being run?

saleem-latif:
@maxrothman This role will run at the end of [devstack](https://github.com/edx/configuration/blob/saleem-latif/SOL-1789/playbooks/vagrant-devstack.yml#L54) and [sandbox](https://github.com/edx/configuration/blob/saleem-latif/SOL-1789/util/jenkins/ansible-provision.sh#L393) provisioning. This is added at the end to make sure that lms, studio and ecommerce setup along with their databases is complete. 
[Here](https://github.com/edx/configuration/tree/master/playbooks/roles/oauth_client_setup) is another role that I used as a reference.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -41,7 +41,8 @@ ECOMMERCE_SECRET_KEY: 'Your secret key here'
 ECOMMERCE_TIME_ZONE: 'UTC'
 ECOMMERCE_LANGUAGE_CODE: 'en-us'
 ECOMMERCE_EDX_API_KEY: 'PUT_YOUR_API_KEY_HERE'  # This should match the value set for edxapp
-ECOMMERCE_ECOMMERCE_URL_ROOT: 'http://localhost:8002'
+ECOMMERCE_ECOMMERCE_DOMAIN: 'localhost:8002'
+ECOMMERCE_ECOMMERCE_URL_ROOT: 'http://{{ ECOMMERCE_ECOMMERCE_DOMAIN  }}'
```
maxrothman, line 45: should be https

maxrothman:
Reaching into roles and pointing at their vars files is an antipattern and causes unexpected behavior. If a role needs access to state provided by another role, then it's usually better suited to being a part of that role. I'd need to understand better what the management command does, but I strongly suspect that what you're trying to do is better suited to an initial data migration.

playbooks/roles/ecommerce/defaults/main.yml
saleem-latif: this value is used in devstack, and https does not work on devstack that is why we have to use http here.

mattdrayer: Is is possible to split apart this value and have it depend on a protocol parameter that can be set based on whether it is a devstack installation or something different?

mattdrayer: Something like `'{{ ECOMMERCE_ECOMMERCE_PROTOCOL }}://{{ ECOMMERCE_ECOMMERCE_DOMAIN }}'`

saleem-latif:
@maxrothman I looked at making tasks of automating otto a part of either edxapp or ecommerce. but the problem with that is we only want this task running for sandbox or devstack provisioning and not for any other scenario. So, adding task in edxapp or ecommerce would cause them running in all palybooks that include these roles.

playbooks/roles/ecommerce/defaults/main.yml
saleem-latif: @mattdrayer I do not think that would be necessary as this variable is overridden from external variables. e.g. in case of sandbox provisioning [this](https://github.com/edx/configuration/blob/master/util/jenkins/ansible-provision.sh#L288) will override `ECOMMERCE_ECOMMERCE_PROTOCOL`.

saleem-latif: according to my understanding this variable is used only for devstack and local installations and for other installations it will be overridden by external variables.

mattdrayer:
@saleem-latif @maxrothman -- we need to wrap this PR up in the next 24 hours in order to begin work on another job -- is the current state of this change set acceptable-enough to move forward?  If not, what concrete modifications need to be made in order for the branch to be able to merge?

saleem-latif:
@maxrothman I also looked at data migrations approach, you are suggesting we use django data migrations to inject this data right ? According to my understanding we can only use static data in these migrations, However, for otto automation we need DNS name for ecommerce and edxapp which are not known until sandbox provision has been initiated so I do not think that data migration is going to serve our purpose. 
Please let me know if I am missing something, it would be helpful to see an example usage if available.

maxrothman:
@mattdrayer did you ever update @saleem-latif based on the conversation we had?

mattdrayer:
I gave a summary of our discussion and asked to explore a couple different approaches, however I am hoping you can provide a more detailed description of what exactly you would like to see in the way of a solution.

maxrothman:
Ok, @saleem-latif's last comment didn't seem to reflect that. We might want to take this discussion to the ticket, but I'd like to start by seeing more concrete requirements.

mattdrayer:
I would prefer to discuss on the ticket, yes

PR #3336, opened 2016-08-23
We don't want to use a file, we have an ssh agent

jibsheet:
Also, this file/path won't exist later on sandboxes which also
source this file when running /edx/bin/update

@feanil 

Built a sandbox plain and from scratch to test
http://jenkins.edx.org:8080/job/ansible-provision/8857/consoleFull
http://jenkins.edx.org:8080/job/ansible-provision/8860/console


fredsmith:
👍

PR #3337, opened 2016-08-23
Improvements to install_stack.sh

nedbat:


nedbat:
@gsong What do you think?

feanil:
:+1:

util/install/install_stack.sh
```diff
@@ -106,6 +115,7 @@ exec > >(tee install-$(date +%Y%m%d-%H%M%S).log) 2>&1
 echo "Capturing output to install-$(date +%Y%m%d-%H%M%S).log."
 
 export OPENEDX_RELEASE=$release
+echo "Installing release '$OPENEDX_RELEASE'"
```
gsong, line 118: I think `${ENV_VAR}` is the preferred interpolation these days.

util/install/install_stack.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-# Stop if any command fails
+# Setting OPENEDX_DEBUG makes this more verbose.
+if [[ ! -z "$OPENEDX_DEBUG" ]]; then
```
gsong: Would `[[ $OPENEDX_DEBUG ]]` be sufficient?

util/vagrant/upgrade.sh
```diff
@@ -1,6 +1,11 @@
 #!/usr/bin/env bash
 
-# Stop if any command fails
+# Setting OPENEDX_DEBUG makes this more verbose.
+if [[ ! -z "$OPENEDX_DEBUG" ]]; then
```
gsong: Same here, would `[[ $OPENEDX_DEBUG ]]` be sufficient?

util/install/install_stack.sh
nedbat: I'm told there is no difference between the two if there are no options being used, and no ambiguity about the ends of the name.

gsong: Functionally there's no difference. I just like to be consistent.

util/install/install_stack.sh
nedbat: OK, made this change.

util/install/install_stack.sh
nedbat: I usually try for the more noise-free option.

gsong:
👌

PR #3338, opened 2016-08-23
Improved argument parsing

nedbat:
The release defaults to $OPENEDX_RELEASE, so that it usually doesn't
have to be provided. The switches now come before the positional
arguments.


nedbat:
@feanil @gsong Another improvement to install_stack.sh

gsong:
👍 

@nedbat You are going into shell script nerdom here.

PR #3339, opened 2016-08-23
Mark the start and end time of the installation

nedbat:


nedbat:
@gsong @feanil Another improvement, just suggested by sboday in Slack.

gsong:
👍

PR #3341, opened 2016-08-24
Sandboxes don't run rabbit behind an ELB

jibsheet:
If you had a sandbox behind an ELB, this would take the whole sandbox
in/out of the ELB during an ansible run?

@edx/devops


fredsmith:
👍

PR #3342, opened 2016-08-24
Pull down ec2 and route53 from ansible 2 for assume role support

jibsheet:
Our 1.9 versions only pass keys as credentials.  Ansible 2 supports aws sts
token based auth.  I've only updated launch_ec2 since that is the only role
currently using assume role.

Note this upgrades boto from 2.32.1 to 2.33.0 (the smallest version bump that supports a new ec2 argument passed by ansible2).  I didn't go all the way to 2.42.0, but can test that in a separate PR if we want.

@edx/devops

### DIFF
--- a/playbooks/library/ec2_2_1_1_0
+++ b/playbooks/library/ec2_2_1_1_0
@@ -1,0 +1,1434 @@
+#!/usr/bin/env python
+# This file is part of Ansible
+#
+# Ansible is free software: you can redistribute it and/or modify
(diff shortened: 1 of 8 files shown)


jibsheet:
Testing the existing ansible provision with the upgrades

http://jenkins.edx.org:8080/job/ansible-provision/8880/console
and I also cut a stage AMI, which isn't impacted by launch_ec2, but does involve boto.

fredsmith:
LGTM 👍

PR #3343, opened 2016-08-24
Docker workaround for ansible provisioning.

jibsheet:
Because /var/home_jenkins on our jenkins test containers is mounted from
~/path/to/git/repo/.docker_volumes/jenkins_home you end up with an ssh
socket for ansible which is way past the 108 character unix limit for
filenames of sockets.

@edx/devops

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -22,6 +22,13 @@ env
 export PYTHONUNBUFFERED=1
 export BOTO_CONFIG=/var/lib/jenkins/${aws_account}.boto
 
+# docker on OS-X includes your Mac's home directory in the socket path
+# that SSH/Ansible uses for the control socket, pushing you over
+# the 108 character limit.
+if [ -f /.dockerenv ]; then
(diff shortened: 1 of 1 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -22,6 +22,13 @@ env
 export PYTHONUNBUFFERED=1
 export BOTO_CONFIG=/var/lib/jenkins/${aws_account}.boto
 
+# docker on OS-X includes your Mac's home directory in the socket path
+# that SSH/Ansible uses for the control socket, pushing you over
+# the 108 character limit.
+if [ -f /.dockerenv ]; then
```
fredsmith, line 28: is there a reason not to do this everywhere?

jibsheet: We don't want to use /tmp in general (shared space, security concerns, etc) so I didn't stick it in ansible.cfg.  This is why ansible defaults to $HOME/.ansible/cp/.

I couldn't figure out a way to make ansible.cfg dynamic or I would have tried that.

fredsmith:
👍

PR #3344, opened 2016-08-24
Be more careful about what env we log in ansible-provision

jibsheet:
Now that it can use AWS tokens for auth, we need to be careful about
just dumping env into logs (and thus into splunk).

@edx/devops

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -18,7 +18,10 @@
 # - environment
 # - name_tag
 set -x
-env
+
+# Seeing the environment is fine, spewing secrets to the log isn't ok
+env | grep -v AWS | grep -v ARN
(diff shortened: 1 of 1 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -18,7 +18,10 @@
 # - environment
 # - name_tag
 set -x
-env
+
+# Seeing the environment is fine, spewing secrets to the log isn't ok
+env | grep -v AWS | grep ARN
```
fredsmith: do we really only want to see output that has the string 'ARN', or should both grep's be -v?

jibsheet: No, you're totally right, fixed.

fredsmith:
👍

PR #3346, opened 2016-08-24
add the other tools needed to test the android app

estute:
## Configuration Pull Request

### DIFF
--- a/playbooks/roles/android_sdk/tasks/main.yml
+++ b/playbooks/roles/android_sdk/tasks/main.yml
@@ -38,7 +38,7 @@
   become: yes
   become_user: "{{ android_user }}"
 # Use the android sdk manager to install the build tools necessary for the edx mobile app
-- name: Install Android API levels
+- name: Install other Android tools
   shell: "echo 'y' | {{ android_home }}/tools/android update sdk --no-ui --filter {{ ','.join(android_tools) }}"
   become: yes
   become_user: "{{ android_user }}"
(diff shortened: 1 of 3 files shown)


estute:
@jzoldak @benpatterson @jibsheet

playbooks/roles/android_sdk/tasks/main.yml
```diff
@@ -37,3 +37,8 @@
   shell: "echo 'y' | {{ android_home }}/tools/android update sdk --no-ui --filter {{ ','.join(android_build_targets) }}"
   become: yes
   become_user: "{{ android_user }}"
+# Use the android sdk manager to install the build tools necessary for the edx mobile app
+- name: Install Android API levels
```
jzoldak: copy/pasted. should be updated.

jibsheet: Alternatively, you could combine the tasks and just do a with_items since you're running the 'android update sdk' twice.  I'd be fine with just a name: change as JZ suggests.

estute: Oops, changing the name. I know it seems a bit redundant, but I plan on adding more tests, and separating the two download types (build targets and other) will help as they must be tested differently.

estute:
@jzoldak updated

playbooks/roles/android_sdk/tasks/main.yml
jibsheet: Makes sense.

jibsheet:
👍

benpatterson:
👍

PR #3334, opened 2016-08-23
SOL-2037: Make sure ssh private keys are not leaked into the jenkins job logs

saleem-latif:
## Configuration Pull Request

**Description:**
After the merge of sandbox-secure, setting a git ssh key for the themes role, it began leaking an ssh private key into the jenkins job logs. We're going to have to rotate this key now.

@mattdrayer , @jibsheet -- FYI

Make sure that the following steps are done before merging
- [x] @devops team member has commented with :+1:
- [ ] are you adding any new default values that need to be overriden when this goes live?  
  - [ ] Open a ticket (DEVOPS) to make sure that they have been added to secure vars.
  - [ ] Add an entry to the CHANGELOG.

PR #3348, opened 2016-08-25
Revert "SOL-2037: Make sure ssh private keys are not leaked into the jenkins job logs"

jibsheet:
Reverts edx/configuration#3334

As documented on the PR, this breaks builds (including builds kicked off to test it).

@fredsmith @feanil thumb please to clean up master for builds.


fredsmith:
👍

PR #3349, opened 2016-08-25
Fix openstack requirements installation

haikuginger:
We discovered that due to a timing issue, the previous Swift storage role didn't work properly with Eucalyptus. There was a migration that required storage to be available, whereas the Swift package wasn't installed until later. This configuration change results in the Swift package being installed while other OpenEdX requirements are installed, meaning that storage is immediately available, and all migrations work properly.

**Dependencies**: None

**Sandbox URL**: http://swift-pr-testing.stage.opencraft.hosting/

**Testing instructions**:
1. Using the sandbox, which is provisioned with Swift storage, upload an image to the forums for the demo course. Observe that the image is embedded correctly, with an accurate path, pointing to a Swift storage location, and that viewing the forum once posted shows that image inline.
2. Using the sandbox, enroll in the demo course and answer some questions. Then, log into the admin panel for that course, and observe that CSV analytics files can be downloaded from Swift storage.

**Reviewers**
- [ ] @itsjeyd 
- [ ] edX reviewer[s] TBD

### DIFF
--- a/playbooks/roles/openstack/tasks/main.yml
+++ b/playbooks/roles/openstack/tasks/main.yml
@@ -50,17 +50,8 @@
     dest: "{{ COMMON_OBJECT_STORE_LOG_SYNC_SCRIPT }}"
   when: COMMON_OBJECT_STORE_LOG_SYNC
 
-# Install openstack python requirements into {{ edxapp_venv_dir }}
-- name : Install python requirements
-  # Need to use command rather than pip so that we can maintain the context of our current working directory;
-  # some requirements are pathed relative to the edx-platform repo.
-  # Using the pip from inside the virtual environment implicitly installs everything into that virtual environment.
-  command: >
-    {{ edxapp_venv_dir }}/bin/pip install {{ COMMON_PIP_VERBOSITY }} -i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w -r {{ openstack_requirements_file }}
-    chdir={{ edxapp_code_dir }}
-  sudo_user: "{{ edxapp_user }}"
-  environment: "{{ edxapp_environment }}"
-  when: edxapp_code_dir is defined
-  tags:
-    - install
-    - install:app-requirements
+# We want to ensure that OpenStack requirements are installed at the same time as edxapp
+# requirements so that they're available during the initial migration.
+- name: make sure Openstack Python requirements get installed
+  set_fact:
+    edxapp_requirements_files: "{{ edxapp_requirements_files + [openstack_requirements_file] }}"
--- a/playbooks/roles/edxapp/meta/main.yml
+++ b/playbooks/roles/edxapp/meta/main.yml
@@ -8,3 +8,5 @@ dependencies:
     theme_users:
       - "{{ edxapp_user }}"
     when: "{{ EDXAPP_ENABLE_COMPREHENSIVE_THEMING }}"
+  - role: openstack
+    when: "{{ EDXAPP_USE_SWIFT_STORAGE }}"


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=3349

playbooks/roles/openstack/tasks/main.yml
```diff
@@ -50,17 +50,6 @@
...
-  command: >
-    {{ edxapp_venv_dir }}/bin/pip install {{ COMMON_PIP_VERBOSITY }} -i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w -r {{ openstack_requirements_file }}
-    chdir={{ edxapp_code_dir }}
-  sudo_user: "{{ edxapp_user }}"
-  environment: "{{ edxapp_environment }}"
-  when: edxapp_code_dir is defined
-  tags:
-    - install
-    - install:app-requirements
+- name: make sure Openstack Python requirements get installed
```
itsjeyd, line 55: @haikuginger Nit: Add a comment explaining why this task does not take care of installing requirements itself, perhaps?

itsjeyd:
:+1:
- [x] I tested this: Works as advertised in PR description (I was able to add a forum post containing an image, and to download CSV reports from the instructor dashboard)
- [x] I read through the code
- [x] Includes documentation (in the form of comments)

openedx-webhooks:
Thanks for the pull request, @haikuginger! I've created [OSPR-1409](https://openedx.atlassian.net/browse/OSPR-1409) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

gsong:
@haikuginger FYI: Looks like devops will be picking this up in early November.

playbooks/edx_sandbox.yml
```diff
@@ -40,6 +40,8 @@
     - role: mongo
       when: "'localhost' in EDXAPP_MONGO_HOSTS"
     - { role: 'rabbitmq', rabbitmq_ip: '127.0.0.1' }
+    - role: openstack
+      when: EDXAPP_SETTINGS == 'openstack'
```
e0d: I think it would be surprising to override the settings file and influence this behavior, though it's terse, I think explicit would be better.

playbooks/roles/openstack/tasks/main.yml
```diff
@@ -50,17 +50,9 @@
...
-  command: >
-    {{ edxapp_venv_dir }}/bin/pip install {{ COMMON_PIP_VERBOSITY }} -i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w -r {{ openstack_requirements_file }}
-    chdir={{ edxapp_code_dir }}
-  sudo_user: "{{ edxapp_user }}"
-  environment: "{{ edxapp_environment }}"
-  when: edxapp_code_dir is defined
-  tags:
-    - install
-    - install:app-requirements
+
```
e0d: So does this currently depend on proper ordering of role includes in the play?  If so, that seems a little brittle.  Would meta including this into edxapp, but conditionalizing it be better?  I don't love conditionals sprinkled everywhere, but both ensure that this ran before the edxapp role and that the referenced vars were in scope.

e0d (commented):
First pass with questions.

playbooks/roles/openstack/tasks/main.yml
haikuginger: So, the trouble is that right now, there's a task that runs during the EdX database migration that relies on datastorage being present. If that's the case, and we're using OpenStack SWIFT storage, then the initial migration will fail if the openstack requirements aren't already installed.

We could certainly do something like we currently do for [private requirements](https://github.com/edx/configuration/blob/b10da53da28358ee82f56307ec30eaf9c329e292/playbooks/roles/edxapp/tasks/deploy.yml#L181) if it's OK to have the OpenStack requirements inside the edxapp role.

e0d: I understand the ordering issue, but a meta include would ensure that this was run before the edxapp role, so there are implementation alternatives.

I realize also that we have some screwed up migrations, can you point to the particularly problematic migration?  It may be that the right thing to do here is fix it.

haikuginger: @e0d, I believe this is the problematic migration. I'm running a provision without this fix to double-check:

https://github.com/edx/edx-platform/blob/6d2468d0dcf782a7561ecc61946146d6fae2762f/lms/djangoapps/certificates/migrations/0003_data__default_modes.py#L24

I'll also do some digging to see if there's an existing Ansible variable that would be better to use than `EDXAPP_SETTINGS`.

haikuginger: Found it; this is the migration that's throwing an error for me at the moment, although I suspect that others would likely come up if we were to fix it:

https://github.com/edx/django-wiki/blob/edx_release/wiki/migrations/0001_initial.py

It's not clear whether modifying this migration to prevent the issue from happening would actually solve the problem, as there are plenty of other locations that use FileFields. So, I think we should do as you suggest, and meta-include this role in edxapp.

Or, better yet, since this is a single requirement, we could honestly probably just put it in one of the main requirements files and be just fine; it wouldn't be universally used, but it'd hardly hurt anything to have it.

haikuginger:
@e0d, I've updated this per discussion. Note that I've added a new variable, `EDXAPP_USE_SWIFT_STORAGE`, as no existing variable seemed appropriate.

e0d:
@haikuginger can you please rebase to see if it fixed the license error from VersionEye.

Also, would one enable swift storage separately from enabling openstack?

e0d:
@haikuginger ping.

haikuginger:
@e0d, sorry for the delay. Rebasing now.

In regards to "enabling Swift storage" vs. enabling OpenStack, I believe that Swift storage is the definitive feature of enabling OpenStack.

haikuginger:
Rebased from 9132ae4.

playbooks/roles/openstack/tasks/main.yml
e0d: @haikuginger what ended up happening here?

haikuginger: By doing the meta-include, we bypassed the order fragility. Based on the number of places we use FileFields in our models, I determined that an actual fix would be much more in depth than fixing this one migration, and so doing the meta-include solves the problem and also brings the OpenStack role more in line with existing conventions.

e0d (approved):
I'm happy with this, @feanil is going to have a quick look.

feanil:
:+1:  Looks good to me.

haikuginger:
Rebased and squashed from 3942ade.

jibsheet:
This is failing in our CI.

UndefinedError: 'EDXAPP_USE_SWIFT_STORAGE' is undefined

Where is that variable defaulted to false?  I'll revert this before bed tonight so I don't have to read failure mails all weekend.

haikuginger:
@jibsheet, sorry for the snafu. I tested provisioning it with the variable in question set, but the green checkmark gave me a false sense of security, and I didn't think to test without it. I've added a commit to add that default in #3464.

jibsheet:
@haikuginger Thanks for the quick fix - in the future, please don't merge on a Friday night, especially not in the middle of a holiday weekend.  This would have been much easier to triage in the office.

jibsheet:
Ah, and to explain the green check mark.  I suspect it's because the docker build on travis only runs

```
ansible-playbook edxapp.yml -t "install:base,install:configuration,install:app-requirements,install:code"
```

The openstack role is untagged, so Ansible may have totally trimmed it out of the call tree.  I would have to test further to confirm that theory.

However - you may wish to add standard tagging to that role.

haikuginger:
@jibsheet, noted! I'll take a look at the tagging process to see what we can do to make the testing fail correctly in future.

jibsheet:
@haikuginger This completely breaks our builds from scratch on AWS.
I will now be reverting it.

haikuginger:
@jibsheet, I'm sorry to hear that! I'm unclear on why this (in conjunction with #3464) would affect AWS builds; can you provide any additional detail as to what's going on so that we can correct the issue?

PR #3350, opened 2016-08-26
Pin the demo course whenever we install it

nedbat:
The demo course was unpinned, which caused problems when the base devstack box and the devstack installation straddled a change to the demo course.  We need to make a change to how the demo course creates users so that devstack installs on master will work, but this change will make Eucalyptus work.


nedbat:
@feanil demo course creates the users, but only if the checkout of demo course changed.  So we have to make sure the provisioning step during "release" gets the same version of the course as the "base" step got.

Not sure what the better fix will be.

nedbat:
@edx/devops not sure who is around today to look at this

feanil:
:+1:

PR #3352, opened 2016-08-26
Switch to manage_user from create_user

jibsheet:
manage_user doesn't try to create a new user, but it doesn't handle
enrollments, so we have to enroll the users in a separate task using
another management command.

Additionally, it takes only password hashes, not passwords, so we have
to provide a pre-salted and pre-generated database string.

### DIFF
--- a/playbooks/roles/demo/defaults/main.yml
+++ b/playbooks/roles/demo/defaults/main.yml
@@ -18,16 +18,20 @@ demo_code_dir: "{{ demo_app_dir }}/edx-demo-course"
 demo_repo: "https://{{ COMMON_GIT_MIRROR }}/edx/edx-demo-course.git"
 demo_course_id: 'course-v1:edX+DemoX+Demo_Course'
 demo_version: "master"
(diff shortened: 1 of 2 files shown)


feanil:
:+1:   @nedbat FYI

nedbat:
:+1: Thanks! :)

benpatterson:
When verifying, please also make sure that the users are enrolled in the same courses as previously (e.g., the performance demo course). Thx for coordinating a fix.

jibsheet:
@benpatterson that's handled in a separate task in this PR.  Turns out the testcourses/demo roles continue to be bizarrely intermingled.

@cpennington FYI since your change exposed the failure mode.

benpatterson:
Ok thanks for making sure it's covered @jibsheet

jibsheet:
@feanil @fredsmith please check updates since initial branch had trouble passing passwords on the command line and a few other failures around usernames.

feanil:
:+1:

PR #3355, opened 2016-08-29
Copy XBlock configuraiton files to the remote server

jibsheet:
This is simpler than doing the scp ourselves and passing a remote path
to ansible-playbook.  It will also simplify any future jenkins jobs we
may end up with.

@edx/devops please review
@cahrens FYI since you wrote the playbook

### DIFF
--- a/playbooks/edx-east/populate_configuration_model.yml
+++ b/playbooks/edx-east/populate_configuration_model.yml
@@ -37,10 +37,22 @@
     python_path: /edx/bin/python.edxapp
     manage_path: /edx/bin/manage.edxapp
   tasks:
+    - name: Create a temp directory
+      shell: mktemp -d /tmp/ansible_xblock_config.XXXXX
+      register: xblock_config_temp_directory
+    - name: Copy config file to remote server
+      copy:
+        src: "{{ file }}"
+        dest: "{{ xblock_config_temp_directory.stdout }}/{{ file | basename }}"
+      register: xblock_config_file
     - name: Manage xblock configurations
       shell: >
         {{ python_path }} {{ manage_path }} lms --settings=aws
-        populate_model -f {{ file | quote }} -u {{ user }}
+        populate_model -f {{ xblock_config_file.dest | quote }} -u {{ user }}
       register: command_result
       changed_when: "'Import complete, 0 new entries created' not in command_result.stdout"
     - debug: msg="{{ command_result.stdout }}"
+    - name: Clean up tempdir
+      file:
+        path: "{{ xblock_config_temp_directory.stdout }}"
+        state: absent


playbooks/edx-east/populate_configuration_model.yml
```diff
@@ -37,10 +37,22 @@
     python_path: /edx/bin/python.edxapp
     manage_path: /edx/bin/manage.edxapp
   tasks:
+    - name: Create a temp directory
+      shell: mktemp -d /tmp/ansible_xblock_config.XXXXX
+      register: xblock_config_temp_directory
+    - name: Copy config file to remote server
+      copy:
+        src: "{{ file }}"
```
cahrens, line 45: Can you update the usage instructions above (see line 9)? I'm not sure how this should be called now.

jibsheet: Ah, you call it the same way, it just takes care of copying the local file to the server for you, so you can say -e"file=/Users/kevin/git/edx/xblock-configuration/foo.json" and Ansible copies foo.json into a directory on the remote server and runs the management command using that path.

This avoids me needing to scp the three files up and get the paths correct when passing them to ansible.

cahrens: OK-- I've never run it against a remote server, so I wasn't aware of that detail. I've only ssh'd into a sandbox, installed the files there, and then run it directly (per https://openedx.atlassian.net/wiki/display/TNL/Adding+a+New+XBlock+on+Production#AddingaNewXBlockonProduction-Sandboxes).

That will still work, correct?

jibsheet: Yep - I didn't know those existed, but I just tested them on my own sandbox.
Because you tell ansible to use a local connection in those commands, it just copies the file around on the sandbox in the same way it would copy files from my laptop to stage when running the playbook (or in the future from jenkins to stage/prod).

cahrens: Thanks for testing it out!

playbooks/edx-east/populate_configuration_model.yml
```diff
@@ -37,10 +37,22 @@
     python_path: /edx/bin/python.edxapp
     manage_path: /edx/bin/manage.edxapp
   tasks:
+    - name: Create a temp directory
+      shell: mktemp -d /tmp/ansible_xblock_config.XXXXX
```
e0d, line 41: {{ ansible_date_time.epoch }}  rather than XXXXX maybe.

e0d:
One quibble, but :+1:

playbooks/edx-east/populate_configuration_model.yml
jibsheet: As noted in person, this is mktemp's template for "Pick me a 5 character random string and ensure uniqueness".

PR #3356, opened 2016-08-29
We moved to using the same xqwatcher as everything else

feanil:
@edx/devops 

We moved to using the standard supervisorctl but forgot to update this config here for the restart.


jibsheet:
👍

PR #3358, opened 2016-08-30
add symlink to jarsigner for cleaner scripting

estute:
@benpatterson @jibsheet

### DIFF
--- a/playbooks/roles/oraclejdk/tasks/main.yml
+++ b/playbooks/roles/oraclejdk/tasks/main.yml
@@ -52,6 +52,7 @@
     - java
     - javac
     - javaws
+    - jarsigner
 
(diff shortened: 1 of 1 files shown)


benpatterson:
@estute `jarsigner` would also point to the java implementation? Why is this necessary?

No objections at this point; just want to understand.

estute:
@benpatterson this is just to make scripting a little cleaner/reliable (a.k.a don't need to use a full, versioned, path)

benpatterson:
Thanks @estute 

It would be great if you updated the commit message to explain a bit more. Then merge when ready as far as I'm concerned.

👍

jibsheet:
👍

PR #3359, opened 2016-08-30
add default configuration for neo4j

adampalay:
@fredsmith , here's the default configuration for neo4j. I'll file a devops ticket to add the real configuration to secure.

@Qubad786 

associated platform PR: https://github.com/edx/edx-platform/pull/13103
## Configuration Pull Request

  - [x] Open a ticket ([DEVOPS-4701](https://openedx.atlassian.net/browse/DEVOPS-4701)) to make sure that they have been added to secure vars.

@fredsmith , I looked into adding to the changelog, but it doesn't look like the changelog has any organizational rule. There are many entries for each given role, and it's unclear if new entries should go on the top or bottom of the file. In fact, people seem to have done both: https://github.com/edx/configuration/blame/adam%2Fneo4j-config/CHANGELOG.md. What's the right move?

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -187,6 +187,16 @@ EDXAPP_ENABLE_READING_FROM_MULTIPLE_HISTORY_TABLES: True
 EDXAPP_GIT_REPO_DIR: '/edx/var/edxapp/course_repos'
 EDXAPP_GIT_REPO_EXPORT_DIR: '/edx/var/edxapp/export_course_repos'
 
+EDXAPP_NEO4J_CONFIG: !!null
+  # Setting this to null as a default. Here's what an example configuration
+  # would look like:
+  # bolt: True
+  # password: 'password'
+  # user: 'neo4j'
+  # https_port: 7473
+  # host: 'localhost'
+  # secure: True
+
 EDXAPP_ONLOAD_BEACON_SAMPLE_RATE: 0.0
 
 EDXAPP_FINANCIAL_REPORTS:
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -187,6 +187,14 @@ EDXAPP_ENABLE_READING_FROM_MULTIPLE_HISTORY_TABLES: True
 EDXAPP_GIT_REPO_DIR: '/edx/var/edxapp/course_repos'
 EDXAPP_GIT_REPO_EXPORT_DIR: '/edx/var/edxapp/export_course_repos'
 
+EDXAPP_NEO4J_CONFIG:
```
feanil: Is there a feature flag that enables/disables NEO4J on the edxapp side?  What happens if neo4j isn't actually setup on localhost?

fredsmith:
Unless this is going to work with localhost, neo4j/password on sandboxes or devstacks, we should set a null configuration in the ansible defaults and put in a comment for how to configure it.

playbooks/roles/edxapp/defaults/main.yml
adampalay: @feanil there isn't a feature flag because right now it's just a management command. Right now, if neo4j isn't set up, the management command will throw an error

adampalay:
@fredsmith , when you say:

> we should set a null configuration in the ansible defaults and put in a comment for how to configure it.

I don't know what that means. What's a null configuration? Are we asking users to go into their devstack's lms.auth.json and configure it manually?

adampalay:
@fredsmith updated the config to default to `!!null`. The code on the LMS-side is already making sure that there's a configuration: https://github.com/edx/edx-platform/pull/13103/files#diff-96e37d56c53da641c0f356efc9c13b2cR160

fredsmith:
👍  when tests pass

adampalay:
and it's ok that I don't have a changelog entry?

@Qubad786 please review

fredsmith:
@adampalay I don't think this needs a changelog entry.

Qubad786:
@adampalay will everyone be configuring it with their own configs? (by overriding this null config?) Or there will be standard configuration documented somewhere?

adampalay:
@Qubad786 we'll all be configuring our own configs if we install it locally

Qubad786:
Looks great 👍

PR #3360, opened 2016-08-31
Add task to take tags from the instance and put them on the AMI

macdiesel:
@edx/pipeline-team This is part of the work to get the tags off an instance. TE-1497.

PTAL.


feanil:
:+1:

PR #3362, opened 2016-09-01
Don't install into a subdirectory

nedbat:
Now the install happens in the current directory, with logs in a 'logs'
subdirectory.

Other improvements: curl now fails if the Vagrantfile can't be
downloaded, because the RELEASE value is invalid.  Curl also doesn't
print as large a progress bar.  The script fails early if there's no
RELEASE specified.

### DIFF
--- a/util/install/install_stack.sh
+++ b/util/install/install_stack.sh
@@ -118,6 +118,11 @@ else
     release=$OPENEDX_RELEASE
 fi
 
+if [[ ! $release ]]; then
+    echo "You must specify RELEASE, or define OPENEDX_RELEASE before running."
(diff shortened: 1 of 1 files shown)


nedbat:
@gsong @edx/devops Opinions?
cc: @cptvitamin

gsong:
@nedbat Definite improvement. 👍

nedbat:
@gsong I'll take that as an approval to merge.  We usually use :+1: as an unambiguous signal.

gsong:
@nedbat I edited my comment above. I guess GitHub doesn't send out update notifications. 👍

PR #3366, opened 2016-09-06
edxapp migrations to run are now seperated in to file per database

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/playbooks/continuous_delivery/run_migrations.yml
+++ b/playbooks/continuous_delivery/run_migrations.yml
@@ -55,7 +55,7 @@
     when: APPLICATION_NAME != "edxapp"
 
   - name: generate list of edxapp unapplied migrations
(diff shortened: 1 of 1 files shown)


doctoryes:
These changes make total sense and are correct. The first actual migration would have consumed the second unapplied migrations' output. 👍

feanil:
:+1:

playbooks/continuous_delivery/run_migrations.yml
```diff
@@ -79,8 +83,7 @@
       fail_on_missing: True
       mode: 0700
     with_items:
-      - "*{{ unapplied_migrations_output }}"
-      - "{{ migration_output }}"
+      - "{{ migration_files.stdout }}"
```
doctoryes: Perhaps this won't matter soon when migrations are run together, but it'd be nice to transfer the unapplied migrations to the artifacts dir _before_ running migrations. That way, even if migrations fail, the unapplied migrations will be available to view for debugging purposes.

macdiesel: Any task that fails within a job will cause the job to fail and artifact are not uploaded. Users will need to rely on the information in the console when a job fails. We should be sure to output any information needed to troubleshoot a defect to stderr/stdout.

doctoryes: 👍

PR #3367, opened 2016-09-07
Allow ecommerce gunicorn workers to be regularly restarted

rlucioni:
Leverages gunicorn's `max-requests` option to restart workers after they've processed a configured number of requests. ECOM-5448.

This PR adds a new default value, `ECOMMERCE_GUNICORN_MAX_REQUESTS`, that needs to be configured once this is merged.

@e0d @jibsheet @edx/ecommerce


clintonb:
👍

jibsheet:
👍 If it's straightforward, I'd love to see it added to the cookiecutter example role (I forget how much of that is in this repo vs the other one).

clintonb:
See https://github.com/edx/configuration/tree/master/playbooks/roles/ansible-role-django-ida for the template. Are there plans to roll this out to other IDAs?

rlucioni:
Thanks @clintonb. I don't think there are plans to roll this out elsewhere, but adding it to the cookiecutter role would make it more readily available to future services. @jibsheet do we really want to make using this option easier? It seems like something that should require a conversation/PR to use.

jibsheet:
@rlucioni We'd still need the PR that sets the default in the secure configuration, and presumably that would come from a ticket where the conversation happened.  If you think folks might just change defaults/main.yml and want to make that harder, I guess it's fine?

rlucioni:
@jibsheet fair. Added.

jibsheet:
👍  Thanks.  Feel free to merge when Travis finishes up.

PR #3374, opened 2016-09-13
A script for making box files smaller

nedbat:
This is to be run in the guest OS before "vagrant package". I didn't
symlink it to /edx/bin because it isn't meant for adopters to be using
after installation.

### DIFF
--- a/playbooks/roles/edx_ansible/tasks/deploy.yml
+++ b/playbooks/roles/edx_ansible/tasks/deploy.yml
@@ -53,6 +53,7 @@
     mode: "0755"
   with_items:
     - { src: 'show-repo-heads.j2', dest: 'show-repo-heads' }
+    - { src: 'pre-box.j2', dest: 'pre-box' }
   tags:
     - install
     - install:configuration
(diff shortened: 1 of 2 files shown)


nedbat:
@edx/devops This reduces a box file by 25-30%. Tweaks welcome.

maxrothman:
Have you tested this? The last 2 lines make me nervous.

nedbat:
@maxrothman I have tested it.  The dd command will run and error out when the disk is full.  Then the /empty file will be deleted.  I also have experience with doing it wrong (with && between the two commands so that the file is never deleted).

Is there something I can do to assuage your fears?

maxrothman:
If bash has -e set (via the environment or something) the second command will never be run. You might want to add an explicit `||` to fix that.

How long does it take to run? Does doing so stress the host machine?

nedbat:
It takes about four minutes to run. VBoxHeadless on the host is at 200% cpu (a quarter of the laptop). If I weren't looking at Activity Monitor, I wouldn't have noticed.

I've added `|| true` to the failing command.

jibsheet:
👍

PR #3378, opened 2016-09-13
add libraries/packages for android emulation and more tests

estute:
## Configuration Pull Request

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/android_testing_tools.yml
+++ b/playbooks/roles/jenkins_worker/tasks/android_testing_tools.yml
@@ -1,0 +1,16 @@
+---
+
+# Install Python packages used for testing the edX Android app on Jenkins
+# Note: normally, we use a virtual environment for pip installs on a Jenkins 
+# worker, but since this is the only instance of python being used on the machine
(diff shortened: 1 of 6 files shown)


estute:
@jzoldak @jibsheet

jzoldak:
LGTM 👍

playbooks/roles/jenkins_worker/tasks/android_testing_tools.yml
```diff
@@ -0,0 +1,16 @@
+---
+
+# Install Python packages used for testing the edX Android app on Jenkins
+# Note: normally, we use a virtual environment for pip installs on a Jenkins 
+# worker, but since this is the only instance of python being used on the machine
+# we install them as system packages. If we start requiring more python, or
+# add other jobs to this worker, please reconsider this.
+- name: Install Android testing requirements using pip
```
jibsheet, line 8: Is using a virtualenv significantly more complicated?  Normally we default to everything in a virtualenv in our jenkins, although I understand the specifics of this work.

jibsheet (approved):
One small question about "if virtualenv is easy, can we just do it?" but I understand if you want to avoid complexity for now.

PR #3381, opened 2016-09-15
Added CD playbook to update search index

clintonb:
ECOM-5134

### DIFF
--- a/playbooks/continuous_delivery/haystack_update_index.yml
+++ b/playbooks/continuous_delivery/haystack_update_index.yml
@@ -1,0 +1,40 @@
+# This playbook will run the update_index management command. It is intended solely for usage by services using
+# Haystack.
+#
+# Required variables for this playbook:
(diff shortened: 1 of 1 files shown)


clintonb:
@edx/pipeline-team please review

playbooks/continuous_delivery/update_index.yml
```diff
@@ -0,0 +1,40 @@
+# This playbook will run the update_index management command. It is intended solely for usage by services using
```
macdiesel: Since this is only for services using haystack could you change the name of this file to reflect this fact?

macdiesel (changes requested):
One nit, otherwise 👍

playbooks/continuous_delivery/update_index.yml
clintonb: Done.

PR #3388, opened 2016-09-19
Extra optional identifier for CD AMI names

macdiesel:
@edx/pipeline-team PTAL.

### DIFF
--- a/playbooks/continuous_delivery/create_ami.yml
+++ b/playbooks/continuous_delivery/create_ami.yml
@@ -26,6 +26,7 @@
 #   - hipchat_token                   - API token to send messages to hipchat
 #   - hipchat_room                    - ID or name of the room to send the notification
 #   - hipchat_url                     - URL of the hipchat API  (defaults to v1 of the api)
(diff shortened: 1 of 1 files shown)


doctoryes (commented):
So - this change is for the case when you want to create an AMI with the same `"{{ edx_environment }} -- {{ deployment }} -- {{ play }} -- {{ app_version }}"` and yet have a different AMI ID?

macdiesel:
@doctoryes we accomplish 2 things via this PR:
- Each run of a B/M/D pipeline will always generate an AMI artifact. Prior to this change an AMI would only be created if the sha of the app repository was different from the last run. 
- If changes to other repositories are made this new AMI will capture those changes (configuration, config-secure, etc). 

On a really basic level, each run of a pipeline should generate a new AMI. We were not doing this before this change.

doctoryes:
Agreed - 👍 

Will we be using the pipeline run ID for the extra number?

macdiesel:
I modified the pattern to use the pipeline counter.

PR #3397, opened 2016-09-28
Make sure user has correct associated group

saleem-latif:
## Configuration Pull Request

Hi @mattdrayer , @jibsheet , @fredsmith 

Kindly review these changes, this PR fixes the issue of invalid groups for edx-themes user, issue was discussed in DEVOPS-4749.

**Problem Description:**
On sandbox we are facing an issue that whenever a new theme is added to edx-themes and is updated via `git pull`, the new directories will have a group `users` instead of `www-themes`. 

After digging around the problem, I found that `user` module of ansible is assigning a default group `users` to any newly created users that does not explicitly defines a group, which differs from previously observed default behavior. Previously, ansible used to assign a default group with name same as the new user's username.

Although the sandbox problem can be avoided by using the preferred workflow of updating themes i.e. using `/edx/bin/update` command instead of manually updating git repo. It would still be a good idea to assign `www-data` as the default group for `edx-themes` user on sandbox.

**Description of changes:**
In this PR, I have updated `add_user` to pass in (while calling ansible module `user`) a default group for the new user, this group passed in as `group_name` argument would sometimes have same value as `user_name` argument but it can have any other value e.g. for `edx-themes` user we use `www-data` as user's primary group.

- [x] Have completed [Ops ansible testing checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)

### DIFF
--- a/playbooks/roles/add_user/tasks/main.yml
+++ b/playbooks/roles/add_user/tasks/main.yml
@@ -34,6 +34,7 @@
   user:
     name: "{{ user_name }}"
     home: "{{ user_home }}"
+    group: "{{ group_name }}"
     createhome: yes
     shell: /bin/false
     generate_ssh_key: yes
--- a/playbooks/roles/edx_themes/tasks/main.yml
+++ b/playbooks/roles/edx_themes/tasks/main.yml
@@ -52,3 +52,9 @@
     append: yes
   with_items: theme_users
   when: theme_users is defined
+
+- name: update .bashrc to set umask value
+  lineinfile:
+    dest: "{{ themes_home }}/.bashrc"
+    line: "umask 002"
+    state: present


playbooks/roles/add_user/tasks/main.yml
```diff
@@ -34,6 +34,10 @@
   user:
     name: "{{ user_name }}"
     home: "{{ user_home }}"
+    append: yes
```
mattdrayer: I know we don't have an automated test harness specifically for these configuration changes, so is there additional information/context that should be included in the commentary to describe this update?

mattdrayer:
@jibsheet @fredsmith -- just FYI that more information has been added to the description of this PR

saleem-latif:
@mattdrayer , @fredsmith, @e0d  currently I am facing an issue while testing this PR, after the sandbox has been created, if we login as `edx-themes` user and create directory inside `/edx/var/edx-themes` directory it will have `0755` as default permissions but I want any new directory created by `edx-themes` to have `0775` permissions.
The problem arises if one adds a new theme and pulls the changes via `git pull` then newly created theme will have `0755` permission set and subsequent asset compilation errors out.
Any thoughts on how to resolve this problem ? I have tried many solutions but none seems to be working so far.

jibsheet:
@saleem-latif You want to set the UMASK for the user (you can google umask to learn more about it).  Right now, users have 022, you want 002.  Ansible's user module should let you set this up in the profile for the user, but I don't have an example at hand.

Once umask is set, files will default to 0775 rather than 0755.

saleem-latif:
@jibsheet I could not find any ansible module that allows setting the value `umask`, so I tried [this](https://github.com/edx/configuration/pull/3397/files#diff-0b63e279148a5b9b2c6a59f2032ee932R48) approach but it seems to have no effect and `umask` still returns `0022` value.
Can you take a look at let me know what am I doing wrong ?

arbabnazar:
I didn't test it but you can use like this, create `.bashrc` file inside the user home directory with the `umask` content. You can use [lineinfile](http://docs.ansible.com/ansible/lineinfile_module.html). Hope that help you

jibsheet:
@saleem-latif as Arbab notes, you really need this to go into a profile or bashrc.  It's not something you can set as a one-off, it's an attribute of a user's shell environment.

saleem-latif:
@arbabnazar Thanks, the approach you suggested works perfect, 
@jibsheet , @fredsmith Kindly take a look at the changes.

jibsheet:
👍  @fredsmith Will you comment on .bashrc vs .profile. I think .bashrc will do what Saleem wants, but wanted someone else to think on it.

saleem-latif:
@fredsmith Any thoughts ? is this PR good for merge ?

fredsmith (approved):
Looks good - bashrc is the right place for this.

PR #3399, opened 2016-09-29
ECOM-5847 ( Adding email functionality in DISCOVERY IDA )

awais786:
ECOM-5847
Adding configuration details for email backend.
## Configuration Pull Request


awais786:
This is the [PR](https://github.com/edx/course-discovery/pull/360/files) for the discovery-IDA which contains django-ses in requirements and production.py var updation.

awais786:
@bderusha  @clintonb Please review this PR.

clintonb (approved):
Squash your commits.

clintonb:
Also, please provide a more informative commit message than the ticket number. This article has a good guideline: http://chris.beams.io/posts/git-commit/.

awais786:
Ticket for DEVOPS
https://openedx.atlassian.net/browse/DEVOPS-4866

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -101,7 +101,26 @@ DISCOVERY_MEDIA_STORAGE_BACKEND:
...
+# Eample config for smtp backend
+# DISCOVERY_EMAIL_CONFIG:
+#   EMAIL_BACKEND: 'django.core.mail.backends.smtp.EmailBackend'
+#   EMAIL_HOST: 'localhost'
+#   EMAIL_PORT: 25
+#   EMAIL_USE_TLS: False
+#   EMAIL_HOST_USER: ''
+#   EMAIL_HOST_PASSWORD: ''
+
+DISCOVERY_EMAIL_CONFIG:
```
bderusha: This is a non-standard approach to handling configuration and has caused us some pain in the past.  I am wondering why we don't just flatten this structure and load the normal way.

@awais786 @clintonb ?

clintonb: I prefer the flat approach; but, the decision was made to use the nested approach to make changes easier for OpenEdX users. I will defer to @jibsheet and @e0d regarding whether we should continue this approach.

jibsheet: I don't have a firm opinion, others do, we should just keep the easier to override version

Question - we use IAM roles, not AWS keys/secrets.  Should I set AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY to blank or leave them unset to cause @awais786 's code to work properly with an IAM role?

clintonb: @jibsheet what are we doing for ecommerce? That's the only other IDA with which I have SES experience. According to https://github.com/django-ses/django-ses/issues/80, leaving them unset may work.

jibsheet: @clintonb looks like we just don't specify anything in the config.  So, probably "leave them unset".  Just want to make sure Awais' code will do the right thing with the config if I leave them unset.

clintonb: Same package, so we should have the same behavior.

e0d: @clintonb there's very limited configuration in the ECOM role around email, so I don't see it as an exemplar for anything other than It "Just Works!"

This looks more relevant:

https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L104

It would be useful to confirm that a placeholder string in the access key and id don't trigger a different behavior than not defining them at all.

Jumping into a repl on a sanbox would allow this.

clintonb: @awais786 @bderusha I will leave it to you guys to test this and ensure these defaults will work. I suspect we may need to use `null` instead of `'SET-ME-PLEASE'`.

awais786: @e0d @clintonb In ecommerce backend is set in [production.py](https://github.com/edx/ecommerce/blob/master/ecommerce/settings/production.py#L19)
No other setting is provided in configuration. 

I am thinking to do the same for discovery. 
Defining the following settings in production.py only and no need for this PR.

# Email configuration

```
EMAIL_BACKEND = 'django_ses.SESBackend'
AWS_SES_REGION_NAME = 'us-east-1'
AWS_SES_REGION_ENDPOINT = 'email.us-east-1.amazonaws.com'
```

bderusha: @awais786 I don't know if this is a right approach.  I would prefer @e0d's suggestion of implementing this like edxapp https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L104

awais786: If these keys AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY are not require then can we remove them ?

bderusha: They are required.  https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L43

awais786: So instead of set-me-please --> "None".

bderusha: And flatten please.  It doesn't seem like we need to nest these params within a  `DISCOVERY_EMAIL_CONFIG` parent var

bderusha (approved):
👍

awais786:
@e0d @jibsheet would you please review this again.

e0d:
:+1:

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -101,7 +101,24 @@ DISCOVERY_MEDIA_STORAGE_BACKEND:
   MEDIA_URL: '{{ DISCOVERY_MEDIA_URL }}'
   
 DISCOVERY_STATICFILES_STORAGE: 'django.contrib.staticfiles.storage.StaticFilesStorage'
-  
+
+
+# You can set different email backends with django:
+# https://docs.djangoproject.com/en/1.9/topics/email/#email-backends
+# Eample config for smtp backend
```
clintonb: This example is more confusing than helpful since the settings are backend-specific. Simply linking to the Django page should be sufficient.

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -101,7 +101,24 @@ DISCOVERY_MEDIA_STORAGE_BACKEND:
...
+# EMAIL_BACKEND: 'django.core.mail.backends.smtp.EmailBackend'
+# EMAIL_HOST: 'localhost'
+# EMAIL_PORT: 25
+# EMAIL_USE_TLS: False
+# EMAIL_HOST_USER: ''
+# EMAIL_HOST_PASSWORD: ''
+
+DISCOVERY_EMAIL_BACKEND: 'django_ses.SESBackend'
```
clintonb, line 107: Double quotes

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -101,7 +101,24 @@ DISCOVERY_MEDIA_STORAGE_BACKEND:
...
+DISCOVERY_AWS_ACCESS_KEY_ID: "None"
```
clintonb: This will produce the string `"None"` rather than `null`, which I believe is what is desired.

awais786: According to this  ['None' will be written out as null](https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L41)

awais786: @clintonb Please check my last comment.

clintonb: Confirmed this is correct with @e0d via HipChat.

clintonb:
👍

playbooks/roles/discovery/defaults/main.yml
jibsheet: @bderusha Now that these are flattened, how does a non-SES using community member deploying openedx pass variables for other backends in?  You'll note that edxapp exposes these variables, but they're just ignored in our configuration (because we use SES as the backend).

https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L105-L109

bderusha: @jibsheet That's a good point.

We have accounted for this in the private.py file https://github.com/edx/course-discovery/pull/360/files#diff-18d677de5a4aac33e7d83c977323c460R7

However that only handles the local dev case, not community config.  @awais786 please add the config vars to this per Kevin's link and your private.py file.

awais786: @bderusha @jibsheet @e0d Please review this PR.

bderusha (approved):

bderusha:
@jibsheet @e0d Can we get a final approval from one of you?

jibsheet (approved):

playbooks/roles/discovery/defaults/main.yml
awais786: @jibsheet @e0d Please review this PR.

awais786:
@e0d @jibsheet should i merge this PR now or wait for the DEVOPS ticket to get merge ?
https://openedx.atlassian.net/browse/DEVOPS-4866

PR #3400, opened 2016-09-29
turn off logging for ec2 facts

estute:
## Configuration Pull Request


jibsheet (approved):
Just curious which fact was too hot for the logs?
But is totally fine (although you will loathe this some day when debugging).

estute:
@jzoldak @benpatterson

jzoldak:
👍

estute:
For posterity's sake, the ec2_facts action logs aws credentials and such, which is not a good idea.

PR #3402, opened 2016-09-30
Script for assuming role by an automated process

jibsheet:
@edx/devops

@e0d this will let us remove the copied code from RunAnsible.groovy

### DIFF
--- a/util/jenkins/assume-role.sh
+++ b/util/jenkins/assume-role.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+# assume-role function for use by machine services that don't use MFA to assume role.
+# source this into your bash script and then
+# set +x
+# assume-role(${AWS_MFA_ARN})
+# set -x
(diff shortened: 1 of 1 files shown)


util/jenkins/assume-role.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+
+# assume-role function for use by machine services that don't use MFA to assume role.
+# source this into your bash script and then
+# set +x
+# assume-role(${AWS_MFA_ARN})
+# set -x
+# You can skip set +-x if you aren't concerned about leaking the tokens in your logs.
```
maxrothman: Should we instead put set +x in the function? This looks like a leak waiting to happen.

jibsheet: Good idea.  I left in the suggestion to set +x outside too since it'll prevent leaking your MFA ARN, but I like knowing that we're much less likely to leak tokens.

maxrothman:
👍

PR #3403, opened 2016-10-01
trivial change

e0d:
## Configuration Pull Request


jibsheet (approved):
As long as there isn't a tab on that line

PR #3404, opened 2016-10-04
Using bin_path in the apparmor template

e0d:
## Configuration Pull Request

OPS-1766


e0d:
ugh, too many literal indexes, but :+1:   after CI passes, I've restarted it.  It failed previously though the reason was not clear from the logs.

e0d:
:+1:

PR #3407, opened 2016-10-06
Update Ansible configuration location for edxmako

andy-armstrong:
This configuration change is in support of the following platform PR: https://github.com/edx/edx-platform/pull/13683

@nasthagiri @fredsmith please review.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Role: edxapp
+  - Renamed edxmako to openedx.core.djangoapps.edxmako
 - Role: common
   - Renamed `COMMON_AWS_SYNC` to `COMMON_OBJECT_STORE_LOG_SYNC`
   - Renamed `COMMON_AWS_SYNC_BUCKET` to `COMMON_OBJECT_STORE_LOG_SYNC_BUCKET`
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -966,14 +966,14 @@ lms_auth_config:
               OPTIONS:
                 default_class: 'xmodule.hidden_module.HiddenDescriptor'
                 fs_root:  "{{ edxapp_course_data_dir }}"
-                render_template: 'edxmako.shortcuts.render_to_string'
+                render_template: 'openedx.core.djangoapps.edxmako.shortcuts.render_to_string'
             - NAME: 'draft'
               ENGINE: 'xmodule.modulestore.mongo.DraftMongoModuleStore'
               DOC_STORE_CONFIG: "{{ EDXAPP_LMS_DRAFT_DOC_STORE_CONFIG }}"
               OPTIONS:
                 default_class: 'xmodule.hidden_module.HiddenDescriptor'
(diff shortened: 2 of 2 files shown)


nasthagiri:
Is there a BWC issue with this change?  Do we ever have a case where a newer version of the configuration changes use older version of edx-platform or vice versa?

andy-armstrong:
@nasthagiri Great question. If we build hotfixes with the latest configuration then that could cause problems. I believe that those AMIs are built using the configuration from the AMI being patched, in which case that should be fine.

For production releases, the configuration and code is baked into the same AMI, so I think that should be fine. Same thing for sandboxes.

One issue that is important is that this breaks devstack. I'll need to send out an email to edx-code and eng-all saying the configuration changes that need to be made locally. I don't know how a new devstack is built, as that might be a problem if eucalyptus developers pick up the production configuration. @nedbat, do you have any advice about this change?

Another option is that I could make edxapp have its own setup.py so that its namespace doesn't change. Maybe I should even bring it out into its own repo, although currently it has a lot of platform dependencies. @doctoryes, do you have any thoughts?

I think I'll put these changes on hold for now until we reach agreement on the right way to move forward.

nasthagiri:
@andy-armstrong Alternatively, you can create (temporary?) migration code that automatically does the conversion when needed.  This is what we did for the modulestore configuration update when Split was introduced: https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/modulestore/modulestore_settings.py#L9

doctoryes:
@andy-armstrong I think you're right - while AMIs and sandboxes would still work as long as you merged both this change and the accompanying edx-platform change at roughly the same time. But devstack would be broken since it's `{lms|cms}.auth.json` files would remain unchanged. @nasthagiri 's suggestion of adding conversion code would work.

I'm not clear on what you're suggesting about edxapp having it's own setup.py _or_ bringing it out into a separate repo. We can discuss it next week.

andy-armstrong:
Thanks @doctoryes. I was thinking about adding an explicit setup.py for edxmako and then pip installing it so that it can have its own namespace. This is how a number of the packages work in common/lib, for example this one:

https://github.com/edx/edx-platform/tree/master/common/lib/calc

The get installed like this:

https://github.com/edx/edx-platform/blob/master/requirements/edx/local.txt

I originally created a PR to move all of these libraries to openedx (https://github.com/edx/edx-platform/pull/13707), but I'm now thinking that they shouldn't live there because they don't install themselves into the openedx namespace. Maybe we should have a new location for libraries that live in edx-platform but install themselves -- perhaps `edx-platform/packages`?

The next logical thought is to move them into their own repos, but the challenge with edxmako is that it accesses edx-platform internals to implement some of its methods, e.g.

https://github.com/edx/edx-platform/blob/master/common/djangoapps/edxmako/shortcuts.py#L25

Let's work it out next week.

nedbat:
@andy-armstrong an alternative to asking everyone to hand-edit their configurations is to make the module also be importable as "edxmako.shortcuts".  You can do this by assigning a key in sys.modules.

andy-armstrong:
Closing as the proper solution doesn't require a configuration change.

PR #3412, opened 2016-10-12
Avoid using the same register twice.

smarnach:
This fixes a minor bug with "sshd_config" being used as a register twice, so the second usage overwrites the result of the first one.  It also removes the unused template for `sshd_config`.


openedx-webhooks:

To automatically create an OSPR issue for this pull request, just visit this link: https://openedx-webhooks.herokuapp.com/github/process_pr?number=3412&repo=edx%2Fconfiguration

smarnach:
@clintonb This should fix the issue you experienced.  I applied this patch to the branch `clintonb/ecomworker-docker` and tried building a docker image using

```
docker build -t ecomworker:latest -f docker/build/ecomworker/Dockerfile .
```

The `vhost` role now finishes successfully, but the build stiil errors out in the `aws` role when trying to set the MTU:

```
TASK: [aws | Set the MTU to 1500 temporarily] ********************************* 
failed: [127.0.0.1] => {"changed": true, "cmd": "/sbin/ifconfig eth0 mtu 1500 up", "delta": "0:00:00.023622", "end": "2016-10-12 14:53:41.029833", "rc": 255, "start": "2016-10-12 14:53:41.006211", "warnings": []}
stderr: SIOCSIFMTU: Operation not permitted
SIOCSIFFLAGS: Operation not permitted
```

However, that seems to be an unrelated problem.

openedx-webhooks:
Thanks for the pull request, @smarnach! I've created [OSPR-1504](https://openedx.atlassian.net/browse/OSPR-1504) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

clintonb:
Thanks @smarnach. We shouldn't be running the `aws` plays for Docker. I will need to update the `ecomworker` play with tags, similar to our other Docker-based IDAs and Dockerfiles.

clintonb (approved):
I have verified this resolves my issues, and I am now able to build an ecomworker Docker image.

clintonb:
@edx/devops please review.

e0d:
@clintonb @smarnach my question is why the ecom worker container in invoking vhost at all, it should be docker -> containter ->common, not aws -> vhost -> common.

clintonb:
@e0d this is a facepalm moment for me. You are correct that this is not the default pathway for Docker builds. Adding the appropriate tags to the `ecomworker` play resolves my issues.

@smarnach I apologize for the trouble. I do not believe these changes are necessary to resolve my Docker issues.

smarnach:
@clintonb I'm glad you found a solution for your problem.  I'm not very familiar with Docker in general and with how the Docker build in this repo should work, so I didn't know whether the `vhost` role was supposed to be included.

@e0d This PR also contains a minor bugfix and a clean-up.  I'll remove the unneeded bits, but the rest might be worth preserving.

playbooks/roles/vhost/templates/etc/ssh/sshd_config.j2
```diff
@@ -1,90 +0,0 @@
-# {{ ansible_managed }}
```
e0d, line 1: @smarnach this file is simply not used?

e0d (approved):
If the removed template was simply not used anywhere, which seems to be the case, I approve.

playbooks/roles/vhost/templates/etc/ssh/sshd_config.j2
smarnach: @e0d This file is unused since the only point where it was used was removed:

https://github.com/edx/configuration/commit/924a7a01ba2ad4322b4ca5dd08431d3fff00f4f2#diff-3c978bc80846b9f43e8b5e01e2df78adL35

PR #3414, opened 2016-10-12
WIP: E0d/xenial

e0d:
## Configuration Pull Request

@nedbat I've re-based this branch I had going a while ago.  With a few additional tweaks, I was able build edxapp docker container.

### DIFF
--- a/playbooks/roles/supervisor/tasks/main.yml
+++ b/playbooks/roles/supervisor/tasks/main.yml
@@ -115,13 +115,18 @@
     - install
     - install:base
 
+- debug: msg="dist-> {{ ansible_distribution_major_version|int }}"
+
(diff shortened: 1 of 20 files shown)


playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -137,7 +144,7 @@
     group: root
   when: >
     supervisor_service == "supervisor" and disable_edx_services and not devstack
-    and ansible_distribution == 'Ubuntu' and ansible_distribution_major_version|int < 16
+    and (ansible_distribution_release == 'precise' or ansible_distribution_release == 'trusty')
```
nedbat, line 147: I'm not up on best ansible practices, but can we encapsulate this logic centrally and then have a condition like `using_systemd` and `using_upstart`?  This seems error-prone.

PR #3416, opened 2016-10-13
Added Dockerfile for E-Commerce Worker service

clintonb:
ECOM-5862


docker/plays/ecomworker.yml
```diff
@@ -0,0 +1,17 @@
+- name: Deploy edX Ecommerce Worker
+  hosts: all
+  become: True
+  gather_facts: True
+  vars:
+    ENABLE_DATADOG: False
+    ENABLE_SPLUNKFORWARDER: False
+    ENABLE_NEWRELIC: False
+  roles:
+    - aws
```
e0d: Should be:

```
roles:
  - common_vars
  - docker
  - ecomworker
  ...
```

clintonb: Updated.

e0d:
:+1:

e0d (approved):
Looks good.

clintonb:
We will revisit this in the future if/when we move this into the Docker-based devstack.

PR #3421, opened 2016-10-14
instruct nginx to add a X-Queue-Start header

pwnage101:
## Configuration Pull Request

This header gets added to requests proxied by nginx to various edx apps.
The New Relic agent, if there is one, reads this header to calculate the
time each request spends between the web server (nginx) and the
application server (django or some other cms).

The specific intention of this commit is to measure the time requests
spend queued at the gunicorn layer.

This work is part of PERF-372.
## Testing

I created a NR-enabled sandbox using this branch, and drove LMS load using locust.  By varying the number of locust clients, I was able to fill the gunicorn request queue:

[2016-10-14-152240_1040x396_scrot](https://rpm.newrelic.com/accounts/1391042/notes/6311)

In particular, pay attention to the new layer "Request Queuing", and how it suddenly increases (then plateaus) when the one gunicorn worker cannot handle the increased simultaneous requests without queuing.


jibsheet (approved):
Too bad, you're like a week away from us finishing up a refactor of these templates so there's only 1 or 2 rather than 12.

This change looks great.

PR #3422, opened 2016-10-14
Jibsheet/disable automated upgrade 16.04

jibsheet:
Two parts.

Run systemctl disable-timers to turn off running apt-get update/upgrade on boot and at 0600, 1800.
http://www.hiroom2.com/2016/05/18/ubuntu-16-04-auto-apt-update-and-apt-upgrade/

This is similar to what we do earlier in this file for 12.04 and 14.04

Additional bugfix.  It was impossible to build xqwatcher from an existing xqwatcher AMI because it tried to make the virtualenv every time.

I've tested this by building
xqwatcher from scratcher
xqwatcher from that AMI

edxapp from scratch and from existing (12.04)
discovery from scratch and from existing (14.04)

@edx/devops

### DIFF
--- a/playbooks/roles/security/tasks/security-ubuntu.yml
+++ b/playbooks/roles/security/tasks/security-ubuntu.yml
@@ -22,6 +22,13 @@
     mode: "0644"
   when: SECURITY_UNATTENDED_UPGRADES
 
+- name: Disable unattended-upgrades if Xenial (16.04)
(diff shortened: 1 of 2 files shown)


maxrothman (approved):

maxrothman:
You should look into the failing test, but this looks good.

PR #3426, opened 2016-10-17
Reorder AMI names so they sort better in the console

macdiesel:
@maxrothman @jibsheet  PTAL. Better sorting of AMIs generated by pipelines in the AWS console.


maxrothman:
👍

PR #3430, opened 2016-10-24
Adding PUBLISHER_FROM_EMAIL for sending emails.

awais786:
## Configuration Pull Request

ECOM-6141


awais786:
This is the email address `publisher-notification@edx.org`

awais786:
Ticket for DEVOPS
https://openedx.atlassian.net/browse/DEVOPS-4975

awais786:
@edx/ecommerce Please review this PR.

bderusha (approved):

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -119,6 +119,8 @@ DISCOVERY_EMAIL_USE_TLS: False
 DISCOVERY_EMAIL_HOST_USER: ''
 DISCOVERY_EMAIL_HOST_PASSWORD: ''
 
+DISCOVERY_PUBLISHER_FROM_EMAIL: 'publisher@example.com'
```
clintonb: Can we set this to null so that no emails are sent if the value is not properly defined?

clintonb (changes requested):
We should avoid fake data that seems real, so it's obvious there are errors and the system fails in a manner that we can find those errors during smoke testing.

fredsmith (approved):

awais786:
@fredsmith should I merge this PR or wait for the DEVOPS ticket to get complete ?

awais786:
@clintonb Please review this PR again.

clintonb (approved):

PR #3432, opened 2016-10-25
install pysqlite for devstack and jenkins workers

estute:
## Configuration Pull Request

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -1156,6 +1156,8 @@ edxapp_debian_pkgs:
   - libfreetype6-dev
   # cffi needs libffi-dev
   - libffi-dev
+  - python-dev
+  - libsqlite3-dev
 
 # Deploy Specific Vars
 edxapp_lms_variant: lms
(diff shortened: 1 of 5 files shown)


estute:
@jz @cpennington

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -151,6 +151,10 @@
     - install
     - install:app-requirements
 
+# Override default dependencies for installing on a devstack
+- include_vars: ../vars/devstack.yml
+  when: devstack == true
```
cpennington: This is the one part of this PR that seems questionable to me. It seems like we shouldn't be adding explicit references to devstack specific things into the generic deploy role (and should instead be adding the devstack specific things to the devstack role or to code that actually builds the devstack, as an override).

estute:
adding @jibsheet for a devops opinion

estute:
This change is related to https://github.com/edx/edx-platform/pull/13826

cpennington (approved):
It's too bad that YAML doesn't have a way to append to lists, so that we have to duplicate the full set of requirements files.

jibsheet (approved):
Per some offline conversation with @estute - as long as we can still cut edx-platform AMIs (slightly different build process ) then this is unfortunate but fine.  I did a quick bit of digging on ansible workarounds, but they're all dictionary based, not list based.

estute:
AMIs cut successfully.

PR #3433, opened 2016-10-25
remove updates available check from MOTD

fredsmith:


jibsheet (approved):
Assuming we've cut/booted an AMI and made sure we can still log in, 👍

PR #3434, opened 2016-10-26
Add a new check which allows checking for extra config

jibsheet:
@edx/devops Please have a look

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -1,4 +1,13 @@
 ---
+- name: Check Configuration Sources
+  fail:
+    msg: "Configuration Sources Checking (COMMON_EXTRA_CONFIGURATION_SOURCES_CHECKING) is enabled, you must define {{ item }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/common/tasks/main.yml
```diff
@@ -1,4 +1,13 @@
 ---
+- name: Check Configuration Sources
+  fail:
+    msg: "Configuration Sources Checking is enabled, you must defined {{ item }}"
```
maxrothman: typo: "defined"

maxrothman: This error message might appear cryptic to someone who didn't know much about the config sources checking system. I'd say the comment in defaults should be duplicated here, or you should have a pointer to it, or something.

maxrothman (commented):
I'll give this a I-can't-believe-this-works-but-I-guess-we'll-ship-it-anyways thumb on general functionality 😞 👍

jibsheet:
@maxrothman Pushed a documentation update, let me know if it makes sense and I'll squash so I can merge and move forward with updating all the private repos.

maxrothman:
👍

PR #3435, opened 2016-10-27
Use unattended-upgrade to only take security updates

jibsheet:
This allows for security updates to be applied during artifact creation, but not applying other updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -192,3 +192,7 @@
 
 - Role: discovery
   - Added `PUBLISHER_FROM_EMAIL` for sending emails to publisher app users.
+
+- Role: security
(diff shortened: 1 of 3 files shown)


jibsheet:
@edx/devops Please Review - https://openedx.atlassian.net/browse/OPS-1853

maxrothman:
Why the dry run first?

jibsheet:
@maxrothman the dry-run prints out things, unattended-upgrade is much quieter about what it's doing (no verbose flag, just debug which is OMG).  I also figured if we had an error in this script we'd see "I want to upgrade these 7" and be able to reason more about why the actual upgrade application failed.

maxrothman (approved):
No verbose flag? Ew.

At least I'll forget about this in a week.

On second thought, can you add a comment so I don't have to ask you the same question in a week?

PR #3440, opened 2016-10-28
add security upgrade flag to te jenkins plays

estute:
---

i know that some of these are ceremonial, in the sense that they are not actually what master is created from, but just wanted to keep current so we can get there some day.

### DIFF
--- a/playbooks/edx-east/jenkins_admin.yml
+++ b/playbooks/edx-east/jenkins_admin.yml
@@ -5,6 +5,8 @@
   gather_facts: True
   vars:
     serial_count: 1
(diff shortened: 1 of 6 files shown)


benpatterson:
👍 for me although I should point out that jenkins_admin and jenkins_tools are devops Jenkins instances

jibsheet (approved):
I'm fine having these applied to our jenkinses (what's the plural of jenkins, jenki?).  I hadn't gotten around to it.  We also tend to run with a private foo-jenkins.yml file for extra config where some of this is set.

PR #3441, opened 2016-10-31
update packer scripts to use ansible bootstrap script

estute:

### DIFF
--- a/util/packer/jenkins_worker.json
+++ b/util/packer/jenkins_worker.json
@@ -3,9 +3,11 @@
     "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
     "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
     "playbook_remote_dir": "/tmp/packer-edx-playbooks",
+    "venv_dir": "/edx/app/edx_ansible/venvs/edx_ansible",
     "ami": "{{env `JENKINS_WORKER_AMI`}}",
     "test_platform_version": "{{env `TEST_PLATFORM_VERSION`}}",
-    "delete_or_keep": "{{env `DELETE_OR_KEEP_AMI`}}"
+    "delete_or_keep": "{{env `DELETE_OR_KEEP_AMI`}}",
+    "remote_branch": "{{env `REMOTE_BRANCH`}}"
   },
   "builders": [{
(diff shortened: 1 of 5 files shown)


util/packer/jenkins_worker.json
```diff
@@ -28,65 +30,29 @@
...
-    "inline": ["sudo apt-get -y install libmysqlclient-dev"]
+    "source": "../../util/install/ansible-bootstrap.sh",
+    "destination": "{{user `playbook_remote_dir`}}/ansible-bootstrap.sh"
   }, {
     "type": "shell",
     "inline": ["cd {{user `playbook_remote_dir`}}",
-      "virtualenv packer-venv",
-      ". packer-venv/bin/activate",
-      "pip install -q -r requirements.txt"]
+      "export CONFIGURATION_VERSION='{{user `remote_branch`}}'",
```
benpatterson, line 39: I don't think this env variable will persist into the next step (`ansible-local`). Does it need to be declared again in the next stanza?

estute: No, it is used in the bootstrap script, not the plays

benpatterson: Well, here's what I'm after... One of the things you can do currently is run the jenkins-worker play according to the branch. So that way, we can make changes and test them via the build-ami job. Is that still possible with this change?

util/packer/jenkins_worker_simple.json
```diff
@@ -17,58 +20,32 @@
...
-      ". packer-venv/bin/activate",
-      "pip install -q -r requirements.txt"]
+      "export CONFIGURATION_VERSION='{{user `remote_branch`}}'",
+      "sudo bash ./ansible-bootstrap.sh" ]
   }, {
     "type": "ansible-local",
     "playbook_file": "../../playbooks/edx-east/jenkins_worker.yml",
     "playbook_dir": "../../playbooks",
-    "command": "source {{user `playbook_remote_dir`}}/packer-venv/bin/activate && ansible-playbook",
+    "command": ". {{user `venv_dir`}}/bin/activate && ansible-playbook",
```
benpatterson, line 45: you may want to say `source` explicitly. That'll work for shell or bash. Not a blocker though...thumb coming momentarily.

benpatterson:
👍

PR #3442, opened 2016-10-31
Jibsheet/use configuration source checking

jibsheet:
@edx-ops/devops OPS-1797

### DIFF
--- a/util/jenkins/build-ami.sh
+++ b/util/jenkins/build-ami.sh
@@ -88,6 +88,14 @@ if [[ ! -z "$configurationprivaterepo" ]]; then
   fi
 fi
 
+configurationinternal_params=""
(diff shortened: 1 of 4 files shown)


e0d:
:+1:

util/jenkins/build-ami.sh
```diff
@@ -88,6 +88,14 @@ if [[ ! -z "$configurationprivaterepo" ]]; then
   fi
 fi
 
+configurationinternal_params=""
```
e0d, line 91: why no _ here?

e0d (approved):

util/jenkins/build-ami.sh
jibsheet: Honestly, just because the rest of build-ami.sh code does configurationprivate_params and I was matching style.

PR #3444, opened 2016-11-01
OPS-1850: use nodeenv for edxapp

maxrothman:
@edx/devops @bjacobel

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -109,37 +109,6 @@
     - install:code
     - install:app-requirements
 
-# Set the npm registry
-# This needs to be done as root since npm is weird about
-# chown - https://github.com/npm/npm/issues/3565
-- name: Set the npm registry
-  shell:
-    npm config set registry '{{ COMMON_NPM_MIRROR_URL }}'
-    creates="{{ edxapp_app_dir }}/.npmrc"
-  environment: "{{ edxapp_environment }}"
-  tags:
-    - install
-    - install:app-requirements
-
-# Set the npm registry permissions
-- name: Set the npm registry permissions
-  file:
-    path: "{{ edxapp_app_dir }}/.npmrc"
-    owner: "{{ edxapp_user }}"
-    group: "{{ edxapp_user }}"
-  tags:
-    - install
-    - install:app-requirements
-
-# Node play that need to be run after platform updates.
-- name: Install edx-platform npm dependencies
-  shell: npm install chdir={{ edxapp_code_dir }}
-  become_user: "{{ edxapp_user }}"
-  environment: "{{ edxapp_environment }}"
-  tags:
-    - install
-    - install:app-requirements
-
 - name: Create the virtualenv to install the Python requirements
   command: >
     virtualenv {{ edxapp_venv_dir }}
@@ -237,6 +206,49 @@
   - "{{ sandbox_post_requirements }}"
   become_user: "{{ edxapp_user }}"
   when: not EDXAPP_PYTHON_SANDBOX
+  tags:
+    - install
+    - install:app-requirements
+
+- name: create nodeenv
+  shell: >
+    {{ edxapp_venv_dir }}/bin/nodeenv {{ edxapp_nodeenv_dir }} --node={{ edxapp_node_version }} --prebuilt
+  args:
+    creates: "{{ edxapp_nodeenv_dir }}"
+  tags:
+    - install
+    - install:app-requirements
+
+# Set the npm registry
+# This needs to be done as root since npm is weird about
+# chown - https://github.com/npm/npm/issues/3565
+- name: Set the npm registry
+  shell: >
+    npm config set registry '{{ COMMON_NPM_MIRROR_URL }}'
+  args:
+    creates: "{{ edxapp_app_dir }}/.npmrc"
+  environment: "{{ edxapp_environment }}"
+  tags:
+    - install
+    - install:app-requirements
+
+# Set the npm registry permissions
+- name: Set the npm registry permissions
+  file:
+    path: "{{ edxapp_app_dir }}/.npmrc"
+    owner: "{{ edxapp_user }}"
+    group: "{{ edxapp_user }}"
+  tags:
+    - install
+    - install:app-requirements
+
+- name: install node dependencies
+  npm:
+    executable: "{{ edxapp_node_bin }}/npm"
+    path: "{{ edxapp_code_dir }}"
+    production: yes
+  environment: "{{ edxapp_environment }}"
+  become_user: "{{ edxapp_user }}"
   tags:
     - install
     - install:app-requirements
--- a/playbooks/roles/common_vars/defaults/main.yml
+++ b/playbooks/roles/common_vars/defaults/main.yml
@@ -47,7 +47,7 @@ COMMON_ENVIRONMENT: 'default_env'
 COMMON_DEPLOYMENT: 'default_deployment'
 COMMON_PIP_VERBOSITY: ''
 COMMON_PYPI_MIRROR_URL: 'https://pypi.python.org/simple'
-COMMON_NPM_MIRROR_URL: 'http://registry.npmjs.org'
+COMMON_NPM_MIRROR_URL: 'https://registry.npmjs.org'
 COMMON_UBUNTU_APT_KEYSERVER: "http://keyserver.ubuntu.com/pks/lookup?op=get&fingerprint=on&search="
 
 COMMON_EDX_PPA: "deb http://ppa.edx.org {{ ansible_distribution_release }} main"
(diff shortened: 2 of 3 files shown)


bjacobel:
@maxrothman How should I go about testing this?

maxrothman:
I think building a sandbox off of this branch would work.

bjacobel:
ok, nodeenv.sandbox.edx.org is launching

maxrothman:
@bjacobel it's probably going to fail because I don't think nodeenv is in edxapp's requirements

EDIT: verified. Apps that use nodeenv have it in their requirements.txt, edx-platform does not. I'll make a quick PR.

maxrothman:
@bjacobel have you had a chance to test yet?

bjacobel:
@maxrothman I haven't, I didn't check the first sandbox I built after you said it would probably fail. Now that edx/edx-platform#13890 is merged I will build a box with `configuration_version` as `origin/max/edxapp-nodeenv` and see how it goes.

bjacobel:
@maxrothman Got an error building the sandbox: http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/9885/

Could be I just configured something wrong, I've never launched one with a non-master configuration version before. You probably know much more about this than me, you want to just ping me when there's a box I can SSH into and poke around?

maxrothman:
It's an issue with my code. Hang on, I'll fix it.

EDIT: @bjacobel fixed

maxrothman:
@bjacobel I built this sandbox from scratch. Though the build failed, it got past the edxapp section so you should be able to use it for testing.

bjacobel:
I SSHed in and did some poking around, everything seemed ok up until the point I started trying to run things with node modules that (I think?) should be installed on sandboxes.

```
 wheatley > ~ > code > edx-pattern-library > master > $ > ssh nodeenv.sandbox.edx.org
Welcome to Ubuntu 12.04.5 LTS (GNU/Linux 3.2.0-101-virtual x86_64)


*******************************************************************
*     ___                         _ __  __                        *
*    / _ \ _ __ ___ _ _    ___ __| |\ \/ / (R)                    *
*   | |_| | '_ \ -_) ' \  / -_) _` | >  <                         *
*    \___/| .__/___|_|_|  \___\__,_|/_/\_\                        *
*         |_|                                                     *
*                                                                 *
* This system is for the use of authorized users only.  Usage of  *
* this system may be monitored and recorded by system personnel.  *
*                                                                 *
* Anyone using this system expressly consents to such monitoring  *
* and is advised that if such monitoring reveals possible         *
* evidence of criminal activity, system personnel may provide the *
* evidence from such monitoring to law enforcement officials.     *
*                                                                 *
* Need help? https://open.edx.org/getting-help                    *
*                                                                 *
*******************************************************************
This system will be terminated on Nov 10, 2016
Last login: Fri Nov  4 14:15:39 2016 from dhcp-18-189-127-242.dyn.mit.edu
[nodeenv] bjacobel@nodeenv i-12022c8b:~$ which node
/usr/bin/node
[nodeenv] bjacobel@nodeenv i-12022c8b:~$ node --version
v0.10.37
[nodeenv] bjacobel@nodeenv i-12022c8b:~$ which npm
/usr/bin/npm
[nodeenv] bjacobel@nodeenv i-12022c8b:~$ npm version
{ http_parser: '1.0',
  node: '0.10.37',
  v8: '3.14.5.9',
  ares: '1.9.0-DEV',
  uv: '0.10.36',
  zlib: '1.2.8',
  modules: '11',
  openssl: '1.0.1l',
  npm: '1.4.28' }
[nodeenv] bjacobel@nodeenv i-12022c8b:~$ npm bin
/home/bjacobel/node_modules/.bin
[nodeenv] bjacobel@nodeenv i-12022c8b:~$ sudo su edxapp -s /bin/bash
edxapp@nodeenv:/home/bjacobel$ source ~/edxapp_env
edxapp@nodeenv:/home/bjacobel$ cd /edx/app/edxapp/edx-platform
edxapp@nodeenv:~/edx-platform$ which node
/edx/app/edxapp/nodeenvs/edxapp/bin/node
edxapp@nodeenv:~/edx-platform$ node --version
v0.10.37
edxapp@nodeenv:~/edx-platform$ which npm
/edx/app/edxapp/nodeenvs/edxapp/bin/npm
edxapp@nodeenv:~/edx-platform$ npm --version
1.4.28
```

(everything looked ok up to this point)

```
edxapp@nodeenv:~/edx-platform$ npm bin
/edx/app/edxapp/edx-platform/node_modules/.bin
edxapp@nodeenv:~/edx-platform$ ls -al $(npm bin)
ls: cannot access /edx/app/edxapp/edx-platform/node_modules/.bin: No such file or directory
edxapp@nodeenv:~/edx-platform$ $(npm bin)/eslint
bash: /edx/app/edxapp/edx-platform/node_modules/.bin/eslint: No such file or directory
edxapp@nodeenv:~/edx-platform$ ls -al /edx/app/edxapp/edx-platform/node_modules
ls: cannot access /edx/app/edxapp/edx-platform/node_modules: No such file or directory
edxapp@nodeenv:~/edx-platform$
```

Can't find any node modules in this nodeenv. Or maybe I'm not actually in the nodeenv? It's confusing because I named the sandbox nodeenv. But none of the node modules I would expect to be installed are in `/edx/app/edxapp/nodeenvs/edxapp/lib` either.

maxrothman:
@bjacobel let's connect later this morning and figure it out.

bjacobel:
:thumbsup: maxrothman.sandbox.edx.org is looking good with these changes

maxrothman:
Tests are running in https://github.com/edx/configuration/pull/3449 because this requires a change on master in edx-platform but the tests run on release.

maxrothman:
Tests have passed in #3449

e0d:
I'm :+1:  for this change.

@benpatterson is there anything you'd like to see before merging this?

benpatterson:
I'd look for some solution in the jenkins_worker role; that's currently using the chrislea PPA and that's how we'll be executing tests....

Where did you see that in the overall plan? Do we move one without the other?

e0d:
@benpatterson  Hmm, I'm a little surprised the worker is using different code -- I hadn't realized that when we discussed last week.  I think they should be the same, but that the key thing is that we are using the same node and npm versions.  The chrislea PPA is not available for xenial, so I think moving the nodeenv is the right thing to do.  Who can we talk to about your usecases for node/npm?

benpatterson:
@e0d Oh, I am totally on board with moving to nodeenv. I was moreso pointing out [this](https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_worker/tasks/system.yml#L29-L31). I think the reason that's there today might be legacy...when it was originally added, we did not have the `edxapp_common` role. So arguably it's a matter of cleaning up that side of things.

What I was also inferring was that using nodeenv has not been built into edxapp CI yet. Currently it's using the node/npm versions it finds on the machine it's running on (supplied by the `jenkins_worker` role). The questions at this juncture would revolve around how we'd want to roll out changes here. If you merge this and deploy using this, then there is the risk of different node environments. If, OTOH, we build this onto edx-platform first, then there is the risk of waiting for that to land (might be a couple of weeks? if we wait for the Tools team).

Happy to take this offline...it's a somewhat strategic/logistical set of questions. But I think we want the same thing here ultimately: both the test and prod infrastructures to be as similar as possible.

e0d:
I'd really like to take the steps to ensure that we're testing the same we we deploy.  Doesn't feel like this would be a huge effort, but I'd like to talk  through the repercussions.

@nedbat FYI

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -241,6 +210,49 @@
     - install
     - install:app-requirements
 
+- name: create nodeenv
+  shell: >
+    {{ edxapp_venv_dir }}/bin/nodeenv {{ edxapp_nodeenv_dir }} --node={{ edxapp_node_version }} --prebuilt
```
benpatterson, line 215: If you use `--force` then this can be idempotent

maxrothman: The `creates` arg below means that it won't attempt to create it when the directory is present. However, this won't replace a nodeenv with the wrong node version. Are you saying we should ditch `creates` and only use `--force`? That's not how we're doing it elsewhere.

I think the right approach might be to make that change in all the places that create nodeenvs in a separate PR.

benpatterson: Ah I missed the `creates` arg. No, you are right...that's how idempotency is achieved. I was more concerned with running the same script twice than I was changing the node version. Perhaps you need to register the node version and ensure that's what was created?

maxrothman: The more I think about it, the more I think `creates` actually does the wrong thing. I think `--force` is safer, but we should do it in a separate PR in all the places that make nodeenvs.

benpatterson:
Closing the loop on this, I am 👍

playbooks/roles/edxapp/tasks/deploy.yml
benpatterson: Agreed ^

benpatterson:
🎉

PR #3445, opened 2016-11-01
Removed AWS access key and secret variables for the Catalog Service

clintonb:
These defaults are conflicting with the overrides.

ECOM-6238


clintonb:
@edx/devops please review
FYI @edx/ecommerce

jibsheet:
This should fix the problem, but would be good if we left those commented out or something, just as breadcrumbs for external users standing up catalog.

PR #3446, opened 2016-11-02
test kernel patch is downloaded

estute:
@benpatterson @jibsheet 
This may be overkill, but just figured i would get an opinion.

### DIFF
--- a/playbooks/roles/jenkins_worker/defaults/main.yml
+++ b/playbooks/roles/jenkins_worker/defaults/main.yml
@@ -40,3 +40,12 @@ pip_accel_reqs:
   - "pip-accel==0.21.1"
   # pip-accel only makes the s3 functionality available if boto is installed
   - "boto=={{ common_boto_version }}"
+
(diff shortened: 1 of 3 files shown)


jibsheet:
I'd be ok with this as a one-off playbook for scanning for vulns in known hosts
But I know for us personally, we're moving away from the GHOST.c and bash checks in security since updates should be taken (and will be unnecesary on 16.04 which was never vulnerable).

I think having a more proactive pattern of applying security updates is the right way forward.

If you want to run this on your jenkins workers I won't object, but I wouldn't want to having anything other than a 'scan for vulns' option for our infra.

estute:
I guess this was more of a how much do we trust unattended-updates. I hope that if it were to timeout or not download, it would error out appropriately and the play would fail. Unfortunately, it's kind of a tough thing to test

jibsheet:
Oh, if you want to test that, spin up a box in the testeng account, block outbound https using a security group or the local firewally and run the security play against it.

estute:
Fair enough.

PR #3453, opened 2016-11-08
Feanil/test sandbox build

feanil:
Configuration Pull Request

---

  - [x] Sandbox Build Works from New Jenkins
  - [x] Sandbox Build works from Old Jenkins
  - [x] AMIs Still Build correctly
  - [x] AMIs still build correctly from scratch

When ansible is going over a NAT Gateway or other intermdiate
network devices, the connection can time out if that device has
its own timeout setting for no traffic.  Setting the ServerAliveInterval
will ensure that there is data sent over the wire ever 30 seconds so the
connection does not get closed while running ansible tasks that take
over 5 minutes.

### DIFF
--- a/docker/plays/ansible.cfg
+++ b/docker/plays/ansible.cfg
@@ -4,3 +4,5 @@ jinja2_extensions=jinja2.ext.do
 roles_path=../plays:../../playbooks/roles
 library=../../playbooks/library
 
+[ssh_connection]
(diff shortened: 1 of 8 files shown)


feanil:
Note: The other settings in the ssh_args are there because those are the defaults that get overridden when we set the ssh_args configuration ourselves.

e0d:
:+1:

PR #3455, opened 2016-11-08
Define new persistent grades queue

efischer19:
WIP branch for sandbox testing, will update as needed.

--

Configuration Pull Request

---

  - [ ] devops team member has commented with :+1:

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -283,8 +283,6 @@ EDXAPP_TIME_ZONE: 'America/New_York'
 EDXAPP_LMS_DOC_LINK_BASE_URL: 'https://edx.readthedocs.io/projects/open-edx-learner-guide'
 EDXAPP_CMS_DOC_LINK_BASE_URL: 'https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course'
 
-EDXAPP_RECALCULATE_GRADES_ROUTING_KEY: 'edx.lms.core.low'
-
 EDXAPP_TECH_SUPPORT_EMAIL: 'technical@example.com'
 EDXAPP_CONTACT_EMAIL: 'info@example.com'
 EDXAPP_BUGS_EMAIL: 'bugs@example.com'
@@ -499,6 +497,9 @@ EDXAPP_CELERY_WORKERS:
     concurrency: 1
     monitor: False
     max_tasks_per_child: 1
+EDXAPP_RECALCULATE_GRADES_ROUTING_KEY: 'edx.lms.core.default'
+EDXAPP_LMS_CELERY_QUEUES: "{{ edxapp_workers|selectattr('service_variant', 'equalto', 'lms')|map(attribute='queue')|map('regex_replace', '(.*)', 'edx.lms.core.\\\\1')|list }}"
+EDXAPP_CMS_CELERY_QUEUES: "{{ edxapp_workers|selectattr('service_variant', 'equalto', 'cms')|map(attribute='queue')|map('regex_replace', '(.*)', 'edx.cms.core.\\\\1')|list }}"
 
 EDXAPP_DEFAULT_CACHE_VERSION: "1"
 EDXAPP_OAUTH_ENFORCE_SECURE: True
@@ -1027,6 +1028,7 @@ lms_env_config:
   AUTH_DOCUMENTATION_URL: "{{ EDXAPP_AUTH_DOCUMENTATION_URL }}"
   DOC_LINK_BASE_URL: "{{ EDXAPP_LMS_DOC_LINK_BASE_URL }}"
   RECALCULATE_GRADES_ROUTING_KEY: "{{ EDXAPP_RECALCULATE_GRADES_ROUTING_KEY }}"
+  CELERY_QUEUES: "{{ EDXAPP_LMS_CELERY_QUEUES }}"
 
 cms_auth_config:
   <<: *edxapp_generic_auth
@@ -1057,6 +1059,7 @@ cms_env_config:
   SITE_NAME:  "{{ EDXAPP_CMS_SITE_NAME }}"
   GIT_REPO_EXPORT_DIR: "{{ EDXAPP_GIT_REPO_EXPORT_DIR }}"
   DOC_LINK_BASE_URL: "{{ EDXAPP_CMS_DOC_LINK_BASE_URL }}"
+  CELERY_QUEUES: "{{ EDXAPP_CMS_CELERY_QUEUES }}"
 
 # install dir for the edx-platform repo
 edxapp_code_dir: "{{ edxapp_app_dir }}/edx-platform"


efischer19:
Sandbox building at http://jenkins.edx.org:8080/job/ansible-provision/9945/

jibsheet:
@efischer19 Oh, right.  I'm going to fix the underlying default in the configuration repo, since I'm not 100% sure we're going to push grading onto everyone's deployment of rabbit queues (I suspect this PR may end up as a secure override).

efischer19:
@jibsheet @feanil I've confirmed this works on [my sandbox](https://efischer19.sandbox.edx.org/), using these changes and the recently-merged https://github.com/edx/edx-platform/pull/13995.

Do you want to merge the grades queue changes here, or move it into configuration-secure?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -282,7 +282,13 @@ EDXAPP_TIME_ZONE: 'America/New_York'
 EDXAPP_LMS_DOC_LINK_BASE_URL: 'https://edx.readthedocs.io/projects/open-edx-learner-guide'
 EDXAPP_CMS_DOC_LINK_BASE_URL: 'https://edx.readthedocs.io/projects/open-edx-building-and-running-a-course'
 
-EDXAPP_RECALCULATE_GRADES_ROUTING_KEY: 'edx.lms.core.low'
+EDXAPP_RECALCULATE_GRADES_ROUTING_KEY: 'edx.lms.core.grades'
```
feanil: Make this `edx.lms.core.default' by default.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -282,7 +282,13 @@ EDXAPP_TIME_ZONE: 'America/New_York'
...
+EDXAPP_CELERY_QUEUES:
+  - 'edx.lms.core.low'
+  - 'edx.lms.core.default'
+  - 'edx.lms.core.high'
+  - 'edx.lms.core.high_mem'
+  - '{{EDXAPP_RECALCULATE_GRADES_ROUTING_KEY}}'
```
feanil: Remove this from the list of celery queues.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -498,6 +504,10 @@ EDXAPP_CELERY_WORKERS:
     concurrency: 1
     monitor: False
     max_tasks_per_child: 1
+  - queue: grades
```
feanil: Remove this block.

feanil:
Per @jibsheet 's comment earlier, we will want to override this just for us and not make everyone build a new queue.  I commented on what parts to change because we do want to keep the change that wires the secure config into the correct env files.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1024,6 +1034,7 @@ lms_env_config:
   AUTH_DOCUMENTATION_URL: "{{ EDXAPP_AUTH_DOCUMENTATION_URL }}"
   DOC_LINK_BASE_URL: "{{ EDXAPP_LMS_DOC_LINK_BASE_URL }}"
   RECALCULATE_GRADES_ROUTING_KEY: "{{ EDXAPP_RECALCULATE_GRADES_ROUTING_KEY }}"
+  CELERY_QUEUES: "{{ EDXAPP_CELERY_QUEUES }}"
```
feanil: We'll want one of these for both the CMS and LMS right?

efischer19:
@feanil I'm not sure which comments you're referring to above - did you mean your comments on the edx-platform PR that merged earlier today?

As I understand things, it seems that the path forward is to close this PR, then make the same changes in secure config (which I do not have access to).

feanil (changes requested):

feanil:
Sorry, still getting used to this new review UI.  Request for changes submitted.

efischer19:
No worries @feanil, I wondered if that was what happened.

I've addressed your comments in my latest commit.

feanil:
@efischer19 I had another thought,  we currently build out these queue names from the var here: https://github.com/edx/configuration/blob/533fe8e2b647d1a40e0321c129a1489742dd148f/playbooks/roles/edxapp/defaults/main.yml#L481

Is it possible to generate the list you have here from that var so we don't have to repeat ourselves?

efischer19:
Let me give it a shot, I'll get back to you

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -498,6 +496,15 @@ EDXAPP_CELERY_WORKERS:
...
     max_tasks_per_child: 1
+EDXAPP_RECALCULATE_GRADES_ROUTING_KEY: 'edx.lms.core.default'
+EDXAPP_LMS_CELERY_QUEUES:
+  - edx."{{ item.service_variant }}".core."{{ item.queue }}"
+  with_items: "{{ EDXAPP_CELERY_WORKERS }}"
+  when: item.service_variant == 'lms'
+EDXAPP_CMS_CELERY_QUEUES:
+  - edx."{{ item.service_variant }}".core."{{ item.queue }}"
+  with_items: "{{ EDXAPP_CELERY_WORKERS }}"
+  when: item.service_variant == 'cms'
```
efischer19: @feanil I'm having trouble defining a variable using (what I would call in python) a list comprehension. My googling is failing; do you know how I can write the equivalent of this python code in yaml?

```
EDXAPP_LMS_CELERY_QUEUES = [
    "edx.{}.core.{}".format(worker.service_variant, worker.queue)
    for worker in EDXAPP_CELERY_WORKERS if worker.service_variant="lms"
]
```

I keep coming back to 2 issues:
- string interpolation is not a cut-and-dry operation at this level of yaml, it seems to happen in the jinja templates elsewhere (as seen in https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/templates/workers.conf.j2#L2)
- `with_items` seems to be more correctly used in tasks than variable definitions.

Am I barking up the wrong tree entirely? Would I be better off trying to modify `lms_env_config` in a jinja template like https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/templates/edxapp_env.j2?

feanil: Jinja provides a map filter, I believe you can use that in conjunction with the `format` function.

http://jinja.pocoo.org/docs/dev/templates/#map

If that doesn't work, I think we can deal with it later, I wouldn't spend too much more time on it than trying this out.

efischer19:
@jibsheet @feanil I think I have this working now - is there further testing we can do to confirm?

efischer19:
So it turns out this isn't working. Two inconveniences keep thwarting me:
- [it's very hard to set_fact a dynamic list](http://stackoverflow.com/questions/29399581/using-set-facts-and-with-items-together-in-ansible) 
- variable injection (modifying the previously set lms_env_config) doesn't seem to be a supported operation

efischer19:
I spoke with Kevin about this on my way out - I think the easiest way forward is to hardcode the values as we had been earlier, as that's cleaner than pushing this logic down into the lms.env.json.j2 template.

efischer19:
The latest commit gets us the lists we want, but I'm still unsure how to set those values into the `lms_env_config` variable.

```
root@efischer19:/edx/etc/playbooks# /edx/bin/ansible-playbook ./run_role.yml -i "localhost," -e role=edxapp -c local --limit "localhost" --tags "install:configuration"

...

TASK: [{{role}} | define service variant specific celery queues] ************** 
ok: [localhost]

TASK: [{{role}} | efischer19] ************************************************* 
ok: [localhost] => (item=lms) => {
    "item": "lms",
    "var": {
        "lms_celery_queues": [
            "edx.core.lms.low",
            "edx.core.lms.default",
            "edx.core.lms.high",
            "edx.core.lms.high_mem",
            "edx.core.lms.efischerdebug"
        ]
    }
}
ok: [localhost] => (item=cms) => {
    "item": "cms",
    "var": {
        "cms_celery_queues": [
            "edx.core.cms.low",
            "edx.core.cms.default",
            "edx.core.cms.high"
        ]
    }
}

...

PLAY RECAP ******************************************************************** 
INFO:ansible.callback_plugins.task_timing:{{role}} | create helper scripts for managing edxapp ----------------------------- 0.79s
INFO:ansible.callback_plugins.task_timing:{{role}} | create auth and application yaml config ------------------------------- 0.79s
INFO:ansible.callback_plugins.task_timing:{{role}} | create application and auth config ------------------------------------ 0.74s
INFO:ansible.callback_plugins.task_timing:{{role}} | create symlinks from the venv bin dir and repo dir -------------------- 0.24s
INFO:ansible.callback_plugins.task_timing:{{role}} | writing {{ item }} supervisor script ---------------------------------- 0.21s
INFO:ansible.callback_plugins.task_timing:{{role}} | writing edxapp and celery supervisor scripts -------------------------- 0.20s
INFO:ansible.callback_plugins.task_timing:{{role}} | add gunicorn configuration files -------------------------------------- 0.19s
INFO:ansible.callback_plugins.task_timing:{{role}} | enable {{ item }} supervisor script ----------------------------------- 0.14s
INFO:ansible.callback_plugins.task_timing:supervisor | Remove symlink for supervisor cfg ----------------------------------- 0.12s
INFO:ansible.callback_plugins.task_timing:supervisor | Create supervisor master config ------------------------------------- 0.08s
INFO:ansible.callback_plugins.task_timing:
Playbook run_role finished: 2016-11-22 15:13:22.911247, 20 total tasks.  0:00:04.201485 elapsed. 

localhost                  : ok=18   changed=2    unreachable=0    failed=0  
```

efischer19:
Alrighty @feanil, we're fully tested and ready to ride here. Okay to merge this?

jibsheet:
I'm ok with this, would love @feanil to also take a look since you folks may have talked in person and had more context. 👍

feanil (approved):

PR #3458, opened 2016-11-09
These are expanded in common.py, but too late

jibsheet:
We have to use the full name when overriding.

@edx/devops
@efischer19


efischer19 (approved):
👍

feanil:
:+1:

PR #3460, opened 2016-11-10
Install Python 3.5 for tools-edx-jenkins

rlucioni:
New ecom team scripts are written in Python 3. If we want to run them with tools-jenkins, we need Python 3 available.

ECOM-5658

@jibsheet as discussed. FYI @edx/devops @edx/ecommerce.

### DIFF
--- a/playbooks/roles/tools_jenkins/tasks/main.yml
+++ b/playbooks/roles/tools_jenkins/tasks/main.yml
@@ -30,3 +30,12 @@
   when: jpi_files
   notify:
     - restart Jenkins
+
+- name: Install system packages
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ jenkins_tools_debian_pkgs }}"
+  tags:
+    - install
+    - install:system-requirements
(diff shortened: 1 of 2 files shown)


playbooks/roles/tools_jenkins/tasks/main.yml
```diff
@@ -30,3 +30,11 @@
...
   notify:
     - restart Jenkins
+
+- name: Install system packages
+  apt:
+    name: "{{ item }}"
+    state: present
+  with_items: "{{ jenkins_tools_debian_pkgs }}"
+  tags:
+    - install:system-requirements
```
jibsheet, line 41: Can you add a - install here also?  Ansible doesn't support globbing (--tags install:*) so for now we double tag a lot of things.?

jibsheet (approved):
Add the extra tag and go ahead and merge.

playbooks/roles/tools_jenkins/tasks/main.yml
rlucioni: Sure, added.

clintonb (approved):

PR #3463, opened 2016-11-12
Remove .npmrc from jenkins worker.

benpatterson:
Will resolve TE-1835.

Not posting full URL, but this was verified in our packer build 3869 tonight. (All builds are failing for packer, without this fix.)

Configuration Pull Request
---


benpatterson:
CC @jzoldak @estute

jibsheet (approved):
GitHub isn't making it easy to tell if you emptied out .npmrc.j2 or deleted it, but I assume we can safely delete it now.

benpatterson:
@jibsheet locally I did `git rm` FWIW. Thanks for the quick look. Merging.

PR #3464, opened 2016-11-12
Add default value to fix failure

haikuginger:
This PR adds a default value for EDXAPP_USE_SWIFT_STORAGE that was missing initially in #3349.

@jibsheet, FYI.


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=3464"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

haikuginger:
@jibsheet, I ran a quick deploy and verified that edxapp no longer errors out with this variable missing.

jibsheet (approved):

PR #3465, opened 2016-11-14
Sets FORUM_API_KEY to EDXAPP_COMMENTS_SERVICE_KEY if available

pomegranited:
`EDXAPP_COMMENTS_SERVICE_KEY` must match `FORUM_API_KEY` for edxapp to communicate with the forum/cs_comments_service API.  

Currently, `EDXAPP_COMMENTS_SERVICE_KEY` and `FORUM_API_KEY` just happen to have the same default value, and both variables must be set to allow authentication:

[edxapp/defaults/main.yml:120](https://github.com/edx/configuration/blob/47dd7dee991b9ee85f7cf6164fc1587db0f652a1/playbooks/roles/edxapp/defaults/main.yml#L120):
```yaml
EDXAPP_COMMENTS_SERVICE_KEY:  'password'
```

[forum/defaults/main.yml:L27](https://github.com/edx/configuration/blob/47dd7dee991b9ee85f7cf6164fc1587db0f652a1/playbooks/roles/forum/defaults/main.yml#L27):
```yaml
FORUM_API_KEY: "password"
```

This change retains the original default value of `password` for both variables.

**Reviewers**
- [x] @mtyaka 
- [x] @feanil 
- [ ] @e0d


openedx-webhooks:
Thanks for the pull request, @pomegranited! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=3465"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

openedx-webhooks:
Thanks for the pull request, @pomegranited! I've created [OSPR-1579](https://openedx.atlassian.net/browse/OSPR-1579) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

mtyaka:
I think this change makes a lot of sense, thanks @pomegranited!

:+1:
- [x] I tested this: I spawned a new instance where I set the `EDXAPP_COMMENTS_SERVICE_KEY` variable to a random value, but set `FORUM_API_KEY` undefined. The instance was deployed successfully and the forums are working correctly.
- [x] I read through the code
- [ ] ~~Includes documentation~~ (backwards compatible change; documentation not needed)

feanil:
:+1:   @e0d second review.

maxrothman:
This is problematic because it assumes that the edxapp role will be run every time the forums role is run, which is not necessarily true if they're on different servers.

The default will keep it from failing outright, but I'm wary of introducing dependencies across roles. That's something we've generally tried to avoid.

As a result, I'm 👎

pomegranited:
@maxrothman I didn't think it assumes the edxapp role is run on the forum server?  But it does assume that you've got a unified set of variables that you're using when provisioning the edxapp and forum servers.  

I looked around to see whether there was a canonical way of sharing variables between services and didn't find one.  Could you point me towards an example?

maxrothman:
Sorry, I should clarify. We and many others happen to use a system in which overrides for all roles are kept in the same place so a role can get access to other roles' overrides, but that's more of an implementation detail than a guarantee. We do however try to guarantee that running any role by itself with no overrides "does the right thing".

There currently is no pattern for sharing variables between services, hence the duplicated key. If you look, you'll see that lack-of-a-pattern used in lots of places. In our overrides, we solve it by having both variables be overridden with a reference to the same value, e.g. `thing: foo; place1: "{{ thing }}"; place2: "{{ thing }}"`.

We've made a conscious choice in the past to avoid having roles reach into each others' "namespaces" even if that means that values aren't properly normalized. It's definitely a tradeoff, but I think it's the correct one.

pomegranited:
@maxrothman We've repaired the issue on our side by just setting both variables, so we're ok without this code change.

Would you be ok with me replacing this PR with a simple comment on each variable, noting that the values must match?

maxrothman:
@pomegranited definitely! That would be useful.

pomegranited:
@maxrothman Makes this PR a bit gutless, but hey :smile:

b737cb8 replaces code change with a comment.  Squashed with the original https://github.com/edx/configuration/commit/d0ef48676b7127655fb0cb33415dbd13e1b3f28a.

@feanil , @mtyaka Are you still ok with this?

mtyaka:
The comments will be very helpful 👍

maxrothman (approved):

pomegranited:
@feanil Is it ok for me to merge this PR?

maxrothman:
(Merge at will, commander)

feanil:
:+1:

PR #3469, opened 2016-11-15
Mark timing logs with their own sourcetype

cpennington:
Configuration Pull Request
---


e0d (approved):
👍

PR #3470, opened 2016-11-15
Switch back to ondrej's mysql PPA

jibsheet:
Longer reasoning in the comment in the file

@edx/devops

FYI @nedbat

### DIFF
--- a/docker/plays/mysql.yml
+++ b/docker/plays/mysql.yml
@@ -1,0 +1,8 @@
+- name: Deploy MySQL 5.6
+  hosts: all
+  sudo: True
+  gather_facts: True
+  roles:
+    - common_vars
+    - docker
+    - mysql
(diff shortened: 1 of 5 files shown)


nedbat:
@jibsheet so this should be correct for xenial also?

jibsheet:
@nedbat correct.  You mentioned you had picked it to your branch, but I don't think you pushed (I got errors on your branch until I brought this in manually).

docker/plays/mysql.yml
```diff
@@ -0,0 +1,8 @@
+- name: Deploy autom
```
e0d: nit: spelling

e0d (approved):
one nit, but I approve.

nedbat:
@jibsheet oops, right, I didn't push the cherry-pick.

PR #3472, opened 2016-11-16
Upgrade Hadoop and Hive

brianhw:
Configuration Pull Request
---

* Upgrade Hadoop from 2.3.0 to 2.7.2.
* Upgrade Hive from 0.2.0 to 1.0.0.

Configuration changes:
* Disable datanucleus.autoCreateSchema.
* Instead, call Hive's schematool -initSchema

### DIFF
--- a/playbooks/roles/analytics_pipeline/tasks/main.yml
+++ b/playbooks/roles/analytics_pipeline/tasks/main.yml
@@ -229,3 +229,10 @@
   tags:
     - install
     - install:configuration
+
+- name: Initialize Hive metadata store schema
+  shell: ". {{ HADOOP_COMMON_CONF_DIR }}/hadoop-env.sh && {{ HIVE_HOME }}/bin/schematool -dbType mysql -initSchema"
+  become_user: "{{ hadoop_common_user }}"
+  tags:
+    - install
+    - install:base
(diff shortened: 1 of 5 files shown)


playbooks/roles/hive/templates/hive-site.xml.j2
```diff
@@ -23,6 +23,6 @@
   </property>
   <property>
     <name>hive.metastore.schema.verification</name>  
-    <value>true</value>
```
mulby: Do we actually want this to be `true` but actually call the schema tool to initialize the schema?

That is, perhaps we want to remove the `autoCreateSchema` property and instead add an ansible task that does something like:

``` bash
schematool -dbType mysql -initSchema
```

mulby (approved):
Generally looks good (assuming it works)

We may want to consider an explicit call to the schematool.

brianhw:
This has incorporated commented from Gabe's review of an earlier iteration.  I have built a box and tested that this works on some edx-analytics-pipeline acceptance tests.  The PR is ready for review again.  @mulby in particular, since I can't invite you to review a PR you have already approved.

azafty468 (approved):
👍 . It looks like hadoop switched from a release name to a hadoop name.  Is that correct?  Or is there a release version that is more appropriate for us to use?

brianhw:
My understanding is that https://archive.apache.org/dist/hadoop/common/ is now where they keep all of their releases.   The old location only goes up to 2.5.0.

mulby:
Still LGTM

feanil (approved):

PR #3477, opened 2016-11-16
Updates for 16.04 MongoDB

jibsheet:
This adds a config file and backports the systemd module so we
can use it elsewhere as needed.

@edx/devops

### DIFF
--- a/playbooks/roles/mongo_2_6/tasks/main.yml
+++ b/playbooks/roles/mongo_2_6/tasks/main.yml
@@ -118,6 +118,27 @@
     - install
     - install:configuration
 
+- name: Add mongod systemd configuration on 16.04
+  template:
+    src:  "etc/systemd/system/mongod.service.j2"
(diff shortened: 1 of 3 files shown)


jibsheet:
@nedbat on this branch I can get all the way to Notifier failing to syncdb which sounds like the sort of problem you can debug.

playbooks/roles/mongo_2_6/tasks/main.yml
```diff
@@ -118,6 +118,28 @@
...
 
+- name: Add mongod systemd configuration on 16.04
+  template:
+    src:  "etc/systemd/system/mongod.service.j2"
+    dest: "/etc/systemd/system/mongod.service"
+  when: ansible_distribution_release == 'xenial'
+  register: template_systemd_configuration
+  notify:
+    - restart mongo
+  tags:
```
e0d, line 128: move the when.

e0d (dismissed):
One small change, then feel free to merge.

PR #3476, opened 2016-11-16
Bump the sandbox version

jibsheet:
The playbook the CI job runs trips over the PPA upgrade in
https://github.com/edx/configuration/pull/3470

@edx/devops

PR #3479, opened 2016-11-16
Revert "Bump the sandbox version"

jibsheet:
Reverts edx/configuration#3476

Go back to the old PPA on 12.04 in order to maintain backcompat with devstacks.
This has the side-effect of impacting sandboxes, since they're both 12.04.

https://github.com/edx/configuration/pull/3478


e0d:
:+1:

PR #3484, opened 2016-11-18
Make Vagrantfiles less hard-coded, and more uniformly configurable

nedbat:
I needed to test a fix in the notifier repo, but could not because NOTIFIER_VERSION was not available in the Vagrantfile.  This makes all the version variables overridable, with less repetition.

### DIFF
--- a/vagrant/base/devstack/Vagrantfile
+++ b/vagrant/base/devstack/Vagrantfile
@@ -13,6 +13,22 @@ if ENV["VAGRANT_GUEST_IP"]
   vm_guest_ip = ENV["VAGRANT_GUEST_IP"]
 end
 
+# These are versioning variables in the roles. Each can be overridden, first
+# with OPENEDX_RELEASE, and then with a specific environment variable of the
+# same name but upper-cased.
+VERSION_VARS = [
+    'edx_platform_version',
+    'configuration_version',
+    'certs_version',
+    'forum_version',
+    'xqueue_version',
+    'demo_version',
+    'NOTIFIER_VERSION',
+    'ECOMMERCE_VERSION',
+    'ECOMMERCE_WORKER_VERSION',
+    'PROGRAMS_VERSION',
+]
+
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Creates a devstack from a base Ubuntu 12.04 image for virtualbox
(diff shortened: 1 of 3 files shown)


nedbat:
@edx/devops @cpennington Thoughts?  I can make similar changes to other Vagrantfiles once we are done with these.

vagrant/base/devstack/Vagrantfile
```diff
@@ -13,6 +13,22 @@ if ENV["VAGRANT_GUEST_IP"]
   vm_guest_ip = ENV["VAGRANT_GUEST_IP"]
 end
 
+# These are versioning variables in the roles. Each can be overridden, first
+# with OPENEDX_RELEASE, and then with a specific environment variable of the
+# same name but upper-cased.
+VERSION_VARS = [
+    'edx_platform_version',
```
e0d, line 20: Why not converge format now?

vagrant/release/devstack/Vagrantfile
```diff
@@ -84,6 +68,30 @@ openedx_releases.default = "eucalyptus-devstack-2016-09-01"
...
+    exit 1
+fi
+export PYTHONUNBUFFERED=1
+source /edx/app/edx_ansible/venvs/edx_ansible/bin/activate
+cd /edx/app/edx_ansible/edx_ansible/playbooks
+
+EXTRA_VARS="#{extra_vars_lines}"
+CONFIG_VER="#{rel || 'master'}"
+
+ansible-playbook -i localhost, -c local run_role.yml -e role=edx_ansible -e configuration_version=$CONFIG_VER $EXTRA_VARS
```
e0d, line 92: Do we remember why two call was necessary here, makes things significantly slower than they could be.

e0d (commented):
Generally looks great, with a couple small questions.

vagrant/base/devstack/Vagrantfile
nedbat: Just a general instinct to make small changes, but I can go through and make them all upper-case if you like.

vagrant/release/devstack/Vagrantfile
nedbat: I assumed this was the usual two-step: get ansible (etc) right, then use ansible to do all the rest.  Changing this definitely feels outside the scope of this PR.

vagrant/release/devstack/Vagrantfile
```diff
@@ -8,37 +8,21 @@ VAGRANTFILE_API_VERSION = "2"
...
-  CONFIG_VER="master"
-fi
-
-ansible-playbook -i localhost, -c local run_role.yml -e role=edx_ansible -e configuration_version=$CONFIG_VER $EXTRA_VARS
-ansible-playbook -i localhost, -c local vagrant-devstack.yml -e configuration_version=$CONFIG_VER $EXTRA_VARS
-
-SCRIPT
+# These are versioning variables in the roles. Each can be overridden, first
+# with OPENEDX_RELEASE, and then with a specific environment variable of the
+# same name but upper-cased.
```
cpennington, line 13: It doesn't look like you implemented the per-app variable overrides in this vagrant file. Am I missing something?

nedbat: Oh, you are right! It didn't exist at all in the original version, and I didn't add it either! oops.

nedbat: Added.

cpennington (approved):
The casing-gymnastics that you're going through make me wonder whether it would be better to just standardize across the whole repo, but I think it's good to go as-is.

nedbat:
I'll wait on making all the version variables uppercase.  I'd rather move on to the other Vagrantfiles first.

PR #3485, opened 2016-11-18
Change to wiredTiger

jibsheet:
If this play is being run against an existing mmapv1 mongo 3.0 install,
make sure MONGO_STORAGE_ENGINE is overridden.

This will facilitate shipping 3.0 with newer release.

@edx/devops
We're not using this role on sandboxes/devstacks so this only impacts deployers (such as us).


maxrothman (approved):
This change could really impact external deployers, so can you send an email to the ops list?

jibsheet:
@maxrothman Yep, my plan was notify and merge after Turkey.  I was waiting to see if we would have the config hangout this week to also mention it.

PR #3486, opened 2016-11-21
Add an install:app-configuration that only generates yml/json

jibsheet:
This is helpful for uses that want to generate just config for
edxapp or any of our IDAs


jibsheet:
@edx/devops This assists in the work we're doing to autobake and test configuration (standardizes tags for IDAs and edxapp)

feanil:
:+1:

feanil (approved):

PR #3496, opened 2016-11-23
Allow for extra dev packages to be installed on docker images.

feanil:
Review: @e0d  @clintonb @jibsheet 

Here is what I was thinking regarding this conversation(https://github.com/edx/ecommerce/pull/1026#discussion_r89206094).

### DIFF
--- a/docker/build/ecommerce/ansible_overrides.yml
+++ b/docker/build/ecommerce/ansible_overrides.yml
@@ -13,3 +13,10 @@ ECOMMERCE_DATABASES:
     PORT: '3306'
     ATOMIC_REQUESTS: true
     CONN_MAX_AGE: 60
+
+# Add test specific packages.
+ECOMMERCE_DEBIAN_EXTRA_PKGS:
+  - firefox
+  - xvfb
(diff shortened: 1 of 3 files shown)


clintonb (approved):
This will be needed for all IDAs.

jibsheet (approved):

e0d:
@feanil I have some questions about this.  Let me write them out and we can discuss.

feanil:
@clintonb I'm happy to merge this but will not be making the changes to the other IDAs.  If we're happy with this as an approach to achieve this, I think in general, we should be documenting the best practice and let the IDA teams update as it makes sense to them.  Thoughts?

feanil:
@clintonb @e0d thinks this merits further discussion and so I setup a meeting for tomorrow to do that.  Given that this is about designing a new approaching to keep test and production settings separate, I think this should not hold up the work I have previously done.  Once we have decided on a final approach, the IDA teams can update their testing based on what is decided.  In the meantime I'd like to merge https://github.com/edx/ecommerce/pull/1026 as is.

PR #3497, opened 2016-11-23
Entrypoint can't be overridden, it should not be used this way.

feanil:
Configuration Pull Request
---


feanil:
FYI @edx/ecommerce

rlucioni (approved):

clintonb (approved):

jibsheet (approved):
Should I even ask about xqueue?

PR #3498, opened 2016-11-23
SEC-220: Block external logged in checking

maxrothman:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -116,6 +116,20 @@ error_page {{ k }} {{ v }};
   {% include "python_lib.zip.j2" %}
   {% include "common-settings.j2" %}
 
+  {% if NGINX_EDXAPP_EMBARGO_CIDRS -%}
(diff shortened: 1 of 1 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -116,6 +116,15 @@ error_page {{ k }} {{ v }};
   {% include "python_lib.zip.j2" %}
   {% include "common-settings.j2" %}
 
+  {% if EDXAPP_LMS_ENABLE_BASIC_AUTH|bool %}
```
feanil: Won't this result in basic auth being turned on for all urls?  I believe we don't  want this for the other clauses and that is why they have been specifically pulled out.

maxrothman: Which URLs do we not want basic auth for? I made this change because I noticed it probably should be on more than just the `/` location (e.g. `/api`).

feanil:
:+1: once embargoes have been tested.

feanil (approved):

PR #3504, opened 2016-11-28
We don't want to recreate these files

jibsheet:
Rabbit's logs go to /var/log/rabbitmq/ these files are our internal
queue monitoring logs and they will be recreated by
/edx/app/rabbitmq/log-rabbitmq-queues.sh

Without this, we end up with > 10K empty files which makes logrotate and
splunk sad.

@edx/devops


feanil:
:+1:

PR #3505, opened 2016-11-28
remove pip installs for android utils

estute:
adb used to be in tools and adding a symlink makes for nicer scripting.
Also, the mobile app now uses a makefile with make requirements, so no need for the pip installs here


benpatterson:
What kind of effect will this have on the time it takes to get an environment ready?

estute:
@benpatterson negligible (seconds). I did this so that in case requirements change, mobile can manage it without a change here. It does open up to pip timeouts, but this is very small risk IMO

benpatterson:
👍

jzoldak:
Nice. 👍 

If we want to post codecov coverage in non-python repos, we can (and should) use the [bash uploader](http://docs.codecov.io/docs/about-the-codecov-bash-uploader).

jibsheet (approved):

PR #3507, opened 2016-11-29
Turn discovery docker build back on.

feanil:


jibsheet (approved):
Looks good - assuming you'll remove the # in common now that tests have run.

PR #3509, opened 2016-11-29
Ned/xenial native

nedbat:
Configuration Pull Request
---


nedbat:
@jibsheet @saleem-latif I removed "no_log: true" lines because an installation failed on one of those steps, and I was frustrated to have no information about what it was trying to do. I didn't see any reason to censor those steps.  I know "no_log: true" is used to avoid putting sensitive information into logs, but I don't see what sensitive information is in the git_clone tasks.

These lines have a long history of coming and going:
@saleem-latif you put them in with https://github.com/edx/configuration/commit/0fcedc07a01b10d169c09b2d0112010371b62242
@jibsheet you took them out in https://github.com/edx/configuration/commit/22a3d1c4651d30508aea578ca287f8b550560a06
@saleem-latif you put them back in at https://github.com/edx/configuration/commit/e51511fcadd575d2554b731aad5f32544086c2c9

Can we come to a stable agreement about whether we want them or don't want them?

saleem-latif:
@nedbat `git_clone` takes repo info (including ssh keys) as argument ([ref](https://github.com/edx/configuration/blob/master/playbooks/roles/edx_themes/meta/main.yml#L21), [ref](https://github.com/edx/configuration/blob/master/playbooks/roles/git_clone/defaults/main.yml#L18-L29)), these ssh keys are logged if we do no use `no_log: true`.

nedbat:
@saleem-latif OK, I can see that "Install read-only ssh key" should have "no_log" set, and I've put that back.  But the other operations in that file don't need it, right?

saleem-latif:
@nedbat I think all the tasks that contain `GIT_REPOS` variable will need `no_log: true` option. There is another option [[link](http://stackoverflow.com/questions/39189549/how-can-i-hide-skipped-tasks-output-in-ansible)] to use a custom callback to suppress log messages but I am not much familiar with that, may be @jibsheet can give better insight on that.

nedbat:
OK, I understand now about the no_log setting, and I've removed that change. Thanks.

jibsheet:
I spoke to Ned in person, and believe Saleem covered most of the issues.
I've considered replacing true with GIT_CLONE_SUPPRESS_LOGS: true in their defaults, so that if something fails (or even for all devstacks) we could -e GIT_CLONE_SUPPORT_LOGS=false temporarily.  But that would require a big comment because it's too easy to log secrets (even as discussed with Ned, on tasks that don't appear to consume a key directly.  Ansible is too chatty.).

nedbat:
@edx/devops These changes are complete: I can finish a Native installation from this branch.   I'm not sure how you'd like to handle the conditionalization that should happen before this is merged to master.

feanil:
@nedbat when you say conditionalization do you mean just for jdk7 vs jdk8 or for other things as well?

jibsheet:
As Feanil asks, I'm really only concerned about breaking insights in some way before we get them deployed on 16.04 (I don't know if jdk8 works on 12.04)

nedbat:
Other places the might need conditionals:  the script stops if the OS isn't Xenial.  If I merge this to master, we won't be able to build sandboxes on Precise, right?  Similarly, the versions of C compiler.

jibsheet:
edX sandboxes do not use sandbox.sh
OpenCraft uses edx-sandbox.yml but I don't believe they use the script.  @smarnach Can you chime in on how you build your sandboxes?

At this point, master of configuration is becoming less and less precise friendly.

smarnach:
@jibsheet We are using the edx_sandbox.yml playbook, but not the sandbox.sh script, so as long as the edx_sandbox.sh playbook keeps working for now, all is fine.

nedbat:
@edx/devops The Travis test is failing with:
```
TASK: [edx_service | Install a bunch of system packages on which edx_service relies] *** 

failed: [127.0.0.1] => (item=libmysqlclient-dev,build-essential,gettext,openjdk-8-jdk) => {"failed": true, "item": "libmysqlclient-dev,build-essential,gettext,openjdk-8-jdk"}

msg: No package matching 'openjdk-8-jdk' is available
```
Please advise.

jibsheet:
@nedbat I think that's the insights role.  It'll need conditionalizing following the pattern we used elsewhere.  We need to add a ticket to https://openedx.atlassian.net/browse/OPS-1524 for insights/analytics-api which is when we'll get to it @fredsmith

nedbat:
OK, so we don't need to fix that failing Travis test before merging?  Any other concerns?

jibsheet:
Well, something needs to be fixed before merging, otherwise anything that impacts the insights role will fail travis tests and I don't think it's ok to have known failing tests.

jibsheet:
@nedbat I've written https://openedx.atlassian.net/browse/OPS-1923 we'll groom and see where it fits in the sprint tomorrow

gsong:
@nedbat Should this PR be closed in favor of one from <https://github.com/edx/configuration/tree/ned/test-ficus>?

nedbat:
Yes. :)

PR #3510, opened 2016-11-30
install cached android emu on jenkins workers

estute:
@jzoldak @benpatterson PTAL
@jibsheet 

this is a *hopefully* temporary fix for https://code.google.com/p/android/issues/detail?id=228113. I pulled the emulator off of my machine and hosted it on s3.

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/test_android_worker.yml
+++ b/playbooks/roles/jenkins_worker/tasks/test_android_worker.yml
@@ -37,3 +37,11 @@
   assert:
     that:
       - "item.stat.exists == True"
+# TEMP: until we either change to using a different system image OR google hosts
+# the android 23 image again
+- name: Verify that cached Android sys image is installed
+  shell: "stat {{ android_home }}/system-images/android-23"
+  register: android_23_stat
+- assert:
+    that:
+        - "android_23_stat.rc == 0"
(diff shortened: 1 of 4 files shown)


estute:
In case this is not a good fix, the next move would be to change the emulators the mobile team uses (of course all other system images are still available ;) ), but this has caused tests to fail in the past, and I am trying to minimize work for them. Let me know what you guys think.

playbooks/roles/jenkins_worker/tasks/test_android_worker.yml
```diff
@@ -37,3 +37,11 @@
   assert:
     that:
       - "item.stat.exists == True"
+# TEMP: until we either change to using a different system image OR google hosts
+# the android 23 image again
```
benpatterson, line 41: nit: reference the issue url that you pointed out to us in the main body of the pull request

benpatterson:
I have a nit, but 👍 either way

estute:
@feanil, in case Kevin is swamped

jibsheet:
Please include the issue link per Ben's request and make an Ostrich joke.
We should consider an ansible library module for curl/checksum.

jibsheet (approved):

playbooks/roles/jenkins_worker/tasks/test_android_worker.yml
estute: Done. added it in the actual play that does the installing

PR #3513, opened 2016-12-01
Allow workers to queue tasks across service variants

efischer19:
# [TNL-5408](https://openedx.atlassian.net/browse/TNL-5408)

@jibsheet @nasthagiri 

I believe this fixes the issue described in the ticket. The problem (as uncovered by Nimisha's observations) is that, when a *worker* tries to queue a task across service variants, it can't because we only ever told the lms and cms main processes how to do that (see https://github.com/edx/configuration/pull/3111)

There's probably a cleaner way to do this, but I wanted to get the basic idea recorded while I had a grip on it.

---

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -1029,6 +1029,7 @@ lms_env_config:
   DOC_LINK_BASE_URL: "{{ EDXAPP_LMS_DOC_LINK_BASE_URL }}"
   RECALCULATE_GRADES_ROUTING_KEY: "{{ EDXAPP_RECALCULATE_GRADES_ROUTING_KEY }}"
   CELERY_QUEUES: "{{ EDXAPP_LMS_CELERY_QUEUES }}"
+  ALTERNATE_WORKER_QUEUES: "cms"
 
 cms_auth_config:
   <<: *edxapp_generic_auth
@@ -1060,6 +1061,7 @@ cms_env_config:
   GIT_REPO_EXPORT_DIR: "{{ EDXAPP_GIT_REPO_EXPORT_DIR }}"
   DOC_LINK_BASE_URL: "{{ EDXAPP_CMS_DOC_LINK_BASE_URL }}"
   CELERY_QUEUES: "{{ EDXAPP_CMS_CELERY_QUEUES }}"
+  ALTERNATE_WORKER_QUEUES: "lms"
 
 # install dir for the edx-platform repo
 edxapp_code_dir: "{{ edxapp_app_dir }}/edx-platform"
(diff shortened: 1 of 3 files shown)


efischer19:
@jibsheet low priority on reviewing this - going to pick up some other work today that we're trying to get merged into platform before the release cut tomorrow.

nasthagiri:
sweet!
And we can verify that this is caused by course imports running asynchronously when we test this on a sandbox.

feanil:
Is there a reason this couldn't go into lms_env config and cms_env config instead?  ie. Into the output json files?  Since the workers and the app servers share the json files, this would have prevented this issue.  Generally, we want to be picking up settings from the config files as much as possible and only use the environment for settings that need to be in the environment because they are out of our control, like how newrelic or django might read settings they care about.

efischer19:
@feanil That's definitely something we can do. I believe the reason we originally defined `ALTERNATE_SERVICE_QUEUES` in lms/cms.conf was to have it linked closely to where `SERVICE_VARIANT` is defined ([see comments here](https://github.com/edx/edx-platform/pull/12689/files#r66989240)) but there's nothing forcing it to stay there if you think it would be better off in the json files.

feanil:
@efischer19, makes sense.  But I think you could put them in here: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L996 and https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L1057 would ensure the same thing.

efischer19:
Yeah, that ought to work well. I'll fix that up when I get back into this PR. Thanks!

efischer19:
I've confirmed this change and its companion https://github.com/edx/edx-platform/pull/14090 to work successfully on my sandbox, even when importing a course.

@jibsheet or @feanil Can I get a review on both PRs?

nasthagiri (approved):

jibsheet:
Since this is now in lms.env.json / cms.env.json do we need to set the environment variable in the workers template?

efischer19:
@jibsheet I didn't think it was needed, as the workers will read from lms/cms.env.json, correct?

jibsheet:
@efischer19 Correct - I should been clearer - I was trying to say "please delete the thing in workers.conf.j2 and it'll read from lms/cms env.json"

efischer19:
Ah, I follow now. I could have been clearer in describing the point of this PR - we caused this bug by *never* defining `ALTERNATE_WORKER_QUEUES` in workers.conf.j2, so there's nothing to delete.

jibsheet:
Oh, interesting.  So, workers isn't loading it from lms.env.json?  It looks like we're currently sending it to the worker process in two ways, so wanted to try and limit that down to one if we could.

jibsheet (approved):

PR #3515, opened 2016-12-01
WIP: Packer build of Xenial base vagrant box

nedbat:


nedbat:
In retrospect, we're better off using community-maintained templates, like boxcutter: https://github.com/boxcutter/ubuntu

PR #3516, opened 2016-12-01
Capture both kinds of node bin dirs

jibsheet:
Nodeenv contains npm
node_modules is where npm installs things which are used
say paver test_js and friends

Caused by 7ca232dc83e5d6570b5aec42d3b33a5e25c95e5f

It's possible a better solution is to ensure that nodeenvs/edxapp/bin/activate is sourced in bashrc instead


jibsheet:
I'd be interested in @bjacobel and @arizzitano 's thoughts on sourcing bin/activate vs jamming the path into PATH

FYI @edx/devops I wrote https://openedx.atlassian.net/browse/OPS-1917 which we should groom in the morning and I'll need someone to help unbreak js testing on new devstacks.

jibsheet:
Oh, nevermind, sourcing nodeenv doesn't get your npm's destination directory.
All packaging is terrible.
I suspect this is the correct solution but someone will need to assist in testing and reading commit messages and explanations.

jibsheet:
paver test_js works on a devstack provisioned with this

bjacobel:
This looks good to me, I'll have my laptop spin up a devstack using this during demos to check it out.

bjacobel:
Actually... I might not know how to have two devstacks at once.

jibsheet:
@bjacobel  What I do is 
```
cd devstack
vagrant halt
mkdir ../devstack2
cd ../devstack2
Maybe copy all the git repos from ../devstack to ../devstack2
curl down Vagrantfile from https://raw.githubusercontent.com/edx/configuration/master/vagrant/release/devstack/Vagrantfile
Add CONFIG_VER="jibsheet/node-devstacks" on line 38 after CONFIG_VER is defined to master and before ansible-playbook is called
```

I can help configure / debug in person if needed, I'm in the office today.

feanil (approved):

PR #3518, opened 2016-12-02
Update other Vagrantfiles to be overridable like devstack

nedbat:


nedbat:
@edx/devops @edx/analytics Trying to keep Vagrantfiles in sync.  Next up: make this code importable from a common place.

feanil (approved):

PR #3519, opened 2016-12-02
Jibsheet/revert openstack overriding aws

jibsheet:
This reverts https://github.com/edx/configuration/pull/3349 and https://github.com/edx/configuration/pull/3464 and exposes a fatal flaw in the design of vhost + (openstack|aws)

Ansible parses and sets all the variables during initial playbook load, aws sets VHOST_NAME to aws, but openstack is included and sets it to openstack last.  No tasks are run, but variables are still evaluated.

Once this PR was merged, our builds began having /edx/app/openstack and other openstack directories.  Luckily, the aws directories still existed, and the aws role is still run, however builds from a fresh AMI (such as for our 16.04 work) fail because /edx/app/aws does not exist.

There's a deeper conversation here about how to correctly separate openstack and aws without clobbering variables like this and more places in the future.

@haikuginger 
@edx/devops  Please Review, I'll build a test AMI from it


haikuginger:
@jibsheet, thanks for the ping on this. I agree that this obviously needs some more consideration. I find it odd that variables from an include are loaded if that include isn't actually run, but if that's what Ansible is doing, then we need to consider that.

@bradenmacdonald, @antoviaque, FYI on this.

feanil (approved):

PR #3524, opened 2016-12-05
ECOM-6463: correct embargo URL

maxrothman:


jibsheet:
@adampalay I saw you were chatting to @fredsmith about this change, so I'll ask you to review, possibly with Fred if he has more context.

adampalay:
cool! 👍  from me.

@fredsmith and I were also talking about just having `/embargo` load the view that `/embargo/blocked-message/courseware/embargo/` loads, but this totally does the job too!

adampalay (approved):

jibsheet:
If it's easy to just grab /embargo in a urls.py, I'd love that and we can fix back after, but this seems like a quick win to give people something more than a 404

adampalay:
@jibsheet , it looks like it will be doable but not _super_ straightforward.

fredsmith (approved):

PR #3530, opened 2016-12-08
Make the configuration dir if it's not there already.

benpatterson:
Configuration Pull Request
---

Text of PR:

I encountered the following error on a Xenial box which came with git 2.7.4 installed.

```
+ PATH=/tmp/bootstrap/bin:/usr/local/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin
+ git clone /tmp/configuration
fatal: repository '/tmp/configuration' does not exist
```

This change ensures that does not happen.

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -142,6 +142,7 @@ PATH="${PYTHON_BIN}":${PATH}
 
 # Install the configuration repository to install 
 # edx_ansible role
+mkdir -p ${CONFIGURATION_DIR}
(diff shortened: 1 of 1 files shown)


benpatterson:
nope. not what i needed.

benpatterson:
for anyone curious, i ran the bootstrap script using shell. I ignorantly thought that it would pick up the `#!/usr/bin/env bash` setting. It was important for git clone because it did not read the `${CONFIGURATION_REPO}` variable

running `bash <ansible bootstrap.sh>` resolved the problem

PR #3541, opened 2016-12-10
Database parameter overrides for analytics.

macdiesel:
@feanil Could I please ask you to create the overrides for the new variables in this file?

Thanks


feanil:
The insights changes should go here: https://github.com/edx/configuration/blob/master/playbooks/roles/insights/defaults/main.yml#L63 as well.

macdiesel:
@feanil good catch. Should we do anything about those docker overrides?

feanil:
I think the docker overrides are fine as they are right now.

feanil (approved):

pomegranited:
Hey @macdiesel and @feanil CC @mulby @bdero - it looks like the [`INSIGHTS_DATABASE_HOST` change](https://github.com/edx/configuration/pull/3541/commits/b5ba71cb9c7c86f09e20f04c90221f43d6097720#diff-a665eb3a7388490059d7b0d1d18be43eR14) breaks analyticstack.

At first, ansible complains about not having a `DOCKER_TLD` variable defined, but really, the database host should be `localhost` for the analytics devstack.

We can fix this by patching the [analyticstack/Vagrantfile](https://github.com/edx/configuration/blob/master/vagrant/release/analyticstack/Vagrantfile) to pass `INSIGHTS_DATABASE_HOST` as an extra variable, but is it seems like the Insights playbook shouldn't depend on docker variables?

playbooks/roles/insights/defaults/main.yml
```diff
@@ -61,15 +61,20 @@ INSIGHTS_LEARNER_API_LIST_DOWNLOAD_FIELDS: !!null
 
 INSIGHTS_DATABASE_NAME: 'dashboard'
 
+INSIGHTS_DATABASE_USER: rosencrantz
+INSIGHTS_DATABASE_PASSWORD: secret
+INSIGHTS_DATABASE_HOST: "db.{{ DOCKER_TLD  }}"
```
macdiesel, line 66: @pomegranited should we just change this to be localhost?

pomegranited:
@macdiesel That'd be great, if it doesn't mess up your docker stuff.  Specifically, the original value was [`127.0.0.1`](https://github.com/edx/configuration/pull/3541/files#diff-d5a061c67ccc49b36d6ed7a21101e215L25).

playbooks/roles/insights/defaults/main.yml
pomegranited: @macdiesel That'd be great, if it doesn't mess up your docker stuff.  Specifically, the original value was [`127.0.0.1`](https://github.com/edx/configuration/pull/3541/files#diff-d5a061c67ccc49b36d6ed7a21101e215L25).

PR #3547, opened 2016-12-12
Add role for installing Docker tools

gsong:
@nedbat This adds a `docker-tools` Ansible role that you can include in the `edx_sandbox` playbook.

### DIFF
--- a/docker/build/docker-tools/Dockerfile
+++ b/docker/build/docker-tools/Dockerfile
@@ -18,6 +18,8 @@ ADD . /edx/app/edx_ansible/edx_ansible
 WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
 
 COPY docker/build/docker-tools/ansible_overrides.yml /
-RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook docker-tools.yml \
+RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook docker-tools.yml \
     -c local -i '127.0.0.1,' \
     -t 'install'
+RUN which docker
+RUN which docker-compose
--- a/docker/build/docker-tools/Dockerfile
+++ b/docker/build/docker-tools/Dockerfile
@@ -1,0 +1,23 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/docker-tools/Dockerfile .
+#
+# This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
+# with the currently checked-out configuration repo.
+
+FROM edxops/xenial-common:latest
+MAINTAINER edxops
+
+ENV PROGRAMS_VERSION=master
+ENV REPO_OWNER=edx
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+COPY docker/build/docker-tools/ansible_overrides.yml /
+RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook docker-tools.yml \
+    -c local -i '127.0.0.1,' \
+    -t 'install'
--- a/playbooks/roles/docker-tools/defaults/main.yml
+++ b/playbooks/roles/docker-tools/defaults/main.yml
@@ -1,0 +1,13 @@
+cache_valid_time: 3600
+
+docker_tools_deps_deb_pkgs:
+    - apt-transport-https
+    - ca-certificates
+    - python-pip
+
+docker_apt_keyserver: "hkp://ha.pool.sks-keyservers.net:80"
+docker_apt_key_id: "58118E89F3A912897C070ADBF76221572C52609D"
+docker_repo: "deb https://apt.dockerproject.org/repo ubuntu-xenial main"
+
+docker_group: "docker"
+docker_users: []
--- a/playbooks/roles/docker-tools/tasks/main.yml
+++ b/playbooks/roles/docker-tools/tasks/main.yml
@@ -1,0 +1,44 @@
+# Install docker-engine and docker-compose
+# Add users to docker group
+---
+- name: add docker group
+  group:
+    name: "{{ docker_group }}"
+
+- name: add users to docker group
+  user:
+    name: "{{ item }}"
+    groups: "{{ docker_group }}"
+    append: yes
+  with_items: "{{ docker_users }}"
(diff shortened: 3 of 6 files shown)


playbooks/roles/docker-tools/tasks/main.yml
```diff
@@ -0,0 +1,42 @@
...
+    state: started
+
+- name: install docker-compose
+  pip:
+    name: "docker-compose"
+
+- name: add docker group
+  group:
+    name: "{{ docker_group }}"
+
```
jibsheet: In my testing last week for some hackathon work, I had to make the group and then start (or restart) the docker service so that the socket file would work properly.  If this works for you, great, but I'm a little surprised.

playbooks/roles/docker-tools/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
...
+    - python-pip
+
+docker_apt_keyserver: "hkp://ha.pool.sks-keyservers.net:80"
+docker_apt_key_id: "58118E89F3A912897C070ADBF76221572C52609D"
+docker_repo: "deb https://apt.dockerproject.org/repo ubuntu-xenial main"
+
+docker_group: "docker"
+docker_users:
+    - edxapp
+    - vagrant
```
jibsheet: I think in the role this should be an empty list and that the playbook should pass these users.
I could for example see another role which I've half written for metrics collection using this, but passing in a metrics user.

jibsheet (commented):
Looks great, but 2 things I noticed.

playbooks/roles/docker-tools/defaults/main.yml
gsong: @jibsheet Sounds like a plan. I'll move the variable up to the playbook level.

playbooks/roles/docker-tools/tasks/main.yml
gsong: @jibsheet I tried this quite a few times and didn't have problems with it. I'll verify again to make sure.

playbooks/roles/docker-tools/defaults/main.yml
gsong: @jibsheet Just to be sure, I'll move the tasks farther up the chain.

playbooks/roles/docker-tools/tasks/main.yml
gsong: @jibsheet Just to be sure, I'll move the tasks farther up the chain.

playbooks/roles/docker-tools/defaults/main.yml
```diff
@@ -0,0 +1,13 @@
...
+    - apt-transport-https
+    - ca-certificates
+    - python-pip
+
+docker_apt_keyserver: "hkp://ha.pool.sks-keyservers.net:80"
+docker_apt_key_id: "58118E89F3A912897C070ADBF76221572C52609D"
+docker_repo: "deb https://apt.dockerproject.org/repo ubuntu-xenial main"
+
+docker_group: "docker"
+docker_users: []
```
gsong, line 13: @jibsheet Per your recommendation, I’ve set `docker_users` to an empty list at the role level. @nedbat This means you’ll have to add appropriate user(s) at the playbook level. I’m pretty sure the only user who needs to run docker commands right now would be `vagrant`.

playbooks/roles/docker-tools/tasks/main.yml
```diff
@@ -0,0 +1,44 @@
+# Install docker-engine and docker-compose
+# Add users to docker group
+---
+- name: add docker group
```
gsong, line 4: @jibsheet I’ve moved group creation and user assignment to be the first tasks.

jibsheet (commented):
Looks good - although do we want to add a quick docker test to this so that CD (Travis) will test changes to the role?  This should be straightforward to do - I'll push you a sample commit.

The other quick change we should do is to add the task tags:
https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle+Tags

jibsheet:
@gsong a1a9484 is my push to get travis to run tests on this role.  Ran locally, may need to weak a bit for this in travis.  Oh - I wonder if Travis isn't running anything because this is set to merge into Ned's branch, not master.. Hmm, that's annoying.

docker/build/docker-tools/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+MAINTAINER edxops
+
+ENV PROGRAMS_VERSION=master
+ENV REPO_OWNER=edx
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+COPY docker/build/docker-tools/ansible_overrides.yml /
```
gsong, line 20: Did you mean to check in `ansible_overrides.yml`? Or is this a developer specific file?

jibsheet: Pushed up.  It's just a stub for this role.

jibsheet:
I pushed up an updated Dockerfile and the overrides.
Do you want to retarget this PR at master, I think this would be fine to land into master and then rebase the other work on top of that.  It'll also give us a chance to see the travis tests run.

Also - feel free to squash my commits down into one, I didn't want to rebase anything in case you were still working on something.

gsong:
@jibsheet I changed the PR target to `master` but it doesn’t seem to be kicking off Travis CI.

jibsheet:
@gsong I deployed the hammer of open/close and now travis is going.

gsong:
@jibsheet Travis says ✅

jibsheet:
@gsong Can you trim off the first two commits of the branch (Ned's devstack work) so we can keep this as a separate thing?  After that we should be able to merge in the morning (I'll be in late, but will touch base).

gsong:
@jibsheet OK, I’ve rebased just the docker-tools changes on top of `master`.

jibsheet:
@gsong I rebased onto master (Feanil just upgraded us to Ansible 2.x) and will let tests run (I also squished my 3 commits into 1)

jibsheet:
@gsong Ok, we found a bug in Feanil's Ansible 2.0 merge, fixed that, rebased again, and now I'll get someone to quickly look at my Dockerfile changes and then we'll be good to merge when Travis finishes.

feanil (approved):

jibsheet:
@gsong I'll leave this for you to merge - ship when ready.  Thanks for your help reorganizing and getting it worked out to go into master.

PR #3408, opened 2016-10-07
Max/one nginx to rule them all

maxrothman:
This is still in-progress, but could use some directional review.

PR #3549, opened 2016-12-13
Added common play for Django-based IDAs

clintonb:
ECOM-6600

### DIFF
--- a/playbooks/roles/edx_django_service/defaults/main.yml
+++ b/playbooks/roles/edx_django_service/defaults/main.yml
@@ -1,0 +1,183 @@
+---
+edx_django_service_name: ' NOT-SET '
+edx_django_service_repo: '{{ edx_django_service_name }}'
+edx_django_service_home: '{{ COMMON_APP_DIR }}/{{ edx_django_service_name }}'
+edx_django_service_user: '{{ edx_django_service_name }}'
+edx_django_service_use_python3: true
+
+# This should be overwritten at the time Ansible is run.
+edx_django_service_is_devstack: false
+
+edx_django_service_has_static_assets: true
+
+edx_django_service_wsgi_name: '{{ edx_django_service_name }}'
+
+edx_django_service_name_devstack_logs:
+  - '{{ supervisor_log_dir }}/{{ edx_django_service_name }}-stdout.log'
+  - '{{ supervisor_log_dir }}/{{ edx_django_service_name }}-stderr.log'
+
+
+edx_django_service_version: 'master'
+edx_django_service_git_identity: null
+edx_django_service_django_settings_module: null
+
+edx_django_service_code_dir: '{{ edx_django_service_home }}/{{ edx_django_service_name }}'
+edx_django_service_venv_dir: '{{ edx_django_service_home }}/venvs/{{ edx_django_service_name }}'
+edx_django_service_venv_bin_dir: '{{ edx_django_service_venv_dir }}/bin'
+
+edx_django_service_nodeenv_dir: '{{ edx_django_service_home }}/nodeenvs/{{ edx_django_service_name }}'
+edx_django_service_nodeenv_bin: '{{ edx_django_service_nodeenv_dir }}/bin'
+edx_django_service_node_modules_dir: '{{ edx_django_service_code_dir }}/node_modules'
+edx_django_service_node_bin: '{{ edx_django_service_node_modules_dir }}/.bin'
+edx_django_service_node_version: '{{ common_node_version }}'
+
+edx_django_service_environment_default:
+  DJANGO_SETTINGS_MODULE: '{{ edx_django_service_django_settings_module }}'
+  PATH: '{{ edx_django_service_nodeenv_bin }}:{{ edx_django_service_venv_dir }}/bin:{{ ansible_env.PATH }}'
+edx_django_service_environment_extra: {}
+edx_django_service_environment: '{{ edx_django_service_environment_default | combine(edx_django_service_environment_extra) }}'
+
+edx_django_service_migration_environment_default:
+  DB_MIGRATION_USER: '{{ COMMON_MYSQL_MIGRATE_USER }}'
+  DB_MIGRATION_PASS: '{{ COMMON_MYSQL_MIGRATE_PASS }}'
+edx_django_service_migration_environment: '{{ edx_django_service_environment|combine(edx_django_service_migration_environment_default) }}'
+
+edx_django_service_debian_pkgs_default:
+  - gettext
+  - libffi-dev  # Needed to install the Python cryptography library for asymmetric JWT signing
+  - libmemcached-dev
+  - libmysqlclient-dev
+  - libssl-dev
+edx_django_service_debian_pkgs_extra: []
+edx_django_service_debian_pkgs: '{{ edx_django_service_debian_pkgs_default + edx_django_service_debian_pkgs_extra }}'
+
+edx_django_service_gunicorn_extra: ''
+edx_django_service_gunicorn_extra_conf: ''
+edx_django_service_gunicorn_host: '127.0.0.1'
+edx_django_service_gunicorn_port: null
+edx_django_service_gunicorn_timeout: 300
+edx_django_service_gunicorn_workers: 2
+edx_django_service_gunicorn_worker_class: 'gevent'
+
+edx_django_service_cors_whitelist: []
+
+nginx_edx_django_service_gunicorn_hosts:
+  - 127.0.0.1
+
+edx_django_service_hostname: '~^((stage|prod)-)?{{ edx_django_service_name }}.*'
+edx_django_service_nginx_port: '1{{ edx_django_service_gunicorn_port }}'
+edx_django_service_ssl_nginx_port: '4{{ edx_django_service_gunicorn_port }}'
+edx_django_service_enable_basic_auth: false
+edx_django_service_basic_auth_exempted_paths_default:
+  - api
+edx_django_service_basic_auth_exempted_paths_extra: []
+edx_django_service_basic_auth_exempted_paths: '{{ edx_django_service_basic_auth_exempted_paths_default + edx_django_service_basic_auth_exempted_paths_extra }}'
+
+edx_django_service_newrelic_appname: '{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-{{ edx_django_service_name }}'
+
+edx_django_service_repos:
+  - PROTOCOL: '{{ COMMON_GIT_PROTOCOL }}'
+    DOMAIN: '{{ COMMON_GIT_MIRROR }}'
+    PATH: '{{ COMMON_GIT_PATH }}'
+    REPO: '{{ edx_django_service_repo }}.git'
+    VERSION: '{{ edx_django_service_version }}'
+    DESTINATION: '{{ edx_django_service_code_dir }}'
+    SSH_KEY: '{{ edx_django_service_git_identity }}'
+
+edx_django_service_secret_key: null
+edx_django_service_language_code: 'en-us'
+
+edx_django_service_data_dir: '{{ COMMON_DATA_DIR }}/{{ edx_django_service_name }}'
+edx_django_service_media_root: '{{ edx_django_service_data_dir }}/media'
+edx_django_service_media_url: '/media/'
+
+edx_django_service_static_root: '{{ COMMON_DATA_DIR }}/{{ edx_django_service_name }}/staticfiles'
+edx_django_service_staticfiles_storage: 'django.contrib.staticfiles.storage.StaticFilesStorage'
+
+edx_django_service_media_storage_backend:
+  DEFAULT_FILE_STORAGE: 'django.core.files.storage.FileSystemStorage'
+  MEDIA_ROOT: '{{ edx_django_service_media_root }}'
+  MEDIA_URL: '{{ edx_django_service_media_url }}'
+
+edx_django_service_memcache: [ 'memcache' ]
+
+edx_django_service_caches:
+  default:
+    BACKEND: 'django.core.cache.backends.memcached.MemcachedCache'
+    KEY_PREFIX: '{{ edx_django_service_name }}'
+    LOCATION: '{{ edx_django_service_memcache }}'
+
+edx_django_service_default_db_host: 'localhost'
+edx_django_service_default_db_name: '{{ edx_django_service_name }}'
+edx_django_service_default_db_atomic_requests: false
+edx_django_service_db_user: 'REPLACE-ME'
+edx_django_service_db_password: 'password'
+
+edx_django_service_databases:
+  default:
+    ENGINE: 'django.db.backends.mysql'
+    NAME: '{{ edx_django_service_default_db_name }}'
+    USER: '{{ edx_django_service_db_user }}'
+    PASSWORD: '{{ edx_django_service_db_password }}'
+    HOST: '{{ edx_django_service_default_db_host }}'
+    PORT: '3306'
+    ATOMIC_REQUESTS: '{{ edx_django_service_default_db_atomic_requests }}'
+    CONN_MAX_AGE: 60
+
+edx_django_service_social_auth_edx_oidc_key: '{{ edx_django_service_name }}-key'
+edx_django_service_social_auth_edx_oidc_secret: '{{ edx_django_service_name }}-secret'
+edx_django_service_social_auth_redirect_is_https: false
+
+edx_django_service_oauth_public_url_root: '{{ EDXAPP_LMS_PUBLIC_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
+edx_django_service_oauth_url_root: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
+edx_django_service_oidc_logout_url: '{{ EDXAPP_LMS_PUBLIC_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
+edx_django_service_oidc_issuer: '{{ edx_django_service_oauth_url_root }}'
+
+edx_django_service_jwt_audience: '{{ EDXAPP_JWT_AUDIENCE | default("SET-ME-PLEASE") }}'
+edx_django_service_jwt_issuer: '{{ edx_django_service_oauth_url_root }}'
+edx_django_service_jwt_secret_key: '{{ EDXAPP_JWT_SECRET_KEY | default("lms-secret") }}'
+
+edx_django_service_session_expire_at_browser_close: false
+
+edx_django_service_jwt_auth:
+  JWT_ISSUERS:
+    - AUDIENCE: '{{ edx_django_service_jwt_audience }}'
+      ISSUER: '{{ edx_django_service_jwt_issuer }}'
+      SECRET_KEY: '{{ edx_django_service_jwt_secret_key }}'
+
+edx_django_service_extra_apps: []
+
+edx_django_service_config_default:
+  LANGUAGE_CODE: '{{ edx_django_service_language_code }}'
+  SECRET_KEY: '{{ edx_django_service_secret_key }}'
+  TIME_ZONE: 'UTC'
+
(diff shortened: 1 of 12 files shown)


clintonb:
@edx/devops thoughts on this?
FYI @edx/ecommerce

clintonb:
Examples of this configuration in action are available at #3550.

cpennington:
How does this relate to `ansible-role-django-ida`?

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,260 @@
...
+# Dependencies:
+#
+#
+# Example play:
+#
+#
+
+- name: set devstack fact
+  set_fact:
+    edx_django_service_is_devstack: true
```
maxrothman: Shouldn't this be the value of the `edx_django_service_is_devstack` var you have in defaults.yml?

jibsheet:
Is the goal here to simplify the interface for edx_service?  Is there code changes we should be pushing into edx_service itself instead?

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,260 @@
...
+- name: install python3.5
+  apt:
+    name: "{{ item }}"
+  with_items:
+    - python3.5
+    - python3.5-dev
+  when: edx_django_service_use_python35
+  tags:
+    - install
+    - install:system-requirements
```
maxrothman: I'm not sure about this approach. In abbey.py, we simply install the standard python3-pip and python3-dev packages, which don't make backwards-incompatible changes on Ubuntu LTS's (unless it's necessary for a security patch). And if we're not doing the deadsnakes thing we can just use the existing apt_packages mechanism.

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,260 @@
...
+  template:
+    src: "edx/app/nginx/sites-available/app.j2"
+    dest: "{{ nginx_sites_available_dir }}/{{ edx_django_service_name }}"
+    owner: root
+    group: "{{ common_web_user }}"
+    mode: 0640
+  notify: reload nginx
+  tags:
+    - install
+    - install:vhosts
```
maxrothman, line 238: Having a single template for all IDAs is going to be tricky (I know because I've tried). Almost all IDAs customize the template in some way, so I'm hesitant to straightjacket future ones.

maxrothman (commented):
Thanks for factoring all of this stuff out, it'll be really good to stop repeating all that code.

Out of curiosity, was there a reason you didn't modify the existing edx_service role?

clintonb:
@cpennington Once the refactoring is approved/close to finalization, `ansible-role-django-ida` will be updated to rely on the new role (or the existing `edx_service` role if this is all munged together).

clintonb:
@jibsheet @maxrothman I wasn't aware of what else depended on `edx_service`, so I opted to make a new role and modify the ECOM services with which I am most familiar. My initial goal was to keep the scope as low as possible for the Hackathon.

Should this all be merged into `edx_service`? There are some unfamiliar (to me) roles that rely on `edx_service`:

* harstorage
* locust
* xqwatcher

playbooks/roles/edx_django_service/tasks/main.yml
clintonb: I would like to avoid more vars, if possible, by relying on the tag.

playbooks/roles/edx_django_service/tasks/main.yml
clintonb: Install Python 3.5 for all? Fine with me.

playbooks/roles/edx_django_service/tasks/main.yml
clintonb: Good point. I'll spend some time diff-ing, and determining if we can add a hook to insert extra configuration. My preferences:

1. Use the default.
2. Add extra configuration to the default (via YAML dict or raw string?).
3. Override the file entirely.

maxrothman:
The branch I mentioned is in #3408

playbooks/roles/edx_django_service/tasks/main.yml
maxrothman: I'm not sure I'm comfortable with the default case being devstack.

clintonb: This step is only run if you use the devstack or devstack:install tags.

clintonb:
@jibsheet @maxrothman I took a look at the non-IDA roles that rely on `edx_service`. Given that they aren't Django projects, it doesn't make sense to merge my changes into `edx_service`. Doing so would mean we would have to add conditionals to disable certain tasks (e.g. running migrations, compiling assets).

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
```diff
@@ -0,0 +1,80 @@
...
+{% else %}
+  {% set default_site = "" %}
+{% endif %}
+
+upstream {{ edx_django_service_name }}_app_server {
+{% for host in nginx_edx_django_service_gunicorn_hosts %}
+    server {{ host }}:{{ edx_django_service_gunicorn_port }} fail_timeout=0;
+{% endfor %}
+}
+
```
clintonb, line 16: TODO: Add support for CORS header

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
```diff
@@ -0,0 +1,80 @@
...
+    try_files /staticfiles/$file =404;
+  }
+
+  location ~ ^/media/(?P<file>.*) {
+    root {{ COMMON_DATA_DIR }}/{{ edx_django_service_name }};
+    try_files /media/$file =404;
+  }
+
+  location / {
+    try_files $uri @proxy_to_app;
```
clintonb: TODO: Restore basic auth

`{% include "basic-auth.j2" %}`

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
```diff
@@ -0,0 +1,80 @@
...
+  }
+
```
clintonb: TODO: Disable basic auth for the `/api` path.

```
# The API should be secured with OAuth 2.0 or or JWT.
location /api {
  try_files $uri @proxy_to_app;
}
```

We might also consider a list of exempted paths. For example, ecommerce needs the `/payment` path publicly available as well.

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
```diff
@@ -0,0 +1,80 @@
...
+  {% if NGINX_ROBOT_RULES|length > 0 %}
```
clintonb: TODO: Consider using `robots.j2` in the nginx play.

clintonb:
@maxrothman I identified a few areas for improvement on the nginx configuration file. All of the IDAs, with the exception of edxapp, follow the same pattern. The only deviations are the disabling of basic auth for certain paths and usage of CORS, both of which can be abstracted in a manner suitable for our current usage.

playbooks/roles/edx_django_service/tasks/main.yml
maxrothman: There is no tag filtering done by default, thus if you run without any tags options, you'll run this task, thus the default case would be devstack.

jibsheet:
@clintonb we have a large nginx refactor we'd like to do (not currently scheduled) but I agree about the needless repetition there.

We put a review of this in for (we think) next sprint OPS-2093

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,177 @@
...
+edx_django_service_name: ' NOT-SET '
+edx_django_service_repo: '{{ edx_django_service_name }}'
+edx_django_service_home: '{{ COMMON_APP_DIR }}/{{ edx_django_service_name }}'
+edx_django_service_user: '{{ edx_django_service_name }}'
+edx_django_service_use_python3: true
+
+# This can be overwritten by setting either the devstack or devstack:install tag.
+edx_django_service_is_devstack: false
+
+edx_django_service_has_static_assets: true
```
feanil: Shouldn't this be false by default? Or were you thinking true since most apps have assets?

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,177 @@
...
+  - '{{ nginx_log_dir }}/access.log'
+  - '{{ nginx_log_dir }}/error.log'
+  - '{{ supervisor_log_dir }}/{{ edx_django_service_name }}-stdout.log'
+  - '{{ supervisor_log_dir }}/{{ edx_django_service_name }}-stderr.log'
+
+
+# TODO Override with global
+edx_django_service_version: 'master'
+edx_django_service_git_identity: !!null
+edx_django_service_django_settings_module: null
```
feanil: Why use two different values of null here and in the line above it?

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,177 @@
...
+edx_django_service_debian_pkgs_default:
+  - gettext
+  - libffi-dev  # Needed to install the Python cryptography library for asymmetric JWT signing
+  - libmemcached-dev
+  - libmysqlclient-dev
+  - libssl-dev
+edx_django_service_debian_pkgs_extra: []
+edx_django_service_debian_pkgs: '{{ edx_django_service_debian_pkgs_default + edx_django_service_debian_pkgs_extra }}'
+
+# TODO These should be overridden as globals.
```
feanil: Does that mean these should be all_caps_vars, or do you mean something else by globals?

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,177 @@
...
+edx_django_service_gunicorn_extra: ''
+edx_django_service_gunicorn_extra_conf: ''
+edx_django_service_gunicorn_host: '127.0.0.1'
+edx_django_service_gunicorn_port: null
+edx_django_service_gunicorn_timeout: 300
+edx_django_service_gunicorn_workers: 2
+edx_django_service_gunicorn_worker_class: 'gevent'
+
+# TODO Sane default?
+edx_django_service_cors_whitelist: []
```
feanil: If this means that CORS is off by default then yes, sane default.

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,177 @@
...
+edx_django_service_repos:
+  - PROTOCOL: '{{ COMMON_GIT_PROTOCOL }}'
+    DOMAIN: '{{ COMMON_GIT_MIRROR }}'
+    PATH: '{{ COMMON_GIT_PATH }}'
+    REPO: '{{ edx_django_service_repo }}.git'
+    VERSION: '{{ edx_django_service_version }}'
+    DESTINATION: '{{ edx_django_service_code_dir }}'
+    SSH_KEY: '{{ edx_django_service_git_identity }}'
+
+edx_django_service_secret_key: 'REPLACE-ME'
```
feanil: REPLACE-ME is a valid key, we should set this to null since modern djangos will throw an error if the key is not set.

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,252 @@
...
+- name: build virtualenv with python2.7
+  command: "virtualenv --python=python2.7 {{ edx_django_service_venv_dir }}"
+  become_user: "{{ edx_django_service_user }}"
+  when: not edx_django_service_use_python3
+  tags:
+    - install
+    - install:system-requirements
+
+# NOTE (CCB): Ideally we should use the pip Ansible command,
+# but that doesn't seem to work with the Python 3.x virtualenv.
```
feanil, line 65: Is this still true?

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,252 @@
...
+  command: make production-requirements
+  args:
+    chdir: "{{ edx_django_service_code_dir }}"
+  become_user: "{{ edx_django_service_user }}"
+  environment: "{{ edx_django_service_environment }}"
+  tags:
+    - install
+    - install:app-requirements
+
+# FIXME Only run for devstack
```
feanil: FIXME

playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -78,17 +77,18 @@
...
-    path: "{{ item }}"
-    state: directory
-    owner: "syslog"
-    group: "syslog"
-  with_items:
-    - "{{ COMMON_LOG_DIR }}/{{ edx_service_name }}"
-  tags:
-    - install
-    - install:base
+# TODO Determine if any services actually use this directory. The IDAs do not.
```
feanil: I believe these are used by the IDAs in their production logging config.

feanil (commented):
Before this can land, we'll also need to update the pipelines to make sure that runs of the IDA playbooks that use this role will skip various tags. migrate, and devstack for sure but there may be others.

playbooks/roles/edx_django_service/tasks/main.yml
feanil: Instead of setting this here, let's set it on the commandline and not be in devstack by default.

playbooks/roles/edx_django_service/defaults/main.yml
clintonb: At this point, the only IDA without static assets might be analytics-data-api.

playbooks/roles/edx_django_service/defaults/main.yml
clintonb: Fixed. Using `null`.

playbooks/roles/edx_django_service/defaults/main.yml
clintonb: Yes, all-caps. This comment has been removed.

playbooks/roles/edx_django_service/defaults/main.yml
clintonb: Great. Removed the TODO.

playbooks/roles/edx_django_service/defaults/main.yml
clintonb: Done.

playbooks/roles/edx_django_service/tasks/main.yml
clintonb: Yes.

playbooks/roles/edx_django_service/tasks/main.yml
clintonb: Okay. Okay. Geez!

playbooks/roles/edx_django_service/tasks/main.yml
clintonb: Great idea! Done.

playbooks/roles/edx_service/tasks/main.yml
clintonb: Reverted.

clintonb:
@edx/devops ready for review!

clintonb:
@edx/devops can I get a review, please?

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,185 @@
...
+# This should be overwritten at the time Ansible is run.
+edx_django_service_is_devstack: false
+
+edx_django_service_has_static_assets: true
+
+edx_django_service_wsgi_name: '{{ edx_django_service_name }}'
+
+edx_django_service_name_devstack_logs:
+  - '{{ nginx_log_dir }}/access.log'
+  - '{{ nginx_log_dir }}/error.log'
```
jibsheet: We ran into issues on vagrant devstack with a similar stanza
https://github.com/edx/configuration/pull/3814

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,185 @@
...
+
+edx_django_service_code_dir: '{{ edx_django_service_home }}/{{ edx_django_service_name }}'
+edx_django_service_venv_dir: '{{ edx_django_service_home }}/venvs/{{ edx_django_service_name }}'
+edx_django_service_venv_bin_dir: '{{ edx_django_service_venv_dir }}/bin'
+
+edx_django_service_nodeenv_dir: '{{ edx_django_service_home }}/nodeenvs/{{ edx_django_service_name }}'
+edx_django_service_nodeenv_bin: '{{ edx_django_service_nodeenv_dir }}/bin'
+edx_django_service_node_modules_dir: '{{ edx_django_service_code_dir }}/node_modules'
+edx_django_service_node_bin: '{{ edx_django_service_node_modules_dir }}/.bin'
+edx_django_service_node_version: '6.9.4'
```
jibsheet: Can we use '{{ common_node_version }}'

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -0,0 +1,185 @@
...
+  DB_MIGRATION_USER: '{{ COMMON_MYSQL_MIGRATE_USER }}'
+  DB_MIGRATION_PASS: '{{ COMMON_MYSQL_MIGRATE_PASS }}'
+edx_django_service_migration_environment: '{{ edx_django_service_environment|combine(edx_django_service_migration_environment_default) }}'
+
+edx_django_service_debian_pkgs_default:
+  - gettext
+  - libffi-dev  # Needed to install the Python cryptography library for asymmetric JWT signing
+  - libmemcached-dev
+  - libmysqlclient-dev
+  - libssl-dev
```
jibsheet, line 50: I wonder at what point these belong even in common (we needed them for the notifier recently) 
I don't think it's worth changing here

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,245 @@
...
+    src: "edx/app/app/app_gunicorn.py.j2"
+    dest: "{{ edx_django_service_home }}/{{ edx_django_service_name }}_gunicorn.py"
+  become_user: "{{ edx_django_service_user }}"
+  tags:
+    - install
+    - install:configuration
+
+- name: install python3
+  apt:
+    name: "{{ item }}"
```
jibsheet, line 35: Does this want a when: edx_django_service_use_python3?

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -0,0 +1,245 @@
...
+- name: build virtualenv with python3
+  command: "virtualenv --python=python3 {{ edx_django_service_venv_dir }}"
+  become_user: "{{ edx_django_service_user }}"
+  when: edx_django_service_use_python3
+  tags:
+    - install
+    - install:system-requirements
+
+- name: build virtualenv with python2.7
+  command: "virtualenv --python=python2.7 {{ edx_django_service_venv_dir }}"
```
jibsheet, line 55: Won't this blow up if we run it twice on the same directory (as happens in AMI builds when it reuses an old AMI?)
I feel like we generally use 
args:
  create: pip
as our ripcord for that

jibsheet (commented):
Looks good, just had a few questions / things

@fredsmith did you still want to review?

clintonb:
@jibsheet thanks for the feedback.
@fredsmith let me know if you'd like to take a look.

@edx/devops thoughts on the manual test strategy below? Is there any testing that can/should be automated?

1. Build sandbox with ecommerce and credentials. Confirm basic app functionality works, and logs in correct locations.
2. Deploy to loadtest environment for ecommerce, credentials, and discovery. Confirm basic app functionality works via loadtests. Confirm Splunk logging works as expected.

feanil:
That seems reasonable to me.

clintonb:
Sandbox validation completed.

feanil (approved):

PR #3554, opened 2016-12-13
Install a version of pip and setup tools that ansible 2.x is happy with.

feanil:
@doctoryes  @macdiesel


doctoryes:
Container built successfully & installed the updated versions of pip/setuptools.

👍

PR #3555, opened 2016-12-13
Using the correct Django settings file for the programs Docker image

clintonb:
ECOM-6570


clintonb:
@jibsheet this should resolve the broken build for #3544. The devstack settings do _NOT_ use syslog (since the service is not running), which is one of the default log handlers.

https://github.com/edx/programs/blob/master/programs/settings/devstack.py#L6-L8

jibsheet (approved):

jibsheet:
Cool - that's what I assumed it was (wrong settings file)

PR #3559, opened 2016-12-14
remove ec2_ami_2_0_0_1 module so we begin using the one built in

macdiesel:
@edx/pipeline-team PTAL. Fixes https://openedx.atlassian.net/browse/TE-1851

Yes this was the only place it was used.


doctoryes:
Full of win 👍

PR #3561, opened 2016-12-14
Set virtualbox network to virtio for speed

nedbat:
This made installing the devstack go about 20% faster for me.


nedbat:
@edx/devops @gsong OK with this?

gsong (approved):
Sounds good to me.

jibsheet (approved):
Anything to make vagrant less slow.

PR #3565, opened 2016-12-14
Update bootstrap to be able to run a subset and run a subset when launching an instance.

feanil:
Configuration Pull Request
---

@edx/devops We need something like this to be able to build sandboxes from scratch on xenial.

### DIFF
--- a/playbooks/edx-east/edx_provision.yml
+++ b/playbooks/edx-east/edx_provision.yml
@@ -22,6 +22,13 @@
     name_tag: sandbox-temp
     elb: false
     ec2_vpc_subnet_id: subnet-cd867aba
+    instance_userdata: |
+      #!/bin/bash
+      set -x
+      set -e
+      export RUN_ANSIBLE=false;
+      wget https://raw.githubusercontent.com/edx/configuration/{{ configuration_version }}/util/install/ansible-bootstrap.sh -O - | bash;
+    launch_wait_time: 5
   roles:
     - role: launch_ec2
       keypair: "{{ keypair }}"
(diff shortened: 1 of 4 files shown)


playbooks/edx-east/edx_provision.yml
```diff
@@ -22,6 +22,10 @@
     name_tag: sandbox-temp
     elb: false
     ec2_vpc_subnet_id: subnet-cd867aba
+    bootstrap_userdata: >
+        #!/bin/env bash -xe
+        export RUN_ANSIBLE=true;
+        wget https://raw.githubusercontent.com/edx/configuration/feanil/sandbox_16.04/util/install/ansible-bootstrap.sh -O - | sudo bash;
```
feanil: TODO: The branch name here will need to change to the branch of config that the ansible_provision job is running from, and will default to master.

feanil:
@edx/devops http://jenkins.edx.org:8080/job/ansible-provision-ansible-2/29/console

Successful build from scratch from this branch.

playbooks/roles/launch_ec2/tasks/main.yml
```diff
@@ -95,3 +96,7 @@
     port: 22
     delay: 10
   with_items: "{{ ec2.instances }}"
+
+- name: Wait for python to install
+  pause:
+    minutes: "{{ launch_ec2_wait_time }}"
```
fredsmith, line 102: It looks like this whole role has no default values.  Is that what we want?  Seems like it could/should intuit some defaults from common vars

fredsmith (approved):
Looks good, but there's some existing technical debt on the launch ec2 role that we may be in a position to deal with.

playbooks/roles/launch_ec2/tasks/main.yml
feanil: That is how the rest of this has been written, the value of this comes from the edx_provision.yml playbook.

If we want to fix that we should write a separate ticket for that tech debt, but honestly I think that is superceded by fully replacing edx-provision and how we build sandboxes in general.

PR #3571, opened 2016-12-16
We don't need git_2_0_1 any more, since we are on Ansible 2 now

nedbat:


nedbat:
@edx/devops I think this is simple enough...

feanil (approved):

nedbat:
@feanil should I be concerned about the travis failures?

PR #3582, opened 2016-12-28
dockerize the analyticstack

mulby:
This pull request converts the Vagrant based analyticstack to a docker-compose based model.

It allows docker containers to be built for the following services:
* analytics_pipeline
* analytics_api
* insights
* luigid (the luigi central scheduler)

To build the containers, I ran the following commands from the root of the configuration repository with an up-to-date virtualenv activated:

```
docker build -t edxops/xenial-common:latest -f docker/build/xenial-common/Dockerfile .
docker build -t edxops/precise-common:latest -f docker/build/precise-common/Dockerfile .
docker build --build-arg ANALYTICS_PIPELINE_VERSION=gabe/central-scheduler -t edxops/analytics_pipeline:latest -f docker/build/analytics_pipeline/Dockerfile .
docker build -t edxops/insights:latest -f docker/build/insights/Dockerfile .
docker build -t edxops/analytics_api:latest -f docker/build/analytics_api/Dockerfile .
docker build -t edxops/luigid:latest -f docker/build/luigid/Dockerfile .
```

Once that's done you should be able to run:

```
git clone https://github.com/edx/edx-analytics-pipeline
git checkout gabe/central-scheduler
make dev.watch
```

This PR introduces the following patterns, some of which may be controversial or already solved in other ways that I don't know about:

1. I introduced a `devstack:start` tag that indicates tasks that should be run every time the docker container starts up. I use this tag to do stuff like change hostnames of databases and other services that may change depending on the docker-compose.yml used to deploy the service. It also does stuff like start rsyslog and initialize databases.
2. I added many of these `devstack:start` tasks to the docker-specific playbook directly since they didn't seem appropriate to run as part of the role. I also didn't want to accidentally pollute the production AMI building process with devstack specific stuff. Not sure if tags are being explicitly passed into ansible during that build process or not.
3. I got rid of the `ansible_overrides.yml` file at container build time, and instead dynamically generate one at container run time that uses values from envrionment variables to populate most of the ansible vars. This allows you to conveniently configure the container by specifying environment vars in the docker-compose.yml file.
4. I define a `docker-entrypoint.sh` file that runs the playbook using the `docker:start` tag, starts supervisor and then runs `exec "$@"` this allows you to easily customize the logs that are output and the command that is run inside the container once everything is started. Similarly, I define a default `docker-command.sh` which defines the default behavior.

Questions:
* Is the "install:system-requirements" tag appropriate for all of the Sqoop and Hive tasks? They are just CLI apps that are not neatly packaged.

Outstanding issues:
- [ ] Static assets are not handled particularly well. I managed to get Insights to live re-compile assets when pages are loaded, but they are still stored all over the place. Some in STATIC_ROOT (the dynamically compiled ones) and the rest in a subdirectory within the edx-analytics-dashboard repo. This is not ideal since it means that when we mount the edx-analytics-dashboard repo into the container it doesn't include all of the bower-fetched assets, so we have to download all of those bower-fetched assets into the repo on the host machine. Ideally, we could move the "node_modules" and "bower_components" directories *outside* of the code dir and instead store that stuff inside the container that isn't mounted over.
- [ ] When the data pipeline container spins up, it doesn't read new entrypoints from setup.cfg so if you add a new tasks module, you can't actually run it with `launch-task` until you run `make develop-local` to refresh the entrypoints.

Instructions for actually running the containers built here can be found here: https://github.com/edx/edx-analytics-pipeline/blob/5ff23e332d9ebf0dfa82b0cf32a5eb89dacfd203/README.md

This is closely related to:
- https://github.com/edx/edx-analytics-pipeline/pull/337
- https://github.com/edx/edx-analytics-dashboard/pull/598

FYI: @e0d @rlucioni @feanil @maxrothman @brianhw @HassanJaveed84 @pomegranited @bradenmacdonald @dsjen @thallada 

TODO:


docker/build/analytics_api/Dockerfile
```diff
@@ -1,14 +1,18 @@
...
-COPY docker/build/analytics_api/ansible_overrides.yml /
-RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook analytics_api.yml -i '127.0.0.1,' -c local -t "install:base,install:system-requirements,install:configuration,install:app-requirements,install:code" -e@/ansible_overrides.yml
-WORKDIR /edx/app/
-CMD ["/edx/app/supervisor/venvs/supervisor/bin/supervisord", "-n", "--configuration", "/edx/app/supervisor/supervisord.conf"]
-EXPOSE 443 80
+RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook analytics_api.yml \
+    -i '127.0.0.1,' -c local \
+    -t "install:base,install:system-requirements,install:configuration,install:app-requirements,install:code,devstack" \
+    -e "ANALYTICS_API_VERSION=$ANALYTICS_API_VERSION" \
+    -e "analytics_api_gunicorn_host=0.0.0.0"
```
mulby, line 14: I think in a docker container we always want gunicorn bound to 0.0.0.0 inside the container. Access to ports is managed by the docker port forwarding stuff, so I didn't make this configurable.

I could see us defining a set of static ansible var overrides for build time (vs. the runtime ones defined in the docker-entrypoint.sh scripts).

docker/plays/analytics_api.yml
```diff
@@ -5,7 +5,67 @@
   vars:
     serial_count: 1
   serial: "{{ serial_count }}"
+
+  pre_tasks:
+    - name: rsyslog started
```
mulby, line 10: I've found this to be somewhat unreliable. Occasionally later tasks will complain that "/dev/log" doesn't exist... has anyone seen this before? @feanil @clintonb?

docker/plays/analytics_api.yml
```diff
@@ -5,7 +5,67 @@
...
+        state: started
+      tags:
+        - devstack:start
+
+    - name: databases ready
+      wait_for:
+        host: "{{ item.HOST }}"
+        port: "{{ item.PORT }}"
+        delay: 2
+      with_items: "{{ ANALYTICS_API_DATABASES.values() }}"
```
mulby, line 22: This is a bit unfortunate, but because these tasks need to run before the `analytics_api` role (to ensure the database has been created in MySQL before we run `migrate`) this `ANALYTICS_API_DATABASES` var *must* be overridden in the runtime ansible_overrides.yml file generated by docker-entrypoint.sh.

docker/plays/insights.yml
```diff
@@ -5,6 +5,96 @@
...
+    # static files to the common data dir.
+    - name: insights can write to the data dir
+      file:
+        path: "{{ COMMON_DATA_DIR }}/{{ insights_service_name }}"
+        state: directory
+        mode: "0775"
+      tags:
+        - devstack:start
+
+    - name: set configuration vars
```
mulby: Now that we are on ansible 2, I might be able to get away with using the `combine()` filter instead of this hack.

docker/plays/insights.yml
```diff
@@ -5,6 +5,96 @@
...
+        - key: DEBUG
+          value: "true"
+        - key: ENABLE_INSECURE_STATIC_FILES
+          value: "true"
+      tags:
+        - devstack:start
+
+    # Since the docker container mounts the user's working directory as a volume, it likely doesn't contain the
+    # "node_modules" directory that is needed to run bower
+    - name: install node dependencies
```
mulby, line 74: can we move this outside of `/edx/app/edx-analytics-dashboard/analytics_dashboard`? It feels like a natural peer for `venvs`. I could imagine the directory structure looking something like this:

```
/edx/app/edx-analytics-dashboard/
    analytics_dashboard/ (where the git repo is checked out)
    venvs/ (where the python virtualenv lives)
    node_modules/ (where the node modules live)
    bower_components/ (where the bower components live)
```
I just haven't fiddled around with npm and bower to get them to write their output elsewhere. I suspect we would also need to tell Insights where these external bower_components are so that it can find them when it searches for static files.

playbooks/roles/luigid/templates/luigid.conf.j2
```diff
@@ -0,0 +1,14 @@
+description     "luigid"
```
mulby, line 1: Probably need to replace this with a systemd unit

docker/plays/insights.yml
thallada: We're planning in our next sprint to upgrade to webpack and hopefully get rid of bower in the process (installing those packages with npm instead). `node_modules` is currently in the repo root, though. It looks like there might be some somewhat hacky ways of forcing npm to install in different folder.

docker/build/analytics_api/Dockerfile
tobz: I think this is fine / accurate.

tobz (commented):
So, generally, I'm OK with most of the Ansible changes.

It does all feel ever so slightly anti-pattern-y, though.  The running of Ansible everytime the container starts, tailing logs to keep the container going instead of having supervisord/the daemon be the entrypoint.

The Vagrant setup is the Vagrant setup but it feels like we should strive for Docker containers that, when you start them, are running the intended service, and that's that.  All the Ansible bits should be baked into CI somewhere.

clintonb:
My work for a Docker-based devstack should allay @tobz's concerns: https://github.com/edx/configuration/commits/clintonb/docker-devstack

clintonb:
Also, see https://github.com/edx/configuration/pull/3549.

mulby:
@tobz, @clintonb 

For the container command, would you prefer I run supervisor in the foreground? OEP-5 was pretty clear I need to run supervisor and not gunicorn directly. That said, I want the container to output the `/edx/var/log/foo/edx.log`, not the supervisor log. The supervisor log is not terribly interesting (by comparison). This was the primary motivation for the script that waits for that file to exist and then tail it. Maybe we could make supervisor output the application log too somehow?

As for the ansible at run time... the docker paradigm is clearly configuration via environment vars or complex configuration files mounted as volumes into the container. Given that container hostnames can change depending on the compose file that is provided, I didn't want to bake some static hostnames into all of the containers. The pattern I see in other containers is environment vars that are either used by the application directly or are written out to various config files when the container starts. I feel pretty strongly opposed to hard-coding hostnames into our services that can't be changed at container run time. It hamstrings the container from being used for *anything else*. I would be open to mounting the config files at runtime... that's definitely an option, but it requires replicating the entire `insights.yml` file in every repo that uses its image in their docker-compose.yml.

Then there's the problem of initializing the various data stores (creating databases, setting up folders in HDFS etc). When a container first boots up it needs to ensure those external data stores have been initialized properly. I tried to find ways to ensure that it mostly noops if it has already initialized (CREATE DATABASE IF NOT EXISTS etc).

Finally there is the "bower install" problem, where we mount the user's working directory from the host into the container and it doesn't contain any of the bower components, so we have to run bower again to pull of that into their mounted repo.

I agree this is suboptimal, but it felt like the best of a lot of bad options.

mulby:
Also @cpennington

clintonb (commented):
Clearing my review queue.

feanil:
I think it's safe to close this as we are unlikely to merge it.

PR #3583, opened 2016-12-28
Changes for Ficus test builds

nedbat:
Configuration Pull Request
---


feanil:
@nedbat  it looks like many of these changes are 16.04 only, without any backward compatible conditionals for 12.04.  This means merging this will break current 12.04 builds (because we have not migrated EVERYTHING to 16.04) yet. Can you add the 12.04 task back in with a conditional so we can migrate incrementally.

nedbat:
@feanil I'm not sue which 12.04 task you mean? In sandbox.sh?

playbooks/roles/browsers/tasks/main.yml
```diff
@@ -1,79 +1,137 @@
...
 - assert:
     that: "phantomjs.stat.exists"
+  tags:
+    - install
+    - install:system-requirements
 
-- name: create xvfb upstart script
-  template: src=xvfb.conf.j2 dest=/etc/init/xvfb.conf owner=root group=root
+- name: create xvfb systemd service
+  template: src=xvfb.service.j2 dest=/etc/systemd/system/xvfb.service owner=root group=root
```
feanil, line 125: @nedbat This is the first one I saw: It would be nice to have both tasks and conditionalize on distribution so we don't break 12.04 when this lands.

nedbat: @gsong Can you make a pull request into this branch with the conditionalization? Or convince @feanil that we don't need it? :)

gsong:
@feanil Do we plan on supporting mixed Ubuntu 12 & 16 environment? If not, can we just manage via Git refs which version of the playbooks/roles should be used?

feanil:
@gsong the issue is that not everything can be upgraded at the same time.  The browsers role is used by devstacks and by testeng jenkins workers and I don't want to commit to both landing in lock-step before we can merge the rest of these changes.  FWIW, I'm currently working on getting the jenkins workers running on 16.04

gsong:
@feanil For Ficus purposes, we can keep using this release branch and cherry-pick from/rebase onto `master` as needed.

In a separate email conversation, you mentioned:

> Prod courses.edx.org is not yet running on xenial. Mostly we've held off on doing so until we have jenkins doing so as well.  Which is what I'm currently working on.  Once Jenkins is running on xenial, we can deploy xenial to production.

Is the expectation that we’ll be completely on Xenial in the next couple weeks? We’re trying to gauge whether it’s worth the effort to have conditional roles based on OS release.

/cc @benpatterson

gsong:
@nedbat Can we close PR and remove this branch?

PR #3585, opened 2017-01-03
Ecommerce is no longer optional in edx_sandbox.yml

nedbat:
Configuration Pull Request
---


nedbat:
@antoviaque @bradenmacdonald I'm told that I am the only edXer using edx_sandbox.yml any more, but that OpenCraft still does.  What do you think of removing the optionalness of ecommerce?  I can leave it, and set the variable instead if you still want it.

bradenmacdonald:
@nedbat We would appreciate having it around for a couple more months. At some point this year we will completely transition to the `edx-stateless.yml` playbook, so then it won't matter to us.

nedbat:
OK, no problem.

PR #3586, opened 2017-01-03
Remove occurrences of ANSIBLE_REPO/VERSION.

doctoryes:
@feanil When looking at something unrelated today, I noticed references to the Ansible fork, which we've now abandoned with the Ansible 2.2 upgrade, correct? Are the changes in this PR then appropriate?


e0d:
@feanil @nedbat this LGTM if it didn't happen elsewhere.

nedbat:
Looks like ansible-bootstrap.sh sets this variable, but nothing it calls uses it. :+1:

PR #3587, opened 2017-01-03
Change migrate --list to showmigrations in pre_supervisor_checks

thallada:
Django removed `migrate --list` in version 1.10 in place of `showmigrations` which has the same output format.

We're unable to deploy the Analytics API because we updated to 1.10 and the `migrate --list` command fails.

All the other services listed have Django 1.8 or later so I switched them too (Django 1.8 added showmigrations). But, I didn't want to touch LMS, CMS, or xqueue since they seem to have their own special ways of checking for migrations that I didn't want to break.

Configuration Pull Request
---

  - are you adding any new default values that need to be overriden when this goes live?  I don't think so?

FYI: @dsjen @kevin @doctoryes


dsjen:
Wow, thank you for uncovering that!

doctoryes (approved):

jibsheet (approved):
You'll need to run a fresh deploy pipeline for this to be built into the image.

PR #3590, opened 2017-01-05
Don't use virtio for dogwood or eucalyptus boxes

nedbat:
The "virtio" nictype is faster, but needs support built into the box file.  Before Ficus, we didn't have support, so don't use it for older boxes.

@jlajoie @staubina @gsong 

---


gsong (approved):

staubina (approved):

jibsheet (approved):

nedbat:
This is now cherry-picked onto ned/test-ficus.

PR #3592, opened 2017-01-05
motd.tail support for Xenial

nedbat:
motd.tail isn't a thing in Ubuntu 14.04 and beyond.  The simplest way to keep it working is to add a file to /etc/update-motd.d that displays /etc/motd.tail

@e0d there was some thought that your xenial branch had a fix for this, but I didn't see it.


nedbat:
BTW, if this technique rubs you the wrong way, I can do it differently.

feanil (approved):
Verified here: feanil.sandbox.edx.org

PR #3593, opened 2017-01-05
Move to 16.04 base sandbox images.

feanil:
@edx/devops


feanil:
There need to be corresponding changes in the jenkins UI but those will have to be after this lands.

jibsheet (approved):

PR #3595, opened 2017-01-09
Basic support for smallfiles

e0d:
@nedbat @jibsheet I think we should include this too as it makes running small installations much more sane.

Configuration Pull Request
---


jibsheet (approved):
Can you file a ticket (or update one of our existing Mongo 3.0 on devstack/sandbox) tickets to cover setting this over there (but overriding it back off for production deploys)

nedbat (approved):

feanil:
@e0d did you already make the PRs to turn this off in our secure config?  I don't see them linked here and I didn't see one for edx-secure at least.  Did it get merged along with some other changes?

PR #3599, opened 2017-01-10
Feanil/credentials 16.04

feanil:
@rlucioni @clintonb corresponding updates to the base docker image.  Once this merges I'll push up this container and then update the test branch(https://github.com/edx/credentials/pull/148) to use the latest image.

I verified locally that this is still needed to keep the dev environment unbroken in 16.04

### DIFF
--- a/playbooks/roles/credentials/templates/edx/app/credentials/devstack.sh.j2
+++ b/playbooks/roles/credentials/templates/edx/app/credentials/devstack.sh.j2
@@ -11,11 +11,25 @@ case $COMMAND in
 
         {{ supervisor_venv_bin }}/supervisord --configuration {{ supervisor_cfg }}
 
+	# Needed to run bower as root. See explaination around 'discovery_user=root'
+	echo '{ "allow_root": true }' > /root/.bowerrc
+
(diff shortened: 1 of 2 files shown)


clintonb (approved):

playbooks/roles/credentials/templates/edx/app/credentials/devstack.sh.j2
```diff
@@ -11,11 +11,25 @@ case $COMMAND in
 
         {{ supervisor_venv_bin }}/supervisord --configuration {{ supervisor_cfg }}
 
+	# Needed to run bower as root. See explaination around 'discovery_user=root'
```
jibsheet, line 14: s/discovery/credentials

jibsheet (approved):
One copy pasta

fredsmith (approved):
I have strong reservations about running services as root, but we'll assume some technical debt with the understanding that this is devstack only and not the default entrypoint for the container.

feanil:
I made [OPS-1978](https://openedx.atlassian.net/browse/OPS-1978) for the follow-on work to move the node_modules directory out of the source directory as a part of the epic to support docker devstacks.

PR #3605, opened 2017-01-11
Feanil/ecom worker 16.04

feanil:
@edx/devops @rlucioni @clintonb Please review.

It turns out we didn't make a docker container for ecomworker.  Building it here so we can run the tests on top of it for the 16.04 work.

### DIFF
--- a/docker/build/ecomworker/Dockerfile
+++ b/docker/build/ecomworker/Dockerfile
@@ -10,9 +10,6 @@
 FROM edxops/xenial-common:latest
 MAINTAINER edxops
 
-ENV ECOMMERCE_WORKER_VERSION=master
-ENV REPO_OWNER=edx
-
 ADD . /edx/app/edx_ansible/edx_ansible
 
 WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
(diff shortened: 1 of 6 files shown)


jibsheet (approved):

util/parsefiles_config.yml
```diff
@@ -27,3 +27,4 @@ weights:
   - elasticsearch: 7
   - docker-tools: 3
   - tools_jenkins: 8
+  - ecomworker: 4
```
rlucioni, line 30: What is this and why is it 4?

docker/build/ecomworker/Dockerfile
```diff
@@ -0,0 +1,24 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/discovery/Dockerfile .
```
rlucioni: I 👀  your copypasta. Replace discovery with ecomworker.

clintonb (approved):

docker/build/ecomworker/Dockerfile
clintonb: s/discovery/ecomworker

rlucioni (approved):

util/parsefiles_config.yml
mikedikan: i'm guessing an ordering mechanism for build ordering, but i'm curious too

feanil: This is the amount of time it takes to build the docker container during testing.  It's used to evenly shard out all the test builds across the 4 travis shards we use.

feanil:
Thanks everyone, copy pasta consumed.  Will merge once the tests pass again.

util/parsefiles_config.yml
rlucioni: Got it, thanks!

mikedikan (approved):

PR #3607, opened 2017-01-12
temporarily install google chrome stable

estute:
[TE-1923](https://openedx.atlassian.net/browse/TE-1923)
This fixes the failing packer jobs, which were failing due to lettuce tests, which failed because they couldn't get a chrome instance up. The jobs began failing on Jan 4, due to a security flaw that was patched in libnss3. The patch was incompatible with our pinned version of chrome, and newer versions of chrome no longer support 12.04, so let's just go to the future and use Xenial.
 
@benpatterson @jzoldak 
2 things to be concerned with here:
* I used the google-chrome-stable package as opposed to a pinned version, but I see the rationale for using a pinned version in a testing environment. I would love your input.
* Regardless of if I pinned it or used the version in the apt-repo, the current version of chrome is notoriously slow (see [sitespeed update](http://support.speedcurve.com/testing-process/chrome-55-variability)) and since google doesn't believe in distributing older versions, we have to use it. As soon as 56 is released and deemed stable, let's use that.

@feanil @gsong 
This tackles part of [jenkins-16.04](https://github.com/edx/configuration/tree/feanil/jenkins-16.04), but for the sake of fixing our scripts, just wanted to focus on chrome. I know there is a separate Firefox issue.

### DIFF
--- a/playbooks/roles/browsers/tasks/main.yml
+++ b/playbooks/roles/browsers/tasks/main.yml
@@ -71,9 +71,42 @@
 - assert:
     that: "phantomjs.stat.exists"
 
-- name: create xvfb upstart script
-  template: src=xvfb.conf.j2 dest=/etc/init/xvfb.conf owner=root group=root
+- name: create xvfb upstart script for Precise and Trusty (12.04 and 14.04)
+  template:
+    src: xvfb.conf.j2
+    dest: /etc/init/xvfb.conf
+    owner: root
+    group: root
+  when: ansible_distribution_release == 'precise' or ansible_distribution_release == 'trusty'
+  tags:
+    - install
+    - install:configuration
 
-- name: start xvfb
+- name: start xvfb upstart script for Precise and Trusty (12.04 and 14.04)
   shell: start xvfb
   ignore_errors: yes
+  when: ansible_distribution_release == 'precise' or ansible_distribution_release == 'trusty'
+  tags:
+    - install
+    - install:configuration
+
+- name: create xvfb systemd service for Xenial (16.04)
+  template:
+    src: xvfb.service.j2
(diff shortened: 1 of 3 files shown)


estute:
note: need to make sure to cherry pick this into ficus/eucalyptus

gsong:
@estute We should compare notes re: browser installations. I've done work at <https://github.com/edx/configuration/tree/open-release/ficus.master/playbooks/roles/browsers>.

estute:
@gsong thanks for the heads up, let's talk tomorrow about it
@benpatterson @jzoldak after talking it over, I pinned the chrome version to favor stability of tests

jzoldak:
👍 
LGTM with it pinned at 55.
@estute can you please put in a reminder ticket in the backlog if there isn't one already for upgrading to version 56 when in comes out?

playbooks/roles/browsers/tasks/main.yml
```diff
@@ -71,9 +71,17 @@
 - assert:
     that: "phantomjs.stat.exists"
 
-- name: create xvfb upstart script
-  template: src=xvfb.conf.j2 dest=/etc/init/xvfb.conf owner=root group=root
+- name: create xvfb systemd service
+  template: src=xvfb.service.j2 dest=/etc/systemd/system/xvfb.service owner=root group=root
```
gsong: Pretty sure devops is going to want this as a dict rather than a string.

benpatterson:
LGTM

estute:
@feanil updated since our conversation earlier today

jzoldak:
re-reviewed the latest changes. 👍

e0d (approved):
LGTM with the updates for upstart + systemd

PR #3612, opened 2017-01-18
Enable and forward query logs for neo4j.

benpatterson:
This will help us gather important metrics related to adoption and usage. Knowledge base reference [here](https://neo4j.com/developer/kb/how-do-i-log-parameter-values-into-the-query-log-file/).

When deploying this, we'll need to enable the splunk forwarder (e.g., set `COMMON_ENABLE_SPLUNKFORWARDER`).

Configuration Pull Request
---

### DIFF
--- a/playbooks/edx-east/neo4j.yml
+++ b/playbooks/edx-east/neo4j.yml
@@ -4,6 +4,13 @@
   gather_facts: True
   vars:
     CLUSTER_NAME: 'coursegraph'
+    SPLUNKFORWARDER_LOG_ITEMS:
+      - source: '/etc/neo4j/log/query.log'
+        recursive: true
+        index: 'testeng'
+        sourcetype: coursegraph_query
+        followSymlink: false
+
   roles:
     - role: nginx
       nginx_template_dir: "../roles/neo4j/templates/edx/app/nginx/sites-available"
@@ -13,3 +20,5 @@
         - coursegraph
 #    - aws
     - neo4j
+    - role: splunkforwarder
+      when: COMMON_ENABLE_SPLUNKFORWARDER
--- a/playbooks/roles/neo4j/defaults/main.yml
+++ b/playbooks/roles/neo4j/defaults/main.yml
@@ -29,3 +29,4 @@ neo4j_listen_address: "0.0.0.0"
 # Properties file settings
 neo4j_https_settings_key: "dbms.connector.https.address"
 neo4j_http_settings_key: "dbms.connector.http.address"
+neo4j_enable_query_logging_key: "dbms.logs.query.enabled"
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -63,9 +63,18 @@
     - install
     - install:configuration
 
+- name: turn on logging for queries
+  lineinfile:
+    dest: "{{ neo4j_server_config_file }}"
+    regexp: "{{ neo4j_enable_query_logging_key }}="
+    line: "{{ neo4j_enable_query_logging_key }}=true"
+  tags:
+    - install
+    - install:configuration
+
 - name: restart neo4j
-  service: 
-    name: neo4j 
+  service:
+    name: neo4j
     state: restarted
   tags:
     - manage


benpatterson:
@fredsmith @adampalay 

Can you guys take a look?

A couple questions for Fred inline.

playbooks/edx-east/neo4j.yml
```diff
@@ -4,6 +4,13 @@
   gather_facts: True
   vars:
     CLUSTER_NAME: 'coursegraph'
+    SPLUNKFORWARDER_LOG_ITEMS:
+      - source: '/etc/neo4j/log/query.log'
```
benpatterson, line 8: I'm pretty sure this is the path, but @fredsmith could you look? Not sure if it would appear there before it's enabled, but it would be good to know beforehand. Related, is there a way I could test this out?

playbooks/edx-east/neo4j.yml
```diff
@@ -4,6 +4,13 @@
...
+        recursive: true
+        index: 'testeng'
+        sourcetype: coursegraph_query
```
benpatterson, line 11: I just made up this `sourcetype`. Is that ok? Or, does it need to exist already on the splunk head?

playbooks/roles/neo4j/defaults/main.yml
```diff
@@ -29,3 +29,4 @@ neo4j_listen_address: "0.0.0.0"
 # Properties file settings
 neo4j_https_settings_key: "dbms.connector.https.address"
 neo4j_http_settings_key: "dbms.connector.http.address"
+neo4j_enable_query_logging_key: "dbms.logs.query.enabled"
```
benpatterson, line 32: I'm assuming this is already in the conf file, but commented out. It will not work if this setting cannot be found in the file. Would be great to confirm this either by Fred looking or by having a way to test these changes.

benpatterson:
Also if, instead of Fred, this should be handled by the oncall person, just LMK

adampalay:
I don't feel qualified to affirm the details of the configuration, but I can +1 the idea! Note that we make queries during the update job, so we'd just want to filter those ones out when reporting on the number and kinds of queries folks are making

benpatterson:
@adampalay is the update job run on a certain cadence? once a day at a certain time?

That said, this should capture the query itself, and we can likely filter that out with regex.

adampalay:
@benpatterson , right now it's run once a week, but I'm certain we can filter out those queries in splunk (for example, no one should be doing `CREATE` or `DELETE`, which the script does).

playbooks/roles/neo4j/defaults/main.yml
fredsmith: Looks like it's not currently there.

```
ubuntu@ip-10-254-3-218:~$ grep dbms.logs /etc/neo4j/neo4j.conf
#dbms.logs.http.enabled=true
#dbms.logs.gc.enabled=true
#dbms.logs.gc.options=-XX:+PrintGCDetails -XX:+PrintGCDateStamps -XX:+PrintGCApplicationStoppedTime -XX:+PrintPromotionFailure -XX:+PrintTenuringDistribution
#dbms.logs.gc.rotation.keep_number=5
#dbms.logs.gc.rotation.size=20m
#dbms.logs.debug.rotation.size=20m
#dbms.logs.debug.rotation.keep_number=7
```

playbooks/edx-east/neo4j.yml
fredsmith: there's no current query log, but the rest of the neo4j logs go in /var/log/:

```
$ ls /var/log/neo4j/
debug.log  neo4j.log
```

playbooks/edx-east/neo4j.yml
fredsmith: I think this needs to be defined before we can start using it.

fredsmith:
@benpatterson see comments inline - I think we'll need to change the path to the query log, and define the source type before this can be run on the neo4j infrastructure.

playbooks/edx-east/neo4j.yml
```diff
@@ -4,6 +4,13 @@
   gather_facts: True
   vars:
     CLUSTER_NAME: 'coursegraph'
+    SPLUNKFORWARDER_LOG_ITEMS:
+      - source: '/etc/neo4j/log/query.log'
+        recursive: true
+        index: 'testeng'
```
adampalay, line 10: @benpatterson should we have a new index for coursegraph (like coursegraph)? Or should we keep it under the prod-edx index?

benpatterson: I don't think it should be under prod-edx.... but sounds like you don't think it should be under 'testeng'?

If we create another index, I believe there needs to be additional configuration changes, but I don't know the details of how to effect them.

adampalay: tbh, I'm not really sure what index this should be under. Testeng doesn't quite feel right, because it's not part of our testing or quality infrastructure. I was thinking prod-edx b/c it has prod data and prod systems write to it directly.

In general, I don't have strong feelings here, just raising the question

fredsmith: I think prod-edx is fine

benpatterson:
Fellas, turns out we need enterprise edition for this level of logging. We may consider the access log but I've been unable to actually capture any logs on nginx on a vagrant instance. It needs further troubleshooting. 

Keeping this open for now, but wanted to ping with an update.

benpatterson:
@adampalay I'm closing this down since we can't do much on the free version. going the nginx route might be worthwhile to pursue.

PR #3614, opened 2017-01-19
disable auto apt updates in xenial

estute:
Since I started using a Xenial ami in our packer builds nearly all of our builds were failing (perhaps the only ones that succeeded beat the automatic updates race). 
@feanil @jibsheet for the devops approval
@benpatterson @jzoldak just to let you know

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -92,6 +92,17 @@ fi
 
 EDX_PPA="deb http://ppa.edx.org ${SHORT_DIST} main"
 
+# Temporarily turn off the automatic updates for Xenial (16.04) to avoid 
+# conflicts with this script
+if [ $SHORT_DIST == "xenial" ]; then
+    systemctl stop apt-daily.service
+    systemctl kill --kill-who=all apt-daily.service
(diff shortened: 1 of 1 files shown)


jibsheet:
This happens during the security role 
https://github.com/edx/configuration/blob/master/playbooks/roles/security/tasks/security-ubuntu.yml#L25-L30
Please note that before a reboot, it'll look like some of the timers are still running if you run systemctl list-timers (cutting an AMI counts as a reboot).

feanil (approved):
I think stopping this for now and proceeding makes sense.   That way we don't make a decision about whether we want this on or off here.  I think this is good to merge for now and we can improve on it if we want to consolidate the logic that is in the security role higher up.

gsong:
@estute and @feanil Unfortunately this change is causing issues when we’re building Vagrant box for devstack:

```
==> default: + '[' xenial == xenial ']'
==> default: + systemctl stop apt-daily.service
==> default: + systemctl kill --kill-who=all apt-daily.service
==> default: Failed to kill unit apt-daily.service: Unit apt-daily.service is not loaded.
```

Let’s figure out what’s the best way to resolve this.

/cc @nedbat

PR #3618, opened 2017-01-23
add test packer script

estute:
Not a PR... used for investigating some packer scripts


e0d:
@estute going concern, or can be closed?

openedx-webhooks:
Although this pull request is already closed, I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #3623, opened 2017-01-23
Conditionally kill the `apt-daily.service` if running

gsong:
`systemctl kill` fails if the service isn't running, which is the case with barebones devstack Ubuntu boxes.

Let's make sure `apt-daily.service` is running before trying to kill it.

/cc @nedbat @estute @feanil 

---

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -92,11 +92,14 @@ fi
 
 EDX_PPA="deb http://ppa.edx.org ${SHORT_DIST} main"
 
-# Temporarily turn off the automatic updates for Xenial (16.04) to avoid 
+# Temporarily turn off the automatic updates for Xenial (16.04) to avoid
 # conflicts with this script
 if [ $SHORT_DIST == "xenial" ]; then
-    systemctl stop apt-daily.service
-    systemctl kill --kill-who=all apt-daily.service
+    APT_DAILY=apt-daily.service
+    if systemctl --state=active | grep -q $APT_DAILY; then
+        systemctl stop $APT_DAILY
+        systemctl kill --kill-who=all $APT_DAILY
+    fi
     while lsof |grep -q /var/lib/dpkg/lock; do
         echo "Waiting for apt to release the dpkg lock..."
         sleep 5


util/install/ansible-bootstrap.sh
```diff
@@ -92,11 +92,14 @@ fi
...
 
-# Temporarily turn off the automatic updates for Xenial (16.04) to avoid 
+# Temporarily turn off the automatic updates for Xenial (16.04) to avoid
 # conflicts with this script
 if [ $SHORT_DIST == "xenial" ]; then
-    systemctl stop apt-daily.service
-    systemctl kill --kill-who=all apt-daily.service
+    APT_DAILY=apt-daily.service
+    systemctl stop $APT_DAILY
+    if systemctl --state=running | grep -q $APT_DAILY; then
```
feanil: Why not put the `systemctl stop` in the conditional as well?

gsong: I can definitely move it there. I’m actually not certain why we need both `stop` and `kill`. @estute?

estute:
@gsong can you please check the conditions in the if block. It doesn't seem to map to the current state of apt.daily and therefore is not killing the service

gsong:
@estute Can you sent the results of `sudo systemctl list-units | grep apt-daily.service` so I can see what states it’s in?

estute:
@gsong 
"apt-daily.service                                 loaded activating start     start Daily apt activities'"

estute:
It works if you check for --state=active

gsong:
@estute I made the change.

But @e0d has some questions about the whole approach of `systemctl kill` in addition to looking for dpkg lock, so he may have something to say as well.

e0d:
@estute @feanil @jibsheet @nedbat @gsong chatted with Stuart a bit tonight and have more context.

Looking at this a bit, my current thinking is that dealing the apt-daily shouldn't be the responsibility of this script.  It could be defensive and bale out with a useful error, but it should be the responsibility of the caller to ensure that apt-get could be run and that concurrency issues or race conditions are managed.

Other thoughts?

clintonb:
What issue does this `systemctl` call solve? When I build locally with the calls commented out, I have no (noticeable?) issues. Can we just remove the whole block?

gsong:
@clintonb This is a problem only experienced at AWS, my preference is to not solve the problem at this script level (per @e0d), if possible.

estute:
@clintonb there is a race condition for access to the dpkg lock between the unattended-updates spawned by the apt-daily.service and the bootstrap script when using this to provision new AMIs.

estute:
@gsong putting in a pr for that

jibsheet:
Echoing what I noted to Ed.  When I solved this issue (by disabling the updates in the security role).  I did so because the daily timers only seemed to run on second boot of an AMI.  The initial boot of an AMI from Ubuntu didn't have the daily timer enabled / didn't cause issues.  We were seeing problems when we would boot a new machine from the snapped AMI and (fresh ubuntu AMI, run edxapp play, take AMI.  Boot AMI, run edxapp play to update to master - blam, apt/dpkg lock).

I don't think I understand the scenario where Stu was running into the timers existing, and I agree that we shouldn't be doing this disabling in the bootstrap script if at all possible.

feanil:
Closing this since https://github.com/edx/configuration/pull/3627 supersedes it.

PR #3627, opened 2017-01-24
remove systemctl calls to release dpkg lock

estute:
This was caused by the fact that we were using an AMI that had been manually booted once before (and thus triggered the apt services)
@edx/devops will move this into the tools scripts for the short term, and use the appropriate plays when I get a chance.
CC @gsong @clintonb


jibsheet (approved):

clintonb:
@edx/devops now that this is merged, would you push a new xenial-common image to Docker Hub, please? DEVOPS-5331

PR #3628, opened 2017-01-24
temp add stop-automatic-updates to packer scripts

estute:
@jibsheet @feanil @benpatterson @jzoldak 

The only reason (as Kevin helped me realize) that we were seeing an issue with dpkg locks is because there is still a manual step to how we provision our AMIs, meaning 2 separate boots of a vanilla AMI which triggers the apt service. In the interest of getting our packer scripts up and running to unblock 16.04 testing, this TEMPORARY fix will do the trick, but we do plan on provisioning the AMIs programmatically.

### DIFF
--- a/util/packer/stop-automatic-updates.sh
+++ b/util/packer/stop-automatic-updates.sh
@@ -1,0 +1,23 @@
+#!/usr/bin/env bash
+
+# Kill the apt services holding a dpkg lock, so that the ansible-bootstrap
+# script can run without conflicts.
+
+# NOTE: this is a temporary fix. Instead, we should be doing what devops does,
+# and first run the security+common roles on a vanilla AMI, which will disable
+# unattended-updates and set up users. Then we can feel free to run the
+# ansible bootstrap without any problems.
+
+set -xe
+
+if grep -q 'Xenial Xerus' /etc/os-release; then
+    systemctl stop apt-daily.service
+    systemctl kill --kill-who=all apt-daily.service
+    # Our jenkins job for building AMIs will timeout, even if the lock is
+    # never released.
+    while lsof |grep -q /var/lib/dpkg/lock; do
+        echo "Waiting for apt to release the dpkg lock..."
+        sleep 5
+    done
+fi
+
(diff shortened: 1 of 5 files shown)


jibsheet (approved):
Looks good to me, but your teammates should probably comment.

util/packer/stop-automatic-updates.sh
```diff
@@ -0,0 +1,16 @@
...
+# Kill the apt services holding a dpkg lock, so that the ansible-bootstrap
+# script can run without conflicts.
+
+set -xe
+
+if grep -q 'Xenial Xerus' /etc/os-release; then
+    systemctl stop apt-daily.service
+    systemctl kill --kill-who=all apt-daily.service
+    while lsof |grep -q /var/lib/dpkg/lock; do
+        echo "Waiting for apt to release the dpkg lock..."
```
feanil: Are there any concerns about this being in an infinite loop if the kill fails?

feanil (approved):

util/packer/stop-automatic-updates.sh
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env bash
+
+# Kill the apt services holding a dpkg lock, so that the ansible-bootstrap
+# script can run without conflicts.
+
```
jzoldak, line 5: Pls add a comment in the script to the effect of:
* We know that this could be an infinite loop, it's not hardened and we are not concerned. Any errors we can troubleshoot via log files.
* This workaround will not be necessary if we run the part of the ansible scripts where we put our ssh keys on the boxes, as this would also turn off the daily update service.

jzoldak (approved):
Small comment about an additional comment in the script. 
Otherwise looks good. 👍

benpatterson:
I'm assuming yes, but... any reason why you can't disable the service?

util/packer/stop-automatic-updates.sh
estute: Done.

benpatterson:
@estute answered all my questions offline. I am with you now.

👍

PR #3629, opened 2017-01-25
Xenial browser requirements

gsong:
* Remove OS version conditionals
* Pin Firefox to v45.*

@estute, @feanil and I have agreed that we don't need to worry about conditionals around OS versions.

/cc @nedbat 

---


estute:
:+1: 
CC: @benpatterson

gsong:
Since @estute gave the 👍, and we’re the two most interested party, I’m going to go ahead and merge.

PR #3631, opened 2017-01-25
Provide broker host var to ecomworker role

### LINKED ISSUE #3611: ecomworker can't connect to celery
From Slack: https://openedx.slack.com/archives/ops/p1484672679001328

> the error message that the others have mentioned, it appears in */edx/var/log/ecomworker/edx.log* and says:
> 
> Jan 11 18:17:29 ip-172-31-20-176 [service_variant=ecomworker][celery.worker.consumer][env:no_env] ERROR [ip-172-31-20-176  1229] [consumer.py:364] - consumer: Cannot connect to amqp://celery:**@172.31.20.176:5672//: [Errno 111] Connection refused.
> 
> This is because RabbitMQ listens on port 5672 only on the *localhost* interface, while module *ecomworker* is trying to contact it via IP address assigned to the *eth0* interface. I believe the offending piece of configuration is set in the $BROKER_URL environment variable. In the already installed system, this can be modified in files */edx/etc/ecommerce.yml* and */edx/etc/ecomworker.yml* by setting *BROKER_URL:* *amqp://celery:celery@localhost:5672*. Here, *localhost* is used instead of the IP address of *eth0* interface. (Then, restart your machine to make sure it works).

This is using open-release/ficus.1rc1
cc @gsong

gsong:
Fix #3611, which is a **blocker for Ficus release**.

Set the default broker host for ecomworker to be 127.0.0.1 in alignment with how we set up rabbitmq.

/cc @nedbat 

---

### DIFF
--- a/playbooks/edx_sandbox.yml
+++ b/playbooks/edx_sandbox.yml
@@ -55,6 +55,7 @@
     - role: ecommerce
       when: SANDBOX_ENABLE_ECOMMERCE
     - role: ecomworker
+      ECOMMERCE_WORKER_BROKER_HOST: 127.0.0.1
       when: SANDBOX_ENABLE_ECOMMERCE
     - programs
     - analytics_api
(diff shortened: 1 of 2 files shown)


nedbat:
@edx/devops ping

e0d (approved):

PR #3632, opened 2017-01-25
prefix migration output with the database name

macdiesel:
@edx/pipeline-team PTAL.


doctoryes:
Ah - good ole' CSMH. 👍

jibsheet:
ahem, CSMHE - because CSMH wasn't a long enough class name.

PR #3637, opened 2017-01-26
install pinned chromium from s3

estute:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/browsers/defaults/main.yml
+++ b/playbooks/roles/browsers/defaults/main.yml
@@ -7,6 +7,7 @@ browser_deb_pkgs:
     - libcurl3
     - xdg-utils
     - gdebi
+    - chromium-codecs-ffmpeg-extra=49.0.2623.108-0ubuntu1.1233
 
 # Debian packages we host in S3 to ensure correct browser version
 # Both Chrome and FireFox update their apt repos with the latest version,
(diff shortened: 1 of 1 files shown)


playbooks/roles/browsers/defaults/main.yml
```diff
@@ -7,13 +7,17 @@ browser_deb_pkgs:
...
 
 # Debian packages we host in S3 to ensure correct browser version
 # Both Chrome and FireFox update their apt repos with the latest version,
 # which often causes spurious acceptance test failures.
 browser_s3_deb_pkgs:
     - { name: "firefox-mozilla-build_42.0-0ubuntu1_amd64.deb", url: "https://s3.amazonaws.com/vagrant.testeng.edx.org/firefox-mozilla-build_42.0-0ubuntu1_amd64.deb" }
-    - { name: "google-chrome-stable_55.0.2883.87-1_amd64.deb", url: "https://s3.amazonaws.com/vagrant.testeng.edx.org/google-chrome-stable_55.0.2883.87-1_amd64.deb" }
+    - { name: "chromium-browser_49.0.2623.108-0ubuntu1.1233_amd64.deb", url: "https://s3.amazonaws.com/vagrant.testeng.edx.org/chromium-browser_49.0.2623.108-0ubuntu1.1233_amd64.deb" }
+
+# Temporarily NOT installing google-chrome
```
feanil, line 19: Please add more to this comment to say why we're not installing chrome.

playbooks/roles/browsers/defaults/main.yml
```diff
@@ -7,13 +7,17 @@ browser_deb_pkgs:
     - libcurl3
     - xdg-utils
     - gdebi
+    - chromium-codecs-ffmpeg-extra=49.0.2623.108-0ubuntu1.1233
```
feanil, line 10: Can you add a comment that this can be removed when we stop using chromium?

feanil (changes requested):

PR #3639, opened 2017-01-27
Don't enrol the staff user in the demo course if the staff user wasn't created.

smarnach:
When `DEMO_CREATE_STAFF_USER` is set to false, the user won't be created, but the demo role will try to enrol the user anyway, resulting in an error.  This fix makes sure the staff user is only enroled if it was created.

This fix should also be cherry-picked on the Ficus release branch.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/demo/tasks/deploy.yml
+++ b/playbooks/roles/demo/tasks/deploy.yml
@@ -16,22 +16,23 @@
   become_user: "{{ common_web_user }}"
   when: demo_checkout.changed
 
-- name: create some test users
-  shell: "{{ demo_edxapp_venv_bin }}/python ./manage.py lms --settings={{ demo_edxapp_settings }} --service-variant lms manage_user {{ item.username}} {{ item.email }} --initial-password-hash {{ item.hashed_password | quote }}"
-  args:
-    chdir: "{{ demo_edxapp_code_dir }}"
-  become_user: "{{ common_web_user }}"
-  with_items: "{{ demo_test_users }}"
-  when: demo_checkout.changed
+- name: build staff and test user list
+  set_fact:
+    demo_test_and_staff_users: "{{ demo_test_users + [demo_staff_user] }}"
+  when: DEMO_CREATE_STAFF_USER
 
-- name: create staff user
-  shell: "{{ demo_edxapp_venv_bin }}/python ./manage.py lms --settings={{ demo_edxapp_settings }} --service-variant lms manage_user staff staff@example.com --initial-password-hash {{ demo_hashed_password | quote }} --staff"
+- name: build staff and test user list
+  set_fact:
+    demo_test_and_staff_users: "{{ demo_test_users }}"
+  when: not DEMO_CREATE_STAFF_USER
+
+- name: create some test users
+  shell: "{{ demo_edxapp_venv_bin }}/python ./manage.py lms --settings={{ demo_edxapp_settings }} --service-variant lms manage_user {{ item.username}} {{ item.email }} --initial-password-hash {{ item.hashed_password | quote }}{% if item.is_staff %} --staff{% endif %}"
   args:
     chdir: "{{ demo_edxapp_code_dir }}"
   become_user: "{{ common_web_user }}"
-  when:
-    - demo_checkout.changed
-    - DEMO_CREATE_STAFF_USER
+  with_items: "{{ demo_test_and_staff_users }}"
+  when: demo_checkout.changed
 
 - name: enroll test users in the demo course
   shell: "{{ demo_edxapp_venv_bin }}/python ./manage.py lms --settings={{ demo_edxapp_settings }} --service-variant lms enroll_user_in_course -e {{ item.email }} -c {{ demo_course_id }}"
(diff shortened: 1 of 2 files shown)


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=3639&repo=edx%2Fconfiguration"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

playbooks/roles/demo/tasks/deploy.yml
```diff
@@ -41,8 +41,9 @@
   with_items:
       - "{{ demo_test_users }}"
       - { email: 'staff@example.com' }
-  when: demo_checkout.changed
-
+  when:
+    - demo_checkout.changed
+    - DEMO_CREATE_STAFF_USER or item.email != 'staff.example.com'
```
e0d: Do we need both DEMO_CREATE_STAFF_USER and COMMON_CREATE_STAFF_USER?

smarnach: @e0d Sorry, `COMMON_CREATE_STAFF_USER` does not exist.  It was just a typo in the PR description.

pomegranited: Typo: you've got `staff.example.com`, but should be `staff@example.com`.

pomegranited: I've never liked that `staff@example.com` is hard-coded into this playbook.  Can we have staff list like `demo_test_users`, e.g. `demo_staff_users` ?

I'm open to discussion whether this is out of scope for this change, but I've suggested it since we're adding a third occurrence of the string.

pomegranited (changes requested):

pomegranited (approved):
:+1:

- [x] I tested this by deploying a new edxapp server with `DEMO_CREATE_STAFF_USER: false`, and verifying that the staff user is not created, but the other default users are.
- [x] I read through the code
- [x] <strike>I checked for accessibility issues</strike>
- [x] <strike>Includes documentation</strike> - not required

feanil:
@smarnach perhaps the right way forward is to just have one list of demo users? This is getting quite messy.

smarnach:
@feanil Well, the code already was a bit messy.  I tried to clean it up while keeping backwards-compatibility.  If we use just a single list of users, the only way to achieve what `DEMO_CREATE_STAFF_USER: false` currently does is to copy the whole list of remaining users from the defaults file.  Another change in behaviour compared to the current situation is that the staff user would also get enrolled in the test courses, while currently only the `demo_test_users` get enrolled.  If you are fine with these backwards-incompatible changes, I'm happy to make the required changes.

smarnach:
@feanil Since this fixes a bug in Ficus, it would be nice if we could get this into the release.  I'm happy to change it in the way you requested if you are fine with the implications.

@nedbat Once this is merged, it would be great if this could be cherry-picked on the Ficus branch.

feanil:
I'm fine with leaving it as is, given backwards compatibility.  However, what was the reason for removing the course modes associated with the users?

smarnach:
@feanil The course modes aren't used anymore since [this commit](https://github.com/open-craft/configuration/commit/d238da6304c192924442e22cff23c6f8b15b75ed), so I didn't see a point in keeping them.  The `enroll_user_in_course` managment command now used to enroll users does not support to choose a mode.  It always enrols users in the default mode.

feanil:
Then I wonder if we should change the user-ids and e-mails as well.  Otherwise, people testing might make an assumption that the user is a certain kind of user.  We should have done it when it was no-longer relevant but we can do it now so it's clear in the future.

smarnach:
@feanil Renaming the users might make sense given that the current names are misleading.  However, these user accounts are documented in many places (e.g. [here](https://github.com/edx/edx-platform/wiki/Frequently-Asked-Questions#which-are-default-accounts-on-edx) and [here](https://github.com/edx/edx-documentation/blob/master/en_us/install_operations/source/installation/installation_options.rst#default-accounts)), and I suspect that there could be some automated tests that depend on the current names, so we might create more confusion than we clear up.

An alternative to changing the names would be to reintroduce a way of enrolling users in a specific course mode via the `enroll_user_in_course` management command.  I don't know how useful that would actually be in practice.

Our main goal with this PR was to fix a bug in the Ficus release.  Since we are already very close to the release, I'd like to keep the changes as small and as backwards-compatible as possible, so my preferred option would be to merge this PR as it is.

openedx-webhooks:
Thanks for the pull request, @smarnach! I've created [OSPR-1666](https://openedx.atlassian.net/browse/OSPR-1666) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

feanil (approved):

feanil:
@smarnach  makes sense, thanks for finding this debt, I'll put something on our backlog to clean it up. Hopefully before the next named release.

smarnach:
Thanks @feanil!

@nedbat Could you please consider cherry-picking this on the Ficus release branch?  It's a small bug fix for a bug only occurring when `DEMO_CREATE_STAFF_USER` is set to `False`, which probably very few people do.

nedbat:
I've cherry-picked this to Ficus.

smarnach:
Thanks, @nedbat!

PR #3648, opened 2017-01-30
Pin nodeenv's node version to 6.9.4 LTS for Platform, Ecommerce, Insights and Credentials

bjacobel:
Fixes [FEDX-299](https://openedx.atlassian.net/browse/FEDX-299).

Currently, new devstacks are unable to be built from the `open-release/ficus.master` tag. Devstack provisioning repeatedly fails on the following step:

```
TASK [ecommerce : Install bower dependencies] **********************************
==> default: fatal: [localhost]: FAILED! => {"changed": true, "cmd": ". /edx/app/ecommerce/nodeenvs/ecommerce/bin/activate && /edx/app/ecommerce/ecommerce/node_modules/.bin/bower install --production --config.interactive=false", "delta": "0:00:02.984544", "end": "2017-01-30 16:12:40.111830", "failed": true, "rc": 1, "start": "2017-01-30 16:12:37.127286", "stderr": "module.js:472\n    throw err;\n    ^\n\nError: Cannot find module 'internal/fs'\n    at Function.Module._resolveFilename (module.js:470:15)\n    at Function.Module._load (module.js:418:25)\n    at Module.require (module.js:498:17)\n    at require (internal/module.js:20:19)\n    at evalmachine.<anonymous>:18:20\n    at Object.<anonymous> (/edx/app/ecommerce/ecommerce/node_modules/bower/lib/node_modules/graceful-fs/fs.js:11:1)\n    at Module._compile (module.js:571:32)\n    at Object.Module._extensions..js (module.js:580:10)\n    at Module.load (module.js:488:32)\n    at tryModuleLoad (module.js:447:12)", "stdout": "", "stdout_lines": [], "warnings": []}
```

This is due to:

**Bug 1**: the Ecom nodeenv is not passing a specific Node version to `nodeenv`. The behavior of this for `nodeenv` results in installing the latest node with a prebuilt binary for that system, which for Ubuntu 16.04 is Node 7.4.0.

[The error above is a known issue with Bower and Node 7](https://github.com/bower/bower/issues/2393), and in general we haven't tested any edX applications with Node 7 yet, so we want to pin to Node 6.9.4 (most recent in the LTS branch) for now. I have made this change to all edX apps that do not specify a node version and are using nodeenv - Insights and Credentials are also affected.

**Bug 2**: the `ficus-devstack-2017-01-11` Vagrant box appears to ship with a Node 7.4.0 nodeenv inside `/edx/app/ecommerce/nodeenvs/ecommerce` - that is, there is a nodeenv in this folder before the ecom role runs during provisioning. This causes the play "Create nodeenv" in `ecommerce/tasks/main.yml` to fail silently on every installation with `ficus-devstack-2017-01-11` as a base, because the nodeenv cannot be created in a non-empty directory. Thus, Node 7.4.0 is still the Node version in the ecom nodeenv post-provisioning, even if you manually specify 6.9.4 in the nodeenv creation command in the playbook. To fix this, I've ~~added a step to the Ecom playbook that deletes the nodeenv if it exists before it creates a new nodeenv using the pinned version~~ made nodeenv run with `--force` every time.

You can test this locally with:

```
$> export OPENEDX_RELEASE=open-release/ficus.master
$> curl -OL https://raw.github.com/edx/configuration/$OPENEDX_RELEASE/vagrant/release/devstack/Vagrantfile
$> CONFIGURATION_VERSION=bjacobel/ecom-node-version vagrant up
```

Tagging @clintonb and @rlucioni as reviewers on this as they made similar changes to Discovery last week. I don't think review from Analytics should be needed as they have never pinned a specific version of Node before.

Also tagging @nedbat and @gsong because of the issues with the Vagrant box, ~~which we may want to fix rather than adding the deletion step to the playbook, although I don't know if that is possible now that that specific named box has shipped~~. @nedbat, I believe this is a release blocker for Ficus.

Also tagging @jibsheet because ✨ devops ✨.

FYI: @andy-armstrong @dsjen @thallada

### DIFF
--- a/playbooks/roles/credentials/tasks/main.yml
+++ b/playbooks/roles/credentials/tasks/main.yml
@@ -43,9 +43,7 @@
     - install:app-requirements
 
 - name: create nodeenv
-  shell: "{{ credentials_venv_dir }}/bin/nodeenv {{ credentials_nodeenv_dir }} --prebuilt"
-  args:
-    creates: "{{ credentials_nodeenv_dir }}"
+  shell: "{{ credentials_venv_dir }}/bin/nodeenv {{ credentials_nodeenv_dir }} --node={{ credentials_node_version }} --prebuilt --force"
   become_user: "{{ credentials_user }}"
   tags:
     - install
--- a/playbooks/roles/credentials/tasks/main.yml
+++ b/playbooks/roles/credentials/tasks/main.yml
@@ -42,6 +42,17 @@
     - install
     - install:app-requirements
 
+# See https://github.com/ekalinin/nodeenv/issues/182
+# This can be removed once https://github.com/ekalinin/nodeenv/pull/183 is merged
+- name: Remove nodeenv src
+  file:
+    path: "{{ credentials_nodeenv_dir }}/src"
+    state: absent
+  become_user: "{{ credentials_user }}"
+  tags:
+    - install
+    - install:app-requirements
+
 - name: create nodeenv
   shell: "{{ credentials_venv_dir }}/bin/nodeenv {{ credentials_nodeenv_dir }} --node={{ credentials_node_version }} --prebuilt --force"
   become_user: "{{ credentials_user }}"
--- a/playbooks/roles/ecommerce/tasks/main.yml
+++ b/playbooks/roles/ecommerce/tasks/main.yml
@@ -41,9 +41,7 @@
     - install:app-requirements
 
 - name: Create nodeenv
-  shell: "{{ ecommerce_venv_dir }}/bin/nodeenv {{ ecommerce_nodeenv_dir }} --prebuilt"
-  args:
-    creates: "{{ ecommerce_nodeenv_dir }}"
+  shell: "{{ ecommerce_venv_dir }}/bin/nodeenv {{ ecommerce_nodeenv_dir }} --node={{ ecommerce_node_version }} --prebuilt --force"
   become_user: "{{ ecommerce_user }}"
   tags:
     - install
--- a/playbooks/roles/ecommerce/tasks/main.yml
+++ b/playbooks/roles/ecommerce/tasks/main.yml
@@ -40,6 +40,17 @@
     - install
     - install:app-requirements
 
+# See https://github.com/ekalinin/nodeenv/issues/182
+# This can be removed once https://github.com/ekalinin/nodeenv/pull/183 is merged
+- name: Remove nodeenv src
+  file:
+    path: "{{ ecommerce_nodeenv_dir }}/src"
+    state: absent
+  become_user: "{{ ecommerce_user }}"
+  tags:
+    - install
+    - install:app-requirements
+
 - name: Create nodeenv
   shell: "{{ ecommerce_venv_dir }}/bin/nodeenv {{ ecommerce_nodeenv_dir }} --node={{ ecommerce_node_version }} --prebuilt --force"
   become_user: "{{ ecommerce_user }}"
--- a/playbooks/roles/ecommerce/tasks/main.yml
+++ b/playbooks/roles/ecommerce/tasks/main.yml
@@ -52,6 +52,7 @@
     executable: "{{ ecommerce_nodeenv_bin }}/npm"
     path: "{{ ecommerce_code_dir }}"
     production: yes
+    state: latest
   become_user: "{{ ecommerce_user }}"
   environment: "{{ ecommerce_environment }}"
   tags:
--- a/playbooks/roles/common_vars/defaults/main.yml
+++ b/playbooks/roles/common_vars/defaults/main.yml
@@ -45,7 +45,7 @@ common_directories:
   - path: "/etc/logrotate.d/hourly"
   - path: "/etc/rsyslog.d/50-default.conf"
     state: absent
-    
+
 COMMON_ENVIRONMENT: 'default_env'
 COMMON_DEPLOYMENT: 'default_deployment'
 COMMON_PIP_VERBOSITY: ''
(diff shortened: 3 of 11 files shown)


gsong:
@nedbat I'll take the lead on reviewing this.

bjacobel:
I've just realized that this should probably be a PR into `open-release/ficus.master`, so I'll repoint it.

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -40,8 +40,17 @@
     - install
     - install:app-requirements
 
+- name: Remove nodeenv if it exists
```
jibsheet: Does this mean we have to reinstall on every ansible run?

playbooks/roles/credentials/tasks/main.yml
```diff
@@ -43,7 +43,7 @@
     - install:app-requirements
 
 - name: create nodeenv
```
jibsheet: I'm surprised this doesn't need the same blow-away-and-recreate logic from ecommerce

playbooks/roles/ecommerce/tasks/main.yml
jibsheet: Should we instead be testing a base Vagrant build, and when a new image is published it'll have the correct version?

jibsheet (commented):
Looking back at 
https://github.com/edx/configuration/pull/3444#pullrequestreview-7528803
It appears we were going to move to --force in the nodeenv calls to force versioning.  Should we do that here (and in edxapp) ?  It appears that change was lost after we merged the PR.
@maxrothman Is there a ticket about that change?

playbooks/roles/credentials/tasks/main.yml
bjacobel: Probably need to do more testing around this. I'm not even sure where the `credentials` app is located in devstack, (there's nothing in `/edx/app` for credentials?) so I can't check that it actually got the right node version in the nodeenv post-provisioning.

playbooks/roles/ecommerce/tasks/main.yml
bjacobel: I didn't realize nodeenv had a `--force` option, that is probably better. I can make that change for all five environments using nodeenv (edxapp, insights, credentials, ecom, discovery) if people agree.

playbooks/roles/credentials/tasks/main.yml
clintonb: The Credentials Service was never added to devstack. The E-Commerce and Programs Services are the only two supported ECOM services. We kinda dropped the ball on the other two. 😟

clintonb (approved):

gsong:
For the purpose of understanding what’s going on, in the current Vagrant devstack base box (ficus-devstack-2017-01-11), there are two nodeenvs installed:

* /edx/app/edxapp/nodeenvs/edxapp: v6.9.2
* /edx/app/ecommerce/nodeenvs/ecommerce: v7.4.0

bjacobel:
I think we do want the capability in the Ansible playbook to wipe out an existing nodeenv if there is one already on the image (upgrading to a newer node version than shipped on the box is the obvious case) so I will investigate the `--force` option and see if that does what we need without perf hit in the case where a nodeenv with the right version already exists (as in the case of edxapp right now)

Edit: If we build and test that capability it means it's not a huge deal that `ficus.master` shipped with 7.4.0 for ecom, I think - nobody uses the image without first running provision on it, right?

gsong:
> If we build and test that capability it means it's not a huge deal that ficus.master shipped with 7.4.0 for ecom, I think - nobody uses the image without first running provision on it, right?

@bjacobel That’s right. But we’ll most likely build a new base image for the next release anyway.

gsong:
> …so I will investigate the --force option and see if that does what we need without perf hit in the case where a nodeenv with the right version already exists (as in the case of edxapp right now)

The performance hit will be minimal since we’re using the `--prebuilt` option. I think it’s OK to use `--force` all the time.

gsong:
Is there a reason why we wouldn’t pin our node version for all projects as a default, and only override when a specific exceptional case comes up? This way we can also consolidate the node installation into one role, and just pass in the necessary variables at role inclusion time.

bjacobel:
@gsong IIRC it's been discussed, but we didn't do it at the time because not all apps were using nodeenv yet and we weren't sure it was the right time (or tool) to generalize the process. I think it sounds like a good idea, though - at the very least, a good start would be just having `6.9.4` as a common variable somewhere.

Where do we put vars like that?

ccing @maxrothman into this, who out of devops knows the most about nodeenv.

rlucioni (approved):

gsong:
> …a good start would be just having 6.9.4 as a common variable somewhere.
>
> Where do we put vars like that?

I think a good candidate would be the `common_vars` role.

bjacobel:
I've filed an issue on the nodeenv repo about one of the strange things we've seen while working on this: `nodeenv <path> --node=6.9.4 --force` will not result in a Node 6.9.4 nodeenv being created at `<path>` if a Node 7.4 one exists there already (ie, it doesn't downgrade nodeenvs, at least not between these two specific versions).

https://github.com/ekalinin/nodeenv/issues/182

gsong:
@bjacobel Feel free to merge at any time.

gsong (approved):

bjacobel:
If we merge this, won't Ecom still be using 7.4? As the nodeenv downgrade isn't currently working, even with `--force`.

gsong:
@bjacobel Yes. I'm working on building a new base box, and will update the `open-release/ficus.master` branch accordingly. The base box will have the proper node versions that we specify.

In the meantime, the work-around is to just blow away `/edx/app/ecommerce/nodeenvs/ecommerce` and re-provision the box.

bjacobel:
@jibsheet With your thumb, I think this is ready to merge. George helped fix the last few pieces of the puzzle with PRs into this one.

clintonb (commented):
Remember to squash all of these commits to as few as make sense.

gsong:
@bjacobel I'll take care of squashing and merging.

PR #3649, opened 2017-01-31
Make sure npm install always happens

gsong:
This forces the npm module to always perform install.


bjacobel (approved):
LGTM, except that we're missing out on being able to do this to credentials because the `npm install` is done [inside the app's Makefile](https://github.com/edx/credentials/blob/master/Makefile#L49).

PR #3650, opened 2017-01-31
Fullstack is now a native install, in Vagrant

nedbat:
Make "fullstack" be the same installation as "native" (edx_sandbox.yml), but in a Vagrantfile.

  - [ ] @gsong review?

### DIFF
--- a/util/install/install_stack.sh
+++ b/util/install/install_stack.sh
@@ -123,11 +123,6 @@ if [[ ! $release ]]; then
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


gsong:
@nedbat Do you want to make the associated change to `util/install/install_stack.sh` in this PR as well?

gsong (approved):
I’ve tested this fully by:

1. Creating a base box
2. Spinning up a release VM
3. Verifying all the services are running

PR #3651, opened 2017-01-31
Remove nodeenv src to address node rollback bug

gsong:
I tested this on a local devstack successfully.


bjacobel (approved):
FYI, that PR to nodeenv merged, but we also need it to be released to pypi, which it doesn't look like the maintainer plans on doing for this bugfix.

gsong:
@bjacobel Re: your PR, saw that this morning. 👏

PR #3654, opened 2017-02-01
migrating file from gist

e0d:
Configuration Pull Request @nedbat 
---


playbooks/sample_vars/password.yml
```diff
@@ -0,0 +1,74 @@
+# This file is used to generate overriden, unique secrets
+# for OPENEDx deployment environments
```
nedbat: "Open edX"

nedbat (approved):

nedbat (changes requested):
Shouldn't EDXAPP_EDXAPP_SECRET_KEY be in this list? Actually, there are many `*_SECRET_KEY` values that maybe should be in here?

nedbat:
On SECRET_KEY: I don't see a recommendation about length, and a 35-character random key is much better than what we are doing now for Open edX installers.

nedbat:
Seems like these lines should be part of this process:
```
        ANALYTICS_API_SECRET_KEY: 'Your secret key here'
        CREDENTIALS_SECRET_KEY: 'SET-ME-TO-A-UNIQUE-LONG-RANDOM-STRING'
        DISCOVERY_SECRET_KEY: 'Your secret key here'
        ECOMMERCE_SECRET_KEY: 'Your secret key here'
        EDX_NOTES_API_SECRET_KEY: "CHANGEME"
        EDXAPP_EDXAPP_SECRET_KEY: "DUMMY KEY CHANGE BEFORE GOING TO PRODUCTION"
        EDXAPP_PROFILE_IMAGE_SECRET_KEY: placeholder_secret_key
        INSIGHTS_SECRET_KEY: 'YOUR_SECRET_KEY_HERE'
        NOTIFIER_LMS_SECRET_KEY: "PUT_YOUR_SECRET_KEY_HERE"
        PROGRAMS_SECRET_KEY: 'Your secret key here'
```

nedbat (approved):

PR #3655, opened 2017-02-01
Record AMI tags when creating an AMI for continuous delivery

cpennington:
Configuration Pull Request
---

### DIFF
--- a/playbooks/continuous_delivery/templates/local/ami_template.yml.j2
+++ b/playbooks/continuous_delivery/templates/local/ami_template.yml.j2
@@ -1,3 +1,4 @@
 ami_id: {{ ami_register.image_id }}
 ami_message: {{ ami_register.msg }}
 ami_state: {{ ami_register.state }}
+{{ ami_tags.tags | to_nice_yaml }}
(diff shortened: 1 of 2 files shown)


playbooks/continuous_delivery/templates/local/ami_template.yml.j2
```diff
@@ -1,3 +1,4 @@
 ami_id: {{ ami_register.image_id }}
 ami_message: {{ ami_register.msg }}
-ami_state: {{ ami_register.state }}
\ No newline at end of file
+ami_state: {{ ami_register.state }}
+{{ ami_tags.tags | to_nice_yaml }}
```
macdiesel, line 4: as opposed to yucky yaml? =)

macdiesel (approved):

playbooks/continuous_delivery/templates/local/ami_template.yml.j2
cpennington: Indeed. `to_yaml` includes `{}` and makes it a single line, so it would conflict w/ the manual interpolation above.

PR #3657, opened 2017-02-01
Put the password generator into a real .sh file

nedbat:
Configuration Pull Request
---

### DIFF
--- a/util/install/generate-passwords.sh
+++ b/util/install/generate-passwords.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+#
+# Read a list of Ansible variables that should have generated values, and make
+# a new file just like it, with the generated values.
+
+TARGET=${OPENEDX_RELEASE-master}
+wget -q "https://raw.githubusercontent.com/edx/configuration/$TARGET/playbooks/sample_vars/passwords.yml" -O passwords-template.yml
+
+while IFS= read -r line; do
+    # Make a random string. SECRET_KEY's should be longer.
+    length=35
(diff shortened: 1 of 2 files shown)


e0d:
@nedbat on master, can update.

nedbat:
@edx/devops @gsong thoughts?

nedbat:
@edx/devops @gsong :ping_pong:

gsong (approved):

util/install/generate-passwords.sh
```diff
@@ -0,0 +1,18 @@
...
+
+while IFS= read -r line; do
+    # Make a random string. SECRET_KEY's should be longer.
+    length=35
+    if [[ $line == *SECRET_KEY* ]]; then
+        length=100
+    fi
+    REPLACE=$(LC_ALL=C < /dev/urandom tr -dc 'A-Za-z0-9' | head -c$length)
+    # Change "!!null"-to-end-of-line to the password.
+    echo "${line/%!!null*/'$REPLACE'}"
```
gsong: @nedbat I think the single ticks need to be escaped: `\'$REPLACE\'`

nedbat: I tested it: with backslashes, the resulting yaml file has backslash-quote.  Without the backslashes, the yaml file correctly has just quotes.

nedbat:
@edx/devops this makes master match the versions of these files in Ficus. Any concerns?

gsong:
Not something we need to address with this specific PR, but will need to revisit at some point is #3716.

jibsheet (approved):

PR #3658, opened 2017-02-01
We set LANG in lms.conf and cms.conf but missed it for Celery

jibsheet:
Probably needs some sort of Changelog.  Also, on 16.04, seems to default to having a LANG of utf8 for us.

Biggest fun is how to adequately test this.

@edx/devops 
@jmbowman  I would be interested to know if merely this line in workers.conf for the cms_default_1 worker is sufficient to fix it.

### DIFF
--- a/playbooks/roles/edxapp/templates/workers.conf.j2
+++ b/playbooks/roles/edxapp/templates/workers.conf.j2
@@ -1,7 +1,7 @@
 {% for w in edxapp_workers %}
 [program:{{ w.service_variant }}_{{ w.queue }}_{{ w.concurrency }}]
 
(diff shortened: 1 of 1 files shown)


feanil (approved):

jmbowman:
I added just `LANG=en_US.UTF-8` to the cms_default_1 worker config, and it did in fact allow the test file to import successfully on my 12.04 sandbox.

jibsheet:
@jmbowman Ok - so my inclination is to fix it this way, make the environments more consistent.
Do you want to chat with your team and see if they can think of reasons this might fail and need testing?  I can ask @benpatterson about what the e2e tests in stage catch and whether this is something we'd want to manually test (cc @cpennington )

jzoldak:
@jibsheet what platform features would use the celery workers? Does uploading assets do it? We upload assets in the e2e tests against staging. 
We do not currently import a course. However all the underlying bits and even the methods that we call do exist so it should be easy to add.

jibsheet:
@jzoldak I don't think uploads go through it.  I actually think most things don't touch the filesystem (which is where this bug was discovered).  The easiest way I know to get a good list of tasks run is to go to https://splunk.edx.org/en-US/app/search/reports and run "Celery tasks by queue" - the fulfillment tasks you can ignore, but it'll show you lms and cms

jzoldak:
@raeeschachar @Muddasser FYI

jmbowman:
@jibsheet Where are we on getting this deployed?  It's still holding back async course import, and I'm mostly done with the code changes for async export which would also be blocked by it.

jibsheet:
Per discussion - we're going to merge, test on stage tomorrow, monitor on prod the following day and then the async import can go back out.

PR #3659, opened 2017-02-01
Updated JWT authentication settings for Credentials Service

clintonb:
This brings the service inline with our other services.

ECOM-6938


clintonb:

This will be coupled with a forthcoming PR for edx/credentials.

clintonb:
Relies on https://github.com/edx/credentials/pull/163.

schenedx (approved):
👍

PR #3665, opened 2017-02-02
Added an additional JWT issuer to the credentials service

clintonb:
This restores backward-compatibility with the current method of token issuance betwen LMS and the Credentials Service.

ECOM-6938


clintonb:
It appears we were a bit too forward-thinking with #3659. LMS is still issuing client-specific tokens.

@schenedx please review

schenedx (approved):
👍

PR #3666, opened 2017-02-02
create pip3 alias and modify sudoers for python3

macdiesel:
@doctoryes @cpennington PTAL.

### DIFF
--- a/docker/build/go-agent/Dockerfile
+++ b/docker/build/go-agent/Dockerfile
@@ -61,7 +61,12 @@ RUN usermod -aG docker go
 
 # Assign the go user root privlidges
 RUN printf "\ngo      ALL=(ALL:ALL) NOPASSWD: /usr/bin/pip, /usr/local/bin/pip\n" >> /etc/sudoers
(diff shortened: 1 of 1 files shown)


cpennington:
I'm confused: http://packages.ubuntu.com/trusty/all/python3-pip/filelist suggests that 14.04 python3-pip should have `/usr/bin/pip3` in it.

macdiesel:
and dpkg agrees with this:
```
root@70d571d93646:/data# dpkg -L python3-pip
/.
/usr
/usr/bin
/usr/bin/pip3
```

It' is not on the container though:
```
root@70d571d93646:/data# ls -alh /usr/bin/pip* 
-rwxr-xr-x 1 root root 283 Jun 29  2016 /usr/bin/pip2
```

apt-search says it is installed in the system:
```
root@70d571d93646:/data# apt-cache pkgnames | grep python3-pip 
python3-pipeline
python3-pip
```

Not sure what's up here. We can dig more on this if you would like.

cpennington:
It seems worth looking at a little bit, just so we don't build up debt around python3 that ends up confusing us later.

PR #3670, opened 2017-02-03
fixing bugs reported in slack

e0d:
Configuration Pull Request @nedbat additional overrides.
---


playbooks/sample_vars/passwords.yml
```diff
@@ -63,6 +63,8 @@ EDXAPP_MONGO_USER: 'edxapp'
 FORUM_MONGO_USER: "cs_comments_service"
 FORUM_MONGO_PASSWORD: !!null
 
+RABBIT_ADMIN_PASSWORD: !!NULL
```
nedbat: This needs to be "!!null"

nedbat (changes requested):

nedbat (approved):

PR #3672, opened 2017-02-03
Syntax enhancements

gsong:
* Quote URL to prevent globbing
* `read -r` to prevent backslash mangling
* Use bash substring substitution


nedbat:
Not sure what glowing means in this context, but changes look good, and TIL about `${NAME/%pat/repl}'

gsong:
@nedbat Autocorrect globbing → glowing. 😕

PR #3674, opened 2017-02-06
Added Sailthru templates setting for ecomworker

clintonb:
ECOM-6975


clintonb:
FYI @edx/ecommerce @edx/devops #neverendingstory

PR #3675, opened 2017-02-06
Silently fail on missing requirements

nedbat:
The way this was before, in a new virtualenv, you would get this:
```
$ make requirements
  File "util/balancecontainers.py", line 1, in <module>
    import yaml
ImportError: No module named yaml
  File "util/balancecontainers.py", line 1, in <module>
    import yaml
ImportError: No module named yaml
pip install -qr pre-requirements.txt --exists-action w
pip install -qr requirements.txt --exists-action w
```
This would alarm cautious installers.  Now you get:
```
$ make requirements
pip install -qr pre-requirements.txt --exists-action w
pip install -qr requirements.txt --exists-action w
```

### DIFF
--- a/util/balancecontainers.py
+++ b/util/balancecontainers.py
@@ -1,11 +1,20 @@
-import yaml
-import os
-import pathlib2
-import itertools
-import sys
 import argparse
 import logging
+import os
+import sys
+
+try:
+    # This script is used by docker.mk at parse-time, which means when you run
+    # "make requirements" to install the required Python packages, this script
+    # runs before its requirements are installed. That means this import will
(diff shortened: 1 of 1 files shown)


nedbat:
@edx/devops @cpennington @clintonb I'm not sure if balancecontainers is used elsewhere.  This might make its failures less obvious if it is used without its requirements.
BTW: Note that the original failing module (yaml) wasn't actually used in the script, so it is no longer imported, but the same problem would happen with pathlib2.

clintonb (commented):
I am unfamiliar with this portion of the codebase. From a general Python point-of-view, it looks fine.

feanil (approved):

PR #3682, opened 2017-02-08
Use the ubuntu keyserver to retrieve the edX PPA key.

feanil:
We already do this in the common role but not in bootstrap or abbey roles.

Review: @nedbat @edx/devops 

The key must have already been on the ubuntu keyserver.  When I thought I was uploading it for the first time I updated it so that it's no longer expired.  The mit keyserver has been flaky in the past, we should try to consistently use the ubuntu keyserver if we can.

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -48,7 +48,7 @@ PYTHON_BIN="${VIRTUAL_ENV}/bin"
 ANSIBLE_DIR="/tmp/ansible"
 CONFIGURATION_DIR="/tmp/configuration"
 EDX_PPA="deb http://ppa.edx.org precise main"
-EDX_PPA_KEY_SERVER="hkp://pgp.mit.edu:80"
+EDX_PPA_KEY_SERVER="keyserver.ubuntu.com"
 EDX_PPA_KEY_ID="B41E5E3969464050"
 
 cat << EOF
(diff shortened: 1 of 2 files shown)


feanil:
http://keyserver.ubuntu.com/pks/lookup?op=vindex&search=ppa.edx.org&fingerprint=on

jibsheet (approved):
Assuming this passes ok on the docker tests (Ed and I saw some weird networking failures, which I think is why it specified an HKP port)

nedbat:
There's one more reference to pgp.mit.edu, (https://github.com/edx/configuration/blob/491f909de5f040f232bc214240282810861aea72/docker/build/go-agent/files/docker_install.sh#L33), but in a list of keyservers. Do we want to keep it?

nedbat:
During my latest installation testing, I had to run the installer script three times during two separate installs because of pgp.mit.edu flakiness.  What do we have to do to get this merged?

feanil:
I pushed up a commit that will run all the containers which should hopefully provide enough confidence that this will not break that.

feanil:
@e0d @jibsheet all docker containers are building and I was able to kick off an AMI from scratch that did not have any issues retrieving the key from this keyserver.

nedbat (approved):

nedbat:
I've cherry-picked this commit onto ficus.

PR #3683, opened 2017-02-08
TNL-6510: Initialize elasticsearch for forums.

robrap:
Configuration Pull Request
---

### Sandbox
- [X] sandbox ([https://robrap.sandbox.edx.org/](https://robrap.sandbox.edx.org/))

### Post-review
- [x] Rebase and squash commits

### DIFF
--- a/playbooks/roles/forum/tasks/deploy.yml
+++ b/playbooks/roles/forum/tasks/deploy.yml
@@ -26,7 +26,7 @@
   tags:
     - install
     - install:configuration
-  
+
 - name: create the supervisor wrapper
   template:
     src: "{{ forum_supervisor_wrapper|basename }}.j2"
@@ -60,6 +60,17 @@
     - install
     - install:app-requirements
 
+- name: initialize elasticsearch
+  command: "{{ forum_code_dir }}/bin/rake search:initialize"
+  args:
+    chdir: "{{ forum_code_dir }}"
+  become_user: "{{ forum_user }}"
+  environment: "{{ forum_environment }}"
+  when: migrate_db is defined and migrate_db|lower == "yes"
+  tags:
+    - migrate
+    - migrate:db
+
   # call supervisorctl update. this reloads
   # the supervisorctl config and restarts
   # the services if any of the configurations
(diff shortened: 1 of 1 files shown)


robrap:
@maxrothman: FYI: I'm provisioning a sandbox now to test this.

robrap:
@maxrothman @dianakhuang: Discussions was properly initialized on the sandbox with no intervention.

dianakhuang (approved):
It might make sense to also get @maxrothman to test whether or not this is skipped in environments where it shouldn't be run.

maxrothman (commented):
This looks good to me, though I'd like one more review from a member of @edx/devops.

maxrothman (approved):

robrap:
@dianakhuang: Agreed.  But, because it is idempotent, it also really wouldn't matter at this point if it gets re-run.  I'll let @maxrothman determine how and if testing is necessary.

maxrothman:
I'm reasonably confident that this'll be skipped when it should be, since `migrate_db` is only defined in playbooks where it should be run, so in all other cases, the `migrate_db is defined` check should fail. Still would like a second review though.

playbooks/roles/forum/tasks/deploy.yml
```diff
@@ -60,6 +60,17 @@
     - install
     - install:app-requirements
 
+- name: initialize elasticsearch
+  command: bin/rake search:initialize
```
e0d: If the env script is correct, the complete relative path should not be necessary.  I guess my question is is this working around an issue with the env script?  If absolute safety is what you crave,  ```{{ forum_code_dir }}/bin/rake``` is probably better.

jibsheet (approved):

e0d (approved):
nit, but approve.

playbooks/roles/forum/tasks/deploy.yml
robrap: @e0d: It was mainly copy/paste from [discovery code](https://github.com/edx/configuration/blob/master/playbooks/roles/discovery/tasks/main.yml#L97-L106).

Just tell me your preference and I'm happy to fix.

PR #3684, opened 2017-02-09
added celery queue variable

syed-awais-ali:
Configuration Pull Request
---


jibsheet:
You need to use this somewhere so it's written out into the env (or auth, I don't know which you want) files for lms and cms.

Look at how EDXAPP_CELERY_BROKER_VHOST is defined, but then used in generic_env_config to actually be written to the file.

jibsheet (commented):
If this is going in both cms and lms, use generic_env_config (you can chase down that it's used in both places if you're interested).

feanil (approved):

PR #3685, opened 2017-02-09
Make sure we have pip and pip3 at the right version pointing to the r…

cpennington:
…ight python

Configuration Pull Request
---


docker/build/go-agent/Dockerfile
```diff
@@ -50,15 +50,13 @@ RUN printf "\ngo      ALL=(ALL:ALL) NOPASSWD: /usr/bin/pip3, /usr/local/bin/pip3
 # Must upgrade to latest before pinning to work around bug
 # https://github.com/pypa/pip/issues/3862
```
feanil: @cpennington is this bug no longer an issue?

cpennington: It appears it's not.

macdiesel (approved):

PR #3686, opened 2017-02-09
TNL-6278 Adds in xvfb and firefox to tools_jenkins for bok-choy tests

jlajoie:
Adds the xvfb and firefox packages to tools_jenkins to support ora2 bok-choy testing.

Tested building with [precise](https://hub.docker.com/r/edxops/precise-common/) and [trusty](https://hub.docker.com/r/edxops/trusty-common/)

Should be merged alongside https://github.com/edx/jenkins-job-dsl-internal/pull/73 and https://github.com/edx/jenkins-job-dsl/pull/117

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/tools_jenkins/defaults/main.yml
+++ b/playbooks/roles/tools_jenkins/defaults/main.yml
@@ -1,4 +1,5 @@
 ---
+
 jenkins_tools_version: "1.651.3"
 
 jenkins_tools_plugins:
@@ -26,7 +27,7 @@ jenkins_tools_plugins:
   - { name: "tmpcleaner", version: "1.2" }
   - { name: "jobConfigHistory", version: "2.15" }
   - { name: "build-timeout", version: "1.17.1" }
-  - { name: "timestamper", version: "1.8.4" } 
+  - { name: "timestamper", version: "1.8.4" }
   - { name: "postbuild-task", version: "1.8" }
   - { name: "notification", version: "1.10" }
   - { name: "shiningpanda", version: "0.23" }
--- a/docker/build/tools_jenkins/Dockerfile
+++ b/docker/build/tools_jenkins/Dockerfile
@@ -1,4 +1,4 @@
-FROM edxops/precise-common:latest
+FROM edxops/trusty-common:latest
 MAINTAINER edxops
 
 USER root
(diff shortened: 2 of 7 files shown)


playbooks/roles/tools_jenkins/defaults/main.yml
```diff
@@ -1,4 +1,8 @@
 ---
+jenkins_deb_pkgs:
+    - xvfb
+    - firefox
```
jibsheet: We pin some specific versions of FireFox in edxapp, @estute probably has the best current knowledge of what we pin.  Should we be pinning this to the same versions?  Otherwise there's a chance you'll get a version of firefox that plays poorly with the tests you're writing.

jibsheet (commented):
I want to sort out what version of firefox to target, but otherwise this looks good to go.

playbooks/roles/tools_jenkins/defaults/main.yml
jibsheet: @estute sorry - extra context - this is to assist with moving ORA2 jobs and running tests.  I suspect there will be some fallout of the worker conversation next week that impacts this long-term, but in the short term I want to make sure they get a sane default firefox that we know works with bokchoy and other things.

estute: @jlajoie we are using pinned versions of FireFox for bok-choy testing (version bumps can create very sporadic results wrt flaky tests). We are currently using FireFox 42 but should cut over to FireFox 45 in the next couple of days. Perhaps you could just include the browser play, which installs the browser and configures xvfb to run on boot (https://github.com/edx/configuration/blob/master/playbooks/roles/browsers/tasks/main.yml)

estute: Also, installing xfvb as a default service on init should obviate the need for the xvfb plugin, but depending on where this job lands, that may not be tenable. @jibsheet ?

jibsheet: Using the role makes sense, although we should probably chat about using workers, since it seems more and more that's what we want here.

jlajoie: For what it's worth I was able to get the expected results with xvfb running as a service through init.d, so I can go ahead and remove the plugin and include the role. 

It sounds like there's still some discussion about how this should be handled in the future once there are workers to execute these type of tests, but as of yet seems undecided? Is this an acceptable approach in the meantime knowing that this should be refactored down the road?

jibsheet: Yes, for now you should switch to using the browser role - Stu and I are meeting to discuss changes to how we would run some other worker related things on Monday, so I wanted to make sure we touch on this then.

estute:
@jlajoie aside from the upstart issue with Docker, LGTM

jibsheet:
@jlajoie Can you rebase this now that my fix for the jenkins_master role has landed?

jlajoie:
Can do

jibsheet:
ship it 👍

PR #3687, opened 2017-02-09
enable rpc service for devstacks

estute:
Starting with Ficus (first devstack to run Xenial and therefore defaulting to use systemd for init), I have run into multiple FS locks when running tests (while locally running a host system running systemd). 

@edx/devops I would like your opinion: Should this be something contained within the play, or instructions for Linux users?
CC @benpatterson

### DIFF
--- a/playbooks/roles/local_dev/tasks/main.yml
+++ b/playbooks/roles/local_dev/tasks/main.yml
@@ -132,3 +132,10 @@
     line: "127.0.0.1 localhost preview.localhost"
     state: present
   become: yes
+
+# Enable the rpc-stat service to handle NFS locks correctly
+- name: enable and start rpc-stat service
+  systemd:
+    name: rpc-statd
+    enabled: yes
+    state: started


jibsheet:
This seems fine to me, but I would want to be sure from @nedbat and @gsong that it's either a no-op or a benefit for OS-X users.

gsong:
@estute @jibsheet Is there some way to test this change on OS X/macOS?

jibsheet:
@gsong I'd probably just take a normal ficus devstack running on an OS-X host and run the systemctl task his play does and then see if tests are better/worse.  Stu can speak to which suites were breaking on him (python or bokchoy, I assume bokchoy).

estute:
@gsong I was seeing this when running the bokchoy servers. Running "paver test_bokchoy --serversonly" was causing an fslock conflict

openedx-webhooks:
Although this pull request is already closed, I'm still watching it for updates.

There is nothing you have to do. No action is needed from your side.  Thanks.

PR #3688, opened 2017-02-09
Handle the edxapp worker case with two ASGs but no ELBs

jibsheet:
Previously it would exit without a message but no inventory.

### DIFF
--- a/playbooks/active_instances_in_asg.py
+++ b/playbooks/active_instances_in_asg.py
@@ -26,6 +26,7 @@ import botocore.exceptions
 import sys
 from collections import defaultdict
 from os import environ
+from itertools import chain
 
 class ActiveInventory():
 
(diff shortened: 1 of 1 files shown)


playbooks/active_instances_in_asg.py
```diff
@@ -61,6 +62,12 @@ def run(self,asg_name):
                 # When we have more than a single active ASG, we need to bail out as we don't know what ASG to pick an instance from
                 print("Multiple active ASGs - unable to choose an instance", file=sys.stderr)
                 return
+            # If we found no active groups, because there are no ELBs (edxapp workers normally)
+            # print a sensible reason why we failed
+            elbs = list(chain.from_iterable([group['LoadBalancerNames'] for group in matching_groups]))
+            if len(active_groups) == 0 and len(elbs) == 0:
```
jibsheet: More pythonic ways to express lines 67/68 would be interesting

jibsheet:
@edx/devops please review
FYI @rlucioni

playbooks/active_instances_in_asg.py
rlucioni: For line 68:
```py
if not (active_groups or elbs):
    ...
```

jibsheet: Thanks, I don't know why I was so fixated on using len()

fredsmith (approved):

feanil (approved):
This implies that we can't use this inventory script to run tasks on the workers.  Is the plan to use ec2.py in the cases where we need to run things against the workers?

jibsheet:
@feanil We can use it on the workers, we only get in here if there is more than 1 ASG.  There's a ticket up for grooming to write something we can use to properly target workers.  This was a patch to try and give them a better error message while we fix the underlying issue.

PR #3692, opened 2017-02-13
Updated credentials, discovery and ecommerce Dockerfiles

clintonb:
- Added file to allow bower to install as root
- Added xvfb, firefox, and gettext packages for testing and local development

### DIFF
--- a/docker/build/ecommerce/Dockerfile
+++ b/docker/build/ecommerce/Dockerfile
@@ -1,10 +1,15 @@
+# docker build -f docker/build/ecommerce/Dockerfile . -t edxops/ecommerce
 FROM edxops/xenial-common:latest
 MAINTAINER edxops
 
-RUN apt-get update
-
 ADD . /edx/app/edx_ansible/edx_ansible
+
 WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+RUN echo '{ "allow_root": true }' > /root/.bowerrc
+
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/ecommerce @edx/devops the tests for the Catalog Service used to finish in about 7-8 minutes. Since moving to testing on Docker, they now take about 15 minutes. Much of that time is due to the tests being slower (perhaps because of host file system?); however, we can start chipping away at the run time by installing these packages in the container rather than every time we run a test.

docker/build/ecommerce/Dockerfile
```diff
@@ -1,11 +1,16 @@
+# docker build -f docker/build/ecommerce/Dockerfile . -t edxops/ecommerce
```
rlucioni, line 1: Did you leave this here intentionally?

rlucioni (approved):
Looks good. Will you be uploading updated images when this merges?

docker/build/ecommerce/Dockerfile
clintonb: Yes, for convenience.

clintonb:
@rlucioni already pushed.

feanil:
https://github.com/edx/ecommerce/pull/1026#discussion_r89206094

feanil:
@clintonb I feel like you have forgotten this conversation ^

PR #3700, opened 2017-02-15
Add ES migration scripts

maxrothman:
Configuration Pull Request
---

### DIFF
--- a/util/elasticsearch/verify-index-copy.py
+++ b/util/elasticsearch/verify-index-copy.py
@@ -1,0 +1,346 @@
+# -*- coding: utf-8 -*-
+"""
+Verifies that an index was correctly copied from one ES host to another.
+"""
+
+import itertools
+import pprint
+import random
+
+from deepdiff import DeepDiff
+from elasticsearch import Elasticsearch
+from elasticsearch.helpers import scan
+from argparse import ArgumentParser
+
+
+description = """
+Compare two Elasticsearch indices
+"""
+
+SCAN_ITER_STEP = 50
+SCAN_MATCH_THRESHOLD = .9
+
+RANDOM_CHECK_SIZE = 10
+RANDOM_CHECKS_BEFORE_RESET = 100
+
+def parse_args():
+    """
+    Parse the arguments for the script.
+    """
+    parser = ArgumentParser(description=description)
+
+    parser.add_argument(
+        '-o', '--old', dest='old', required=True, nargs=2,
+        help='Hostname and index of old ES host, e.g. https://localhost:9200 content'
+    )
+    parser.add_argument(
+        '-n', '--new', dest='new', required=True, nargs=2,
+        help='Hostname of new ES host, e.g. https://localhost:9200 content'
+    )
+    parser.add_argument(
+        '-s', '--scan', dest='scan', action="store_true",
+        help='Run a full scan comparison instead of a random selection.'
+    )
+    parser.add_argument(
+        '-c', '--check-percentage', dest='check_percentage', type=float, default=.1,
+        help='Percentage of randomly found docs to check between old and new indices (default: .1)'
+    )
+
+    return parser.parse_args()
+
+
+def grouper(iterable, n):
+    """
+    Collect data into fixed-length chunks or blocks
+    from the import itertools recipe list: https://docs.python.org/3/library/itertools.html#recipes
+    """
+    # grouper('ABCDEFG', 3, 'x') --> ABC DEF Gxx"
+    args = [iter(iterable)] * n
+    return itertools.izip_longest(*args)
+
+
+def docs_match(old_doc, new_doc):
+    """
+    Return True if the the docs match, minus the ignorable fields
+
+    Args:
+        old_doc: a dict of an elasticsearch doc from the old cluster
+        new_doc: a dict of an elasticsearch doc from the new cluster
+    """
+    """
+    example doc:
+    {'dictionary_item_added': {
+        "root['_source']['_id']",
+        "root['_source']['abuse_flaggers']",
+        "root['_source']['anonymous']",
+        "root['_source']['anonymous_to_peers']",
+        "root['_source']['at_position_list']",
+        "root['_source']['author_username']",
+        "root['_source']['closed']",
+        "root['_source']['comment_count']",
+        "root['_source']['historical_abuse_flaggers']",
+        "root['_source']['pinned']",
+        "root['_source']['thread_type']",
+        "root['_source']['visible']",
+        "root['_source']['votes']",
+        "root['found']"},
+     'dictionary_item_removed': {
+        "root['_source']['id']",
+        "root['_source']['thread_id']",
+        "root['_source']['votes_point']",
+        "root['exists']"},
+     'values_changed': {
+        "root['_index']": {
+            'new_value': u'content_20170324145539907',
+            'old_value': u'content_20151207225034'},
+        "root['_source']['body']": {
+            'new_value': u'encryption neglect hypothesize polluters wining pitiably prophetess apostrophe foretelling assignments diaphragms trustees scroll scruffs shrivels characterizes digraph lasted sharked rewind chamoix charier protoplasm rapports isolated upbraid mortgaged cuddled indefinitely sinful insaner slenderized cemetery deject soundly preventable',
+            'old_value': u'embellishing orbitals complying alternation welching sepulchered grate blench placenta landslide dependance hurdle predicted chaplet earsplitting assess awol necrosis freeways skipper delicatessen sponsorship bellboys antiseptics gabardines admittedly screechier professional roughness educations nutting valences iridescence deductions'},
+        "root['_source']['title']": {
+            'new_value': u'southpaw afterward playgoers roughed requites arrived byplay ninetieth textural rental foreclosing',
+            'old_value': u'guttersnipes corduroys ghostly discourtesies'},
+        "root['_source']['updated_at']": {
+            'new_value': u'2017-03-29T18:51:19Z',
+            'old_value': u'2017-03-28T12:58:02Z'},
+        "root['_version']": {
+            'new_value': 20,
+            'old_value': 1}}}
+"""
+    ignorable_fields = [
+        "root['exists']",
+        "root['found']",
+        "root['_index']",
+        "root['updated_at']",
+        "root['_version']",
+        "root['_score']",
+    ]
+    diff_types = ['dictionary_item_added', 'dictionary_item_removed', 'values_changed']
+    diff_doc = DeepDiff(old_doc, new_doc)
+
+    if 'values_changed' not in diff_doc:
+        diff_doc['values_changed'] = set()
+
+    #if this fails something is horribly wrong
+    if set(diff_doc.keys()) != set(diff_types):
+        print 'ERROR: expected to be diffing dictionaries, got something else! id: {}'.format(
+            new_doc['_id'])
+
+    for diff_type in diff_types:
+        for field in ignorable_fields:
+            if diff_type in diff_doc:
+                #values_changed is a set, the other two are dicts
+                if isinstance(diff_doc[diff_type], set):
+                    diff_doc[diff_type].discard(field)
+                else:
+                    diff_doc[diff_type].pop(field, None)
+
+    return all(len(diff_doc[diff_type]) == 0 for diff_type in diff_types)
+
+
+def find_matching_ids(es, index, ids, docs):
+    """
+    Finds out how many of the ids in the given ids are in the given index in the given
+    ES deployment.
+
+    We also compare documents to ensure that those still match, skipping a few fields
+    (see docs_match() for which ones).
+
+    Args:
+        es - Elasticsearch instance corresponding to the cluster we want to check
+        index - name of the index that we want to check
+        ids - a list of dictionaries of the form {'_id': <id>} of the ids we want to check.
+        docs - a dictionary of the form {'<id>': document}, where "document"s are full ES docs
+    """
+    body = {'docs': ids}
+
+    search_result = es.mget(index=index, body=body)
+    matching = 0
+    for elt in search_result['docs']:
+        # Checks whether or not there was a document matching the id at all.
+        # 'exists' is 0.9.x
+        # 'found' is 1.5.x
+        if elt.get('exists', False) or elt.get('found', False):
+            if docs_match(docs[elt['_id']], elt):
+                matching += 1
+            else:
+                print 'FAILURE: Documents with id {id} do not match: '.format(
+                    id=elt['_id']
+                ) + repr({'diff': DeepDiff(docs[elt['_id']], elt), 'new': elt, 'old': docs[elt['_id']]})
+        else:
+            print 'ERROR: Document with id {id} missing: {doc}'.format(
+                id=elt['_id'], doc=docs[elt['_id']]
+            )
+    return matching
+
+
+def scan_documents(old_es, new_es, old_index, new_index):
+    """
+    Scan for matching documents
+
+     In order to match the two indices without having to deal with ordering issues,
+     we pull a set of documents from the old ES index, and then try to find matching
+     documents with the same _id in the new ES index. This process is batched to avoid
+     making individual network calls to the new ES index.
+    """
+
+    matching = 0
+    total = 0
+    old_iter = scan(old_es, index=old_index)
+    for old_elts in grouper(old_iter, SCAN_ITER_STEP):
+
+        old_elt_ids = []
+        old_elt_docs = {}
+        for elt in old_elts:
+            if elt is not None:
+                old_elt_ids.append({'_id': elt['_id']})
+                old_elt_docs[elt['_id']] = elt
+
+        matching += find_matching_ids(new_es, new_index, old_elt_ids, old_elt_docs)
+        total += len(old_elt_ids)
+        if total % 100 == 0:
+            print 'processed {} items'.format(total)
+
+    ratio = float(matching)/total
+    print "{}: scanned documents matching ({} out of {}, {:.6}%)".format(
+        'OK' if ratio > SCAN_MATCH_THRESHOLD else 'FAILURE', matching, total, ratio * 100
+    )
+
+
+def random_checks(old_es, new_es, old_index, new_index, total_document_count, check_percentage):
+    """
+    Check random documents
+    This is meant to be a random search trying to spot checks on whether
+    or not data was moved over correctly. Runs a lot faster than the full scan.
+    """
+
+    total = 0
+    matching = 0
+    current_offset = -1
+    while float(total) / total_document_count < check_percentage:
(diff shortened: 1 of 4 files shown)


util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,58 @@
+import requests
+from argparse import ArgumentParser
+
+description = """
+Compare two Elasticsearch indices
```
robrap: Should this compare mappings as well?

util/elasticsearch/incremental-reindex.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+#
+# Thin wrapper around rake search:catchup.
```
robrap: Maybe mentioned that this is specific to cs_comments_service.

util/elasticsearch/incremental-reindex.sh
```diff
@@ -0,0 +1,38 @@
...
+#  source ../forum_env; ./incremental-reindex.sh INDEX [WINDOW] [SLEEP_TIME] [BATCH_SIZE]
+#
+# Args:
+#   INDEX       The index to re-index
+#   WINDOW      Number of minutes ago to re-index from
+#   SLEEP_TIME  Number of seconds to sleep between re-indexing
+#   BATCH_SIZE  Number of documents to index per batch
+#
+# Example:
+#   ./incremental-reindex.sh content -1hour
```
robrap: How does `-1hour` work?  This gets translated into minutes?

util/elasticsearch/incremental-reindex.sh
```diff
@@ -0,0 +1,38 @@
...
+# indexing can only yield correct results, the only risk of setting WINDOW too large
+# is poor performance.
+#
+# Usage:
+#  source ../forum_env; ./incremental-reindex.sh INDEX [WINDOW] [SLEEP_TIME] [BATCH_SIZE]
+#
+# Args:
+#   INDEX       The index to re-index
+#   WINDOW      Number of minutes ago to re-index from
+#   SLEEP_TIME  Number of seconds to sleep between re-indexing
```
robrap: Just a note that the WINDOW doesn't account for the SLEEP_TIME.  If you had it sleep for an hour with a window of 30 minutes, you'd miss documents.  Maybe add a comment, or add a dummy check that `SLEEP_TIME < WINDOW * 60`.

util/elasticsearch/verify-index-copy.py
robrap: FYI: I'm not reviewing this because I think it is under development.  Let me know when it is ready.

util/elasticsearch/incremental-reindex.sh
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env bash
```
robrap: Do we need an incremental script using logstash as well for teams?  Maybe that is part of a later story?

robrap (changes requested):

util/elasticsearch/incremental-reindex.sh
maxrothman: I have a strong hunch that we we can move the entire teams and notes indices in a small enough amount of time that we can do it during downtime (i.e. <5m), so such a script would not be necessary.

robrap: I'm not clear on what "during downtime" means?  But, what I think you are saying is that we can copy everything both before and after we are running just the updated instances.  Is that what you are saying?

maxrothman: My plan for those apps is the following:
* Remove instances from the loadbalancer
* move the index to the new cluster
* Add the instances to the loadbalancer

Since the index is moved completely during downtime, there's no need for a catchup script.

robrap: Do you mean search downtime or app downtime?  Are you saying we will have LMS downtime for this migration? I didn't think we would have downtime.

maxrothman: I was under the impression we could get downtime if we needed it. We could also just do search downtime if that's possible. As long as nothing can write to the index.

feanil:
I thought these scripts were going to go into the forums repo here(https://github.com/edx/cs_comments_service/tree/master/scripts/es) replacing the old stream2es scripts.

maxrothman:
That was the original plan, but since `copy-index.sh` will be used for all 3 apps and not just forums, it seemed to make more sense to put them in config in the same way we did for the mongo migration.

feanil:
Don't forget to PR a removal of the files in forums and to add a readme there to point to these scripts when they are finalized.

maxrothman:
Yep, already have the PR, waiting to merge it.

dianakhuang:
@maxrothman I added a first round of the script. Haven't had a chance to test it on something like load-test yet, but I figured I'd let you take a look at it to make sure it looks the way you want.

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,209 @@
...
+    )
+    parser.add_argument(
+        '-m', '--max-scan', dest='max_scan', type=int, default=100000,
+        help='Number of documents to scan through in simple scan. (default: 100000)'
+    )
+
+    return parser.parse_args()
+
+
+def grouper(iterable, n, fillvalue=None):
```
maxrothman: fillvalue does not appear to be used.

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,209 @@
...
+    parser.add_argument(
+        '-n', '--new', dest='new', required=True, nargs=2,
+        help='Hostname of new ES host, e.g. https://localhost:9200 content'
+    )
+    parser.add_argument(
+        '-mt', '--match-threshold', dest='match_threshold', type=float, default=.9,
+        help='Percentage of matching docs between old and new indices req. for an OK (default: .9)'
+    )
+    parser.add_argument(
+        '-cp', '--check-percentage', dest='check_percentage', type=float, default=.1,
```
maxrothman: Nit: single-dash options should only be one letter long

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,209 @@
...
+        '-o', '--old', dest='old', required=True, nargs=2,
+        help='Hostname and index of old ES host, e.g. https://localhost:9200 content'
+    )
+    parser.add_argument(
+        '-n', '--new', dest='new', required=True, nargs=2,
+        help='Hostname of new ES host, e.g. https://localhost:9200 content'
+    )
+    parser.add_argument(
+        '-mt', '--match-threshold', dest='match_threshold', type=float, default=.9,
+        help='Percentage of matching docs between old and new indices req. for an OK (default: .9)'
```
maxrothman: FWIW we're not planning on using this option. We're just going to run at the error rate in the log.

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,209 @@
...
+
+    search_result = es.mget(index=index, body=body)
+    matching = 0
+    for elt in search_result['docs']:
+        # in ES 0.9.x
+        if 'exists' in elt:
+            matching += 1 if elt['exists'] else 0
+        # in ES 1.x
+        elif 'found' in elt:
+            matching += 1 if elt['found'] else 0
```
maxrothman: It looks like this function is just checking whether the id is found, but doesn't compare the content of the documents.

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,209 @@
...
+        '-mt', '--match-threshold', dest='match_threshold', type=float, default=.9,
+        help='Percentage of matching docs between old and new indices req. for an OK (default: .9)'
+    )
+    parser.add_argument(
+        '-cp', '--check-percentage', dest='check_percentage', type=float, default=.1,
+        help='Percentage of randomly found docs to check between old and new indices (default: .1)'
+    )
+    parser.add_argument(
+        '-m', '--max-scan', dest='max_scan', type=int, default=100000,
+        help='Number of documents to scan through in simple scan. (default: 100000)'
```
maxrothman: Is there a particular reason for this option? I don't really understand the use case

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,209 @@
...
+        ids = [{'_id': elt['_id']} for elt in results['hits']['hits']]
+        matching += find_matching_ids(new_es, new_index, ids)
+        num_elts = len(ids)
+        current_checked += num_elts
+        current_offset += num_elts
+
+        if current_checked % 100 == 0:
+            print 'processed {} items'.format(current_checked)
+
+    ratio = float(matching) / current_checked
```
maxrothman: Nit: you can do the cast to float in the argument definition

maxrothman (commented):
It seems like there's some confusion about the requirements. I'm looking the script to do a full scan OR a random check, configurable via an option, not one then the other.

It'd also be good to print the IDs and bodies of non-matching documents, rather than just the total.

util/elasticsearch/verify-index-copy.py
dianakhuang: If we're going to scan every document, why do we even need to do the random check?

maxrothman: The random check trades confidence for speed. We'll run the random check before making the cluster live, then do the slow, full scan afterwards while the cluster is active.

TL/DR: we want to gain some confidence before go-live and more confidence later asynchronously.

dianakhuang:
@maxrothman I believe I have addressed your comments, so this should be ready for testing.

util/elasticsearch/verify-index-copy.py
```diff
@@ -68,85 +65,77 @@ def find_matching_ids(es, index, ids):
...
+                print 'Document with id {id} does not match body: {body}'.format(
+                    id=elt['_id'], body=bodies[elt['_id']]
+                )
+                print 'Found body: {body}'.format(
+                    body=elt['_source']['body']
+                )
+        else:
+            print 'Document missing with id: {id}, body: {body}'.format(
+                id=elt['_id'], body=bodies[elt['_id']]
+            )
```
maxrothman: Nit: prefix the logging lines with something easy to search, like I did on https://github.com/edx/configuration/pull/3700/files#diff-7a0dc9e15666855efef7dce951d51334R198

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,238 @@
...
+        '-n', '--new', dest='new', required=True, nargs=2,
+        help='Hostname of new ES host, e.g. https://localhost:9200 content'
+    )
+    parser.add_argument(
+        '-s', '--scan', dest='scan', action="store_true",
+        help='Run a full scan comparison instead of a random selection.'
+    )
+    parser.add_argument(
+        '-c', '--check-percentage', dest='check_percentage', type=float, default=.1,
+        help='Percentage of randomly found docs to check between old and new indices (default: .1)'
```
robrap, line 46: I was imagining that we would need an argument for a timestamp and only compare documents updated before that timestamp.  @maxrothman: I think the detailed migration plan for forums would help clarify these requirements.

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,238 @@
...
+                id=elt['_id'], body=bodies[elt['_id']]
+            )
+    return matching
+
+
+def scan_documents(old_es, new_es, old_index, new_index):
+    """
+    Scan for matching documents
+
+     In order to match the two indeces without having to deal with ordering issues,
```
robrap: typo: indeces

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,238 @@
...
+     we pull a set of dcouments from the old ES index, and then try to find matching
```
robrap: typo: dcouments

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Args:
+        es - Elasticsearch instance corresponding to the cluster we want to check
+        index - name of the index that we want to check
+        ids - a list of dictionaries of the form {'_id': <id>} of the ids we want to check.
+        bodies - a dictionary of the form {'<id>': '<text body>'}
+    """
+    body = {'docs': ids}
+
+    search_result = es.mget(index=index, body=body)
```
robrap: Out of curiosity, are there parts of the document we are not comparing?  If there is, can you make that explicit in the comment so it is either documented or we can choose to compare.

util/elasticsearch/verify-index-copy.py
```diff
@@ -0,0 +1,238 @@
+"""
+Verifies that an index was correctly copied from one ES host to another.
```
robrap, line 3: As discussed, should compare mappings.

robrap (changes requested):

util/elasticsearch/verify-index-copy.py
dianakhuang: Yup, there is a bunch of metadata that we're not comparing here. I'll add a comment.

util/elasticsearch/verify-index-copy.py
maxrothman: Unfortunately this is a bit of a chicken-and-egg thing: I can't finish making a detailed migration plan until I've iterated and tested one using this script. I also suspect this script will be iterated on as the plan development continues.

What would be the goal of the timestamp filtering?

robrap: I guess I don't see it that way, but we all work differently.  I would write my initial detailed plan and review it and then test and iterate.

A rough detailed plan would be:
- Create new ES cluster.
- Create new cs_comments_service instance (AMI earlier).
- Run bin/rake search:initialize on new ES cluster to create the alias and mappings.
- Run logstash copy from old ES to new ES at time=T.
- Validate documents using @dianakhuang script, comparing old ES to new ES for documents updated before T (or T-5 minutes).
- Make one post manually (for later validation).
- Switch auto-scaling group to new instances.
- Wait for draining of old instances.
- Run catchup from T - 1 hour.
- Double check that new manual post is searchable after catchup.

You can see in step above how I would use the timestamp, to avoid comparing to changes that continue after the copy.

maxrothman: Here was broadly what I was thinking:
* Create new ES cluster.
* Create new cs_comments_service instance (AMI earlier).
* Run bin/rake search:initialize on new ES cluster to create the alias and mappings.
* Run logstash copy from old ES to new ES at time=T.
* Run catchup for T - however long it took to run the copy
* Validate randomly sampled documents using script from old ES to new ES
* Make one post manually (for later validation).
* Switch auto-scaling group to new instances.
* Wait for draining of old instances.
* Run catchup from T - however long it's been since the last catchup.
* Double check that new manual post is searchable after catchup.

Since the time it takes to run the first catchup will be relatively small, I'm not worried about the few documents added in that amount of time, thus I don't think timestamp search is necessary. I'm not expecting 0 errors from the script, just close to 0.

util/elasticsearch/verify-index-copy.py
```diff
@@ -176,6 +179,41 @@ def random_checks(old_es, new_es, old_index, new_index, total_document_count, ch
     )
 
 
+def check_mappings(old_mapping, new_mapping):
```
maxrothman: FWIW Difflib might make this much more simple

util/elasticsearch/verify-index-copy.py
```diff
@@ -176,6 +179,41 @@ def random_checks(old_es, new_es, old_index, new_index, total_document_count, ch
...
+    """
+    Verify that the two mappings match in terms of keys and properties
+    """
+    mappings_match = True
+    # Check mapping types
+    for old_key, new_key in zip(old_mapping.keys(), new_mapping.keys()):
```
maxrothman: Doesn't this mean you'll get the keys in arbitrary order, meaning there's no guarantee you'll get matching keys together?

dianakhuang: Hmm, according to @robrap the order is important for these keys, and what we get out of the elasticsearch library has already been converted into a dictionary. I'll talk it over and see what our options are.

util/elasticsearch/verify-index-copy.py
dianakhuang: Difflib seems to be more about comparing linear sequences rather than weird nested structures like this. We're not doing a straight comparison on everything, because newer ES has more fields in the mappings than the old.

util/elasticsearch/verify-index-copy.py
robrap: @maxrothman 
1. If you have a couple of problems, how do you know whether or not you can ignore them?  Can you see the timestamp of the document, and will you know if it will be fixed during the reindex?
2. I thought you had a later step of doing a full validation?  Doesn't that have to end at a specific time, or does the script just make sure that all docs in the old are in the new?  @dianakhuang?

util/elasticsearch/verify-index-copy.py
robrap: @dianakhuang: I don't see the comment.  Also, as @maxrothman mentioned the other day, we need to determine if we can really ignore this other metadata?  
Or, did you add but just not update the comment and I am missing the code?

dianakhuang: I believe the convo with @maxrothman was specifically about mappings. For this, I believe we agreed that checking the body and the id was sufficient. I added the comment to the docstring here: https://github.com/edx/configuration/pull/3700/files#diff-7a0dc9e15666855efef7dce951d51334R66

util/elasticsearch/verify-index-copy.py
maxrothman: 1. Conflicting documents will have to be hand-inspected. Anything the script can do to make that easier (e.g. deepdiff) would be appreciated.
2. I forgot that in my plan above. After the whole thing is done, I will run the full validation. It'll scan every document in the old cluster and make sure they're in new (and output conflicts when there are any)

robrap (approved):

robrap:
@feanil @MichaelRoytman: What should happen to this PR?  Should it just be closed?  Does anything need to be merged or fixed anywhere?

michaelroytman:
@robrap I'm not sure which of (if any) of these scripts would be useful moving forward, since I was not directly involved in resolving the issues with the elasticsearch indexes during the last push that you and @feanil were involved in with notes and edx-search. It looks like from the runbook updates Feanil made that none of these resources are necessary for a re-index. Is that true for the forums as well? Or does it remain as complicated as described in the original runbook, because those scripts are used there.

feanil:
The forums is as complicated as described here, so we should merge this since that part of the run book mentions this.

feanil:
I've squased this, I'll merge once it's passing tests.

PR #3705, opened 2017-02-16
We don't use this bot.

feanil:
dedxsy

@edx/devops


maxrothman (approved):
sad panda

jibsheet (approved):
Can we remove
```
playbooks/sample_vars/passwords.yml
58:XSY_PASSWORD: !!null
```

And then

image

PR #3706, opened 2017-02-17
Source python virtual env in analytics_api_env

nadeemshahzad:
@edx/devops Please review. Thanks


e0d:
We were discussing this in Cambridge and I think we want to change the way we are doing env files for the python virtual env activation.

We want to source the /home/venv/service/bin/activate file so that PYTHON_HOME is set and the bash prompt is updated with the virtual env name.

nadeemshahzad:
@e0d I have added virtual env activation in env file

e0d (approved):

PR #3709, opened 2017-02-22
add the ec2 plugin + deps

jibsheet:
https://wiki.jenkins-ci.org/display/JENKINS/Amazon+EC2+Plugin


jibsheet:
@edx/devops

feanil (approved):
NIT: I know we haven't been doing this before, but is it possible to add reason why we need to add the plugin as a comment in this file?  I know it might change since once it's available anyone can use it but having some info mapping these to reasons why they're here seems useful.

jibsheet:
The NIT seems reasonable.  I've also been thinking about alphabetizing or at least organizing the file (one issue with organizing is that multiple plugins have git as a dep, etc)

PR #3710, opened 2017-02-23
rabbitmq memory alerting

nadeemshahzad:


playbooks/roles/rabbitmq/templates/edx/app/rabbitmq/log-rabbitmq-memory.py.j2
```diff
@@ -0,0 +1,22 @@
...
+log_file="{{ rabbitmq_log_dir }}/rabbitmq.memory.log"
+
+def run_cmd(cmd):
+    process=Popen(cmd,stdout=PIPE,stderr=PIPE,shell=True)
+    output=process.communicate()[0]
+    return output
+
+if __name__=='__main__':
+
+    process_mem=run_cmd("/usr/sbin/rabbitmqctl status | grep total | awk -F',|}' 'NR==1{print $2}'")
```
feanil: This python script seems like overkill, why not just make it a shell script that runs this command and appends the output to the relevant file along with a timestamp?

feanil:
You'll also need to setup a log-rotate rule for the new output file that you are generating.

feanil:
This generally looks good, but you're still missing a log rotation rule for the new file.

nadeemshahzad:
@feanil Log file is going to "rabbitmq_log_dir" on which log rotation is already working.

playbooks/roles/rabbitmq/defaults/main.yml
```diff
@@ -5,6 +5,7 @@ rabbitmq_app_dir: "{{ COMMON_APP_DIR }}/rabbitmq"
 rabbitmq_data_dir: "{{ COMMON_DATA_DIR }}/rabbitmq"
 rabbitmq_log_dir: "{{ COMMON_LOG_DIR }}/rabbitmq"
 rabbitmq_cfg_dir: "{{ COMMON_CFG_DIR }}/rabbitmq"
+rabbitmq_mem_log: "{{ COMMON_LOG_DIR }}/rabbitmq.memory.log"
```
feanil: Memory log location should go under `rabbitmq_log_dir`

playbooks/roles/rabbitmq/templates/etc/logrotate.d/rabbitmq.j2
```diff
@@ -1,6 +1,6 @@
 # We want to hit the top level queues and any vhost queues
 # such as notifier and fulfillment
-{{ rabbitmq_log_dir }}/*.log {{ rabbitmq_log_dir }}/*/*.log {
+{{ rabbitmq_log_dir }}/*.log {{ rabbitmq_log_dir }}/*/*.log {{ rabbitmq_mem_log }}{
```
feanil: Onec the memory file is under the `rabbitmq_log_dir` you don't need to make this change.

feanil (changes requested):
@nadeemshahzad good point, if you just put your file under the `rabbitmq_log_dir` it should get picked up automatically since it inds in `.log`

nadeemshahzad:
@edx/devops Please review this

feanil (approved):

PR #3711, opened 2017-02-23
turn off mongo auth for platform jenkins

estute:


jibsheet:
Bah- sorry.  I went through the mental / typing exercises with devstacks and sandboxes and forgot about the jenkins worker mongo.

jibsheet (approved):

PR #3722, opened 2017-03-01
Improve docs and become www-data

jibsheet:
This wasn't previously required, which is odd, but we cannot read lms.*.json
without it.

DEVOPS-5543


jibsheet:
@edx/devops

feanil (approved):

PR #3723, opened 2017-03-01
Add a playbook for installing and configuring Graphite.

tobz:
This playbook will install `carbon-cache` and `graphite-api` to provide a minimalistic starting point for metrics collection.  A server would be ready to accept metrics (from anything that could speak the Graphite line protocol, such as `statsd`) and be a target for a Grafana installation after running this playbook on its own.

It currently installs all components from their upstream source (albeit with defaults that point to tags, and not master/development branches).  It uses systemd, and so is tailored specifically to Ubuntu 16.04.

There is a small "master" playbook -- `edx-monitoring` -- for running the role, and in the future, supporting the culmination of other playbooks to build out an all-in-one monitoring server: Graphite, Grafana, etc.

Configuration Pull Request
---


tobz:
Point of order: I cribbed this playbook from an existing playbook (https://github.com/azavea/ansible-graphite) and so the LICENSE from the original project is included, with an amendment at the top pointing back to the original project.

This should be kosher given that it's the Apache 2.0 license.

e0d:
@tobz high-level question, can you summarize what changes you made from the source ansible artifacts and why you chose this one?

tobz:
High-level changes:
- make it use a virtualenv
- install graphite-api over graphite-web (don't need all the web baggage if running Grafana on top)
- make it install gunicorn (because we're using nginx, so no mod_wsgi)
- use systemd (instead of SysV) to ensure both carbon-cache and graphite-api get supervised properly

I mostly chose this one because it worked off of the tagged source, had a suitable license for derivations, and was one of the smaller ones that I found.  Many of them were versions that appeared to be tailored to the individual's own needs / company needs.

playbooks/roles/graphite/handlers/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: Restart Carbon
```
e0d: We've been typically avoiding handlers as their behavior can be surprising and we typically don't do rolling upgrades.  We've been using tags to target tasks to give us flexibility around which sets are run at any invocation:

https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle+Tags

playbooks/roles/graphite/tasks/carbon.yml
```diff
@@ -0,0 +1,49 @@
+---
+- name: checkout Carbon
+  git: repo=https://github.com/graphite-project/carbon.git
+       dest={{ graphite_root }}/src/carbon
+       version={{ GRAPHITE_CARBON_VERSION }}
+
+- name: install Carbon dependencies
```
e0d: why not use the pip module?

playbooks/roles/graphite/tasks/carbon.yml
```diff
@@ -0,0 +1,49 @@
...
+            dest={{ graphite_root }}/conf/{{ item }}.conf
+  with_items:
+    - carbon
+    - storage-schemas
+    - storage-aggregation
+  notify:
+    - Restart Carbon
+
+- name: set permissions for storage
+  file: path={{ graphite_root }}/storage
```
e0d: Feels like we should separate out the storage root from the app root.  It may be that by default graphite_storage is set to {{ graphite_root }}/storage, but I'd think we'd relocate the storage at scale.

playbooks/roles/graphite/tasks/graphite-api.yml
```diff
@@ -0,0 +1,49 @@
+---
+- name: checkout Graphite API
+  git: repo=https://github.com/brutasse/graphite-api.git
```
e0d: repo should be a var I think, makes groking from defaults easier, supports forks, etc.

playbooks/roles/graphite/tasks/graphite-api.yml
```diff
@@ -0,0 +1,49 @@
...
+       dest={{ graphite_root }}/src/api
+       version={{ GRAPHITE_API_VERSION }}
+
+- name: install Graphite API dependencies
+  command: "{{ graphite_root }}/bin/pip install -r {{ graphite_root }}/src/api/requirements.txt"
```
e0d: pip module

playbooks/roles/graphite/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+
+- name: create service account for Carbon
+  user: name=carbon
+        system=yes
+        home={{ graphite_root }}
+        shell=/bin/false
+        group=graphite
+        state=present
+
+- name: create common Graphite directories
```
e0d: I like doing this using with_items and data:

https://github.com/edx/configuration/blob/master/playbooks/roles/vhost/defaults/main.yml#L18

playbooks/roles/graphite/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+- name: create Graphite source directories
+  file: path={{ graphite_root }}/src
+        state=directory
+        owner=graphite
+        group=graphite
+
+- name: install Graphite dependencies
+  apt: pkg={{ item }}
+       state=present
+  with_items:
```
e0d: use a var: https://github.com/edx/configuration/blob/master/playbooks/roles/common_vars/defaults/main.yml#L110

playbooks/roles/graphite/defaults/main.yml
```diff
@@ -0,0 +1,34 @@
+---
```
e0d, line 1: Can you add boilerplate and explanatory comment, c.f., https://github.com/edx/configuration/blob/master/playbooks/roles/vhost/tasks/main.yml#L2

playbooks/roles/graphite/templates/carbon/conf/carbon.conf.j2
```diff
@@ -0,0 +1,23 @@
+[cache]
+USER = carbon
+CACHE_WRITE_STRATEGY = max
+MAX_CACHE_SIZE = inf
+USE_FLOW_CONTROL = True
+WHISPER_FALLOCATE_CREATE = True
+MAX_CREATES_PER_MINUTE = 100
+MAX_UPDATES_PER_SECOND = 1000
+LINE_RECEIVER_INTERFACE = 0.0.0.0
```
e0d: Seems like some of these are things we'd want to var out.  Listening on 0.0.0.0 for sure seems like something we'd want to make obvious in the defaults.  Other things we'd expect ourselves or others to reasonably change should be as well.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/graphite.j2
```diff
@@ -0,0 +1,52 @@
+#
```
e0d, line 1: Play will need to include nginx.

playbooks/roles/graphite/templates/graphite-api/systemd/graphite-api.service.j2
```diff
@@ -0,0 +1,20 @@
...
+Requires=graphite-api.socket
+
+[Service]
+StandardOutput=syslog
+StandardError=syslog
+User=graphite
+Group=graphite
+WorkingDirectory={{ graphite_root }}
+Environment=GRAPHITE_API_CONFIG={{ graphite_root }}/conf/graphite-api.yaml
+ExecStart={{ graphite_root }}/bin/gunicorn -w 2 graphite_api.app:app
```
e0d: I think workers should be configurable.

playbooks/roles/graphite/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
+---
+- name: create Graphite user group
+  group: name=graphite
```
e0d: we should make this a var and the group so we can pass to templates later.

playbooks/roles/graphite/templates/graphite-api/systemd/graphite-api.service.j2
```diff
@@ -0,0 +1,20 @@
+[Unit]
+Description=Graphite-API service
+Requires=graphite-api.socket
+
+[Service]
+StandardOutput=syslog
+StandardError=syslog
+User=graphite
```
e0d: Use user/group vars defined above.

playbooks/roles/graphite/tasks/main.yml
```diff
@@ -0,0 +1,68 @@
...
+- name: initialize virtualenv for Graphite
+  command: "virtualenv {{ graphite_root }}"
+  args:
+    creates: "{{ graphite_root }}/bin/activate"
+
+- include: whisper.yml
+- include: carbon.yml
+- include: graphite-api.yml
+
+- name: Find all example configuration files
```
e0d: what's wrong with the examples?

e0d:
I think we should have a docker CI build that can run in travis:

https://github.com/edx/configuration/tree/master/docker/build

e0d (commented):
First pass.

playbooks/roles/graphite/tasks/main.yml
tobz: Nothing, really.  There's just like 10 of them and it felt like it was just a little harder than it ought to be to peek at the file listing and see the actual configuration files being used.

playbooks/roles/graphite/tasks/carbon.yml
tobz: There's a pip module? Sweet! I'll use that.

playbooks/roles/graphite/tasks/main.yml
tobz: Good point.  Will do.

playbooks/roles/graphite/defaults/main.yml
tobz: I addressed this in `tasks/main.yml` specifically.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/graphite.j2
tobz: Yup, this is now taken care of.

playbooks/roles/graphite/handlers/main.yml
tobz: I opted to stick with handlers for the time being.

tobz:
@edx/devops Since Ed is out, can someone take another pass at this for review?

fredsmith:
@tobz can you add tags on all tasks according to https://openedx.atlassian.net/wiki/display/OpenOPS/Role+Life-cycle+Tags ?  We've been doing this for all ansible code that gets merged.

You should also re-write your handlers as tasks, and make sure they're tagged appropriately.

tobz:
@fredsmith This should be all set now.

playbooks/roles/graphite/handlers/main.yml
fredsmith: I think this can go away now.

fredsmith (changes requested):
I think we can remove the handler definitions, then :+1:

tobz:
Ah right, my mistake.  Fixed.

fredsmith (approved):

fredsmith:
looks good to merge :+1:

PR #3725, opened 2017-03-01
mattdrayer/ENT-244: Add ENTERPRISE_SERVICE_URL setting

mattdrayer:
Configuration Pull Request
---

@mjfrey @bradenmacdonald @bdero @fredsmith I think this change will address the exception noted in ENT-244.


mjfrey (approved):

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -142,6 +142,7 @@ ECOMMERCE_SERVICE_CONFIG:
 
   COURSE_CATALOG_API_URL: '{{ ECOMMERCE_COURSE_CATALOG_URL }}/api/v1/'
   ENTERPRISE_API_URL: '{{ ECOMMERCE_ENTERPRISE_URL }}/api/v1/'
+  ENTERPRISE_SERVICE_URL: '{{ ECOMMERCE_ENTERPRISE_URL }}/enterprise/'
```
mjfrey: Isn't this /api/v1/enterprise?

mjfrey (changes requested):

playbooks/roles/ecommerce/defaults/main.yml
bdero: I think `ECOMMERCE_ENTERPRISE_URL` already has `/enterprise/` at the end of it, judging by the line right above.

bdero (changes requested):

playbooks/roles/ecommerce/defaults/main.yml
mattdrayer: @bdero this is what I was looking at: https://github.com/edx/configuration/pull/3725/files/869a4392448c35dee30a0a7e397f648963805173#diff-dd4f3df8a0026e7f194781ec48f0338eR134

mattdrayer: @mjfrey this is what I was referencing:  https://github.com/edx/ecommerce/blob/master/ecommerce/settings/base.py#L580

mattdrayer: I opted to leave `ENTERPRISE_API_URL` as-is in this PR versus removing it, although we are overriding the value here: https://github.com/edx/ecommerce/blob/master/ecommerce/settings/production.py#L87

...so maybe it is okay to remove?  Trying to avoid breaking something...

bdero: @mattdrayer Ah ok, I think my confusion is stemming from the default [ENTERPRISE_API_URL](https://github.com/edx/configuration/pull/3725/files#diff-dd4f3df8a0026e7f194781ec48f0338eR144) being wrong.

Can we correct it to use the new ECOMMERCE_SERVICE_URL default you made?: 
```
ENTERPRISE_SERVICE_URL: '{{ ECOMMERCE_ENTERPRISE_URL }}/enterprise/'
ENTERPRISE_API_URL: '{{ ECOMMERCE_SERVICE_URL }}/api/v1/'
```

bdero (changes requested):

playbooks/roles/ecommerce/defaults/main.yml
bdero: Sorry, NVM, I just realized it's not wrong, I'm confusing the enterprise app URLs with the ecommerce URLs..

bdero (approved):

playbooks/roles/ecommerce/defaults/main.yml
mattdrayer: @bdero I think your previous comment would work, and it emulates what is currently defined in settings/base.py + settings/production.py -- I'd be okay making this change, although I think the ideal thing to do is remove `ENTERPRISE_API_URL` entirely from config -- I was just hesitant to do this without some confirmation

mattdrayer:
@bdero see my latest commit -- this is what I am describing

bdero:
@mattdrayer Yup, I went ahead and double checked, it seems reasonable to remove it as you just did.

mattdrayer:
Okay, I'm going to move forward with this change then -- will tidy up the PR and ping DevOps for final review -- @mjfrey did you have anything further you want to see?

mjfrey (approved):

mjfrey:
👍

fredsmith (approved):

PR #3727, opened 2017-03-02
update firefox version test for jenkins workers

estute:
@jibsheet @benpatterson This should be the last missed commit.


benpatterson:
👍 

How has this not-been-failing all along though?

estute:
Because this is a merge into ficus. The packer jobs default to running on master

benpatterson:
Och i missed the branch.

But: ship it!

PR #3733, opened 2017-03-06
Add xvfb setting to prevent chrome hangs.

benpatterson:
Configuration Pull Request
---


benpatterson:
@estute @jzoldak @michaelyoungstrom 

Heads up I'm closing this, too. Platform does this in the start-up scripts somewhere IIRC, but putting it into the chrome configuration is a more scalable kind of fix. So feel free to re-implement if you have the cycles/think it's worth it.

PR #3735, opened 2017-03-06
Switch from git_add_course to standard import

jibsheet:
git_add_course uses the platform's sysadmin dashboard which assumes it
can create an xlog mongo database and write into it without auth
configured.  This is not something we want to spend more time enabling,
so switch to our 'standard' import technique.

This relies on the demo role more than I would like, but we have a
backlog task to clean that role up, and can touch this at the same time
(this role has always imported a bunch of vars from demo, so I'm not
 adding any new deps).

Test build https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/73/console

https://openedx.atlassian.net/browse/DEVOPS-5587

@edx/devops

### DIFF
--- a/playbooks/roles/testcourses/tasks/deploy.yml
+++ b/playbooks/roles/testcourses/tasks/deploy.yml
@@ -1,9 +1,9 @@
 ---
-- name: import the test courses from github
-  shell: "{{ demo_edxapp_venv_bin }}/python /edx/bin/manage.edxapp lms git_add_course --settings={{ demo_edxapp_settings }} \"{{ item.github_url }}\""
-  become_user: "{{ common_web_user }}"
-  when: item.install == True
+- include: import_course.yml
+  when: course.install == True
   with_items: "{{ TESTCOURSES_EXPORTS }}"
+  loop_control:
+    loop_var: course
 
 - name: enroll test users in the testcourses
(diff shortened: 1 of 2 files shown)


feanil:
The [demo role cleanup ticket](https://openedx.atlassian.net/browse/OPS-2034) is mostly fore cleaning up users, what sort of additional information should be included in that ticket for cleanup around this role? or am I looking at the wrong ticket?

jibsheet:
This role has an implicit "import the demo users" dep in the task below the checkouts that I tweaked
https://github.com/edx/configuration/blob/386ed68a87f6d45b3c5773d2e7c6faccc9863306/playbooks/roles/testcourses/tasks/deploy.yml#L15
so I assumed that this was going to be dragged into that cleanup. I'll add some notes.

I kind of think we can rewrite testcourses in terms of the demo role, but that was out of scope for "fix nightly CI" in my opinion.

feanil (dismissed):
Once the ticket has been updated.

mattdrayer:
@mattdrayer lurks...

jibsheet:
I'm running a rebuild here
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/79/console
but this now correctly imports and enrolls in my testing

feanil (approved):

mattdrayer:
🚀

PR #3736, opened 2017-03-06
script to shovel between rabbitmq clusters

nadeemshahzad:
@edx/devops Please review

### DIFF
--- a/util/rabbitmq/shovel.py
+++ b/util/rabbitmq/shovel.py
@@ -1,39 +1,52 @@
-#!/usr/bin/python
-
 import argparse
 import subprocess
+import requests
 
 parser=argparse.ArgumentParser(description='Shovels between RabbitMQ Clusters')
 parser.add_argument('--src_host',action='store',dest='src_host')
 parser.add_argument('--dest_host',action='store',dest='dest_host',default='127.0.0.1')
 parser.add_argument('--src_user',action='store',dest='src_user')
+parser.add_argument('--src_user_pass',action='store',dest='src_user_pass')
 parser.add_argument('--dest_user',action='store',dest='dest_user')
-parser.add_argument('--src_pass',action='store',dest='src_pass')
-parser.add_argument('--dest_pass',action='store',dest='dest_pass')
-parser.add_argument('--src_queue',action='store',dest='src_queue')
-parser.add_argument('--dest_queue',action='store',dest='dest_queue')
-parser.add_argument('--name',action='store',dest='name')
+parser.add_argument('--dest_user_pass',action='store',dest='dest_user_pass')
 
 args=parser.parse_args()
 
-SRC_URI="\"amqp://%s:%s@%s\"" % (args.src_user,args.src_pass,args.src_host)
-DEST_URI="\"amqp://%s:%s@%s\"" % (args.dest_user,args.dest_pass,args.dest_host)
-SRC_QUEUE="\"%s\"" % args.src_queue
-DEST_QUEUE="\"%s\"" % args.dest_queue
-SHOVEL_NAME="%s" % args.name
-SHOVEL_ARGS="{\"src-uri\": %s, \"src-queue\": %s,\"dest-uri\": %s,\"dest-queue\": %s}" % (SRC_URI,SRC_QUEUE,DEST_URI,DEST_QUEUE)
+src_uri="\"amqp://%s:%s@%s\"" % (args.src_user,args.src_user_pass,args.src_host)
+dest_uri="\"amqp://%s:%s@%s\"" % (args.dest_user,args.dest_user_pass,args.dest_host)
+port=15672
 
-def run_cmd(cmd):
-    subprocess.call(cmd,shell=True)
+def list_vhosts():
+    url="http://%s:%d/api/vhosts" % (args.src_host,port)
+    response=requests.get(url,auth=(args.src_user,args.src_user_pass))
+    vhosts=[v['name'] for v in response.json() if v['name'].startswith('/')]
+    return vhosts
+
+def list_queues():
+    for vhost in list_vhosts():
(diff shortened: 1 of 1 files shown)


feanil:
@nadeemshahzad this script currently only takes one queue name as input, I think what we want is the ability to point the script at one cluster and it would find all the queues on that cluster that were not `celereyev` queues and run shovel for each of them.

nadeemshahzad:
@feanil I have updated script to iterate through queues of old cluster and create shovel for each of them except **celeryev** and **.pidbox** queues

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+parser.add_argument('--src_host',action='store',dest='src_host')
+parser.add_argument('--dest_host',action='store',dest='dest_host',default='127.0.0.1')
+parser.add_argument('--src_user',action='store',dest='src_user')
+parser.add_argument('--src_user_pass',action='store',dest='src_user_pass')
+parser.add_argument('--dest_user',action='store',dest='dest_user')
+parser.add_argument('--dest_user_pass',action='store',dest='dest_user_pass')
+
+args=parser.parse_args()
+
+src_uri="\"amqp://%s:%s@%s\"" % (args.src_user,args.src_user_pass,args.src_host)
```
feanil: Use single quotes on the outside so you don't have to escape the double quotes.

Also use `str.format()` instead of `%`

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+dest_uri="\"amqp://%s:%s@%s\"" % (args.dest_user,args.dest_user_pass,args.dest_host)
```
feanil: Same note as the line above.

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+port=15672
+
+def list_vhosts():
+    url="http://%s:%d/api/vhosts" % (args.src_host,port)
```
feanil: Use `str.format()`

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+    response=requests.get(url,auth=(args.src_user,args.src_user_pass))
+    vhosts=[v['name'] for v in response.json() if v['name'].startswith('/')]
```
feanil: You should do error checking on the response before trying to pull out the json response.

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+    return vhosts
+
+def list_queues():
+    for vhost in list_vhosts():
+        url="http://%s:%d/api/queues/%s" % (args.src_host,port,vhost)
```
feanil: Use `str.format()`

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+        response=requests.get(url,auth=(args.src_user,args.src_user_pass))
+        queues=[q['name'] for q in response.json()]
```
feanil: Add error checking before reading out of the request response.

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+        return queues
+
+def create_shovel(shovel,arg):
+    cmd="/usr/sbin/rabbitmqctl set_parameter shovel %s '%s'" % (shovel,arg)
```
feanil: Use `str.format()`

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+    subprocess.call(
```
feanil: Use [check_output](https://docs.python.org/3.5/library/subprocess.html#subprocess.check_output) instead. If the call fails, check the exception and store the output.  At the end of the run, it should produce a list of all queues where shovel call failed and the output of the call.

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,52 @@
...
+    --dest_host <dest_host_IP> --dest_user <dest_rabbitmq_user> --dest_user_pass <user_pass>
+    """
+        
+    for queue in list_queues():
+        """ 
+        Ignore queues celeryev and *.pidbox to shovel
+        """
+        q=queue.split('.')
+        if (q[0]!='celeryev' and q[-1]!='pidbox'):
+            args="{\"src-uri\": %s, \"src-queue\": \"%s\",\"dest-uri\": %s,\"dest-queue\": \"%s\"}" % (src_uri,queue,dest_uri,queue)
```
feanil: Use single quotes on the outside so you don't have to escape all the double quotes.

Use `str.format()` instead of `%`

feanil (changes requested):

nadeemshahzad:
@feanil Updated error handling and template formatting

util/rabbitmq/shovel.py
```diff
@@ -0,0 +1,70 @@
...
+        """ 
+        Ignore queues celeryev and *.pidbox to shovel
+        """
+        q=queue.split('.')
+        if (q[0]!='celeryev' and q[-1]!='pidbox'):
+            args='{{"src-uri": "{}", "src-queue": "{}","dest-uri": "{}","dest-queue": "{}"}}'.format(src_uri,queue,dest_uri,queue)
+            shovel_output=create_shovel(queue,args)
+            if shovel_output is not None:
+               output.append(shovel_output)
+    print "\n".join(output)
```
feanil: Can you also include the name of the queue that failed or is that already provided in the output?

feanil (approved):
One last comment to address then I think this is good to go.

nadeemshahzad:
@feanil Added queue name to output if call failed

PR #3743, opened 2017-03-08
Explicitly allow read access to gocd log directory.

feanil:
Configuration Pull Request
---

@edx/pipeline-team the re-install made this directory only readable by the go user even though the files in it are readable by everyone. So splunk was unable to list the directory to figure out the names of the files.

### DIFF
--- a/playbooks/roles/go-server/tasks/main.yml
+++ b/playbooks/roles/go-server/tasks/main.yml
@@ -83,6 +83,17 @@
     group: "{{ GO_SERVER_GROUP }}"
     force: no
 
+- name: ensure everyone can read go-server log files
+  file:
+    path: "/var/log/go-server"
+    state: directory
+    mode: "0755"
+    owner: "{{ GO_SERVER_USER }}"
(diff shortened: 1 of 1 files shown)


benpatterson:
😍

playbooks/roles/go-server/tasks/main.yml
```diff
@@ -83,6 +83,17 @@
...
+- name: ensure everyone can read go-server log files
+  file:
+    path: "/var/log/go-server"
+    state: directory
+    mode: "0755"
+    owner: "{{ GO_SERVER_USER }}"
+    group: "{{ GO_SERVER_GROUP }}"
+  tags:
+    - install
+    - install:base
```
macdiesel, line 95: with just these tags will we be able to run this on an existing go instance?

doctoryes (approved):

playbooks/roles/go-server/tasks/main.yml
feanil: Can, and did.  Logs are already flowing back into splunk.

PR #3744, opened 2017-03-08
Updated country population task for ecommerce

clintonb:
We now make use of the management command's initial-only flag silently exit if the countries table is already populated.

ECOM-7422


clintonb:

jibsheet (approved):

rlucioni (approved):
Pretty cool that we're finally able to use `--initial-only` option we contributed. I'd forgotten about it!

clintonb:
So. Long. Ago.

PR #3748, opened 2017-03-10
Adds EDXAPP_ENTERPRISE_ENROLLMENT_API_URL

pomegranited:
These settings are used by edx-enterprise, an app which runs in edxapp.

* `EDXAPP_ENTERPRISE_ENROLLMENT_API_URL`: is used for server-side requests to the LMS Enrollments API.
* `EDXAPP_ENTERPRISE_PUBLIC_ENROLLMENT_API_URL`: is used for client-side requests.

Common templates modifies [`/etc/hosts` to alias the FQDN of the edxapp server to `127.0.0.1`](https://github.com/edx/configuration/blob/0aa9bfa07853d319c487de1690309d2b3cf38c2b/playbooks/roles/common/templates/hosts.j2#L1).  Servers that sit behind a load balancer and cannot decrypt their own SSL therefore cannot contact their own FQDN via https.

Rather than try and change the `/etc/hosts` entry, and potentially disrupt the basic assumptions of most edxapp servers, we've decided to just use two URLs.

cf: https://github.com/edx/edx-enterprise/pull/67, https://github.com/edx/edx-platform/pull/14671

**Sandbox URL**: sandbox is being provisioned.

* LMS: https://pr14671.sandbox.opencraft.hosting/
* Studio: https://studio-pr14671.sandbox.opencraft.hosting/

**Deployment targets**: edx.org and edge.edx.org

**Reviewers**:

- [ ] @haikuginger 
- [ ] @edx/devops TBD

Configuration Pull Request
---


openedx-webhooks:
Thanks for the pull request, @pomegranited! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=3748"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

pomegranited:
PR not required; will just change defaults in edx-platform.

PR #3750, opened 2017-03-13
Devstack workarounds

robrap:
These commits are here for discussion.

The first commit fixes (or is a work around for) an ecommerce related issue.  I'm not sure if this breaks anything on sandboxes or any other environments.  It does get us through a devstack build.
```
TASK [ecommerce : Migrate] fails with django.db.utils.OperationalError: (1050, "Table 'thumbnail_kvstore' already exists") in the error trace.
```

The second is a work around for a mongo auth issue for the bokchoy test db.  I do not know what the proper fix should be.  Here is the original error:
```
 File "/edx/app/edxapp/venvs/edxapp/local/lib/python2.7/site-packages/pymongo/helpers.py", line 213, in _check_command_response
    raise OperationFailure(msg % errmsg, code, response)
pymongo.errors.OperationFailure: command SON([('createIndexes', u'events'), ('indexes', [{'name': u'time_-1', 'key': SON([('time', -1)])}])]) on namespace test.$cmd failed: not authorized on test to execute command { createIndexes: "events", indexes: [ { name: "time_-1", key: { time: -1 } } ] }
```

FYI: @jibsheet @clintonb @jlajoie @benpatterson

---


playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -68,6 +68,17 @@
     - install
     - install:app-requirements
 
+- name: fake thumbnails
```
clintonb, line 71: This should not be run in any real environments (e.g. stage, production).

clintonb (changes requested):

playbooks/roles/ecommerce/tasks/main.yml
robrap: I probably should have added a comment.  Either way, I'm not sure how this should be managed, so I'm looking for some help from @edx/devops.

clintonb:
@robrap in the past I have seen others/been advised to do the following:

**Sandbox**
1. Make a new sandbox from scratch.
2. Ask DevOps to cut a new base image.

**Devstack**
1. Develop a workaround script, and send it to edx-code.
2. Rebuild/redeploy the devstack image from scratch.

robrap:
Thanks @clintonb.  I think my PR is the best workaround for now.  Not sure if you think we should send it out as a workaround.
I opened a ticket with devops (DEVOPS-5611) to hopefully get to the full resolution of both these issues.

playbooks/roles/mongo_2_6/defaults/main.yml
```diff
@@ -31,7 +31,7 @@ MONGO_USERS:
 
 MONGO_CLUSTERED: false
 MONGO_BIND_IP: 127.0.0.1
-MONGO_AUTH: true
+MONGO_AUTH: false
```
jibsheet, line 34: I would rather not change this in just the 2.6 role.

I believe the correct fixes are either:

* create the mongo DBs for testing in edxlocal or here (edxlocal is better since it's devstack contained) and that should be as easy as a new var and a new task.
* disable MONGO_AUTH only in the devstack playbook, which is vagrant-devstack.yml in the playbooks directory (and make sure to get analyticsstack and friends).  That's as trivial as adding to the existing vars: declaration

I don't have time to write any of these fixes because I'm on PTO after today - but am happy to explain the approaches more if desired.

robrap: Thanks @jibsheet.  The 2.6 role change was just a temp quick fix for myself.

Can you clarify which of your two ideas is simplest and would get the job done?  Also, if you could drop any code or point to anything that would help with implementation, that would be great.

jibsheet: The vagrant playbooks are here
https://github.com/edx/configuration/tree/master/playbooks
and the var definition process should be pretty obvious

The more correct fix of edxlocal would involve a new variable
https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/defaults/main.yml that lists databases and users, you can model it on
https://github.com/edx/configuration/blob/master/playbooks/roles/mongo_2_6/defaults/main.yml#L22-L30

And then you copy/paste a task into https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/tasks/main.yml that steals the task from
https://github.com/edx/configuration/blob/master/playbooks/edx-east/create_mongo_users.yml

robrap:
FYI: @nedbat

robrap:
@clintonb: Can you provide @nedbat and I a more detailed description of what broke this thumbnail and what is fixing it, so we can understand how and if a redeploying the devstack image from scratch will resolve this.  Note that that process will be done using ficus branches, so if the problem/fix is not on those branches, this may not resolve.  Thank you.

clintonb:
@robrap @nedbat https://django-oscar.readthedocs.io/en/releases-1.4/releases/v1.3.html#dependency-changes. The migration needs to be faked if running against an existing database. If starting from scratch, the migration will be run as normal, so there is no need to fake it.

playbooks/roles/mongo_2_6/defaults/main.yml
robrap: @jibsheet Do you mean to literally copy the full contents of `create_mongo_users.yml` into `edxlocal/tasks/main.yml`?  Or a subset (lines?), or other?  Thanks.

robrap: @fredsmith: Feel free to answer in @jibsheet's place while he is on vacation.  Thanks.
Also, what is the quickest way to test?  Is there a way to just kick this off?

robrap: @fredsmith: @jibsheet answered the coding question.  Here is the code:
https://github.com/edx/configuration/blob/master/playbooks/edx-east/create_mongo_users.yml#L33-L44

I'll just run provision to test.

robrap: @fredsmith Can you help me understand this?  It's clear, except for the part about "analyticsstack and friends"?

Earlier comment from @jibsheet:
> disable MONGO_AUTH only in the devstack playbook, which is vagrant-devstack.yml in the playbooks directory (and make sure to get analyticsstack and friends). That's as trivial as adding to the existing vars: declaration

robrap: There were too many test dbs, including unit tests, and I think they had random names, so I'm voting to turn off MONGO_AUTH for devstacks.  I'm still not sure about the "analyticsstack and friends" comment above.

robrap:
Closing this PR in favor of [this new PR](https://github.com/edx/configuration/pull/3756) that attempts more permanent fixes for these issues.  I'm going to leave this branch until we merge the other PR, because others are using it as a workaround.

PR #3751, opened 2017-03-13
Add ability to delete EBS volumes upon EC2 instance termination.

doctoryes:
The current default is False, according to the docs:
https://docs.ansible.com/ansible/ec2_module.html

This change allows instance-launching code to override the setting and delete EBS volumes upon EC2 instance termination.

---

### DIFF
--- a/playbooks/continuous_delivery/launch_instance.yml
+++ b/playbooks/continuous_delivery/launch_instance.yml
@@ -11,6 +11,9 @@
 #   - ec2_region              - The region the server should be brought up in
 #   - ec2_instance_type       - The instance type to use
 #   - ebs_volume_size         - Size in GB for the root volume
+#   - delete_ebs_volumes_on_termination
+#                             - True to delete any EBS volumes attached to an EC2 instance upon its termination;
+#                               False to leave the EBS volumes un-deleted (default).
 #   - ec2_timeout             - Time in seconds to wait for an ec2 instance become available
 #   - ec2_assign_public_ip    - (yes/no) should the instance have a public IP address?
 #   - ami_id                  - overrides 'base_ami_id' (used when passing in the ami.yml file produced by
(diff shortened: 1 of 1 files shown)


macdiesel (approved):
The ansible here looks fine, is there a companion gomatic PR to enable this on stage? 

Also looking at the larger picture here @fredsmith @jibsheet do we have a janitor that cleans up these EBS volumes in production?

jibsheet:
We do have a janitor
http://admin.edx.org:8080/job/janitor-edx/
https://github.com/edx-ops/sysadmin/tree/master/util/janitor

It's designed to be pluggable
You could use it

doctoryes:
@macdiesel I have made the companion PR now:
https://github.com/edx/edx-gomatic/pull/271

fredsmith:
What problem does this solve that Minos and the janitor process doesn't?  Minos checks volumes for important data that needs to be backed up before allowing instances to terminate, and the janitor process deletes unused volumes that have been confirmed safe to delete.  I think improving this method is preferable to allowing AWS to delete volumes on instance shutdown.

doctoryes:
@fredsmith This ticket and work was initiated by a situation we had in the last couple of months where we ran over some existing EBS quota. I never knew the details - @macdiesel @feanil do you know why the volumes weren't being properly deleted by Minos and the janitor process?

macdiesel:
Janitors currently only clean up AGS and danging instances. I would need @feanil to provide further context of why we want this in stage but not in prod, but I seem to remember something about a data dump after termination, not sure.

jibsheet:
Doing it in stage but not prod sounds a lot like relying on minos (which ensures data is in S3 before allowing a machine to terminate) but I'll let @feanil talk it out with you folks, since I was away for the outage that caused this.

feanil:
Unfortunately we can't do this.  When we build AMIs built from an instance that has ebs volume deletion turned on will inherit that setting.  This means any new instances launched from that AMI would also have volume deletion turned on by default which we don't want for prod volumes at least.  There is a jenkins job that runs every 4 hours that will cleanup volumes for which we have snapshots.  But looking at the job that makes snapshots, it only runs daily.  I have updated that job to run every 4 hours as well so volumes should get cleaned up more quickly than before.  We can further tweak these jobs the code for them lives in the sysdamin repo.

doctoryes:
Closing per @feanil 's note above.

PR #3753, opened 2017-03-13
Remove compile translations step in insights role

thallada:
We keep the [compiled translation files in git](https://github.com/edx/edx-analytics-dashboard/tree/master/analytics_dashboard/conf/locale) already and have [an automated job](https://github.com/edx/edx-analytics-dashboard/pulls?q=is%3Apr+author%3Aedx-transifex-bot) that compiles them regularly, so there's no need to keep this step.

I thought this might fix stage-edx-insights from failing when trying to build AMIs, but it looks like it still failed when I tried to build an AMI with Alton using this branch. Regardless, I think this still makes sense to merge.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/insights/tasks/main.yml
+++ b/playbooks/roles/insights/tasks/main.yml
@@ -106,15 +106,6 @@
     - assets
     - assets:gather
 
-- name: compile translations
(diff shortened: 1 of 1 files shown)


dsjen (approved):
👍

jibsheet (approved):

PR #3754, opened 2017-03-14
Use safer clean that protects private files.

robrap:
Configuration Pull Request
---
This should merge after @nedbat's PR to fix the edx-platform `make clean`: https://github.com/edx/edx-platform/pull/14677.

This command will protect private files, rather than blowing them away.  This is convenient for devstacks.

@fredsmith Can you review? 


robrap:
@fredsmith: This doesn't work.  I think it would need conditionals and call "make clean" only if the Makefile exists, otherwise call git clean directly.  I don't have time for this now, so I'm just going to close this.
FYI: @nedbat

PR #3755, opened 2017-03-14
remove ES cluster members var override for analyticstack

mulby:
@edx/devops

This was causing analyticstack builds to fail by writing this incorrect setting to the elasticsearch.yml file (clearly problematic).

```yaml
discovery.zen.ping.unicast.hosts: ['[',']']
```

FYI - @azafty468 @brianhw @HassanJaveed84 @iloveagent57


tobz (approved):

jibsheet (approved):
Does it default to an empty list fine now?  Or is there more to change here.  Otherwise, it's clearly generating the wrong thing.

mulby:
No more to change here. It works properly with this change.

PR #3756, opened 2017-03-14
Fix devstack issues DEVSTACK-28 and DEVSTACK-29

robrap:
Configuration Pull Request
---

Fixes issues DEVSTACK-28 and DEVSTACK-29:
- Adding ecommerce fake thumnail migration when needed.
- Turning off MONGO_AUTH for devstacks.

### DIFF
--- a/playbooks/roles/ecommerce/tasks/main.yml
+++ b/playbooks/roles/ecommerce/tasks/main.yml
@@ -68,6 +68,24 @@
     - install
     - install:app-requirements
 
+# This is a hacked fix for the fact that the table `thumbnail_kvstore` exists in
+# some environments, which won't need the 3rd party newly introduced migration
+# to create this table, so we fake the migration.
+# This is required for the Ginkgo release.
+# TODO: Delete this task for the Hawthorn release.
+- name: fake thumbnails
+  shell: >
+    table_exists=`mysql -uroot -ss -e "SELECT EXISTS(SELECT * FROM information_schema.tables WHERE table_schema = '{{ ECOMMERCE_DEFAULT_DB_NAME }}' AND table_name = 'thumbnail_kvstore')"`;
+    if [ "$table_exists" -eq "1" ]; then {{ ecommerce_venv_dir }}/bin/python ./manage.py migrate thumbnail 0001 --fake; fi;
+  args:
+    chdir: "{{ ecommerce_code_dir }}"
+  become_user: "{{ ecommerce_user }}"
+  environment: "{{ ecommerce_environment }}"
+  when: migrate_db is defined and migrate_db|lower == "yes"
+  tags:
+    - migrate
+    - migrate:db
+
 - name: Migrate
   shell: >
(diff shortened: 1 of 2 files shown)


robrap:
This has been tested on a fresh devstack.  I imagine I should test this on a sandbox and from-scratch sandbox @clintonb?

clintonb (approved):
Yes, please test with an existing installation. A new sandbox should suffice.

robrap:
Will do @clintonb...once sandbox provisioning is working again in general.

robrap:
Building sandbox now: http://jenkins.edx.org:8080/job/ansible-provision/11374/
Will be: https://robrap.sandbox.edx.org

robrap:
FYI: @fredsmith @clintonb: The sandbox I had created did not use "reconfigure", and it passed.  @mattdrayer was using "reconfigure" on a new sandbox and ran into the thumbnail issue.  So, I gave him a new branch with this fix (he was also doing an ES upgrade).  If that passes, it means this was tested on a sandbox with and without reconfigure selected, as well as a fresh devstack.

mattdrayer:
Unfortunately the sandbox build failed, stating "Access denied for user 'migrate'@'localhost' to database 'discovery'\"

See http://jenkins.edx.org:8080/view/Ansible/job/ansible-provision/11376/console for details

robrap:
@mattdrayer I know what that issue is as well.  I have a workaround that I added to the configuration branch.  I think the real fix may need a DEVOPS ticket which I can help explain.
Can you first check with DEVOPS or on a sandbox that it is still using mongo 2.6?  If it upgraded to mongo 3.0, you'll need to switch to "MONGO_AUTH: false" in the mongo 3.0 file.  You are welcome to do that preemptively on the branch you are testing.  Just see the commit I just added and the change will be obvious.

robrap:
FYI: @clintonb @fredsmith: I'll need to create a separate ticket for sandboxes and the mongo auth issue (see @mattdrayer's comment), but I don't plan on fixing that one.  

I'm not sure of the fix.  It seems that the way the configuration happens now, MONGO_AUTH is on for everything that goes on the single sandbox, but is probably off on the IDAs, so IDAs like discovery never test for this and it first fails when hitting a sandbox.  That just a guess.  Not sure how to correct this so that this change gets pushed and fixed to all the IDAs by default.  I'm planning on opening a DEVOPS ticket for this, since that is where "MONGO_AUTH: true" originated.  Thanks.

mattdrayer:
@robrap I SSH'ed to the failed sandbox and this is what I see:

```
mattdrayer@mtd4 i-04fb17ed0f49f7a79:~$ mongod --version
db version v2.6.12
2017-03-15T15:59:20.999+0000 git version: d73c92b1c85703828b55c2916a5dd4ad46535f6a
```

fredsmith (approved):

robrap:
FYI: It turns out @mattdrayer's issue was a MySQL issue, and not a mongo issue.  I added DEVOPS-5634 for this.

PR #3760, opened 2017-03-16
Allow custom version tags to be applied when creating an AMI

cpennington:
Configuration Pull Request
---

### DIFF
--- a/playbooks/continuous_delivery/create_ami.yml
+++ b/playbooks/continuous_delivery/create_ami.yml
@@ -27,6 +27,8 @@
 #   - hipchat_room                    - ID or name of the room to send the notification
 #   - hipchat_url                     - URL of the hipchat API  (defaults to v1 of the api)
 #   - extra_name_identifier           - Makes each AMI unique if desired - Default: 0
+#   - version_tags                    - A mapping of {app: [repo, version], ...}, used to generate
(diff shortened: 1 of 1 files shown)


playbooks/continuous_delivery/create_ami.yml
```diff
@@ -89,6 +89,16 @@
       resource: "{{ ami_register.image_id }}"
       tags: "{{ instance_tags.tags }}"
 
+  - name: Add any version tags that were passed on the commandline
+    ec2_tag:
+      region: "{{ ec2_region }}"
+      resource: "{{ ami_register.image_id }}"
+      tags: "{
+          {% for name, (repo, version) in version_tags.items() %}
```
macdiesel, line 100: Could you please add version_tags to the docs at the top of this file.

macdiesel (approved):
Looks fine, please add documentation for the new variable.

playbooks/continuous_delivery/create_ami.yml
cpennington: Will do. I'll also add a default value that's empty, so that this doesn't break if it's not supplied.

cpennington: Done

PR #3761, opened 2017-03-17
Add a playbook to install Grafana.

tobz:
This is a playbook to install Grafana.

We pull from their PackageCloud PPA, so most things are taken care of for us, and all we do is provide a templatized configuration file and management tasks.

There's also an nginx site to go along with it.

---

### DIFF
--- a/playbooks/roles/grafana/defaults/main.yml
+++ b/playbooks/roles/grafana/defaults/main.yml
@@ -1,0 +1,501 @@
+---
+
+grafana_apt_pkg: grafana
+
+GRAFANA_NGINX_PORT: 80
+
+# Application mode.
+grafana_app_mode: "production"
+
+# Data paths.
+grafana_paths_data: "/var/lib/grafana"
+grafana_paths_logs: "/var/log/grafana"
+grafana_paths_plugins: "{{ grafana_paths_data }}/plugins"
+
+grafana_paths:
+  data: "{{ grafana_paths_data }}"
+  logs: "{{ grafana_paths_logs }}"
+  plugins: "{{ grafana_paths_plugins }}"
+
+# Server basics.
+grafana_server_protocol: "http"
+grafana_server_http_addr: "127.0.0.1"
+grafana_server_http_port: 3000
+grafana_server_domain: "localhost"
+grafana_server_enforce_domain: false
+grafana_server_root_url: "%(protocol)s://%(domain)s:%(http_port)s/"
+grafana_server_router_logging: false
+grafana_server_static_root_path: "public"
+grafana_server_enable_gzip: false
+grafana_server_cert_file: ""
+grafana_server_cert_key: ""
+
+grafana_server:
+  protocol: "{{ grafana_server_protocol }}"
+  http_addr: "{{ grafana_server_http_addr }}"
+  http_port: "{{ grafana_server_http_port }}"
+  domain: "{{ grafana_server_domain }}"
+  enforce_domain: "{{ grafana_server_enforce_domain }}"
+  root_url: "{{ grafana_server_root_url }}"
+  router_logging: "{{ grafana_server_router_logging }}"
+  static_root_path: "{{ grafana_server_static_root_path }}"
+  enable_gzip: "{{ grafana_server_enable_gzip }}"
+  cert_file: "{{ grafana_server_cert_file }}"
+  cert_key: "{{ grafana_server_cert_key }}"
+
+# Database.
+grafana_database_type: "sqlite3"
+grafana_database_host: 127.0.0.1:3306
+grafana_database_name: "grafana"
+grafana_database_user: "root"
+grafana_database_password: ""
+grafana_database_ssl_mode: "disable"
+grafana_database_ca_cert_path: ""
+grafana_database_client_key_path: ""
+grafana_database_client_cert_path: ""
+grafana_database_server_cert_name: ""
+grafana_database_path: "grafana.db"
+
+grafana_database:
+  type: "{{ grafana_database_type }}"
+  host: "{{ grafana_database_host }}"
+  name: "{{ grafana_database_name }}"
+  user: "{{ grafana_database_user }}"
+  password: "{{ grafana_database_password }}"
+  ssl_mode: "{{ grafana_database_ssl_mode }}"
+  ca_cert_path: "{{ grafana_database_ca_cert_path }}"
+  client_key_path: "{{ grafana_database_client_key_path }}"
+  client_cert_path: "{{ grafana_database_client_cert_path }}"
+  server_cert_name: "{{ grafana_database_server_cert_name }}"
+  path: "{{ grafana_database_path }}"
+
+# Sessions.
+grafana_session_provider: "file"
+grafana_session_provider_config: "sessions"
+grafana_session_cookie_name: "grafana_sess"
+grafana_session_cookie_secure: false
+grafana_session_session_life_time: 86400
+grafana_session_gc_interval_time: 86400
+
+grafana_session:
+  provider: "{{ grafana_session_provider }}"
+  provider_config: "{{ grafana_session_provider_config }}"
+  cookie_name: "{{ grafana_session_cookie_name }}"
+  cookie_secure: "{{ grafana_session_cookie_secure }}"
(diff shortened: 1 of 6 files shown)


playbooks/roles/grafana/tasks/main.yml
```diff
@@ -0,0 +1,81 @@
...
+    - install:system-requirements
+
+- name: install PackageCloud PPA
+  apt_repository:
+    # This is  Debian Jessie repository, when we use Ubuntu Xenial, yes.  It's the latest Debian repository they
+    # populate for their official packages.  It does work, so, *shrug*.
+    repo: "deb https://packagecloud.io/grafana/stable/debian/ jessie main"
+    update_cache: yes
+    state: present
+  tags:
```
e0d, line 44: should we limit to particular OS in a when?

e0d: Typically, we've been making these vars, a little fiddly maybe, but slightly better from a transparency POV.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/grafana.j2
```diff
@@ -0,0 +1,51 @@
+#
+# {{ ansible_managed }}
+#
+{# This prevents the injected comment from eating the server
+   directive.  There's probably a better way of doing this,
+   but I don't know it currently.
+#}
+{% raw %}
```
e0d: Not a deal breaker, but http://jinja.pocoo.org/docs/2.9/templates/#whitespace-control

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/graphite.j2
```diff
@@ -20,7 +20,6 @@ upstream graphite_app_server {
 }
 
 server {
-
   server_name graphite.*;
   listen {{ GRAPHITE_NGINX_PORT }} {{ default_site }};
   client_max_body_size 1M;
```
e0d, line 25: would we be likely to change this over time?  If so, make it a var.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/grafana.j2
```diff
@@ -0,0 +1,51 @@
...
+{%- endif -%}
+
+upstream grafana_app_server {
+  server 127.0.0.1:3000 fail_timeout=0;
+}
+
+server {
+  server_name grafana.*;
+  listen {{ GRAFANA_NGINX_PORT }} {{ default_site }};
+  client_max_body_size 1M;
```
e0d, line 18: Seems this should be synched with the value below, so a var makes sense.

playbooks/roles/grafana/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+grafana_analytics_google_tag_manager_id: ""
+
+grafana_analytics:
+  reporting_enabled: "{{ grafana_analytics_reporting_enabled }}"
+  check_for_update: "{{ grafana_analytics_check_for_update }}"
+  google_analytics_ua_id: "{{ grafana_analytics_google_analytics_ua_id }}"
+  google_tag_manager_id: "{{ grafana_analytics_google_tag_manager_id }}"
+
+# Security.
+grafana_security_admin_user: "admin"
```
e0d, line 101: typically we set credentials to values that make it obvious they should be changed.

playbooks/roles/grafana/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+grafana_auth__anonymous_org_role: "Viewer"
+
+grafana_auth__anonymous:
+  enabled: "{{ grafana_auth__anonymous_enabled }}"
+  org_name: "{{ grafana_auth__anonymous_org_name }}"
+  org_role: "{{ grafana_auth__anonymous_org_role }}"
+
+# Authorization (GitHub).
+grafana_auth__github_enabled: "false"
+grafana_auth__github_allow_sign_up: "truee"
```
e0d: typo, also wonder why all these boolean values are strings.

playbooks/roles/grafana/defaults/main.yml
```diff
@@ -0,0 +1,501 @@
...
+grafana_paths_plugins: "{{ grafana_paths_data }}/plugins"
+
+grafana_paths:
+  data: "{{ grafana_paths_data }}"
+  logs: "{{ grafana_paths_logs }}"
+  plugins: "{{ grafana_paths_plugins }}"
+
+# Server basics.
+grafana_server_protocol: "http"
+grafana_server_http_addr: "0.0.0.0"
```
e0d: Is that what we want for the default?

e0d (commented):
Generally looks good, few questions and comments.

playbooks/roles/grafana/tasks/main.yml
tobz: Probably?  I was working off the assumption that we are not likely to install this stuff on CentOS, RHEl, etc, any time soon, and that there'd be massive work to make it possible.. so it could be deferred.

Maybe a better question is whether or not we should gate the playbook working at all by doing something more upstream?  For example, configuring the service makes no sense without installing it, but a `when` directive would simply stop the package install, if I understand the docs correctly, and let other things proceed (and then fail).

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/graphite.j2
tobz: The maximum body size? Uh, not likely to change, AFAIK.  That would be inbound requests, and I can't see one being over 1MB to begin with, really.  It feels totally safe to cap and leave at 1MB.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/grafana.j2
tobz: Sorry, what is "this" in this case, and what should it be synced with?

playbooks/roles/grafana/defaults/main.yml
tobz: Good point.  I'll make it localhost.

playbooks/roles/grafana/defaults/main.yml
tobz: Fair.  I'll change the admin password to "changeme".

playbooks/roles/grafana/defaults/main.yml
tobz: I did a copy/paste of the key/values from another cookbook as a starting point, and so I missed the fact these were all booleans-as-string.  I'll convert them to proper booleans.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/graphite.j2
e0d: If that's the case, fine, but the "this" above was referring to the fact that you are setting the same value in two places with a static value.  Not a big deal, but might save future someone a deploy if changing it once, updated both settings.

playbooks/roles/grafana/tasks/main.yml
e0d: There's a fail module that could gate the role:
http://docs.ansible.com/ansible/fail_module.html

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/graphite.j2
tobz: Oh, I see what you're saying.  I mean, I could make these use variables, but I wouldn't use a shared variable.  One thing is Grafana and one thing is Graphite.  What do you use as the prefix for that variable?  Why is a variable in one playbook affecting things in another?

(We obviously define things like HTTP ports, etc, to be used by the nginx playbook but the templates in nginx serve at the behest of other playbooks, so they are inextricably tied.)

e0d: I had missed they are related to the different applications.  Let's just leave this then.

tobz:
@e0d I think all of your feedback has been addressed.

e0d (approved):

PR #3762, opened 2017-03-17
Remove the specific version AMI version tagging in favor of using the version_tags variable

cpennington:
This shouldn't merge until https://github.com/edx/edx-gomatic/pull/280 merges.

Configuration Pull Request
---


doctoryes (approved):

feanil:
Was the alton diff functionality updated to match the new tagging scheme?

cpennington:
The tags generated are identical.

cpennington:
At least, they are for all existing tags. I didn't update alton to be generic (if it wasn't already).

PR #3769, opened 2017-03-21
Use gevent workers for discovery

rlucioni:
Some API endpoints on the discovery service make long, blocking calls to external services (i.e., Elasticsearch). Combined with synchronous Gunicorn workers, this can cause requests to pile up and the service to respond with 503s. By using asynchronous workers, the service should be less likely to 503 while it responds to a long-running request involving Elasticsearch queries. Discovery already installs gevent in production.

@edx/ecommerce @edx/devops


feanil:
Was this load tested?

rlucioni:
@feanil no. Do we stand to gain significant value from investing the time in load testing this? It's not the first time we're using gevent workers.

Our loadtest-edx-discovery pipeline isn't set up to deploy a branch of configuration. If you're not comfortable shipping this without load testing, I'll need help deploying an AMI built with this branch. I could look into modifying the pipelines so they accept a branch as an environment variable, but I'm not sure this one instance warrants that.

davec-edx:
@rlucioni 

It might be worth throwing a bunch of requests at stage and at least
measuring the CPU and memory load, maybe recovery time. Then deploying this
to stage and retest. Not a full load test for sure but will provide some
insight into issues.

rlucioni:
@feanil and I caught up about this offline. Some libraries do strange things when run with gevent, so I'll do a quick load test before and after applying this change as a sanity check.

rlucioni:
@feanil I'm glad I load tested this! Switching to gevent workers alone destroys the service's response time and throughput. Here's an overview:



I'm going to test a few other theories, but it's safe to say this won't merge in its current state.

mikedikan:
Wow

rlucioni:
I'm going to open a PR against edx-internal with a different but related change. Turns out that tuning the Gunicorn worker count to match the instance type we're running on pays significant dividends.

PR #3772, opened 2017-03-22
Added an Elasticsearch Dockerfile for devstack

clintonb:
This Dockerfile is based on the official image (rather than built from scratch by us) and includes the elasticsearch-head plugin.


clintonb:
@edx/devops please review

PR #3775, opened 2017-03-22
ubuntu-restricted-extras requires multiverse

jibsheet:
This is not enabled by default, enable it for machines using the
browsers role.

Without this I cannot build docker containers for tools-jenkins.
Going to test the build-packer job

@edx/devops please review
FYI @benpatterson


feanil (approved):

jibsheet:
https://build.testeng.edx.org/view/packer/job/build-packer-ami/5269/console

jibsheet:
FYI @estute in case in causes any worker issues (although my build was clean)

PR #3776, opened 2017-03-22
Updated the Credentials Service to use Python 3.5

clintonb:
ECOM-7552

### DIFF
--- a/playbooks/roles/credentials/tasks/main.yml
+++ b/playbooks/roles/credentials/tasks/main.yml
@@ -30,17 +30,42 @@
     - install
     - install:configuration
 
-- name: install application requirements
-  pip:
-    requirements: "{{ credentials_requirements_base }}/{{ item }}"
-    # Ansible will initialize this virtualenv if it's missing.
-    virtualenv: "{{ credentials_venv_dir }}"
-    state: present
-  become_user: "{{ credentials_user }}"
-  with_items: "{{ credentials_requirements }}"
+- name: add deadsnakes repository
+  apt_repository:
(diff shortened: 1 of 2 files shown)


clintonb:
@edx/devops @edx/ecommerce please review.

feanil:
Looks like the docker build failed @clintonb

playbooks/roles/credentials/tasks/main.yml
```diff
@@ -30,17 +30,42 @@
...
+    - install
+    - install:system-requirements
+
+- name: install nodenv
   pip:
-    requirements: "{{ credentials_requirements_base }}/{{ item }}"
-    # Ansible will initialize this virtualenv if it's missing.
-    virtualenv: "{{ credentials_venv_dir }}"
-    state: present
+    name: "nodeenv"
```
rlucioni, line 61: Why do we need to mess with nodeenv as part of this?

clintonb:
Updating the code to retain backward-compatibility with Python 2.7.

playbooks/roles/credentials/tasks/main.yml
clintonb: We use a separate nodeenv for sandboxes and Vagrant-based devstacks. I'm not sure how we managed to build AMIs without installing this package via `pip`. It's no longer in `production.txt`.

clintonb:
Merge https://github.com/edx/credentials/pull/201 first

clintonb:
@feanil @rlucioni thumbs?

rlucioni (approved):

PR #3778, opened 2017-03-23
Add Enterprise Service worker username and API URL

mattdrayer:
Configuration Pull Request
---

@fredsmith here is the configuration update for communicating with the Enterprise service.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -650,6 +650,11 @@ EDXAPP_BLOCK_STRUCTURES_SETTINGS:
   # Maximum number of retries per task.
   TASK_MAX_RETRIES: 5
 
+# Configuration settings needed for the LMS to communicate with the Enterprise service.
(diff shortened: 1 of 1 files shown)


fredsmith (approved):

mattdrayer:
@fredsmith before I merge this PR, just want to confirm that we have the relevant setting(s) defined in our internal configuration spec (assuming it is required)

fredsmith:
@mattdrayer I don't think we need to override these settings, do we?

fredsmith (changes requested):
We should use the edxapp API base URL variable instead of localhost

fredsmith:
https://github.com/edx/edge-internal/pull/10 and https://github.com/edx/edx-internal/pull/34 @jibsheet FYI

fredsmith (approved):

jibsheet (approved):

mattdrayer:
Awesome -- yeah when I looked again at the `localhost:8000` URL I had the same thought and that is why I asked.  The change LGTM -- I'm assuming we'll want to squash the commits into one before merging, yes?  Happy to do that after the checks are complete.

PR #3783, opened 2017-03-29
infrastructure for building an AMI for a loadtest driver jenkins worker

pwnage101:
* Add a new ansible role "loadtest_driver" for setting up any instance
  for running load tests.
* Make the existing "locust" role depend on loadtest_driver and pull out
  the redundant tasks.  The "locust" role is outside of the AMI-building
  code path, but we do use it for something else and want to keep common
  tasks shared.
* Make the "jenkins_worker" role depend on the new "loadtest_driver"
  role.
* Create a new ansible playbook for setting up loadtest drivers.
* create a new packer script to build an AMI for loadtest drivers.

Configuration Pull Request
---

### DIFF
--- a/playbooks/edx-east/jenkins_worker_loadtest_driver.yml
+++ b/playbooks/edx-east/jenkins_worker_loadtest_driver.yml
@@ -1,0 +1,12 @@
+# Configure a Jenkins worker instance
+# This has all the requirements to run load tests.
+
+- name: Configure instance(s)
+  hosts: jenkins_worker
+  become: True
+  gather_facts: True
+  vars:
+    loadtest_driver_worker: True
+  roles:
+    - aws
+    - jenkins_worker
--- a/playbooks/roles/jenkins_worker/meta/main.yml
+++ b/playbooks/roles/jenkins_worker/meta/main.yml
@@ -45,3 +45,7 @@ dependencies:
     android_apt_libraries:
       - lib32stdc++6
       - lib32z1
+
+  # dependencies for loadtest driver worker
+  - role: loadtest_driver
+    when: loadtest_driver_worker is defined
--- a/playbooks/roles/loadtest_driver/defaults/main.yml
+++ b/playbooks/roles/loadtest_driver/defaults/main.yml
@@ -1,0 +1,14 @@
+---
+
+# ulimit variables
+ulimit_config:
+  - domain: '*'
+    type: soft
+    item: nofile
+    value: 4096
+  - domain: '*'
+    type: hard
+    item: nofile
+    value: 4096
+
+ulimit_conf_file: "/etc/security/limits.conf"
--- a/playbooks/roles/loadtest_driver/meta/main.yml
+++ b/playbooks/roles/loadtest_driver/meta/main.yml
@@ -1,0 +1,15 @@
+---
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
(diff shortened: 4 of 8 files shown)


estute (approved):
With regards to the Jenkins/Packer parts. I can't speak for the locust files

pwnage101:
I did just confirm that the one task that I moved out of the locust role and into the loadtest_driver role ("Increase file descriptor limit of the system") is still indirectly triggered by the locust role, as expected.  That means this change does *not* introduce a regression on the existing loadtest workflow.

downstream job which just shows what I mean: http://jenkins.edx.org:8080/job/run-ansible-playbook/1499/console

playbooks/edx-east/jenkins_worker_loadtest_driver.yml
```diff
@@ -0,0 +1,14 @@
...
+# This has all the requirements to run load tests.
+
+- name: Configure instance(s)
+  hosts: jenkins_worker
+  become: True
+  gather_facts: True
+  vars:
+    serial_count: 1
+    loadtest_driver: True
+  serial: "{{ serial_count }}"
```
jibsheet: I know this is just copy/pasta but please remove serial_count and serial.  There's no benefit to your use case.

playbooks/roles/jenkins_worker/meta/main.yml
```diff
@@ -45,3 +45,7 @@ dependencies:
     android_apt_libraries:
       - lib32stdc++6
       - lib32z1
+
+  # dependencies for loadtest driver worker
+  - role: loadtest_driver
+    when: loadtest_driver_worker is defined
```
jibsheet, line 51: Looks like the var up above in the playbook is loadtest_driver ? You should probably keep the worker suffix for consistency with the other plays.

jibsheet (commented):
Looks like one variable mismatch and a small cleanup and you should be good to go.

playbooks/roles/jenkins_worker/meta/main.yml
pwnage101: Huh, I guess I didn't fully verify that the AMI was built correctly.  Sure enough, the loadtest_driver role was not triggered: https://build.testeng.edx.org/view/packer/job/build-packer-ami/5345/consoleFull

playbooks/edx-east/jenkins_worker_loadtest_driver.yml
pwnage101: What does this variable do anyway?  On first grep, I couldn't actually figure out where it gets used.

pwnage101: ah, looks like an ansible feature http://docs.ansible.com/ansible/playbooks_delegation.html

pwnage101:
kicking off new build to test fix: https://build.testeng.edx.org/view/packer/job/build-packer-ami/5372/

pwnage101:
@jibsheet build succeeded and loadtest_driver role executed, thumbs up?

playbooks/edx-east/jenkins_worker_loadtest_driver.yml
jibsheet: Yep, if running on multiple machines, do them $serial at a time (makes sense for something like Mongo where the first one is primary, and the others run later and join the cluster).  But you're running against a single machine, so it does no harm, but is just noise.

jibsheet (approved):

PR #3784, opened 2017-03-30
matrix-auth was being overridden by Jenkins (down to 1.1)

jibsheet:
This appears to be because it wasn't pinned and the shipped version
of the plugin won.  We have pinning code in jenkins_master that we
can use, but we have to copy down that list to add ours (ansible
doesn't allow "recursive references" of a variable passed back to
itself.

Also upgrade DSL plugin to get better error handling.
https://github.com/jenkinsci/job-dsl-plugin/wiki#release-notes
https://github.com/jenkinsci/job-dsl-plugin/wiki/Migration#migrating-to-152

Continue process of trying to group plugins and their deps.

### DIFF
--- a/playbooks/roles/tools_jenkins/defaults/main.yml
+++ b/playbooks/roles/tools_jenkins/defaults/main.yml
@@ -5,7 +5,7 @@ jenkins_tools_version: "1.651.3"
 jenkins_tools_plugins:
   - { name: "cloudbees-folder", version: "5.12" }
   - { name: "structs", version: "1.2" }
-  - { name: "job-dsl", version: "1.50" }
+  - { name: "job-dsl", version: "1.52" }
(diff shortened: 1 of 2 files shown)


jibsheet:
I've rebuilt the docker container and am testing now.
@edx/devops 
FYI @rlucioni because I think you have the most "different" DSL (the translations code).  I'll run that locally and make sure there are no warnings.

rlucioni:
Thanks for the heads up @jibsheet, let me know if I can help.

feanil (approved):

jibsheet:
@rlucioni BTW, translations seeded fine once I found and replaced one lingering reference to edx-jenkins-job-dsl (the old name of the -internal repo).  It was in the seed job configuration, not in your code.

PR #3785, opened 2017-03-30
Removed unused settings for the Credentials Service

clintonb:
LEARNER-406


clintonb:

I'll file a ticket to remove the internal/private/secure values sometime tomorrow.

rlucioni (approved):

PR #3787, opened 2017-03-31
Disable app preload in gunicorn for LMS & Studio.

ormsbee:
Since the earliest days of edX, we've run gunicorn with preload_app as
True. This was primarily because the first handful of courses were using
the XMLModuleStore, which was extremely slow to initialize because it
had to read entire courses from disk and load them into memory at
startup.

XMLModuleStore courses stopped being created with the introduction of
Studio, and we've long since removed them from edx.org. Note: This
should not be confused with importing courses that are authored in XML.
XMLModuleStore courses are courses who use the disk instead of Mongo
for content persistence, and require a server restart for content edits
to show.

This commit sets things back to how Django is more commonly deployed
with gunicorn. It will make server startup time somewhat more expensive
because each worker process does init work, but it should make it
easier to safely enable features like database connection pooling
without having to worry about shared resources.

[PERF-305]

Configuration Pull Request
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,5 @@
+- Role: edxapp
+  - Set preload_app to False in gunicorn config for LMS and Studio.
 - Role: analytics_api
   - Added `ANALYTICS_API_AGGREGATE_PAGE_SIZE`, default value 10.  Adjust this parameter to increase the number of
     aggregate search results returned by the Analytics API, i.e. in course_metadata: enrollment_modes, cohorts, and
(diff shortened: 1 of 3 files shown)


feanil:
:clap:  @ormsbee Please also add an entry to the top of the Changelog for this.

ormsbee:
@feanil: Changelog entry added. I spun up a sandbox at https://ormsbee.sandbox.edx.org / https://studio-ormsbee.sandbox.edx.org, and everything seems to be working. Good to merge?

feanil (approved):
:shipit:

PR #3789, opened 2017-04-03
tentative stab at adding newrelic monitoring to coursegraph (SUST-81)

adampalay:
Configuration Pull Request

@fredsmith , may you please provide some comments and guidance? Here are my questions:

1. When do we use the `lineinfile` module instead of the `template` module? I used the `template` module for the newrelic.yml configuration file. Should I have used `lineinfile` instead?
2. Are the settings sensible for newrelic.yml? They're just the default ones
3. I didn't include newrelic.jar in this PR, because I wasn't sure if it was somehow specific to our newrelic installation.
4. How can I test these configuration changes?

---

### DIFF
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -87,7 +87,8 @@
     src: newrelic.jar
     dest: /etc/neo4j
     # just guessing? No idea what mode we'd need
-    mode: "0644"
+    mode: 0644
+    owner: neo4j
   tags:
     - install
     - install:configuration
@@ -95,7 +96,7 @@
 - template:
     src: ../templates/etc/neo4j/newrelic.yml.j2
     dest: /etc/neo4j/newrelic.yml
-    # owner: do I need one?
+    owner: neo4j
     mode: 0644
 
 - name: restart neo4j
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -72,9 +72,35 @@
     - install
     - install:configuration
 
+- name: set to run neo4j with the newrelic javaagent
+  # see https://docs.newrelic.com/docs/agents/java-agent/installation/java-agent-manual-installation
+  lineinfile:
+    dest: "{{ neo4j_server_config_file }}"
+    regexp: "wrapper.java.additional.2="
+    line: "wrapper.java.additional.2=/etc/neo4j/newrelic.jar"
+  tags:
+    - install
+    - install:configuration
+
+- name: store jar file for javaagent
+  copy:
+    src: newrelic.jar
+    dest: /etc/neo4j
+    # just guessing? No idea what mode we'd need
+    mode: "0644"
+  tags:
+    - install
+    - install:configuration
+
+- template:
+    src: ../templates/etc/neo4j/newrelic.yml.j2
+    dest: /etc/neo4j/newrelic.yml
+    # owner: do I need one?
+    mode: 0644
+
 - name: restart neo4j
-  service: 
-    name: neo4j 
+  service:
+    name: neo4j
     state: restarted
   tags:
     - manage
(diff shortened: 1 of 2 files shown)


adampalay:
@benpatterson , would also be curious to get your thoughts on `template vs `lineinfile`

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -72,9 +72,35 @@
...
+    dest: "{{ neo4j_server_config_file }}"
+    regexp: "wrapper.java.additional.2="
+    line: "wrapper.java.additional.2=/etc/neo4j/newrelic.jar"
+  tags:
+    - install
+    - install:configuration
+
+- name: store jar file for javaagent
+  copy:
+    src: newrelic.jar
```
feanil, line 87: What version of agent is this?  Also can we download it on demand from newrelic as a part of the playbook.  Basically provide a URL to the agent jar and it gets downloaded onto the machine if it's a different version from the one that is currently installed there.  The way you have it now, I think we would need a code change before we can update the agent if it's already there on the remote machine.

benpatterson:
@adampalay the decision between `template` vs `lineinfile` for me comes down to whether or not a configuration file will arrive for free when something is installed. So, if installing the agent includes pulling down a yml file, then lineinfile sounds good to me. If, OTOH, the configuration file always has to be put in separately, I might instead have a pared-down yml file with only the configuration settings you want (and comments explaining what they are, and where to find the list of all conf options). HTH?

playbooks/roles/neo4j/tasks/main.yml
adampalay: It looks like we can't download it as part of the playbook—the various methods of downloading it from newrelic all involve clicking a button, while authenticated, to get a zip file that contains `newrelic.jar`: https://docs.newrelic.com/docs/agents/java-agent/installation/java-agent-manual-installation. I don't see a way just to get that agent itself.

I don't know how we'd get newrelic.jar onto the box without including it in a configuration file.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -72,9 +72,35 @@
...
+    # just guessing? No idea what mode we'd need
+    mode: "0644"
+  tags:
+    - install
+    - install:configuration
+
+- template:
+    src: ../templates/etc/neo4j/newrelic.yml.j2
+    dest: /etc/neo4j/newrelic.yml
+    # owner: do I need one?
```
fredsmith: you do need an owner.  make it the 'neo4j' user.

playbooks/roles/neo4j/tasks/main.yml
fredsmith: Because of the reasons adam cited, I think that pulling this in via a copy command is the right way to do this.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -72,9 +72,35 @@
...
+    line: "wrapper.java.additional.2=/etc/neo4j/newrelic.jar"
+  tags:
+    - install
+    - install:configuration
+
+- name: store jar file for javaagent
+  copy:
+    src: newrelic.jar
+    dest: /etc/neo4j
+    # just guessing? No idea what mode we'd need
```
fredsmith, line 89: needs an owner, 'neo4j'

fredsmith (changes requested):

playbooks/roles/neo4j/tasks/main.yml
adampalay: @feanil , other thoughts?

feanil: For things like this in the past we've uploaded them to s3.  I think that's better so that we have a backup of it that is not on a developer machine.  There is also a concern that if the file exists on the destination machine, it won't get updated, we should either do some version checks or always copy the file over to the destination.

adampalay: @feanil that makes sense! That way we can put it behind auth too!

edx-webhook:
Thanks for the pull request, @adampalay! I've created [OSPR-2354](https://openedx.atlassian.net/browse/OSPR-2354) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

mduboseedx:
@edx/devops Is this something that is still worth pursuing?

PR #3793, opened 2017-04-04
Rabbit will only use 40% of RAM, allow this to go up

jibsheet:
The linked documentation comes out ao .7 for a t2.large
We can also prevent rabbit from going to disk until more memory
is full, but that isn't needed right now.

### DIFF
--- a/playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
+++ b/playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
@@ -5,5 +5,6 @@
 {# 
    Note: That these names should include the node name prefix. eg. 'rabbit@hostname'
 #}
-  {cluster_nodes, {['{{ RABBITMQ_CLUSTERED_HOSTS|join("\',\'") }}'], disc}}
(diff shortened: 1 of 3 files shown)


playbooks/roles/rabbitmq/templates/etc/rabbitmq/rabbitmq.config.j2
```diff
@@ -5,5 +5,6 @@
 {# 
    Note: That these names should include the node name prefix. eg. 'rabbit@hostname'
 #}
-  {cluster_nodes, {['{{ RABBITMQ_CLUSTERED_HOSTS|join("\',\'") }}'], disc}}
+  {cluster_nodes, {['{{ RABBITMQ_CLUSTERED_HOSTS|join("\',\'") }}'], disc}},
+  {vm_memory_high_watermark, '{{ RABBITMQ_VM_MEMORY_HIGH_WATERMARK }}' }
```
feanil: I think the tests are failing because of the single quotes around the number.  I think it's expecting a float instead of a string.

jibsheet: Yep - I have some local changes related to this

jibsheet:
@feanil  ready for a re-review.  Want to merge so I can rebase the rabbit upgrade on top.

feanil (approved):

PR #3795, opened 2017-04-04
Updated base Dockerfiles

clintonb:


clintonb:
Thoughts @gsong?

FYI @rlucioni

rlucioni (approved):

gsong (approved):

clintonb:
@edx/devops any input/approvals?

feanil:
What is generating zombie processes?

clintonb:
We don't necessarily have zombie processes. Some processes may not respond appropriately to the exit signal, which could lead to Docker Compose timing out and potentially shutting down databases in the middle of write operations.

@gsong can provide a better explanation.

gsong:
I believe gunicorn expects SIGINT or SIGQUIT, and `manage runserver` expects SIGINT. Docker sends SIGTERM, so we need something to make that translation, such as [Tini](https://github.com/krallin/tini).

feanil:
Looks like supervisor handles all the signals except for SIGUSR1 in a resonable way: http://supervisord.org/running.html#signals

I'm concerned with making an unofficial container the base of our containers.  I think it's not a reasonable thing to do from a security perspective.

gsong:
@feanil I think the intent is that there are cases where we won't use supervisor. We can also just install Tini manually in the Dockerfile instead of using the image.

```
ARG TINI_VERSION
ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini /tini
RUN chmod +x /tini
```

You can then override entry points using Compose files for containers that need Tini:

```
entrypoint:
  - /tini
  - "--"
```

feanil:
@gsong It seems strange to have two tools that do similar things, in my mind, supervisor is just a python based init engine, not sure what the benefit is to having tini instead.  I buy that there might be slightly extra overhead with this, but I think before we commit to TINI we should update the OEP and get consensus.  This is a small technical change but not a small technology decision.

feanil:
At the very least the OEP should be corrected.

PR #3798, opened 2017-04-05
Dockerfile for notifier on 16.04

michaelroytman:
Configuration Pull Request
---


feanil:
@MichaelRoytman please add this to the list of docker files that get built as part of the travis tests.

feanil:
Assuming that passes, I think this is good to go.  Let me know when that's done and I'll re-review.

michaelroytman:
@feanil All set! If you could take a second look, that would be great. Thanks.

feanil (approved):

PR #3799, opened 2017-04-06
Quiet an ansible deprecation warnings

jibsheet:
[DEPRECATION WARNING]: Instead of sudo/sudo_user, use become/become_user and
make sure become_method is 'sudo' (default).
This feature will be removed in a
future release. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.

@edx/devops Please review - ansible has gotten noisier about this.

### DIFF
--- a/docker/plays/analytics_api.yml
+++ b/docker/plays/analytics_api.yml
@@ -1,6 +1,6 @@
 - name: Deploy Analytics API
   hosts: all
-  sudo: True
+  become: True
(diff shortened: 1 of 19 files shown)


feanil (approved):

jibsheet:
Ran all docker builds as expected https://travis-ci.org/edx/configuration/builds/219324989

PR #3800, opened 2017-04-07
Run edxapp on Docker for Devstack

jmbowman:
Changes to support building functional Docker images for edxapp devstack.  The main points:

* Override some variables to allow using mysql, memcached, etc. from other containers
* Include static assets in the devstack container
* Add a devstack.sh template conditional on the "devstack" tag, for consistency with other edX service containers
* Change Docker devstack's default LMS and Studio ports to make conflicts with Vagrant devstack less likely
* Add a new "edxapp_npm_production" variable to allow Docker devstack to install node devDependencies
* Configure Docker devstack's gunicorn to auto-reload on code changes
* Add nginx configuration
* Break out Elasticsearch host and port into separate variables which combine to form the current one
* Stream a concatenation of all relevant log files from the container
* Use edxapp's home directory instead of root's when gathering static assets (to use the correct npm package cache directory, `$HOME/.npm`)

---

### DIFF
--- a/docker/build/edxapp/Dockerfile
+++ b/docker/build/edxapp/Dockerfile
@@ -2,26 +2,25 @@
 #
 # From the root of configuration:
 #
-# docker build -f docker/build/edxapp/Dockerfile .
+# docker build -f docker/build/edxapp/Dockerfile . -t edxops/edxapp:latest
 #
 # This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
 # with the currently checked-out configuration repo.
 
+
 FROM edxops/precise-common:latest
 MAINTAINER edxops
-
 USER root
-RUN apt-get update
+CMD ["/edx/app/supervisor/venvs/supervisor/bin/supervisord", "-n", "--configuration", "/edx/app/supervisor/supervisord.conf"]
 
 ADD . /edx/app/edx_ansible/edx_ansible
-COPY docker/build/edxapp/ansible_overrides.yml /
 WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
 COPY docker/build/edxapp/ansible_overrides.yml /
-RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook \
-    edxapp.yml -i '127.0.0.1,' -c local \
-    -e "EDXAPP_PYTHON_SANDBOX=false" \
-    -e@/ansible_overrides.yml \
-    -t "install:base,install:configuration,install:system-requirements,install:app-requirements,install:code"
-WORKDIR /edx/app
-CMD ["/edx/app/supervisor/venvs/supervisor/bin/supervisord", "-n", "--configuration", "/edx/app/supervisor/supervisord.conf"]
-EXPOSE 8000 8010
+
+RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook edxapp.yml \
+    -c local -i '127.0.0.1,' \
+    -t 'install,assets,devstack' \
+    --extra-vars="@/ansible_overrides.yml"
+
+EXPOSE 18000 18010
--- a/docker/build/edxapp/ansible_overrides.yml
+++ b/docker/build/edxapp/ansible_overrides.yml
@@ -1,10 +1,29 @@
 ---
-
-DOCKER_TLD: "edx"
-
-# prevents Travis from giving up on the build
-COMMON_PIP_VERBOSITY: "-vvvv"
-
-EDXAPP_MYSQL_HOST: "db.{{ DOCKER_TLD }}"
+EDXAPP_SETTINGS: 'devstack_docker'
+EDXAPP_MEMCACHE: ['edx.devstack.memcached:11211']
+EDXAPP_MYSQL_HOST: "edx.devstack.mysql"
 EDXAPP_MONGO_HOSTS:
-  - "mongo.{{ DOCKER_TLD }}"
+  - "edx.devstack.mongo"
+
+devstack: true
+migrate_db: false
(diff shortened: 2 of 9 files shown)


clintonb (approved):

docker/plays/edxapp.yml
```diff
@@ -1,11 +1,19 @@
...
   hosts: all
-  sudo: True
+  become: True
   gather_facts: True
   vars:
     serial_count: 1
   serial: "{{ serial_count }}"
   roles:
     - common_vars
-    - docker
```
feanil: Why was the docker role removed?

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -36,6 +36,17 @@
...
 
+- name: write devstack script
+  template:
+    src: "devstack.sh.j2"
+    dest: "{{ edxapp_app_dir }}/devstack.sh"
+    owner: "{{ supervisor_user }}"
+    group: "{{ common_web_user }}"
+    mode: 0744
+  tags:
+    - devstack
```
feanil: Rather than using a devstack tag, you sholud conditionalize this on `devstack` var.  The reason for this is that we don't skip this tag when building production AMIs. Optionally you could add skipping this tag to the Pipelines and then this would be fine.

clintonb: Example: https://github.com/edx/configuration/pull/3549/files#diff-6bb3ff2dddf172c72df3a61b5de44e12R95

docker/plays/edxapp.yml
jmbowman: I'm not sure exactly, that change predated my work on this.  It doesn't look like the docker role (or the container role it lists as a dependency) really has any content yet; do we plan to actually populate those or deprecate and remove them?  @clintonb , do you recall why you [originally removed this](https://github.com/edx/configuration/commit/3900bbf57ea63b01bd7ff63938b26dbc3082b12e)?

gsong (approved):

playbooks/roles/edxapp/tasks/main.yml
jmbowman: Should I just add the condition [like here](https://github.com/edx/configuration/pull/3549/files#diff-6bb3ff2dddf172c72df3a61b5de44e12R135) (but keying off of the `devstack` variable instead), or also replace all usage of the devstack tag with the appropriate install sub-tag instead?

docker/plays/edxapp.yml
feanil: There are vars in them that affect things and they are the equivalent of the aws role in the edx-east playbooks which allow us a common place to commit common changes related to docker.  I think it should be added back.

playbooks/roles/edxapp/tasks/main.yml
feanil: the `edx_django_service_is_devstack` is specific to the new role that clinton linked to.  The edxapp role does not rely on that yet so it should not use that var, I think for now you should use the `devstack` var.

docker/plays/edxapp.yml
clintonb: docker inherits container which inherits common. edxapp already uses common.

feanil: The goal is not just to inherit common which is independent of architecture but to inherit docker which can have docker specific config in them.  We found that we needed something like this for AWS and adding it own after the fact is more costly than having an obvious place to put docker specific settings that are common to all edx containers.

clintonb: Got it. Thanks for clarifying.

playbooks/roles/edxapp/tasks/main.yml
jmbowman: Right, that was the plan; I was just wondering if I should also keep the reference to the devstack and devstack:install tags or replace them with install and install:base.

feanil: I think you can leave the devstack and devstack:install tags.

jmbowman:
@feanil Does this look ok now with the latest changes?

feanil (approved):

docker/build/edxapp/Dockerfile
```diff
@@ -2,26 +2,25 @@
 #
 # From the root of configuration:
 #
-# docker build -f docker/build/edxapp/Dockerfile .
+# docker build -f docker/build/edxapp/Dockerfile . -t edxops/edxapp:devstack
```
clintonb: :latest

docker/build/edxapp/ansible_overrides.yml
```diff
@@ -1,10 +1,29 @@
...
-# prevents Travis from giving up on the build
-COMMON_PIP_VERBOSITY: "-vvvv"
+EDXAPP_LMS_GUNICORN_EXTRA_CONF: 'reload = True'
 
-EDXAPP_MYSQL_HOST: "db.{{ DOCKER_TLD }}"
-EDXAPP_MONGO_HOSTS:
-  - "mongo.{{ DOCKER_TLD }}"
+EDXAPP_NO_PREREQ_INSTALL: 0
+COMMON_SSH_PASSWORD_AUTH: "yes"
+EDXAPP_LMS_BASE: "edx.devstack.edxapp:18000"
```
clintonb: s/edxapp/lms

docker/build/edxapp/ansible_overrides.yml
```diff
@@ -1,10 +1,29 @@
...
+EDXAPP_CMS_BASE: "edx.devstack.edxapp:18010"
```
clintonb: s/edxapp/studio

clintonb (approved):

PR #3801, opened 2017-04-10
Determine default network interface name instead of hardcoding to eth0.

mtyaka:
Depending on the hardware configuration, the default network interface is not always named `eth0` in Ubuntu 16.04. On some EC2 VM types, the interface is named `ens3`, for example.

This patch adds a task to determine the interface name using `ip route` command instead of     hardcoding to `eth0`.
    
Also, since `/etc/network/interfaces.d/<interface-name>.cfg` files no longer exist in Ubuntu 16.04, this patch adds a script to `/etc/network/if-up.d` to adjust the MTU value at network startup instead.

**JIRA Ticket**: [OSPR-1746](https://openedx.atlassian.net/browse/OSPR-1746)

**Merge deadline**: None

**Partner information**: 3rd party-hosted open edX instance

---

### DIFF
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -31,7 +31,7 @@
     - deploy
 
 - name: Get name of default network interface
-  shell: "ip route show 0/0 | awk '{print $NF}'"
+  shell: "ip route show 0/0 | grep -oP '(?<=dev )[[:alnum:]]+'"
   register: get_interface_name
   when: ansible_distribution in common_debian_variants
 
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -30,13 +30,8 @@
   tags:
     - deploy
 
-- name: Get name of default network interface
-  shell: "ip route show 0/0 | grep -oP '(?<=dev )[[:alnum:]]+'"
-  register: get_interface_name
-  when: ansible_distribution in common_debian_variants
-
 - name: Set the MTU to 1500 temporarily
-  shell: "/sbin/ifconfig {{ get_interface_name.stdout }} mtu 1500 up"
+  shell: "/sbin/ifconfig {{ ansible_default_ipv4.interface }} mtu 1500 up"
   when: ansible_distribution in common_debian_variants
 
 - name: Set the MTU to 1500 permanently
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -30,24 +30,23 @@
   tags:
     - deploy
 
+- name: Get name of default network interface
+  shell: "ip route show 0/0 | awk '{print $NF}'"
+  register: get_interface_name
+  when: ansible_distribution in common_debian_variants
+
 - name: Set the MTU to 1500 temporarily
-  shell: /sbin/ifconfig eth0 mtu 1500 up
+  shell: "/sbin/ifconfig {{ get_interface_name.stdout }} mtu 1500 up"
   when: ansible_distribution in common_debian_variants
 
-- name: Check for eth0.cfg
-  stat:
-    path: /etc/network/interfaces.d/eth0.cfg
-  register: eth0_cfg
+- name: Set the MTU to 1500 permanently
+  template:
+    dest: /etc/network/if-up.d/mtu
+    src: mtu.j2
+    mode: 0755
+    owner: root
+    group: root
   when: ansible_distribution in common_debian_variants
-
-- name: Set the MTU to 1500 inside eth0.cfg file
-  lineinfile:
-    dest: /etc/network/interfaces.d/eth0.cfg
-    regexp: "^post-up /sbin/ifconfig eth0 mtu 1500"
-    line: "post-up /sbin/ifconfig eth0 mtu 1500"
-    insertafter: "^iface"
-  when: ansible_distribution in common_debian_variants and eth0_cfg.stat.exists
-
 #
 # End dealing with Jumbo frames issue in mixed MTU deployements in AWS
 #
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @mtyaka! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=3801&repo=edx%2Fconfiguration"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

smarnach:
@mtyaka The actual issue here seems to be the granularity of roles.  The `aws` role is meant to be used for instances running on AWS.  We abuse this role to only get the log syncing to S3.  In the future, more stuff might be added to the `aws` role that only works when actually running on AWS.

So my proposal would be to split the S3 log syncing out to a separate role, and make the `aws` role depend on that new role.  This change would be backwards compatible, and allow us to just include the new S3 log sync role.

mtyaka:
@smarnach I agree that it would make sense to split the log syncing logic out of the `aws` role, however I've also run into this MTU issue when deploying to AWS where the name of the interface seems to depend on the instance type. I've seen it when using a `m4.large` instance type in the `us-east-1` region.

smarnach:
@mtyaka That seems to be a separate bug.  I would presume that on AWS we would want to always set the MTU, regardless of the name of the interface, though I don't understand in what circumstances the step is necessary in the first place.

It might be possible to get the interface name from the Ansible or EC2 facts.  In any case, it's easy to parse the interface name from the output of `ip route show 0/0`.

So my preferred solution would be to fix these issues separately.  However, since I don't know exactly when we need to set the MTU, your solution might cover all cases correctly, and it's simpler than what I propose.

pomegranited:
@mtyaka I agree with @smarnach..  I think [we always want MTU 1500](http://docs.aws.amazon.com/AWSEC2/latest/WindowsGuide/network_mtu.html#jumbo_frame_instances) for traffic outside of the VPC, so it's better to set the MTU for the default network interface, regardless of whatever it's called.

Can you parse the interface name [as he suggests](https://github.com/edx/configuration/pull/3801#issuecomment-292904228), and set the MTU on that interface?

mtyaka:
@smarnach @pomegranited Thanks, I agree always setting MTU would be best. I'll try to do that; thanks for the hint. I'll also have to find a way to set the MTU permanently. The [current solution](https://github.com/open-craft/configuration/blob/f108e9bfa2fe66684bf21888ae37d82fec1176fd/playbooks/roles/aws/tasks/main.yml#L43-L55) does not work on Ubuntu 16.04 at all, because `eth0.cfg` file does not exist. There's only `50-cloud-init.cfg` file inside `/etc/network/interfaces.d/` and I'll have to figure out how it all works.

pomegranited:
@mtyaka Is there a `/etc/network/if-up.d` dir?  If so, you might be able to create a new file that just sets the MTU on ifup.  cf https://ubuntuforums.org/showthread.php?t=903788

mtyaka:
@pomegranited Oh nice! Yes `/etc/networks/if-up.d` exists, thanks for the tip :)

feanil:
@mtyaka I like @pomegranited 's idea better than just making the tasks conditional on eth0,  all the interfaces should have the same MTU of 1500p

mtyaka:
Thanks everyone! I updated the patch to determine the interface name using `ip route show 0/0` instead of hardcoding to `eth0`. I also replaced the `/etc/network/interfaces.d/<interface-name>.cfg` configuration file (which no longer exists on Ubuntu 16.04) with a script in `/etc/network/if-up.d/`, which sets the MTU at network startup.

edx-webhook:
Thanks for the pull request, @mtyaka! I've created [OSPR-1746](https://openedx.atlassian.net/browse/OSPR-1746) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

playbooks/roles/aws/tasks/main.yml
```diff
@@ -30,24 +30,23 @@
   tags:
     - deploy
 
-- name: Set the MTU to 1500 temporarily
-  shell: /sbin/ifconfig eth0 mtu 1500 up
+- name: Get name of default network interface
+  shell: "ip route show 0/0 | awk '{print $NF}'"
```
smarnach: The interface name is not guaranteed to be the last bit in the output; e.g. on my zesty box, it looks like this:

    $ ip route show 0/0
    default via 192.168.178.1 dev enp0s25 proto static metric 100 

So I propose to grep for the bit following "dev" instead, which I expect to be more robust:

    ip route show 0/0 | grep -oP '(?<=dev )[[:alnum:]]+'

mtyaka: Ah, nice catch, thanks @smarnach! I updated the code.

jibsheet:
Can you use this same var in the dhclient.conf that's templated in from vhost?  Otherwise that'll go out of skew (with that change, I think your branch hits all the eth0s)

mtyaka:
Thanks @jibsheet, I updated dhclient.conf.j2 to use default interface name.

I also discovered that ansible already has a built in variable that provides info about the default network interface, so I removed the task that was parsing the name out of `ip route show` command.

jibsheet:
@mtyaka thanks, I restarted the test shard that failed (looked like a network timeout to me).
I'll try to rereview, but may be Friday/Monday given other constraints.

mtyaka:
Thanks @jibsheet!

jibsheet:
@smarnach Did you want to look this over again?  You had some early input.  I'm expecting to finish a rereview/test and merging on Monday.

smarnach:
@jibsheet I took another look, and everything looks good to me.

jibsheet (approved):

jibsheet:
I'll click merge in the morning, thanks again.

mtyaka:
Thanks @jibsheet!

PR #3805, opened 2017-04-12
Be able to override the django sql CONN_MAX_AGE setting.

feanil:
Configuration Pull Request
---

@edx/devops @ormsbee


jibsheet (approved):

ormsbee (approved):
Argh. For some reason I thought we were just splatting these straight from the JSON file. Sorry for not making this PR myself. :-(

PR #3808, opened 2017-04-12
Make changes for running analyticstack on 16.04.

brianhw:
* Replace llvm-gcc with gcc.
* Define new Hadoop services.
* Add retry when first using Hadoop.
* Add procps to debian packages to address bug.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/analytics_pipeline/tasks/main.yml
+++ b/playbooks/roles/analytics_pipeline/tasks/main.yml
@@ -127,21 +127,46 @@
     - install
     - install:app-requirements
 
-- name: Ensure hdfs services are started
+- name: reload systemd configuration
+  command: systemctl daemon-reload
+  tags:
+    - install
+    - install:configuration
+
+- name: enable Hadoop services
   service:
-    name: hdfs
+    name: "{{ item }}"
+    enabled: yes
+  with_items: "{{ hadoop_common_services }}"
+  tags:
+    - install
+    - install:configuration
+
+- name: start Hadoop services
+  service:
+    name: "{{ item }}"
     state: started
+  with_items: "{{ hadoop_common_services }}"
   tags:
     - manage
     - manage:start
 
-- name: Ensure map reduce services are started
+- name: stop Hadoop services
   service:
-    name: yarn
-    state: started
+    name: "{{ item }}"
+    state: stopped
+  with_items: "{{ hadoop_common_services }}"
   tags:
-    - manage
+    - manage:stop
+
+- name: restart Hadoop services
+  service:
+    name: "{{ item }}"
+    state: restarted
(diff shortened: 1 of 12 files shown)


brianhw:
This has finally been tested against a matching version of edx-analytics-pipeline (see https://github.com/edx/edx-analytics-pipeline/pull/379), and all acceptance tests pass.   (Or, rather, no acceptance tests fail.)

@edx/devops This is ready for final review.     FYI:  @tobz , @HassanJaveed84 , @azafty468.

@nedbat FYI. Once merged, ficus.master should be ready for ficus.3.

brianhw:
@tobz I cloned your system definitions -- hopefully I did it the right way.

playbooks/roles/analytics_pipeline/tasks/main.yml
```diff
@@ -127,21 +127,38 @@
...
-    name: hdfs
+    name: "{{ item }}"
+    enabled: yes
     state: started
+  with_items: "{{ hadoop_common_services }}"
   tags:
-    - manage
-    - manage:start
+    - install
+    - install:configuration
```
fredsmith: I think this is still manage, manage:start, not install, install:configuration.   Can you separate enabled and state into two tasks and tag the enabled:true one install, install:configuration and the state: started one manage, manage:start ?

playbooks/roles/hadoop_common/templates/hdfs-datanode.service.j2
```diff
@@ -0,0 +1,12 @@
+[Unit]
```
fredsmith: Can all these files go in the templates/etc/systemd/system/ directory in the role?  This is our standard practice for templates.

fredsmith (changes requested):
Two minor chages, otherwise looks good.

playbooks/roles/analytics_pipeline/tasks/main.yml
brianhw: I'll give it a try -- thanks!

playbooks/roles/hadoop_common/templates/hdfs-datanode.service.j2
brianhw: Yes, I will.

nedbat (approved):
Seems good, though I know nothing about hadoop-world.

brianhw:
@fredsmith I made the changes you requested, rebuilt an analyticstack box, and ran some edx-analytics-pipeline acceptance tests against it successfully.  I ran some tests both before and after a "vagrant halt", to confirm that pipeline services were available.   Awaiting a thumb.

fredsmith (approved):

PR #3811, opened 2017-04-13
We need to be able to run on sandboxes/devstacks

jibsheet:
We'll safely skip the install of rabbitmq if something is already
installed, so warning is sufficient for our purposes.

@edx/devops

### DIFF
--- a/playbooks/roles/rabbitmq/defaults/main.yml
+++ b/playbooks/roles/rabbitmq/defaults/main.yml
@@ -11,10 +11,6 @@
 # Defaults for role rabbitmq
 # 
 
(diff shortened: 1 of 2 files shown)


jibsheet:
@jbarciauskas FYI

jbarciauskas:
Rewriting history will never feel normal to me

jibsheet:
Not clear how we're rewriting history?  Just removing too vigorous a test?

fredsmith (approved):

jbarciauskas:
Sorry, was being silly. Just a general  c on git workflows.

jbarciauskas:
*comment

PR #3812, opened 2017-04-13
Fixed CORS for credentials

clintonb:
LEARNER-568


clintonb:
@edx/devops please review

jibsheet (approved):

PR #3817, opened 2017-04-14
ENT-306: Added a feature flag to show/hide account activation message on edx-platform dashboard

saleem-latif:
Hi @mattdrayer , @brittneyexline , @feanil 
Please take a look at this PR. It adds FEATURE Flag to configuration introduced in https://github.com/edx/edx-platform/pull/14874.

Configuration Pull Request
---

  - [ ] Completed [Ops Ansible test checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)  


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -635,6 +636,9 @@ EDXAPP_COMPREHENSIVE_THEME_VERSION: ""
 EDXAPP_SOCIAL_AUTH_SAML_SP_PRIVATE_KEY: ''
 EDXAPP_SOCIAL_AUTH_SAML_SP_PUBLIC_CERT: ''
 
+# Account activation message flag
+EDXAPP_DISPLAY_ACCOUNT_ACTIVATION_MESSAGE_ON_DASHBOARD: true
```
feanil: Does this need to be a permanent feature flag? Are three times when you wouldn't want this to be true?

brittneyexline: Hi @feanil , there is a possibility that this could be changed to false (though unlikely), but I'm curious if there are other options for flagging that you think would be more appropriate?

haikuginger: I'm confused as to why the default value for this variable appears _after_ its use in `EDXAPP_FEATURES`. Can you clarify this?

haikuginger (changes requested):

playbooks/roles/edxapp/defaults/main.yml
saleem-latif: @haikuginger I have seen usages of variables before their declaration and they work, But I am moving variable declaration before its usage to be on the safe side.

feanil:
@brittneyexline responding to the conversation on the outdated diff.  Feature flags in the dictionary are permanent and hard to change.  It sounds like this is a thing where we want the feature flag so that we can push out the code while leaving the flag off but then turn it on so that the feature can be flipped independent of the release, is that right?  If so, then this is a really good use case for a temporary waffle flag.

saleem-latif:
closing this for now, I will reopen it once https://openedx.atlassian.net/browse/ENT-363 is prioritized and pulled into a sprint.

PR #3819, opened 2017-04-14
Credentials CORS fixes

clintonb:
Finally figured out where I misunderstood nginx

LEARNER-568


jibsheet:
nginx is so misunderstood

clintonb:
Coincidentally, you mentioned if statements this morning!

PR #3820, opened 2017-04-14
[ENT-327] Set SESSION_EXPIRE_AT_BROWSER_CLOSE for IDAs

haikuginger:
This pull request adds the `EDXAPP_SESSION_EXPIRE_AT_BROWSER_CLOSE` setting. This is applied to edxapp as well as all relevant IDAs. When set to True, this setting ensures that, by default, sessions expire on browser close.

For the LMS, the only implication is for SSO users; users logging in using the standard login view will have their session explicitly set regardless. SSO users will have their sessions set to an explicit expiration time by the `social` app if that information is available; otherwise, their sessions will expire on browser close.

For the IDAs, the implication is that sessions will terminate on browser close. However, this should have minimal UI implications, since the IDAs process authentication through OIDC. They will, from the user's perspective, inherit the LMS's session state, whether that's long-lived (from selecting the "keep me logged in" checkbox) or short-lived (by leaving the same checkbox blank).

A sandbox will be available shortly demonstrating the implications of these changes on the LMS; it does require a small code change for the LMS to retrieve this variable from ENV_TOKENS. None of the IDAs require code changes to make use of this variable.

New variables:

- `EDXAPP_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `ECOMMERCE_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `CREDENTIALS_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `INSIGHTS_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `DISCOVERY_SESSION_EXPIRE_AT_BROWSER_CLOSE`

---


openedx-webhooks:

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?repo=edx%2Fconfiguration&number=3820"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -52,6 +52,8 @@ CREDENTIALS_LOGOUT_URL: '{{ CREDENTIALS_URL_ROOT }}/logout/'
 CREDENTIALS_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
 CREDENTIALS_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
 
+CREDENTIALS_SESSION_EXPIRE_AT_BROWSER_CLOSE: '{{ EDXAPP_SESSION_EXPIRE_AT_BROWSER_CLOSE }}'
```
clintonb: We should not tie our services together in this manner. The bits for JWT above are the exception, not the rule. Set the default to `false`, and override `CREDENTIALS_SESSION_EXPIRE_AT_BROWSER_CLOSE` in edx/edx-internal. This should be done for all IDAs.

clintonb (changes requested):

playbooks/roles/credentials/defaults/main.yml
haikuginger: Okay; thanks for the advice!

haikuginger: @clintonb, done.

clintonb (approved):
Squash your commits. @edx/devops is there anything we need to do to get this into edx/edx-internal, or is it a simple matter of adding the variables there?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -428,6 +428,7 @@ EDXAPP_MICROSITE_CONFIGURATION: {}
 EDXAPP_COURSES_WITH_UNSAFE_CODE: []
 EDXAPP_SESSION_COOKIE_DOMAIN: ""
 EDXAPP_SESSION_COOKIE_NAME: "sessionid"
+EDXAPP_SESSION_EXPIRE_AT_BROWSER_CLOSE: false
```
bradenmacdonald: Please add a comment here indicating that this variable currently only affects SSO sessions, since the regular app configures normal login sessions differently. I think without that explanation, this variable will seem to be broken based on a face-value understanding of what it does.

(I agree that this variable name makes sense though since it matches the django variable that it overrides).

bradenmacdonald: I would also like to know whether this change affects Google/Facebook login users - if it doesn't we should mention that in the comment as well, and if it does we should consider whether we still want to use this approach, making those users re-login every time. I don't know what % of edx.org users use google/fb but I imagine it's not insignificant.

brittneyexline (commented):
I agree with @bradenmacdonald 's comment about the potential effect on Google/Facebook SSO users. If the behavior affects them that is problematic - which I suppose is why our original plan was for this behavior to be configurable by SSO provider.

haikuginger:
This should not affect Facebook/Google SSO, because the backends for those providers support explicitly-set expiry times, which the `social` app will set when logging the user in. This should only affect users logging in with backends (e.g., SAML) that do not currently support a session expiry time.

brittneyexline (approved):
Thanks for the clarification @haikuginger

feanil:
If the vars are specific to SSO, along with the comment that @bradenmacdonald asked for, what about changing the variables to be specific to SSO?

eg. `EDXAPP_SSO_SESSION_EXPIRE_AT_BROWSER_CLOSE`

haikuginger:
@feanil, we could do that, but they're only specific to SSO because we implement normal signin that way - as @bradenmacdonald mentioned, the name is canonical to Django, so I'd prefer to keep it that way.


@bradenmacdonald, @brittneyexline:

Results of testing with Facebook and Google were not great. Facebook doesn't provide an expiry parameter, so we fell back to the default session length. Google does provide an expiry parameter.

Unfortunately, it turns out that we're not actually using those at the moment, so even if both did provide long-lived expiry parameters, we'd still set session-level cookies instead of long-lived ones.

Luckily, in the course of investigating this, I did figure out what I think is a _very_ hackish way to do this.

For a given `AuthBackend` we control (that's only SAML; we use the other ones as they're given to us by social), we can override a method `extra_data`, and add additional data to the `JSONField` that gets stored on the `UserSocialAuth` database entry for a user. This includes the `expires` value, which is what gets used to determine the length of a session (when that feature is enabled). So that's one part.

We also need to actually appear to enable that feature... but only when using a SAML provider that has early session expiry enabled. We can do this by adding another override to our existing custom Django TPA strategy.

Combined, these overrides will allow SAML providers to set a flag that allows sessions to be expired on browser close, rather than on the platform default.

I still think this PR is worth doing; we definitely want to set this value to True for the relevant IDAs to get the full effect. I'll be adding the above to the existing edx-platform PR.

clintonb:
> Luckily, in the course of investigating this, I did figure out what I think is a very hackish way to do this.

No hacks. Hacks will almost assuredly lead to problems down the road. Given the low priority of this functionality, if this cannot be done without hacks don't do it at all.

bradenmacdonald:
@haikuginger Thanks for investigating Facebook and Google. I think that given those findings, we can't use this approach for the platform (though it does make sense for the IDAs). I think the best way forward is to modify the python-social code to properly provide a way for the SAML backend to control its session expiry time, and then to modify our SAMLProviderConfig class to allow us to configure the session expiry on a per-provider basis, via the django admin.

>  we can override a method extra_data, and add additional data to the JSONField that gets stored on the UserSocialAuth database entry for a user. This includes the expires value, which is what gets used to determine the length of a session (when that feature is enabled).

I really don't like that part of this. We should be changing this in the backend code and python-social code only. Settings about how the provider works should not be being persisted into the UserSocialAuth table. I agree with Clinton that we shouldn't go for a hacky fix here. Don't shy away from making changes to python-social or to python-saml in order to get a clean approach that we can live with. We'll do a story next sprint to upstream them.

haikuginger:
@bradenmacdonald, okay, sounds good. Regardless, the time spent on this was necessary to figure exactly what modifications of that nature would need to look like.

haikuginger:
@bradenmacdonald, your comment request has been addressed.

To summarize, we've found a non-hackish way to implement exactly what we need in edx/python-social-auth#1 and #14900. However, we still need this configuration change so that we can apply short-lived sessions to ecommerce and the other OIDC-based IDAs, and #14897 is still "nice to have" so that we can take advantage of this setting in future without any code changes.

@brittneyexline, if you can confirm your 👍 with that in mind, that'd be great.

Subsequent of this change, we'll need to set the following variables to `True` for production:

- `ECOMMERCE_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `CREDENTIALS_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `INSIGHTS_SESSION_EXPIRE_AT_BROWSER_CLOSE`
- `DISCOVERY_SESSION_EXPIRE_AT_BROWSER_CLOSE`

brittneyexline (approved):
confirming my 👍

haikuginger:
@bradenmacdonald, I need your 👍 ASAP, please.

bradenmacdonald:
@haikuginger Is there a reason to still include `EDXAPP_SESSION_EXPIRE_AT_BROWSER_CLOSE` in this PR since we won't need it, and its behavior is somewhat unintuitive? I'd rather just leave it out.

Otherwise I'm 👍  on this approach but have not yet been able to test this out until the mentioned sandbox is available.

haikuginger:
@bradenmacdonald, not for our specific use case. The name is slightly confusing, given what our code does, but it is a standard Django setting, so that ought to counteract it somewhat.

haikuginger:
@bradenmacdonald, you can check a sandbox at haikuginger.sandbox.edx.org; I've added you as an ssh user, so you can ssh in, take a look at the edx-platform and ecommerce vars files, as well as enter the Django shell for either and verify that the setting is accessible there.

bradenmacdonald:
@haikuginger The main thing I want to test is the OIDC logout, but I can't log in to https://ecommerce-haikuginger.sandbox.edx.org/coupons/ . It just gives a server error due to a missing SiteConfiguration - am I doing something wrong?

Also, as mentioned on https://github.com/edx/edx-platform/pull/14900 , we need to be aware that these settings don't seem to be of use for browsers like Chrome and Firefox that don't respect session cookies by default.

I'm going to be travelling this week, so don't consider me a blocker on this PR. I'm good with the approach, though I'd like someone to have verified it with the IDA(s) on the sandbox, and I'd prefer that we exclude the edx-platform change for now, since it's potentially confusing and we don't have a use case for it.

haikuginger:
Agreed: re the edx-platform PR. We'll wait on that until we can figure out the appropriate language to avoid confusion. It's no longer an essential part of this process anyway.

haikuginger:
Updated from 09bd930cd; removed the edxapp vars, since they seem to be a source of confusion. Will merge once we have a green build.

haikuginger:
Testable at https://ecommerce-haikuginger.sandbox.edx.org/; log in using OIDC, then quit the browser. Open it again, and verify that you're no longer logged into Ecommerce. Merging now.

PR #3821, opened 2017-04-14
Updated CORS configuration for credentials

clintonb:
We have invested *numerous* man-hours in attempting to provide a
properly-secured CORS implementation. Unfortunately, our efforts
have only resulted in a lot of frustration, failure, and new
knowledge. For now, given the low risk of exposing these files, we
will move forward with the insecure option.

LEARNER-568


clintonb:
FYI @edx/devops going with the insecure option for now. Ironically, I probably could spent the last 48-72 hours solving the static assets on S3 problem, thereby avoiding having to touch nginx altogether. (facepalm)

clintonb:
P.S. I will limit this to just fonts on Tuesday of next week. For now, I just want to move to something else. I've been working on this for 16+ hours per day for the last week.

PR #3824, opened 2017-04-19
bake pip requirements into the loadtest worker AMI

pwnage101:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
+++ b/playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
@@ -1,0 +1,36 @@
+---
+# Create a virtualenv for edx-load-tests by installing the requirements and
+# packaging the virtualenv.
+
+- name: Create shallow clone of edx-load-tests
+  git:
+    repo: https://github.com/edx/edx-load-tests.git
+    dest: "{{ jenkins_home }}/shallow-clone"
+    version: "master"
+    depth: 1
+  become_user: "{{ jenkins_user }}"
+
+- name: Install most edx-load-tests requirements using pip
+  pip:
+    requirements: "{{ jenkins_home }}/shallow-clone/requirements/{{ item }}"
+    extra_args: "--exists-action=w"
+    virtualenv: "{{ jenkins_home }}/edx-venv"
+    virtualenv_command: virtualenv
+  with_items:
+    - base.txt
+  become_user: "{{ jenkins_user }}"
+
+# Archive the current state of the virtualenv as a starting point for new
+# builds.  The edx-venv directory is deleted and then recreated cleanly from
+# the archive by the jenkins build scripts.
+#
+# TODO: after we migrate to ansible 2.3+, change this task to use the archive
+# module.  http://docs.ansible.com/ansible/archive_module.html
+- name: Create a clean virtualenv archive
+  command: "tar -cpzf edx-venv_clean.tar.gz edx-venv"
+  args:
+    chdir: "{{ jenkins_home }}"
+  become_user: "{{ jenkins_user }}"
+
+- name: Remove the shallow-clone
+  file: path={{ jenkins_home }}/shallow-clone state=absent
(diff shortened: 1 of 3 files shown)


util/packer/jenkins_worker_loadtest.json
```diff
@@ -2,6 +2,7 @@
   "variables": {
     "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
     "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
+    "new_relic_key": "{{env `NEW_RELIC_KEY`}}",
```
ormsbee, line 5: Devops is ok with this from a usage/billing point of view, right?

ormsbee:
@jibsheet: Do you have time to review this?

util/packer/jenkins_worker_loadtest.json
pwnage101: I spoke with Ben who added newrelic support to the platform workers last month, and he said the cost was insubstantial or somehow lumped into some larger package.  Would like to hear what devops knows.

pwnage101: I personally don't even care if these instances get newrelic, this fix was just a little easier than the alternative (modifying an ansible role to conditionally install newrelic).

jibsheet: NewRelic as a system monitor (CPU/RAM etc) is baked into our package.
New Relic as an app monkeypatcher is $$

jibsheet:
@ormsbee I'll look this and Troy's DSL over tomorrow

playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
```diff
@@ -0,0 +1,33 @@
+---
+# Create a virtualenv for edx-load-tests by installing the requirements and
+# packaging the virtualenv.
+
+- name: Create shallow clone of edx-load-tests
+  git:
+    repo: https://github.com/edx/edx-load-tests.git
+    dest: "{{ jenkins_home }}/shallow-clone"
+    version: "master"
```
jibsheet, line 9: Do you ever see wanting to bake a load test driver with not-master?  You could make this "{{ LOADTEST_DRIVER_VERSION }}" and pass it down with -e in the packer.  It would mean ensuring that the packer builder has an arg for you.

Just a thought, not anything you need to change, probably 20 minutes of work when you do need it eventually.

playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
```diff
@@ -0,0 +1,33 @@
...
+    virtualenv_command: virtualenv
+  with_items:
+    - base.txt
+  become_user: "{{ jenkins_user }}"
+
+# Archive the current state of the virtualenv as a starting point for new
+# builds.  The edx-venv directory is deleted and then recreated cleanly from
+# the archive by the jenkins build scripts.
+- name: Create a clean virtualenv archive
+  command: "tar -cpzf edx-venv_clean.tar.gz edx-venv"
```
jibsheet: When we eventually get to 2.3, we should replace a lot of things with http://docs.ansible.com/ansible/archive_module.html

jibsheet (approved):
Just a few notes for future dev, the code as is looks good.

playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
pwnage101: Since baking AMIs and installing them into jenkins is still a manual process, supporting baking arbitrary branches doesn't win us very much.

Later down the line when we have some infrastructure around running packer, saving that AMI as a jenkins artifact, and passing that through something to reconfigure jenkins master (doesn't exist yet AFAIK), I may start to think about this.

playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
pwnage101: ah, will add a note.

playbooks/roles/jenkins_worker/tasks/loadtest_driver_worker.yml
jibsheet: Sounds good to me

PR #3825, opened 2017-04-19
chmod and chown the edx_ansible_venv_dir to make vagrant provision more reliable

macdiesel:

### DIFF
--- a/playbooks/roles/edx_ansible/tasks/deploy.yml
+++ b/playbooks/roles/edx_ansible/tasks/deploy.yml
@@ -10,6 +10,16 @@
     - install
     - install:code
 
+- name: Ensure directory permissions for Install edx_ansible venv requirements
+  command: chmod -R 755 "{{ edx_ansible_venv_dir }}"
+  become: yes
+  become_user: root
+
+- name: Ensure directory owner for Install edx_ansible venv requirements
+  command: chown -R "{{ edx_ansible_user }}" "{{ edx_ansible_venv_dir }}"
+  become: yes
+  become_user: root
+
 - name: Install edx_ansible venv requirements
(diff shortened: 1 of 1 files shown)


doctoryes (approved):
I hit this problem when configuring a new open-release/ficus.2 Vagrant devstack. I solved it using `sudo chown -R edx-ansible.edx-ansible /edx/app/edx_ansible/venvs/edx_ansible/lib` as the vagrant user - but this Ansible addition would work better.

fredsmith (approved):

jibsheet:
While this may help devstacks, in Slack someone is reporting failed sandboxes and @benpatterson is reporting failing packer builds.

please revert until it can be tested further.

doctoryes:
Is the gauntlet of tests which need to be run for particular configuration changes written down anywhere? If so, perhaps a link to it should be in the repo commit message.

jibsheet:
@doctoryes  https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist

If you want to PR it into the .github/PULL_REQUEST_TEMPLAT.md I won't object.
I expect it may want some updating for the GoCD world we now live in.

benpatterson:
@doctoryes @jibsheet speaking for the packer builds, that's something we *could* hook up; however, we haven't added it to CI in here because it can take 1.5-2 hrs (part of what it does is build the venv from scratch).

jibsheet:
Yes, we don't run our full CI because the hourly one takes 70 minutes and the from scratch one takes a few hours.  We provide instructions to run them yourself and examine the output on potentially breaking changes (link above).  Same link refers to packer I believe.

PR #3828, opened 2017-04-20
Always reboot when taking an AMI

jibsheet:
Not doing this means we can't guarantee filesystem integrity
and we've seen this bear out recently when several 16.04 instances had 0
length minos installs (the last task to run before the AMI is cut)


jibsheet:
@edx/devops

fredsmith (approved):

PR #3829, opened 2017-04-20
Added Vary header to credentials /static/ path

clintonb:
This header will provide more information to downstream caches so they know which headers to consider when determining a match. This is especially important to ensure browsers do not cache incorrect CORS headers when switching between sites/tenants.

DEVOPS-5807


clintonb:
FYI @rlucioni @edx/devops

rlucioni:
@clintonb does this require any changes to the configuration of CloudFront distributions for credentials? Do we need to whitelist this header?

rlucioni (approved):

clintonb:
The Origin header should already be whitelisted. We may need to invalidate caches again.

rlucioni:
For reference: http://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/RequestAndResponseBehaviorCustomOrigin.html#ResponseCustomRemovedHeaders

PR #3833, opened 2017-04-25
Added asymmetric JWS key settings for edxapp

clintonb:
ECOM-3629


clintonb:
@edx/devops please review

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -913,6 +913,8 @@ generic_env_config:  &edxapp_generic_env
   LOG_DIR:  "{{ COMMON_DATA_DIR }}/logs/edx"
   JWT_ISSUER: "{{ EDXAPP_LMS_ISSUER }}"
   JWT_EXPIRATION: '{{ EDXAPP_JWT_EXPIRATION }}'
+  JWT_PRIVATE_SIGNING_KEY: null
```
jibsheet: Do you mean null or !!null?

clintonb: Question I should have asked 1-2 years ago: what's the difference?

jibsheet (approved):

playbooks/roles/edxapp/defaults/main.yml
clintonb: Updated.

PR #3834, opened 2017-04-25
splunk-s3-backup-play

nadeemshahzad:
@edx/devops Kindly review

### DIFF
--- a/playbooks/edx-east/splunk_config_backup.yml
+++ b/playbooks/edx-east/splunk_config_backup.yml
@@ -1,0 +1,19 @@
+- name: Backup splunk configurations
+  hosts: all
+  vars:
+    splunk_config_dir: "/opt/splunk/etc"
+    splunk_config_archive: "{{ date }}.tar.gz"
+    splunk_s3_backup_tempdir: "/var/tmp"
+    splunk_s3_bucket: "{{ SPLUNK_S3_BACKUPS_BUCKET }}"
+    date: "{{ lookup('pipe', 'date +%Y-%m-%d') }}"
+  tasks:
+    - name: archive splunk configuration dir
+      command: >
+        tar -cpzf  "{{ splunk_s3_backup_tempdir }}/{{ splunk_config_archive }}" {{ splunk_config_dir }}
+    - name: copy tarball to s3 bucket
+      command: >
+        aws s3 cp "{{ splunk_s3_backup_tempdir }}/{{ splunk_config_archive }}" s3://{{ splunk_s3_bucket }}
+    - name: cleanup backup file
+      file:
+        path: "{{ splunk_s3_backup_tempdir }}/{{ splunk_config_archive }}"
+        state: absent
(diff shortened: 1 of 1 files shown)


playbooks/edx-east/splunk_config_backup.yml
```diff
@@ -0,0 +1,19 @@
...
+  hosts: all
+  vars:
+    splunk_config_dir: "/opt/splunk/etc"
+    splunk_config_archive: "{{ date }}.tar.gz"
+    splunk_s3_backup_tempdir: "/var/tmp"
+    splunk_s3_bucket: "{{ SPLUNK_S3_BACKUPS_BUCKET }}"
+    date: "{{ lookup('pipe', 'date +%Y-%m-%d') }}"
+  tasks:
+    - name: archive splunk configuration dir
+      command: >
```
jibsheet: Is the > just because of the line length?

playbooks/edx-east/splunk_config_backup.yml
```diff
@@ -0,0 +1,19 @@
+- name: Backup splunk configurations
+  hosts: all
+  vars:
+    splunk_config_dir: "/opt/splunk/etc"
+    splunk_config_archive: "{{ date }}.tar.gz"
+    splunk_s3_backup_tempdir: "/var/tmp"
+    splunk_s3_bucket: "{{ SPLUNK_S3_BACKUPS_BUCKET }}"
+    date: "{{ lookup('pipe', 'date +%Y-%m-%d') }}"
```
jibsheet: Can we include more of the timestamp (minutes/hour should suffice)

Just thinking I would totally run this play to back up before making a config change, and that would overwrite the backup from earlier in the day.

That way I could have splunk-backup.2017-04-02T2048.tar.gz

playbooks/edx-east/splunk_config_backup.yml
```diff
@@ -0,0 +1,19 @@
+- name: Backup splunk configurations
+  hosts: all
+  vars:
+    splunk_config_dir: "/opt/splunk/etc"
+    splunk_config_archive: "{{ date }}.tar.gz"
```
jibsheet: Can you consume a prefix of some kind?  we run several splunk servers and should back all of them up (I know the ticket is concentrated on one in particular)

nadeemshahzad: @jibsheet Added prefix with splunk server host id

playbooks/edx-east/splunk_config_backup.yml
nadeemshahzad: @jibsheet Include (hour/minutes) in timestamp and make date persistent.

playbooks/edx-east/splunk_config_backup.yml
nadeemshahzad: @jibsheet It was for line length , but it is ok to remove >

jibsheet (approved):
Changes look good - I'll want to see how you pass in the args in the Jenkins DSL, but I think this is good.

PR #3836, opened 2017-04-26
Set issuer for IDAs that use OpenID Connect

clintonb:
This should have been set prior to https://github.com/edx/auth-backends/pull/9 being merged. We did not catch this problem until now because the library currently used for JWT validation, pyjwt, simply skips issuer validation if the issuer is set to None.

LEARNER-693


clintonb:
FYI @edx/learner @edx/devops @douglashall

rlucioni (approved):
Is this issuer already included in the tokens? I'm wondering if deploying this will present any challenges.

clintonb:
@rlucioni our ID tokens have always had an issuer. The IDAs have not validated that issuer for the last 14 months. This will not cause an issue for existing installations. It will resolve the issue we saw this morning with Credentials and the upgraded authentication backend.

schenedx (approved):
👍

clintonb:
Travis is timing out (probably due to Credentials static assets taking a while to compile): https://travis-ci.org/edx/configuration/jobs/226128960. This change is unrelated to static assets, so I am merging. I will work with @edx/devops to increase the timeout or find another resolution to the build problem.

clintonb:
The build error was due to the changing of the `social` library to `social_core`. This has been fixed on edx/credentials.

PR #3838, opened 2017-04-27
Add check to see if hive metastore is initialized.

brianhw:
As part of a recent Hive upgrade, an additional task was added to analytics_pipeline role to initialize the Hive schema using a call to schematool. However, this was not idempotent – it failed when it had already been run. 

This PR addresses that, by checking first whether Hive finds version information in its metastore.  See DE-109.

### DIFF
--- a/playbooks/roles/analytics_pipeline/tasks/main.yml
+++ b/playbooks/roles/analytics_pipeline/tasks/main.yml
@@ -230,9 +230,18 @@
     - install
     - install:configuration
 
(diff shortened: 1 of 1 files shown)


brianhw:
This has been successfully been built into a box.  And Hassan has also used the branch to run release/analyticstack/Vagrantfile to reprovision.  (That covers the initial case and the second-run case.)

feanil (approved):

tobz (approved):
LGTM!

PR #3843, opened 2017-04-30
Automate required vagrant plugin installation for any number of plugins.

UmanShahzad:
Configuration Pull Request
---

---

Don't think the checklist above is really relevant for this change, except the first point, so ignoring it unless someone says otherwise.

### Problem

Save new users some typing (`vagrant plugin install PLUGIN1`, `vagrant plugin install PLUGIN2`, ...).

### Solution

In case `vagrant` needs more plugins in the future, one can simply add them to the list in the method call `install_plugins` in `Vagrantfile` and automate installation for all of them (given '`y`').

Because `vagrant` doesn't seem to detect the additional installed plugins mid-run, one has to manually kill the current `vagrant` process and restart it with Ruby's `exec`.

Long story short, this should save future users some typing! Especially if more plugins become required.

### Future Work

- Plugin version locking.

@bradenmacdonald For recommending me to commit this feature here.


openedx-webhooks:
Thanks for the pull request, @UmanShahzad! I've created [OSPR-1757](https://openedx.atlassian.net/browse/OSPR-1757) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

feanil:
@nedbat I think you or @gsong would be the right people to review this.

vagrant/.vagrantplugins.rb
```diff
@@ -0,0 +1,33 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
```
nedbat: I would rather not have editor mode files in every source file. I think emacs and vim users probably have their editors configured to recognize .rb files as Ruby.

vagrant/.vagrantplugins.rb
```diff
@@ -0,0 +1,33 @@
+# -*- mode: ruby -*-
```
nedbat: Why name this as a hidden file?  Let's find a good place for a real file with a real name.

nedbat (commented):
@UmanShahzad have you sent in your contributor's agreement? Also, full disclosure: we are in the process of building out a Docker-based infrastructure.  We likely won't get rid of Vagrantfiles just yet, but they are definitely not our preferred technology for the longer-term.

I'll add a few comments now, but we do need to get that contributor's agreement. Thanks for the pull request! :)

gsong:
@nedbat: @UmanShahzad is good on the CLA. 👍

vagrant/.vagrantplugins.rb
UmanShahzad: I agree -- this is only what I originally saw a similar solution doing and thought it may be the preferred method. But looking at it again it could be better. Please consider the new change and let me know if it's more reasonable!

vagrant/.vagrantplugins.rb
UmanShahzad: Good point, didn't catch that, even though I'm using vim myself -_- Removed it.

UmanShahzad:
@nedbat Thanks for reviewing the PR! I've handed in the CLA as @gsong noted and made some changes according to your review.

I definitely understand your moving to Docker. I hope I can stay in touch to help on that if needed. Let me know if this change is worth merging.

vagrant/base/analyticstack/Vagrantfile
```diff
@@ -1,6 +1,8 @@
 Vagrant.require_version ">= 1.8.7"
-unless Vagrant.has_plugin?("vagrant-vbguest")
-  raise "Please install the vagrant-vbguest plugin by running `vagrant plugin install vagrant-vbguest`"
+
+unless ARGV.empty?
```
nedbat: This is probably a fine line, but I don't understand it: why check if ARGV is empty?  Also, unless the "require" is a problem, we could check it inside the install_plugins function, and avoid the duplicated if-statements.

vagrantplugins.rb
```diff
@@ -0,0 +1,34 @@
+def install_plugins(plugins)
```
nedbat: Let's put this in the vagrant directory.
Also, the configuration repo is inconsistent, but we've decided on 4-space indents, so this new file should follow that rule.  We have a .editorconfig file if you have the Vim plugin.

nedbat: (actually, new lines in the Vagrantfiles should also follow that 4-space indent. sorry to be picky...)

nedbat (changes requested):

vagrant/base/analyticstack/Vagrantfile
UmanShahzad: So the reason for checking if ARGV is empty is to exempt the normal, argument-less `vagrant` command from raising the plugin error or asking for input (i.e. if you just typed `vagrant` to see the subcommand list, with either the old version or this one, then if you don't have the plugin it throws the error; I assume that isn't intentional so we exempt the requirement check in this case).

Unfortunately commands like `vagrant help` are still not exempt with what's here, so I think your recommendation to encapsulate this logic inside of `install_plugins` may be a fine idea to keep it hidden and still cover all cases.

vagrantplugins.rb
UmanShahzad: No no, that's fine, I am picky myself :) Thanks for that, I'll follow it.

UmanShahzad:
@nedbat Updated -- let me know if you feel the ARGV situation should be handled differently.

nedbat:
Thanks, looks good!

nedbat (approved):

nedbat:
@UmanShahzad I had to revert this for now, a developer had an issue with the require_relative path.

nedbat:
@UmanShahzad ugh, I feel terrible about this.  The way we install devstack is with install_devstack.sh, which curls the Vagrantfile without the rest of the configuration repo.  It also installs the plugins for you.

UmanShahzad:
@nedbat No worries! Nothing terrible about it. I walked through `install_stack.sh` and saw where the plugin installation occurs. It's explicitly called (i.e. `vagrant plugin install vagrant-vbguest` written straight in the script). Automating it for further plugins is trivial:

```
for plugin in vagrant-vbguest vagrant-plugin2 vagrant-plugin3 ...
do
    vagrant plugin install $plugin
done
```

But considering only a single plugin's being used right now and you guys are moving to Docker, this is totally unnecessary.

Hopefully I can help contribute elsewhere next time :) Thanks for reviewing!

nedbat:
@UmanShahzad it's been a pleasure working with you.  I hope you can do more in the future. Thanks.

PR #3849, opened 2017-05-02
timeout crons

nadeemshahzad:
@edx/devops Please review

### DIFF
--- a/playbooks/roles/rabbitmq/tasks/main.yml
+++ b/playbooks/roles/rabbitmq/tasks/main.yml
@@ -118,7 +118,7 @@
 - name: Set up a cron job to run queue script
   cron:
     name: "log-queue-lenghts"
-    job: "{{ rabbitmq_app_dir }}/log-rabbitmq-queues.sh >/dev/null 2>&1"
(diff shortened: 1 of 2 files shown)


playbooks/roles/rabbitmq/tasks/main.yml
```diff
@@ -127,7 +127,7 @@
 - name: Set up a cron job to run the script
   cron:
     name: "log-rabbitmq-memory-usage"
-    job: "{{ rabbitmq_app_dir }}/log-rabbitmq-memory.sh >/dev/null 2>&1"
+    job: /usr/bin/timeout 5 {{ rabbitmq_app_dir }}/log-rabbitmq-memory.sh >/dev/null 2>&1
```
feanil: Does it make more sense to make the number a variable so we don't repeat it in both of these jobs and it's easier to change?

nadeemshahzad: @feanil Yes, this make change easier. I have changed number to var in play.

feanil (approved):

PR #3856, opened 2017-05-04
Mention the internal repos

jibsheet:
While reviewing @doctoryes' PR I realized we never tweaked this template.

@edx/devops


doctoryes (approved):
Nit: How many contributors understand what an internal value is versus a secure value? I wonder if a pointer to a further explanation is appropriate here? Perhaps this topic is covered elsewhere before the PR creation phase.

jibsheet:
Agreed - it's hard because it's mostly an edx.org convention.
We're working on the docs for *-internal vs *-secure repos, so I assume I'll come back and link to those later.
But currently we're fielding a number of support tickets that say "Please add this to the secure repo" and we reply "Please open a PR against the internal repo"

PR #3861, opened 2017-05-09
Updated discovery play to use edx_django_service play

clintonb:
- This removes the duplication across the various IDA configurations
- Updated Dockerfile for Docker-based devstack

LEARNER-817


clintonb:
@edx/devops please review
FYI @edx/docker-devstack-working-group @edx/learner 

This will be the first of three PRs to update credentials, discovery, and ecommerce to use the `edx_django_service` role.

clintonb:
Our Docker devstack images have been running this configuration for the past five months.

feanil:
Have you built AMIs from this?

clintonb:
loadtest deployments are failing due to https://github.com/edx/configuration/blob/f8cc6369a8e1b556fb92b9ad4960bb252daa5eb4/playbooks/roles/discovery/defaults/main.yml#L127-L133. I'm open to ideas for workarounds. (I'm working on another story at the moment, so I can't dedicate any attention to this this week.)

jibsheet:
Ah, the discovery.yml playbook in edx-east overrides that.
This looks like an oversight in where the default is declared vs some documentation.

I'm on a bunch of other things this week, but I probably know more than @e0d the original author about current usage so can help later.

clintonb:
I'll ping you next week, @jibsheet. Thanks!

playbooks/edx-east/discovery.yml
```diff
@@ -9,8 +9,6 @@
     CLUSTER_NAME: 'discovery'
   roles:
     - aws
-    - role: automated
```
clintonb, line 12: @e0d @jibsheet please advise on the need for this role. Only discovery and edxapp are using it. Is the intention to use this for all IDAs? If so, it might make sense to add it to `edx_django_service`. For now I have removed it to see if there are other issues with deploying to loadtest-edx-discovery.

jibsheet: If you want to run management commands from jenkins, you need it.
Otherwise you don't and we'll add it later.
I don't think it hurts to have it if none of the VARS exist, also might help people if the var scaffolding exists.

clintonb: This role is now part of `edx_django_service`.

playbooks/edx-east/discovery.yml
```diff
@@ -9,8 +9,6 @@
...
-      AUTOMATED_USERS: "{{ DISCOVERY_AUTOMATED_USERS | default({}) }}"
```
feanil, line 13: You removed this line but did not update meta/main.yml to pass the `DISCOVERY_AUTOMATED_USERS` to the `edx_django_service` role.

feanil (changes requested):

playbooks/edx-east/discovery.yml
clintonb: @feanil, does it actually need to be overridden, or are the defaults in `playbooks/roles/edx_django_service/defaults/main.yml` sufficient for the long-term?

feanil: The defaults in `playbooks/roles/edx_django_service/defaults/main.yml` is just an example.  DISCOVERY_AUTOMATED_USERS needs to exist as a variable because that is what we override in our internal config to define all the various commands that the automated user is allowed to run.

clintonb: Restored. In the future should we consider moving the overrides in edx-internal to configuration? What about the override makes it edX-specific/private?

clintonb:
Related PRs
* Add OAuth URLs to edx-internal: https://github.com/edx/edx-internal/pull/95
* Update pipelines to use loadtest-<play> branch for edx-internal: https://github.com/edx/edx-gomatic/pull/341

clintonb:
@feanil ready for final review
FYI @edx/learner

playbooks/edx-east/discovery.yml
jibsheet: We should be able to define SERVICE_AUTOMATED_USERS as an empty dictionary and override in internal.  The dictionary contains public ssh keys and commands we run in production, so is not good for public config.

feanil (approved):

clintonb:
I will merge/deploy this on Monday since I don't have time/attention to dedicate to debugging if anything goes wrong...not that anything will go wrong.

PR #3862, opened 2017-05-10
We found once a minute too frequent for these for splunk

jibsheet:


jibsheet:
As a note - I'd really like us to stop using cron without cron_file http://docs.ansible.com/ansible/cron_module.html 

It means it ends up in root's crontab -l rather than somewhere you might expect, like cron.d

feanil (approved):

PR #3864, opened 2017-05-10
Parameterize Dockerfiles for Open edX releases

nedbat:
Configuration Pull Request
---

Add some parameterization to the Dockerfiles so we can make images for Open edX releases. (Much more is needed to complete that work, this is a start).

This provides for choosing a different branch other than master ($OPENEDX_RELEASE), and for adding a component to the image names ($IMAGE_PREFIX).

The IMAGE_PREFIX would be empty for master builds, so the images will be named as they are today: "edxops/edxapp:latest".  For Ficus, IMAGE_PREFIX would be "ficus/", so the image would be "edxops/ficus/edxapp:latest".

These files use a feature newly available in Docker 17.05: the use of ARGS in the FROM line.

### DIFF
--- a/docker/build/edxapp/Dockerfile
+++ b/docker/build/edxapp/Dockerfile
@@ -8,7 +8,8 @@
 # with the currently checked-out configuration repo.
 
 
-FROM edxops/xenial-common:latest
+ARG IMAGE_PREFIX
+FROM edxops/${IMAGE_PREFIX}xenial-common:latest
 MAINTAINER edxops
 USER root
 CMD ["/edx/app/supervisor/venvs/supervisor/bin/supervisord", "-n", "--configuration", "/edx/app/supervisor/supervisord.conf"]
@@ -18,9 +19,12 @@ WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
 
 COPY docker/build/edxapp/ansible_overrides.yml /
 
+ARG OPENEDX_RELEASE=master
+ENV OPENEDX_RELEASE=${OPENEDX_RELEASE}
 RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook edxapp.yml \
     -c local -i '127.0.0.1,' \
     -t 'install,assets,devstack' \
+    --extra-vars=edx_platform_version=${OPENEDX_RELEASE} \
     --extra-vars="@/ansible_overrides.yml"
 
 EXPOSE 18000 18010
--- a/docker/build/xenial-common/Dockerfile
+++ b/docker/build/xenial-common/Dockerfile
@@ -1,5 +1,5 @@
 FROM ubuntu:xenial
-MAINTAINER edxops 
+MAINTAINER edxops
 
 # Set locale to UTF-8 which is not the default for docker.
 # See the links for details:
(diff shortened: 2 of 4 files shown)


nedbat:
@clintonb @gsong @feanil opinions?

docker/build/edxapp/Dockerfile
```diff
@@ -8,7 +8,8 @@
 # with the currently checked-out configuration repo.
 
 
-FROM edxops/xenial-common:latest
+ARG IMAGE_PREFIX
```
feanil, line 11: Is the default for this just an empty string?  What are other valid values?

nedbat: Yes, it defaults to empty string.  I'm not sure the best way to indicate that.  It works now if not provided.

feanil: Let's add a comment just so that clear.  Also is there anything we can say about how to figure out what are valid values for this arg?  It's fine if there isn't  but would be useful to put something here if there is.

clintonb: What are some example values?

clintonb (commented):
This seems right. I'd like to better understand the naming scheme.

docker/build/edxapp/Dockerfile
gsong: @clintonb I believe the idea is to have Open edX specific prefixes, e.g. `ficus-`.

docker/build/xenial-common/Dockerfile
```diff
@@ -1,5 +1,5 @@
 FROM ubuntu:xenial
-MAINTAINER edxops 
+MAINTAINER edxops
```
gsong, line 2: `MAINTAINER` instruction is deprecated: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated

docker/build/xenial-common/Dockerfile
```diff
@@ -10,7 +10,8 @@ ENV LANG C.UTF-8
 
 ENV ANSIBLE_REPO="https://github.com/edx/ansible"
 ENV CONFIGURATION_REPO="https://github.com/edx/configuration.git"
-ENV CONFIGURATION_VERSION="master"
+ARG OPENEDX_RELEASE=master
+ENV CONFIGURATION_VERSION="$OPENEDX_RELEASE"
```
gsong: Can we standardize on `${var}` syntax?

gsong:
@clintonb @feanil It's also worth noting that the parameterized `FROM` instruction is a feature of edge channel Docker CE.

docker/build/edxapp/Dockerfile
nedbat: I should have explained more, sorry.  Master would make "edxops/edxapp:latest", and Ficus would make "edxops/ficus/edxapp:latest".  $IMAGE_PREFIX would be "ficus/"

nedbat: I've added the description to the top of the PR.

feanil: @nedbat any reason not to put that info in this file or a README under the `docker/build` directory?

nedbat: Great idea.

docker/build/xenial-common/Dockerfile
nedbat: I thought the maintainer was supposed to be an email address, but I can't find where I read that, and I don't know what email we would use.  I'll leave this to a following PR.

docker/build/edxapp/Dockerfile
nedbat: I've added details to the docker/README.md (i'll convert to .rst in a follow-on).

nedbat (commented):
I also used ${} everywhere.

feanil (approved):

clintonb (approved):

PR #3871, opened 2017-05-16
Install docker on tools jenkins.

feanil:
Configuration Pull Request
---

@edx/devops

### DIFF
--- a/playbooks/roles/tools_jenkins/defaults/main.yml
+++ b/playbooks/roles/tools_jenkins/defaults/main.yml
@@ -77,3 +77,5 @@ jenkins_tools_debian_pkgs:
     - python-pycurl
     - psmisc
     - mysql-client-core-5.6
+
+jenkins_tools_docker_user: 'jenkins-docker-proxy'
--- a/playbooks/roles/tools_jenkins/defaults/main.yml
+++ b/playbooks/roles/tools_jenkins/defaults/main.yml
@@ -78,4 +78,4 @@ jenkins_tools_debian_pkgs:
     - psmisc
     - mysql-client-core-5.6
 
-jenkins_tools_docker_user: 'jenkins-docker-proxy'
+JENKINS_TOOLS_DOCKER_USER: 'jenkins-docker-proxy'
--- a/playbooks/roles/tools_jenkins/templates/99-tools-jenkins.j2
+++ b/playbooks/roles/tools_jenkins/templates/99-tools-jenkins.j2
@@ -1,0 +1,1 @@
+{{ jenkins_user }} ALL=({{ JENKINS_TOOLS_DOCKER_USER }}) SETENV:NOPASSWD:/usr/bin/docker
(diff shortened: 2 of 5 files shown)


jibsheet (approved):
Do we have any concerns having jenkins itself with docker socket access?  Should we make it a user and su / sudo before running docker (from a security standpoint).

feanil:
Would we then use the automator role to allow it to only sudo specific commands?

The docker image builds alone would be a lot of commands to allow, let's talk in parking lot tomorrow to figure out how this might work.

jibsheet:
I wasn't thinking of the automator role although I guess we could have jenkins ssh back to itself, but that seems like a lot.

I was thinking that the job would just sudo su -s /bin/bash dockeruser and that way just random groovy or Java in Jenkins wouldn't have direct access to the socket?  Jenkins would still be able to sudo of course, and you could get there from here, but it at least disallows writing to it directly and issuing commands directly.

feanil:
Spoke with @jibsheet offline and we agreed that proxying the commands through another user gives you a extra layer of security without too much cost.

**TODO**: The jenkins user needs to be able to sudo as the jenkins-docker-proxy.  So the tools_jenkins role will need to drop a sudoers file in sudoers.d to allow this.

playbooks/roles/tools_jenkins/defaults/main.yml
```diff
@@ -66,3 +66,6 @@ jenkins_tools_bundled_plugins:
   - "pam-auth"
   - "ssh-credentials"
   - "ssh-slaves"
+
+jenkins_tools_docker_user: 'jenkins-docker-proxy'
+tools_jenkins_sudoers_template: "roles/tools_jenkins/templates/99-tools-jenkins.j2"
```
jibsheet: I thought you could just say 99-tools-jenkins.j2 and the template module will default to the current role's templates/ directory.  I also don't know if this needs to be a var, but I'm not too worried about that.

playbooks/roles/tools_jenkins/defaults/main.yml
```diff
@@ -66,3 +66,6 @@ jenkins_tools_bundled_plugins:
   - "pam-auth"
   - "ssh-credentials"
   - "ssh-slaves"
+
+jenkins_tools_docker_user: 'jenkins-docker-proxy'
```
jibsheet: I think this should be an upper case var because it's something we expect you to override, not a role internal variable.

playbooks/roles/tools_jenkins/templates/99-tools-jenkins.j2
```diff
@@ -0,0 +1 @@
+{{ jenkins_user }} ALL=({{ jenkins_tools_docker_user }}) SETENV:NOPASSWD:/usr/bin/docker
```
jibsheet: I assume you didn't need wildcarding for this to work?

jibsheet (approved):
Some small style suggestions and a question, but looks good.

michaelroytman:
@kevin If you're interested, re: wildcarding:

screen shot 2017-06-22 at 3 32 14 pm

jibsheet:
Shipit - I'll reread man 5 sudoers next time I have insomnia

PR #3876, opened 2017-05-19
Added PhantomJS to Jenkins workers

clintonb:
LEARNER-509

### DIFF
--- a/playbooks/roles/jenkins_worker/defaults/main.yml
+++ b/playbooks/roles/jenkins_worker/defaults/main.yml
@@ -11,6 +11,7 @@ jenkins_debian_pkgs:
     - libffi-dev
     - python-dev
     - libsqlite3-dev
+    - phantomjs
 
 # packer direct download URL
 packer_url: "https://releases.hashicorp.com/packer/0.8.6/packer_0.8.6_linux_amd64.zip"


clintonb:
@edx/devops @edx/testeng please review. What is involved in getting this deployed?

estute:
@clintonb we already have phantomJS installed on the Jenkins workers (via the browsers role - https://github.com/edx/configuration/blob/master/playbooks/roles/browsers/defaults/main.yml). However, we are planning on getting rid of it

clintonb:
@estute Why do you want to get rid of it?

Question for @edx/devops: any opposition to my adding Firefox and PhantomJS to the tools_jenkins Docker image? In order to actually run my tests I either need to do that, or run a worker (for which we do not currently have a Docker image).

estute:
@clintonb because PhantomJS either unmaintained or will be soon, and just creates more tools spread. We currently only use it for a11y testing, and plan to do that with FF (or eventually headless chrome)

clintonb:
This has been a bit painful. Does no one test their DSL locally? If so, how do they test e2e tests without a browser?

estute:
@clintonb Ah, sorry if that was misleading. The only use of Phantom JS in the a11y tests. We use Chrome and FF for all other testing. The e2e tests are run with FF

jibsheet:
We run the browsers role on tools_jenkins for the ora2 tests, so you have what they have.

PR #3877, opened 2017-05-22
Added yarn to IDAs

clintonb:
The credentials, discovery, ecommerce, and edx_django_service roles now include yarn as a replacement for npm.

### DIFF
--- a/playbooks/roles/yarn/tasks/main.yml
+++ b/playbooks/roles/yarn/tasks/main.yml
@@ -1,0 +1,41 @@
+---
+# Configures Yarn package sources, and installs the package.
+# Adapted from https://github.com/fubarhouse/ansible-role-yarn.
+
+- name: "Yarn | GPG"
+  become: yes
+  become_user: root
+  apt_key:
+    url: https://dl.yarnpkg.com/debian/pubkey.gpg
+    state: present
+
+- name: "Yarn | Ensure Debian sources list file exists"
+  become: yes
+  become_user: root
+  file:
+    path: /etc/apt/sources.list.d/yarn.list
+    owner: root
+    mode: 0644
+    state: touch
+
+- name: "Yarn | Ensure Debian package is in sources list"
+  become: yes
+  become_user: root
+  lineinfile:
+    dest: /etc/apt/sources.list.d/yarn.list
+    regexp: 'deb http://dl.yarnpkg.com/debian/ stable main'
+    line: 'deb http://dl.yarnpkg.com/debian/ stable main'
+    state: present
+
+- name: "Yarn | Update APT cache"
+  become: yes
+  become_user: root
+  apt:
+    update_cache: yes
+
+- name: "Yarn | Install"
+  become: yes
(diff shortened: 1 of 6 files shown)


clintonb:
Currently, only the Credentials Service uses yarn: https://github.com/edx/credentials/pull/243.

clintonb:
FYI @edx/learner 
@edx/devops please review/advise on the new role.

rlucioni (approved):
Looks like you'll have to rebuild this once https://github.com/edx/credentials/pull/246 lands to get a clean build.

playbooks/roles/yarn/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
+---
+# Configures Yarn package sources, and installs the package.
+# Adapted from https://github.com/fubarhouse/ansible-role-yarn.
+
+- name: "Yarn | GPG"
+  become: yes
```
feanil, line 6: We generally use `true` instead of `yes`

playbooks/roles/yarn/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  become_user: root
```
feanil, line 7: You don't need this.  By default `become: true` will become root.

playbooks/roles/yarn/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  apt_key:
+    url: https://dl.yarnpkg.com/debian/pubkey.gpg
+    state: present
+
+- name: "Yarn | Ensure Debian sources list file exists"
```
feanil, line 12: Use `apt_repository` instead?

http://docs.ansible.com/ansible/apt_repository_module.html

playbooks/roles/yarn/tasks/main.yml
```diff
@@ -0,0 +1,41 @@
...
+  become: yes
+  become_user: root
+  apt:
+    update_cache: yes
+
+- name: "Yarn | Install"
+  become: yes
+  become_user: root
+  package:
+    name: yarn
```
feanil, line 40: Pull this out into `defaults/main.yml` as a list that's passed into here.  That's a general pattern we use and helps keep things flexible.

feanil (changes requested):

thallada:
We are interested in using this in Insights too!

clintonb:
@thallada I'm holding off on this since Node 8.x and npm 5 incorporate similar changes. My preference is to not include an extra dependency if one we already have achieves the same goals. If you want to move forward with `yarn`, feel free to take over this branch.

thallada:
I didn't realize npm 5 had all those features. I'll upgrade to that instead, thanks.

feanil:
@clintonb please close and delete the branch if we're done with it.

clintonb:
Note to future viewers: we will use Node 8.x, once an LTS is released, and npm 5 which implements similar features.

PR #3899, opened 2017-06-07
add a couple of a11y tests to the packer smoke test script

estute:
@jzoldak @arizzitano 
I added 2 a11y tests to the smoke test script that gets run whenever we build a new jenkins worker. It was missing before


estute:
Build can be seen here: https://build.testeng.edx.org/job/build-packer-ami/6209

estute:
Updated build https://build.testeng.edx.org/view/All/job/build-packer-ami/6269

PR #3900, opened 2017-06-07
Skip elb operations in CreateSandbox job with full_edx_installation

nadeemshahzad:
@edx/devops Review please

PR #3902, opened 2017-06-07
PLAT-1228 Support bok-choy in edxapp Docker devstack

jmbowman:
Making the following changes to enable running bok-choy tests in devstack Docker images:

* Install the mongo and mysql clients required for test setup (*only* for devstack Docker images)
* Set some environment variables to point selenium at the correct separate container running a selenium standalone server with Firefox
* Refactor the Debian package list and environment variable settings to be cleanly overridden in order to accomplish the goals above

In the long term, we probably want to refactor the test setup in order to avoid needing the database clients in the base edx-platform container (they add almost 250 MB to its size).  But the immediate goal is to just get it working.

---

  - [x] Are you adding any new default values that need to be overridden when this change goes live? - No

### DIFF
--- a/docker/build/edxapp/ansible_overrides.yml
+++ b/docker/build/edxapp/ansible_overrides.yml
@@ -26,3 +26,12 @@ EDXAPP_INSTALL_PRIVATE_REQUIREMENTS: true
 EDXAPP_SEARCH_HOST: 'edx.devstack.elasticsearch'
 
 EDXAPP_PYTHON_SANDBOX: false
+
+edxapp_debian_pkgs_extra:
+  - mongodb-clients
+  - mysql-client
+
+edxapp_environment_extra:
+  SELENIUM_BROWSER: 'firefox'
+  SELENIUM_HOST: 'edx.devstack.firefox'
+  SELENIUM_PORT: '4444'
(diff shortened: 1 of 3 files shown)


jmbowman:
I attempted to create a sandbox to test this, but it failed with the error that https://github.com/edx/configuration/pull/3900 is intended to fix.  I'll retry once that's merged.

docker/build/edxapp/ansible_overrides.yml
```diff
@@ -26,3 +26,12 @@ EDXAPP_INSTALL_PRIVATE_REQUIREMENTS: true
 EDXAPP_SEARCH_HOST: 'edx.devstack.elasticsearch'
 
 EDXAPP_PYTHON_SANDBOX: false
+
+edxapp_debian_pkgs_extra:
+  - mongodb-clients
```
jibsheet, line 31: We'll want to fix this later to be not-2.6

jibsheet (approved):
Looks good, if you skip xqueue you shouldn't see an ELB error on sandboxes, but feel free to link to it if you're seeing it more.

Your travis test looks network fail related, but I didn't look deeply.

Please make sure the sandbox does build before we merge.

jmbowman:
Successfully created the sandbox and ran through the checklist testing steps for it.  Should I continue on with the rest of the checklist?  It looks like it involves building another 2 sandboxes and 2 AMIs.

jibsheet:
I would merge since it's effectively a NOOP outside of the docker environment.
We'll let you know if it causes pipeline problems (you can build the non-scratch AMI if you want to be extra cautious).

PR #3905, opened 2017-06-09
New base AMI

jibsheet:
DEVOPS-6012 which drops the django celery tables
Newer clean 16.04 AMI so that it'll get newer kernel modules.

FYI @efischer19 

@edx/devops  please review


jibsheet:
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/1242/console
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/1240/console

feanil (approved):

PR #3895, opened 2017-06-06
Remove local version of ansible modules taken from ansible-core

nadeemshahzad:
@edx/devops Kindly review

PR #3907, opened 2017-06-09
Revert "Remove local version of ansible modules taken from ansible-core"

macdiesel:
Reverts edx/configuration#3895

Before we can do this we need to upgrade to boto 3. This is currently breaking builds:
https://gocd.tools.edx.org/go/tab/build/detail/stage-insights/31/build_ami/1/stage_edx

### DIFF
--- a/playbooks/continuous_delivery/cleanup.yml
+++ b/playbooks/continuous_delivery/cleanup.yml
@@ -49,7 +49,7 @@
       state: absent
 
   - name: Send Hipchat notification cleanup has finished
-    hipchat:
+    hipchat_2_0_0_1:
       api: "{{ hipchat_url }}"
       token: "{{ hipchat_token }}"
       room: "{{ hipchat_room }}"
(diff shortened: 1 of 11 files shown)


doctoryes:
@macdiesel How did you figure out that boto3 is a prerequisite?

macdiesel:
I ran in to this a few months back when I tried to remove these items myself. There is a ticket on the delivery backlog to do this work.

doctoryes (approved):
👍  Could you please link to the referred ticket?

macdiesel:
Here are the related items I was able to find, I think I dropped it at the time in favor of other feature work:
https://openedx.atlassian.net/browse/DELIVERY-2
https://github.com/edx/configuration/pull/3559
https://github.com/edx/configuration/pull/3566

I remember my brief investigation that there was a defect in our version of ansible when using boto 2. creationDate is an attribute of image that is only available in boto 3. Therefore if we want to remove our custom create_ami we need to update to boto 3.  There could be other issues as well related to some of these modules that may use boto and shouldn't update until that update is complete.

PR #3919, opened 2017-06-15
increase sshd security

fredsmith:
Suggestions from https://stribika.github.io/2015/01/04/secure-secure-shell.html

### DIFF
--- a/playbooks/roles/vhost/tasks/main.yml
+++ b/playbooks/roles/vhost/tasks/main.yml
@@ -76,7 +76,7 @@
     line: "LogLevel VERBOSE"
   register: sshd_config_line1
 
-- name: Update sshd logging to VERBOSE
+- name: Update sshd password authentication
   lineinfile:
     dest: /etc/ssh/sshd_config
     state: present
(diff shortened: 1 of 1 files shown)


jibsheet:
Can you confirm that this results in a machine reachable from os-x?  I know there were some issues around keys and ciphers in Sierra (mostly around retiring DSA, which we're also doing) but older os-x releases may have an old ssh client.  I'm not sure what our fleet runs, Wes might know.

I assume we can test by spinning up a sandbox with this branch pretty easily.

fredsmith:
My concern about retiring the keys is that it causes the ssh host key to change, so people who have strict host key checking on will get a possibly unexpected warning after running this.  The rest of the changes are invisible to the client.

I can spin up a sandbox for testing.

jibsheet:
Luckily, for most users, that's tools-edx-gp.edx.org and we can communicate it.

fredsmith:
@jibsheet if you have the ability to test from the systems you have concerns about, fredsmith.sandbox.edx.org is up with this branch.   https://sshcheck.com/server/fredsmith.sandbox.edx.org/

playbooks/roles/vhost/tasks/main.yml
```diff
@@ -76,21 +76,82 @@
...
     state: present
     regexp: "^PasswordAuthentication .*"
     line: "PasswordAuthentication {{ COMMON_SSH_PASSWORD_AUTH }}"
   register: sshd_config_line2
 
+- name: secure ssh algorithms
+  lineinfile:
+    name: /etc/ssh/sshd_config
+    regexp: '^KexAlgorithms'
+    insertafter: '^ServerKeyBits'
```
smarnach, line 91: `ServerKeyBits` is for SSH protocol version 1, and not supported anymore in the OpenSSH version included in Ubuntu 17.04, so I suggest inserting the additional lines in a different place (or simply at the end of the file).

smarnach (commented):
@fredsmith Thanks for pinging me.  I had a quick look at the PR, but I haven't yet verified whether this will cause any problems for us.  I will try to do so by Monday.

jibsheet:
Should we be concerned about fixing 
```
ssh-rsa	RSA with SHA-1 hash 
```
while we're here?

I'll poll developers for old os-x (I have sierra and it's fine with that).

smarnach:
I tested this, and this is unlikely to cause any problems for us, so all good from our side.

I'm not a cryptography expert, so it would take me a significant amount of time to judge the quality of the recommendations on https://stribika.github.io/2015/01/04/secure-secure-shell.html.  In cases like these, when I'm not really competent to judge something myself, I try to judge the party that is trying to sell it to me.  However, I couldn't find out anything about the author of the guide you linked, not even their real name.

How about following the recommendations on https://wiki.mozilla.org/Security/Guidelines/OpenSSH instead?  The Mozilla recommendations come from a reputed security team and seem more up to date (though they might be too restrictive, not sure).

playbooks/roles/vhost/tasks/main.yml
myvcc657: Perbaiki

myvcc657:
`reslove all erroe file`

PR #3926, opened 2017-06-19
ENT-433: Provision a default admin user for developer environments

shrlyhe:
Configuration Pull Request
---

### DIFF
--- a/playbooks/edx-east/manage_edxapp_users_and_groups.yml
+++ b/playbooks/edx-east/manage_edxapp_users_and_groups.yml
@@ -10,7 +10,7 @@
 #
 # Overview:
 # This playbook ensures that the specified users and groups exist in the targeted
-# edxapp cluster. 
+# edxapp cluster.
 #
 # Users have the following properties:
 #   - username (required, str)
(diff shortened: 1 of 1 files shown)


fredsmith:
Hi,  This isn't the right location for this change. 


  -  It looks like you want to set a default user account for sandbox systems?   This ansible playbook isn't run on sandboxes, as far as I know.  It is only run on our production systems and we don't want to add this user there.  
  - Your syntax is also incorrect - ansible playbooks take variables in the 'vars' section of the playbook, not in the root of the playbook.  See below where python_path, etc. are defined.  
  - By convention, we don't set default variables in playbooks, we set them in roles. See where vars_files pulls in `roles/common_vars/defaults/main.yml`    manage_edxapp_users_and_groups is a unique playbook, too.  most of our playbooks use environment specific overrides, not inline static includes like this one does.

I think you should update the sandbox playbook to run the same tasks as this playbook, and add the users you want created to https://github.com/edx/sandbox-internal/blob/master/ansible/vars/developer-sandbox.yml Then make the tasks you copy from this playbook take the users out of a named dictionary.

jibsheet:
The demo role sets other users up and seems like the right place to do this, especially if we want them in all environments.  But there's a ticket about unbundling demo and democourses that may impact this work.

shrlyhe:
@fredsmith Thanks for the comment! I'm not sure if I have access to https://github.com/edx/sandbox-internal/blob/master/ansible/vars/developer-sandbox.yml, I'm getting a 404 error.

PR #3927, opened 2017-06-19
Corrected OAuth logout URL default

clintonb:
PLAT-1573


playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -211,7 +211,7 @@ COMMON_EXTRA_CONFIGURATION_SOURCES: []
 
 COMMON_OAUTH_PUBLIC_URL_ROOT: 'http://127.0.0.1:8000/oauth2'
 COMMON_OAUTH_URL_ROOT: '{{ COMMON_OAUTH_PUBLIC_URL_ROOT }}'
-COMMON_OAUTH_LOGOUT_URL: '{{ COMMON_OAUTH_PUBLIC_URL_ROOT }}/logout'
+COMMON_OAUTH_LOGOUT_URL: 'http://127.0.0.1:8000/logout'
```
jibsheet: Shouldn't http://127.0.0.1:8000 be in a var that then gets interpolated into PUBLIC_URL_ROOT and LOGOUT_URL?  Assuming we could then have only 1 to set when we turn it into courses.edx.org.

clintonb: Sure. Suggestions for the new variable name?

clintonb:
Related: https://github.com/edx/edge-internal/pull/28

playbooks/roles/common_vars/defaults/main.yml
feanil: `COMMON_OAUTH_BASE_URL`?

clintonb:
More changes to edx-internal and edge-internal will be made once this is incorporated. This is fine to merge for now.

jibsheet (approved):

PR #3932, opened 2017-06-21
New sandbox base image

jibsheet:
DEVOPS-6079


jibsheet:
@edx/devops Please review, resolves the djcelery tables that were re-added when we cut the ES 1.5 base box.

feanil (approved):

PR #3938, opened 2017-06-22
 script to find asgs and active instances for clusters

hchen15:
Configuration Pull Request
---


util/cluster_instance_monitoring.py
```diff
@@ -0,0 +1,91 @@
+import boto3
+import argparse
+import sys
+import yaml
+import pprint
```
feanil: You can just do `from pprint import pprint` and then you don't have to instantiate your own pretty printer.

util/cluster_instance_monitoring.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+    Determines if a given cluster has at least one ASG and at least one active instance.
+
+    Input: 
+    cluster_file: a yaml file containing a dictionary of triples that specify the particular cluster to monitor.
+    The keys of each entry in the dictionary are 'env', 'deployment', and 'cluster', specifying the environment, deployment,
+        and cluster to find ASG's and active instances for. 
+
+    """
+    pp = pprint.PrettyPrinter()
```
feanil: See above comment.  You should be able to delete this line.

util/cluster_instance_monitoring.py
```diff
@@ -0,0 +1,91 @@
...
+
+    f = open(cluster_file)
```
feanil: Use a context manager instead.  See the last paragraph of [this section](https://docs.python.org/2/tutorial/inputoutput.html#methods-of-file-objects) if you haven't used them before.

util/cluster_instance_monitoring.py
```diff
@@ -0,0 +1,91 @@
...
+    cluster_map = yaml.safe_load(f)
+    f.close()
+
+    region = 'us-east-1'
```
feanil: Let's make this a configurable option that can be passed in from the command-line.

util/cluster_instance_monitoring.py
```diff
@@ -0,0 +1,91 @@
...
+        
+        if not cluster_asgs:
+            not_matching_triples += [triple]
+        else:
+            all_matching_asgs += cluster_asgs
+
+    #The ASG's for the triples that have no active instances
+    no_active_instances_asgs = []
+
+    #check to make sure each ASG has at least 1 running instance
```
feanil: You don't need to make sure that each ASG has a instance in it, just that one ASG for a triple has an instance in it.  The case can occur where an ASG doesn't have instances in it because it is currently being terminated.  We don't care about this if there is a second ASG for the same triple that can support traffic.

util/cluster_monitoring_requirements.txt
```diff
@@ -0,0 +1 @@
+boto3==1.4.4
```
feanil: Any reason that this can't be added to the toplevel requirements file?  I would just add it there instead of making a new file if it's just this.

feanil (changes requested):

jibsheet:
git rm the requirements file you added, it's still here but empty.

feanil (approved):

PR #3940, opened 2017-06-22
Who supervises the supervisor?

efischer19:
InstructorTasks (celery tasks) are dying before they finish. From what I can determine, what's happening is:

- systemd tells supervisor to stop (SIGTERM)
- supervisor helpfully relays that through to the processes it supervises, which are the python processes executing our celery tasks.
  - at this point, no further tasks will be started, but supervisor will allow running tasks to finish
  - long running tasks (grade reports, bulk email) continue to execute, preventing their process from going down
- after a period of time (~90 seconds?), systemd decides that supervisor has had more than enough time and terminates forcefully via SIGKILL
  - this signal is then forwarded through to supervisor's children (the python processes), and they die too.

[Here's a splunk search showing the above sequence playing out](https://splunk.edx.org/en-US/app/search/search?sid=1498156537.1897804)

The fix is to set `TimeoutStopSec` to the same 5 day period we've defined for `TimeoutStartSec`, which can both be defined via `TimeoutSec`

```
TimeoutStartSec=
Configures the time to wait for start-up. If a daemon service does not signal start-up completion within the configured time, the service will be considered failed and will be shut down again. Takes a unit-less value in seconds, or a time span value such as "5min 20s". Pass "infinity" to disable the timeout logic. Defaults to DefaultTimeoutStartSec= from the manager configuration file, except when Type=oneshot is used, in which case the timeout is disabled by default (see systemd-system.conf(5)).

TimeoutStopSec=
Configures the time to wait for stop. If a service is asked to stop, but does not terminate in the specified time, it will be terminated forcibly via SIGTERM, and after another timeout of equal duration with SIGKILL (see KillMode= in systemd.kill(5)). Takes a unit-less value in seconds, or a time span value such as "5min 20s". Pass "infinity" to disable the timeout logic. Defaults to DefaultTimeoutStopSec= from the manager configuration file (see systemd-system.conf(5)).

TimeoutSec=
A shorthand for configuring both TimeoutStartSec= and TimeoutStopSec= to the specified value.
```
https://www.freedesktop.org/software/systemd/man/systemd.service.html

---

### DIFF
--- a/playbooks/roles/supervisor/templates/etc/init/supervisor-systemd.service.j2
+++ b/playbooks/roles/supervisor/templates/etc/init/supervisor-systemd.service.j2
@@ -28,12 +28,17 @@ PermissionsStartOnly=true
 User={{ supervisor_service_user }}
 
 Type=forking
-TimeoutStartSec=432000
+TimeoutSec=432000
 
 ExecStart={{ supervisor_venv_dir }}/bin/supervisord --configuration {{ supervisor_cfg }}
 ExecReload={{ supervisor_venv_dir }}/bin/supervisorctl reload
 ExecStop={{ supervisor_venv_dir }}/bin/supervisorctl shutdown
 
+# Trust supervisor to kill all its children
+# Otherwise systemd will see that ExecStop ^ comes back synchronously and say "Oh, I can kill everyone in this cgroup"
+# https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStop=
+# https://www.freedesktop.org/software/systemd/man/systemd.kill.html
+KillMode=none
 
 [Install]
 WantedBy=multi-user.target


efischer19:
FYI @feanil @jibsheet - I'm going to try to simulate this ~~in a sandbox~~on loadtest to confirm the issue and its fix, will update here when I have it

Edit: sandboxes aren't doing the thing for whatever reason (tasks are being hard-killed immediately, not after 90 seconds as I expected to see), will repro in loadtest w/ Kevin tomorrow morning

efischer19:
Updated "what's happening": Systemd is calling supervisor.service's defined `ExecStop`, then helpfully decides that "all processes remaining for a service are terminated according to the KillMode= setting", [per the docs](https://www.freedesktop.org/software/systemd/man/systemd.service.html#ExecStop=). Further, `KillMode` defaults to `control-group`, which ensure that "all remaining processes in the control group of this unit will be killed on unit stop", [per those docs](https://www.freedesktop.org/software/systemd/man/systemd.kill.html).

So the initial ExecStop command would send a SIGTERM to all the celery processes, then systemd would send them *another* one. As a further source of confusion, celery will try to register a new SIGTERM handler after the first signal, so the two signals firing in rapid succession led to a race condition a non-trivial amount of times. This means that sometimes the second signal would kill the process, other times we fell back to the default 90-second timeout mentioned in this PR's initial comment.

So the net result of this PR merging is twofold - systemd will stop sending SIGTERM to celery processes twice during this play, and the timeout before sending a final SIGKILL has increased from the default 90 seconds to 5 days.

jibsheet (approved):

jibsheet:
@edx/devops I've approved this, since most of the actual code changes are Eric's, but given the sensitivity of the shutdown process, I'd like more eyes.

We've built an AMI from this and deployed it in to loadtest and I've watched it do the correct thing with Terminating:Wait|Proceed both on workers with a grade report running and on frontend workers, GCing everything.

BTW, this likely was killing frontend workers abruptly, but at least those were out of the ELB when killed.

feanil (approved):

PR #3941, opened 2017-06-22
Updated ecommerce play to use edx_django_service play

clintonb:
- This removes the duplication across the various IDA configurations
- Updated Dockerfile for Docker-based devstack

LEARNER-818

Testing Plan (Login, payments)
- [x] Local (Docker)
- [x] Sandbox
- [x] Loadtest environment (See https://github.com/edx/edx-internal/pull/235#issuecomment-339369663 for a note about issues with PayPal in the loadtest environment. Still okay to proceed.)


clintonb:
FYI @edx/devops I'm currently deploying and testing in loadtest. Looking to get this merged in the next 1-2 weeks, depending on the fallout from testing and my schedule.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -20,33 +20,18 @@ ECOMMERCE_PIP_EXTRA_ARGS: "-i {{ COMMON_PYPI_MIRROR_URL }}"
 ECOMMERCE_NGINX_PORT: "18130"
 ECOMMERCE_SSL_NGINX_PORT: 48130
 
+ECOMMERCE_MEMCACHE: [ 'memcache' ]
```
feanil: This host won't exist in sandboxes, generally the defaults should work on sandboxes.  Can we add this in the docker ansible overrides instead and default this to `localhost`?

feanil (commented):
Quick comment.  Happy to do a larger review when you think this is ready.

playbooks/roles/ecommerce/defaults/main.yml
jibsheet: Also, if prod doesn't have a memcache, shouldn't devstacks use a process cache until prod does (to avoid any fun persistence bugs)

clintonb:
Hmm...we never setup loadtest for payments. 😭

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -245,27 +199,18 @@ ecommerce_home: "{{ COMMON_APP_DIR }}/{{ ecommerce_service_name }}"
...
-ecommerce_gunicorn_timeout: "300"
-
-ecommerce_log_dir: "{{ COMMON_LOG_DIR }}/{{ ecommerce_service_name }}"
-
-ecommerce_requirements_base: "{{ ecommerce_code_dir }}/requirements"
-ecommerce_requirements:
-  - production.txt
-  - optional.txt
 
+# TODO Find a way to access edx_django_service_environment in ecommerce/tasks
```
clintonb: @edx/devops any idea what I can do here? `edx_django_service_environment` is created in a dependent role, but requires some values passed in in `meta/main.yml`. If I use `edx_django_service_environment` in `ecommerce/tasks/main.yml`, it does not have the composed value (e.g. `DJANGO_SETTINGS_MODULE` is set to `null`). How can get the fully-composed value, if possible?

clintonb: I'm considering using [`include_role`](http://docs.ansible.com/ansible/latest/include_role_module.html) to dynamically include the role from the parent. The tasks in `ecommerce/tasks/main.yml` move to to `ecommerce/tasks/dependent.yml` and, based on the presence of certain variables, the `edx_django_service` task will call `include_role`.

jibsheet: I'm not actually sure what the right way to pass it back is.  Can you register something?  I'm surprised it isn't composed in ecommerce because meta'd roles should run first (and it's all one global state anyway).

clintonb: Solution from @feanil:

Pass arrays of dicts for the custom management commands. Each dict contains the command/args and the condition under which the command executes. There are two arrays—one for migrate commands, the other for install commands.

I'll get around to this next week.

clintonb:
@edx/learner please review

staubina:
Linking for reference per our conversation this morning

https://github.com/edx/configuration/pull/4150

Issued with ECOMMERCE_JWT_ISSUERS
and 
Changing the tasks to use different tags

docker/plays/ecommerce.yml
```diff
@@ -6,6 +6,7 @@
     serial_count: 1
   serial: "{{ serial_count }}"
   roles:
-    - common_vars
-    - docker
-    - ecommerce
+    - nginx
+    - role: ecommerce
+      nginx_default_sites:
```
feanil: nginx_default_site setting should be under the nginx role not under the ecommerce role correct?

playbooks/edx-east/edx_continuous_integration.yml
```diff
@@ -14,7 +14,6 @@
       - xqueue
       - xserver
       - analytics_api
-      - ecommerce
```
feanil, line 17: Is this being removed from here because the ecommerce role is now installing its own nginx profile?

playbooks/roles/ecommerce/defaults/main.yml
clintonb: Fixed!

clintonb:
@edx/devops @edx/learner please review

playbooks/edx-east/edx_continuous_integration.yml
clintonb: Yes.

docker/plays/ecommerce.yml
clintonb: That definitely makes sense. I copied the behavior from discovery: https://github.com/edx/configuration/blob/b84eb4096f20c3759d9e670d0dc01f73d2b8627c/docker/plays/discovery.yml. Not sure this is needed. Will dig deeper shortly.

clintonb: Alright, I got to the bottom of this. Discovery was wrong. `nginx_default_sites` only applies to the nginx role. Since we stopped using nginx on devstack, we never noticed the problem. I have corrected this and done a basic test to confirm that using supervisor+nginx+gunicorn mostly works.

The issues we previously saw with permissions on static assets remain but that is fine given given that we don't use those services for development. This will need to be resolved if we ever support Docker for production usage.

clintonb:
@feanil ready for final review! 🎉

docker/plays/discovery.yml
```diff
@@ -6,7 +6,7 @@
     serial_count: 1
   serial: "{{ serial_count }}"
   roles:
-    - nginx
-    - role: discovery
+    - role: nginx
       nginx_default_sites:
```
feanil, line 10: You have the default site for discovery but not the the `nginx_sites` list.  Will this cause the site profile to not get installed?

feanil: NM. forgot that the discover role now deals with installing the nginx profile.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -168,27 +150,18 @@ ECOMMERCE_BROKER_HOST: '{{ ansible_default_ipv4.address }}'
...
-  COURSE_CATALOG_API_URL: '{{ ECOMMERCE_COURSE_CATALOG_URL }}/api/v1/'
   ENTERPRISE_SERVICE_URL: '{{ ECOMMERCE_ENTERPRISE_URL }}/enterprise/'
   ECOMMERCE_URL_ROOT: '{{ ECOMMERCE_ECOMMERCE_URL_ROOT }}'
-  LMS_URL_ROOT: '{{ ECOMMERCE_LMS_URL_ROOT }}'
-  LMS_HEARTBEAT_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/heartbeat'
-  ENROLLMENT_API_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/api/enrollment/v1/enrollment'
-  COMMERCE_API_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/api/commerce/v1/'
-  LMS_DASHBOARD_URL: '{{ ECOMMERCE_LMS_URL_ROOT }}/dashboard'
+
+  # TODO Update this service and ecomworker to only use the central issuer.
```
feanil: Still a relevant TODO?  Is there a ticket for it if it's follow on work?

feanil (approved):
One small question on a TODO but otherwise looks good.  I keep forgetting that the nginx stuff is now a part of the service role.

playbooks/roles/ecommerce/defaults/main.yml
clintonb: Still relevant. I'll add the ticket number to the comment.

https://openedx.atlassian.net/browse/LEARNER-3041

clintonb:
Hmm...GitHub isn't let me delete the branch for some reason.

smarnach:
This PR broke the `edx_sandbox.yml` playbook, since it still lists `ecommerce` in `nginx_sites`.  I'm currently testing the obvious fix of simply removing that line (see [this commit](https://github.com/open-craft/configuration/commit/57ba199f00391de33b926d26b10dce83d9fc5322)).  I'll send a PR once I've verified that this works.

clintonb:
@smarnach sorry about that. Thanks for the forthcoming PR.

@edx/devops how can we prevent this in the future? Do we have anything that uses `edx_sandbox.yml`?

feanil:
I'm not sure who uses this playbook, I think it's used by various open-source examples.  @nedbat might know better.

smarnach:
OpenCraft is actually working on an automated test for that playbook.  It will continuously test the master branch, so it will only catch problems after the PR is merged, but it will still be helpful to get automatically notified.

clintonb:
It might be helpful to incorporate your test(s) into our Travis run @smarnach. That way it gets run for every PR and we are being proactive.

smarnach:
@clintonb Xavier discussed the approach with edX.  We actually want to run a full provisioning of a sandbox to test this, or maybe two in different configurations.  This takes about two hours on our infrastructure, so it would be rather inconvenient in Travis, and would also require quite a bit of additional resources.  We will start with the version that only tests the master branch, and will see where we can go from there.

nedbat:
(Sorry, just catching up to this.)  edx_sandbox.yml is the basis of the Native installation method we support for Open edX.  A grep of the configuration repo would find it used there, and for the fullstack Vagrant image (which is meant to mirror the Native installation).

The breakage mentioned in this discussion: was it fixed somehow?

smarnach:
@nedbat This was fixed in #4179.  (I didn't explicitly comment here, since the PR got automatically linked above.)

PR #3962, opened 2017-06-30
Cache all static assets for 5 minutes by default?

feanil:
Configuration Pull Request
---


jibsheet (approved):

ormsbee (approved):
I'm fine with this PR going forward, but I think that it points to a root problem at the app layer that I'll take a quick look into.

feanil:
@ormsbee makes sense, we should definitely fix as much as we can to to be hashed.  Though I think a default 5 minutes on all assets seems like a reasonable safety net so gonna merge this as well. I'll keep an eye out during the Wednesday deploy in-case it causes any issues.

PR #3965, opened 2017-07-03
Update Insights config for webpack

thallada:
We've had a [separate branch](https://github.com/edx/edx-analytics-dashboard/commits/thallada/kill-bower) in Insights for a while that removes bower and replaces require.js with webpack and it is finally getting ready to merge.

Please don't merge this PR until we are ready to merge the Insights PR because these changes aren't backwards compatible with the old system and GoCD builds will break until we merge the Insights PR.

**Question**: You can see from the commit history on this that I tried to switch Insights to Node 8.x. But, I got a pretty [indecipherable error in GoCD](https://gocd.tools.edx.org/go/tab/build/detail/loadtest-insights/39/build_ami/1/loadtest_edx#tab-console) that appeared to be coming from an ansible module, and I can't reproduce it locally. Do we have any IDAs using the latest version of Node? It's not required for this to merge, I was just hoping to get the [package.lock](https://docs.npmjs.com/files/package-locks) npm feature which is in the latest npm version bundled with Node 8.x.

To test this, I deployed to https://loadtest-edx-insights.edx.org/ using GoCD.

Configuration Pull Request
---

  - [X] Are you adding any new default values that need to be overridden when this change goes live? No. If so:
    - [X] Update the appropriate internal repo (be sure to update for all our environments)
    - [X] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [X] Add an entry to the CHANGELOG.
  - [X] Have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?

@edx/educator-dahlia


jibsheet:
I'm not aware of anyone else trying node 8 (based on a quick grep of the repo) but you could ask Learner directly.

As for the error.  Two things come to mind.  It appears to be reusing the node 6 nodeenv, could that cause issues?  Alternately, it could be an ansible incompatibility with newer node, but I'd have to look harder at that.  We could test the first one by building from scratch into loadtest and seeing what happens.

jibsheet:
As for the PR, this should get a CHANGELOG especially since it's backwards incompatible, and we should figure out how this coincides with ginkgo's branches.

jibsheet (approved):
My only question is if we should clean out the old bower directory or if that could cause issues?

We can build from scratch, but could also mark the directory as state:absent in a task to make sure it goes away for anyone else.

thallada:
@jibsheet I believe webpack should completely ignore the old bower_components directory so it shouldn't cause issues. If it isn't too difficult though, I think building from scratch would be a good idea just in case.

jibsheet:
Sounds good.  I think building from scratch is still manual and you may not have access to make the change in gocd, but any of us should be able to do it, just let us know when you're ready to run the pipeline.

jibsheet:
BTW - please squash down before merging, into whatever logical set of commits should go out.

PR #3968, opened 2017-07-06
Upgrade neo4j to 3.2.1

adampalay:
Configuration Pull Request
---

@fredsmith here's a PR to upgrade neo4j from 3.0.3—its current version—to 3.2.2, the most recent. As part of the upgrade, there are few things that I think will be prudent to do:

* As part of the [upgrade](https://neo4j.com/guides/upgrade/), neo4 changes the way it stores data. In order to mitigate the risk of the the updated data storage format, it makes a backup of the data before it does anything. As a result, we need to make sure we have 2x the storage that neo4j is currently taking up.
* It probably doesn't hurt to download a backup of the database before we run the update anyway
* We need to bring down neo4j while we do the update

    - [ ] ~Update the appropriate internal repo (be sure to update for all our environments)~
    - [ ] ~If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.~

### DIFF
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -53,6 +53,15 @@
     - install
     - install:base
 
+- name: turn off authentication (we're accessing neo4j from behind a private vpn)
+  lineinfile:
+    dest: "{{ neo4j_server_config_file }}"
+    regexp: "dbms.security.auth_enabled="
+    line: "dbms.security.auth_enabled=false"
+  tags:
+    - install
+    - install:configuration
+
 - name: set neo4j heap size
   lineinfile:
     dest: "{{ neo4j_server_config_file }}"
--- a/playbooks/roles/neo4j/tasks/main.yml
+++ b/playbooks/roles/neo4j/tasks/main.yml
@@ -65,8 +65,11 @@
 - name: set neo4j heap size
   lineinfile:
     dest: "{{ neo4j_server_config_file }}"
-    regexp: "dbms.memory.heap.max_size="
-    line: "dbms.memory.heap.max_size={{ neo4j_heap_max_size }}"
+    regexp: "{{ item }}="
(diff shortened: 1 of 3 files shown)


adampalay:
I've tested this upgrade locally, using a docker image of neo4j with a few local courses. I also tested this upgrade on a sandbox, but without data in it.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -92,6 +113,16 @@
     - install
     - install:base
 
+- name: set log dir for neo4j
```
adampalay: this shouldn't be here :)

adampalay:
@fredsmith  this is ready for a new review.

playbooks/roles/neo4j/tasks/main.yml
```diff
@@ -45,17 +53,36 @@
     - install
     - install:base
 
+- name: turn off authentication (we're accessing neo4j from behind a private vpn)
```
fredsmith: @adampalay if we're expecting people outside of edx to be using this, we should put this behind a when: feature flag and override it in configuration internal.  I think the default should be that authentication is on.

fredsmith (changes requested):

playbooks/roles/neo4j/tasks/main.yml
adampalay: @fredsmith fair enough.

adampalay:
@fredsmith this better? (corresponding branch: https://github.com/edx/edx-internal/compare/neo4j-server-auth?expand=1)

fredsmith (approved):

adampalay:
@fredsmith , thanks for the 👍 . I'm going to hold off merging this now until [OPS-2265](https://openedx.atlassian.net/browse/OPS-2265) is completed. Let me know how I can assist on that ticket.

adampalay:
@fredsmith , updated and rebased, just a heads up!

PR #3969, opened 2017-07-06
Estute/define plugins for build jenkins

estute:
This pr is not intended to be merged. It is only for reference (for https://openedx.atlassian.net/browse/TE-2105)

In the future, we plan to restructure all of the Jenkins roles, to work with https://github.com/edx/jenkins-configuration (part of which is captured here https://openedx.atlassian.net/browse/TE-2114).


estute:
@jzoldak the one plugin i wasn't sure about was http_request. If I recall correctly, this was just something we played around with when we tried the interceptor bake-off (jenkins version)

feanil:
@estute instead of adding more content to the jenkins_master role, I think it would be more useful to pull all the build jenkins related config into a 'build_jenkins' role similar to what we do for admin jenkins and tools jenkins.  Is this the right PR to be having that conversation?

estute:
@feanil yes. That is exactly what we are planning on doing. I do not intend to merge this. Just couldn't find a good spot to put this without doing all of the ansible work now (which we probably plan to do next sprint)

estute:
Wasn't sure if a gist, jira or wiki would be better for this

jzoldak:
LGTM. Thx @estute

PR #3975, opened 2017-07-10
use jenkins per-job credentials, not filesystem credentials

fredsmith:


feanil (approved):

feanil (changes requested):
We should close this.  The change will need to be different.

PR #3976, opened 2017-07-10
Add playbook for installing oauth2_proxy.

tobz:
This is a weaponized version of bit.ly's oauth2_proxy, which provides a reverse proxy implementation that allows you to add authn/authz in front of a web application(s).  I've been using this in Data Engineering for our metrics front-end, but will be extending the usage to other DE-maintained applications and want to start legitimizing the set-up.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/oauth2_proxy/defaults/main.yml
+++ b/playbooks/roles/oauth2_proxy/defaults/main.yml
@@ -1,0 +1,60 @@
+---
+oauth2_proxy_app_dir: "{{ COMMON_APP_DIR }}/oauth2_proxy"
+oauth2_proxy_conf_dir: "{{ COMMON_CFG_DIR }}/oauth2_proxy"
+
+oauth2_proxy_user: "oauth2_proxy"
+
+# We define this tuple here separately because we need to know it for downloading the right tarball.  Given that they
+# bake in both the version number -- which doesn't always match the actual Git tag they release off -- and the Go version,
+# it's nearly impossible to use only `oauth2_proxy_version` to build a valid URL.
+oauth2_proxy_version: "2.2.0"
+oauth2_proxy_version_tuple: "2.2.0.linux-amd64.go1.8.1"
+oauth2_proxy_pkg_name: "oauth2_proxy-{{ oauth2_proxy_version_tuple }}"
+oauth2_proxy_release_url: "https://github.com/bitly/oauth2_proxy/releases/download/v2.2/{{ oauth2_proxy_pkg_name }}.tar.gz"
+oauth2_proxy_release_sha256: "1c16698ed0c85aa47aeb80e608f723835d9d1a8b98bd9ae36a514826b3acce56"
+
+oauth2_proxy_listen_port: 4180
+oauth2_proxy_listen_addr: "0.0.0.0"
+oauth2_proxy_upstreams: ["localhost:80"] # List of address:port values acting as upstreams/backends.
+oauth2_proxy_request_logging: true
+oauth2_proxy_pass_basic_auth: true # Pass Basic Authorization header to upstream(s).
+oauth2_proxy_pass_user_headers: true # Passes X-Forwarded-User and X-Forwarded-Email to upstream(s).
+oauth2_proxy_pass_host_header: true # Pass original Host header to upstream(s).  If false, Host header will come from upstream address.
+oauth2_proxy_pass_access_token: true # Pass OAuth access token via X-Forwarded-Access-Token header to upstream(s).
+oauth2_proxy_email_domains: ["example.com"] # Which e-mail domains, if any, to validate for.  Needed for things like validating a specific G Suite apps domain, etc.
+oauth2_proxy_provider: "google" # OAuth provider type.
+oauth2_proxy_client_id: "CHANGEME-OAUTH2-CLIENT-ID" # OAuth client ID.
+oauth2_proxy_client_secret: "CHANGEME-OAUTH2-CLIENT-SECRET" # OAuth client secret.
+oauth2_proxy_custom_templates_dir: "" # Directory having template overrides for the login/error pages.
+oauth2_proxy_cookie_name: "_oauth2_proxy" # Client-side browser cookie name.
+oauth2_proxy_cookie_secret: "CHANGEME-COOKIE-SECRET" # Cookie encryption secret.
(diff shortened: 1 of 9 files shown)


playbooks/roles/oauth2_proxy/defaults/main.yml
```diff
@@ -0,0 +1,60 @@
...
+
+oauth2_proxy_listen_port: 4180
+oauth2_proxy_listen_addr: "0.0.0.0"
+oauth2_proxy_upstreams: ["localhost:80"] # List of address:port values acting as upstreams/backends.
+oauth2_proxy_request_logging: true
+oauth2_proxy_pass_basic_auth: true # Pass Basic Authorization header to upstream(s).
+oauth2_proxy_pass_user_headers: true # Passes X-Forwarded-User and X-Forwarded-Email to upstream(s).
+oauth2_proxy_pass_host_header: true # Pass original Host header to upstream(s).  If false, Host header will come from upstream address.
+oauth2_proxy_pass_access_token: true # Pass OAuth access token via X-Forwarded-Access-Token header to upstream(s).
+oauth2_proxy_email_domains: ["foo.com"] # Which e-mail domains, if any, to validate for.  Needed for things like validating a specific G Suite apps domain, etc.
```
feanil: NIT: `example.com` is what we've used for other such things, change this to that, for easier grepping of what to change.

jibsheet:
Realized on starting to read this that it's relevant to @fredsmith and @adampalay 's desire to figure out auth and neo4j.

playbooks/roles/oauth2_proxy/defaults/main.yml
```diff
@@ -0,0 +1,60 @@
...
+oauth2_proxy_provider: "google" # OAuth provider type.
+oauth2_proxy_client_id: "my-lil-client-id.apps.googleusercontent.com"
```
feanil: This is likely going to be overridden by the end user right?  If so it should be an all caps var along with any other vars that you expect to always be overriding.

playbooks/roles/oauth2_proxy/handlers/main.yml
```diff
@@ -0,0 +1,8 @@
+---
+- name: restart the oauth2_proxy service
```
feanil: We have generally avoided handlers because they don't work well with a role getting used multiple times with different config.  The last config persists.  Instead we generally just always restart at the end of `tasks/main.yml` or restart conditionally on specific changes(like config files changed.)

playbooks/roles/oauth2_proxy/meta/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+dependencies:
+  - role: common
+    tags:
+      - always
```
feanil: Why this tag?

adampalay:
@jibsheet , how so?

playbooks/roles/oauth2_proxy/meta/main.yml
tobz: So, I did this with the grafana/graphite stuff, too.

For some reason, the common role does not run in its entirety if I include it and am simply using the "install" tag.  If it doesn't run, supervisor doesn't seem to install.  Making it always run .... makes everything work.

I'm not sure if you guys run other roles with another tag that "wakes up" the rest of the common playbook, but this was a simple fix.

playbooks/roles/oauth2_proxy/handlers/main.yml
tobz: Ah yeah, woops.  I used the forums playbook as a skeleton, but meant to go back and add tasks for manage:start/stop/etc.

I'll do that.

playbooks/roles/oauth2_proxy/meta/main.yml
feanil: Usually on the first run, we run without tags.  I'm fine with this if you put in a comment about the reasoning in if you want to keep it.

feanil:
@adampalay this can be used as a proxy between the end-user and neo4j so that only users who have oauth credentials(edx account for example) can log in.

playbooks/roles/oauth2_proxy/meta/main.yml
tobz: Yeah, I'll add a comment explaining the reasoning.

playbooks/roles/oauth2_proxy/defaults/main.yml
tobz: Roger.

playbooks/roles/oauth2_proxy/defaults/main.yml
tobz: Understood.  Will change.

tobz:
Alright, all the code review comments should be addressed at this point.

tobz:
Can this get another pass?

/cc @edx/devops

playbooks/roles/oauth2_proxy/templates/oauth2_proxy_supervisor.conf.j2
```diff
@@ -0,0 +1,9 @@
+[program:oauth2_proxy]
+command={{ oauth2_proxy_app_dir }}/oauth2_proxy -config {{ oauth2_proxy_conf_dir }}/oauth2_proxy.cfg
+priority=999
+user={{ oauth2_proxy_user }}
+stdout_logfile={{ supervisor_log_dir }}/%(program_name)s-stdout.log
+stderr_logfile={{ supervisor_log_dir }}/%(program_name)s-stderr.log
+killasgroup=true
+stopasgroup=true
```
jibsheet: My only question is if you need both of these.  It sends the signal both to the parent process and all the children (instead of expecting the parent process to wait/reap children).  It's caused us issues in celery before, although we still use it in gunicorn.  

I didn't know if this was a conscious choice, or copy/pasta from our other supervisor templates.

jibsheet (approved):
Just one question about the supervisor config, otherwise looks good to ship.

playbooks/roles/oauth2_proxy/templates/oauth2_proxy_supervisor.conf.j2
tobz: This was a copypasta.  I'll drop them, since this is a Go binary that runs a single process with no children.

PR #3979, opened 2017-07-11
Adding all defaults required for VIDEO_IMAGE_SETTINGS

macdiesel:
CMS requires all of these values be in VIDEO_IMAGE_SETTINGS. This is currently preventing devstack docker from starting Studio.
---


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -673,6 +673,10 @@ EDXAPP_VIDEO_IMAGE_MAX_AGE: 31536000
 EDXAPP_VIDEO_IMAGE_SETTINGS:
   VIDEO_IMAGE_MAX_BYTES : 2097152
   VIDEO_IMAGE_MIN_BYTES : 2048
+  STORAGE_KWARGS:
+    location: "{{ edxapp_media_dir }}/"
```
robrap, line 677: 1. Seems like `edxapp_media_dir` and `EDXAPP_MEDIA_URL` should match case.
2. I don't know whether these both need/want the trailing "/"?  How is this being tested?

macdiesel: 1. I'm just working with what's already in the system, changing the case of this variable is out of the scope of this PR.
2. The trailing slash seems like the safer choice. Also it's what is already in the python settings file and we know it works in both prod and vagrant devstack.

jibsheet: Please don't change the variable case.  While the variable case is often wrong in the edxapp play, too much configuration is baked on it.

macdiesel: I 100% promise I will not change any variable case.

Could I get a thumb? =)

jmbowman (approved):

macdiesel:
@muhammad-ammar & @muzaffaryousaf these settings are overridden here:
https://github.com/edx/edx-internal/blob/master/ansible/vars/edx.yml#L628-L639

I see that location and base_url are not specified there. If this PR merges, will the application behave correctly in production even though we would be passing it kwargs that are not valid/used when configured for s3?

mushtaqak:
@macdiesel I don't think these values would cause any issue. I have checked sandbox is working correct for this config.

I think we would be fine for prod but @jibsheet can answer better.

macdiesel:
yeah it's in a if config.DEBUG block so it's good for prod and it solves the devstack issue as well.

PR #3984, opened 2017-07-12
Add Dockerfile for Selenium Firefox w/mp4 video support

jmbowman:
The [official Selenium docker containers](https://github.com/SeleniumHQ/docker-selenium) don't include the multimedia libraries required to support [mp4 playback on Firefox](https://github.com/SeleniumHQ/docker-selenium/pull/217) (partially because they add non-trivially to the container's size).  Quite a few of our JavaScript and acceptance tests count on this working, so this Dockerfile builds an image (derived from one of the official Selenium images) which we can use to run our tests in the latest Firefox with that particular obstacle resolved.

We'll probably want to rebuild and upload this automatically whenever the Dockerfile changes (like when we want to upgrade to a newer known-working version of the upstream image), but I didn't see anything in this repo to enable that.

I've also added support for generating a random, random-length password for VNC connections each time the container starts (and added a Dockerfile for a Chrome Selenium image which does this as well).  Once this is merged and we upload the images, I'll create a PR for the devstack repo which uses these images and adds a "vnc-passwords" make target which retrieves the current passwords from the container logs.

---

### DIFF
--- a/docker/build/chrome/Dockerfile
+++ b/docker/build/chrome/Dockerfile
@@ -1,0 +1,19 @@
+FROM selenium/standalone-chrome-debug:3.4.0-einsteinium
+MAINTAINER edxops
+
+USER root
+
+# Install a password generator
+RUN apt-get update -qqy \
+  && apt-get -qqy install \
+  pwgen \
+  && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+USER seluser
+
+CMD export VNC_PASSWORD=$(pwgen -s -1 $(shuf -i 10-20 -n 1)) \
+  && x11vnc -storepasswd $VNC_PASSWORD /home/seluser/.vnc/passwd \
+  && echo "Chrome VNC password: $VNC_PASSWORD" \
+  && /opt/bin/entry_point.sh
+
+EXPOSE 4444 5900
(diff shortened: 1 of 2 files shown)


jibsheet:
@MichaelRoytman Can you point @jmbowman to the Docker CI work in the DSL repo, I suspect firefox could be added to that pretty easily, although having it target changes specifically in this Dockerfile would be more work for someone implementing it.

michaelroytman:
@jmbowman This [job](https://github.com/edx/jenkins-job-dsl/blob/master/devops/jobs/DockerCI.groovy) runs on tools-edx-jenkins and builds and pushes up Docker images on changes to either configuration or to the IDA's codebase. Therefore, on a change to the Dockerfile (which would be a change to the configuration repository), the job should be triggered, and a new Docker image should be pushed up. However, it would build and push on every configuration change. We don't yet selectively build and push images based on the configuration changes.

The configuration file for the job lives [here](https://github.com/edx/edx-internal/blob/master/tools-edx-jenkins/docker-ci.yml). You should be able to add Firefox to that dictionary in a way similar to rabbitmq to create a job for continuously pushing this Dockerfile. We'll just have to be sure to reseed the jobs. If you don't have access to that repo, let me know, and I can give you access, since we've been rolling out permissions on an as-needed basis.

Does that sound like what you're looking for? Please let me know! Thanks.

jmbowman:
@MichaelRoytman That sounds good, and I in fact don't seem to have access to that repo.  We'll just need to add these Chrome and Firefox images to the list being built and uploaded.

michaelroytman (approved):

PR #3986, opened 2017-07-13
Remove old variables, add our more modern style of overrides

jibsheet:


michaelroytman (approved):

jibsheet:
@thallada would you mind double checking the variable deletions here?  It looks like Jarvis pulled them out years ago, but I always like to check.

thallada (approved):
I checked that we don't use those variables in the Analytics API config anymore. I'm assuming the real ANALYTICS_API_DEFAULT_* and ANALYTICS_API_REPORTS_* values are being defined in the secure config repo?

jibsheet:
@thallada Yeah, right now secure is overriding all of ANALYTICS_API_DATABASES: I'm trying to just override the passwords and then putting the database name, username and hostname in the internal config where folks can see it.  But that was hard to do when overriding the whole structure.

CHANGELOG.md
```diff
@@ -330,4 +330,8 @@
 - Role: insights
   - Removed `bower install` task
   - Replaced r.js build task with webpack build task
-  - Removed `./maange.py compress` task
\ No newline at end of file
+  - Removed `./maange.py compress` task
```
thallada: Nit: If you could fix my spelling mistake, that'd be great maan.

CHANGELOG.md
```diff
@@ -330,7 +330,11 @@
 - Role: insights
   - Removed `bower install` task
   - Replaced r.js build task with webpack build task
-  - Removed `./maange.py compress` task
+  - Removed `./mange.py compress` task
```
feanil: Mange?  I don't think we want to give the software mange.

feanil (approved):

feanil (approved):

PR #3990, opened 2017-07-17
Added Apple Pay merchant certificates to ecommerce role

clintonb:
LEARNER-1845

### DIFF
--- a/playbooks/roles/ecommerce/tasks/main.yml
+++ b/playbooks/roles/ecommerce/tasks/main.yml
@@ -170,6 +170,28 @@
     - install
     - install:configuration
 
+- name: Create Apple Pay certificates directory
+  file:
+    path: "{{ ecommerce_apple_pay_merchant_certificate_directory }}"
+    state: directory
+    owner: "{{ supervisor_user }}"
+    group: "{{ common_web_user }}"
+  tags:
+    - install
+    - install:configuration
+
+- name: Write Apple Pay merchant certificates
+  copy:
+    content: "{{ ECOMMERCE_APPLE_PAY_MERCHANT_CERTIFICATE }}"
+    dest: "{{ ecommerce_apple_pay_merchant_certificate_path }}"
+    owner: "{{ supervisor_user }}"
+    group: "{{ common_web_user }}"
+    mode: "0644"
+  no_log: true
+  tags:
+    - install
+    - install:configuration
+
 - name: Setup the ecommence env file
   template:
     src: "./{{ ecommerce_home }}/{{ ecommerce_service_name }}_env.j2"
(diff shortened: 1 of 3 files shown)


clintonb:
@edx/devops please review. I have tested this on my sandbox and confirmed the certificates are written as expected.

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -170,6 +170,29 @@
...
+
+- name: Write Apple Pay merchant certificates
+  copy:
+    content: "{{ item.value }}"
+    dest: "{{ ecommerce_apple_pay_merchant_certificate_directory }}/{{ item.key }}.pem"
+    owner: "{{ supervisor_user }}"
+    group: "{{ common_web_user }}"
+    mode: "0644"
+  no_log: true
+  with_dict: "{{ ECOMMERCE_APPLE_PAY_MERCHANT_CERTIFICATES }}"
```
jibsheet: Why is this a dict while up above we hardcode the edx.pem path into apple_pay_merchant_id_certificate_path
Do we actually have multiple certs or could this just be a simple multiline-yaml-string?

clintonb: Good question. Now that I have thought this through further, it makes no sense to configure this in a multi-tenant fashion. Only one CyberSource account is used, and there is no concept of profiles (as we do for credit card checkout). This should be flattened to a single certificate.

jibsheet (approved):
@feanil worth you looking at the permissions, but since we need app access, I think this is as good as we get to have it for this use case, rest of it looks fine to me.

playbooks/roles/ecommerce/tasks/main.yml
```diff
@@ -170,6 +170,28 @@
     - install
     - install:configuration
 
+- name: Create Apple Pay certificates directory
+  file:
+    path: "{{ ecommerce_apple_pay_merchant_certificate_directory }}"
+    state: directory
+    owner: "{{ supervisor_user }}"
+    group: "{{ common_web_user }}"
```
jibsheet, line 178: Feanil couldn't leave this note himself yesterday, so capturing for him.
common_web_user is the user that the gunicorn processes run as (www-data on sandboxes/prod) and while it's unfortunate that this is a user that also runs nginx, we'll have to live with it unless we're going to blow up the whole IDA configuration.

clintonb: No 💥, please.

jibsheet: Yeah, this was mostly a historical note about how we chose these users, I'm not expecting you to fix the IDA user roles today.

PR #3993, opened 2017-07-17
ENT-537: Fix SANDBOX_EDXAPP_USERS devstack error

shrlyhe:
Configuration Pull Request
---


clintonb:
How is this tested? [This line](https://github.com/edx/configuration/blob/31526b8f4cd29c95f7d3a0e7e758ded8ffe925d5/playbooks/roles/demo/tasks/deploy.yml#L31-L31) should always fail given that `SANDBOX_EDXAPP_USERS` is not declared anywhere.

Also, squash your commits. A one line change shouldn't need two commits.

shrlyhe:
@clintonb shouldn't that task skip since the conditional check doesn't pass? I'm not sure how to test besides re-provisioning devstack locally

clintonb:
@shrlyhe yes. My point is `SANDBOX_EDXAPP_USERS` needs a default set (ideally in `demo/defaults/main.yml`). Even setting it to an empty value is fine. It's also worth noting the pluralized name implies `SANDBOX_EDXAPP_USERS` is an array`; however, on the line in question the variable is being wrapped in array brackets. What is the type of `SANDBOX_EDXAPP_USERS`? Is it a dict, an array, or something else?

As for testing, you should be able to easily test your configuration branch for sandboxes with the Jenkins job. See https://openedx.atlassian.net/wiki/display/EdxOps/Sandboxes.

Testing for Docker Devstack should be straightforward, but I'm happy to help if it isn't.

1. Build an image (`make docker.build.edxapp` in edx/configuration)
2. Run Docker devstack provisioning (`make dev.provision` in edx/devstack) and ensure the [demo play](https://github.com/edx/devstack/blob/master/provision-lms.sh#L31) successfully executes.

shrlyhe:
@clintonb thanks for your comments. I made the suggested changes and ran Docker devstack provisioning successfully. Sandbox looks ok too: https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/2652/

clintonb (approved):

playbooks/roles/demo/tasks/deploy.yml
```diff
@@ -28,7 +28,7 @@
 
 - name: build staff, admin, and test user list
   set_fact:
-    demo_test_admin_and_staff_users: "{{ demo_test_and_staff_users + [SANDBOX_EDXAPP_USERS] }}"
+    demo_test_admin_and_staff_users: "{{ demo_test_and_staff_users + SANDBOX_EDXAPP_USERS }}"
```
feanil: If that list is empty by default, why not always run this task and get rid of the when condition?    Then you don't need two slightly different copies of the same task.

feanil: Once you do this, you can get rid of the 'SANDBOX_CREATE_ADMIN_USER' var.

shrlyhe: @feanil should I remove the var from sandbox-internal as well?

feanil (approved):
:+1: once tests pass.

PR #3995, opened 2017-07-17
Insights: Pass THEME_SCSS as an env var to webpack task

thallada:
After merging the Insights webpack work we realized we broke theming and Insights was always loading the open-edx theme. This change will pass the INSIGHTS_THEME_SCSS variable to the webpack task (via environmental variables) which an upcoming Insights PR will read to find the theme to use when building the static assets. Django no longer handles theming, so I removed it from the INSIGHTS_CONFIG django settings.

I tested this out on https://loadtest-edx-insights.edx.org/ and verified that the edx theme (not open-edx) is loaded.

Configuration Pull Request
---

  - [X] Are you adding any new default values that need to be overridden when this change goes live? No, I am changing an existing variable. If so:
    - [X] Update the appropriate internal repo (be sure to update for all our environments)
    - [X] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [X] Add an entry to the CHANGELOG.
  - [X] Have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)? On loadtest, yes.

### DIFF
--- a/playbooks/roles/insights/tasks/main.yml
+++ b/playbooks/roles/insights/tasks/main.yml
@@ -58,10 +58,10 @@
     production: yes
     state: latest
   become_user: "{{ insights_user }}"
+  environment: "{{ insights_environment }}"
   tags:
     - install
     - install:app-requirements
-  environment: "{{ insights_environment }}"
 
 - name: migrate
   shell: "DB_MIGRATION_USER='{{ COMMON_MYSQL_MIGRATE_USER }}' DB_MIGRATION_PASS='{{ COMMON_MYSQL_MIGRATE_PASS }}' {{ insights_venv_dir }}/bin/python {{ insights_manage }} migrate --noinput"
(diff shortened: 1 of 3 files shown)


efischer19 (commented):
I'm 👍 on this, but I believe we usually wait for an ops thumb before merging config PRs.

playbooks/roles/insights/tasks/main.yml
```diff
@@ -82,6 +82,7 @@
   tags:
     - assets
     - assets:gather
+  environment: "{{ insights_environment }}"
```
jibsheet: style nit - usually tags: is the last part of a task, can you push this up a few lines.

jibsheet (approved):

PR #4004, opened 2017-07-20
PLAT-1659 Remove git checkout from edxapp

jmbowman:
The edxapp container currently includes the /edx/app/edxapp/edx-platform directory which gets completely overridden in devstack by a volume representing the host's local edx-platform git checkout.  A local build indicates that removing this unused directory reduces the image size by about 1.35 GB (the compressed download size difference will be somewhat less).  Given that we download new versions of this image all the time, this seems like a pretty substantial win.

Can anybody think of good reasons why we shouldn't do this?  It would make it impossible to use the image without a local edx-platform checkout, but all of our current tooling assumes that you have that.  If we want to support Docker in production, we'll definitely need an image that has this directory, but we'd want to make enough other changes for that use case that it should be a separate image anyway.

---

### DIFF
--- a/docker/build/edxapp/Dockerfile
+++ b/docker/build/edxapp/Dockerfile
@@ -29,6 +29,7 @@ RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook edxapp.yml 
     --extra-vars=edx_platform_version=${OPENEDX_RELEASE} \
     --extra-vars="@/ansible_overrides.yml" \
     --extra-vars="@/devstack.yml" \
-    --extra-vars="@/devstack/ansible_overrides.yml"
+    --extra-vars="@/devstack/ansible_overrides.yml" \
+    && rm -rf /edx/app/edxapp/edx-platform
 
 EXPOSE 18000 18010


jmbowman:
@edx/docker-devstack-working-group This could use review, and in particular careful thought about potential pitfalls of doing this.

clintonb:
> It would make it impossible to use the image without a local edx-platform checkout...

This is my only concern. However, considering I haven't exercised this option in months, it's not a strong concern.

clintonb (approved):

jbarciauskas:
My question is, how do we do this in a way that supports a fullstack deployment approach in the future? Fullstack does *not* require a git checkout as such, but just a copy of the relevant code base

jibsheet:
For fullstack, you could just RM as a separate step and tag before and after as fullstack devstack
or make fullstack a shallow clone to save space

clintonb:
Image layering.

```
FROM edxops/edxapp:latest-without-checkout

RUN git clone https://github.com/edx/edx-platform.git /edx/app/edxapp/edx-platform
RUN paver update_assets 
```

This new image would be tagged as `edxops/edxapp:fullstack`, or whatever scheme we come up with.

jbarciauskas:
@clintonb right, that makes sense, just making sure we're not locking ourselves out or anything

macdiesel (approved):

macdiesel:
@gsong 🎉

gsong:
@macdiesel Can't wait to see the new image size.

jmbowman:
@gsong Regarding image sizes: https://hub.docker.com/r/edxops/edxapp/tags/ .  Not very precise measurements, but dropped from about 2 GB to about 1 GB.

gsong:
@jmbowman 👍

PR #4006, opened 2017-07-21
First stab at getting the sqlite fix script in

bmedx:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/files/edx/app/edxapp/upgrade_sqlite.sh
+++ b/playbooks/roles/edxapp/files/edx/app/edxapp/upgrade_sqlite.sh
@@ -1,0 +1,18 @@
+#!/usr/bin/env bash
+
+#
+# Script to update sqlite3 to a version which will not segfault
+# in out Django unit tests.
+#
+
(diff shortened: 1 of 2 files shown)


feanil:
Did the conversation about making the unit tests use MySQL already happen somewhere else?  I believe we use MySQL when testing with Jenkins don't we?

jibsheet:
Jenkins uses in-memory sqlite for unit tests (mysql for bokchoy)
https://github.com/edx/edx-platform/blob/master/lms/envs/test.py#L192-L200

feanil:
Ah, was confusing the two.  Got it.

bmedx:
These changes are specifically for devstack anyway. I believe, we're using sqlite for performance reasons there.

macdiesel (approved):

feanil (changes requested):
Instead of putting a shell script on the remote server and executing, all of that could be done via ansible.  Rather than making it a part of the edxapp role, it should just be a new role(something like `fix_docker_sqlite` maybe?) that does this work and is included in the docker devstack playbooks.

bmedx:
@feanil Ok, I'll kill this branch and start from scratch on a new PR then.

PR #4020, opened 2017-07-27
PLAT-1590 New container for edx-e2e-tests

jmbowman:
We want to be able to run the edx-platform e2e tests in Docker Devstack; this will help making development of those tests easier, sanity check major code changes locally, and let us run smoke tests against our built container images.  This test repo does browser automation and assumes a local checkout of edx-platform; in Docker Devstack, we can just reuse the existing Selenium containers and edx-platform containers and git checkout.  So all we really need is a new container with a pre-configured virtualenv, with the test repo content being mapped in from the host at runtime as with the edxapp and IDA containers.

I've already merged a PR with some minor fixes that were needed for edx-platform settings, and have an open [edx-e2e-tests PR](https://github.com/edx/edx-e2e-tests/pull/204) with the changes needed for the tests to run and pass reliably under Docker.

---

### DIFF
--- a/docker/build/e2e/Dockerfile
+++ b/docker/build/e2e/Dockerfile
@@ -1,0 +1,47 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/e2e/Dockerfile . -t edxops/e2e:latest
+
+
+FROM edxops/python:2.7
+MAINTAINER edxops
+USER root
+CMD bash -c 'while true; do sleep 60; done'
+
+ENV ACCESS_TOKEN="" \
+    BASIC_AUTH_USER="" \
+    BASIC_AUTH_PASSWORD="" \
+    CONFIG_ROOT="/edx/app/e2e" \
+    USER_LOGIN_EMAIL="staff@example.com" \
+    USER_LOGIN_PASSWORD="edx" \
+    COURSE_ORG="edX" \
+    COURSE_NUMBER="E2E-101" \
+    COURSE_RUN="course" \
+    COURSE_DISPLAY_NAME="Manual Smoke Test Course 1 - Auto" \
+    EDXAPP_CMS_DOC_LINK_BASE_URL="http://edx.readthedocs.io/projects/open-edx-building-and-running-a-course" \
+    GLOBAL_PASSWORD="" \
+    LMS_BASE_URL="edx.devstack.lms:18000" \
+    LMS_PROTOCOL="http" \
+    SELENIUM_BROWSER="chrome" \
+    SELENIUM_HOST="edx.devstack.chrome" \
+    SELENIUM_PORT="4444" \
+    STAFF_USER_EMAIL="" \
+    STUDIO_BASE_URL="edx.devstack.studio:18010" \
+    STUDIO_PROTOCOL="http" \
+    VIRTUAL_ENV="/edx/app/e2e/venvs/e2e"
+
+COPY docker/build/e2e/devstack.sh /edx/app/e2e/
+
+RUN apt-get update -qqy \
+    && apt-get -qqy install \
+    libxml2-dev \
+    libxslt1-dev \
+    && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
+    && git clone https://github.com/edx/edx-e2e-tests.git ${CONFIG_ROOT}/edx-e2e-tests \
+    && cd ${CONFIG_ROOT}/edx-e2e-tests \
+    && pip install -r requirements/base.txt \
+    && paver install_pages \
+    && rm -rf /edx/app/e2e/edx-e2e-tests/ \
+    && chmod u+x /edx/app/e2e/devstack.sh
--- a/docker/build/e2e/Dockerfile
+++ b/docker/build/e2e/Dockerfile
@@ -1,47 +1,26 @@
 # To build this Dockerfile:
 #
-# From the root of configuration:
+# From the root of the edx-e2e-tests repository:
 #
-# docker build -f docker/build/e2e/Dockerfile . -t edxops/e2e:latest
-
+# docker build -f Dockerfile . -t edxops/e2e:latest
 
 FROM edxops/python:2.7
 MAINTAINER edxops
-USER root
-CMD bash -c 'while true; do sleep 60; done'
 
-ENV ACCESS_TOKEN="" \
-    BASIC_AUTH_USER="" \
-    BASIC_AUTH_PASSWORD="" \
-    CONFIG_ROOT="/edx/app/e2e" \
-    USER_LOGIN_EMAIL="staff@example.com" \
-    USER_LOGIN_PASSWORD="edx" \
-    COURSE_ORG="edX" \
-    COURSE_NUMBER="E2E-101" \
-    COURSE_RUN="course" \
-    COURSE_DISPLAY_NAME="Manual Smoke Test Course 1 - Auto" \
-    EDXAPP_CMS_DOC_LINK_BASE_URL="http://edx.readthedocs.io/projects/open-edx-building-and-running-a-course" \
-    GLOBAL_PASSWORD="" \
-    LMS_BASE_URL="edx.devstack.lms:18000" \
-    LMS_PROTOCOL="http" \
-    SELENIUM_BROWSER="chrome" \
-    SELENIUM_HOST="edx.devstack.chrome" \
-    SELENIUM_PORT="4444" \
-    STAFF_USER_EMAIL="" \
-    STUDIO_BASE_URL="edx.devstack.studio:18010" \
-    STUDIO_PROTOCOL="http" \
-    VIRTUAL_ENV="/edx/app/e2e/venvs/e2e"
-
-COPY docker/build/e2e/devstack.sh /edx/app/e2e/
-
+# Install system libraries needed for lxml
 RUN apt-get update -qqy \
     && apt-get -qqy install \
     libxml2-dev \
     libxslt1-dev \
-    && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
-    && git clone https://github.com/edx/edx-e2e-tests.git ${CONFIG_ROOT}/edx-e2e-tests \
-    && cd ${CONFIG_ROOT}/edx-e2e-tests \
-    && pip install -r requirements/base.txt \
+    && rm -rf /var/lib/apt/lists/* /var/cache/apt/*
+
+ADD . /edx-e2e-tests
+WORKDIR /edx-e2e-tests
+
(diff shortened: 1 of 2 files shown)


docker/build/e2e/Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+USER root
+CMD bash -c 'while true; do sleep 60; done'
+
+ENV ACCESS_TOKEN="" \
+    BASIC_AUTH_USER="" \
+    BASIC_AUTH_PASSWORD="" \
+    CONFIG_ROOT="/edx/app/e2e" \
+    USER_LOGIN_EMAIL="staff@example.com" \
+    USER_LOGIN_PASSWORD="edx" \
+    COURSE_ORG="ArbiRaees" \
```
feanil: Can we change the name of this course to not be a person's name?  Maybe something like, 

```
COURSE_ORG="edx"
COURSE_NUMBER="e2e"
COURSE_RUN="course"
```
This is generally more informative.

docker/build/e2e/Dockerfile
```diff
@@ -0,0 +1,45 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/e2e/Dockerfile . -t edxops/e2e:latest
+
+
+FROM edxops/xenial-common:latest
```
clintonb: We have base images for Python: https://github.com/edx/edx-docker-base. Ubuntu seems a bit overkill for this purpose.

docker/build/e2e/Dockerfile
jmbowman: I'm up for that, but what's here is what gets imported by the course tarball in the edx-e2e-tests repo which all the tests were designed around.  Should I change that also?

docker/build/e2e/Dockerfile
jmbowman: Thanks, I'll try that out and see if it works.

clintonb: If it doesn't it's no big deal since we will have already downloaded Xenial for the other apps.

docker/build/e2e/Dockerfile
feanil: Unless it would be extremely hard to do so, I think we should change the course so it's purpose is obvious.  Knowing that a course called 'ArbiRaees' is the one you're supposed to be using with e2e tests seems like a level of indirection that doesn't need to exist.

jmbowman:
@feanil @clintonb Any further suggestions, or does this look good to go?

docker/build/e2e/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    VIRTUAL_ENV="/edx/app/e2e/venvs/e2e"
+
+COPY docker/build/e2e/devstack.sh /edx/app/e2e/
+
+RUN apt-get update -qqy \
+    && apt-get -qqy install \
+    libxml2-dev \
+    libxslt1-dev \
+    && rm -rf /var/lib/apt/lists/* /var/cache/apt/* \
+    && git clone https://github.com/edx/edx-e2e-tests.git ${CONFIG_ROOT}/edx-e2e-tests \
```
clintonb: Why is this all one command? We lose the benefits of layer caching with this setup.

docker/build/e2e/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+    && cd ${CONFIG_ROOT}/edx-e2e-tests \
+    && pip install -r requirements/base.txt \
+    && paver install_pages \
+    && rm -rf /edx/app/e2e/edx-e2e-tests/ \
```
clintonb: Why is this being deleted? Mounting a volume here might make sense if I am developing tests. However, if I just want to run them, it would be ideal if I didn't have to checkout another repo.

docker/build/e2e/Dockerfile
```diff
@@ -0,0 +1,47 @@
...
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/e2e/Dockerfile . -t edxops/e2e:latest
+
+
+FROM edxops/python:2.7
+MAINTAINER edxops
+USER root
+CMD bash -c 'while true; do sleep 60; done'
```
clintonb: ???

docker/build/e2e/devstack.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+cd /edx/app/e2e/edx-e2e-tests
```
clintonb: You don't need a separate file for this. Use [`WORKDIR`](https://docs.docker.com/engine/reference/builder/#workdir) in the Dockerfile.

clintonb (changes requested):
See comments for requested changes.

Not required, but it might make more sense for the Dockerfile to live alongside the code, similar to what we do in https://github.com/edx/credentials/. The repository of Dockerfiles is an anti-pattern that we only use because we rely on Ansible. Given that there is no reliance on Ansible, I see no reason for the Dockerfile to live in this repo. You also get the benefit of having Travis build images rather than needing another Jenkins job.

docker/build/e2e/Dockerfile
jmbowman: Mainly because of the performance impact of having multiple layers, but it probably would make sense to at least break it down into 2 layers (1 for the system updates, 1 for everything depending on the current repo content

docker/build/e2e/Dockerfile
jmbowman: Because that "install_pages" paver command checks out a full copy of edx-platform, and we currently expect people to have a local checkout of multiple repos when using devstack.  I could trim it to just delete the edx-platform checkout, the e2e repo itself isn't all that big

docker/build/e2e/devstack.sh
jmbowman: Ah, thanks.  This file originally also activated the virtualenv, but that went away with the switch from xenial to the minimal python image.

docker/build/e2e/Dockerfile
jmbowman: This is just to be able to launch the container along with the others in devstack's docker-compose.yml so it can find the appropriate volumes and other containers.  Would you rather see a default test command here, overridden in docker-compose.yml?  Or is there a better way of handling this?

docker/build/e2e/Dockerfile
clintonb: Performance impact is news to me. Can you share more details?

docker/build/e2e/Dockerfile
clintonb: Default test command makes sense. Given that this container is not necessarily needed for development, it doesn't make sense to include it in the default `docker-compose.yml`. That seems like a waste of CPU cycles.

docker/build/e2e/Dockerfile
clintonb: Does `paver install_pages` need to be run? Looking at the command, it seems that it's installing a bunch of packages which (strangely) all live in one repo. I'm guessing moving those to separate repos is out of scope, but we should probably track that as a debt ticket.

This is dependency hell. I will leave it to you to decide how you want to deal with it.

jmbowman:
One reason for managing the builds with Jenkins vs. Travis is that this image also has a dependency on edx-platform; it might be easier for us to have dual triggers for image builds in Jenkins.  @feanil , do you have strong feelings about whether this Dockerfile should live here or the edx-e2e-tests repo?

feanil:
If it needs to be triggered by commits to mulitple repos, I think this is the right place for it.  If it just needs to be triggered by changes to one repo, it should live in that repo.

docker/build/e2e/Dockerfile
jmbowman: But it needs a volume mount from edx-platform and access to Selenium container ports which are only visible to other running containers; it can't successfully run without either docker-compose configuration or a bunch of extra parameters to a `docker run` command.  An extra make command with all that pre-configured could work, but then you also lose the benefit of automatically getting updated images via `make pull`, etc.

docker/build/e2e/Dockerfile
jmbowman: The official guideline is [here](https://docs.docker.com/engine/userguide/eng-image/dockerfile_best-practices/#minimize-the-number-of-layers), but I can't find a good summary of the rationale for it in one place.  The reasons I've seen mentioned include:

* If you add a bunch of files (apt cache, tarball you extract and delete later, etc.) in one layer and then remove them in a later one, you still end up downloading them as part of the earlier layer.
* Docker uses a layered filesystem, so the more layers the container has, the longer it takes to find a file which is provided by one of the earlier layers (because it checks all the later ones first).
* There's a small amount of size and performance overhead in creating, storing, and loading each layer, so an image with fewer layers is slightly faster to build and load and takes a little less disk space.

Here's one reasonably good [writeup of the issues involved](https://www.ctl.io/developers/blog/post/optimizing-docker-images/).

docker/build/e2e/Dockerfile
jmbowman: Yeah, untangling that is out of scope for just "get the existing e2e tests running against Docker Devstack".  The tests currently use some code from edx-platform, so it has to be installed in this container's python environment.

docker/build/e2e/Dockerfile
clintonb: I understand those concerns. I am concerned that we are building another monolith in edx/devstack with all the extra services that are being added there. I find myself disabling many of the containers when I start devstack because I know I won't need them. This conversation should be moved elsewhere, but should we consider separate Docker Compose files for separate cases (e.g. docker-compose-e2e.yml knows about this container and the browsers)? I continue to not like wrapping commands in Make/Paver targets because that takes away an opportunity for devs to learn more about how everything works. Anecdotally, I think I've spent more time debugging Paver than it's saved me in typing commands.

/rant

docker/build/e2e/devstack.sh
jmbowman: Another reason for keeping this file, though: the `devstack` repo's Makefile assumes that each container has one: https://github.com/edx/devstack/blob/master/Makefile#L80

clintonb: It doesn't have to: https://github.com/edx/devstack/blob/master/Makefile#L82-L86

jmbowman:
This is probably going to move to the `edx-e2e-tests` repo; I just checked in a version that's close to what I'm probably going to propose as an enhancement over the one that was just checked in from https://github.com/edx/edx-e2e-tests/pull/201 .  I'll close out this PR once we confirm that having just the one Dockerfile there will work for all the intended use cases.

jmbowman:
Successfully uploaded the first auto-built image based on the Dockerfile over in the edx-e2e-tests repo, so closing this.

PR #3974, opened 2017-07-10
[asadiqbal08/ENT-322] Add 'enterprise_worker' user to LMS user provisioning

asadiqbal08:
TASK: [ENT-322](https://openedx.atlassian.net/browse/ENT-322)
The user is creating successfully on [sandbox](https://enterprise-worker-user-3.sandbox.edx.org)

Configuration Pull Request
---

Make sure that the following steps are done before merging:

  - [ ] A DevOps team member has approved the PR.
  - [ ] Are you adding any new default values that need to be overridden when this change goes live? If so:
    - [ ] Update the appropriate internal repo (be sure to update for all our environments)
    - [ ] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [ ] Add an entry to the CHANGELOG.
  - [ ] Have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?

PR #4022, opened 2017-07-28
Revert "[asadiqbal08/ENT-322] Add 'enterprise_worker' user to LMS user provisioning"

feanil:
Reverts edx/configuration#3974

Reverting this because it seems to be failing on our production deploys.  It looks like perhaps the defaults here mismatch the settings for this user in stage? 

https://gocd.tools.edx.org/go/tab/build/detail/STAGE_edxapp_B/575/run_play/1/run_play_job

@douglashall FYI


douglashall:
@feanil Thanks for the heads up!

PR #4026, opened 2017-08-01
ENT-572: Create permutation configuration file

shrlyhe:
Configuration Pull Request

This is the permutation JSON file that will be fed into the Python script that will do the actual permutation action (https://openedx.atlassian.net/browse/ENT-574). Only the fields that are currently supported by the course creation script have been included. Other fields (description, availability, title, language, pacing_type) will be included once support has been configured in the course creation script 

---

### DIFF
--- a/util/course-permutation-tool/permutations.json
+++ b/util/course-permutation-tool/permutations.json
@@ -1,0 +1,54 @@
+{
+    "permutation_data": {
+        "start": [
+            "past",
+            "future",
+            null
+        ],
+        "end": [
+            "past",
+            "future",
+            null
+        ],
+        "seats": [
+            [
+                "audit"
+            ],
+            [
+                "verified"
+            ],
+            [
+                "audit",
+                "verified"
+            ],
+            [],
+            null
+        ],
+        "display_name": [
+            "International Project Management",
+            "Cybersecurity Fundamentals",
+            "",
+            null
+        ],
+        "mobile_available": [
+            true,
+            false,
+            null
+        ]
+    },
+    "default_data": {
+        "start": "past",
+        "end": "future",
+        "seats": [
+            {
+                "type": [
+                    "audit",
+                    "verified"
+                ],
+                "upgrade_deadline": "future"
+            }
+        ],
+        "display_name": "International Project Management",
+        "mobile_available": true
+    }
+}
--- a/util/course-permutation-tool/README.rst
+++ b/util/course-permutation-tool/README.rst
@@ -1,0 +1,11 @@
+Course Permutation Tool for Developer Environments
+########################
+
+This is a tool to add default courses to developer environments, specifically for
+devstack and sandboxes. The goal is for developers to have access to courses with
+metadata that matches that in production, and to provide a way to generate course
+permutations. It will be comprised of a permutations JSON file, which
+includes permutation options and default values, and also a Python script that will
+generate the final file that will get passed into a course creation script.
+
+More info to come once finalized


util/permutations.json
```diff
@@ -0,0 +1,152 @@
+{
+    "permutation_data": {
+        "start": [
+            "2018-08-27T04:00:00Z",
```
feanil: Are these start dates meant to me always in the future?  That won't be true in a year for this date.

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+            "2017-03-06T05:00:00Z",
+            null
+        ],
+        "end": [
+            "2018-10-30T00:00:00Z",
```
feanil: Same question, is this meant to be always in the future?

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+                "price": "0.00",
+                "currency": "USD",
+                "upgrade_deadline": null,
+                "sku": "7B55983"
+            },
+            {
+                "type": "verified",
+                "price": "150.00",
+                "currency": "USD",
+                "upgrade_deadline": "2018-02-27T23:59:00Z",
```
feanil: Same question here, if it's supposed to be in the future, this will not work as expected in a year.

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+            }
+        ],
+        "pacing_type": [
+            "instructor_paced",
+            "self_paced",
+            null
+        ],
+        "marketing_url": [
+            "https://www.edx.org/course/ritxpm9001x?utm_source=mattdrayer_edx&utm_medium=affiliate_partner",
+            "https://www.edx.org/course/cybersecurity-capstone-ritx-cyber525x?utm_source=mattdrayer_edx&utm_medium=affiliate_partner",
```
feanil: You should remove the querystring from both of these URLs.

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+        "image": {
+            "width": null,
+            "height": null,
+            "src": "https://www.edx.org/sites/default/files/course/image/promoted/international_project_378x225.jpg",
+            "description": null
+        },
+        "pacing_type": "instructor_paced",
+        "marketing_url": "https://www.edx.org/course/international-project-management-ritx-pm9003x-1?utm_source=mattdrayer_edx&utm_medium=affiliate_partner",
```
feanil: Remove the query string.

util/permutations.json
shrlyhe: Would it be a good idea to pass in options such as "now", "past", and "future", and then dynamically generate the dates in the Python script instead?

feanil: You may even want something where you can say `X days in the future`, `X days in the past`, so you can test multiple future dates if you wanted.

shrlyhe: Sure, I can look into adding that as a feature in the script

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+                "upgrade_deadline": "2017-01-10T00:00:00Z",
+                "sku": "E29ADD9"
+            },
+            {
+                "type": null,
+                "price": null,
+                "currency": null,
+                "upgrade_deadline": null,
+                "sku": null
+            }
```
brittneyexline: I would add an empty json object and null here as well.
Since a course can have multiple modes available, there might have to be some limited permutation logic for this particular field to generate lists of different modes available. Or we could put the likely combinations here. We should talk about this offline probably...

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+                "height": null,
+                "src": "https://www.edx.org/sites/default/files/course/image/promoted/international_project_378x225.jpg",
+                "description": null
+            },
+            {
+                "src": null,
+                "height": null,
+                "width": null,
+                "description": null
+            }
```
brittneyexline: add an empty json object and null here as options.

util/permutations.json
```diff
@@ -0,0 +1,152 @@
...
+        ],
+        "pacing_type": [
+            "instructor_paced",
+            "self_paced",
+            null
+        ],
+        "marketing_url": [
+            "https://www.edx.org/course/ritxpm9001x",
```
brittneyexline: this is actually a little trickier than i thought now that i'm seeing this... if it's possible, we might want to generate this based on the course key, ie `course/course-v1:edX+DemoX+Demo_Course/about`
then, this would point to the course about page for the course created on the local environment. the only other consideration is that null and "" happen too, so we would also want to do permutations. we can talk about this offline too.

brittneyexline (commented):
looks pretty good, there are a couple things we should talk about how to approach when you get a chance.

util/permutations.json
shrlyhe: ok, I plan on generating the key dynamically in the script, so I can attach that to the marketing URL and generate it dynamically there, as well

shrlyhe: what do you mean by "null" and ""? are they different from the array values i've already included?

feanil:
@shrlyhe instead of putting this json file in the top level of the `util` directory, can we make a sub folder for this and the scripts that will be relying on it?  Maybe we could also add a `README` file there to explain what this is for and how to use it in the future as we flesh that out?

util/course-permutation-tool/permutations.json
```diff
@@ -0,0 +1,56 @@
+{
+    "permutation_data": {
+        "start": [
+            "now",
```
feanil: For start date, now and past are effectively the same aren't they since your usage of the course will always be after "now".

util/course-permutation-tool/permutations.json
```diff
@@ -0,0 +1,56 @@
...
+            "past",
+            "future",
+            null
+        ],
+        "end": [
+            "now",
```
feanil: Same here, I think now and past are redundant, since now becomes the past pretty quickly.

feanil (commented):
This seems like a pretty good start to me.  If it looks good to @brittneyexline who has more context, I think this part can land.

brittneyexline (approved):
:shipit:

PR #4028, opened 2017-08-02
jdmulloy/ops1849/remove 12.04

jdmulloy:
Configuration Pull Request
---


jdmulloy:
I have performed all the test listed on the  [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist) and I've tested that vagrant up works for vagrant/base/fullstack and vagrant/base/devstack.

vagrant/base/analyticstack/Vagrantfile
```diff
@@ -66,10 +66,6 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.provider :virtualbox do |vb|
     vb.customize ["modifyvm", :id, "--memory", MEMORY.to_s]
     vb.customize ["modifyvm", :id, "--cpus", CPU_COUNT.to_s]
-
-    # Allow DNS to work for Ubuntu 12.10 host
-    # http://askubuntu.com/questions/238040/how-do-i-fix-name-service-for-vagrant-client
-    vb.customize ["modifyvm", :id, "--natdnshostresolver1", "on"]
```
feanil, line 72: Let's check to make sure we don't need this anymore.  I'm a bit worried that we need it on 16.04 since it's not behind a conditional.

feanil (commented):
Generally looks good, we should also delete this folder:

https://github.com/edx/configuration/tree/master/docker/build/precise-common

playbooks/edx-east/edx_provision.yml
```diff
@@ -6,8 +6,8 @@
     keypair: continuous-integration
     instance_type: t2.medium
     security_group: sandbox-vpc
-    # ubuntu 12.04
-    ami: ami-f478849c
+    # ubuntu 16.04 - 20170721
+    ami: ami-cd0f5cb6
```
jibsheet, line 10: Did we decide if this was used?  I know I deferred to @feanil

playbooks/edx-east/retire_host.yml
```diff
@@ -15,7 +15,7 @@
   pre_tasks:
     - set_fact:
         STOP_ALL_EDX_SERVICES_EXTRA_ARGS: "--no-wait"
-      when: ansible_distribution_release == 'precise' or ansible_distribution_release == 'trusty'
+      when: ansible_distribution_release == 'trusty'
```
jibsheet: Can we drop the whole set_fact?
This only runs on edxapp machines and we don't have any of those on precise or trust?

vagrant/base/analyticstack/Vagrantfile
jdmulloy: I just checked that DNS resolution works in both vagrant VMs I created. I also realized after the fact that just building the VMs involves downloading many things over the internet, so if DNS wasn't working it would have broken very badly.

playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -157,7 +157,6 @@ common_debian_variants:
 
 # We only have to install old Python for these releases:
 old_python_ppa_releases:
-  - precise
   - trusty
```
jibsheet, line 160: Can we kill them both and kill this whole var?

playbooks/roles/mongo_3_0/defaults/main.yml
```diff
@@ -14,7 +14,7 @@ mongo_log_dir: "{{ COMMON_LOG_DIR }}/mongo"
 mongo_journal_dir: "{{ COMMON_DATA_DIR }}/mongo/mongodb/journal"
 mongo_user: mongodb
 
-MONGODB_REPO: "deb http://repo.mongodb.org/apt/ubuntu precise/mongodb-org/3.0 multiverse"
+MONGODB_REPO: "deb http://repo.mongodb.org/apt/ubuntu trusty/mongodb-org/3.0 multiverse"
```
jibsheet: I wonder how this will interact with our infrastructure which apparently is running precise binaries on trusty.
I'd suggest switching to the ansible var, but it would probably blow up sandboxes
We should also run a test of this against a mongo server, I can help with that.

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -140,7 +140,7 @@
     group: root
   when: >
     supervisor_service == "supervisor" and disable_edx_services and not devstack
-    and (ansible_distribution_release == 'precise' or ansible_distribution_release == 'trusty')
+    and (ansible_distribution_release == 'trusty')
```
jibsheet: Do we use supervisor on trusty?  This is less for @jdmulloy and more for the rest of @edx/devops

jibsheet (commented):
Can we also delete precise-common's Dockerfile now?

playbooks/edx-east/edx_provision.yml
feanil: I think it's overridden at various levels that call this script but is used if you call this directory.  If someone is doing that, I didn't want to break it.

playbooks/roles/common_vars/defaults/main.yml
feanil: Don't we still have trusty machines that  run common?

jibsheet: yep- sorry, misread the role.  Ah well.  Soon.

playbooks/roles/mongo_3_0/defaults/main.yml
feanil: Let's pull this change out of this PR so we can land the rest while we're testing this.  Though, Joe did bring up a sandbox from his branch and was able to get into courseware so we may just be able to :man_shrugging:  and move along?

playbooks/roles/supervisor/tasks/main.yml
feanil: Only the IDAs use pre_supervisor and they all run on 16.04, I think we can get rid of this task and the associated template.

playbooks/roles/mongo_3_0/defaults/main.yml
jibsheet: Sandboxes would work, you'd get a 14.04 binary on 16.04, I can check his sandbox.
I'm worried it would try to change the binaries running on our mongo servers 😬

jibsheet: Sandboxes work because they run 2.6
I think no good can come from this change.  Pull it out and we'll delete it with the 2.6 role when 3.2 ships.

jibsheet: To clarify - go back to precise and add a comment

jdmulloy: @jibsheet I have committed the change back to precise and included a comment explaining the situation/plan.

playbooks/roles/supervisor/tasks/main.yml
jdmulloy: Removed

playbooks/roles/supervisor/tasks/main.yml
```diff
@@ -115,18 +115,6 @@
     - install
     - install:base
 
-  # 12.04, 14.04, etc.
-- name: Create supervisor upstart job
```
feanil: @jdmulloy I think you deleted the wrong thin, this is the supervisor upstart definition.  This is still needed because we do run services on supervisor in Trusty.  It's the task below this one for pre_supervisor that needs to be removed.

playbooks/edx-east/retire_host.yml
jdmulloy: Should I also remove all the instances of STOP_ALL_EDX_SERVICES_EXTRA_ARGS from playbooks/roles/stop_all_edx_services/handlers/main.yml, or is it still useful?

playbooks/roles/supervisor/tasks/main.yml
jdmulloy: Fixed

playbooks/edx-east/retire_host.yml
jibsheet: Yeah, just remove this whole pre_tasks/set_fact.  We don't need to set this anymore.
We'll leave STOP_ALL_EDX_SERVICES_EXTRA_ARGS in the role in case anyone else needs it (like docker).

playbooks/roles/supervisor/tasks/main.yml
jibsheet: The file has been removed (pre_supervisor.conf.j2) but this task needs to go away since it still uses the template.

jibsheet (approved):

PR #4031, opened 2017-08-04
password min and max length constraints added 

tasawernawaz:
Configuration Pull Request
---
[LEARNER-1492](https://openedx.atlassian.net/browse/LEARNER-1492)

These settings can be tested on https://tasawer-config.sandbox.edx.org/

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -706,6 +706,10 @@ EDXAPP_ICP_LICENSE: !!null
 # Base Cookie Domain to share cookie across edx domains
 EDXAPP_BASE_COOKIE_DOMAIN: "{{ EDXAPP_LMS_SITE_NAME }}"
 
+# The minimum and maximum length for the password account field
+EDXAPP_PASSWORD_MIN_LENGTH: 2
+EDXAPP_PASSWORD_MAX_LENGTH: 75
+
 #-------- Everything below this line is internal to the role ------------
 
 #Use YAML references (& and *) and hash merge <<: to factor out shared settings
@@ -1157,6 +1161,8 @@ lms_env_config:
   CELERY_QUEUES: "{{ EDXAPP_LMS_CELERY_QUEUES }}"
   ALTERNATE_WORKER_QUEUES: "cms"
   ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES: "{{ EDXAPP_ENTERPRISE_COURSE_ENROLLMENT_AUDIT_MODES }}"
+  PASSWORD_MIN_LENGTH: "{{ EDXAPP_PASSWORD_MIN_LENGTH }}"
+  PASSWORD_MAX_LENGTH: "{{ EDXAPP_PASSWORD_MAX_LENGTH }}"
 
 cms_auth_config:
   <<: *edxapp_generic_auth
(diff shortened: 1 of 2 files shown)


tasawernawaz:
@edx/devops @edx/learner-spartans Please review.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -706,6 +706,10 @@ EDXAPP_ICP_LICENSE: !!null
 # Base Cookie Domain to share cookie across edx domains
 EDXAPP_BASE_COOKIE_DOMAIN: "{{ EDXAPP_LMS_SITE_NAME }}"
 
+# The minimum and maximum length for the password account field
+PASSWORD_MIN_LENGTH: 2
```
feanil: Is this the current actual default?  What happens if someone changes it?  Are people with shorter passwords forced to reset their passwords right away or is it just that new password resets will enforce the rule?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -706,6 +706,10 @@ EDXAPP_ICP_LICENSE: !!null
...
+PASSWORD_MAX_LENGTH: 75
```
feanil: Why this default max? Is this due to some database constraint?

feanil (commented):
There is also existing password constraint logic in the system.  @fredsmith has more info I believe.  How is this different from that?

fredsmith:
This code sets logic for the existing system that is in use previously, so we don't have to abuse edxapp_env_extra :+1: 

FYI, there are also settings for password complexity, https://github.com/edx/edx-platform/wiki/Optional-Password-Policy-Enforcement, although NIST is no longer recommending that passwords have complexity requirements, and now only recommends length and uniqueness tests:  https://pages.nist.gov/800-63-3/sp800-63b.html#sec5

playbooks/roles/edxapp/defaults/main.yml
schenedx: Maybe rename this to `EDXAPP_PASSWORD_MIN_LENGTH` to keep the names consistent?

playbooks/roles/edxapp/defaults/main.yml
schenedx: Maybe rename this to `EDXAPP_PASSWORD_MAX_LENGTH` to keep the names consistent?

schenedx (approved):
Two minor naming comments. Rest LGTM

playbooks/roles/edxapp/defaults/main.yml
fredsmith: NIST recommends these numbers be 8 and (at least) 64.  We're using 128 as the max in other places, and I can't see any reason not to set this at 128 as well.

playbooks/roles/edxapp/defaults/main.yml
fredsmith: according to @jibsheet, users will not be able to log in with a password that doesn't meet the password requirements, even though they can set one with the password reset process.

tasawernawaz: @schenedx Do you mean only here not at https://github.com/edx/configuration/blob/e3dfca780920422b203fea192265316879a76d27/playbooks/roles/edxapp/defaults/main.yml#L1164, because we do at both places then obviously we need to update the LMS code as well where ever we are using these variables.

schenedx: @tasawernawaz Yes, make the change only here. Then reference this variable in line #1164. We don't need to change the actual settings name at all.

feanil: @tasawernawaz @schenedx is it possible to accept the existing password but enforce the complexity on registrations and resets so that we can incrementally improve this?

tasawernawaz: @feanil We already have a ticket for removing length enforcement from login page. https://openedx.atlassian.net/browse/LEARNER-1209

PR #4036, opened 2017-08-08
Fixed SSO for sandboxes

ahsan-ul-haq:
LEARNER-2107

Configuration Pull Request
---

    - [x] Update the appropriate internal repo (be sure to update for all our environments)
    - [x] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.


playbooks/roles/credentials/defaults/main.yml
```diff
@@ -147,7 +147,7 @@ NGINX_CREDENTIALS_GUNICORN_HOSTS:
 CREDENTIALS_EXTRA_APPS: []
 
 CREDENTIALS_JWT_AUDIENCE: '{{ EDXAPP_JWT_AUDIENCE | default("SET-ME-PLEASE") }}'
```
clintonb: The audience and secret key need to be updated, too.

clintonb (changes requested):
Ensure that every OAuth/JWT-related variable is updated to rely on the common vars: https://github.com/edx/configuration/blob/master/playbooks/roles/common_vars/defaults/main.yml#L213-L221

playbooks/roles/credentials/defaults/main.yml
ahsan-ul-haq: Couldn't find audience in Ecommerce. Also we don't need to add them in ansible-provision.sh

clintonb: Ecommerce doesn't matter here. This should be updated to refer to `COMMON_JWT_AUDIENCE`. The eventual goal is that we have a single place, `common_vars`, to change values for all IDAs.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -147,7 +147,7 @@ NGINX_CREDENTIALS_GUNICORN_HOSTS:
...
-CREDENTIALS_JWT_ISSUER: '{{ CREDENTIALS_OAUTH_URL_ROOT }}'
+CREDENTIALS_JWT_ISSUER: '{{ COMMON_JWT_ISSUER }}'
 CREDENTIALS_JWT_SECRET_KEY: '{{ EDXAPP_JWT_SECRET_KEY | default("lms-secret") }}'
```
clintonb: `COMMON_JWT_SECRET_KEY`

clintonb (changes requested):
`CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_ISSUER` should point to `CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_ISSUER`.
`CREDENTIALS_OAUTH_URL_ROOT` should point to `COMMON_OAUTH_URL_ROOT`
`CREDENTIALS_OIDC_LOGOUT_URL` should point to `COMMON_OAUTH_LOGOUT_URL `

Make similar changes for the other IDAs. You'll know you're done when there are no references to edxapp for OAuth or JWT-related variables.

clintonb:
Nit: update your commit message to describe the solution (Fixed SSO for sandboxes), not the problem.

playbooks/roles/credentials/defaults/main.yml
jibsheet: Agree with Clinton - if we're going to change one of these vars, we should change them all, otherwise it's just confusing (and prevents us from being able to clean up the tech debt of the old settings)

Will this require secure changes, or are passthrough dummy vars still in place?  I forget what Clinton had us do in the backend.

clintonb: Unsecure defaults were setup by https://github.com/edx/edx-internal/pull/95. Secure overrides were also setup via DEVOPS ticket.

ahsan-ul-haq:
@clintonb made the suggested changes

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -53,8 +53,8 @@ CREDENTIALS_DJANGO_SETTINGS_MODULE: "credentials.settings.production"
 CREDENTIALS_DOMAIN: 'credentials'
 CREDENTIALS_URL_ROOT: 'http://{{ CREDENTIALS_DOMAIN }}:18150'
 CREDENTIALS_LOGOUT_URL: '{{ CREDENTIALS_URL_ROOT }}/logout/'
-CREDENTIALS_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
-CREDENTIALS_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
+CREDENTIALS_OAUTH_URL_ROOT: '{{ COMMON_OAUTH_URL_ROOT | default("http://127.0.0.1:8000") }}/oauth2'
```
clintonb: Ditch the defaults to avoid confusion and a potential source of bugs. The `common_vars` play will always be used.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -44,8 +44,8 @@ ECOMMERCE_DATABASES:
 ECOMMERCE_VERSION: "master"
 ECOMMERCE_DJANGO_SETTINGS_MODULE: "ecommerce.settings.production"
 
-ECOMMERCE_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
-ECOMMERCE_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
+ECOMMERCE_OAUTH_URL_ROOT: '{{ COMMON_OAUTH_URL_ROOT | default("http://127.0.0.1:8000") }}/oauth2'
```
clintonb: Ditch the defaults for these lines, too.

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -146,9 +146,9 @@ NGINX_CREDENTIALS_GUNICORN_HOSTS:
 
 CREDENTIALS_EXTRA_APPS: []
 
-CREDENTIALS_JWT_AUDIENCE: '{{ EDXAPP_JWT_AUDIENCE | default("SET-ME-PLEASE") }}'
-CREDENTIALS_JWT_ISSUER: '{{ CREDENTIALS_OAUTH_URL_ROOT }}'
-CREDENTIALS_JWT_SECRET_KEY: '{{ EDXAPP_JWT_SECRET_KEY | default("lms-secret") }}'
+CREDENTIALS_JWT_AUDIENCE: '{{ COMMON_JWT_AUDIENCE | default("SET-ME-PLEASE") }}'
```
clintonb: Same as above.

util/jenkins/ansible-provision.sh
```diff
@@ -269,6 +269,7 @@ COMMON_USER_INFO:
 USER_CMD_PROMPT: '[$name_tag] '
 COMMON_ENABLE_NEWRELIC_APP: $enable_newrelic
 COMMON_ENABLE_DATADOG: $enable_datadog
+COMMON_JWT_ISSUER: "https://${deploy_host}/oauth2"
```
clintonb: You want to define `COMMON_OAUTH_BASE_URL` here. Everything else is composed of it. See https://github.com/edx/configuration/blob/master/playbooks/roles/common_vars/defaults/main.yml#L212.

clintonb (changes requested):
Go ahead and squash your commits for the next round. This PR is small, and we don't gain anything from seeing the fixups. Also, I imagine the next round will get approved, and we can merge this.

util/jenkins/ansible-provision.sh
ahsan-ul-haq: @clintonb should it stay` http://127.0.0.1:8000` as defined in `common_vars`

clintonb: No. `COMMON_OAUTH_BASE_URL: "https://${deploy_host}"`

ahsan-ul-haq:
@clintonb made the change

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -53,8 +53,8 @@ CREDENTIALS_DJANGO_SETTINGS_MODULE: "credentials.settings.production"
 CREDENTIALS_DOMAIN: 'credentials'
 CREDENTIALS_URL_ROOT: 'http://{{ CREDENTIALS_DOMAIN }}:18150'
 CREDENTIALS_LOGOUT_URL: '{{ CREDENTIALS_URL_ROOT }}/logout/'
-CREDENTIALS_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
-CREDENTIALS_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
+CREDENTIALS_OAUTH_URL_ROOT: '{{ COMMON_OAUTH_URL_ROOT }}/oauth2'
```
clintonb: You don't need a path. See the values of the variables in https://github.com/edx/configuration/blob/master/playbooks/roles/common_vars/defaults/main.yml#L212.

clintonb (changes requested):
Once you've removed the paths, make sure you create a sandbox (with ecommerce and credentials) to actually test these changes.

ahsan-ul-haq:
@clintonb does the change seem fine now?

clintonb (approved):
Looks good. Please test on a sandbox before merging.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -44,8 +44,8 @@ ECOMMERCE_DATABASES:
 ECOMMERCE_VERSION: "master"
 ECOMMERCE_DJANGO_SETTINGS_MODULE: "ecommerce.settings.production"
 
-ECOMMERCE_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
-ECOMMERCE_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
+ECOMMERCE_OAUTH_URL_ROOT: '{{ COMMON_OAUTH_URL_ROOT }}'
```
feanil: You no longer use this variable below, does something else still use it?

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -53,8 +53,8 @@ CREDENTIALS_DJANGO_SETTINGS_MODULE: "credentials.settings.production"
 CREDENTIALS_DOMAIN: 'credentials'
 CREDENTIALS_URL_ROOT: 'http://{{ CREDENTIALS_DOMAIN }}:18150'
 CREDENTIALS_LOGOUT_URL: '{{ CREDENTIALS_URL_ROOT }}/logout/'
-CREDENTIALS_OAUTH_URL_ROOT: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/oauth2'
-CREDENTIALS_OIDC_LOGOUT_URL: '{{ EDXAPP_LMS_ROOT_URL | default("http://127.0.0.1:8000") }}/logout'
+CREDENTIALS_OAUTH_URL_ROOT: '{{ COMMON_OAUTH_URL_ROOT }}'
```
feanil: You no longer use this variable below, does something else still use it?

ahsan-ul-haq: yes it is being use in three other places as per I know.

playbooks/roles/ecommerce/defaults/main.yml
ahsan-ul-haq: yes this is also being used on other places

feanil: @ahsan-ul-haq can you point me to where?  This seems confusing because the app doesn't use it's own OAUTH_URL_ROOT setting.

ahsan-ul-haq: [here](https://github.com/edx/configuration/blob/b05150e32f8430e38d08376315bd631ced9d1d90/playbooks/roles/ecommerce/defaults/main.yml#L77) it is being used

clintonb: Use `COMMON_OIDC_ISSUER` instead of the role-specific variable. A goal here, in addition to getting sandboxes working, is to consolidate these variables as much as possible. As the person who probably wrote many of them, I find it confusing to have to mentally evaluate these variables when I need to make changes. The more we can get to using `COMMON_*` values, the better.

ahsan-ul-haq:
@clintonb I made some changes does this look like moving towards right direction?

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -146,17 +146,17 @@ NGINX_CREDENTIALS_GUNICORN_HOSTS:
 
 CREDENTIALS_EXTRA_APPS: []
 
-CREDENTIALS_JWT_AUDIENCE: '{{ EDXAPP_JWT_AUDIENCE | default("SET-ME-PLEASE") }}'
-CREDENTIALS_JWT_ISSUER: '{{ CREDENTIALS_OAUTH_URL_ROOT }}'
-CREDENTIALS_JWT_SECRET_KEY: '{{ EDXAPP_JWT_SECRET_KEY | default("lms-secret") }}'
+CREDENTIALS_JWT_AUDIENCE: '{{ COMMON_JWT_AUDIENCE }}'
```
clintonb: If these are no longer used, they should be removed.

clintonb (approved):
There is still room for improvement, but this is good enough for now. The remaining variables will be taken care of when we consolidate to the common role.

Again, make sure this is tested on a sandbox before merging. Also, squash commits.

playbooks/roles/ecommerce/defaults/main.yml
feanil: Right, so we should be able to delete the role specific `ECOMMERCE_OAUTH_URL_ROOT` from this role when this PR is ready.

clintonb: @feanil given that the main goal here is to get SSO working on sandboxes are you okay with holding off on any additional variable consolidation? I am willing to let this go as-is (once tested). Additional cleanup can happen later, or resources can dedicated to [LEARNER-812](https://openedx.atlassian.net/browse/LEARNER-812).

feanil: @clintonb my biggest concern is that there is a variable called `ECOMMERCE_OAUTH_URL_ROOT` does not actually set the ecommerce app's 'OAUTH_URL_ROOT'  If you think that it will be fixed quickly enough, I think we should add a TODO comment to that effect and merge this.

clintonb: I went ahead and finished the cleanup so as not to bog down @ahsan-ul-haq. My goal was expose someone else to configuration. That goal has been sufficiently met.

feanil:
@ahsan-ul-haq please also add a CHANGELOG entry since we are removing "public" variables.  And make a DEVOPS support ticket to clean up our internal/secure config.

clintonb:
Sandbox building: https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/3421/console

clintonb:
I tested this on my sandbox, and it "works". By that I mean the Django settings are correct. However, logins fail because the tenants aren't setup in the respective services and the OAuth clients (in LMS) aren't not linked to users. This PR is good to merge (after commits are squashed), but at least two more commits/PRs will be needed:

1. Create service users, and associate them with the OAuth clients.
2. Create tenants for ecommerce and credentials at provision time.

ahsan-ul-haq:
@feanil updated changelog @clintonb removed unused vars.

CHANGELOG.md
```diff
@@ -359,3 +359,15 @@
 - Role: edxapp
   - Added `PASSWORD_MIN_LENGTH` for password minimum length validation on reset page.
   - Added `PASSWORD_MAX_LENGTH` for password maximum length validation on reset page.
+
+- Role: credentials
+  - Removed `CREDENTIALS_OAUTH_URL_ROOT` which is no longer used, as `COMMON_OAUTH_URL_ROOT` from `common_vars` is used
```
clintonb: These can be more succinct, shorter: Replaced `x` with `y` from `common_vars`

playbooks/roles/credentials/defaults/main.yml
```diff
@@ -66,7 +64,7 @@ CREDENTIALS_LANGUAGE_CODE: 'en_US.UTF-8'
 CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_KEY: 'SET-ME-TO-A-UNIQUE-LONG-RANDOM-STRING'
 CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_SECRET: 'SET-ME-TO-A-UNIQUE-LONG-RANDOM-STRING'
 CREDENTIALS_SOCIAL_AUTH_REDIRECT_IS_HTTPS: false
-CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_ISSUER: '{{ CREDENTIALS_OAUTH_URL_ROOT }}'
+CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_ISSUER: '{{ COMMON_JWT_ISSUER }}'
```
clintonb: This variable is no longer needed. Replace its usages with the `COMMON_JWT_ISSUER`.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -56,25 +53,24 @@ ECOMMERCE_EDX_API_KEY: 'PUT_YOUR_API_KEY_HERE'  # This should match the value se
...
 ECOMMERCE_JWT_SECRET_KEYS:
-  - '{{ ECOMMERCE_JWT_SECRET_KEY }}'
+  - '{{ COMMON_JWT_SECRET_KEY }}'
 
 # Used to automatically configure OAuth2 Client
 ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_KEY : 'ecommerce-key'
 ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_SECRET : 'ecommerce-secret'
 ECOMMERCE_SOCIAL_AUTH_REDIRECT_IS_HTTPS: false
-ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_ISSUER: '{{ ECOMMERCE_OAUTH_URL_ROOT }}'
+ECOMMERCE_SOCIAL_AUTH_EDX_OIDC_ISSUER: '{{ COMMON_JWT_ISSUER }}'
```
clintonb: This variable is no longer needed. Replace its usages with the `COMMON_JWT_ISSUER`.

ahsan-ul-haq:
@clintonb Do you think we need to create new service users for credentials and Ecommerce? Also can you explain how we can create tenants for ecommerce and credentials at provision time?

clintonb:
Yes. Every service needs its own service user. That's how we do it in production. That's how we should do it everywhere else. There should already be an example of this for ecommerce.

Provisioning sites is a matter of calling the appropriate management command. Here is an example for ecommerce: https://github.com/edx/devstack/blob/master/provision-ecommerce.sh#L7. Search the management commands in edx/credentials to find the appropriate one there. When you do, please update Docker Devstack, too, since others refuse to step up and take responsibility for doing so. :(

schenedx (approved):
👍

feanil (approved):

clintonb (approved):

clintonb:
Remember to squash commits.

andy-armstrong:
FYI, it appears that this caused a production issue when rolled out this morning:

https://openedx.atlassian.net/browse/LEARNER-2235

PR #4040, opened 2017-08-10
mroytman/fix duplicate image output bug and incorrect image identification bug

michaelroytman:
Configuration Pull Request
---

### DIFF
--- a/util/parsefiles.py
+++ b/util/parsefiles.py
@@ -343,24 +343,34 @@ def _get_modified_dockerfiles(files, git_dir):
     for f in files:
         file_path = pathlib2.Path(git_dir, f)
         if file_path in candidate_files:
-            items.add(_get_play_name(file_path))
+            play = items.add(_get_play_name(file_path))
+
+            if play is not None:
+                items.add(play)
     return items
 
 
 def _get_play_name(path):
 
     """
-    Gets name of play from the filepath, which is
-    the directory following occurence of the word "build".
+    Gets name of play from the filepath, which is the token
+    after either "docker/build" in the file path.
 
     Input:
     path: A path to the changed file under docker/build dir
     """
-    # get individual parts of a file path
-    dirs = path.parts
-    # name of play
-    return dirs[dirs.index("build")+1]
 
+    # attempt to extract Docker image name from file path; splits the path of a file over
+    # "docker/build/", because the first token after "docker/build/" is the image name
+    suffix = (str(path)).split(str(os.path.join('docker', 'build', '')))
+
+    # if file path contains "docker/build/"
+    if len(suffix) > 1:
+        # split suffix over separators to file path components separately
+        suffix_parts = suffix[1].split(os.sep)
+        # first token will be image name; <repo>/docker/build/<image>/...
+        return suffix_parts[0]
+    return None
 
 def arg_parse():
 
@@ -422,8 +432,5 @@ if __name__ == '__main__':
     # Add playbooks to the list whose docker file has been modified
     modified_docker_files = _get_modified_dockerfiles(change_set, TRAVIS_BUILD_DIR)
 
-
-    # prints Docker plays
-    print " ".join(str(play) for play in docker_plays),
-
-    print " ".join(str(dock) for dock in modified_docker_files)
+    all_plays = set(set(docker_plays) | set( modified_docker_files))
+    print " ".join(all_plays)


michaelroytman:
This fixes the bug in the _get_role_name method; Travis's environment is at "/home/travis/build/edx/configuration/...", so identifying the Docker image/play name by grabbing the next token after the first instance of the word "build" is incorrect, as it resolves to "edx", instead of the name of the Docker image. Instead, we look for the second token after the first instance of the word "docker".

michaelroytman:
Also, I address https://openedx.atlassian.net/browse/OPS-2334 here as well, since I was tinkering with parsefiles.py anyway.

fredsmith (approved):

jdmulloy:
I like the deduplication.

I don't know anything about the script, but I'm curious if it will work correctly outside of travis. If it's not used anywhere outside of Travis then it doesn't matter.

jibsheet (approved):

jibsheet:
Can you retitle the PR for posterity Michael?

michaelroytman:
@jdmulloy The script is only being used on Travis and will be used on in a Jenkins job soon, but I don't expect there to be another instance of the word "docker" in the relevant file paths, so I don't think it should cause the same problem again. I think, ideally, we'd check for the prefix "configuration/docker", but it's slightly more complicated. I could do something like this.

```
def _get_play_name(path):

    """
    Gets name of play from the filepath, which is
    the directory two tokens following occurence of 
    the word "docker"; note that we do this because files 
    in Travis are stored at "/travis/build/edx/...", so using
    the token after "build" incorrectly identifies "edx" as the
    play name.

    Input:
    path: A path to the changed file under docker/build dir
    """
    # get individual parts of a file path
    # split over "configuration/docker/build/", '' is to add trailing slash to avoid removing later
    separate_prefix = (str(path)).split(str(os.path.join('configuration','docker', 'build', '')))
    # if prefix found
    if len(separate_prefix) > 1:
        # split suffix over separators
        suffix_parts = separate_prefix[1].split(os.sep)
        # first token will be image name; configuration/docker/build/<image>/...
        return suffix_parts[0]
    return None
```

Thoughts? I don't actually think we need the conditional, since the method is only used on files under the "configuration/docker" subdirectory anyway.

@jibsheet Sure thing.

util/parsefiles.py
```diff
@@ -351,15 +351,19 @@ def _get_play_name(path):
...
+    play name.
 
     Input:
     path: A path to the changed file under docker/build dir
     """
     # get individual parts of a file path
     dirs = path.parts
     # name of play
-    return dirs[dirs.index("build")+1]
+    return dirs[dirs.index("docker")+2]
```
feanil: If you actually care about things in the 'docker/build' directory, shouldn't we change this test to something ensures that the item after 'docker' is 'build'?

michaelroytman: @feanil Could you look at my comment above? The piece of code up there would check for configuration/docker/build/.

michaelroytman:
@jibsheet @feanil @fredsmith @jdmulloy Could one of you please re-review this? I made some changes in response to Feanil's comment above. Thank you!

util/parsefiles.py
```diff
@@ -343,25 +343,35 @@ def _get_modified_dockerfiles(files, git_dir):
...
     """
-    # get individual parts of a file path
-    dirs = path.parts
-    # name of play
-    return dirs[dirs.index("build")+1]
+
+    # attempt to extract Docker image name from file path
 
 
+    suffix = (str(path)).split(str(os.path.join('docker', 'build', '')))
```
feanil: Add a comment to summarize what this is doing.

util/parsefiles.py
```diff
@@ -422,8 +432,5 @@ def arg_parse():
     # Add playbooks to the list whose docker file has been modified
     modified_docker_files = _get_modified_dockerfiles(change_set, TRAVIS_BUILD_DIR)
 
-
-    # prints Docker plays
-    print " ".join(str(play) for play in docker_plays),
-
-    print " ".join(str(dock) for dock in modified_docker_files)
+    all_plays = set(set(docker_plays) | set( modified_docker_files))
+    print " ".join(str(play) for play in all_plays)
```
feanil: You can replace this line with `print " ".join(all_plays)` that should work correctly.

PR #4041, opened 2017-08-10
ENT-574: Create course permutation script 

shrlyhe:
Configuration Pull Request
---


util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+
+    wrapper_courses_dict = {}  # needed to match course input file creation
+    wrapper_courses_dict["courses"] = results
+
+    with open("test_courses.json", "w") as outfile:
+        json.dump(wrapper_courses_dict, outfile)
+
+
+if __name__ == "__main__":
+    num_args = arg_parse()
```
brittneyexline: what is the name num_args supposed to reflect? is it the actual parsed arguments or the number of arguments parsed? i would rename for clarity.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+def arg_parse():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--seed_data_file', help="Input a permutation configuration file")
+    parser.add_argument('--fields', nargs="*", action="append", default=None,
+                        help="Specify which fields to generate permutations on")
+
+    return parser.parse_args()
+
+
+def parse_field_arguments():
```
brittneyexline: consider renaming to something like `process_field_arguments` since they are already parsed above. also take those parsed arguments as an argument to this function.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+    file = open(num_args.seed_data_file)
```
brittneyexline: as above, `num_args` should have been passed in as an argument - err towards being explicit when utilizing variables generated in different contexts/scopes.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+
+    permutation_data = file_data["permutation_data"]
+
+    default_data_keys = permutation_data.keys()
+
+    fields = {}
+
+    # if no field arguments are given, just return default data
+    if not num_args.fields:
+        default_permutation = file_data["default_data"]
```
brittneyexline: these two lines could be combined into one since default_permutation doesn't get used anywhere else

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+        fields = default_permutation
+    else:
+        field_length = len(num_args.fields[0])
```
brittneyexline: what is num_args.fields[0]? perhaps that should be pulled into a user friendly named variable above.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+                        help="Specify which fields to generate permutations on")
+
+    return parser.parse_args()
+
+
+def parse_field_arguments():
+    file = open(num_args.seed_data_file)
+    file_data = json.load(file)
+
+    default_data = file_data["default_data"]
```
brittneyexline: we should add error handling if these expected arguments are empty or missing.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+    fields = {}
+
+    # if no field arguments are given, just return default data
+    if not num_args.fields:
+        default_permutation = file_data["default_data"]
+        fields = default_permutation
+    else:
+        field_length = len(num_args.fields[0])
+
+        if (field_length > 3):
```
brittneyexline: in general, i think some preprocessing above for num_args.fields with clear names will help with the readability and simplicity of this whole section.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+        # add each command line field to fields dict
+        for permutation_choices in num_args.fields:
+            for i in range(0, field_length):
+                fields[permutation_choices[i]] = permutation_data[permutation_choices[i]]
+
+        # the difference btwn all possible fields and the chosen permutation ones
+        default_fields = list(set(default_data_keys) - set(num_args.fields[0]))
+
+        # add non permutation fields to dict
+        for j in range(0, len(default_fields)):
```
brittneyexline: in python, you should be able to say `for field in default_fields` and use field as the variable in the loop

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+    wrapper_courses_dict["courses"] = results
+
+    with open("test_courses.json", "w") as outfile:
+        json.dump(wrapper_courses_dict, outfile)
+
+
+if __name__ == "__main__":
+    num_args = arg_parse()
+    parse_field_arguments()
+    generate_permutations(parse_field_arguments(), 0, [], {}, {})
```
brittneyexline: why does parse field arguments get called twice? might be better to store the results in a variable and pass the variable in.

since the function you are calling is recursive, i would make a separate function that starts out the recursive function to hid the details of those parameters from the top level caller right here. from reading this, i couldn't say what those other arguments do.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+
+        # add non permutation fields to dict
+        for j in range(0, len(default_fields)):
+            fields[default_fields[j]] = default_data[default_fields[j]]
+
+    return fields
+
+
+def generate_permutations(fields, index, results, courses_dict, fields_dict):
+    all_permutations_keys = fields.keys()
```
brittneyexline: what is the difference between fields and fields_dict if fields is a dict too?

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+    permutation_option = all_permutations_keys[index]
+    permutations_values = fields[permutation_option]
+
+    for i in range(len(permutations_values)):
```
brittneyexline: again, we can just do `for permutation_value in permutation_values`

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+        # add other required default fields to dict
+        courses_dict["number"] = None  # will be generated automatically by course creation script
```
brittneyexline: we should define these defaults at the top of the file as a static variable or constant.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+        # configure enrollment seat settings
+        enrollment_dict = {}
+        enrollment_dict["credit"] = False
+        enrollment_dict["credit_provider"] = "test-credit-provider"
+
+        # add permutation fields to dict
+        fields_dict[permutation_option] = permutations_values[i]
+        # generate dates
+        now = datetime.datetime.now(pytz.UTC)
+        if permutations_values[i] == "future":
```
brittneyexline: define the strings future and past as constants at the top of the file

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+            future = str(now + datetime.timedelta(days=365))
```
brittneyexline: also, this date translation could be abstracted into its own helper function that these two places call to.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+        if permutations_values[i] == None:
+            fields_dict[permutation_option] = None
+        # add audit and verify fields to dict
+        if all_permutations_keys[i] == "audit" and permutations_values[i] == True:
+            enrollment_dict["audit"] = permutations_values[i]
+        if all_permutations_keys[i] == "verify" and permutations_values[i] == True:
+            enrollment_dict["verify"] = True
+
+        if index + 1 < len(all_permutations_keys):
+            generate_permutations(fields, index + 1, results, courses_dict, fields_dict)
```
brittneyexline: i'm a little confused about if recursive calls to this function end up writing to the file too... i would make sure to structure and comment the code so that it is clear what the recursion is doing. i am having trouble knowing if this implementation is correct based on reading this section.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,113 @@
...
+
+        courses_dict["enrollment"] = enrollment_dict
+        courses_dict["fields"] = fields_dict.copy()
+        results.append(courses_dict.copy())
+
+    wrapper_courses_dict = {}  # needed to match course input file creation
+    wrapper_courses_dict["courses"] = results
+
+    with open("test_courses.json", "w") as outfile:
```
brittneyexline: make the file name a constant.

brittneyexline (commented):
This is a good rough draft. Think of it just like writing an essay - you've gotten your first draft down and it says/does what you want it to. But think of the next round of revisions with the mindset that you need to make what you've written obvious to the reader, so they can easily understand the function and purpose of your code. Happy to discuss in person as well if you need more clarification.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+# ./generate_permutations.py --seed_data_file permutations.json --fields field1 field2 field3
+
+import argparse
+import json
+import datetime
+import pytz
+import sys
+
+# course defaults
+course_number = None
```
brittneyexline: usually constants are capitalized

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+def arg_parse():
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--seed_data_file', help="Input a permutation configuration file")
+    parser.add_argument('--fields', nargs="*", action="append", default=None,
+                        help="Specify which fields to generate permutations on")
+
+    return parser.parse_args()
+
+
+def process_field_arguments(parser_args, seed_data_file, field_args):
```
brittneyexline: you should either pass parser_args alone or seed_data_file and field_args. it's probably easier to drop the two args and keep parser_args because you're not using them.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+    try:
+        file = open(parser_args.seed_data_file)
+        file_data = json.load(file)
+        default_data = file_data["default_data"]
+        permutation_data = file_data["permutation_data"]
+        default_data_keys = permutation_data.keys()
+    except IOError:
+        print sys.exc_value
```
brittneyexline: this could be a more informative message - what does it mean if we get an IOError?

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+        sys.exit()
+    except KeyError:
+        print("Could not find key {}".format(sys.exc_value))
+        sys.exit()
+
+    field_args = {}
+
+    # if no field arguments are given, just return default data
+    if not parser_args.fields:
```
brittneyexline: parser_args.fields should be put into its own variable for clarity.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+        field_args = file_data["default_data"]
+    else:
+        num_field_args = len(parser_args.fields[0])
+        if (num_field_args > 3):
+            raise argparse.ArgumentTypeError("Only a max of 3 fields allowed")
+        # add each command line field to fields dict
+        for permutation_choices in parser_args.fields:
+            for i in range(0, num_field_args):
```
brittneyexline: this would be more readable as `for field in parser_args.fields` - python makes it easy to not have to use an iterator for going through lists.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+                try:
+                    field_args[permutation_choices[i]] = permutation_data[permutation_choices[i]]
+                except KeyError:
+                    print "{} is not a field option".format(sys.exc_value)
```
brittneyexline: maybe we should exit in this case?

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+
+        # the difference btwn all possible fields and the chosen permutation ones
```
brittneyexline: please use full words instead of abbreviations, unless it is a well known acronym like URL. also maybe say "calculate the difference..." just so it's clear what the action is with regard to the subject.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+        default_fields = list(set(default_data_keys) - set(parser_args.fields[0]))
+
+        # add non permutation fields to dict
+        for field in default_fields:
+            field_args[field] = default_data[field]
```
brittneyexline: so it seems like field_args has inconsistent value types - above you are setting each key to be the field and the value to a list, and here it seems to be set to a single value. you should probably wrap this in a list for consistency, especially given how you are iterating on lists below.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+
+    wrapper_courses_dict = {}  # needed to match course input file creation
+    wrapper_courses_dict["courses"] = results
+
+    with open(test_courses_file, "w") as outfile:
+        json.dump(wrapper_courses_dict, outfile)
+
+
+def set_course_defaults(courses_dict):
+    courses_dict["number"] = course_number  # will be generated automatically by course creation script
```
brittneyexline: put this comment above the line instead of inline.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+    courses_dict["organization"] = course_organization
+    courses_dict["run"] = course_run
+    courses_dict["user"] = course_user
+    courses_dict["partner"] = course_partner
+    # configure enrollment seat settings
+    enrollment_dict = {}
```
brittneyexline: it seems like you are doing two separate things - populating the courses_dict and creating an enrollment_dict. i would separate these into their own functions because they have nothing to do with each other (that i can tell - if they do it should be made clear).

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+    enrollment_dict["credit"] = False
+    enrollment_dict["credit_provider"] = "test-credit-provider"
+    return enrollment_dict
+
+
+def generate_date_translation(field_values_dict, permutation_option, permutation_value):
+    now = datetime.datetime.now(pytz.UTC)
+    if permutation_value == future_date:
+        future = str(now + datetime.timedelta(days=365))
```
brittneyexline: i think this isn't the right scope for this function - we should be checking the permutation options to see if it is one that expects a date format above, then limit the scope of this function to calculating future or past values.

In the call site above, you'd do something like:
```
if permutation_option in list of fields that require dates:
    permutation_value = calculate_date_value(permutation_value)
field_values_dict[permutation_option] = permutation_value
```

then here it would be:
```
def calculate_date_value(date_const):
    # logic for calculating future or past based on date_const value. if it's invalid, throw an error.
```

util/course-permutation-tool/generate_permutations.py
brittneyexline: i can't tell - does this function need to get called multiple times, or is it a one time set up? if it is only one time, it doesn't have to be its own function, but it should be clear that this set up is necessary for the expected format.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+        generate_date_translation(field_values_dict, permutation_option, permutation_value)
+
+        for permutation_key in all_permutations_keys:
+            # add audit and verify fields to dict
+            if permutation_key == "audit" and permutation_value == True:
+                enrollment_dict["audit"] = permutation_value
+            if permutation_key == "verify" and permutation_value == True:
+                enrollment_dict["verify"] = True
+
+        if index + 1 < len(all_permutations_keys):
```
brittneyexline: if your recursion is just incrementing and index, it seems like you could accomplish this more simply with a loop over the permutation keys...

util/course-permutation-tool/generate_permutations.py
```diff
@@ -0,0 +1,145 @@
...
+            generate_permutations(field_args, index + 1, results, courses_dict, field_values_dict)
+
+        courses_dict["enrollment"] = enrollment_dict
+        courses_dict["fields"] = field_values_dict.copy()
+        results.append(courses_dict.copy())
+
+    wrapper_courses_dict = {}  # needed to match course input file creation
+    wrapper_courses_dict["courses"] = results
+
+    with open(test_courses_file, "w") as outfile:
```
brittneyexline: this will be easier to write tests for if you separate out writing the data to a file into its own function, and just have this return the generated json.

brittneyexline (commented):
This is getting closer, but still needs some cleaning up. Now that you have the bulk of your logic in one function (generate_permutations), you can write tests that take different sample inputs, call the function with them, and compare against expected outputs. See the first few tests here (https://github.com/edx/edx-enterprise/blob/master/tests/test_sap_success_factors/test_sap_successfactors.py) for an example of how you can do this.

util/course-permutation-tool/generate_permutations.py
shrlyhe: I think I encountered this problem earlier, and the way that I fixed it was by wrapping each ```default_data``` field in permutations.json in a list to make it consistent with the way ```permutation_data``` is formatted

util/course-permutation-tool/generate_permutations.py
```diff
@@ -35,39 +35,40 @@ def arg_parse():
...
     except IOError:
-        print sys.exc_value
+        print("Please fix your input file - {}".format(sys.exc_value))
         sys.exit()
     except KeyError:
         print("Could not find key {}".format(sys.exc_value))
         sys.exit()
 
     field_args = {}
+    parsed_field_args = parser_args.fields
```
brittneyexline: since you only care about the first element, you could do:
`parsed_field_args = parser_args.fields[0] if isinstance(parser_args.fields, list) else None`
and then not have to repeatedly access the first index below.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -82,12 +83,25 @@ def generate_permutations(field_args, index, results, courses_dict, field_values
...
-        enrollment_dict = set_course_defaults(courses_dict)
+        # add other required course and enrollment default fields to dict
+        # will be generated automatically by course creation script
+        courses_dict["number"] = COURSE_NUMBER
+        courses_dict["organization"] = COURSE_ORGANIZATION
+        courses_dict["run"] = COURSE_RUN
+        courses_dict["user"] = COURSE_USER
+        courses_dict["partner"] = COURSE_PARTNER
+        # configure enrollment seat settings
+        enrollment_dict = {}
```
brittneyexline: this could be done as:
```
enrollment_dict = {
    'credit': False,
    'credit_provider': "test-credit-provider",
}

File: util/course-permutation-tool/generate_permutations.py [hunk 32]
```diff
@@ -82,12 +83,25 @@ def generate_permutations(field_args, index, results, courses_dict, field_values
...
+        enrollment_dict["credit"] = False
+        enrollment_dict["credit_provider"] = "test-credit-provider"
+
         # add permutation fields to dict
         field_values_dict[permutation_option] = permutation_value
+
         # generate start and end dates
-        generate_date_translation(field_values_dict, permutation_option, permutation_value)
+        if permutation_option in ["start", "end"]:
```
brittneyexline: we should do this based on the field, what would happen if i passed in "start" as the title field? as a user i wouldn't expect a random date to appear instead for the title.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -103,36 +117,31 @@ def generate_permutations(field_args, index, results, courses_dict, field_values
...
         results.append(courses_dict.copy())
 
-    wrapper_courses_dict = {}  # needed to match course input file creation
+    # needed to match course input file creation
+    wrapper_courses_dict = {}
     wrapper_courses_dict["courses"] = results
 
-    with open(test_courses_file, "w") as outfile:
-        json.dump(wrapper_courses_dict, outfile)
+    create_courses_json_file(wrapper_courses_dict)
```
brittneyexline: this should get called from the top level, and this function should return the json that should be passed into this function.

util/course-permutation-tool/generate_permutations.py
```diff
@@ -103,36 +117,31 @@ def generate_permutations(field_args, index, results, courses_dict, field_values
...
+        if date_const == FUTURE_DATE:
+            future = str(now + datetime.timedelta(days=365))
+            return future
+        if date_const == PAST_DATE:
+            past = str(now - datetime.timedelta(days=60))
+            return past
+        if date_const == None:
+            return None
+    except ValueError:
+        print "Dates can only be future or past"
```
brittneyexline: maybe we should exit the script here? and format it the same way as the other exit points above?

edx-webhook:
Thanks for the pull request, @shrlyhe! I've created [OSPR-2086](https://openedx.atlassian.net/browse/OSPR-2086) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

nedbat:
@brittneyexline Should we close this?

brittneyexline:
yea @nedbat, i think we can close this.

PR #4043, opened 2017-08-14
Fixed issue with ecommerce JWT issuers

clintonb:
We should be continue using the role-specific variable here to allow
for JWTs signed by ecomworker.

LEARNER-2235


feanil (approved):

mjfrey (approved):

schenedx (approved):
👍

PR #4045, opened 2017-08-14
add additional CD role path

fredsmith:


feanil:
@fredsmith can you explain why you need the additional role path(update commit message)?  Is it because you're running the pipeline from a different playbook root?  I thought ansible only consumed the config from the folder where you run ansible-playbook.  This PR seems to suggest otherwise.

fredsmith:
@feanil - the pipeline we're building runs ansible after cding to this directory. According to http://docs.ansible.com/ansible/latest/intro_configuration.html#configuration-file, ansible.cfg is consumed from the cwd first. The role path needs to include the roles that are in configuration.

feanil (approved):
I'm fine with this change but how do all the other pipelines work without this?  Do they not `cd` into this directory to run this?  If not, why not?

PR #4046, opened 2017-08-16
change test specifiers for move to pytest

estute:
Packer builds were failing since the platform switched from nose to pytest for bokchoy tests. This fixes the problem

CC @edx/devops


estute:
Build @ https://build.testeng.edx.org/view/packer/job/build-packer-ami/7023

jzoldak (approved):

PR #4048, opened 2017-08-16
Mongo 3.2 role

jdmulloy:
Configuration Pull Request
---


jdmulloy:
Looking at individual commits may be helpful since I started by copying the 3.0 role and then making changes. The combined view in the PR shows everything as an add, so there's no way to see the changes from the 3.0 role.

jdmulloy:
This will give you a proper diff of changes since I copied the 3.0 role.

https://github.com/edx/configuration/compare/1411e5c2a6fbcdc53625f04f9d6d6ed3a1b79a85...jdmulloy/ops2205/mongo_3.2

playbooks/test-mongo.yml
```diff
@@ -0,0 +1,88 @@
+# ansible-playbook -i '54.164.134.57,' -u ubuntu edx-east/mongo_3_2.yml -e@test-mongo.yml
```
feanil: Move this config to https://github.com/edx/configuration/tree/master/playbooks/sample_vars

playbooks/test-mongo.yml
```diff
@@ -0,0 +1,88 @@
...
+
+##### edx-secure/ansible/vars/stage-edx.yml #####
+MONGO_ADMIN_USER: 'admin'
+MONGO_ADMIN_PASSWORD: '794jtB7zLIvDjHGu2gD6wKUU'
+MONGO_MONITOR_USER: 'cloud-manager'
+MONGO_MONITOR_PASSWORD: '7DJ9FTWHJx4TCSPxSmx1k3DD'
+MONGO_BACKUP_USER: 'backup'
+MONGO_BACKUP_PASSWORD: 'XbJA3LouKV5QDv2NQixnOrQj'
+MONGO_CLUSTERED: true
+MONGO_REPL_SET: 'jdmulloy-3.2'
```
feanil: Replace your name with a generic repl_set name.  Maybe `test-repl-set`

playbooks/test-mongo.yml
```diff
@@ -0,0 +1,88 @@
...
+    password: "{{ MONGO_MONITOR_PASSWORD }}"
+    database: "admin"
+    roles: clusterMonitor
+  - user: "{{ MONGO_BACKUP_USER }}"
+    password: "{{ MONGO_BACKUP_PASSWORD }}"
+    database: "admin"
+    roles: backup
+
+##### edx-internal/ansible/vars/stage-edx.yml #####
+EDXAPP_MONGO_DB_NAME: 'jdmulloy-32'
```
feanil: Use a generic DB name.  Maybe `test-mongo-db`

playbooks/test-mongo.yml
```diff
@@ -0,0 +1,88 @@
...
+EDXAPP_MONGO_PORT: 27017
+EDXAPP_MONGO_REPLICA_SET: 'jdmulloy-3.2'
```
feanil: Should refere to the `MONGO_REPL_SET` variable right?

playbooks/test-mongo.yml
feanil: It may also be good to add a comment about how all the secrets in here are examples and should all be replaced.  Maybe change the secrets to have values that obviously need to be replaced.  eg. "PLEASE-SET-ME"

feanil (commented):
Generally looks good.

playbooks/test-mongo.yml
jdmulloy: I can make it more generic. I wasn't intending to leave it like this. I just checked it in so the work could be easily handed off.

playbooks/test-mongo.yml
jdmulloy: Maybe? I copied variable from the real configs until it worked. I don't understand enough about why the real configs are the way they are.

jdmulloy: Maybe we should at some point collapse these variables to one in the real configs, but that's a risky surgery I can't do right now.

jibsheet: EDXAPP_MONGO_REPLICA_SET is an internal variable we use.  We set it in edx-internal to a name based on deployment/environment and then it's consumed into the code that makes users (custom to edx) and our split modulestore configuration.

It comes from the confused time where we had 2 repl sets, one for edxapp one for forums.  Now we (edX) have one repl set with two databases.

We can dedupe it in our config, but out of scope here.  Now that I'm aware of it, I may try to do it while moving secure -> internal

@jdmulloy you shouldn't need it, I'll look at what the rest of this file contains when I get here for real.

playbooks/sample_vars/test-mongo.yml
```diff
@@ -0,0 +1,110 @@
...
+
+##### edx-secure/ansible/vars/stage-edx.yml #####
+MONGO_ADMIN_USER: 'admin'
+MONGO_ADMIN_PASSWORD: 'CHANGEME_794jtB7zLIvDjHGu2gD6wKUU'
+MONGO_MONITOR_USER: 'cloud-manager'
+MONGO_MONITOR_PASSWORD: 'CHANGEME_7DJ9FTWHJx4TCSPxSmx1k3DD'
+MONGO_BACKUP_USER: 'backup'
+MONGO_BACKUP_PASSWORD: 'CHANGEME_XbJA3LouKV5QDv2NQixnOrQj'
+MONGO_REPL_SET: 'test-repl-set'
+mongo_bind_ip: 0.0.0.0
```
jibsheet: This isn't the variable we use anymore.  I believe it worked by accident (because in cluster mode we used Mongo's bind to 0.0.0.0 default), strike it from your sample/test data.
We'll want to change the secure/internal config to use the correct uppercase var.

playbooks/sample_vars/test-mongo.yml
```diff
@@ -0,0 +1,110 @@
...
+MONGO_STORAGE_ENGINE_OPTIONS:
+  wiredTiger:
+    engineConfig:
+      cacheSizeGB: "{{ (( ansible_memtotal_mb / 1024 ) * 0.70 ) | round(1,'ceil') | int }}" # 70% of available ram - 11GB on m4.xlarge
+EDXAPP_MONGO_HOSTS: "{{ MONGO_RS_CONFIG.members|map(attribute='host')|list }}"
+
+##### Testing
+#UNMOUNT_DISKS: true
+
+##### edx-olive/olive-secure/ansible/vars/prod-olivex.yml
```
jibsheet: This is a public repo - I'd rather not have repo paths in it, particularly not repo paths of things we are handing off to another team.

playbooks/sample_vars/test-mongo.yml
```diff
@@ -0,0 +1,110 @@
...
+# Passwords and relication keys in this file are examples and must be changed.
+# You must change any variable with the string "CHANGEME" in it
+
+##### edx-secure/ansible/vars/edx.yml #####
+MONGO_HEARTBEAT_TIMEOUT_SECS: 3
+MONGO_STORAGE_ENGINE: 'wiredTiger'
+MONGO_STORAGE_ENGINE_OPTIONS:
+  wiredTiger:
+    engineConfig:
+      cacheSizeGB: "{{ (( ansible_memtotal_mb / 1024 ) * 0.70 ) | round(1,'ceil') | int }}" # 70% of available ram - 11GB on m4.xlarge
```
jibsheet: Please test without the STORAGE_ENGINE and  STORAGE_ENGINE_OPTIONS override.  Mongo 3.2 uses wiredTiger by default, it should come up properly.  We're also going to test dropping the cacheSizeGB because of some memory issues we've seen.

playbooks/roles/mongo_3_2/defaults/main.yml
```diff
@@ -0,0 +1,73 @@
...
+MONGO_BIND_IP: 127.0.0.1
+MONGO_REPL_SET: "rs0"
+MONGO_AUTH: true
+
+# Cluster member configuration
+# Fed directly into mongodb_replica_set module
+MONGO_RS_CONFIG:
+  members: []
+
+# Storage engine options in 3.0: "mmapv1" or "wiredTiger"
```
jibsheet: Update this comment, this is the 3.2 play (same comment the line below).  mmapv1 and WT are supported in both versions.

playbooks/roles/mongo_3_2/defaults/main.yml
```diff
@@ -0,0 +1,73 @@
...
+  - user: cs_comments_service
+    password: password
+    database: cs_comments_service
+    roles: readWrite
+  - user: edxapp
+    password: password
+    database: edxapp
+    roles: readWrite
+
+MONGO_BIND_IP: 127.0.0.1
```
jibsheet: You removed the use of this in the mongod.conf.j2 however we need to support this.  For single instance edxapp deploys (like sandboxes) we should not be listening on 0.0.0.0, but that is what we now do. 

Please see https://github.com/edx/configuration/pull/3594/files for some sense of the bugfix we had to do because people were running wide open mongods listening on 0.0.0.0.  This was a regression from the mongo installation config file which has set it to 127.0.0.1 for a while https://docs.mongodb.com/manual/reference/configuration-options/#net.bindIp

I think we want to leave it as 127.0.0.1 and add documentation saying "if you're running a cluster, you must set this to 0.0.0.0 or some other variable to allow mongo to talk to other cluster members".  You'll also need to re-add support to the config file.

playbooks/edx-east/mongo_3_2.yml
```diff
@@ -0,0 +1,28 @@
...
+#
+# ADDING A NEW CLUSTER MEMBER
+# If you are adding a member to a cluster, you must be sure that the new machine is not first in your inventory
+# ansible-playbook mongo_3_2.yml -i 10.1.1.1,10.2.2.2,new-machine-ip -e@/path/to/edx.yml -e@/path/to/ed.yml
+- name: Deploy MongoDB
+  hosts: all
+  become: True
+  gather_facts: True
+  roles:
+    - aws
```
jibsheet, line 18: Note for later - we should figure out how to make the AWS part optional (not part of this PR)

playbooks/roles/mongo_3_2/defaults/main.yml
jibsheet: See my later comment where I mention that you copied config that says mongo_bind_ip, but the variable is MONGO_BIND_IP, you should PR that change since it's wrong for 3.0 as well as 3.2.

playbooks/roles/mongo_3_2/files/disable-transparent-hugepages.conf
```diff
@@ -0,0 +1,9 @@
+description     "set transparent hugepage to never"
```
jibsheet: We don't use this anymore, right?  Can we delete it

playbooks/roles/mongo_3_2/defaults/main.yml
```diff
@@ -0,0 +1,73 @@
+mongo_logappend: true
+
+#This way, when mongod receives a SIGUSR1, it'll close and reopen its log file handle
+mongo_logrotate: reopen
+
+mongo_version: 3.2.16
+mongo_port: "27017"
+mongo_extra_conf: ''
+mongo_key_file: '/etc/mongodb_key'
+pymongo_version: 3.2.2
```
jibsheet, line 10: We have a different version in our requirements.txt, can we normalize them

playbooks/roles/mongo_3_2/tasks/main.yml
```diff
@@ -0,0 +1,329 @@
...
+    password: "{{ MONGO_ADMIN_PASSWORD }}"
+    database: admin
+    roles: root
+  when: initialize_replica_set
+  run_once: true
+  tags:
+    - "manage"
+    - "manage:db"
+
+- name: create super user
```
jibsheet: In retrospect, this task seems superfluous.
You either create the ADMIN_USER above when initialize_replica_set or you already have an admin_user.  I think the intent was to log in as MONGO_ADMIN_USER and make sure MONGO_ADMIN_USER was set, but that doesn't actually do anything.  If you want to use it to change the password, you can't even do that here.  

I suggest deleting this task (the when: not initialize_replica_set) unless you saw some use for it that I'm missing.

playbooks/roles/mongo_3_2/templates/create_root.js.j2
```diff
@@ -0,0 +1,12 @@
+// Add super user
```
jibsheet: Do we use this?  I think this is a holdover from the 2.6 role.  It's probably unused in the 3.0 role also.

playbooks/roles/mongo_3_2/templates/mongod.conf.j2
```diff
@@ -0,0 +1,45 @@
+# Do not edit this file directly, it was generated by ansible
```
jibsheet: Switch this to {{ ansible_managed }}

jibsheet (changes requested):
Looks great, one important change around the bind ip and a bunch of small tweaks (mostly deleting things that shouldn't have been in the 3.0 role to start or other cleanups we can do)

playbooks/sample_vars/test-mongo.yml
jdmulloy: That makes sense. I was just trying to keep track of where I got sample data from. Is there some way I can express the source without leaking sensitive info? It gets a bit confusing having info spread out across multiple repos.

playbooks/sample_vars/test-mongo.yml
jdmulloy: I can strike it from the test data that is checked in but I'll still need to set it when I use the file for my testing right?

playbooks/roles/mongo_3_2/files/disable-transparent-hugepages.conf
jdmulloy: Mongo still gives a warning if we don't disable THP.

playbooks/roles/mongo_3_2/defaults/main.yml
jdmulloy: Should I bring the one in requirements.txt up to 3.2 then? In the 3_0 role it was very old and didn't work with 3.2. I'm guessing it's backwards compatible.

jibsheet:
@jdmulloy I just close/opened to attempt to get it into travis' build queue now that github seems marginally healthier

playbooks/roles/mongo_3_2/files/disable-transparent-hugepages.conf
jibsheet: But this is the upstart file, right?  You replaced it with the systemd version, and we don't expect to run this role on anything other than 16.04 and higher?

playbooks/roles/mongo_3_2/defaults/main.yml
jibsheet: I think you've brought up requirements.txt, which is the right fix in my mind

playbooks/sample_vars/test-mongo.yml
jibsheet: Unfortunately, not a great way.  I'll often just say secure vs internal and leave it at that.

playbooks/sample_vars/test-mongo.yml
jibsheet: You'll want to replace it with MONGO_BIND_IP (in your test data)

playbooks/roles/mongo_3_2/files/disable-transparent-hugepages.conf
jdmulloy: Correct. I will remove this file.

playbooks/roles/mongo_3_2/defaults/main.yml
jibsheet: Can you document here that if you're clustering more than one machine with our play, MONGO_BIND_IP should be set to either 0.0.0.0 to listen on all IPs, or set to more specific IPs as needed for their configuration.

jibsheet (approved):
Please add a note about the BIND IP called out in the review comment, add a CHANGELOG entry (mentioning new 3.2 role and the removal of MONGO_CLUSTERED) then squash (As makes sense) and merge.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+- Role: mongo_3_2
+  - Added role for mongo 3.2, not yet in use.
+  - Removed MONGO_CLUSTED variable. In this role mongo replication is always configured, even if there is only one node.
```
jibsheet: CLUSTERED

PR #4052, opened 2017-08-20
Create ansible role that deploys a custom theme

clemente:
This PR includes an ansible role that creates a new theme on deploy. The theme is based on a skeleton theme downloaded from a repository, and adds custom SASS overrides and custom static files on top. Custom files can come from a local directory or from a URL.

**JIRA tickets**: https://openedx.atlassian.net/browse/OSPR-1896

**Discussions**: to be added

**Dependencies**: None

**Screenshots**:

**Sandbox URL**: None

**Partner information**: None

**Deployment targets**: None that I know of

**Merge deadline**: None

**Testing instructions**:

1. In the vagrant VM, enable comprehensive theming in lms.env.json, with theme directory /edx/var/edxapp/themes
1. In the vagrant VM, enter into /edx/app/edx_ansible/edx_ansible and checkout the branch from this PR
2. As "vagrant" user:
```
    export PYTHONUNBUFFERED=1
    source /edx/app/edx_ansible/venvs/edx_ansible/bin/activate
    cd /edx/app/edx_ansible/edx_ansible/playbooks
     # Here comes the part you'll want to customize (this line is in README.rst too):
    ansible-playbook -i localhost, -c local run_role.yml -e role=simple_theme  -e configuration_version=master -e edx_platform_version=master -e EDXAPP_DEFAULT_SITE_THEME=mytheme2 -e '{"SIMPLETHEME_SASS_OVERRIDES": [{"variable": "main-color", "value":"#823456"}, {"variable": "action-primary-bg", "value":"$main-color"}]}' -e EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO="https://github.com/open-craft/edx-theme/" -e EDXAPP_COMPREHENSIVE_THEME_VERSION="harvard-dcex" -e edxapp_user=vagrant -e common_web_group=www-data -e SIMPLETHEME_ENABLE_DEPLOY=true -e '{"SIMPLETHEME_STATIC_FILES_URLS": [{"url": "http://docs.ansible.com/ansible/latest/_static/images/logo_invert.png", "dest":"lms/static/images/logo.png"}, {"url": "http://docs.ansible.com/favicon.ico", "dest":"lms/static/images/favicon.ico"}]}' -e '{"EDXAPP_COMPREHENSIVE_THEME_DIRS":["/edx/var/edxapp/themes"], "EDXAPP_ENABLE_COMPREHENSIVE_THEMING": true}'
```
3. This will create a new theme in the themes directory. It will have the provided values inside. No other theme will be modified.
4. Add the theme through admin. Do update_assets as usual or restart LMS to compile the theme
5. You should see the theme with the changes you requested. It will be based on the skeleton theme you provided in `EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO` and it will have SASS changes on top of that. It will have the logo and favicon from the URLs you passed. You can remove the theme and run again with different values.
6. Try different values, e.g. a new name instead of mytheme2, and don't pass `EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO`/`…_VERSION`.
7. If possible, try to do a full deployment by running the edxapp role against an unprovisioned machine, and verify that after provisioning, it runs openedx with the new theme enabled. 

**Author notes and concerns**:

1. The main use case is to run as part of edxapp role, i.e. to use it when provisioning a new machine. The instructions above are a workaround to test it independently without having to redeploy each time.
2. I'm using EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO to download a base theme. These variables were already present in edxapp for exactly this functionality, but were unused yet.

**Reviewers**
- [ ] @smarnach 
- [ ] @mtyaka 
- [ ] edX reviewer[s] TBD

**Settings**
In defaults/main.yml you'll find the available values, but you must your new values through ansible values.

---


openedx-webhooks:
Thanks for the pull request, @clemente! It looks like you're a member of a company that does contract work for edX. If you're doing this work as part of a paid contract with edX, you should talk to edX about who will review this pull request. If this work is not part of a paid contract with edX, then you should ensure that there is an OSPR issue to track this work in [JIRA](https://openedx.atlassian.net), so that we don't lose track of your pull request.

<a   href="https://openedx-webhooks.herokuapp.com/github/process_pr?number=4052&repo=edx%2Fconfiguration"   rel="nofollow" >Create an OSPR issue for this pull request</a>.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+
+# Default variables for the theme which will be deployed
+# This file contains example variables and colors that you'll want to change
+
+# Default
+themes_dir: "/edx/app/edxapp/themes"
```
smarnach: Isn't there already a variable for this?

If you introduce new variable, they should be prefixed with the role name.  See the [Ansible code conventions](https://openedx.atlassian.net/wiki/spaces/OpenOPS/pages/26837527/Ansible+Code+Conventions) for details.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+# Choose the theme name
+theme_name: "simplet1"
+
+# This role requires ENABLE_COMPREHENSIVE_THEMING, therefore it uses this pattern:
+theme_folder: "{{ themes_dir}}/edx-platform/{{ theme_name }}"
+
+
+# Force the use of comprehensive theming when this role gets included
+EDXAPP_COMPREHENSIVE_THEME_DIRS:
+  - "/edx/app/edxapp/themes/edx-platform"
```
smarnach: I'm actually not sure how the variable precedence works out if we override this here, in particular given that we don't know yet whether the role is run by `include_role`.  It's possible we need to move these overrides to `vars/default.yml` after all, to give them high enough priority.  If we do so, however, we need to make sure that the overrides only take effect if we actually want them.  Ansible is a bit special in this regard – even if a role is disabled by a `when` clause, all it's variables are set anyway.

So if we can't find a reasonable way to make the overrides here, we need to document what variables need to be set to use this role instead.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+  - variable: main-color
+    value: '#123456'
+  - variable: action-primary-bg
+    value: '$main-color'
+  - variable: action-primary-fg
+    value: '#fff'
+  - variable: link-color
+    value: 'red'
+  - variable: action-secondary-bg
+    value: '#07f'
```
smarnach: The default should be something neutral; probably just an empty list.

playbooks/roles/simple_theme/tasks/main.yml
```diff
@@ -0,0 +1,56 @@
...
+- name: Create directory to hold the theme
+  file: path={{ theme_folder }} state=directory
+- name: Create directory to hold SCSS
+  file: path={{ theme_folder }}/lms/static/sass/partials/base state=directory
+- name: Create directory to hold HTML
+  file: path={{ theme_folder }}/lms/templates/static_templates state=directory
+- name: Create directory to hold images
+  file: path={{ theme_folder }}/lms/static/images state=directory
+- name: Create directory to hold images
+  file: path={{ theme_folder }}/lms/static/images/profiles state=directory
```
smarnach: It looks like all these directories are created as the root user, which doesn't seem like a good idea.

You should also use the newer dictionary syntax for module parameters:

    - name: Create dir
      file:
        path: /some/path
        state: directory

See the coding conventions I linked before.

playbooks/roles/simple_theme/templates/lms/static/sass/_lms-overrides.scss.j2
```diff
@@ -0,0 +1,3 @@
+/* Extra SASS as defined by simple_theme: */
```
smarnach: This does not seem like a useful template, since it simply expands to what is in `EDXAPP_THEME_EXTRA_SASS`.  You can use the `copy` role with the `content` parameter to achieve the same purpose.  (If you prefer it the way it is for consistency and flexibility, this is also fine with me.  So up to you.)

smarnach (changes requested):

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+# This file contains example variables and colors that you'll want to change
+
+# Default
+themes_dir: "/edx/app/edxapp/themes"
+
+# Choose the theme name
+theme_name: "simplet1"
+
+# This role requires ENABLE_COMPREHENSIVE_THEMING, therefore it uses this pattern:
+theme_folder: "{{ themes_dir}}/edx-platform/{{ theme_name }}"
```
mtyaka: Themes are usually located in `"/edx/app/edxapp/{{ theme_name }}"`. I don't think the `"/edx-platform/"` in the middle is necessary.

playbooks/roles/simple_theme/defaults/main.yml
mtyaka: Should this be `- "{{ theme_folder }}"` ?

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+#
+
+# Default variables for the theme which will be deployed
+# This file contains example variables and colors that you'll want to change
+
+# Default
+themes_dir: "/edx/app/edxapp/themes"
+
+# Choose the theme name
+theme_name: "simplet1"
```
mtyaka: Can we change the default to something more readable? Perhaps `"simple-theme"`?

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+# This fragment will be inserted in _lms-overrides and will affect all pages
+EDXAPP_THEME_EXTRA_SASS: |
+  .header-global h1.logo a img {
+    height: 50px;
+  }
+  .header-global.slim h2 {
+    width: 60% !important;
+  }
+  .wrapper-footer {
+      border-top: 3px solid $main-color;
```
mtyaka: Nit: indentation is not consistent with the rules above.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    value: '$main-color'
+  - variable: action-primary-fg
+    value: '#fff'
+  - variable: link-color
+    value: 'red'
+  - variable: action-secondary-bg
+    value: '#07f'
+
+
+# Files will be copied from "files" directory in this role to the specified path.
```
mtyaka: Custom files should not be included in the role itself (except as a default example). There should be an ansible variable that lets you specify the folder where static files are located.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+# Directories must be created; there's a task for that in tasks/main.yml
+EDXAPP_THEME_STATIC_FILES:
+  logo.png: '/lms/static/images/logo.png'
+  favicon.ico: '/lms/static/images/favicon.ico'
+
+# Files will be downloaded when the role runs, and copied to the path given in
+# the second value. Directories must be created; there's a task for that
+# in tasks/main.yml
+EDXAPP_THEME_FROM_REMOTE_TO_STATIC_FILES:
+  'https://clix.tiss.edu/wp-content/uploads/2015/11/icon-student-1.png': '/lms/static/images/profiles/default_50.png'
```
mtyaka: I don't think we will need this. I should have removed that from the task description, sorry.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+    value: 'red'
+  - variable: action-secondary-bg
+    value: '#07f'
+
+
+# Files will be copied from "files" directory in this role to the specified path.
+# Directories must be created; there's a task for that in tasks/main.yml
+EDXAPP_THEME_STATIC_FILES:
+  logo.png: '/lms/static/images/logo.png'
+  favicon.ico: '/lms/static/images/favicon.ico'
```
mtyaka: I think we can simplify this. Instead of specifying the source and destination of each file, we could specify a folder that contains all custom static files, and just copy them all to the appropriate location in the theme folder, maintaining the folder hierarchy. Sorry for not thinking of this when I created the task.

For example source folder could look like this:

```
- lms
  - images
    - logo.png
    - favicon.ico
  - js
    - myscript.js
- cms
  - images
    - logo.png
```

would copy the files into the static folders in the theme:

```
- lms
  - static
    - images
      - logo.png
      - favicon.ico
    - js
      - myscript.js
- cms
  - static
    - images
      - logo.png
```

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+# in tasks/main.yml
+EDXAPP_THEME_FROM_REMOTE_TO_STATIC_FILES:
+  'https://clix.tiss.edu/wp-content/uploads/2015/11/icon-student-1.png': '/lms/static/images/profiles/default_50.png'
+
+# These values can be used in Jinja templates
+EDXAPP_THEME_CUSTOM_VARIABLES:
+  hide_registration_link: true
+  terms_of_service: |
+    <p>The contents
+    of my ToS page</p>
```
mtyaka: I think I would prefer if all customizations were empty by default, but code comments containing examples like these will be appreciated.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,80 @@
...
+
+# This fragment will be inserted in _lms-overrides and will affect all pages
+EDXAPP_THEME_EXTRA_SASS: |
```
mtyaka: Same here - there should probably be no extra sass by default, but it would be nice if we kept this example in a code comment.

playbooks/roles/simple_theme/templates/lms/templates/static_templates/tos.html.j2
```diff
@@ -0,0 +1,11 @@
+<%! from django.utils.translation import ugettext as _ %>
+<%inherit file="../main.html" />
+<%block name="pagetitle">${_("Terms of Service")}</%block>
+<section class="container about">
+
+<h1>${_("Terms of Service")}</h1>
+<!-- <h2>On April 1, 2017, adopted Terms of Service, providing as follows:</h2> -->
```
mtyaka: Remove this line.

playbooks/roles/simple_theme/templates/lms/templates/static_templates/tos.html.j2
```diff
@@ -0,0 +1,11 @@
...
+
+{{ EDXAPP_THEME_CUSTOM_VARIABLES['terms_of_service'] }}
+
+</section>
```
mtyaka: We will eventually want to support other static pages (about, contact, privacy policy...), but I think this is fine for first implementation.

mtyaka (commented):
Thanks for working on this @clemente, it's looking very nice.

The main thing missing is ability to specify a different theme template/skeleton, where you can use your own custom theme as the base.

playbooks/roles/simple_theme/defaults/main.yml
clemente: They recommend it in here (and in following pages): http://edx.readthedocs.io/projects/edx-installing-configuring-and-running/en/latest/configuration/changing_appearance/theming/theme_directories.html

smarnach: I think the path should be `/edx/app/edxapp/themes/{{theme _name}}`, which is what we are generally using.  I can't see how the additional layer with `edx-platform` would make sense, since the `ecommerce` theme would go in `/edx/app/ecommerce/themes` anyway.

playbooks/roles/simple_theme/defaults/main.yml
clemente: I implemented this. The part about adding the „static/“ to every path was the hardest (not because of the reg.expr. needed, but because i can't directly copy everything, instead I must process every folder to do the path change). Ansible is not bash and I might have resorted to a strange system; but it works. Any idea of how could have been easier?

playbooks/roles/simple_theme/templates/lms/static/sass/lms-course.scss.j2
```diff
@@ -0,0 +1,2 @@
+@import 'lms/static/sass/lms-course';
+@import 'lms-overrides';
```
mtyaka: This file, as well as `lms-main-v1.scss` and `lms-main-v2.scss` don't have to be jinja templates.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,95 @@
...
+# Set EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO: "https://github.com/open-craft/edx-theme/"
+# and EDXAPP_COMPREHENSIVE_THEME_VERSION: "harvard-dcex"
+#
+
+
+# This variable holds the main path where the role will copy files.
+# EDXAPP_COMPREHENSIVE_THEME_DIRS.0 is usually "/edx/app/edxapp/themes"
+# or "/edx/app/edxapp/themes/edx-platform".
+# If you have more than 1 theme dirs, you'll need to override this internal variable
+theme_folder: "{{ EDXAPP_COMPREHENSIVE_THEME_DIRS.0 }}/{{ EDXAPP_DEFAULT_SITE_THEME }}"
```
smarnach: All variables you introduce should be prefixed with the role name, or at least with a common prefix, so this should be `simpletheme_folder`.  Ansible does not have scopes, so we need to simulate scopes by prefixes.

playbooks/roles/simple_theme/meta/main.yml
```diff
@@ -0,0 +1,15 @@
...
+#
+# edX Configuration
+#
+# github:     https://github.com/edx/configuration
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+
+# At the moment the simple_theme doesn't depend on other roles
```
smarnach: Just make this

    dependencies: []

This says the same as the comment. just more succinctly.  There are also some cases where this file must evaluate to a dictionary (it might be only necessary for ansible-galaxy, but I don't quite remember), so it should not be completely empty.

playbooks/roles/simple_theme/defaults/main.yml
```diff
@@ -0,0 +1,95 @@
...
+# The theme will be deployed to a directory of that name.
+#
+# You have the option to use a skeleton theme. This is the base theme that will be
+# copied to the target machine, and modified afterwards via the customizations
+#  applied by this role's variables.
+#
+# Example: if you have a theme in https://github.com/open-craft/edx-theme/tree/harvard-dcex:
+# Set EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO: "https://github.com/open-craft/edx-theme/"
+# and EDXAPP_COMPREHENSIVE_THEME_VERSION: "harvard-dcex"
+#
```
smarnach: Thanks for writing the documentation!  I think it will be easier to find if you move it to a `README.md` file at the top level of the role.  Also include the documentation from `tasks/main.yml` in the same file.

playbooks/roles/simple_theme/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+# you can run the following ansible-playbook command. It requires you to pass
+# more variables manually because they're not available except when running
+# inside "edxapp" role. For instance you might need to pass edxapp_user (e.g.
+# "vagrant" if you test inside devstack).
+#
+# Example command to test this role, to be run from devstack, from "vagrant" user:
+# export PYTHONUNBUFFERED=1
+# source /edx/app/edx_ansible/venvs/edx_ansible/bin/activate
+# cd /edx/app/edx_ansible/edx_ansible/playbooks
+# ansible-playbook -i localhost, -c local run_role.yml -e role=simple_theme  -e configuration_version=master -e edx_platform_version=master -e EDXAPP_DEFAULT_SITE_THEME=mytheme2 -e '{"SIMPLETHEME_SASS_OVERRIDES": [{"variable": "main-color", "value":"#823456"}, {"variable": "action-primary-bg", "value":"$main-color"}]}' -e EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO="https://github.com/open-craft/edx-theme/" -e EDXAPP_COMPREHENSIVE_THEME_VERSION="harvard-dcex" -e edxapp_user=vagrant -e EDXAPP_SIMPLETHEME_DEPLOY=true -e '{"EDXAPP_COMPREHENSIVE_THEME_DIRS":["/edx/app/edxapp/themes"], "EDXAPP_ENABLE_COMPREHENSIVE_THEMING": true}'
```
smarnach: I'm not sure we need to document this.  This was more of a development tool, but it's not actually the intended use of this role.

playbooks/roles/simple_theme/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+    msg: "Simple-theme needs to know the name the name of the deployed them. Pass it in EDXAPP_DEFAULT_SITE_THEME"
+
+
+- name: Download skeleton theme
+  git:
+    repo: "{{ EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO }}"
+    dest: "{{ theme_folder }}"
+    version: "{{ EDXAPP_COMPREHENSIVE_THEME_VERSION | default('master') }}"
+    force: yes # overwrite local modifications (including the ones by this role)
+  when: EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO != ""
```
smarnach: What happens if `EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO` is not set?  We won't clone a skeleton repo, so will it make sense to continue this role?

playbooks/roles/simple_theme/tasks/main.yml
```diff
@@ -0,0 +1,106 @@
...
+  with_items:
+    - "."
+    - "lms/static/sass/partials/base"
+    - "lms/templates/static_templates"
+
+
+# Adapt to the list of files that define your theme
+- name: Compile the templates
+  template:
+    src: "./{{ item }}.j2"
```
smarnach: Any reason to prefix this with `./`?  I'm just curious.

smarnach (commented):
I just noticed that I have pending comments here since Monday – some of them are probably out of date by now.  I'll send the comments now and remove the ones that don't apply after sending, since the Github UI doesn't give me any better way to do this.

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+      - "EDXAPP_COMPREHENSIVE_THEME_DIRS | length > 0"
+      - "EDXAPP_ENABLE_COMPREHENSIVE_THEMING is defined"
+      - "EDXAPP_ENABLE_COMPREHENSIVE_THEMING"
+    msg: "Simple-theme deployment requires comprehensive theming to be enabled"
+
+- assert:
+    that:
+      - "EDXAPP_DEFAULT_SITE_THEME is defined"
+      - "EDXAPP_DEFAULT_SITE_THEME != ''"
+    msg: "Simple-theme needs to know the name the name of the deployed them. Pass it in EDXAPP_DEFAULT_SITE_THEME"
```
smarnach: Typo: duplicate "the name".

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+
+- name: Download skeleton theme
+  git:
+    repo: "{{ EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO }}"
+    dest: "{{ theme_folder }}"
+    version: "{{ EDXAPP_COMPREHENSIVE_THEME_VERSION | default('master') }}"
+    force: yes # overwrite local modifications (including the ones by this role)
+  when: EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO is defined and EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO != ""
+
+# Adapt to the list of directories that you need created
```
smarnach: If this is supposed to be user-defined, shouldn't we take the list of directories from a variables that can be configured?  Otherwise people would need to modify the role source.

If there is a default set of directories that almost always needs to be created, it's even better to make this _two_ variables:  One for the list of default directories, and one for the additional user-defined directories, so the user does not need to repeat the default directories when overriding the list.

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,108 @@
...
+    state: directory
+    owner: "{{ edxapp_user }}"
+    group: "{{ edxapp_user }}"
+  with_items:
+    - "."
+    - "lms/static/sass/partials/base"
+    - "lms/templates/static_templates"
+
+
+# Adapt to the list of files that define your theme
```
smarnach: Same here – if this is meant to be configurable, it needs to be made configurable.

smarnach (changes requested):
@clemente Here is another set of comments, in addition to the ones sent before.  Please also look at the comments mark as "outdated" by Github.  They are only marked as outdated since the main tasks file was renamed, so they are all still relevant.

playbooks/roles/simple_theme/tasks/main.yml
clemente: I have added more information and labelled it as „testing“.
But it's good that it's there. Whoever uses this role will need to test it in local for some time, and it's not good to hide the information about how to test it.
Also, I think a very common requirement will be to deploy theme changes to a running instance (as Xavier asked), so I wrote about this option too. We can discourage it more strongly, for now, but it actually works.

playbooks/roles/simple_theme/tasks/main.yml
clemente: Well yes, to make it fail if someone writes an absolute path there; that is, to show that you're expected to write relative paths.
I removed because it appeared only once in the file.
In fact I would like to force ansible not to allow accessing absolute paths in the host unless explicitly allowed. But I cannot control this yet.

playbooks/roles/simple_theme/tasks/deploy.yml
clemente: It started being a list of „files that compose your theme“, but now it's actually not needed anymore: you should provide your files through a skeleton theme, and these two tasks only process the internal files used by the simple_theme role, which are only 2 (_lms-overrides.scss and _variables.scss.j2) and you're not expected to change them. The skeleton themes should be built in such a way that they can have these 2 files overwritten with no problem (from @mtyaka 's proposal: https://tasks.opencraft.com/browse/OC-2685?focusedCommentId=44488&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-44488)
I rewrote the descriptions.

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,122 @@
...
+#
+# Simple theme
+#
+# See documentation in README.rst
+
+# Require comprehensive theming
+# EDXAPP_COMPREHENSIVE_THEME_DIRS.0 is usually "/edx/app/edxapp/themes"
+- assert:
+    that:
+      - "EDXAPP_COMPREHENSIVE_THEME_DIRS is defined"
```
smarnach: Are all the "is defined" tests necessary?  I have never needed one.

In general, we define default values for  all variables in `defaults/main.yml`, so they are guaranteed to exist.  This case is a bit special, since we are accessing a variable from the edxapp role.  However, since the simple_theme role is designed to only be run from the edxapp role, I think we can rely on the variables of that role, and let Ansible error out otherwise.  It will error out anywat if the assertion fails.

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,122 @@
...
+# deleted or redownloaded. It would be better to redownload, but for that we
+# need https://github.com/ansible/ansible-modules-core/issues/5292 to be fixed,
+# or to implement a workaround.
+- name: Download skeleton theme
+  git:
+    repo: "{{ EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO }}"
+    dest: "{{ simpletheme_folder }}"
+    version: "{{ EDXAPP_COMPREHENSIVE_THEME_VERSION | default('master') }}"
+    # force: yes  # Disabled due to ansible bug, see above
+  when: EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO is defined and EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO != "" and simpletheme_folder_stat.stat.exists == False
```
smarnach: I would omit the `is defined` test here as well (unless you disagree with the reasoning I gave).

I also dislike comparisons with Boolean values and I prefer `not xxx` over `xxx == False`, but that's more of a style nit, so up to you.

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,122 @@
...
+    when: item.state == 'directory'
+  - name: Create default skeleton (files)
+    copy:
+      src: "{{ item.src }}"
+      dest: "{{ simpletheme_folder }}/{{ item.path }}"
+      owner: "{{ edxapp_user }}"
+      group: "{{ common_web_group }}"
+    with_filetree: "../files/default_skeleton"
+    when: item.state != 'directory'
+  when: (EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO is not defined or EDXAPP_COMPREHENSIVE_THEME_SOURCE_REPO == "") and simpletheme_folder_stat.stat.exists == False
```
smarnach: I would prefer for the condition to go at the top, since this is what I'm used from most other programming languages, like

    if condition:
        do_something_cool()

In Ansible people tend to write the condition at the end, so we should likely follow that convention.

In the condition itself, I'd again remove the `is defined` bit and simplify `== False` – see above.

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,122 @@
...
+
+# Done in a separate step because "git:" doesn't have owner/group parameters
+- name: Adjust owner/group of skeleton theme
```
smarnach: I think this task should go directly after the git task.

playbooks/roles/simple_theme/tasks/main.yml
```diff
@@ -0,0 +1,16 @@
...
+# Simple theme
+#
+# See documentation in README.rst
+
+# This flag should have been called EDXAPP_SIMPLETHEME_DEPLOY
+# but it would be incompatible with include_role because of
+# https://github.com/ansible/ansible/issues/19472
+# Therefore we use "include" here (instead of "include_role")
+# Also, because the edxapp role can be included twice in a playbook, we limit
+# this role to the 2nd run, the one without celery_worker
```
smarnach: I don't think it's clear whether it's the first or the second run, and I don't think this captures the gist of the condition we added.

The idea of being able to run the edxapp role in two different modes is that edX themselves deploy the celery workers to different machines than the web application.  Since there is a lot of overlap between the setups for both, they use the same role with slight variations triggered by the `celery_worker` flag.  In our playbook, we want to install the celery workers into the same environment as the web app, so we include the role twice.

So the actual reason for the condition below is that the theme is only needed for the web app, and it's useles for the celery workers.  Even in contexts where the role is included only once with `celery_worker: True`, we don't want the simple_theme role to run.  So I'd change the comment to something like
> The theme is only needed for the web application, so we skip running the role for the Celery workers.

smarnach (commented):
@clemente Here's another round of comments, mostly stylistic.

playbooks/roles/simple_theme/tasks/deploy.yml
clemente: > However, since the simple_theme role is designed to only be run from the edxapp role,

I'm always taking into account that we might want to run this role outside edxapp too, e.g. for deploying theme changes.
But I'm ok with the ansible role crashing if called in an unsupported way, so I'll remove the „is defined“.

playbooks/roles/simple_theme/tasks/deploy.yml
clemente: Ok. This was copied from somewhere.
I still don't trust ansible boolean variables enough to be sure that „not xxx“ will always work. But if it works for you, I'll use it and try it.

playbooks/roles/simple_theme/tasks/deploy.yml
clemente: > I would prefer for the condition to go at the top,

Good idea, all „when:“ could be written first. But yes, the rest of the code (including edx roles) tend to put it at the end, so I'll leave it there.

Also, it seems to me that ansible have put some effort into making ansible *not* seem a programming language. E.g. no variable assignments, strange variable scoping, no print(), no debug, …

smarnach (approved):
@clemente Thanks for the changes!  Looks good to me now, :+1:.

To prepare this for upstream review, here are a few things that are still left to do:
* Trigger the bot to create an OSPR ticket (see the first comment) and add the ticket number to the PR description.
* Make sure the PR desrciption is up to date – e.g. testing instructions and author notes.

openedx-webhooks:
Thanks for the pull request, @clemente! I've created [OSPR-1896](https://openedx.atlassian.net/browse/OSPR-1896) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

gsong:
@clemente I would also consider squashing some commits.

clemente:
@gsong Of course. I squashed all commits to just 1, because it's just 1 feature. I also rebased on top of master.

clemente:
I added a new feature, to download static files from URL. @smarnach could you review the commit? Soon I'll check the feature as part of the deployment process.
I'll squash commits and rebase after it's reviewed.

smarnach:
@clemente I can't review before Wednesday, so if you want feedback earlier than that please ask someone else.

clemente:
@smarnach that's ok, no hurry, could be late in the week. I tested it and it works, and the review isn't blocking anything from my side.

clemente:
Rebased on top of master.

nedbat:
@smarnach are we still waiting for your review?

smarnach:
@nedbat You are right, sorry about that.  This was an internal miscommunication on our end, so this fell through the cracks.  I will review this on Tuesday.

smarnach:
@clemente The latest changes look good to me.

I haven't tested this myself, but I verified the logs of [this appserver](https://stage.console.opencraft.com/instance/1460/edx-appserver/360/) (only accessible for OpenCraft), and all is looking fine. :+1:

playbooks/roles/simple_theme/tasks/deploy.yml
```diff
@@ -0,0 +1,140 @@
...
+  when: SIMPLETHEME_STATIC_FILES_DIR != ""
+
+
+# Downloading remote files is done in two steps: create directorie + download each file.
+# This step is done after the static files from SIMPLETHEME_STATIC_FILES_DIR have been
+# copied, therefore remote files may overwrite the previously installed static files.
+- block:
+  - name: Create directories for static files to be downloaded
+    file:
+      path: "{{ simpletheme_folder }}/{{ item.dest | regex_replace('/[^/]*$','') }}"
```
smarnach: Can we use the `dirname` filter here instead of `regez_replace`?  (Sorry I did not notice this earlier.)

clemente: I added a commit; I can squash it with the first one if you prefer.
I added this to our other internal branches (1 new PR, 1 new commit in existing PR).

smarnach: Thanks!  please wait for upstream review before squashing.

smarnach:
@nedbat We've finished the internal review on this PR now – sorry for the noise.

clemente:
@nedbat Any news on this review?
We already use this code in our servers and it's working.

nedbat:
@edx/devops is this in your backlog?

coryleeio:
It is:
https://openedx.atlassian.net/browse/OSPR-1896

@feanil

feanil:
I've scheduled this for 2 sprints out.  So we likely won't get to reviewing it until the new year, but it is now scheduled for review.

clemente:
@feanil Great, thanks!

nadeemshahzad (approved):

nadeemshahzad:
Thanks @clemente for the contribution. I have tested PR on sandbox and changes are looking good to me.

openedx-webhooks:
@clemente 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/configuration&entry.867055334=https://github.com/edx/configuration/pull/4052&entry.1484655318=https://github.com/clemente&entry.752974735=2017-08-20+19:58&entry.1917517419=2018-01-09+14:45&entry.2133058324=Yes) so we can improve your experience in the future.

clemente:
@nadeemshahzad thanks for reviewing.

PR #4057, opened 2017-08-23
Define PROCTORING_SETTINGS in generic env

efischer19:
Context:
- https://openedx.atlassian.net/browse/EDUCATOR-1142,
- https://github.com/edx/edx-platform/pull/15743/files

We're currently trying to grab this setting in the cms, but it only exists in the lms for some reason.


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1116,6 +1116,7 @@ generic_env_config:  &edxapp_generic_env
   BASE_COOKIE_DOMAIN: "{{ EDXAPP_BASE_COOKIE_DOMAIN }}"
 
   POLICY_CHANGE_GRADES_ROUTING_KEY: "{{ EDXAPP_POLICY_CHANGE_GRADES_ROUTING_KEY }}"
+  PROCTORING_SETTINGS: "{{ EDXAPP_PROCTORING_SETTINGS }}"
```
efischer19, line 1119: There didn't seem to be any ryhme or reason to the structure/order of the values under `generic_env_config:`, so I just slapped this one on the end. Happy to move it if there is in fact some sort of system I should be following.

jibsheet: At one time it was maybe alphabetized, but honestly now it seems to be a giant jumble puzzle and people add to the end

jibsheet (approved):

jibsheet:
Let tests finish, then go ahead and merge, let us know if we need to rerun the pipeline to get it out tomorrow

efischer19:
Nah, this has never worked since merge and no one externally has even mentioned it - we're fine to wait and let it get picked up on the next normal run.

Thanks for the help!

PR #4058, opened 2017-08-23
Setting up new jenkins roles

michaelyoungstrom:
This PR creates a total of 2 new Jenkins roles: jenkins_common and jenkins_build.

Jenkins_common has been designed as a generic template that other teams can inherit for their own instances of Jenkins. jenkins_build is our instance of Jenkins.

These roles take advantage of our recently created repo: https://github.com/edx/jenkins-configuration in order to bootstrap the setup process for jenkins; programatically configuring plugins, adding in credentials, and generating a seed job. The files needed for defining the desired configuration exist in the `files` directory of jenkins_build, and the groovy scripts are defined by `jenkins_configuration_scripts` in jenkins_common.

My commands for running these:

**Build Jenkins:** `ansible-playbook -i /path/to/host/file ../configuration/playbooks/edx-east/jenkins_build.yml -e@/path/to/general/secrets -e@/path/to/build_jenkins_secrets`

**Test Jenkins:** `ansible-playbook -i /path/to/host/file ../configuration/playbooks/edx-east/jenkins_test.yml -e@/path/to/general/secrets -e@/path/to/test_jenkins_secrets`

Apologies for this becoming such a large PR.

Configuration Pull Request
---


playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -0,0 +1,70 @@
...
+  - 4configureGit.groovy
+  - 4configureHipChat.groovy
+  - 4configureJobConfigHistory.groovy
+  - 4configureMailerPlugin.groovy
+  - 5addSeedJob.groovy
+  - 5createLoggers.groovy
+  - 6importCredentials.groovy
+
+# MAIN CONFIG DEFAULT SECRET
+
```
jzoldak: totally optional style thing here - IMO it would be  more readable with an empty line before these headings but no empty line between the heading and its first item.

playbooks/roles/jenkins_common/tasks/main.yml
```diff
@@ -0,0 +1,91 @@
+---
+- name: Install jenkins specific system packages
+  apt:
+    name: '{{ item }}'
+    state: present
+    update_cache: yes
+  with_items: '{{ jenkins_debian_pkgs }}'
+  tags:
+    - jenkins
```
jzoldak: Is there documentation for when/why you would use certain tags, and what effect doing so would have?

playbooks/roles/jenkins_test/files/ec2/mongo_init_script.sh
```diff
@@ -0,0 +1,16 @@
+# This confirms that mongo is running and is accessible on localhost
```
jzoldak: It seems like most if not all of playbooks/roles/jenkins_test/files is a duplicate of playbooks/roles/jenkins_build/files. Is there a way that we can avoid this repetition? Maybe we could use symbolic links?

My concern is for DRY code, and also I'm concerned that the two folder contents will get unintentionally out of step with each other, and thus test-jenkins will be dissimilar to build-jenkins and no longer a valid environment for verifying functionality before it gets to production.

jzoldak (changes requested):

playbooks/roles/jenkins_build/defaults/main.yml
michaelyoungstrom: Agreed. Fixed

playbooks/roles/jenkins_common/tasks/main.yml
michaelyoungstrom: For the most part I followed the jenkins_master role, but there is some general documentation here: https://openedx.atlassian.net/wiki/spaces/OpenOPS/pages/39584735/Role+Life-cycle+Tags . If you see any you disagree with let me know. A few seemed tough to categorize

playbooks/roles/jenkins_test/files/ec2/mongo_init_script.sh
michaelyoungstrom: Fixed this by getting rid of the jenkins_test role. It really wasnt necessary. The difference in secrets is already handled in our secure repo, and any differences in variables can still be addressed in the actual play for test jenkins

michaelyoungstrom:
CC: @edx/devops

jzoldak (approved):
This looks awesome @michaelyoungstrom!

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -0,0 +1,67 @@
+build_jenkins_user: jenkins
+build_jenkins_version: jenkins_1.651.3
+build_jenkins_war_source: https://s3.amazonaws.com/edx-testeng-tools/jenkins
+build_jenkins_server_name: build.testeng.edx.org
+build_jenkins_git_home: /git
+
+build_jenkins_debian_pkgs:
```
jibsheet: This looks like the same list used in jenkins_common.  You don't need to override and pass unless you're changing it (you can just leave this var out of the meta call later).  I didn't check the rest, but we may be able to prune a few overrides here?

playbooks/roles/jenkins_build/files/ec2_config.yml
```diff
@@ -0,0 +1,680 @@
+---
+CLOUDS:
+    - NAME: 'us-east-1'
```
jibsheet: I would make regions a variable (this is where the secure vs internal thing comes in, but for you folks I'd just stuff all your vars in your secure repo).

Update: imagine I put this like 5 lines later where you have REGION: 'us-east-1' - sentiment stands

playbooks/roles/jenkins_build/files/ec2_config.yml
```diff
@@ -0,0 +1,680 @@
...
+      ACCESS_KEY_ID: '{{ JENKINS_EC2_ACCESS_KEY_ID }}'
+      SECRET_ACCESS_KEY: '{{ JENKINS_EC2_SECRET_ACCESS_KEY }}'
+      USE_INSTANCE_PROFILE_FOR_CREDS: false
+      REGION: 'us-east-1'
+      EC2_PRIVATE_KEY_PATH: '{{ build_jenkins_config_path }}/ec2/id_rsa'
+      INSTANCE_CAP: '250'
+      AMIS:
+          - AMI_ID: 'ami-33580448'
```
jibsheet: these are unlikely to be public AMIs, right

playbooks/roles/jenkins_build/files/git_config.yml
```diff
@@ -0,0 +1,3 @@
+---
+NAME: 'jenkins'
+EMAIL: 'jenkins@edx.org'
```
jibsheet: Don't put edx email addresses in the confirmation repo, folks will try to deploy this and you'll be super confused when you get emails (we generally use jenkins@example.com as a default with a variable to override)

playbooks/roles/jenkins_build/files/mailer_config.yml
```diff
@@ -0,0 +1,9 @@
+---
+SMTP_SERVER: 'email-smtp.us-east-1.amazonaws.com'
+REPLY_TO_ADDRESS: 'edx-platform-build <ci@edx.org>'
+DEFAULT_SUFFIX: '@edx.org'
```
jibsheet: similar to earlier comments about edx.org

playbooks/roles/jenkins_common/defaults/main.yml
```diff
@@ -0,0 +1,26 @@
+jenkins_common_user: jenkins
+jenkins_common_group: edx
+jenkins_common_home: /var/lib/jenkins
+jenkins_common_config_path: /init-configs
+jenkins_common_port: 8080
```
jibsheet, line 5: Looks like we can just use the defaults for many of these in the jenkins_build role

jibsheet (commented):
This looks great.  I called out a few things where we're putting edX specific data in a public repo.

Larger-scale, I wonder if that ec2 config should get pulled back internally since it's *so* specific to what we do, and making everything a var would get tedious (unless you can easily have a list of dictionaries and use ansible/jinja2 to generate that file).
If you do pull it into an internal repo, we'd want to document the use of it for folks

playbooks/roles/jenkins_build/files/ec2_config.yml
```diff
@@ -0,0 +1,680 @@
...
+                  VALUE: 'jenkins'
+                - NAME: 'master'
+                  VALUE: 'build.testeng.edx.org'
+            IDLE_TERMINATION_MINUTES: '10'
+            USE_PRIVATE_DNS_NAME: false
+            INSTANCE_CAP: ''
+            IAM_INSTANCE_PROFILE: '{{ JENKINS_EC2_IAM_INSTANCE_PROFILE }}'
+            USE_EPHEMERAL_DEVICES: false
+            LAUNCH_TIMEOUT: ''
+          - AMI_ID: 'ami-70fea467'
```
estute: Perhaps we should just put the publicly available Xenial amis here. Open source users won't have access to it. Also, not sure if this value registers on the sensitivity scale.

playbooks/roles/jenkins_build/files/main_config.yml
```diff
@@ -0,0 +1,27 @@
...
+GLOBAL_PROPERTIES:
+    ENVIRONMENT_VARIABLES:
+        - NAME: 'BROWSERMOB_PROXY_PORT'
+          VALUE: '9090'
+        - NAME: 'GITHUB_OWNER_WHITELIST'
+          VALUE: '{{ JENKINS_MAIN_GITHUB_OWNER_WHITELIST }}'
+    TOOL_LOCATIONS:
+LOCATION:
+    URL: 'https://build.testeng.edx.org/'
+    ADMIN_EMAIL: 'edx-platform-build <ci@edx.org>'
```
estute: ^not sure if we should have emails in the open

playbooks/roles/jenkins_build/meta/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+dependencies:
+  - common
+  - edxapp_common
+  - browsers
+  - role: jenkins_common
+    jenkins_common_configuration_scripts: '{{ build_jenkins_configuration_scripts }}'
+    jenkins_common_server_name: '{{ build_jenkins_server_name }}'
+
+  - role: docker-tools
```
estute: What is going on here? While I want to get to a point where we are using Docker for our Jenkii, we don't currently. I think we should simplify this effort into generating an identical matching Build Jenkins to what we have today.

estute (approved):
This looks great! I have a question about the docker-tools role, but maybe there is something that I am missing. Otherwise :+1:

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -0,0 +1,62 @@
...
+
+# MAIN CONFIG DEFAULT SECRET
+JENKINS_MAIN_GITHUB_OWNER_WHITELIST: ''
+JENKINS_MAIN_ADMIN_EMAIL: ''
+
+# MAILER CONFIG DEFAULT SECRETS
+JENKINS_MAILER_SMTP_AUTH_USERNAME: ''
+JENKINS_MAILER_SMTP_AUTH_PASSWORD: ''
+JENKINS_MAILER_REPLY_TO_ADDRESS: 'edx-platform-build <ci@edx.org>'
+JENKINS_MAILER_DEFAULT_SUFFIX: '@edx.org'
```
jibsheet: These are still edx.org defaults

playbooks/roles/jenkins_build/files/main_config.yml
```diff
@@ -0,0 +1,27 @@
...
+    DISABLE_REMEMBER_ME: true
+GLOBAL_PROPERTIES:
+    ENVIRONMENT_VARIABLES:
+        - NAME: 'BROWSERMOB_PROXY_PORT'
+          VALUE: '9090'
+        - NAME: 'GITHUB_OWNER_WHITELIST'
+          VALUE: '{{ JENKINS_MAIN_GITHUB_OWNER_WHITELIST }}'
+    TOOL_LOCATIONS:
+LOCATION:
+    URL: 'https://build.testeng.edx.org/'
```
jibsheet: We shouldn't have edx.org real URLs in here (or at least they should be easier to override)

playbooks/roles/jenkins_build/templates/ec2_config.yml.j2
```diff
@@ -0,0 +1,44 @@
+---
```
jibsheet: This looks so much better

playbooks/roles/jenkins_common/defaults/main.yml
jibsheet: stand by this note, can you tell me if I missed something?

playbooks/roles/jenkins_build/meta/main.yml
```diff
@@ -4,5 +4,18 @@ dependencies:
   - edxapp_common
   - browsers
```
estute: Just realized that this is here. Browsers and edxapp_common are unnecessary on the build Jenkins master.

jibsheet (approved):

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -0,0 +1,148 @@
...
+  - 4configureGHPRB.groovy
+  - 4configureGit.groovy
+  - 4configureHipChat.groovy
+  - 4configureJobConfigHistory.groovy
+  - 4configureMailerPlugin.groovy
+  - 5addSeedJob.groovy
+  - 5createLoggers.groovy
+  - 6importCredentials.groovy
+
+# plugins
```
estute, line 18: Sorry to hold this up, but this should an easy fix (or separate PR): The complete list of needed plugins for build jenkins can be found here https://github.com/edx/configuration/pull/3969/files

estute:
:+1: thanks for the update!!

PR #4059, opened 2017-08-23
Determine the size of the root volume not sdf

feanil:
Configuration Pull Request
---


jibsheet (approved):

macdiesel (approved):
Um sure... but some context would help me understand why.

feanil:
Without this change, the default size of the root volume remains 8GB no matter what.  You run into this situation when you build from a base ubuntu AMI which has an 8GB volume.  The expectation is that this var sets the size of the root volume so correct the device name to match that expectation.

PR #4060, opened 2017-08-24
allow custom blacklist rules for splunk forwarder

estute:
We don't want to log the seed job logs. In the future, we will probably expand this, as other jobs are not important enough to index. I find it is more readable when the blacklist is complicated rather than the monitor field, but that's just me.


feanil (approved):

jibsheet (approved):

estute:
@jzoldak I manually tested this on test jenkins, and it indexed correctly.

jzoldak (approved):

PR #4062, opened 2017-08-24
Make edx_service stricter about edx_service_config

jibsheet:
With the defined check, it's possible to have config syntax errors and Ansible will then skip the task and not write a config file.  I contend that we want ansible to bail out if the dict it gets in edx_service_config is invalid.  It appears to be the checking of the variable that evaluates the various interpolations.

Noticed this when manually testing config changes and saw missing files rather than syntax errors.  In particular, this will make our config checking code much better.

### DIFF
--- a/playbooks/roles/edx_service/defaults/main.yml
+++ b/playbooks/roles/edx_service/defaults/main.yml
@@ -17,6 +17,12 @@
 edx_service_name: edx_service
 
 edx_service_repos: []
+
+# A few roles meta this role but don't need a config file written
+# this allows them to not pass a config and the tasks will skip
+# and not write out a config at all.
+edx_service_config: {}
+
 #
 # OS packages
 #
--- a/playbooks/roles/edx_service/tasks/main.yml
+++ b/playbooks/roles/edx_service/tasks/main.yml
@@ -95,7 +95,7 @@
     src: "config.yml.j2"
     dest: "{{ COMMON_CFG_DIR  }}/{{ edx_service_name }}.yml"
     mode: "0644"
-  when: edx_service_config is defined
+  when: edx_service_config
   tags:
     - install
     - install:configuration


jdmulloy:
Nit: The commit message should describe the fix instead of the problem. Perhaps something like "Fix Ansible to bail out if edx_service_config not defined"

jibsheet:
This PR only exists in the current state so I can see what Travis thinks.  I expect messages to be rewritten before I ask for review.

jibsheet:
I also disagree, commit messages that explain the problem being solved are more useful to me.  I can often tell what you changed by reading the diff, but I never know *why* unless you remember and can explain it (hence why I tend to explain things into the commit message).

feanil:
My understanding was that there is a difference between being defined and being valid.  An invalid thing would still be considered defined.

jibsheet:
@feanil With the is defined check, invalid YAML passes a build, but the step is skipped and no file is produced.  Checking for existence bubbles up the error.  Possibly an ansible version change.

```
"msg": "The conditional check 'edx_service_config' failed. The error was: error while evaluating conditional (edx_service_config):
large dictionary
'dict object' has no attribute 'sop_access_key'\n\nThe error appears to have been in '/Users/kevinfalcone/git/edx/configuration/playbooks/roles/edx_service/tasks/main.yml': line 93, column 3, but may\nbe elsewhere in the file depending on the exact syntax problem.\n\nThe offending line appears to be:\n\n\n- name: Write out app config file\n  ^ here\n"
```

I'm guessing lazy evaluation of some kind biting us.  Travis says this is mostly a-ok so I'll prep the rest of the changes when I get a few more things done.  I would like to get to the bottom of this though, otherwise it means our config checker will pass invalid code (since I'm basically just running the config checker locally to test).

jdmulloy (approved):

jibsheet:
@feanil I emailed you the script I'm using to test (since it hardcodes in a bunch of internal configuration structure) so you can replicate and see if you see a better/different solution

playbooks/roles/edx_service/defaults/main.yml
```diff
@@ -17,6 +17,12 @@
 edx_service_name: edx_service
 
 edx_service_repos: []
+
+# A few roles meta this role but don't need a config file writte
```
feanil: `writte` should be `written`

feanil (approved):
I was able to re-produce this bug.

feanil:
FYI: Submitted https://github.com/ansible/ansible/issues/28733

PR #3099, opened 2016-06-08
WL-506: Update configuration ansible script for edxapp to support lms theming

saleem-latif:
Hi @mattdrayer , @fredsmith , @feanil 

Kindly review this PR, it contains changes for lms theming, 

**Description of Changes:**
- We need make following updates to `edxapp` role,
- We need to make sure to clone `edx-themes` repo
- Add `edxapp` user to `edx-themes` group
- Rename `EDXAPP_COMPREHENISVE_THEME_DIR` to `EDXAPP_COMPREHENISVE_THEME_DIRS`
- Add `EDXAPP_ENABLE_COMPREHENSIVE_THEMING` variable to `default/main.yml`
- Update the value of `EDXAPP_COMPREHENISVE_THEME_DIRS` so that it is a list of directories.
- Rename environment variable `COMPREHENISVE_THEME_DIR` to `COMPREHENISVE_THEME_DIRS` in `lms.env.json` and `cms.env.json`

PR #4063, opened 2017-08-25
This was overriding using a nonexistant jenkins var

jibsheet:
Configuration Pull Request
---

Jenkins wasn't updated when this code was changed, so THEME_DIRS
was always null.  We're now going to just set the correct directories
in the internal sandbox config which removes the need for this var
setting here.


jibsheet:
For historical purposes https://github.com/edx/configuration/pull/3099
Note that even if Jenkins had been changed, it would have been passing a string to a list var, so this isn't even a shim of backwards compatibility.
https://github.com/edx/jenkins-job-dsl/pull/202 and https://github.com/edx/sandbox-internal are providing the rest of the fix.

jdmulloy (approved):

feanil (approved):

PR #4075, opened 2017-08-31
Add OAUTH_DELETE_EXPIRED to LMS Envs

bmedx:
Configuration Pull Request
---


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -541,6 +541,7 @@ EDXAPP_DEFAULT_CACHE_VERSION: "1"
 EDXAPP_OAUTH_ENFORCE_SECURE: True
 EDXAPP_OAUTH_EXPIRE_CONFIDENTIAL_CLIENT_DAYS: 365
 EDXAPP_OAUTH_EXPIRE_PUBLIC_CLIENT_DAYS: 30
+EDXAPP_OAUTH_DELETE_EXPIRED: True
```
feanil: What will this do to people other than edX that have a bunch of expired tokens in their DB?  Are they going to see a DB IO spike the first time the deletion code hits?  Also, when/how does the deletion code run?

bmedx: This only operates as individual tokens are used. Currently when a Grant is consumed the expiration time of the Grant is updated to be a day in the past, so performance wise that will just change to a delete instead. There should not be any bulk deletions caused by this change. If non-edx people are using these items as foreign keys for some reason, or have delete signals hooked up to them it will cause a behavior change, but that seems an unlikely circumstance. I can change this to the default of False and add overrides if you think that's better.

feanil: Awesome, thanks.  I think it's safe to leave as is then.   Can you add a comment in the defaults file saying what you have in your github comment just so it's clear to openedx ops consumers.

feanil:
@bmedx this also feels like it's worth an entry at the top of the changelog.

jibsheet:
Yep - I think with the comment and changelog this is 👍

jibsheet (approved):
Squash and merge

PR #4096, opened 2017-09-14
updated format of cert template languages setting

McKenzieW:
Configuration Pull Request
dict --> list of tuples
---

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -308,8 +308,8 @@ EDXAPP_LANG: 'en_US.UTF-8'
 EDXAPP_LANGUAGE_CODE : 'en'
 EDXAPP_LANGUAGE_COOKIE: 'openedx-language-preference'
 EDXAPP_CERTIFICATE_TEMPLATE_LANGUAGES:
-  'en': 'English'
-  'es': 'Español'
+  - ('en', 'English')
+  - ('es', 'Español')
 EDXAPP_TIME_ZONE: 'America/New_York'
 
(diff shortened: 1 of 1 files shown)


McKenzieW:
@edx/devops can I get a quick review on this data structure change to a config variable I recently added?

jibsheet:
@McKenzieW Got a platform PRl ink?  Or was the format wrong and you're fixing to match what platform wants?

McKenzieW:
@jibsheet the format was wrong, this is a fix for what platform wants

jibsheet:
Have you tested the yaml -> json of this on a sandbox or something else that goes through ansible?  As best I can tell, your current version is a list of 2 strings.  You can get a list of lists by doing something like
```
- ["en","English"] 
```
or
```
-
  - en
  - English
```
 but that's a list, not a tuple.  Tuples are harder and I'm honestly not sure ansible is going to support them since pyyaml only does so with tags. http://pyyaml.org/wiki/PyYAMLDocumentation#YAMLsyntax

McKenzieW:
@jibsheet you're right. In all honesty I don't have much experience with yaml and made some incorrect assumptions about how it would be converted. It seems like it will probably make more sense to just update the lms-side logic to deal with the data in its existing format. So no need for this PR anymore.

jibsheet:
@McKenzieW If it's easier on the platform side to do list of lists -> list of tuples, I'm happy to help you get the lists of lists right (although list of dicts -> list of tuples should also be an easy comprehension)

McKenzieW:
@jibsheet Thank you! I think I've got the dict --> tuples logic down though, so we should be all set here.

PR #4106, opened 2017-09-25
create video pipeline role

muhammad-ammar:
Configuration Pull Request
---


muhammad-ammar:
Below provisioning through vagrant works and new `video_pipeline` role executed 
```bash
vagrant box add ubuntu_xenial /Users/mammar/Downloads/ficus-devstack-2017-02-07.box
VAGRANT_VAGRANTFILE=/Users/mammar/Documents/edx/devstack/src/configuration/pipeline-vagrantfile ANSIBLE_CONFIG=/Users/mammar/Documents/edx/devstack/src/configuration/ansible_dev.cfg vagrant up
```

but if I try to execute the below `ansible-playbook` command then it fails at `TASK [user : Grant full sudo access to the edxadmin group]`.

```bash
ANSIBLE_CONFIG=/Users/mammar/Documents/edx/devstack/src/configuration/ansible_dev.cfg ansible-playbook -i 192.168.33.10, --extra-vars="migrate_db=yes disable_edx_services=false edx_django_service_has_static_assets=yes edx_django_service_version=ammar/changes-for-config" /Users/mammar/Documents/edx/devstack/src/configuration/playbooks/edx-east/video_pipeline.yml --ask-sudo-pass -vvv -c local
```

**error with ansible-playbook command**
```
fatal: [local]: FAILED! => {
    "changed": false,
    "checksum": "c2fc4cba65c501225c024ac45c0c795f6f230416",
    "failed": true,
    "invocation": {
        "module_args": {
            "backup": false,
            "content": null,
            "delimiter": null,
            "dest": "/etc/sudoers.d/edxadmin",
            "directory_mode": null,
            "follow": false,
            "force": true,
            "group": "root",
            "mode": 288,
            "original_basename": "tmpvUR2hP",
            "owner": "root",
            "regexp": null,
            "remote_src": null,
            "selevel": null,
            "serole": null,
            "setype": null,
            "seuser": null,
            "src": "/Users/mammar/.ansible/tmp/ansible-tmp-1506428415.54-39676294445235/source",
            "unsafe_writes": null,
            "validate": "visudo -cf %s"
        }
    },
    "msg": "Destination directory /etc/sudoers.d does not exist"
}
```
The weird thing is that `/etc/sudoers.d` exists in the box. I have manually verified that.

**Question:** What should I do to successfully execute the `ansible-playbook` command?
**Question:** What to if I don't want to execute a task for the new service? Foe example `install nodenv` and `create nodeenv` tasks [here](https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/tasks/main.yml). Should I add new tag and skip tags or skip task using `when` conditional?

ansible_dev.cfg
```diff
@@ -0,0 +1,5 @@
+[defaults]
+jinja2_extensions = jinja2.ext.do
+private_key_file = /Users/mammar/.vagrant.d/boxes/ubuntu_xenial/0/virtualbox/vagrant_private_key
+remote_port = 2222
+remote_user = vagrant
```
muhammad-ammar: copied these settings from `vagrant ssh-config` command

feanil:
@muhammad-ammar the `-c local` will cause the command to run on your local machine not on the "remote" vagrant box.   You should remove that bit and then I think it will run on the right machine.

ansible_dev.cfg
feanil: Why do we need to include this in the repo?  You should be able to run vagrant with an ansible provisioner for testing,  take a look at what we do [here](https://github.com/edx/configuration/blob/master/vagrant/base/devstack/Vagrantfile).

pipeline-vagrantfile
```diff
@@ -0,0 +1,57 @@
+Vagrant.require_version ">= 1.8.7"
```
feanil: Vagrant files should not go at the root level, please make a new directory relevant to Veda under the `vagrant` folder.

Ideally we should not use vagrant at all but make a docker based devstack but if this is the technology you are familiar with for now I think it's fine to move us forward.

playbooks/roles/ansible-role-django-ida/tasks/main.yml
```diff
@@ -10,7 +10,7 @@
 
 - name: Create docker directories
   file:
-    path: "../docker/{{ item }}/{{ my_role_name }}"
+    path: '../docker/{{ item }}/{{ my_role_name|replace("_","-") }}'
```
feanil: Why are you replacing the underscores with dashes in the directory names?

playbooks/roles/video_pipeline/defaults/main.yml
```diff
@@ -0,0 +1,118 @@
...
+
+VIDEO_PIPELINE_NGINX_PORT: "1{{ video_pipeline_gunicorn_port }}"
+VIDEO_PIPELINE_SSL_NGINX_PORT: "4{{ video_pipeline_gunicorn_port }}"
+
+VIDEO_PIPELINE_DEFAULT_DB_NAME: 'pipeline'
+VIDEO_PIPELINE_MYSQL: 'localhost'
+# MySQL usernames are limited to 16 characters
+VIDEO_PIPELINE_MYSQL_USER: 'pipeline001'
+VIDEO_PIPELINE_MYSQL_PASSWORD: 'password'
+VIDEO_PIPELINE_MYSQL_PORT: 'password'
```
feanil: Port should be a number.

playbooks/roles/video_pipeline/defaults/main.yml
```diff
@@ -0,0 +1,118 @@
...
+
+VIDEO_PIPELINE_MEMCACHE: [ 'memcache' ]
```
feanil: Video pipeline does not have a memcache by default, right? I think this should be empty by default.

playbooks/roles/video_pipeline/tasks/main.yml
```diff
@@ -0,0 +1,27 @@
...
+#
+#
+# Example play:
+#
+#
+
+- name: Create video pipeline configuration
+  template:
+    src: "templates/instance_config.yaml.j2"
+    dest: "{{ edx_django_service_code_dir }}/instance_config.yaml"
```
feanil: You're referring to the edx_django_service_code_dir variable outside of that role.  This is generally not a good practice as you can't guarantee its value because multiple roles could call the same meta role.  You should make the relevant variable in your defaults/main.yml and then refere to it here.  Even if you do that I would not want it to be in the code dir.

Would it be possible to put this file in a different directory and then point to it from the code directory?  For example, the config dir for the app. `{{ COMMON_CFG_DIR}}/{{ {{ video_pipeline_service_name }}`

feanil (commented):
First pass of the review.  Let me know if anything I said is unclear.

muhammad-ammar:
I should have mentioned this earlier. `ansible_dev.cfg` and `pipeline-vagrantfile` will never be committed, they are just for my local testing against a vagrant box.

playbooks/roles/video_pipeline/tasks/main.yml
muhammad-ammar: Yes, we can definitely put this file in `{{ COMMON_CFG_DIR}}/{{ video_pipeline_service_name }}` but this will require changes in the `edx-video-pipeline` source. Can we go with things as they are now and update this task once we have refactored the `edx-video-pipeline` source?

playbooks/roles/ansible-role-django-ida/tasks/main.yml
muhammad-ammar: `Update docker templates` task in this file will fail without the above change becuase it expects underscore `_`  in the role name directory to be replaced with hyphen `-`. Anyway I will revert this change because I haven't used the cookiecutter.

muhammad-ammar:
@feanil Video processing flow involves [4 repos](https://github.com/edx/terraform/blob/master/plans/veda/sandbox/sandbox.tf#L617-L620)

- https://github.com/edx/edx-video-pipeline
- https://github.com/edx/edx-video-worker
- https://github.com/yro/chunkey
- https://github.com/yro/v_videocompile

The last 2 repos are only used by `edx-video-worker`, So mainly first 2 repos are responsible. `edx-video-worker` encodes studio uploaded video to different formats and `edx-video-pipeline` perform tasks like communicating with external transcript providers for transcript generation. Both repos maintain queues for processing videos. 

My question is how do you recommend/suggest the way new ansible role should be written? Any existing role/example I can look into? I am little bit stuck at what could be best way to handle multiple repos? One way could be to have virtualenv for 2 repos and then they can communicate with each other to accomplish the task? Currently all the repos are installed like [this](https://github.com/edx/terraform/blob/master/plans/veda/sandbox/sandbox.tf#L643-L667).

playbooks/roles/video_pipeline/tasks/main.yml
feanil: That sounds good to me.

feanil:
I would recommend a separate role that sets up the workers is probably the simplest way to move forward.  The worker runs celery correct?  It will be hard to set that up using the edx_django_service role I think but you can perhaps use the edx_service role instead which is a bit more general.

muhammad-ammar:
@feanil I am currently blocked and needs your help. During provisioning, `migrate db` task fails because [database user](https://github.com/edx/configuration/pull/4106/files#diff-913447ca033da373733a8f9a5b3b25afR41) is not created. Database users are created in [edxlocal](https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/defaults/main.yml) role. Should I need to include edxlocal as dependency somewhere?
Also If I want to test the new role on a sandbox then I need to use my configuration branch. Is there anything else I need to do?

feanil:
The sandbox playbook includes edxlocal.  You can include that for your testing in the playbook for running this role..  Remember in production, the database should already have users that are provisioned independent of bringing up the application.

muhammad-ammar:
@feanil I have made more changes. Please have a look. I have some questions also.

* What is the process of accessing a resource (for example, s3 bucket) in ansible which was created by a terraform plan?
* Currently there are lot of variables in the `defaults/main.yml` and those should be overrided. So what is best pattern/way to provide sane default values for sandbox environment? My initial goal is to spin up a sandbox that can be used for video-pipeline testing without doing any configuration manually. 
* If the role is working fine on sandbox environment then what else is needed for production? 
* Anything you mention that should be done before we can mark this role completed? 


Currently I am doing testing as mentioned [here](https://github.com/edx/configuration/tree/master/vagrant). I have tried to run ansible-playbook with vagrant box instead of running vagrant provision every time but no luck.

feanil:
@muhammad-ammar 
- What do you need to do with the s3 bucket? There is an [s3 ansible module](http://docs.ansible.com/ansible/latest/aws_s3_module.html)
- For things that are not secret or edx-specific, I would say that you should just put the correct values in the `defaults/main.yml`.  For edx specific things, you should put them in the sandbox-internal repo and for secrets they should go in the sandbox-secure repo.
- If the role works as desired on a sandbox, then we probably need to figure out what the production config is as well as how we want to run the roles in production(pipelines, manually, etc) and document that process.

muhammad-ammar:
@feanil I tried to [create a sandbox](https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/5198/console) with this new role but db migration failed. This is the same error I faced during my local testing against vagrant box and fixed by adding edxlocal role after a discussion with. Any idea how we can fix that on sandbox? My understanding was that edxlocal role will be executed on sandbox but looks like that this is not the case.

playbooks/edx-east/video_pipeline.yml
```diff
@@ -0,0 +1,13 @@
...
+  hosts: all
+  become: True
+  gather_facts: True
+  vars:
+    ENABLE_NEWRELIC: False
+    migrate_db: 'yes'
+  roles:
+    - role: nginx
+      nginx_default_sites:
+        - pipeline
```
feanil: The name of the site should be `video_pipeline`

util/jenkins/ansible-provision.sh
```diff
@@ -346,10 +355,14 @@ EOF
     fi
 fi
 
-declare -A deploy
-roles="edxapp forum ecommerce credentials discovery notifier xqueue xserver certs demo testcourses"
+video_pipeline="true"
 
+declare -A deploy
+roles="edxapp forum ecommerce credentials discovery video_pipeline notifier xqueue xserver certs demo testcourses"
+echo "Roles ====>"
```
feanil: Is this for debugging only I assume or something else?

vagrant/base/test_playbook/Vagrantfile
```diff
@@ -2,26 +2,33 @@
 # under /playbooks (e.g., "foo.yml"), export its name (e.g., "foo") as the value
 # of the environment variable VAGRANT_ANSIBLE_PLAYBOOK, and execute `vagrant up`
 # from within this directory.
-MEMORY = 2048
+MEMORY = 1024
```
feanil: I assume this is to get things working on your local machine?

vagrant/base/test_playbook/Vagrantfile
```diff
@@ -2,26 +2,33 @@
...
   end
 
+  config.ssh.forward_x11 = true
+
   config.vm.provision :ansible do |ansible|
     ansible.playbook = "../../../playbooks/" + ENV['VAGRANT_ANSIBLE_PLAYBOOK'] + ".yml"
-    ansible.verbose = "extra"
+    ansible.verbose = "vvvv"
+    ansible.extra_vars = {
+        edx_django_service_version: 'ammar/changes-for-config'
```
feanil: This var should only be updated via meta, you should be setting the version of this var that is specific to the video_pipeline role.

muhammad-ammar:
@feanil This is ready for review. Please have a look.

playbooks/edx-east/video_pipeline.yml
muhammad-ammar: Any specific reason for this name?

playbooks/roles/video_pipeline/defaults/main.yml
```diff
@@ -0,0 +1,338 @@
...
+VIDEO_PIPELINE_SG_SCRIPT_KEY: "SET-ME-PLEASE"
+
+VIDEO_PIPELINE_THREEPLAY_FTPHOST: "SET-ME-PLEASE"
+
+VIDEO_PIPELINE_XUETANG_API_URL: "SET-ME-PLEASE"
+VIDEO_PIPELINE_XUETANG_API_SHARED_SECRET: "SET-ME-PLEASE"
+
+# video pipeline config
+
+VIDEO_PIPELINE_CONFIG:
```
muhammad-ammar: @feanil Both the `edx-video-pipeline` and `edx-video-worker` use this same config, So how can I use this config for `video_worker` role?

util/jenkins/ansible-provision.sh
```diff
@@ -204,6 +204,10 @@ CREDENTIALS_NGINX_PORT: 80
 CREDENTIALS_SSL_NGINX_PORT: 443
 CREDENTIALS_VERSION: $credentials_version
 
+VIDEO_PIPELINE_NGINX_PORT: 80
+VIDEO_PIPELINE_SSL_NGINX_PORT: 443
+VIDEO_PIPELINE_VERSION: ammar/changes-for-config
```
muhammad-ammar, line 209: This should be the branch passed from jenkins create sandbox interface. Should I need to create another ticket to update the jenkins UI? Which repo holds the code for that?

playbooks/edx-east/video_pipeline.yml
feanil: Yes, it should map to the role it is relevant to.  Pipeline is not specific enough.

playbooks/roles/video_pipeline/defaults/main.yml
feanil: You may need to pull up the config into its own role that is pulled in via meta dependencies into both the video_pipeline and video_worker roles.   http://docs.ansible.com/ansible/2.4/playbooks_reuse_roles.html#role-dependencies

util/jenkins/ansible-provision.sh
feanil: Right now we are just trying to get the role working, we'll worry about getting it all merged back and added to the automation as a second step.  The goal here is to have ansible we can run manually to add the video_pipeline web server to the sandbox.

muhammad-ammar: @feanil I can now create sandbox with edx-video-pipeline and as par my minimal testing things are working as expected though I haven't tested every single thing. Here is the [sandbox](https://pipeline-veda-config2.sandbox.edx.org/) I created.

Do you have concerns? Is there anything missing in the role? Any suggestions/recommendations? It would be great if you can explain what is required to merge this PR?

playbooks/roles/video_pipeline/meta/main.yml
```diff
@@ -0,0 +1,61 @@
...
+# }
+dependencies:
+  - role: video_common_vars
+    VEDA_STATIC_ROOT: '{{ VIDEO_PIPELINE_STATIC_ROOT }}'
+    VEDA_DJANGO_SECRET_KEY: '{{ VIDEO_PIPELINE_SECRET_KEY }}'
+    VEDA_MYSQL_DB_NAME: '{{ VIDEO_PIPELINE_DEFAULT_DB_NAME }}'
+    VEDA_MYSQL_HOST: '{{ VIDEO_PIPELINE_MYSQL_HOST }}'
+    VEDA_MYSQL_USER: '{{ VIDEO_PIPELINE_MYSQL_USER }}'
+    VEDA_MYSQL_PASSWORD: '{{ VIDEO_PIPELINE_MYSQL_PASSWORD }}'
+    VEDA_URL_ROOT: '{{ VIDEO_PIPELINE_URL_ROOT }}'
```
muhammad-ammar: @feanil I have created a role for config for video-pipeline and video-worker. Is this how you are suggesting?

playbooks/roles/video_pipeline/meta/main.yml
```diff
@@ -0,0 +1,61 @@
...
+    edx_django_service_db_password: '{{ VIDEO_PIPELINE_MYSQL_PASSWORD }}'
+    edx_django_service_default_db_atomic_requests: false
+    edx_django_service_social_auth_edx_oidc_key: '{{ VIDEO_PIPELINE_SOCIAL_AUTH_EDX_OIDC_KEY }}'
+    edx_django_service_social_auth_edx_oidc_secret: '{{ VIDEO_PIPELINE_SOCIAL_AUTH_EDX_OIDC_SECRET }}'
+    edx_django_service_social_auth_redirect_is_https: '{{ VIDEO_PIPELINE_SOCIAL_AUTH_REDIRECT_IS_HTTPS }}'
+    edx_django_service_extra_apps: '{{ VIDEO_PIPELINE_EXTRA_APPS }}'
+    edx_django_service_automated_users: '{{ VIDEO_PIPELINE_AUTOMATED_USERS }}'
+    edx_django_service_use_python3: false
+    nginx_edx_django_service_gunicorn_hosts: '{{ NGINX_VIDEO_PIPELINE_GUNICORN_HOSTS }}'
+    edx_django_service_config: '{{ VEDA_COMMON_CONFIG }}'
```
muhammad-ammar: @feanil `VEDA_COMMON_CONFIG` is defined in `video_common_vars`. Should I expect it to have correct values?

playbooks/edx-east/video_pipeline.yml
muhammad-ammar: ok...Will there be any issue if I use a name with `underscore` or `hyphen` like `video_pipeline`?

feanil: no, that should be fine, it's valid yaml.

playbooks/roles/video_pipeline/meta/main.yml
```diff
@@ -0,0 +1,72 @@
...
+# Example:
+#
+# dependencies:
+#   - {
+#   role: my_role
+#   my_role_var0: 'foo'
+#   my_role_var1: 'bar'
+# }
+dependencies:
+  - role: edx_django_service
```
muhammad-ammar, line 22: @feanil `edx_django_service : migrate database` task was run by `edx_django_service` role. And it expects that `/edx/etc/pipeline.yml` configuration should be present. But the problem is that configuration file is written by `video_common_vars` role which will be executed after `edx_django_service` role. And I can't execute `video_common_vars` first because `/edx/etc/` and other common paths are created by `edx_django_service` role. This is a deadlock. How can we resolve this issue?

playbooks/edx-east/oauth_client_setup.yml
```diff
@@ -9,5 +9,6 @@
     - "roles/ecommerce/defaults/main.yml"
     - "roles/credentials/defaults/main.yml"
     - "roles/discovery/defaults/main.yml"
+    - "roles/video_pipeline/defaults/main.yml"
```
feanil, line 12: Does the video pipeline have oauth_client credentials that need to be setup?

playbooks/edx-east/video_pipeline.yml
```diff
@@ -0,0 +1,13 @@
...
+  become: True
+  gather_facts: True
+  vars:
+    ENABLE_NEWRELIC: False
+    migrate_db: 'yes'
+  roles:
+    - role: nginx
+      nginx_default_sites:
+        - video_pipeline
+    - aws
```
feanil, line 12: This should go first in the list of roles.

playbooks/roles/video_pipeline/meta/main.yml
feanil: @muhammad-ammar I don't understand why you need the video common role?   Especially because you override a bunch of the config in it.  Why not just pull those settings all into the video_pipeline role?

util/jenkins/ansible-provision.sh
feanil: If this gets the veda-web-frontend working I think it will be good to merge after we make sure it is following best practices.  Are the steps you take to create your sandbox documented so I can re-produce this?

feanil (commented):
@muhammad-ammar sorry for the delay in reviewing.

playbooks/edx-east/oauth_client_setup.yml
muhammad-ammar: yes. video_pipeline communicate with platform to update video status. Also video workers communicate with video_pipeline

muhammad-ammar:
@feanil Closing in favor of #4162

PR #4110, opened 2017-09-26
add templates for system properties script

estute:
This was originally part of https://github.com/edx/configuration/pull/4099, but should be separated out into it's own PR, as we will want to make sure that this is applied to build jenkins 1.x (when we do the swap), so that we do not lose any functionality


michaelyoungstrom (approved):
LGTM

jzoldak (approved):

PR #4133, opened 2017-10-05
allow adding authorized_keys from ansible inline

fredsmith:
This feature is documented in the user role, but didn't actually work. I decided that an inline key should take precedence over a github key if both github and inline keys were defined, but not throw an error.

### DIFF
--- a/playbooks/roles/user/tasks/main.yml
+++ b/playbooks/roles/user/tasks/main.yml
@@ -151,10 +151,17 @@
     user: "{{ item.name }}"
     exclusive: yes
     key: "https://github.com/{{ item.name }}.keys"
-  when: item.github is defined and item.get('state', 'present') == 'present'
+  when: item.github is defined and item.get('state', 'present') == 'present' and item.authorized_keys is not defined
   register: task_result
   until: task_result|succeeded
   retries: 5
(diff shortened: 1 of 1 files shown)


playbooks/roles/user/tasks/main.yml
```diff
@@ -151,12 +151,19 @@
     user: "{{ item.name }}"
     exclusive: yes
     key: "https://github.com/{{ item.name }}.keys"
-  when: item.github is defined and item.get('state', 'present') == 'present'
+  when: item.github is defined and item.get('state', 'present') == 'present' and item.authorized_keys is not defined
```
feanil, line 154: Why not do both, if they have github keys get them, but also authorize any other keys they have.x

playbooks/roles/user/tasks/main.yml
```diff
@@ -151,12 +151,19 @@
...
   register: task_result
   until: task_result|succeeded
   retries: 5
   with_items: "{{ user_info }}"
 
+- name: set authorized_keys for users that have them defined in ansible
```
feanil, line 160: This only allows for one authorized key, can we make it so you can have a list of authorized keys in the config and it installs all of them?

PR #4148, opened 2017-10-16
set no-cache on 404s on static files

fredsmith:
Cloudflare caches 404 responses for 5 minutes if no cache-control header is set, but they say that they will expire 404's immediately if they are given a cache-control: no-cache header.  This will hopefully set that value properly.

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/maps.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/maps.j2
@@ -1,0 +1,12 @@
+# nginx maps are defined at the top level and are global
+
+# cache header for static files
+map $status $cache_header_long_lived {
+  default   "max-age=315360000";
+  404       "no-cache";
+}
+
+map $status $cache_header_short_lived {
+  default   "max-age=300";
+  404       "no-cache";
+}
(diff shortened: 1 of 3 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -22,11 +22,18 @@ map $http_user_agent {{ "$limit_bot_" ~ agent.alias }} {
...
       }
 
+
 limit_req_zone {{ "$limit_bot_" ~ agent.alias }} zone=agents:10m rate={{ agent.rate }};
 {% endfor %}
 
 {%- endif %}
 
+# cache header for static files
+map $status $cache_header {
```
feanil: Should this also go in the LMS?  Or in some shared space so the CMS config.

feanil (commented):
Please look to see which files include the static-files.j2 template and update all of them to either have the may or to include a shared maps template.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/maps.j2
```diff
@@ -0,0 +1,8 @@
+# nginx maps are defined at the top level and are global
+
+# cache header for static files
+map $status $cache_header {
+  default   "max-age=315360000";
```
feanil: Add a comment about how long this is.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/static-files.j2
```diff
@@ -40,8 +40,8 @@
 
     {% include "static-files-extra.j2" ignore missing %}
 
-    # Expire other static files immediately (there should be very few / none of these)
-    expires 300;
+    # Non-hashed files (there should be very few / none of these)
+    add_header "Cache-Control" $cache_header always;
```
feanil: We should use a different map for the fall through that defaults to 5 minutes for this again.  Otherwise we break what openedx expects.

feanil (approved):

PR #4151, opened 2017-10-16
Add tag for deploying test jenkins to build jenkins

michaelyoungstrom:
Things that need to be changed when upgrading:
- nginx config
- github oauth (security.yml file)
- splunkforwarder config

Ticket: TE-2277

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_common/tasks/main.yml
+++ b/playbooks/roles/jenkins_common/tasks/main.yml
@@ -164,10 +164,21 @@
     owner: '{{ jenkins_common_user }}'
     group: '{{ jenkins_common_group }}'
   with_items: '{{ jenkins_common_non_plugin_template_files }}'
+  register: templates_copied
   tags:
     - install
     - install:base
     - install:jenkins-configuration
+
+- name: Update Github OAUTH settings when promoting jenkins instance to production
+  template:
+    src: '{{ role_path }}/templates/config/security.yml.j2'
+    dest: '{{ jenkins_common_config_path }}/security.yml'
+    owner: '{{ jenkins_common_user }}'
+    group: '{{ jenkins_common_group }}'
+  when: '"security" in jenkins_common_non_plugin_template_files and templates_copied is not defined'
+  tags:
+    - jenkins:promote-to-production
 
 - name: Copy plugins.yml config file
   template:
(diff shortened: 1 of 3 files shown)


playbooks/roles/jenkins_common/tasks/main.yml
```diff
@@ -164,11 +164,22 @@
...
     - install:base
     - install:jenkins-configuration
 
+- name: Copy security yml file for deploy_jenkins tag
+  template:
+    src: '{{ role_path }}/templates/config/security.yml.j2'
+    dest: '{{ jenkins_common_config_path }}/security.yml'
+    owner: '{{ jenkins_common_user }}'
+    group: '{{ jenkins_common_group }}'
+  when: '"security" in jenkins_common_non_plugin_template_files and templates_copied is not defined'
```
jzoldak, line 179: I'm confused about "jenkins_common_non_plugin_template_files". What are the implications of being defined in that list? From [what is now in master](https://github.com/edx/configuration/blob/aa2d429640e034e77f147c7066ce3ad1a1da05d1/playbooks/roles/jenkins_common/defaults/main.yml#L27) it is a list of both plugin-related and non-plugin-related items.

jzoldak: After these items are changed, do we need to restart jenkins?
If so we should also tag a task that will do so.

michaelyoungstrom: If its not defined in that list, then the yaml file isnt going to get created anways (probably because the user doesnt want github oauth enabled), so we shouldnt create it here either. Good point on restarting jenkins. Ill add that.

playbooks/roles/jenkins_common/tasks/main.yml
```diff
@@ -164,11 +164,22 @@
     owner: '{{ jenkins_common_user }}'
     group: '{{ jenkins_common_group }}'
   with_items: '{{ jenkins_common_non_plugin_template_files }}'
+  register: templates_copied
   tags:
     - install
     - install:base
     - install:jenkins-configuration
 
+- name: Copy security yml file for deploy_jenkins tag
```
jzoldak: Pls rename this to something like "Update the OAUTH settings used to connect to GitHub"

jzoldak (approved):

PR #4162, opened 2017-10-19
veda web frontend role

muhammad-ammar:
Configuration Pull Request
---


muhammad-ammar:
@feanil This is the new role for video pipeline web fronted as discussed.

muhammad-ammar:
@feanil This is not finished but I can create the sandbox and most of the things are working as expected. I am fixing the minor issues. It would be great if you can give your feedback as the future work is dependent on this new role.

muhammad-ammar:
@feanil This jenkins link can be used to create platform sandbox including veda-web-frontend. Is this enough or you need anything else?
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/5701/rebuild/parameterized

muhammad-ammar:
@feanil This is ready for review.

feanil:
@muhammad-ammar how are you testing that the web frontend is working correctly?

feanil:
I tried to go to https://veda-web.sandbox.edx.org:18555 and https://veda-web.sandbox.edx.org:8555 and both failed to connect and did not return a site.

muhammad-ammar:
> I tried to go to https://veda-web.sandbox.edx.org:18555 and https://veda-web.sandbox.edx.org:8555 and both failed to connect and did not return a site.

@feanil I am not sure if the above urls should work or not? I am accessing the video pipeline frontend using https://pipeline-veda-web.sandbox.edx.org/

muhammad-ammar:
https://pipeline-veda-web.sandbox.edx.org/admin/
user: admin
password: admin

playbooks/roles/launch_ec2/tasks/main.yml
```diff
@@ -89,7 +89,7 @@
   delay: 30
   with_nested:
     - "{{ ec2.instances }}"
-    - ['studio', 'ecommerce', 'preview', 'discovery', 'credentials']
+    - ['studio', 'ecommerce', 'preview', 'discovery', 'credentials', 'pipeline']
```
feanil: why not set the DNS to `veda` instead?

playbooks/roles/video_pipeline_base/defaults/main.yml
```diff
@@ -1 +1,266 @@
...
+video_pipeline_base_gunicorn_port: 8555
+VIDEO_PIPELINE_BASE_NGINX_PORT: "1{{ video_pipeline_base_gunicorn_port }}"
+VIDEO_PIPELINE_BASE_SSL_NGINX_PORT: "4{{ video_pipeline_base_gunicorn_port }}"
+
+VIDEO_PIPELINE_BASE_URL_ROOT: "http://127.0.0.1:{{ VIDEO_PIPELINE_BASE_NGINX_PORT }}"
+
+VIDEO_PIPELINE_BASE_SOCIAL_AUTH_EDX_OIDC_KEY: "pipeline-key"
+VIDEO_PIPELINE_BASE_SOCIAL_AUTH_EDX_OIDC_SECRET: "pipeline-secret"
+VIDEO_PIPELINE_BASE_LOGOUT_URL: "{{ VIDEO_PIPELINE_BASE_URL_ROOT }}/logout/"
+VIDEO_PIPELINE_BASE_OAUTH2_URL: "{{ VIDEO_PIPELINE_BASE_URL_ROOT }}/api/val/v0"
```
feanil: Does this endpoint actually exists?  Does veda auth using the lms? or does it have its own credentials?

util/jenkins/ansible-provision.sh
```diff
@@ -204,6 +204,10 @@ CREDENTIALS_NGINX_PORT: 80
 CREDENTIALS_SSL_NGINX_PORT: 443
 CREDENTIALS_VERSION: $credentials_version
 
+VIDEO_PIPELINE_BASE_NGINX_PORT: 80
+VIDEO_PIPELINE_BASE_SSL_NGINX_PORT: 443
+VIDEO_PIPELINE_BASE_VERSION: ammar/changes-for-config-role
```
feanil: This should be set to master.

feanil (commented):
@muhammad-ammar a few comments but generally looks good.

playbooks/roles/video_pipeline_base/defaults/main.yml
muhammad-ammar: @feanil Veda need client id and client secret to communicate with val as mentioned in [Configuring VAL Access](https://openedx.atlassian.net/wiki/spaces/EDUCATOR/pages/162737033/edx-video-pipeline+testing+VEDA+Sandbox) section.

And yes, this endpoint exists. 

@yro Can you share your thoughts on this?

feanil: Cool just wanted to make sure, since I couldn't get to it.

yro: @feanil we do need this endpoint - but we need to figure out a way to setup the VEDA secure repo for this. I'll connect with you and set up some time to get this started.

feanil:
@muhammad-ammar I think if you correct the other to review points, I think this is good to go.

util/jenkins/ansible-provision.sh
muhammad-ammar: @feanil Should this merged with 
```bash
VIDEO_PIPELINE_BASE_VERSION: master
```

OR 

```bash
VIDEO_PIPELINE_BASE_VERSION: $veda_version
```

playbooks/roles/video_pipeline_base/defaults/main.yml
muhammad-ammar: Thanks @yro

util/jenkins/ansible-provision.sh
feanil: It should be set to master and we would override it in the ansible call if it needs to be some other gitref.  `$veda_version` won't work because ansible can't pull items out of the environment directly like that.

playbooks/roles/veda_web_frontend/meta/main.yml
```diff
@@ -1,4 +1,46 @@
...
+    edx_django_service_db_user: '{{ VIDEO_PIPELINE_BASE_MYSQL_USER }}'
+    edx_django_service_db_password: '{{ VIDEO_PIPELINE_BASE_MYSQL_PASSWORD }}'
+    edx_django_service_default_db_atomic_requests: false
+    edx_django_service_social_auth_edx_oidc_key: '{{ VIDEO_PIPELINE_BAS_SOCIAL_AUTH_EDX_OIDC_KEY }}'
+    edx_django_service_social_auth_edx_oidc_secret: '{{ VIDEO_PIPELINE_BAS_SOCIAL_AUTH_EDX_OIDC_SECRET }}'
+    edx_django_service_social_auth_redirect_is_https: '{{ VIDEO_PIPELINE_BAS_SOCIAL_AUTH_REDIRECT_IS_HTTPS }}'
+    edx_django_service_extra_apps: '{{ VEDA_WEB_FRONTEND_EXTRA_APPS }}'
+    edx_django_service_automated_users: '{{ VEDA_WEB_FRONTEND_AUTOMATED_USERS }}'
+    edx_django_service_use_python3: false
+    edx_django_service_config: '{{ VIDEO_PIPELINE_BASE_CONFIG }}'
```
muhammad-ammar: @feanil I want to override `video_pipeline_base_service_name: '{{ veda_web_frontend_service_name }}'`
under `role: edx_django_service` so that `VIDEO_PIPELINE_BASE_CONFIG` has correct values. Is this OK? If not then what is the best way to override?

feanil: @muhammad-ammar if `VIDEO_PIPELINE_BASE_CONFIG` is just config variables, then I think you can just pass in your override as a part of the meta include but you'll have to test that.

muhammad-ammar: Thanks @feanil

muhammad-ammar:
@feanil Today I rebased with master and squashed commits and after that sandbox creation is failing. From the [console logs](https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/6096/consoleFull), it looks like that `edx_service` role is not being executed which is a dependency of `edx_django_service` role.

Can you give a hint what might be wrong?

util/jenkins/ansible-provision.sh
```diff
@@ -349,8 +357,9 @@ EOF
     fi
 fi
 
+veda_web_frontend="true"
 declare -A deploy
-roles="edxapp forum ecommerce credentials discovery notifier xqueue xserver certs demo testcourses"
+roles="edxapp forum ecommerce credentials discovery veda_web_frontend notifier xqueue xserver certs demo testcourses"
```
muhammad-ammar, line 361: @feanil Are you OK with changes in this file for merge?

playbooks/edx-east/edx_continuous_integration.yml
```diff
@@ -39,6 +39,7 @@
     - analytics_api
     - ecommerce
     - credentials
+    - veda_web_frontend
```
muhammad-ammar, line 40: @feanil Also this change is safe to merge?

muhammad-ammar:
@feanil I have some questions. ^^

playbooks/edx-east/edx_continuous_integration.yml
feanil: I think this is good to merge only if we can build a successfull sandbox from scratch with it.  You can find the instructions on how to build from scratch here: https://openedx.atlassian.net/wiki/spaces/EdxOps/pages/28967861/Ops+Ansible+Testing+Checklist

In particular step 2 should work.

util/jenkins/ansible-provision.sh
```diff
@@ -349,8 +357,9 @@ EOF
     fi
 fi
 
+veda_web_frontend="true"
```
feanil, line 359: Why do we need this?

util/jenkins/ansible-provision.sh
feanil: This change seems fine.

muhammad-ammar:
@feanil Any help on this? [Latest build failure](https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/6300/console). I am unable to understand why `edx_service` role is not executing before `edx_django_service`?

muhammad-ammar:
@feanil If i comment out analytics_api, ecommerce, xserver, forum and credentials roles in [edx_continuous_integration.yml](https://github.com/edx/configuration/blob/master/playbooks/edx-east/edx_continuous_integration.yml) then edx_service role for veda_web_frontend role executed.

feanil:
@muhammad-ammar rather than trying to do strange dependency magic here, maybe we should try to make the base role simple and clear.  Make it a role you can call that will output the right files on disk. Here is what I think you should do:
1. Add a task to the `veda_pipeline_base` role that will let you output the templated yaml file.
2. Add common as a dependency of the `veda_pipeline_base` role.
3. Add the `veda_pipeline_base` role as an explicit dependency of the `veda_web_frontend` role.

muhammad-ammar:
@feanil The issue I am facing is that `edx_service` role tasks are not executed which is a dependency of `edx_django_service` role. So I don't know how your suggestion will execute the `edx_service` role tasks before `edx_django_service` role. I have already tried what you suggested. But I want to understand the root cause of issue I mentioned so that we can solve it instead of suppress it.

Regarding `strange dependency magic`, this is the first time I am writing ansible roles and my expectation was that experts will guide me. Anyway thanks for your reply.

feanil:
The issue you're running into is that you are trying to "render" values for the `veda_pipeline_base` role simply by referencing the role.  But the way that ansible dependency logic works means that your overrides aren't being passed in correctly.   If you added the base role to you meta dependencies with the override, I belivee it will also resolve the issue and set the right settings but again. This is much harder to reason about than just having a role that outputs a config file.

muhammad-ammar:
@feanil I have added a task in the `video_pipeline_base` that writes a config file. But `edx_service` role tasks are still not executed. Can you please take a look what I am doing wrong?

muhammad-ammar:
@feanil I have added a task in the `video_pipeline_base` that writes a config file. But `edx_service` tasks were still not executed. So I tried different things and it looks like that `ecommerce` role is doing something due to which `edx_service` role tasks are not running for `veda_web_frontend` role. I will further investigate this to find the root cause of the issue. In the mean time if you have any suggestion, please let me know.

feanil:
@muhammad-ammar someone in the opensource community also found this issue, https://github.com/edx/configuration/pull/4201  You can try that change locally to see if it fixes your issues.

playbooks/roles/veda_web_frontend/tasks/main.yml
```diff
@@ -1 +1,26 @@
...
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Tasks for role veda_web_frontend
+#
+
+
+- name: create OAuth application clients
```
clintonb: This should not be in the tasks for the service's play, but in a separate play. Otherwise, this will run, and fail, with every deployment. Also, database-related tasks should be behind a flag so that they can be disabled when building Docker images.

playbooks/roles/veda_web_frontend/meta/main.yml
```diff
@@ -1,4 +1,46 @@
...
+#
+##
+# Role includes for role veda_web_frontend
+#
 dependencies:
-  - video_pipeline_base
+  - role: video_pipeline_base
+    video_pipeline_base_service_name: '{{ veda_web_frontend_service_name }}'
+  - role: edx_django_service
+    edx_django_service_repo: 'edx-video-pipeline'
```
clintonb: Nearly everything in here should be a variable sourced from `defaults/main.yml`. Hard-coding values is not recommended.

muhammad-ammar: @clintonb Placed the hardcoded values while I was trying to figure out why `edx_service` role tasks are not running for my `veda_web_frontend` role. None of this will be committed.

playbooks/roles/veda_web_frontend/tasks/main.yml
muhammad-ammar: @clintonb I am not sure why this task needs it own role and why it will fail for every deployment? 

Also if this is behind a flag then what thing will decide when to enable/disable this flag?

muhammad-ammar: @clintonb For now I would prefer to remain this task in this role as too much time was wasted due to `edx_service` role tasks were not executed. But we can put a flag on this task but then the question where and when to enable/disable that flag?

muhammad-ammar:
@clintonb Can you please review the changes in this PR? Our initial goal is to setup a sandbox with everything ready for edx video pipeline. This PR is the first step in that direction. `video_pipeline_base`: Base role used by `veda_web_frontend` and other roles
`veda_web_frontend`: Django service which provides a frontend for edx video pipeline 

Let me know if you need any other info. Thanks

playbooks/roles/veda_web_frontend/tasks/main.yml
muhammad-ammar: My understanding developed from [feanil's comment](https://github.com/edx/configuration/pull/4162#discussion_r148696535) was that if sandbox creation is working then this is good to merge but now I am worried after your comment :)

clintonb: Wait. . .you guys have a separate OAuth provider, don't you? That's very much wrong, but out of scope at this time. You really should prioritize fixing that. For now, you should add a comment stating that you are creating a client in the VEDA application, not LMS. Otherwise, others will be confused.

playbooks/edx-east/veda_web_frontend.yml
```diff
@@ -1,5 +1,12 @@
...
   gather_facts: True
   roles:
-    - veda_web_frontend
+    - aws
+    - role: nginx
+      nginx_default_sites:
+        - veda_web_frontend
+    - role: veda_web_frontend
+    - role: splunkforwarder
+      when: COMMON_ENABLE_SPLUNKFORWARDER
```
clintonb, line 12: What about New Relic?

playbooks/roles/veda_web_frontend/meta/main.yml
clintonb: Okay. If you want a review, I want to see Ansible that is pretty much complete. I don't want to fall into the back-and-forth review cycle.

playbooks/roles/veda_web_frontend/meta/main.yml
```diff
@@ -1,4 +1,46 @@
...
+    edx_django_service_version: '{{ VIDEO_PIPELINE_BASE_VERSION }}'
```
clintonb: This is quite confusing. The `veda_web_frontend` play should not be dependent on the `video_pipeline_base` play. What are you trying to achieve with the `video_pipeline_base` play?

playbooks/roles/video_pipeline_base/defaults/main.yml
```diff
@@ -1 +1,263 @@
...
+  ## Encoding Config
+  ffmpeg_compiled: "ffmpeg"
+  ffprobe_compiled: "ffprobe"
+  target_aspect_ratio: 1.7777778
+
+  # ----------
+  ##---
+  # This is a list of encodes and their respective course
+  # boolean matches
+  encode_dict:
```
clintonb: Does this actually need to be in Ansible? Why not make these the defaults in your application's settings?

util/jenkins/ansible-provision.sh
```diff
@@ -304,6 +308,10 @@ COURSE_DISCOVERY_ECOMMERCE_API_URL: "https://ecommerce-${deploy_host}/api/v2"
 DISCOVERY_URL_ROOT: "https://discovery-${deploy_host}"
 DISCOVERY_SOCIAL_AUTH_REDIRECT_IS_HTTPS: true
 
+VIDEO_PIPELINE_DOMAIN: "pipeline-${deploy_host}"
+VIDEO_PIPELINE_BASE_URL_ROOT: "https://{{ VIDEO_PIPELINE_DOMAIN }}"
+EDX_LMS_BASE_URL: "https://{{ EDXAPP_LMS_BASE }}"
```
clintonb: Namespace this appropriately, as we do with `ECOMMERCE_LMS_URL_ROOT` above.

clintonb (changes requested):
I'm not really sure what's going on here. Why are there two new roles when there is only one service? Is your plan to make another dependent role? At present, the mixing of the two roles' variables is confusing, and will probably lead to errors.

Also, I recommend creating a `Dockerfile` so that you can test your Ansible builds locally rather than relying on sandbox builds. This has the added benefit of producing an image that can later be incorporated in Docker Devstack, better enabling local development.

playbooks/roles/veda_web_frontend/tasks/main.yml
muhammad-ammar: I have added a comment. Can you suggest how we should improve this? This OAuth client is required by different video pipeline workers to communicate with video pipeline. 
We have inherited this edx video pipeline and other related things and we are improving it along with adding new features.

playbooks/edx-east/veda_web_frontend.yml
muhammad-ammar: As discussed with devops, new relic is not needed now. Will decide whether we include it or not?

playbooks/roles/veda_web_frontend/meta/main.yml
muhammad-ammar: `video_pipeline_base` is the base role will be used by other video pipeline roles. Can you please mention what is confusing so that we can make it more clear?

playbooks/roles/video_pipeline_base/defaults/main.yml
muhammad-ammar: Good catch. Will do in separate PR.

muhammad-ammar:
@clintonb I have tried to address your feedback. Please let me know if something is still missing.

playbooks/edx-east/veda_web_frontend.yml
clintonb: Why is New Relic not needed now? How will you monitor the service and deal with debugging? New Relic seems like a minor addendum that will have a huge potential payout. If you aren’t going to use it, please add a comment explaining why in this file.

playbooks/roles/veda_web_frontend/meta/main.yml
clintonb: Typically (at least in my limited experience in this code base), roles are independent of one another, unless inherited via meta, as you are doing with edx_django_service. 

Sharing variables across roles is confusing because it implies a dependency, but it’s not clear why there is a dependency. What other roles do you intend to build atop video_pipeline_base? Why?

playbooks/edx-east/oauth_client_setup.yml
```diff
@@ -9,5 +9,6 @@
     - "roles/ecommerce/defaults/main.yml"
     - "roles/credentials/defaults/main.yml"
     - "roles/discovery/defaults/main.yml"
+    - "roles/video_pipeline_base/defaults/main.yml"
```
clintonb: This is where some of the confusion comes in. You say video_pipeline_base is a base/dependency role, yet here it has more precedence than the veda role. Those two things don’t align.

playbooks/roles/veda_web_frontend/defaults/main.yml
```diff
@@ -1 +1,63 @@
...
+
+VEDA_WEB_FRONTEND_REPO: "edx-video-pipeline"
+
+VEDA_WEB_FRONTEND_VERSION: "master"
+
+VEDA_WEB_FRONTEND_MEMCACHE: []
+
+VEDA_WEB_FRONTEND_DJANGO_SETTINGS_MODULE: "VEDA.settings.production"
+
+VEDA_WEB_FRONTEND_SOCIAL_AUTH_REDIRECT_IS_HTTPS: false
```
clintonb: This should default to true.

playbooks/roles/veda_web_frontend/meta/main.yml
```diff
@@ -1,4 +1,46 @@
...
+    edx_django_service_db_user: '{{ VIDEO_PIPELINE_BASE_MYSQL_USER }}'
+    edx_django_service_db_password: '{{ VIDEO_PIPELINE_BASE_MYSQL_PASSWORD }}'
+    edx_django_service_default_db_atomic_requests: false
+    edx_django_service_social_auth_edx_oidc_key: '{{ VIDEO_PIPELINE_BASE_SOCIAL_AUTH_EDX_OIDC_KEY }}'
+    edx_django_service_social_auth_edx_oidc_secret: '{{ VIDEO_PIPELINE_BASE_SOCIAL_AUTH_EDX_OIDC_SECRET }}'
+    edx_django_service_social_auth_redirect_is_https: '{{ VIDEO_PIPELINE_BASE_SOCIAL_AUTH_REDIRECT_IS_HTTPS }}'
+    edx_django_service_extra_apps: '{{ VEDA_WEB_FRONTEND_EXTRA_APPS }}'
+    edx_django_service_automated_users: '{{ VEDA_WEB_FRONTEND_AUTOMATED_USERS }}'
+    edx_django_service_use_python3: false
+    edx_django_service_config: !!null
```
clintonb: Pass a variable, even if it’s an empty dictionary. Otherwise, you’ll never be able to override Django settings without modifying this file.

playbooks/roles/veda_web_frontend/tasks/main.yml
```diff
@@ -1 +1,26 @@
...
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Tasks for role veda_web_frontend
+#
+
+# This is creating a client in VEDA application not LMS.
+- name: create OAuth application clients
+  shell: >
+    {{ veda_web_frontend_venv_dir }}/bin/python {{ COMMON_BIN_DIR }}/manage.{{ veda_web_frontend_service_name }} create_oauth_client --settings={{ VEDA_WEB_FRONTEND_DJANGO_SETTINGS_MODULE }}
```
clintonb: 1. Why does your manage.py file an extension that isn’t .py?
2. You don’t need to specify settings. See the post-migration commands in edx_django_service and ecommerce

playbooks/edx-east/veda_web_frontend.yml
muhammad-ammar: @feanil Can better answer this.

clintonb (changes requested):
Please schedule some time to chat live. That will be faster than commenting back and forth here. I want to understand your future vision for these new roles.

playbooks/roles/veda_web_frontend/meta/main.yml
muhammad-ammar: [veda_delivery_workder](https://github.com/edx/configuration/tree/master/playbooks/roles/veda_delivery_worker) and [veda_intake_worker](https://github.com/edx/configuration/tree/master/playbooks/roles/veda_intake_worker) will be build on top of `video_pipeline_base` role. These workers code live inside the edx-video-pipeline and all three share the same configuration.

playbooks/roles/veda_web_frontend/tasks/main.yml
muhammad-ammar: I am following this pattern. https://github.com/edx/configuration/blob/master/playbooks/roles/oauth_client_setup/tasks/main.yml#L26

`manage.veda` is a symlink to manage.py. All these symlinks exist in `/edx/bin/`

muhammad-ammar:
@clintonb I have addressed more feedback. Please have a look.

clintonb:
@muhammad-amar, you still haven’t answered my questions about the two separate roles. This is the point where a live conversation is needed to resolve the differences. Going back and forth over comments will only lead to lost time. Please schedule a thirty minute meeting so we can work through this issue and get you moving forward.

clintonb:
@muhammad-ammar ^

muhammad-ammar:
@clintonb I have refactored the things like you suggested. Please have a look to verify if I did things correctly?

One question, So now when we want to override a value then we will use `VIDEO_PIPELINE_BASE` variables to override it?

muhammad-ammar:
@clintonb Any idea about this [sandbox failure](https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/6581/console)? I have added `veda_web_frontend defaults yml` in [oauth_client_setup.yml](https://github.com/edx/configuration/pull/4162/files#diff-697bf97608b67eaa1f1148cbbefecd50R12) but even then its failing.

playbooks/roles/veda_web_frontend/defaults/main.yml
```diff
@@ -1 +1,73 @@
...
+
+VEDA_WEB_FRONTEND_DEFAULT_DB_NAME: '{{ VIDEO_PIPELINE_BASE_DEFAULT_DB_NAME }}'
+VEDA_WEB_FRONTEND_MYSQL_HOST: '{{ VIDEO_PIPELINE_BASE_MYSQL_HOST }}'
+VEDA_WEB_FRONTEND_MYSQL_USER: '{{ VIDEO_PIPELINE_BASE_MYSQL_USER }}'
+VEDA_WEB_FRONTEND_MYSQL_PASSWORD: '{{ VIDEO_PIPELINE_BASE_MYSQL_PASSWORD }}'
+
+VEDA_WEB_FRONTEND_OAUTH2_URL: '{{ VIDEO_PIPELINE_BASE_URL_ROOT }}/api/val/v0'
+VEDA_WEB_FRONTEND_LOGOUT_URL: '{{ VIDEO_PIPELINE_BASE_URL_ROOT }}/logout/'
+VEDA_WEB_FRONTEND_SOCIAL_AUTH_EDX_OIDC_KEY: '{{ VIDEO_PIPELINE_BASE_SOCIAL_AUTH_EDX_OIDC_KEY | default("pipeline-key") }}'
+VEDA_WEB_FRONTEND_SOCIAL_AUTH_EDX_OIDC_SECRET: '{{ VIDEO_PIPELINE_BASE_SOCIAL_AUTH_EDX_OIDC_SECRET | default("pipeline-secret") }}'
```
muhammad-ammar, line 72: @clintonb The above variables are declared this way because this file is included in [oauth_client_setup.yml](https://github.com/edx/configuration/pull/4162/files#diff-697bf97608b67eaa1f1148cbbefecd50) and variables are used in [oauth_client_setup/defaults/main.yml](https://github.com/edx/configuration/pull/4162/files#diff-0ab86397f50d79e8a022619b287997ce)

muhammad-ammar:
@clintonb This is final from my side. Your final feedback is required.

playbooks/roles/veda_web_frontend/meta/main.yml
```diff
@@ -1,4 +1,38 @@
...
 dependencies:
-  - video_pipeline_base
+  - role: video_pipeline_base
+    video_pipeline_base_service_name: '{{ veda_web_frontend_service_name }}'
+  - role: edx_django_service
+    edx_django_service_repo: '{{ VEDA_WEB_FRONTEND_REPO }}'
+    edx_django_service_version: '{{ VEDA_WEB_FRONTEND_VERSION }}'
+    edx_django_service_name: '{{ veda_web_frontend_service_name }}'
+    edx_django_service_user: '{{ veda_web_frontend_user }}'
+    edx_django_service_home: '{{ COMMON_APP_DIR }}/{{ veda_web_frontend_service_name }}'
```
clintonb: There is no need to override this.

playbooks/roles/video_pipeline_base/tasks/main.yml
```diff
@@ -1 +1,7 @@
 ---
+
+- name: Write out app config file
```
clintonb, line 3: This is not necessary, and will override the file written by `edx_django_service`.

playbooks/roles/video_pipeline_base/meta/main.yml
```diff
@@ -1 +1,4 @@
 ---
+
+dependencies:
+  - common
```
clintonb: This is not necessary. The `video_pipeline_base` play should only be a container of common variables. It shouldn't actually perform any tasks. `edx_django_service` already depends on `common`.

playbooks/roles/video_pipeline_base/templates/config.yml.j2
```diff
@@ -0,0 +1,4 @@
+---
+# {{ ansible_managed }}
```
clintonb: Remove.

util/jenkins/ansible-provision.sh
clintonb: This question needs to be addressed, or this line removed. If this is needed, please add a comment explaining why, so future engineers don't have to dig through PR comments.

playbooks/roles/video_pipeline_base/defaults/main.yml
```diff
@@ -1 +1,184 @@
...
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for common variables for video_pipeline. This is the base role for all video pipeline roles.
+#
+
+video_pipeline_base_service_name: "video_pipeline_base"
+
+video_pipeline_base_val_base_url: "{{ VEDA_LMS_BASE_URL | default('http://127.0.0.1:8000') }}"
```
clintonb: The prefix should correspond to the play name: `VIDEO_PIPELINE_BASE_LMS_BASE_URL`

util/jenkins/ansible-provision.sh
```diff
@@ -204,6 +204,10 @@ CREDENTIALS_NGINX_PORT: 80
 CREDENTIALS_SSL_NGINX_PORT: 443
 CREDENTIALS_VERSION: $credentials_version
 
+VIDEO_PIPELINE_BASE_NGINX_PORT: 80
+VIDEO_PIPELINE_BASE_SSL_NGINX_PORT: 443
+VEDA_WEB_FRONTEND_VERSION: master
```
clintonb: The value should be derived from Jenkins input. This file is used by the sandbox creation job.

clintonb (changes requested):

util/jenkins/ansible-provision.sh
muhammad-ammar: Currently jenkins UI has no support to do this. Will update this when we add the support.

clintonb: If you aren't going to add it (and I'm not sure why you wouldn't), remove the override.

muhammad-ammar:
> This question needs to be addressed, or this line removed. If this is needed, please add a comment explaining why, so future engineers don't have to dig through PR comments.

`veda_web_frontend="true"` is needed to run the `veda_web_frontend` role. Please see the below shell script snippet from [ansible-provision.sh](https://github.com/edx/configuration/pull/4162/files#diff-32b2548c4c6e77f9c5a605464f58b6a4R361)

```bash
veda_web_frontend="true"
declare -A deploy
roles="edxapp forum ecommerce credentials discovery veda_web_frontend notifier xqueue xserver certs demo testcourses"

for role in $roles; do
    deploy[$role]=${!role}
done
```

As currently jenkins UI has not support to select veda role that's why `veda_web_frontend` variable is set manually here. Other role variables like credentials, discovery are set automatically when you select them from jenkins UI.

@clintonb @feanil

util/jenkins/ansible-provision.sh
muhammad-ammar: @clintonb I can't do/add everything in a single PR. Will definitely do once time comes for that work.

muhammad-ammar:
@clintonb Addressed your feedback.

playbooks/roles/video_pipeline_base/tasks/main.yml
muhammad-ammar: @clintonb Now after removing this, things are messed up again.

`VIDEO_PIPELINE_BASE_CONFIG` is the config that will be written to a file. Now what we do is that we define something like below in `veda_web_frontend` role
```yaml
VEDA_WEB_FRONTEND_SERVICE_CONFIG: '{{ VIDEO_PIPELINE_BASE_CONFIG }}'
```
and then pass `VEDA_WEB_FRONTEND_SERVICE_CONFIG` to `edx_django_service` in `veda_web_frontend/meta/main.yml`. `VIDEO_PIPELINE_BASE_CONFIG` has [STATIC_ROOT](https://github.com/edx/configuration/pull/4162/files#diff-6058521ab03b79de7ee733b8002cc56fR100) . Now the value of `STATIC_ROOT` is constructed to `/edx/var/video_pipeline_base/staticfiles` which is wrong, instead it should be `/edx/var/veda/staticfiles`, This is happening because variables are not overrided when we do below in `veda_web_frontend` role
```yaml
VEDA_WEB_FRONTEND_SERVICE_CONFIG: '{{ VIDEO_PIPELINE_BASE_CONFIG }}'
```

Due to this very reason, we created the task in video_pipeline_base to write the config file and this file will not be overriden by `edx_django_service` because we have disabled the config writing task in the `veda_web_frontend` meta file.

I am soooo tired of this ansible stuff. Too much time has wasted and no elegant solution to this whole unique role is produced. I am re-adding this task and will stay with it unless someone has a better solution to this whole problem.

muhammad-ammar:
@clintonb @feanil I am planning to merge this today. I think it is good to go. Do you have any serious concerns?

playbooks/roles/video_pipeline_base/tasks/main.yml
clintonb: Ideally, you should override the static variables, but do what you need to do.

I understand your frustration, but we have best practices, templates like the edx_django_service play, and cookiecutter project templates to avoid this frustration. I'm not blaming you but, the truth is all of this frustration is due to the fact that VEDA has never adhered to any of our best practices. If the team maintaining this wants to avoid future frustration, you should really *invest* the time to do things as we do them with the other IDAs. All of them follow the same pattern for Ansible, Docker, sandboxes, deployment, etc.

clintonb (approved):

playbooks/edx-east/edx_continuous_integration.yml
feanil: @muhammad-ammar did you run through all the tests on this checklist?

playbooks/edx-east/veda_web_frontend.yml
feanil: @clintonb good question, we are not installing it initially because the initial goal is to get the machines built from ansible in the first place and get splunk setup.  New Relic was not on the table initially because there are multiple pieces of VEDA that are not straightforward to instrument and interrelated so we are going to follow up with that later.

+1 to adding a comment in here explaining why we don't have new relic in this list.

playbooks/roles/edx_ansible/templates/update.j2
```diff
@@ -14,7 +14,7 @@ IFS=","
 
     <repo> - must be one of edx-platform, edx-workers, xqueue, cs_comments_service, credentials, xserver, configuration,
     read-only-certificate-code, edx-analytics-data-api, edx-ora2, insights, ecommerce, course_discovery,
-    notifier
+    notifier, video_pipeline
```
feanil: This is not the entire pipeline, this is just the web frontend we should name it as such in this file to clearly map it back to the service and role name.

playbooks/roles/edx_ansible/templates/update.j2
```diff
@@ -61,6 +61,7 @@ repos_to_cmd["insights"]="$edx_ansible_cmd insights.yml -e 'INSIGHTS_VERSION=$2'
 repos_to_cmd["ecommerce"]="$edx_ansible_cmd ecommerce.yml -e 'ECOMMERCE_VERSION=$2'"
 repos_to_cmd["discovery"]="$edx_ansible_cmd discovery.yml -e 'DISCOVERY_VERSION=$2'"
 repos_to_cmd["notifier"]="$edx_ansible_cmd notifier.yml -e 'NOTIFIER_VERSION=$2'"
+repos_to_cmd["video_pipeline"]="$edx_ansible_cmd veda_web_frontend.yml -e 'VEDA_WEB_FRONTEND_VERSION=$2'"
```
feanil: Same as previous comment.

playbooks/roles/veda_web_frontend/defaults/main.yml
```diff
@@ -1 +1,73 @@
...
+# wiki:       https://openedx.atlassian.net/wiki/display/OpenOPS
+# code style: https://openedx.atlassian.net/wiki/display/OpenOPS/Ansible+Code+Conventions
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+##
+# Defaults for role veda_web_frontend
+#
+
+
+veda_web_frontend_service_name: "veda"
```
feanil: Should we be calling the service `veda_web` instead of just `veda` since that refers to the whole suite.

playbooks/roles/veda_web_frontend/defaults/main.yml
```diff
@@ -1 +1,73 @@
...
+veda_web_frontend_venv_dir: "{{ veda_web_frontend_service_home }}/venvs/{{ veda_web_frontend_service_name }}"
+
+#
+# OS packages
+#
+
+veda_web_frontend_debian_pkgs:
+  - libffi6
+  - libffi-dev
+  - ffmpeg
```
feanil: Why is this needed on the web frontends?

playbooks/roles/veda_web_frontend/tasks/main.yml
feanil: Python is also symlinked in `{{ COMMON_BIN_DIR }}` let's update the call to python to match the call to manage.py

playbooks/roles/video_pipeline_base/meta/main.yml
feanil: +1 remove this dependency.

feanil (commented):
It looks like there were questions still un-answered from my last review as well but I've just done another pass.

muhammad-ammar:
> It looks like there were questions still un-answered from my last review as well but I've just done another pass.

@feanil I have already answered all questions from your last review. Can you share the link to your question which was not answered?

muhammad-ammar:
> @muhammad-ammar did you run through all the tests on this checklist?

@feanil Last time you said that only step 2 is enough to merge this. [Your comment](https://github.com/edx/configuration/pull/4162#discussion_r148696535) 
I have already created sandbox as per step 2 and it works.

playbooks/roles/veda_web_frontend/defaults/main.yml
muhammad-ammar: I will prefer to remain this as it is. I will happy to rename this in a future PR if this is really a concern.

muhammad-ammar:
@feanil Addressed your feedback and built the [sandbox](https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/6987/console) as per step 2.

playbooks/edx-east/veda_web_frontend.yml
feanil: @muhammad-ammar  still waiting for a comment here.

playbooks/edx-east/edx_continuous_integration.yml
feanil: @muhammad-ammar maybe you answered this in a different place?  I'm usually looking at the inline responses since it's easier to keep context that way.

playbooks/roles/veda_web_frontend/defaults/main.yml
feanil: The best practice is that the service name and the role name match up, in this case the role and vars are all `veda_web_frontend` and the service name is `veda`  I think updating both to `veda_web` is the right move.  I'm fine with doing this as a follow on ticket.  Please add a TODO above this line with a comment and a link to the ticket that will do this work.

playbooks/roles/video_pipeline_base/templates/config.yml.j2
feanil: @muhammad-ammar this still needs to happen.

util/jenkins/ansible-provision.sh
feanil: This question is also still outstanding @muhammad-ammar

muhammad-ammar: @feanil https://github.com/edx/configuration/pull/4162#issuecomment-345139371

playbooks/edx-east/veda_web_frontend.yml
muhammad-ammar: @feanil already added. https://github.com/edx/configuration/pull/4162/files#diff-079fe0eafbbd4576346fed9a23124b03R13

playbooks/roles/veda_web_frontend/defaults/main.yml
muhammad-ammar: @feanil Added [TODO](https://github.com/edx/configuration/pull/4162/files#diff-934dc87706379ff0c936c3a872e92a08R14)

muhammad-ammar:
@feanil Addressed your feedback. Let me know if there is something missing.

feanil (approved):

PR #4163, opened 2017-10-19
ENT-594 Put ENTERPRISE_REPORTING_SECRET into the lms AUTH_TOKENS settings

brittneyexline:
This value will be read from the edx-secure config (https://openedx.atlassian.net/browse/DEVOPS-6660), put into the auth tokens for lms (here), and put into the django settings which are used for encrypting the passwords for enterprise reporting configs introduced by https://github.com/edx/edx-platform/pull/16265.

Configuration Pull Request
---


fredsmith (changes requested):
you should set a default for EDXAPP_ENTERPRISE_REPORTING_SECRET, otherwise this will cause errors when it isn't included in a secure override.

fredsmith (approved):

PR #4174, opened 2017-10-25
Setting sql_mode for IDAs to prevent data loss

clintonb:
This setting ensures truncation warnings are escalated as errors. See https://docs.djangoproject.com/en/1.10/ref/databases/#mysql-sql-mode for additional info.


clintonb:
@edx/devops @edx/learner please review

FYI @edx/platform-team @edx/analytics you'll want to do the same for other services.

feanil (approved):

feanil:
@doctoryes something like this should also be done for edx-platform as well.  Can you or someone from the platform team put in a PR?

doctoryes:
@feanil Will do.

doctoryes:
https://openedx.atlassian.net/browse/PLAT-1781 was created to convert edxapp.

PR #4194, opened 2017-11-03
jenkins_common: Programmatically use forked github-api plugin

michaelyoungstrom:
There's a bug in the github-api plugin due to a change in github's API. This is being fixed in v1.90, but is not yet available in the maven repository (which is currently where we get our plugin jpi files). So for now, I've added our version to S3, and updated the ansible to pull the file from there.

Before pulling down our custom version, we have to make sure there aren't any existing github-api files already in the plugins dir, which get downloaded because its a dependency of other git plugins we use.

Configuration Pull Request
---


estute (approved):
:+1:

PR #4221, opened 2017-11-28
Ibrahimahmed443/wl 1043 sandbox automation

ibrahimahmed443:
Configuration Pull Request
---


ibrahimahmed443:
@bill-filler @afzaledx Here's the work in progress related to the ansible scripts.

feanil:
@ibrahimahmed443 this is very specific to edx, our specific DNS setting for whitelabels shouldn't be exposed here.  There should be variables that we override in sandbox-internal that allows us to add more sites.

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -53,7 +53,7 @@ DISCOVERY_ELASTICSEARCH_INDEX_NAME: 'catalog'
 
 DISCOVERY_MEMCACHE: [ 'memcache' ]
 
-DISCOVERY_VERSION: "master"
+DISCOVERY_VERSION: "hasnain-naveed/WL-1295"
```
fredsmith: why?

playbooks/roles/discovery/defaults/main.yml
```diff
@@ -122,7 +122,7 @@ DISCOVERY_REPOS:
     DOMAIN: "{{ COMMON_GIT_MIRROR }}"
     PATH: "{{ COMMON_GIT_PATH }}"
     REPO: 'course-discovery.git'
-    VERSION: "{{ DISCOVERY_VERSION }}"
+    VERSION: "hasnain-naveed/WL-1295"
```
fredsmith: you're setting this to the same as the above.  seems unnecessary.  Also, we shouldn't change the default to a dev branch.

ibrahimahmed443: This is because DISCOVERY_VERSION is overridden from somewhere else. So I had to set this explicitly here. We will remove this code anyway, as the branch gets merged.

ibrahimahmed443:
@fredsmith its ready for another look.

playbooks/roles/whitelabel/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+  become_user: "{{ edxapp_user }}"
+  environment: "{{ edxapp_environment }}"
+  with_items: "{{ WHITELABEL_ORG }}"
+
+# Discovery
+
+- name: Run discovery playbook
+  become: True
+  shell: >
+    /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook -i localhost, -c local
```
fredsmith: the relevant parts of this playbook should either be included in your playbook, or should have their roles meta'ed in as dependencies.

playbooks/roles/whitelabel/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+WHITELABEL_THEME_DIR: "edx-themes"
+
+WHITELABEL_COURSE_TITLE: "test_course"
+WHITELABEL_COURSE_RUN: "2017"
+WHITELABEL_MODULE_STORE: "split"
+
+WHITELABEL_MYSQL_COMMANDS:
+  - "CREATE SCHEMA {{ DISCOVERY_DEFAULT_DB_NAME }};"
+  - "CREATE USER '{{ DISCOVERY_MYSQL_USER }}'@'{{ DISCOVERY_MYSQL }}' IDENTIFIED BY '{{ DISCOVERY_MYSQL_PASSWORD }}';"
+  - "GRANT ALL PRIVILEGES ON {{ DISCOVERY_DEFAULT_DB_NAME }}.* TO '{{ DISCOVERY_MYSQL_USER }}'@'{{ DISCOVERY_MYSQL }}';"
```
fredsmith: the app user should only have the required permissions, not all permissions.

playbooks/edx-east/whitelabel.yml
```diff
@@ -0,0 +1,10 @@
+- name: Setup Whitelabel themes
+  hosts: all
+  become: True
+  gather_facts: True
+  vars_files:
+    - "roles/edxapp/defaults/main.yml"
```
fredsmith: I'm not crazy about this, but I see that there is prior art in edx_provision.yml for including another role's vars.    The right way to do this is to have a vars role that both roles include, but I think that is out of scope here.

fredsmith (changes requested):
We shouldn't be calling ansible-playbook -c local from within a playbook.  Please find a way to do that using roles.

bill-filler:
@ibrahimahmed443 
please address the comments that Fred made last week so we can get this landed

ibrahimahmed443:
@fredsmith I copied the required part from discovery into my playbook [here](https://github.com/edx/configuration/pull/4221/commits/735a0f393d5d19dd8783b2113d062f9a2c62f25c); however, by doing so I started getting the same errors as  we would if we setup discovery by ticking the discovery checkbox from Jenkins. That is, "Access denied for user 'migrate'@'localhost' to database 'discovery'".

If I run the discovery playbook directly as I was doing earlier, this runs successfully. I am not fully aware of the history of why this happens, but it was documented by another team that discovery doesn't work by ticking the checkbox, rather we have to run the playbook after setting up the sandbox. And I was doing a similar thing by ansible-playbook -c local.

fredsmith:
@ibrahimahmed443 have you tried to make sure that the edxlocal role is creating the correct databases before your discovery playbook is run?  

The database appears to be defined in here:  https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/defaults/main.yml#L16

The discoery database should be created (https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/tasks/main.yml#L13) and granted access (https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/tasks/main.yml#L30)

So if that task is run before your failed task above, something is not being passed properly into your task.  If that task is not run before your failed task, you should figure out how to structure the meta dependencies to make it happen first.

bill-filler:
@ibrahimahmed443 
courses no longer configured for WL sites. I see none on HMS, Wharton and mitxpro after configuring the sandbox

ibrahimahmed443:
@bill-filler On Friday, Kashif pointed out that we should be able to specify the "course display name/title" and we found out the management command used to create course does not support that. We created a new branch in LMS: hasnain-naveed/create_course_change which makes this change. Hence for now, while creating a sandbox, this branch needs to be specified in the edxapp version field.

ibrahimahmed443:
@fredsmith Can you give a thumbs up on this PR?

fredsmith (approved):
Looks great 👍

PR #4232, opened 2017-12-04
OPS-2492 Support redis as an ecommerce celery backend in edx/configur…

coryleeio:
…ation Ansible

Configuration Pull Request
---


playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -145,11 +145,11 @@ ECOMMERCE_STATICFILES_STORAGE: 'ecommerce.theming.storage.ThemeStorage'
 # Celery
 ECOMMERCE_BROKER_USERNAME: 'celery'
 ECOMMERCE_BROKER_PASSWORD: 'celery'
-# Used as the default RabbitMQ IP.
 ECOMMERCE_BROKER_HOST: '{{ ansible_default_ipv4.address }}'
-# Used as the default RabbitMQ port.
 ECOMMERCE_BROKER_PORT: 5672
-ECOMMERCE_BROKER_URL: 'amqp://{{ ECOMMERCE_BROKER_USERNAME }}:{{ ECOMMERCE_BROKER_PASSWORD }}@{{ ECOMMERCE_BROKER_HOST }}:{{ ECOMMERCE_BROKER_PORT }}'
+ECOMMERCE_CELERY_BROKER_TRANSPORT: 'amqp'
+ECOMMERCE_CELERY_BROKER_HOSTNAME: '{{ ECOMMERCE_BROKER_HOST }}'
```
feanil: NIT: You should flip this around so the default value for `ECOMMERCE_BROKER_HOST` is `{{ ECOMMERCE_CELERY_BROKER_HOST }}` and then set the default value for `ECOMMERCE_CELERY_BROKER_HOST` to `{{ ansible_default_ipv4.address }}`

That way when it comes time to delete things we can just remove the line for the old var and not have to do any other updates.

feanil (approved):
One nit to fix if you want.

playbooks/roles/ecommerce/defaults/main.yml
jdmulloy: @feanil How will this affect configurations that are currently setting `ECOMMERCE_BROKER_HOST`? What Cory did here is similar to what I did in #4189 for backwards compatibility.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -145,11 +145,11 @@ ECOMMERCE_STATICFILES_STORAGE: 'ecommerce.theming.storage.ThemeStorage'
...
+ECOMMERCE_BROKER_URL: '{{ ECOMMERCE_CELERY_BROKER_TRANSPORT }}://{{ ECOMMERCE_BROKER_USERNAME }}:{{ ECOMMERCE_BROKER_PASSWORD }}@{{ ECOMMERCE_CELERY_BROKER_HOSTNAME }}:{{ ECOMMERCE_BROKER_PORT }}'
```
jdmulloy: Some of these variables say Celery and some don't. Can and/or should we rework it to make them consistent? Unfortunately it will require some changes to existing configs and some backwards compatibility. So I would like opinions on whether or not it's worth it.

jdmulloy (changes requested):

playbooks/roles/ecommerce/defaults/main.yml
coryleeio: Took care of this by removing celery so it would be consistent with the other ecommerce variables if not edxapp, for consistency as we discussed.

playbooks/roles/ecommerce/defaults/main.yml
coryleeio: I went ahead and flipped them for ease, I'd say it's not a big deal if they're not perfectly consistent.  The tickets are pretty straight forward anyway, just having you delete a variable or two for cleanup

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -145,11 +145,11 @@ ECOMMERCE_STATICFILES_STORAGE: 'ecommerce.theming.storage.ThemeStorage'
 # Celery
 ECOMMERCE_BROKER_USERNAME: 'celery'
 ECOMMERCE_BROKER_PASSWORD: 'celery'
-# Used as the default RabbitMQ IP.
-ECOMMERCE_BROKER_HOST: '{{ ansible_default_ipv4.address }}'
-# Used as the default RabbitMQ port.
+ECOMMERCE_BROKER_HOST: '{{ ECOMMERCE_BROKER_HOSTNAME }}'
```
jdmulloy: Now that the diff has been simplified, I'm wondering if we need to rename `ECOMMERCE_BROKER_HOST` to `ECOMMERCE_BROKER_HOSTNAME`

jdmulloy (changes requested):

playbooks/roles/ecommerce/defaults/main.yml
coryleeio: We need to keep ECOMMERCE_BROKER_HOST for backwards compatibility until hawthorne I believe

playbooks/roles/ecommerce/defaults/main.yml
coryleeio: @feanil @jdmulloy I realized actually you can't flip them unless you continue to use HOST in the URL instead of hostname, otherwise you will overwrite what the user may have set HOST to with HOSTNAME

example:
If we use the new value of HOSTNAME in the URL
user has legacy value of HOST set to 10.10.10.10
we set HOSTNAME to '{{ ansible_default_ipv4.address }}' , they dont have HOSTNAME in the config so in the URL we just changed their host to localhost.

We could continue to key the URL off of the old value HOST to fix this, however then you can't just delete the variable anymore.   I'm going to revert this change I made to align it back with the way the other variable is being done for consistency and to not break backwards compatibility.

playbooks/roles/ecommerce/defaults/main.yml
coryleeio: After discussion we removed Hostname keeping only host.

feanil (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+- Role: ecommerce
+  - Added `ECOMMERCE_BROKER_TRANSPORT` with a default value of 'ampq' to be backwards compatible with rabbit.  Set to 'redis' if you wish to use redis instead of rabbit as a queue for ecommerce.
```
jdmulloy, line 5: Can you also mention the equivalent variable for worker?

playbooks/roles/ecomworker/defaults/main.yml
```diff
@@ -33,8 +33,9 @@ ECOMMERCE_WORKER_BROKER_PASSWORD: 'celery'
 ECOMMERCE_WORKER_BROKER_HOST: '{{ ansible_default_ipv4.address }}'
 # Used as the default RabbitMQ port.
 ECOMMERCE_WORKER_BROKER_PORT: 5672
+ECOMMERCE_WORKER_BROKER_TRANSPORT: 'amqp'
```
jdmulloy, line 36: Maybe this should reference ECOMMERCE_BROKER_TRANSPORT.

i.e. `ECOMMERCE_WORKER_BROKER_TRANSPORT: '{{ ECOMMERCE_BROKER_TRANSPORT }}'`

jdmulloy (changes requested):

playbooks/roles/ecomworker/defaults/main.yml
coryleeio: We discussed this and decided not to make the change since it is in a different file

CHANGELOG.md
coryleeio: For our reference - we discussed this and decided to split it into two entries since they are different files

jdmulloy (approved):

PR #3921, opened 2017-06-15
Updated credentials play to use edx_django_service play

clintonb:
- This removes the duplication across the various IDA configurations
- Updated Dockerfile for Docker-based devstack

LEARNER-816

**Testing**
- [x] Docker
- [x] Sandbox
- [x] Loadtest

PR #4234, opened 2017-12-04
Updated credentials to use client-specific JWT issuer

clintonb:
LEARNER-3165

Caused by https://github.com/edx/configuration/pull/3921


clintonb:
LMS is using client-specific credentials: https://github.com/edx/edx-platform/blob/9efa7770f8c06bc37174e1a7da3503eb761953d6/openedx/core/djangoapps/programs/tasks/v1/tasks.py#L54.

I don't remember why we decided to do this, but we should stop. I have created [LEARNER-3441](https://openedx.atlassian.net/browse/LEARNER-3441) to fix this permanently.

feanil (approved):

PR #4246, opened 2017-12-11
Coryleeio/install splunk app via ansible

coryleeio:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/splunk-server/defaults/main.yml
+++ b/playbooks/roles/splunk-server/defaults/main.yml
@@ -74,9 +74,7 @@ SPLUNK_APP_S3_PATH: 's3://edx-splunk/apps'
 SPLUNK_APP_DOWNLOAD_PATH: "{{ splunk_home }}/etc/apps/app_downloads"
   # Source bucket for splunk apps
 
-SPLUNK_APP_TARS: 
-  - file_name: "splunk-app-for-jenkins_107.tgz"
-    md5sum: "253385a0a68caea04e81fdb4d3083dd9"
+SPLUNK_APP_TARS: []
   # A list of dicts that contains tars of splunk apps to install, and their md5sum
   # Download new apps from splunkbase(includes md5 hash):
   # https://splunkbase.splunk.com/apps/
--- a/playbooks/roles/splunk-server/defaults/main.yml
+++ b/playbooks/roles/splunk-server/defaults/main.yml
@@ -70,7 +70,12 @@ SPLUNK_DASHBOARDS: []
   # e.g.
   # - {{ role_path }}/../../../secure-repo/path/to/templates/template.xml.j2
 
-SPLUNK_APP_S3_PATH: 's3://edx-splunk/apps'
+SPLUNK_APP_S3_PATH: 'BUCKET_S3_URL_GOES_HERE' 
+  # This is the bucket where splunk apps are stored
+  # A valid bucket name is not provided above because we dont want the bucket to be claimed by someone else
+  # 
+  # example: s3://my-bucket-goes-here
+
 SPLUNK_APP_DOWNLOAD_PATH: "{{ splunk_home }}/etc/apps/app_downloads"
   # Source bucket for splunk apps
 
--- a/playbooks/roles/splunk-server/defaults/main.yml
+++ b/playbooks/roles/splunk-server/defaults/main.yml
@@ -70,6 +70,21 @@ SPLUNK_DASHBOARDS: []
   # e.g.
   # - {{ role_path }}/../../../secure-repo/path/to/templates/template.xml.j2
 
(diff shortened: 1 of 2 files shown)


playbooks/roles/splunk-server/defaults/main.yml
```diff
@@ -70,6 +70,19 @@ SPLUNK_DASHBOARDS: []
   # e.g.
   # - {{ role_path }}/../../../secure-repo/path/to/templates/template.xml.j2
 
+SPLUNK_APP_S3_PATH: 's3://edx-splunk/apps'
```
feanil: We shouldn't put our bucket names in the public repo. Put in a placeholder here and override this in our internal config.

playbooks/roles/splunk-server/tasks/main.yml
```diff
@@ -160,6 +160,44 @@
...
+  unarchive:
+    copy: no
+    src: "{{ SPLUNK_APP_DOWNLOAD_PATH }}/{{ item.file_name }}"
+    dest: "{{ splunk_home }}/etc/apps"
+  tags:
+    - install
+    - install:configuration
+  with_items:
+  - "{{ SPLUNK_APP_TARS }}"
+
```
feanil, line 211: Should we remove the tar files here to clean up after ourselves?

playbooks/roles/splunk-server/defaults/main.yml
coryleeio: Kk will do.

playbooks/roles/splunk-server/tasks/main.yml
coryleeio: I was leaving it around to prevent it from re-unarchiving if the role is run more than once, but after reading the docs I see that this doesn't do anything since it's using gtar --diff anyway, so i'll go ahead and clean those up.

On the cleanup note - 
The opt/splunk/etc/apps folder contains the apps, but it also contains a searches directory.  

Should I purge this directory with the exception of the search directory?
Otherwise our app install isn't idempotent.

coryleeio: For example if you installed jenkins_splunk_107.tgz then installed jenkins_splunk_108.tgz on the same server you would probably wind up with both installed.

playbooks/roles/splunk-server/defaults/main.yml
```diff
@@ -70,6 +70,24 @@ SPLUNK_DASHBOARDS: []
...
   # - {{ role_path }}/../../../secure-repo/path/to/templates/template.xml.j2
 
+SPLUNK_APP_S3_PATH: 'BUCKET_S3_URL_GOES_HERE' 
+  # This is the bucket where splunk apps are stored
+  # A valid bucket name is not provided above because we dont want the bucket to be claimed by someone else
+  # 
+  # example: s3://my-bucket-goes-here
+
+SPLUNK_APP_DOWNLOAD_PATH: "{{ splunk_home }}/etc/apps/app_downloads"
+  # Source bucket for splunk apps
```
feanil: Comment should talk about the download path.

playbooks/roles/splunk-server/defaults/main.yml
```diff
@@ -70,6 +70,24 @@ SPLUNK_DASHBOARDS: []
...
+
+SPLUNK_APP_TARS: []
+  # A list of dicts that contains tars of splunk apps to install, and their md5sum
+  # Download new apps from splunkbase(includes md5 hash):
+  # https://splunkbase.splunk.com/apps/
+  # Put them in SPLUNK_APP_S3_PATH
+  # Add an entry to this dict
+  # Apps will be installed if md5sum is correct, otherwise the task will fail
+  # See https://docs.splunk.com/Documentation/UnixApp/5.2.3/User/InstalltheSplunkAppforUnixandLinux#Install_the_Splunk_App_for_Unix_and_Linux_from_the_command_line
```
feanil, line 89: Add an example of what the structure of the dict looks like(which keys).

playbooks/roles/splunk-server/tasks/main.yml
```diff
@@ -160,6 +160,56 @@
...
+    owner: "{{ splunk_user }}"
+    group: "{{ splunk_user }}"
+    mode: 0755
+    state: directory
+  tags:
+    - install
+    - install:configuration
+
+- name: discover existing app tars
+  shell: "test -f {{ SPLUNK_APP_DOWNLOAD_PATH }}/{{ item.file_name }}"
```
feanil: NIT: Why not use [stat](http://docs.ansible.com/ansible/latest/stat_module.html) instead?

feanil (approved):
Couple of doc fixes which should be done before merging and a NIT on the test which you can fix if you want to.

PR #4251, opened 2017-12-14
Add EDX_PLATFORM_REVISION env var for Studio/LMS

ormsbee:
Useful for debug purposes, and better to add this via variable than to take the hit to do a dealer.git call on LMS/Studio startup (which was how it was being done before).

Configuration Pull Request
---


ormsbee:
Related to https://github.com/edx/edx-platform/pull/16901

ormsbee:
Tested on sandbox:

```
[ormsbee-sandbox] ormsbee@ormsbee i-0f00478bd1d9d1259:/edx/app/edxapp/edx-platform$ /edx/bin/edxapp-shell-lms
2017-12-14 15:12:57,295 INFO 25043 [dd.dogapi] dog_stats_api.py:66 - Initializing dog api to use statsd: localhost, 8125
Python 2.7.12 (default, Nov 20 2017, 18:23:56)
[GCC 5.4.0 20160609] on linux2
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from django.conf import settings
>>> settings.EDX_PLATFORM_REVISION
u'ormsbee/remove_git_lms'
```

ormsbee:
Ah, need CHANGELOG entry. Will have that shortly.

ormsbee:
@nasthagiri: FYI: the other half of edx/edx-platform#16901

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -982,6 +982,7 @@ generic_env_config:  &edxapp_generic_env
   CREDENTIALS_PUBLIC_SERVICE_URL: "{{ EDXAPP_CREDENTIALS_PUBLIC_SERVICE_URL }}"
   ECOMMERCE_PUBLIC_URL_ROOT: "{{ EDXAPP_ECOMMERCE_PUBLIC_URL_ROOT }}"
   ECOMMERCE_API_URL: "{{ EDXAPP_ECOMMERCE_API_URL }}"
+  EDX_PLATFORM_REVISION: "{{ edx_platform_version }}"
```
nasthagiri, line 985: Not all CAPS?

nasthagiri, line 985: Curious: Where will it get populated from now?

nasthagiri (approved):

playbooks/roles/edxapp/defaults/main.yml
ormsbee: The variable it's being read from already exists. I'm just re-using it.

nasthagiri: hmm.. ok.  just noticed that it's the only variable here that's not in all caps.

ormsbee: Ansible will compile this out to lms.env.json and cms.env.json, though I'm fuzzy on the details. I've tested in the sandbox and it seems to work correctly, though it reflects the branch name and not the commit hash -- the `edx_platform_version` is anything that git will accept as a checkout name (so hash, tag, branch, etc.). I'm hoping this is good enough for our purposes.

ormsbee:
@edx/devops: A tiny addition, if someone has time to review. Thank you.

playbooks/roles/edxapp/defaults/main.yml
fredsmith: our convention is that non-caps variables are not intended to be overriden by environment specific configs.  Otherwise they're OK to use. 👍

fredsmith: I think it should, mostly, come from GoCD materials.  This won't be particularly useful on sandboxes where the platform revision is a static version string and not a hash.

fredsmith (approved):

playbooks/roles/edxapp/defaults/main.yml
ormsbee: Right, the main goal is staging and prod. Cool, thank you very much!

PR #4254, opened 2017-12-17
Limit oauth2 by IP to do at most 20 requests/second.

feanil:
Configuration Pull Request
---

For reference before the issue this afternoon our total RPM was about 30.  This should be plenty for legit traffic but should slow down the rest.

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -15,7 +15,7 @@ upstream lms-backend {
 # Make Zone
 limit_req_zone $cookie_{{ EDXAPP_SESSION_COOKIE_NAME }} zone=cookies:10m rate={{ EDXAPP_COURSES_REQUEST_RATE }};
 
-limit_req_zone $binary_remote_addr zone=iplimit:10m rate=20r/s
+limit_req_zone $http_x_forwarded_for zone=iplimit:10m rate=20r/s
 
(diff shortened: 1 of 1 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -15,6 +15,8 @@ upstream lms-backend {
 # Make Zone
 limit_req_zone $cookie_{{ EDXAPP_SESSION_COOKIE_NAME }} zone=cookies:10m rate={{ EDXAPP_COURSES_REQUEST_RATE }};
 
+limit_req_zone $binary_remote_addr zone=iplimit:10m rate=20r/s
```
ormsbee: Sorry, I'm not very familiar with nginx conf... `binary_remote_addr` gets the real remote IP forwarded in headers and not the address of the ELB, right?

coryleeio (approved):

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
coryleeio: I *think* it is the real remote based on this post:

https://stackoverflow.com/questions/34572486/nginx-rate-limit-and-real-ip-module

ormsbee: Okay. And that's 20 requests a second? I don't really care about the specific number, I just want to understand since the PR says 10.

PR #4255, opened 2017-12-18
Add sqlite fix role to jenkins worker as dependency.

doctoryes:
Configuration Pull Request
---
This configuration change adds the sqlite upgrade fix to Jenkins workers, as we're seeing sqlite3 crashes during Django 1.9+ testing.

@macdiesel @bmedx The `tags` and `with` were added in this PR: https://github.com/edx/configuration/pull/4021 . This PR removes the tags to use the fix elsewhere - will this change cause devstack problems?

@estute @jzoldak Does this change look OK for the Jenkins workers?

### DIFF
--- a/playbooks/roles/devstack_sqlite_fix/tasks/main.yml
+++ b/playbooks/roles/devstack_sqlite_fix/tasks/main.yml
@@ -4,11 +4,6 @@
     path: "{{ SQLITE_FIX_TMP_DIR }}"
     state: directory
     mode: 0775
-  when: devstack is defined and devstack
-  tags:
-    - devstack
-    - devstack:install
-
 
 # Tasks to download and upgrade pysqlite to prevent segfaults when testing in devstack
 - name: Download and unzip sqlite autoconf update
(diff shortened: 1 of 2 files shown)


doctoryes:
@jmbowman @pwnage101 FYI

pwnage101:
shouldn't we also transfer the `devstack` and `devstack:install` tags to the role instantiation at https://github.com/edx/configuration/blob/d3b8f3aab7aedbaf3cd7bf22d09a2cbd219a5b1d/playbooks/edx-east/edxapp.yml#L23-L24 ?

jzoldak:
@doctoryes makes sense to me. we will need to create a new ami for the jenkins worker with the build-packer-ami job, then change the reference to this new ami in the jenkins master configuration.

jmbowman:
Will this actually fix the issue for tox runs?  Tox has its own virtualenv, so I would expect it to just go ahead and `pip install` a new copy of pysqlite which points back at the SQLite version bundled with Python.

doctoryes:
@jmbowman Hmm - I had assumed that the underlying image's version of sqlite would be updated in the process. But looks like `build_static` would put the libs in the Python pysqlite installation itself. As you've pointed out offline, it's strange that we're then not seeing sqlite crashes on Jenkins machines. But - I'm convinced that this PR's changes won't actually fix the sqlite crashes in the tox environment - so I'm closing this PR in favor of a tox-based solution.

Thanks for the reviews, all!

PR #4261, opened 2017-12-20
check for and replace expired key

fredsmith:
The server utils role runs first, and is the first attempt to run "apt-get update", which fails if the existing server has our expired ppa key.  This should detect the expired key and replace it with the updated one.

### DIFF
--- a/playbooks/roles/server_utils/tasks/main.yml
+++ b/playbooks/roles/server_utils/tasks/main.yml
@@ -21,6 +21,14 @@
 #
 #
 
+- name: Check for expired edx key
+  command: "apt-key list | grep {{ SERVER_UTILS_EDX_PPA_KEY_ID }}"
+  register: ppa_key_status
+
+- name: remove expired edx key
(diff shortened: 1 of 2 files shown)


playbooks/roles/server_utils/tasks/main.yml
```diff
@@ -21,6 +21,21 @@
 #
 #
 
+- name: Check for expired edx key
+  command: "apt-key list | grep {{ SERVER_UTILS_EDX_PPA_KEY_ID }}"
+  register: ppa_key_status
+
+- name: remove expired edx key
+  command: "sudo apt-key remove {{ SERVER_UTILS_EDX_PPA_KEY_ID }}"
```
feanil: You don't need to remove the old key and re-get it.  You just need to make it get the key: 

```
apt-key adv --keyserver keys.gnupg.net --recv-keys [KEY]
```

feanil (commented):
This makes sense as a way of working around the current issues but don't like this as a long term thing we leave in the config repo. It is a repetition of what we have in common and we can probably design a better solution to prevent this sort of issues in the future.

fredsmith:
@feanil Updated with a single task to just receive the key.

coryleeio (approved):

PR #4289, opened 2018-01-17
Allow packer job to use bokchoy db cache for smoke tests

michaelyoungstrom:
Configuration Pull Request
---

Summary
---

Our packer job was failing because of the recent paver work we did with respect to bokchoy database caching. As part of the ansible, we run a series of smoke tests on the new image (including a few bokchoy tests). During these tests, it tries to get our cache bucket from s3 but fails (due to a lack of credentials), which causes an Error.

This change temporarily adds a .boto file to the new image so that it can access s3, and removes it at the end (as this information should not be left on the image). I didn't want to use the environment option in the module that kicks off these tests because the keys would be leaked in the logs, and turning on "no_log" is not an adequate solution because it would give us no insight in the event of a test failure.

### DIFF
--- a/playbooks/roles/test_build_server/defaults/main.yml
+++ b/playbooks/roles/test_build_server/defaults/main.yml
@@ -17,3 +17,6 @@
 test_build_server_user: jenkins
 test_build_server_repo_path: /home/jenkins
 test_edx_platform_version: master
+
+DB_CACHE_ACCESS_KEY_ID: ''
+DB_CACHE_SECRET_ACCESS_KEY: ''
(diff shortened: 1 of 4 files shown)


feanil:
Why setup a specific key, why not just give the IAM role of the running machine the relevant permissions?  Also, do the db caches need to be secret? why not just have them be public?

michaelyoungstrom:
@feanil Great points. I'll close this and go that route. Our bucket is public to read but not write to, I'll modify how platform calls the bucket to allow us to read without credentials.

PR #4291, opened 2018-01-18
upgrade ff to 57.04 on jenkins

estute:
Don't merge. WIP


jmbowman:
We've upgraded past these versions since the PR was created, closing.

PR #4309, opened 2018-02-01
Add devpi and devpi_consumer roles and devpi Docker container.

doctoryes:
Configuration Pull Request
---
https://openedx.atlassian.net/browse/PLAT-1912

devpi is a standalone caching layer in front of PyPI. This PR adds roles to:
- install devpi on an instance
- install pip configuration on an edxapp machine to "pip install" pkgs via the proxy instead of directly via PyPI

Also, adds a Dockerfile and supporting files to build a devpi Docker container that is used in Docker devstack.

### DIFF
--- a/docker/build/devpi/Dockerfile
+++ b/docker/build/devpi/Dockerfile
@@ -1,0 +1,44 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/devpi/Dockerfile .
+#
+# This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
+# with the currently checked-out configuration repo.
+
+FROM edxops/xenial-common:latest
+MAINTAINER edxops
+
+ARG ARG_DEVPI_SERVER_VERSION=4.4.0
+ARG ARG_DEVPI_WEB_VERSION=3.2.2
+ARG ARG_DEVPI_CLIENT_VERSION=4.0.0
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+RUN apt-get update
+
+COPY docker/build/devstack/ansible_overrides.yml /devstack/ansible_overrides.yml
+
+RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook devpi.yml \
+    -c local -i '127.0.0.1,' \
+    -t "install,devstack" \
+    --extra-vars="@/devstack/ansible_overrides.yml" \
+    --extra-vars="DEVPI_SERVER_VERSION=$ARG_DEVPI_SERVER_VERSION" \
+    --extra-vars="DEVPI_WEB_VERSION=$ARG_DEVPI_WEB_VERSION" \
+    --extra-vars="DEVPI_CLIENT_VERSION=$ARG_DEVPI_CLIENT_VERSION"
+
+EXPOSE 3141
+VOLUME /data
+
+COPY docker/build/devpi/docker-entrypoint.sh /docker-entrypoint.sh
+RUN chmod +x /docker-entrypoint.sh
+
+USER root
+ENV HOME /data
+WORKDIR /data
+
+ENTRYPOINT ["/docker-entrypoint.sh"]
+CMD ["devpi"]
--- a/docker/build/devpi/docker-entrypoint.sh
+++ b/docker/build/devpi/docker-entrypoint.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+function defaults {
+    : ${DEVPI_SERVERDIR="/data/server"}
+    : ${DEVPI_CLIENTDIR="/data/client"}
+
+    echo "DEVPI_SERVERDIR is ${DEVPI_SERVERDIR}"
+    echo "DEVPI_CLIENTDIR is ${DEVPI_CLIENTDIR}"
+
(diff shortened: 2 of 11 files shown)


docker/build/devpi/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+
+ARG ARG_DEVPI_SERVER_VERSION=4.4.0
+ARG ARG_DEVPI_WEB_VERSION=3.2.2
+ARG ARG_DEVPI_CLIENT_VERSION=4.0.0
+
+ADD . /edx/app/edx_ansible/edx_ansible
+
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+RUN echo '{ "allow_root": true }' > /root/.bowerrc
```
jmbowman: Does devpi actually use Bower?  I didn't see any mention of it in the docs or repo.

playbooks/roles/devpi/defaults/main.yml
```diff
@@ -0,0 +1,35 @@
...
+# Defaults for role devpi
+
+devpi_group: devpi
+devpi_user: devpi
+devpi_app_dir: /home/devpi
+devpi_venv_dir: devpi_venv
+devpi_venv_path: "{{ devpi_app_dir }}/venvs/{{ devpi_venv_dir }}"
+DEVPI_SERVER_VERSION: 4.4.0
+DEVPI_WEB_VERSION: 3.2.2
+DEVPI_CLIENT_VERSION: 4.0.0
```
jmbowman: Do these really need to be specified both here and in the Dockerfile?  Seems like we'll forget to keep them in sync.

docker/build/devpi/docker-entrypoint.sh
```diff
@@ -0,0 +1,41 @@
...
+    export DEVPI_SERVERDIR DEVPI_CLIENTDIR
+}
+
+function initialize_devpi {
+    echo "[RUN]: Initializing devpi-server..."
+    devpi-server --restrict-modify root --start --host 127.0.0.1 --port 3141
+    devpi-server --status
+    devpi use http://localhost:3141
+    devpi login root --password=''
+    devpi user -m root password="${DEVPI_PASSWORD}"
```
jmbowman: Is this defined in a separate secure repo?  On further reflection, I think we'd be fine randomizing and logging this; the other containers shouldn't need the password to download packages by default, and the user can fetch it from the log via a make target when they need to upload packages.

jmbowman (commented):
Looks pretty good, just a few minor questions and suggestions.

playbooks/roles/devpi/defaults/main.yml
doctoryes: They don't - I just couldn't decide where to put the defaults. I'll remove them from here and leave them in the Dockerfile.

docker/build/devpi/Dockerfile
doctoryes: Nope - this is copypasta. I will remove it.

docker/build/devpi/docker-entrypoint.sh
doctoryes: Currently, it's in the docker-compose.yml file in devstack, which isn't where it should go.
Where's the logic that randomizes the firefox/chrome VNC passwords upon container startup?

doctoryes: Nm - I implemented a solution that generates a random password upon the first initialization of the devpi volumes. The password gets stashed into a file in the volume and that file persists across container start/stops, which mirrors the persistence of the root password. Let me know what you think about it.

jmbowman (approved):

doctoryes:
@edx/devops I built a sandbox with this configuration branch - it built successfully fyi.

docker/build/devpi/ansible_overrides.yml
```diff
@@ -0,0 +1 @@
+{}
```
feanil: Is the empty dict necessary?

playbooks/edx-east/edxapp.yml
```diff
@@ -11,6 +11,8 @@
      when: COMMON_ENABLE_AWS_ROLE
    - role: automated
      AUTOMATED_USERS: "{{ EDXAPP_AUTOMATED_USERS | default({}) }}"
+   - role: devpi_consumer
```
feanil: Since devstacks are now docker do you need this if you have it in the edxapp.yml in the docker directory?

feanil (commented):
I didn't review the role, just the integration points and had some questions.

docker/build/devpi/ansible_overrides.yml
doctoryes: This override file isn't used at all - but an empty dict was required to include it in the ansible-playbook call. I'll just remove it.

playbooks/edx-east/edxapp.yml
doctoryes: Right - seems like the `when` will never be true. I'll remove it from this location.

doctoryes:
@feanil I responded to your feedback. Re-review?

doctoryes:
@edx/devops Though this work isn't mission-critical, this PR is blocked on a DevOps review. And devpi can't be added to devstack until this PR merges - FYI.

coryleeio:
This is a pretty big PR and I can't look at it just yet, but i've gone
ahead and retagged all of devops on it so it will come up in our review
queue so it doesn't get lost

coryleeio (approved):

PR #4331, opened 2018-02-16
Add delay to unattended-upgrades after boot

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/security/defaults/main.yml
+++ b/playbooks/roles/security/defaults/main.yml
@@ -16,7 +16,7 @@
 #
 security_role_name: security
 # set to true to enable unattended upgrades nightly
-SECURITY_UNATTENDED_UPGRADES: true
+SECURITY_UNATTENDED_UPGRADES: false
 # set to true to upgrade all packages nightly.  false will only upgrade from security repo.
 SECURITY_UPDATE_ALL_PACKAGES: false
 # set to true to run aptitute safe-upgrade whenever ansible is run
@@ -24,6 +24,9 @@ SAFE_UPGRADE_ON_ANSIBLE: false
 # set to true to run unattended-upgrade during ansible runs.  This is expected to only install security udpates.
 SECURITY_UPGRADE_ON_ANSIBLE: false
 
+apt_timer_directories:
+  - path: "/etc/systemd/system/apt-daily.timer.d"
+  - path: "/etc/systemd/system/apt-daily-upgrade.timer.d"
 
 #
 # OS packages
@@ -37,3 +40,4 @@ security_debian_pkgs:
 security_redhat_pkgs:
   - yum-plugin-security
   - yum-cron
+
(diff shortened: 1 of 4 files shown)


playbooks/roles/security/templates/etc/systemd/system/apt-daily.timer.d/override.conf
```diff
@@ -0,0 +1,2 @@
+[Timer]
+OnBootSec=2h
```
feanil: Let's just set this to 1d so we don't have to worry about it while the AMI is being cut at all.  2h is above the usual build times but we could hit 2h if the build is really slow for some reason(slow connection for example).

feanil (commented):
@nadeemshahzad  I think you also need to deal with the apt-daily-upgrade.timer right?

playbooks/roles/security/defaults/main.yml
```diff
@@ -16,7 +16,7 @@
 #
 security_role_name: security
 # set to true to enable unattended upgrades nightly
-SECURITY_UNATTENDED_UPGRADES: false
+SECURITY_UNATTENDED_UPGRADES: true
```
jibsheet: Do we want this true for every opensource user too or just for us.
Also, are we going to take this on our AMI instances?  I assumed this was only for machines that stay up for a long time (like say jenkins)

feanil: Good point, we should still default this to false here.

I think we should take it on all our machines, with the override change, it should not interfere with any ansible runs and so it should be good to leave it running everywhere.

jibsheet: We should probably sort out if we want the upgrades running on our AMI'd machines, but either way this should default back to false and we can make appropriate overrides.  The rest of this looks fine given we switch out to a day.

nadeemshahzad:
Override configuration will add the delay of 24 hours to run ```apt-daily.service``` to update ache and ```apt-daily-upgrade.service``` to install security updates.
```
~# systemctl list-timers apt-daily.timer
NEXT                         LEFT     LAST PASSED UNIT            ACTIVATES
Tue 2018-02-20 23:58:21 UTC  23h left n/a  n/a    apt-daily.timer apt-daily.service
~# systemctl list-timers apt-daily-upgrade.timer
NEXT                         LEFT     LAST PASSED UNIT                    ACTIVATES
Wed 2018-02-21 00:14:09 UTC  24h left n/a  n/a    apt-daily-upgrade.timer apt-daily-upgrade.service
```

coryleeio (approved):

feanil:
@nadeemshahzad do you have a sandbox running with this change?

feanil:
Is this PR still relevant?  Seems like we haven't had this issue in a while?

nadeemshahzad:
@feanil I think we can keep these changes for machines which we don't build with AMI and we can override` SECURITY_UNATTENDED_UPGRADES: true ` if we need to run for all other machines.
I have build sandbox with changes with override `SECURITY_UNATTENDED_UPGRADES: true` https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/14445/ .

PR #4333, opened 2018-02-16
Add debugging to git clones.

feanil:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/git_clone/tasks/main.yml
+++ b/playbooks/roles/git_clone/tasks/main.yml
@@ -20,6 +20,14 @@
 # Rewrite this task using the ansible git-config module once we'll migrate to Ansible 2.x
 # https://docs.ansible.com/ansible/git_config_module.html#git-config
 
+- name: Debug Info about Repo
+  debug:
+    msg: "{{ item.PROTOCOL }} : {{ item.DOMAIN }} : {{ item.PATH }} : {{ item.REPO }} : {{ item.VERSION }} : {{ item.DESTINATION }}"
+  with_items: "{{ GIT_REPOS }}"
+  tags:
+    - install
+    - install:code
+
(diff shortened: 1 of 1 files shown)


playbooks/roles/git_clone/tasks/main.yml
```diff
@@ -20,6 +20,14 @@
 # Rewrite this task using the ansible git-config module once we'll migrate to Ansible 2.x
 # https://docs.ansible.com/ansible/git_config_module.html#git-config
 
+- name: Debug Info about Repo
+  debug:
+    msg: "{{ item.PROTOCOL }} : {{ item.DOMAIN }} : {{ item.PATH }} : {{ item.REPO }} : {{ item.VERSION }} : {{ item.DESTINATION }}"
+  with_items: "{{ GIT_REPOS }}"
```
jibsheet, line 26: Are you sure ansible won't print out the entire contents of GIT_REPOS?  I thought that's what happened in the other place and why we turned on no_log (especially if someone is running ansible with -v)

feanil: I'll check in the travis runs.  You might be right.  I might not be able to get at this info.

jibsheet: If it does print, we can probably transform GIT_REPOS into another dictionary (or list of strings) using set_fact and having that step set no_log.  Then debug that list/dict.  Maybe.  I wonder if there's a reason I didn't already try that.

feanil: The deletion of a value from the original list is hard I think.  But yea, setting facts would be a good way to do it.

jibsheet: zip_to_dict consumes an args of of the vars we want, and an input of the dict, it might just work

PR #4395, opened 2018-03-20
Add local email as default config for ecomm

georgebabey:
Add local smtp email as default config for ecommerce ansible role

We encountered an issue in our sandbox where ecomm emails for purchases were not being sent out (ENT-913). Tracked it down to trying to the default using EMAIL_BACKEND of SES which would require the email addresses to be registered.

This change modifies the ecommerce ansible role to override the default value in production.yml and set EMAIL_BACKEND to local smtp.  This matches what is already happening within the edxapp role: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L129

Verified in sandbox "gbabey-ecomm.sandbox.edx.org"

---

    - [x] Update the appropriate internal repo (be sure to update for all our environments)


georgebabey:
Corresponding internal change to set EMAIL_BACKEND to SES: https://github.com/edx/edx-internal/pull/318

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -186,6 +191,12 @@ ecommerce_service_config_overrides:
   DEFAULT_SITE_THEME: "{{ ECOMMERCE_DEFAULT_SITE_THEME }}"
   CSRF_COOKIE_SECURE: "{{ ECOMMERCE_CSRF_COOKIE_SECURE }}"
 
+  # default email backed set to local SMTP
```
douglashall: typo: `backed` -> `backend`

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -20,6 +20,11 @@ ECOMMERCE_PIP_EXTRA_ARGS: "-i {{ COMMON_PYPI_MIRROR_URL }}"
 ECOMMERCE_NGINX_PORT: 18130
 ECOMMERCE_SSL_NGINX_PORT: 48130
 
+ECOMMERCE_EMAIL_BACKEND: 'django.core.mail.backends.smtp.EmailBackend'
+ECOMMERCE_EMAIL_HOST: 'localhost'
+ECOMMERCE_EMAIL_PORT: 25
+ECOMMERCE_EMAIL_USE_TLS: False
```
douglashall, line 26: This assumes SMTP is installed and configured on the base machine. Do we know this is true?

douglashall (approved):

playbooks/roles/ecommerce/defaults/main.yml
georgebabey: SMTP is installed an configured in the sandboxes as that is what LMS is also using - confirmed my change on gbabey-ecomm.sandbox.edx.org. I tried to track down where it got installed but it didn't look like it was Ansible so maybe just the base AMI includes it...

douglashall:
You could also remove [this](https://github.com/edx/ecommerce/blob/master/ecommerce/settings/production.py#L21).

georgebabey:
@douglashall If I remove that than there would be no email configured out of the box for production of openedx instances that aren't using our configuration roles. Not sure if that is something we need to be concerned about, but it felt like a more breaking change than just updating our own config overrides.

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -186,6 +191,12 @@ ecommerce_service_config_overrides:
   DEFAULT_SITE_THEME: "{{ ECOMMERCE_DEFAULT_SITE_THEME }}"
   CSRF_COOKIE_SECURE: "{{ ECOMMERCE_CSRF_COOKIE_SECURE }}"
 
+  # default email backend set to local SMTP
```
fredsmith, line 194: This isn't currently overriden by any of our deployments, so there would need to be a corresponding change in edx-internal to switch it back to SES.

playbooks/roles/ecommerce/defaults/main.yml
fredsmith: it's installed by the edx local role:

https://github.com/edx/configuration/blob/master/playbooks/roles/edxlocal/defaults/main.yml#L2

I've confirmed that postfix is not running on the production ecommerce machines.

fredsmith (changes requested):
I'd like to better understand what the current behavior is, and how it will change after we do this.   We'll have to override the config in all production-ish environments to use SES.  We should also have a corresponding changelog entry to make sure people know this is changing. 

You'll need to change the default for all edx environments, including the sandbox, in edx-internal.

Django applications on sandboxes should be sending all mail via SES.  if that doesn't work, we should understand why and fix it, so please submit a ticket with some error logs.

playbooks/roles/ecommerce/defaults/main.yml
georgebabey: That was referenced here: https://github.com/edx/edx-internal/pull/318

georgebabey:
@fredsmith Changlog was updated but given this statement I'm not sure this PR is the right approach:
> "Django applications on sandboxes should be sending all mail via SES. if that doesn't work, we should understand why and fix it, so please submit a ticket with some error logs."

To be clear this is not how it is working today in sandboxes for LMS/CMS with edxapp. See the edxapp role here: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/defaults/main.yml#L129

If that is the case we can throw out this PR and look into updating edxapp instead of ecomm.

feanil:
Closing this PR as it's almost a year old.

PR #4414, opened 2018-03-22
Update build jenkins plugins

michaelyoungstrom:
update plugins


michaelyoungstrom:
This was thumbed and merged, we reverted as we were waiting a bit to deploy the changes.

PR #4407, opened 2018-03-21
Update splunk forwarder role so it can be installed on Amazon Linux OS

schenedx:
Configuration Pull Request
---

Make sure that the following steps are done before merging:

  - [ ] A DevOps team member has approved the PR.
  - [x] Are you adding any new default values that need to be overridden when this change goes live? If so:
    - [x] Add an entry to the CHANGELOG.
  - [ ] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).

PR #4426, opened 2018-03-26
Revert "Update splunk forwarder role so it can be installed on Amazon Linux OS"

jibsheet:
Reverts edx/configuration#4407

Registering the same variable for the deb and rpm download causes the user creation and splunk start to never happen on an ubuntu box, this is preventing me from building IDAs from scratch.

Please feel free to make a fix PR instead, but I'd like to move forward with my testing and I'm currently blocked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,3 @@
-- Role: splunkforwarder
-  - Updated the role so the splunkforwarder can be installed on Amazon Linux OS environment, which is a RHEL variant
-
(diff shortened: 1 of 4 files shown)


feanil (approved):

feanil:
Looks like Simon has made a fix PR, I think we can just fix forward.

schenedx:
@jibsheet Please review my fix forward PR. If that solves your problem, please close this one.

PR #4443, opened 2018-04-03
Move analytics-api to edx_django_service role

georgebabey:
Configuration Pull Request
---
Note this requires https://github.com/edx/edx-analytics-data-api/pull/184 to land first. Specifically these changes relate to configuration changes:
- new make target to just build production requirements
- gevent now a requirement as the gunicorn worker in sandbox deploys
- JWT_AUTH configuration in settings/production.py needs to update as to not overwrite base values

TODO:
- [ ] Update analytics branch to point to master before merging (or when the jenkins sandbox build lands and allows configurable version)


georgebabey:
From a configuration standpoint I think this is ready for review. @fredsmith I'll need to work with you next week to make sure the secrets repo has the correct values to support prod deploys - namely thinking the JWT_AUTH config that we're introducing to this IDA

playbooks/roles/analytics_api/defaults/main.yml
```diff
@@ -103,21 +124,27 @@ ANALYTICS_API_MEDIA_URL: '/static/reports/'
...
+ANALYTICS_API_CSRF_COOKIE_SECURE: false
+
+analytics_api_service_config_overrides:
   API_AUTH_TOKEN: '{{ ANALYTICS_API_AUTH_TOKEN }}'
   STATICFILES_DIRS: ['static']
   STATIC_ROOT: "{{ COMMON_DATA_DIR }}/{{ analytics_api_service_name }}/staticfiles"
+  LMS_BASE_URL: "{{ ANALYTICS_API_LMS_BASE_URL | default('http://127.0.0.1:8000') }}"
+
   # db config
+  ANALYTICS_DATABASE: 'reports'
```
feanil: Where does this config var get used?

georgebabey: This gets used in multiple places in the edx-analytics-data-api code as a reference to the record store in the DB config. Used in [DB Routing](https://github.com/edx/edx-analytics-data-api/blob/5c635010f15650670a46af1ca52c14ad706d8f8a/analyticsdataserver/router.py#L11), [Healthcheck endpoint](https://github.com/edx/edx-analytics-data-api/blob/662acec284687a37761c26216ae0eeb0defbd2be/analyticsdataserver/views.py#L96) and [within the views](https://github.com/edx/edx-analytics-data-api/blob/5c635010f15650670a46af1ca52c14ad706d8f8a/analytics_data_api/v0/views/courses.py#L663)


I will update this to pull from `ANALYTICS_API_REPORTS_DB_NAME` name which is already set to 'reports'  No reason to have to configure that in multiple places

feanil (commented):
Generally looks good, do you have a sandbox with this code running?  Now that the job is seeded, you should be able to run with your version of configuration and test that the analytics API stuff works as expected.

georgebabey:
@feanil This is deployed to sandboxes with both [full_edx_installation_from_scratch](https://analytics-api-georgebabey.sandbox.edx.org/) and [full_edx_installation](https://analytics-api-georgebabey-not-scratch.sandbox.edx.org/) AMIs and both seem functional

georgebabey:
I had to revert the analytics-data-api changes as those changes require configuration set here.

@feanil or @fredsmith can I get a review of this?

feanil (approved):

PR #4451, opened 2018-04-06
remove datadog agents

syed-awais-ali:
Configuration Pull Request
---


feanil (commented):
This should probably not go into the common role since we don't want to affect all users.  Why not just create a new small playbook with the things we need.

playbooks/roles/common_vars/defaults/main.yml
```diff
@@ -225,3 +225,5 @@ COMMON_JWT_SECRET_KEY: 'SET-ME-PLEASE'
 CREATE_SERVICE_WORKER_USERS: True
 
 COMMON_ENABLE_AWS_ROLE: true
+
+COMMON_DISABLE_DATADOG: false
```
feanil: I don't think you need this variable.  It's not used anywhere.

playbooks/roles/datadog-uninstall/meta/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+dependencies:
+  - common
```
feanil: I don't think you need this either.  It doesn't need common to uninstall datadog.

playbooks/roles/datadog-uninstall/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+common_datadog_agent_version: '1:5.10.1-1'
```
feanil: These variables should not be prefixed with common,  they are now in a different role.  prefix with `datadog_uninstall`

coryleeio (approved):

playbooks/roles/datadog-uninstall/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+datadog_uninstall_agent_version: '1:5.10.1-1'
```
coryleeio, line 1: If we specify the version and are wrong we might miss some servers, maybe consider removing it?

coryleeio: Approving because I dont this is worth blocking on

PR #4456, opened 2018-04-09
Limit django access from MIT vpn cidrs

nadeemshahzad:
Configuration Pull Request
---


playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
```diff
@@ -82,6 +82,16 @@ server {
     try_files $uri @proxy_to_app;
   }
 
+{% if NGINX_MIT_VPN_CIDRS %}
```
jibsheet: This is an internal name for this feature, can we name the config option something more generic like DJANGO_ADMIN_RESTRICTION or DJANGO_ADMIN_CIDR_ACCESS or something else?

jibsheet (approved):
Please squash and also leave a note in the changelog.  Good to merge after that.

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
```diff
@@ -82,6 +82,16 @@ server {
     try_files $uri @proxy_to_app;
   }
 
+{% if NGINX_DJANGO_ADMIN_ACCESS_CIDRS %}
```
jibsheet: Does this need to be defaulted to [] for the edx_django_service IDAs?
You have it in nginx, but I wasn't sure if that was rolled in somehow or not (I assumed not)

jibsheet:
@nadeemshahzad I realize that since we're going to try and roll this out IDA by IDA, we can't just have the list of blocks, we need something to turn it on or off per IDA as well.  I think we can keep the global list of CIDR blocks, but we'll need some ENABLE variable to be passed in by each role as we go, and then we can remove it later.

Without that restriction, we'll accidentally turn this on for some of the IDAs before finalizing the ELB setup.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -300,6 +300,16 @@ location ~ ^{{ EDXAPP_MEDIA_URL }}/(?P<file>.*) {
     expires {{ EDXAPP_PROFILE_IMAGE_MAX_AGE }}s;
 }
 
+{% if NGINX_DJANGO_ADMIN_ACCESS_CIDRS %}
```
jibsheet: We'll likely want the same second variable for the nginx role as the edx_django_service role (prefixed differently).  Otherwise when we deploy ecom, it'll turn on for LMS/CMS immediately also.

jibsheet (approved):
I think this is safe to merge since it'll default to off.

PR #4468, opened 2018-04-10
install and configure splunk plugin on jenkins

estute:
Install the Splunk plugin @ version 1.6.4 and provide basic configuration for the groovy scripts defined here: https://github.com/edx/jenkins-configuration/pull/78

### DIFF
--- a/playbooks/roles/jenkins_build/defaults/main.yml
+++ b/playbooks/roles/jenkins_build/defaults/main.yml
@@ -18,6 +18,7 @@ build_jenkins_configuration_scripts:
   - 4configureJobConfigHistory.groovy
   - 4configureMailerPlugin.groovy
   - 4configureMaskPasswords.groovy
+  - 4configureSplunkPlugin.groovy
   - 5createLoggers.groovy
 
 # plugins
@@ -190,6 +191,9 @@ build_jenkins_plugins_list:
   - name: 'shiningpanda'
     version: '0.21'
     group: 'org.jenkins-ci.plugins'
+  - name: 'splunk-devops'
+    version: '1.6.4'
+    group: 'com.splunk.splunkins'
   - name: 'ssh-agent'
     version: '1.14'
     group: 'org.jenkins-ci.plugins'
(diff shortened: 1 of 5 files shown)


playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -276,3 +280,27 @@ build_jenkins_log_list:
 # job config history
 build_jenkins_history_max_days: '15'
 build_jenkins_history_exclude_pattern: 'queue|nodeMonitors|UpdateCenter|global-build-stats|GhprbTrigger'
+
+# splunk
+build_jenkins_splunk_ignored_jobs: '(^((?!edx-(platform|e2e)).)*$)|.*private.*'
```
jmbowman, line 285: That looks fun to validate.  Do we have a way to easily check which existing jobs that regex will and won't ignore?

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -276,3 +280,27 @@ build_jenkins_log_list:
...
+    - metadata.data_source: 'Build Report'
+      metadata.config_item: 'Source Type'
+      metadata.value: 'build_report'
+    - metadata.data_source: 'Build Event'
+      metadata.config_item: 'Source Type'
+      metadata.value: 'build_event'
+    - metadata.data_source: 'Console Log'
+      metadata.config_item: 'Source Type'
+      metadata.value: 'console_log'
+    - metadata.data_source: 'Queue_Information'
```
jmbowman: Why does this one have an underscore in `data_source` when the rest don't?

playbooks/roles/jenkins_common/defaults/main.yml
```diff
@@ -180,3 +180,23 @@ JENKINS_MAILER_SMTP_AUTH_PASSWORD: ''
 # mask passwords
 JENKINS_MASK_PASSWORDS_CLASSES: []
 JENKINS_MASK_PASSWORDS_PAIRS: []
+
+# splunk
+jenkins_common_splunk_enabled: true
+JENKINS_SPLUNK_HOSTNAME: ''
```
jmbowman: Do we have control over the capitalization here, or is something else forcing us to use lower case for some variables and upper case for others?

playbooks/roles/jenkins_common/templates/config/splunk_config.yml.j2
```diff
@@ -0,0 +1,21 @@
+---
+SPLUNK_APP_ENABLED: {{ jenkins_common_splunk_enabled }}
+SPLUNK_HOSTNAME: '{{ JENKINS_SPLUNK_HOSTNAME }}'
+SPLUNK_HOST_PORT: {{ JENKINS_SPLUNK_PORT }}
+USE_SSL: {{ jenkins_common_splunk_use_ssl }}
+SPLUNK_TOKEN: {{ JENKINS_SPLUNK_TOKEN }}
```
jmbowman: Does the value for this need to go in quotes, since it's a string?  Same question for `IGNORED_JOBS_PATTERN` below.

playbooks/roles/jenkins_build/defaults/main.yml
estute: Ah, typo, thanks

playbooks/roles/jenkins_common/defaults/main.yml
estute: This was following the conventions found here: https://openedx.atlassian.net/wiki/spaces/OpenOPS/pages/26837527/Ansible+Code+Conventions

playbooks/roles/jenkins_build/defaults/main.yml
estute: To be honest, I use the groovy shell. We could easily write a job/script that queries Jenkins for all jobs and tries to match them to a supplied regex.

jmbowman (approved):

playbooks/roles/jenkins_common/defaults/main.yml
```diff
@@ -180,3 +180,23 @@ JENKINS_MAILER_SMTP_AUTH_PASSWORD: ''
...
+JENKINS_SPLUNK_PORT: 8088
+jenkins_common_splunk_use_ssl: true
+JENKINS_SPLUNK_TOKEN: ''
+jenkins_common_splunk_raw_events_enabled: true
+jenkins_common_splunk_batch_size: 4096
+JENKINS_SPLUNK_APP_URL: ''
```
michaelyoungstrom: Could you sort out the fields that are in all caps from the rest, similar to the rest of the file? I'm not certain if thats technically the standard but it will keep it consistent with our config and make it easier to read, especially since it should be obvious which fields are intended to be overwritten (all caps).

michaelyoungstrom (approved):

PR #4498, opened 2018-04-20
Fix requirements for jenkins worker

michaelyoungstrom:
We need to run pip from the root platform directory due to relative imports.

Configuration Pull Request
---


jmbowman (approved):
Huh, it looks like it hadn't previously been installing the requirements from local.txt (which is why this was working before).  I guess it had just been waiting for paver to do that, and that was being run from the correct directory.

PR #4513, opened 2018-04-24
Add RETIREMENT_STATES to edxapp to make retirement workflow configurable

bmedx:
Reviewer note - this variable isn't intended to be overridden within edx, but open source partners might need to. I believe that the placement and lack of variable substitution for this is correct for that use case, but am far from certain of it and happy to shuffle things around if this is not the right way to go about it.

Usage is here: https://github.com/edx/edx-platform/pull/18014

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -731,6 +731,13 @@ EDXAPP_RETIRED_USER_SALTS:
   - "OVERRIDE ME WITH A RANDOM VALUE"
   - "ROTATE SALTS BY APPENDING NEW VALUES"
 EDXAPP_RETIREMENT_SERVICE_WORKER_USERNAME: "OVERRIDE THIS WITH A VALID LMS USERNAME"
+# These get loaded into database models per environment via management command
+# These are the required states, environmental overrides are in edx-internal.
+EDXAPP_RETIREMENT_STATES:
+  - "PENDING"
+  - "ERRORED"
+  - "ABORTED"
+  - "COMPLETE"
 
 # Comprehensive Theming
 # Deprecated, maintained for backward compatibility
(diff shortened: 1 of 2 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1248,6 +1248,29 @@ generic_env_config:  &edxapp_generic_env
   RETIRED_EMAIL_DOMAIN: "{{ EDXAPP_RETIRED_EMAIL_DOMAIN }}"
   RETIRED_USER_SALTS: "{{ EDXAPP_RETIRED_USER_SALTS }}"
   RETIREMENT_SERVICE_WORKER_USERNAME: "{{ EDXAPP_RETIREMENT_SERVICE_WORKER_USERNAME }}"
+  RETIREMENT_STATES:
```
feanil: Please pull this out into a toplevel config so it can be updated without updating the entire genaric env config.  See the example for the other retirement vars.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1248,6 +1248,29 @@ generic_env_config:  &edxapp_generic_env
...
+    - "ENROLLMENTS_COMPLETE"
+    - "RETIRING_NOTES"
+    - "NOTES_COMPLETE"
+    - "NOTIFYING_PARTNERS"
+    - "PARTNERS_NOTIFIED"
+    - "RETIRING_LMS"
+    - "LMS_COMPLETE"
+    - "ERRORED"
+    - "ABORTED"
+    - "COMPLETE"
```
jibsheet: Is this the correct default? (what happens for openedx installs that don't have ecom for example)

bmedx:
@feanil @jibsheet I think this is ready for re-review. Override PRs now also exist in internal, edge-internal, and sandbox-internal.

jibsheet (approved):
My only question is that some times I waffle on platform's default not equalling the ansible default, but that's a separate fix if we decide to do it anyway (I tend to have a default in platform with a comment about other options, and then we set those options in the internal repos)

feanil (approved):

doctoryes (approved):
FWIW

PR #4517, opened 2018-04-27
Fix analytics_api role when analytics db name differs from config name

georgebabey:
In https://github.com/edx/configuration/pull/4443/files I mistakenly changed the ANALYTICS_DATABASE to the ANALYTICS_API_REPORTS_DB_NAME. In fact this config refers to the database config identifier, the key in the DATABASES dictionary.

This worked fine in sandboxes where both the config and the db are named 'reports', however when those configs differ, such as on staging, the health check would fail as it tries to connect to the reporting database. This PR fixes that change by setting this value back to 'reports'

Configuration Pull Request
---


playbooks/roles/analytics_api/defaults/main.yml
```diff
@@ -144,7 +144,7 @@ analytics_api_service_config_overrides:
   LMS_BASE_URL: "{{ ANALYTICS_API_LMS_BASE_URL | default('http://127.0.0.1:8000') }}"
 
   # db config
-  ANALYTICS_DATABASE: '{{ ANALYTICS_API_REPORTS_DB_NAME }}'
+  ANALYTICS_DATABASE: 'reports'
```
fredsmith: Can you set this to a new overridable variable with the default set to something that works in sandboxes and devstacks, and then override it in stage and prod?

fredsmith (changes requested):

playbooks/roles/analytics_api/defaults/main.yml
georgebabey: Sure - that makes sense if someone would want to overwrite the entire DATABASE config.

To be clear though this is set to 'reports' across all of our current environments. The problem was ANALYTICS_API_REPORTS_DB_NAME differs which was causing issues in stage

georgebabey:
@fredsmith PR updated. Sandbox to verify is being built.

fredsmith (approved):

PR #4531, opened 2018-05-02
Coryleeio/social data logging fix

coryleeio:
Configuration Pull Request
---


jibsheet (approved):
While I'm fine with this change (although I am puzzled why the map check fails) I'd love to get a second thumb from @feanil

coryleeio:
The map check fails bc the if= is only valid in a location block, I did verify that it works correctly in a location block and does not work correctly outside of it. (it does nothing, and does not complain about being in the wrong context)

The context for the access_logs directive reads:
Context: | http, server, location, if in location, limit_except

http://nginx.org/en/docs/http/ngx_http_log_module.html

feanil (approved):

PR #4552, opened 2018-05-11
WIP journal post migrate

tuchfarber:
Configuration Pull Request
---


tuchfarber:
Where we stand:
This creates a site, org, and journal in the Journals app now. 

Things that are missing:
- Org creation in Journals doesn't create an org in discovery because the API for that model is read only. Don't know the best way around this without changing the view model type and that might require some learner/educator sign-off.
- Needs to add video settings to journal. In the provision script this is done via SQL, but the better option is probably to extend the publish_journals command to accept a list of course IDs. Then we'll need to run the gather_videos command after

PR #4558, opened 2018-05-15
Allow adding custom maintenance pages with static contents.

akovari:
Configuration Pull Request
---

This change allows adding custom static contents to nginx and override the default server error template. The goal is to allow having more complex error pages with custom images, stylesheets etc.

An example of this could be something like:

    NGINX_STATIC_FILES:
      - "error-static/"
    NGINX_SERVER_TEMPLATE: "error-template.j2"


openedx-webhooks:
Thanks for the pull request, @akovari! I've created [OSPR-2409](https://openedx.atlassian.net/browse/OSPR-2409) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

mduboseedx:
jenkins ok to test

mduboseedx:
@edx/devops Could you take a look at this?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -173,6 +173,12 @@ error_page {{ k }} {{ v }};
     {{ NGINX_EDXAPP_LMS_APP_EXTRA }}
   }
 
+  {% for item in NGINX_SERVER_MAINTAINANCE_LOCATIONS %}
+    location {{ item.location }} {
```
mtyaka: I don't think we need this, since a location that serves static files from the static dir [already exists](https://github.com/open-craft/configuration/blob/f7f09d3f577a25515d7d29700864ced5def11d1a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/static-files.j2#L1-L5). You can see it in action by visiting https://courses.edx.org/server/rate-limit.html for example.

But we do need some way to toggle the maintenance page on and off.

Adding something like this (not tested) near the top of the `server { ... }` block might work:

```
if (-f {{ nginx_server_static_dir }}/maintenance.html) {
   rewrite ^(.*)$ /server/maintenance.html last;
}
```

That way you can have a `maintenance.html.disabled` template for example, and just copy it to `maintenance.html` when you want to enable it.

playbooks/roles/nginx/tasks/main.yml
```diff
@@ -251,6 +251,18 @@
...
     - install:configuration
 
+- name: Copy error pages and static contents
+  copy:
+    src: "{{ item.path }}"
+    dest: "{{ nginx_server_static_dir }}"
+    owner: "{{ common_web_user }}"
+    group: "{{ common_web_user }}"
+    mode: "0640"
+  with_items: "{{ NGINX_SERVER_MAINTAINANCE_LOCATIONS }}"
```
mtyaka: Can we rename the variables to `NGINX_SERVER_STATIC_FILES` instead, since it's more generic than just maintenance pages and can be used for any static file?

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -184,3 +184,5 @@ NGINX_EDXAPP_CMS_APP_EXTRA: ""
 NGINX_EDXAPP_LMS_APP_EXTRA: ""
 
 NGINX_DJANGO_ADMIN_ACCESS_CIDRS: []
+
+NGINX_SERVER_MAINTAINANCE_LOCATIONS: []
```
mtyaka: As mentioned below, I suggest to replace this with `NGINX_SERVER_STATIC_FILES` and make it just a list of files (or folders) to be copied over to nginx static dir.

mtyaka (changes requested):

playbooks/roles/nginx/tasks/main.yml
```diff
@@ -241,7 +241,7 @@
 
 - name: Create NGINX server templates
   template:
-    src: "edx/var/nginx/server-static/server-template.j2"
+    src: {{ NGINX_SERVER_TEMPLATE }}
```
mtyaka: The default value for the new `NGINX_SERVER_TEMPLATE` variable needs to be defined in `defaults/main.yml`.

mtyaka (changes requested):
Found a small problem with missing default variable value, but otherwise this is looking good.

playbooks/roles/nginx/tasks/main.yml
akovari: Good catch! Thanks

mtyaka: `{{ NGINX_SERVER_TEMPLATE }}` needs to be wrapped in quotes (`"{{ NGINX_SERVER_TEMPLATE }}"`), otherwise  it's not valid YAML - that's why tests are currently failing.

mtyaka (changes requested):

playbooks/roles/nginx/tasks/main.yml
akovari: Hmm, I fixed it, thanks

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -184,3 +184,6 @@ NGINX_EDXAPP_CMS_APP_EXTRA: ""
 NGINX_EDXAPP_LMS_APP_EXTRA: ""
 
 NGINX_DJANGO_ADMIN_ACCESS_CIDRS: []
+
+NGINX_SERVER_TEMPLATE: "edx/var/nginx/server-static/server-template.j2"
++NGINX_SERVER_STATIC_FILES: []
```
mtyaka: @akovari Sorry for not noticing this earlier, but there's an extra `+` character in front of the variable name.

mtyaka (changes requested):

playbooks/roles/nginx/defaults/main.yml
akovari: Ooops! Thanks @mtyaka ! fixed

mtyaka (approved):

mtyaka:
@akovari I tested this and it works well :+1:

Can you please add an entry to the [CHANGELOG](https://github.com/edx/configuration/blob/master/CHANGELOG.md) to list the new variables and explain what they do?

- [x] I tested this: I used this code to deploy a custom maintenance page to an Open EdX installation.
- [x] I read through the code
- [ ] ~~I checked for accessibility issues~~ n/a
- [ ] Includes documentation - not yet: we should add an entry to the changelog

akovari:
@mtyaka , thanks for the review. I've updated the CHANGELOG

mtyaka:
Rebased from f79f8daa0880857e767c7a33e469a17be88cbfde

mduboseedx:
@mtyaka Is this ready for a review?

mtyaka:
@mduboseedx Yes, this is ready for review.

mduboseedx:
@edx/devops @jdmulloy Could you give this a review when you have a chance?

CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+Role: nginx                                                             
+  - Added `NGINX_SERVER_TEMPLATE` to make the error file template location configurable
+  - Added `NGINX_SERVER_STATIC_FILES` to allow copying static contents to the server static folder. Can be used to deploy static contents for the error pages for example
+  - Added `NGINX_PROXY_INTERCEPT_ERRORS` to be able to use custom static error pages for error responses from the LMS.
+  - When there is file named `maintenance.html` in the `nginx_server_static_dir`, it will be shown if the backend is unavailable
```
jdmulloy, line 5: I'm still working on the full review, but one thing I've noticed and confirmed with testing is that if this file exists it will be served regardless of the backend's availability, which I believe is the desired behavior. This changelog line should be updated to reflect this.

jdmulloy (commented):
This looks reasonable but I still need to do some more testing. There are a few other things we need for this to work for edx.org, but this is a good start. The other goals we have are below, if you would like to address them feel free, but if not we can build on top of your work to implement them.

Thanks

- API endpoints should return a 503 with an empty body
- Static assets should still be served normally to avoid accidental caching by the CDN.

mtyaka:
Thanks for reviewing this @jdmulloy! I'll schedule some time to address your comments in the next two weeks.

jdmulloy:
Hi @mtyaka,
I found out that we already have similar functionality in the nginx config that serves a maintenance html file from S3 if nginx can't reach gunicorn. To activate the maintenance we stop gunicorn, which also prevents writes to the database. We would prefer to adapt the existing code instead of adding a second way of doing a maintenance page. Perhaps you could adapt the existing code to optionally use a local file instead of S3 for those who aren't using S3.

There are some issues with the existing code that I've addressed in PR #4673. Primarily the /heartbeat endpoint returns a 502 which causes the ELB to stop sending the instances traffic, so users never see the maintenance page.

Please let me know if there are issues with the current code for your use case.

Thanks

jdmulloy:
I've also created a playbook to toggle the maintenance page in PR #4676.

jdmulloy:
@mtyaka I am closing this PR for now since I believe PRs #4673 and #4676 cover the same use case. Please let me know if you have a use case that these do not solve, or submit a PR to adapt the existing code to you use case. For instance make it so that a maintenance page can be served from somewhere other than S3.

Thanks

mtyaka:
Thank you @jdmulloy!

I opened #4714 and #4715 to address some use cases that haven't been covered by merged code yet. If you could take a look that would be great!

PR #4561, opened 2018-05-15
update nr display name to instanceid on aws

fredsmith:
This code will set the display name variable in newrelic infrastructure and the hostname variable in nrsysmond to the same string - the hostname plus the AWS instanceid - on instance boot. This is to work around an issue with newrelic infrastructure that fails to differentiate hosts with different instance ids but the same hostname.  By giving each host a completely unique hostname in NR, it will provide more consistent data.


fredsmith:
@jibsheet @feanil this is ready for review.

playbooks/roles/edxapp/templates/edx/app/edxapp/lms.sh.j2
```diff
@@ -0,0 +1,33 @@
...
+export NEW_RELIC_CONFIG_FILE="{{ edxapp_app_dir }}/newrelic.ini"
+if command -v ec2metadata >/dev/null 2>&1; then
+  INSTANCEID=$(ec2metadata --instance-id);
+  HOSTNAME=$(hostname)
+  export NEW_RELIC_PROCESS_HOST_DISPLAY_NAME="$HOSTNAME-$INSTANCEID"
+fi
+export NEW_RELIC_LICENSE_KEY="{{ NEWRELIC_LICENSE_KEY }}"
+{% endif -%}
+
+export PORT="{{ edxapp_lms_gunicorn_port }}"
```
feanil: Should this go into an 'lms_env' file that we can source here?   If that file doesn't already exist, I think it's fine to do that as a follow-up PR.

playbooks/roles/edxapp/templates/edx/app/edxapp/cms.sh.j2
```diff
@@ -0,0 +1,33 @@
...
+
+{% set edxapp_venv_bin = edxapp_venv_dir + "/bin" %}
+
+{% if COMMON_ENABLE_NEWRELIC_APP %}
+{% set executable = edxapp_venv_bin + '/newrelic-admin run-program ' + edxapp_venv_bin + '/gunicorn' %}
+{% else %}
+{% set executable = edxapp_venv_bin + '/gunicorn' %}
+{% endif %}
+
+{% if COMMON_ENABLE_NEWRELIC_APP %}
```
feanil: Why have a second of the same conditional as above?  Why not just put it into the IF clause of the above conditional?

playbooks/roles/edxapp/templates/edx/app/edxapp/lms.sh.j2
fredsmith: there's an edxapp_env file, but not an lms_env file.

feanil (approved):

PR #4576, opened 2018-05-22
update ff container to 3.11

estute:
This will sync the FF versions between docker devstack and Jenkins.  Build Jenkins currently runs Firefox 59.0.2 whereas the latest Firefox container in the edxops group in Dockerhub is 59.0.1. We are seeing some possible discrepancies in behavior of bokchoy tests while trying to debug them, so we should sync these to be able have reproducible tests.

Also, the firefox container was not being built as part of the CI for this repo

### DIFF
--- a/docker/build/chrome/Dockerfile
+++ b/docker/build/chrome/Dockerfile
@@ -1,4 +1,4 @@
-FROM selenium/standalone-chrome-debug:3.4.0-einsteinium
+FROM selenium/standalone-chrome-debug:3.11.0-californium
 MAINTAINER edxops
 
 USER root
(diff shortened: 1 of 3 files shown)


jmbowman (approved):
I suspect the FF version in the container is even older than that; the 3.4.0-einsteinium image comes with FF 54.0.  Firefox has gotten much faster over the past few versions, would be nice if this gets us a noticeable speed boost.

jmbowman:
@estute Are you ok with merging this now?  We're down to just one failing bok-choy test in devstack, which is less than we were at with the previous Firefox image.

PR #4586, opened 2018-05-30
Setting up a jenkins worker docker image

michaelyoungstrom:
This Dockerfile is to be used for building jenkins-worker containers. The plan is to have an EC2 instance on Jenkins spin up a number of these containers, and shard out tests via pytest-xdist.

This file was designed with just unittests in mind to start. Since these assume mongo is running, the Dockerfile also runs the mongo play. And since the EC2 instance will need to ssh into the containers, it also sets up sshd. If you think it makes more sense to work towards keeping a multicontainer approach (might be tricky with fargate), let me know... I figured a "fat" container at least made sense to get the ball rolling.

I plan on keeping the same subnet/ security group for these containers as we use for our EC2 workers, so I decided to create the user ubuntu (in order to keep our ssh config simple... as the user/key will be the same this way).

The dockerfile will depend on the argument `JENKINS_WORKER_KEY_URL`, which points to a url with our desired ssh public key.

### DIFF
--- a/playbooks/roles/edxapp/tasks/service_variant_config.yml
+++ b/playbooks/roles/edxapp/tasks/service_variant_config.yml
@@ -66,6 +66,40 @@
     - install
     - install:configuration
 
+# write the supervisor script for mongod and sshd
+# this should only be run when provisioning a jenkins-worker container
+- name: Set ignore_jenkins_worker_command when no tags present
+  set_fact: ignore_jenkins_worker_command=true
+
+- name: writing mongod and sshd supervisor scripts
+  template:
+    src: "edx/app/supervisor/conf.d.available/{{ item }}.j2"
+    dest: "{{ supervisor_available_dir }}/{{ item }}"
+    owner: "{{ supervisor_user }}"
+    group: "{{ supervisor_user }}"
+    mode: 0644
+  become_user: "{{ supervisor_user }}"
+  with_items:
+    - mongod.conf
+    - sshd.conf
+  when: 'ignore_jenkins_worker_command is not defined'
+  tags:
+    - jenkins-worker
+
+- name: "enable mongod and sshd supervisor script"
+  file:
+    src: "{{ supervisor_available_dir }}/{{ item }}.conf"
+    dest: "{{ supervisor_cfg_dir }}/{{ item }}.conf"
+    state: link
+    force: yes
+  become_user: "{{ supervisor_user }}"
+  with_items:
+    - mongod
+    - sshd
+  when: 'ignore_jenkins_worker_command is not defined'
+  tags:
+    - jenkins-worker
+
 # Enable the supervisor jobs
 - name: "enable {{ item }} supervisor script"
   file:
--- a/docker/build/jenkins_worker/Dockerfile
+++ b/docker/build/jenkins_worker/Dockerfile
@@ -1,0 +1,58 @@
+# To build this Dockerfile:
+#
+# From the root of configuration:
+#
+# docker build -f docker/build/jenkins_worker/Dockerfile .
+#
+# This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
+# with the currently checked-out configuration repo.
+
+# Run the edxapp play with custom ansible overrides
+ARG BASE_IMAGE_TAG=latest
+FROM edxops/xenial-common:${BASE_IMAGE_TAG}
+MAINTAINER edxops
+USER root
+
+ADD . /edx/app/edx_ansible/edx_ansible
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+
+COPY docker/build/jenkins_worker/ansible_overrides.yml /jenkins_worker/ansible_overrides.yml
+COPY docker/build/edxapp/devstack.yml /
+COPY docker/build/devstack/ansible_overrides.yml /devstack/ansible_overrides.yml
+
+ARG OPENEDX_RELEASE=master
+ENV OPENEDX_RELEASE=${OPENEDX_RELEASE}
+RUN sudo /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook edxapp.yml \
+    -c local -i '127.0.0.1,' \
+    -t 'install,assets,devstack,jenkins-worker' \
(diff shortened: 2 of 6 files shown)


docker/build/jenkinsworker/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+# Run the mongo play
+COPY docker/build/mongo/ansible_overrides.yml /mongo/ansible_overrides.yml
+
+RUN mkdir -p /data/db
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook mongo.yml \
+    -i '127.0.0.1,' -c local \
+    -t 'install' \
+    -e@/mongo/ansible_overrides.yml
```
jmbowman: Maybe use `--extra-vars=` instead of `-e` for consistency with the run above?

docker/build/jenkinsworker/Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+
+RUN groupadd ubuntu
+RUN useradd -ms /bin/bash ubuntu -g ubuntu -d /home/ubuntu
+
+ARG JENKINS_WORKER_KEY_URL
+RUN curl ${JENKINS_WORKER_KEY_URL} --create-dirs -o /home/ubuntu/.ssh/authorized_keys
+RUN chown -R ubuntu /home/ubuntu
+
+RUN apt-get clean && \
+    rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
```
jmbowman: For this cleanup to actually reduce the image size, it needs to happen in the same Dockerfile instruction as `apt-get update` and `apt-get install`; otherwise a layer is created containing this data, and then another layer is created that doesn't have it, and both are added together as part of the image.  See https://github.com/edx/configuration/blob/master/docker/build/chrome/Dockerfile#L7 for an example.

docker/build/jenkinsworker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+RUN mkdir /var/run/sshd
+
+RUN groupadd ubuntu
+RUN useradd -ms /bin/bash ubuntu -g ubuntu -d /home/ubuntu
+
+ARG JENKINS_WORKER_KEY_URL
+RUN curl ${JENKINS_WORKER_KEY_URL} --create-dirs -o /home/ubuntu/.ssh/authorized_keys
+
+RUN chown -R ubuntu /home/ubuntu
+RUN chown ubuntu /edx/app/edxapp/edx-platform
```
estute: I would recommend chaining all of these RUN commands. It is considered a good practice and can possibly lead to better performance of the container.

docker/build/jenkinsworker/Dockerfile
```diff
@@ -0,0 +1,58 @@
+# To build this Dockerfile:
```
estute: Nit: I know the directory itself isn't consistent, but could you rename with '_'?

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -57,6 +57,40 @@
     - install
     - install:configuration
 
+# write the supervisor script for mongod and sshd
+# this should only be run when provisioning a jenkins-worker container
+- name: Set ignore_jenkins_worker_command when no tags present
+  set_fact: ignore_jenkins_worker_command=true
```
estute, line 72: Why do you ignore the task unless specified, as opposed to only running it when a flag is defined? Not a big deal, but it was just a little confusing to me.

docker/build/jenkinsworker/Dockerfile
```diff
@@ -0,0 +1,58 @@
...
+RUN mkdir -p /data/db
+WORKDIR /edx/app/edx_ansible/edx_ansible/docker/plays
+RUN /edx/app/edx_ansible/venvs/edx_ansible/bin/ansible-playbook mongo.yml \
+    -i '127.0.0.1,' -c local \
+    -t 'install' \
+    -e@/mongo/ansible_overrides.yml
+
+# Add sshd to enable jenkins master to ssh into containers
+RUN apt-get update
+RUN apt-get install -y openssh-server
```
jmbowman: It looks like you just deleted the cleanup for data added in these `apt-get` runs; if you did the update, install, and cleanup all in the same `RUN` command, it would keep the layer size small.  Example: https://github.com/edx/configuration/blob/master/docker/build/chrome/Dockerfile#L7

playbooks/roles/edxapp/tasks/service_variant_config.yml
michaelyoungstrom: I wasn't sure if there's a way to do that. Do you have an example? By default, if you add a tag to a task, that task will be run both when the tag is specified, as well as when no tags are specified.

estute: ah ok, so we will be passing  'ignore_jenkins_worker_command=false' as an extra arg, then?

michaelyoungstrom: Nope, we won't ever need to pass in an argument for `ignore_jenkins_worker_command` 

Basically, this dummy variable is to protect against the case I mentioned: when someone runs the play with no tags.

When that happens, the var is set to true, and therefore the steps writing and enabling mongod and sshd files doesn't get run (thanks to `when: 'ignore_jenkins_worker_command is not defined'`)

When someone runs the play with the `jenkins-worker` tag, which we will be doing, then the block above doesn't get run, so the variable isn't defined and the mongod/sshd blocks get run.

This might be slightly confusing, but I haven't found a better solution when searching, and I've implemented something similar in the past, here: https://github.com/edx/configuration/blob/f1c79df88b333609db815f05542cae53293903f1/playbooks/roles/jenkins_common/tasks/main.yml#L226-L234

estute: thanks for the clarification

estute (approved):

jmbowman:
@michaelyoungstrom Did you see the Travis complaint about the missing entry in `util/parsefiles_config.yml`?  Other than that, looks good to go.

PR #4596, opened 2018-06-05
Kill edx-east

coryleeio:


playbooks/edx-east/add-ubuntu-key.yml
```diff
@@ -0,0 +1 @@
+../add-ubuntu-key.yml
```
feanil: Why symlink each file individually instead of just symlinking the folder?

coryleeio: Good point  - fixed.

feanil (approved):

PR #4597, opened 2018-06-05
Stop having edxapp: for edxapp:lms edxapp:cms

jibsheet:
Configuration Pull Request
---

As we move to deploying cms and lms separately, we don't need a group.
Having the group specified in each lms.conf cms.conf supervisor config
doesn't work on sandboxes, so I just cleaned up everything that uses
them on sandboxes.

### DIFF
--- a/playbooks/edx-east/ora2.yml
+++ b/playbooks/edx-east/ora2.yml
@@ -26,9 +26,6 @@
       environment:
         PATH: "{{ edxapp_deploy_path }}"
       become_user: "{{ edxapp_user }}"
-      notify:
-        - "restart edxapp"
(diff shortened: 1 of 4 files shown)


playbooks/edx-east/ora2.yml
```diff
@@ -46,8 +46,11 @@
         - deploy
 
   handlers:
-    - name: restart edxapp
-      shell: "{{ supervisorctl_path }} restart edxapp:"
+    - name: restart lms
```
feanil: You need to update the 'notify' statements above to call the new handlers.

feanil (approved):

PR #4598, opened 2018-06-06
Allow nginx to be templated but not enabled

jibsheet:
If you set nginx_skip_enable_sites it will not symlink the items in
nginx_sites.  This means you have to make that change later.  EdX does it
with our pre_supervisor script, the same code that checks on migrations
being complete and also symlinks in supervisor processes to run.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -327,6 +327,16 @@ EDXAPP_COURSES_REQUEST_BURST_RATE: 10
 EDXAPP_COURSES_USER_AGENT_BURST_RATE: 5
 EDXAPP_RATE_LIMITED_USER_AGENTS: []
 
+# Consumed in the edxapp.yml playbook
+# to pass to nginx as nginx_skip_enable_sites
+EDXAPP_NGINX_SKIP_ENABLE_SITES: False
+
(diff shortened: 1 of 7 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -327,6 +327,11 @@ EDXAPP_COURSES_REQUEST_BURST_RATE: 10
 EDXAPP_COURSES_USER_AGENT_BURST_RATE: 5
 EDXAPP_RATE_LIMITED_USER_AGENTS: []
 
+# TODO - default to false so that we always link config and, push this into our repo.
+# Consumed in the edxapp.yml playbook
+# to pass to nginx as nginx_skip_enable_sites
+EDXAPP_NGINX_SKIP_ENABLE_SITES: True
```
feanil: We may want to set this to false by default and override it in internal.  If we set it like this, we'll break the current behavior that others might be relying on for edxapp.  Also, this is probably worth a changelog entry in either case.

jibsheet: Yep - see the TODO part of the comment, this was pure laziness to avoid dealing with edx-internal while I was testing it

feanil (approved):

PR #4606, opened 2018-06-07
Rebase olive on master

coryleeio:


feanil:
@coryleeio this is all a single commit which makes it hard to know which commits are overrides specific to olivex.

PR #4671, opened 2018-07-06
WIP: OPS-3164 NGINX maintenance page

jdmulloy:
This won't be merged, I just need feedback on my changes.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/api_maintenance.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/api_maintenance.j2
@@ -1,0 +1,4 @@
+    # maintenance page
+    if (-f {{ nginx_server_static_dir }}/maintenance.html) {
+      return 503;
+    }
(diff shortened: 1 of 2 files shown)


feanil:
Why do you need to move it into each location block?

feanil:
Since this is a jinja template, should we just do a jinja include of a single shared block?

jdmulloy:
I moved the maintenance page if to the location that proxies to edxapp, so that static assets will still be served normally. This is to prevent the CDN from caching the maintenance page.

In all the locations that have "api" in them I added an if to serve a 503 if the maintenance page exists. Right now it's the standard nginx error page, I can do some work to make it serve empty content, or perhaps an empty json object which would be ```{}```

jdmulloy:
@feanil I can do a shared block.

feanil:
Empty page with the right response code is fine for API endpoints I think.  Can we serve the maintenance page from s3 like we have in the past?

feanil:
@jdmulloy how does this reconcile with https://github.com/edx/configuration/blob/master/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/s3_maintenance.j2

jdmulloy:
@feanil we might be able to serve a page from s3, there is some code in here for it, but I don't yet know how it works. It looks like it requires rerunning ansible to reconfigure nginx, whereas the OSPR solution just requires puting a file in place, which could be done by ansible. I'll try figuring out the s3 solution.

PR #4673, opened 2018-07-06
Maintenance via 502/s3 OPS-3164

jdmulloy:
This is a different way to do a maintenance page than #4671 which leverages our existing s3 proxy solution. Possibly more performant since we only do a file check on the heartbeat and not on all requests.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/cms.j2
@@ -30,6 +30,8 @@ server {
 error_page {{ k }} {{ v }};
   {% endfor %}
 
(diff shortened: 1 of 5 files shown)


jdmulloy:
With these changes stopping supervisor will still cause a maintenance page to be shown via S3. We have to put a file in place to fake the health check so we stay in the ELB. Normally the health check will fail if the LMS is down. I also put in an ugly hack to serve an empty json file on api endpoints.

jibsheet:
I like this.
Need to do a better read of it, but I think it solves a bunch of the original concerns about how to handle the api endpoints, etc.

jibsheet (approved):

jibsheet (approved):
Looks good, assuming tests don't find anything

PR #4675, opened 2018-07-10
do not copy security config for local jenkins dev

estute:
Configuration Pull Request
---

--------
In the past, we exlcuded the security configuration from Jenkins when doing local development because it also sets up github oauth. We have since split the oauth config out of security config, the security config is being included in local development, but this leads to a problem: since we do not have oauth AND do not specify a generic 'admin' user in our permissions, you can't actually do anything on the box.

This makes sure that neither github oauth or security are included in a local dev box

To be honest, I would like to have an override for this so that a local dev environment has a generic admin/permission set so that you can more accurately test, but this unblocks me in my current work.

### DIFF
--- a/playbooks/roles/jenkins_common/tasks/main.yml
+++ b/playbooks/roles/jenkins_common/tasks/main.yml
@@ -173,12 +173,12 @@
     - install:base
     - install:jenkins-configuration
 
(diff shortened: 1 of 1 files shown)


estute:
CC @azafty468

jmbowman (approved):

PR #4681, opened 2018-07-11
xserver sandbox apparmor

fredsmith:
Make sure that xserver is running code under apparmor.


feanil (approved):

nedbat:
@fredsmith @feanil This should work on Hawthorn, right?  The file renaming threw me off a bit, but the key fix here is to use `xserver_venv_sandbox_dir` in the paths in the apparmor config, right?

nedbat:
("work" meaning, once we cherry-pick it to Hawthorn, the fix will work there too.)

feanil:
@nedbat yes, this should get ported to hawthorn and yes, the change in the path for the app-armor profile is the key change.

nedbat:
This has been cherry-picked onto Hawthorn.

PR #4685, opened 2018-07-13
A new playbook to create RDSes

jibsheet:
We'll need to add more to add a replica (command=replicate) or modify (command=modify) but it took long enough to get the infrastructure right and this RDS doesn't have replicas.

Those two playbooks will likely be copy pastas.  But we can use them to increads disk size, etc.

Configuration Pull Request
---


playbooks/create_rds.yml
```diff
@@ -0,0 +1,57 @@
...
+      username: "{{ database_connection.login_user }}" 
+      password: "{{ database_connection.login_password }}"
+      engine_version: "{{ database_engine_version }}"
+      maint_window: "{{ maintenance_window }}"
+      multi_zone: "{{ multi_zone|default('no') }}"
+      vpc_security_groups: "{{ vpc_security_groups }}"
+      subnet: "{{ subnet_group }}"
+      parameter_group: "{{ parameter_group }}"
+      backup_window: "{{ backup_window }}"
+      backup_retention: "{{ backup_retention|default(0) }}"
```
feanil: I'd rather default to having some backups instead of no-backups but honestly just make this a required field, you should always have to think about this.

playbooks/create_rds.yml
```diff
@@ -0,0 +1,57 @@
...
+      instance_name: "{{ database_name }}"
+      db_engine: "{{ database_engine|default('MySQL') }}"
+      size: "{{ database_size }}"
+      instance_type: "{{ instance_type }}"
+      region: "{{ aws_region }}"
+      username: "{{ database_connection.login_user }}" 
+      password: "{{ database_connection.login_password }}"
+      engine_version: "{{ database_engine_version }}"
+      maint_window: "{{ maintenance_window }}"
+      multi_zone: "{{ multi_zone|default('no') }}"
```
feanil: I think don't set a default for this, we should fail if there is nothing explicitly set for this.

feanil (approved):

PR #4686, opened 2018-07-13
Add EDX_REST_API_CLIENT_NAME env variable.

ormsbee:
This is sent as part of the User-Agent by edx-rest-api-client as of version 1.7.2.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_django_service/templates/edx/app/app/app.sh.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/app/app.sh.j2
@@ -20,5 +20,7 @@ fi
 export NEW_RELIC_LICENSE_KEY="{{ NEWRELIC_LICENSE_KEY }}"
 {% endif -%}
 
+export EDX_REST_API_CLIENT_NAME="{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-{{ edx_django_service_name }}"
+
 source {{ edx_django_service_home }}/{{ edx_django_service_name }}_env
 {{ executable }} -c {{ edx_django_service_home }}/{{ edx_django_service_name }}_gunicorn.py {{ edx_django_service_gunicorn_extra }} {{ edx_django_service_wsgi_name }}.wsgi:application
--- a/playbooks/roles/edxapp/templates/edx/app/edxapp/cms.sh.j2
+++ b/playbooks/roles/edxapp/templates/edx/app/edxapp/cms.sh.j2
@@ -27,6 +27,7 @@ export DJANGO_SETTINGS_MODULE="{{ EDXAPP_CMS_ENV }}"
 export SERVICE_VARIANT="cms"
 export PATH="{{ edxapp_deploy_path }}"
 export BOTO_CONFIG="{{ edxapp_app_dir }}/.boto"
+export EDX_REST_API_CLIENT_NAME="{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-studio"
 
 source {{ edxapp_app_dir }}/edxapp_env
 {{ executable }} -c {{ edxapp_app_dir }}/cms_gunicorn.py {{ EDXAPP_CMS_GUNICORN_EXTRA }} cms.wsgi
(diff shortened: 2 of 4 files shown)


ormsbee:
Sorry on the "assign" bit. Wrong box.

jibsheet:
No worries on the assign, but I'm headed out in a few minutes so I leave it to Feanil

jibsheet:
There are a handful of IDAs that don't use edx_django_service yet.  XQueue, notes, and I believe insights and analytics_api.  Do we need consistency?
You should be able to grep service playbooks/roles/{list,of,idas}/meta/main.yml to figure out what uses it and what doesn't.

Also, I can't remember where the conversation went about settings vs ENV because I was sort of snowed in with on-call when that happened, ENV won?

playbooks/roles/edx_django_service/templates/edx/app/app/app.sh.j2
```diff
@@ -20,5 +20,7 @@ fi
 export NEW_RELIC_LICENSE_KEY="{{ NEWRELIC_LICENSE_KEY }}"
 {% endif -%}
 
+export EDX_REST_API_CLIENT_NAME="{{ edx_django_service_name }}"
```
robrap: @jibsheet: Some questions:
1. Where would be a good place to add this to get it into all environments?
2. Would it make sense to just use the E-D-P?

jibsheet: @ormsbee Robert and I were going to talk about this Friday but I was too swamped before he went away.
I'm not sure what he meant by question 1, probably trying to make sure your additions of studio/lms/workers covered my initial note?  I haven't grepped meta.yml to see if any other roles needed the var
question 2 - my inclination is to use the EDP when we can (which should have variables we can use) but I don't know if I know enough about this work to know if using just the app name is better/sufficient

robrap: @jibsheet: You can ignore question 1.  I was asking because you said edx_django_service wasn't used everywhere, and I didn't know that @ormsbee had already made an update to resolve your issue.
You can also ignore my question about EDP, unless this is important to you.  I didn't know if EDP might already be available as an environment variable.

So, if you are good with this then we can squash and merge.  Just let me know.  Thanks.

playbooks/roles/edxapp/templates/edx/app/edxapp/worker.sh.j2
```diff
@@ -15,5 +15,7 @@ fi
 {% set executable = edxapp_venv_bin + '/python' %}
 {% endif %}
 
+export EDX_REST_API_CLIENT_NAME="edxapp-worker"
```
feanil: Rather than setting this here, you may want to set it here: https://github.com/edx/configuration/blob/master/playbooks/roles/edxapp/templates/edx/app/supervisor/conf.d.available/workers.conf.j2#L4

This way you can have a different client name for each type of backend worker we run making it easier to pin-point which queue the requests are coming from.

robrap: Thanks @feanil.  Do you mean adding something like the following to that list?
```
EDX_REST_API_CLIENT_NAME=edx.{{ w.service_variant }}.core.{{ w.queue }}
```
Thanks.

feanil: Yes, adding that to the supervisor profile will give us more info than labeling all the workers as `edxapp-worker`

playbooks/roles/edx_django_service/templates/edx/app/app/app.sh.j2
jibsheet: It's probably better to do
"{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-{{ edx_django_service_name}}"
just so when we look at all the metrics, it's clearer.  Or if you accidentally cross a prod API and a stage client for example, it'll be more obvious.

robrap: @jibsheet: Are these available for lms and studio as well?  For example:
```
export EDX_REST_API_CLIENT_NAME="{{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-lms"
```

jibsheet: Yep, generally speaking COMMON_ prefixed vars should be available in any role (as long as the role correctly meta's in the common_vars role) which studio/lms do.

robrap:
@jibsheet @feanil: I updated the PR.  Thanks.

jibsheet (approved):
Will you squash - tests will take a while, so no real reason to wait for this round to finish.

After that, let's ship it.  Tag me if you need approval on applying the DRF change to things.

PR #4703, opened 2018-07-23
Bump Build Jenkins plugins for security updates

michaelyoungstrom:
Most plugins are small bumps for security patches, but the biggest upgrade is the ssh-credentials plugin, which deprecates the ability to use files on jenkins master for ssh secrets. This PR removes support for such secret type.

Note: Our list of plugins at the moment is essentially the plugins we explicitly want installed, mixed with dependencies we need to pin for some reason... such as a dependency that otherwise would be installed at a version prior to an important release (like a security patch).

It would be cool if we could come up with some sort of pip-tools-esque system instead, utilizing the plugins gradle script we have in the jenkins-configuration repo.... just a thought.


estute (approved):
Concerning your note, I will keep that in mind during my current task, as it is fairly related

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -67,7 +70,7 @@ build_jenkins_plugins_list:
     version: '1.39'
     group: 'org.jenkins-ci.plugins'
   - name: 'credentials'
-    version: '2.1.16'
+    version: '2.1.17'
```
jmbowman: 2.1.18 was released on the 20th, although it doesn't seem to contain any security fixes.

jmbowman (commented):
I can't find anywhere that the *_SSH_LIST variables are given meaningful content; am I missing it, or is this for future use?

playbooks/roles/jenkins_build/defaults/main.yml
michaelyoungstrom: yeah I had bumped this purely because it was needed for the ssh credentials plugin, but i might as well bump to latest after seeing the changelog.

michaelyoungstrom:
@jmbowman The variable is empty by default but can be overwritten with the secrets outside of this repository. It then gets loaded into the template, which gets consumed by the jenkins-configuration repo scripts.

jmbowman (approved):

PR #4706, opened 2018-07-24
Support for Asymmetric JWT Keys

nasthagiri:
Configuration Pull Request
---

This PR goes hand-in-hand with https://github.com/edx/edx-platform/pull/18629 and https://github.com/edx/edx-drf-extensions/pull/39.

### DIFF
--- a/playbooks/roles/edx_django_service/defaults/main.yml
+++ b/playbooks/roles/edx_django_service/defaults/main.yml
@@ -154,6 +154,7 @@ edx_django_service_jwt_auth:
     - AUDIENCE: '{{ edx_django_service_jwt_audience }}'
       ISSUER: '{{ edx_django_service_jwt_issuer }}'
       SECRET_KEY: '{{ edx_django_service_jwt_secret_key }}'
+  JWT_PUBLIC_SIGNING_JWK_SET: '{{ COMMON_JWT_PUBLIC_SIGNING_JWK_SET }}'
 
 edx_django_service_extra_apps: []
 
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -359,10 +359,16 @@ EDXAPP_UNIVERSITY_EMAIL: 'university@example.com'
 EDXAPP_PRESS_EMAIL: 'press@example.com'
 EDXAPP_LMS_ROOT_URL: "{{ EDXAPP_LMS_BASE_SCHEME | default('https') }}://{{ EDXAPP_LMS_BASE }}"
 EDXAPP_LMS_INTERNAL_ROOT_URL: "{{ EDXAPP_LMS_ROOT_URL }}"
+
 EDXAPP_LMS_ISSUER: "{{ COMMON_JWT_ISSUER }}"
 EDXAPP_JWT_EXPIRATION: 30  # Number of seconds until expiration
 EDXAPP_JWT_AUDIENCE: "{{ COMMON_JWT_AUDIENCE }}"
 EDXAPP_JWT_SECRET_KEY: "{{ COMMON_JWT_SECRET_KEY }}"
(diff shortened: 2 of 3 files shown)


feanil:
`COMMON_JWT_SIGNING_JWK_SET` is the public key right?  Would it be more clear to name it `COMMON_PUBLIC_JWT_SIGNING_JWK_SET` and add a comment expanding the acronyms in the name?

nasthagiri:
Yes.  `COMMON_JWT_SIGNING_JWK_SET` is the **set** of public keys that are actively available for verifying JWT tokens in the wild.

Sounds good.  I'll add `PUBLIC` in its name and add a comment.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -359,10 +359,15 @@ EDXAPP_UNIVERSITY_EMAIL: 'university@example.com'
...
 EDXAPP_LMS_ROOT_URL: "{{ EDXAPP_LMS_BASE_SCHEME | default('https') }}://{{ EDXAPP_LMS_BASE }}"
 EDXAPP_LMS_INTERNAL_ROOT_URL: "{{ EDXAPP_LMS_ROOT_URL }}"
+
 EDXAPP_LMS_ISSUER: "{{ COMMON_JWT_ISSUER }}"
 EDXAPP_JWT_EXPIRATION: 30  # Number of seconds until expiration
 EDXAPP_JWT_AUDIENCE: "{{ COMMON_JWT_AUDIENCE }}"
 EDXAPP_JWT_SECRET_KEY: "{{ COMMON_JWT_SECRET_KEY }}"
+EDXAPP_JWT_PUBLIC_SIGNING_JWK_SET: "{{ COMMON_JWT_PUBLIC_SIGNING_JWK_SET }}"
+
+EDXAPP_JWT_SIGNING_ALGORITHM: 'SET-ME-PLEASE'
```
jibsheet: We have links to docs for JWK_SET above, should we do the same for the signing things?  may just be a link back into the platform

playbooks/roles/ecommerce/defaults/main.yml
```diff
@@ -51,6 +51,7 @@ ECOMMERCE_JWT_LEEWAY: 1
 # while migrating to the new key.
 ECOMMERCE_JWT_SECRET_KEYS:
   - '{{ COMMON_JWT_SECRET_KEY }}'
+ECOMMERCE_JWT_PUBLIC_SIGNING_JWK_SET: '{{ COMMON_JWT_PUBLIC_SIGNING_JWK_SET }}'
```
jibsheet: ecommerce uses edx_django_service so shouldn't need another setting of this

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -146,6 +146,7 @@ edx_django_service_oidc_issuer: '{{ COMMON_OIDC_ISSUER }}'
 edx_django_service_jwt_audience: '{{ COMMON_JWT_AUDIENCE }}'
 edx_django_service_jwt_issuer: '{{ COMMON_JWT_ISSUER }}'
 edx_django_service_jwt_secret_key: '{{ COMMON_JWT_SECRET_KEY }}'
+edx_django_service_JWT_PUBLIC_SIGNING_JWK_SET: '{{ COMMON_JWT_PUBLIC_SIGNING_JWK_SET }}'
```
jibsheet: this declares a variable inside the edx_django_service role but doesn't use it.
I suspect you want to define it directly inside edx_django_service_jwt_auth a few lines below, similar to the SOCIAL_AUTH ones.
The other jwt vars exist because of JWT_ISSUERS which gets stuffed into JWT_AUTH.

playbooks/roles/edxapp/defaults/main.yml
jibsheet: We should probably also set this to the same default used in common.py, otherwise it needs to go in internal and openedx installs have to pick it also.

jibsheet (approved):

PR #4707, opened 2018-07-25
OPS-3333 Fix default character sets and collations in mysql playbook

coryleeio:


coryleeio:
This doesn't change the existing tables or databases, it changes new ones if a lower level default is not specified. It does mean the information schema etc may be created under utf8 for new dbs which should be fine.

After:

	mysql> SELECT @@character_set_server,@@collation_server;
	+------------------------+--------------------+
	| @@character_set_server | @@collation_server |
	+------------------------+--------------------+
	| utf8                   | utf8_general_ci    |
	+------------------------+--------------------+
	1 row in set (0.01 sec)

playbooks/roles/mysql/handlers/main.yml
```diff
@@ -0,0 +1,5 @@
+---
```
jibsheet: I believe we've been trying to remove handlers (needless complexity in many cases, and can fail oddly on more complex playbooks)
Can we just do this as the next task in the role?

playbooks/roles/mysql/tasks/main.yml
```diff
@@ -21,4 +21,4 @@
   when: remove_experimental_mysql
 
 - include: mysql.yml
-  when: (mysql_56_installed.rc == 1) or (remove_experimental_mysql)
\ No newline at end of file
+  when: (mysql_56_installed.rc == 1) or (remove_experimental_mysql)
```
jibsheet, line 24: Will you file a ticket, I think we can scrub out the remove_experimental_mysql logic, it was part of an old vagrant devstack failure and should be totally irrelevant with Hawthown

playbooks/roles/mysql/handlers/main.yml
coryleeio: We can - my thought was that you might want to run the mysql playbook without restarting mysql, if we want to bake a restart in we can

coryleeio: thoughts?

playbooks/roles/mysql/tasks/main.yml
coryleeio: Sure - https://openedx.atlassian.net/browse/OPS-3336

playbooks/roles/mysql/handlers/main.yml
jibsheet: we only really use this role for sandboxes built from scratch to my knowledge (and maybe one dangling vagrant analytics thing) so I don't see the value in not restarting

coryleeio: Okie dokie

coryleeio: @jibsheet  Fixed this

playbooks/roles/mysql/tasks/mysql.yml
```diff
@@ -40,10 +40,16 @@
     state: present
   with_items: "{{ mysql_debian_pkgs }}"
 
-- name: Start mysql
+- name: Set default character sets and collations
+  template: >
```
jibsheet: sorry, I missed this earlier.  This is the weird old ansible format, we shouldn't use > and switch to key: arg rather than key=arg (which means they each go on their own line)

jibsheet (approved):
Fix that template syntax and it should be good to go

PR #4722, opened 2018-08-02
fix ACE defaults (hawthorn)

mulby:

Master PR here: https://github.com/edx/configuration/pull/4721


nasthagiri (approved):

nedbat (approved):

jibsheet (approved):
BTW, Based on these being the defaults, we may be able to remove our overrides in edx-internal since they're identical.
This will help us later as we try to squash all these down and have better defaults.

mulby:
@jibsheet agreed re: edx-internal, with the exception of EDXAPP_ACE_CHANNEL_DEFAULT_EMAIL which will have a different value on edx.org

PR #4727, opened 2018-08-07
ENT-1142. Enable CORS for analytics-data-api calls through edx_django_service

georgebabey:
This PR adds an optional `allow_cors_headers` flag for the edx_django_service role. When set
the CORS headers are added for any non-basic-auth endpoint (such as /api)  I added this to edx_django_service in anticipation other IDAs that separate frontends need to talk to will need this same configuration.

This also updates the configuration for analytics-data-api to use this new fields, in addition to
setting the CORS_ORIGIN whitelist (which for edx properties is set in edx-internal).  The edx-portal will be making XHRs to analytics-data-api so this is necessary for that project.

Verified on sandbox georgebabey-data-api-cors.sandbox.edx.org

Configuration Pull Request
---


playbooks/roles/analytics_api/defaults/main.yml
```diff
@@ -142,6 +142,7 @@ ANALYTICS_API_REPORT_DOWNLOAD_BACKEND:
 
 ANALYTICS_API_CSRF_COOKIE_SECURE: false
 
+ANALYTICS_API_ALLOW_CORS_HEADERS: true
```
georgebabey, line 145: We could keep this disabled here and only enable in edx-internal, but I didn't see the harm in keeping it here since the origin whitelist will be empty

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/cors-add-header.j2
```diff
@@ -1 +1,2 @@
-add_header 'Access-Control-Allow-Origin' $cors_origin;
+    add_header 'Access-Control-Allow-Origin' $cors_origin;
+    add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
```
georgebabey, line 2: This nginx include that adds CORS headers already existed, but was only used for static assets previously so didn't have the Allow-Methods (because calls to /static would be GETs it wasn't needed)

Because I'm using the same include, this will open up /static to POST & OPTIONS, but again, we're whitelisting origins so I don't think this is a concern, just calling out the side-effect.

playbooks/roles/analytics_api/defaults/main.yml
douglashall: Do we need a flag at all? Wouldn't we always want to set the CORS headers using the defined whitelist?

georgebabey: If i didn't put this flag in, credentials for example which has a whitelist, but is only whitelisting for static assets would then open up the `/api` routes.  I thought opting-in would make sense for something like this.

brittneyexline (approved):

fredsmith (approved):
LGTM.  Have you tested this on a sandbox?  I'm not sure about the path behavior on jinja2 includes, which is what you're doing in your nginx template. I don't know where the paths are relative from.

georgebabey:
@fredsmith Thanks - verified on a sandbox (georgebabey-data-api-cors.sandbox.edx.org) and the resulting nginx is as expected

PR #4733, opened 2018-08-08
OPS-3198 removing custom redirects

coryleeio:


coryleeio:
For cleanup after removing edxonline

feanil (approved):

PR #4749, opened 2018-08-27
DEVOPS-7790 Add management command users to ecom role

coryleeio:


jibsheet (commented):
Looking at the other roles, won't this fail without edx-internal setting ECOMMERCE_AUTOMATED_USERS?

I think you need to declare it to be an empty dictionary like discovery or journals do

coryleeio:
@jibsheet I think you are right.  Updated.

jibsheet (approved):

BenjiLee (approved):

PR #4756, opened 2018-09-05
configuration for journals frontend app

bill-filler:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/journals/defaults/main.yml
+++ b/playbooks/roles/journals/defaults/main.yml
@@ -57,6 +57,7 @@ JOURNALS_MEMCACHE: [ 'memcache' ]
 JOURNALS_VERSION: "master"
 JOURNALS_DJANGO_SETTINGS_MODULE: "journals.settings.production"
 JOURNALS_URL_ROOT: 'http://localhost:{{ JOURNALS_NGINX_PORT }}'
+JOURNALS_FRONTEND_URL: 'http://localhost:1991'
 JOURNALS_LOGOUT_URL: '{{ JOURNALS_URL_ROOT }}/logout/'
 
 JOURNALS_SECRET_KEY: 'Your secret key here'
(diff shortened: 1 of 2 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -327,6 +327,7 @@ ECOMMERCE_SOCIAL_AUTH_REDIRECT_IS_HTTPS: true
 ecommerce_create_demo_data: true
 
 JOURNALS_URL_ROOT: "https://journals-{{ EDXAPP_LMS_BASE }}"
+JOURNALS_FRONTEND_URL: "https://journalsapp-{{ EDXAPP_LMS_BASE }}"
```
tuchfarber, line 330: I realize we can change it, but we might want to have a conversation with architecture about how we want to name these. I don't have any preferences on naming scheme currently, so I'm fine merging as is.

playbooks/roles/journals/defaults/main.yml
```diff
@@ -57,6 +57,7 @@ JOURNALS_MEMCACHE: [ 'memcache' ]
 JOURNALS_VERSION: "master"
 JOURNALS_DJANGO_SETTINGS_MODULE: "journals.settings.production"
 JOURNALS_URL_ROOT: 'http://localhost:{{ JOURNALS_NGINX_PORT }}'
+JOURNALS_FRONTEND_URL: 'http://localhost:1991'
```
tuchfarber, line 60: Fine with merging as is. Don't know if we should pick a new port for the all the frontends as the cookie cutter defaults to 1991 and we won't be able to run multiple at once. That's more of a long term plan though (aka how we're going to run devstacks with 8+ separate frontends). Something we'll need to think about moving forward.

tuchfarber (approved):

util/jenkins/ansible-provision.sh
bill-filler: these actually get ovewritten by the whitelabel task which gets the proper names out of themes sandbox config

playbooks/roles/journals/defaults/main.yml
bill-filler: same comment as above

fredsmith (approved):

PR #4763, opened 2018-09-07
Moved Nginx log format name override

blarghmatey:
When the `NGINX_LOG_FORMAT_NAME` var is set to something different than `p_combined` it breaks the Nginx configuration as there is no long a `p_combined` format defined for the main nginx config. This also prevents users from being able to specify their own format and using it throughout the various Nginx configs.

Removed the `NGINX_LOG_FORMAT_NAME` setting from the `log_format` definition to allow for the default log format to function properly. Users can define an alternate format with a different name and pass that via the `NGINX_LOG_FORMAT_NAME` to be used in the CMS/LMS etc.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
+++ b/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
@@ -33,7 +33,7 @@ http {
         # Logging Settings
         ##
 
-        log_format {{ NGINX_LOG_FORMAT_NAME }} '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local]  '
+        log_format p_combined '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local]  '
                             '"$request" $status $body_bytes_sent $request_time '
                             '"$http_referer" "$http_user_agent"';
 
--- a/playbooks/roles/nginx/defaults/main.yml
+++ b/playbooks/roles/nginx/defaults/main.yml
@@ -42,6 +42,10 @@ NGINX_SSL_PROTOCOLS: "TLSv1 TLSv1.1 TLSv1.2"
 NGINX_DH_PARAMS_PATH: "/etc/ssl/private/dhparams.pem"
 NGINX_DH_KEYSIZE: 2048
 
+# This can be one of 'p_combined' or 'ssl_combined' by default. If you
+# wish to specify your own format then define it in a configuration file
+# located under `nginx_conf_dir` and then use the format name specified
+# in your configuration file.
 NGINX_LOG_FORMAT_NAME: 'p_combined'
 # When set to False, nginx will pass X-Forwarded-For, X-Forwarded-Port,
 # and X-Forwarded-Proto headers through to the backend unmodified.
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/xqueue.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/xqueue.j2
@@ -9,6 +9,9 @@ server {
 
   {% include "common-settings.j2" %}
 
+  access_log {{ nginx_log_dir }}/access.log {{ NGINX_LOG_FORMAT_NAME }};
+  error_log {{ nginx_log_dir }}/error.log error;
+
   location / {
     {% if XQUEUE_ENABLE_BASIC_AUTH|bool %}
       {% include "basic-auth.j2" %}


openedx-webhooks:
Thanks for the pull request, @blarghmatey! I've created [OSPR-2602](https://openedx.atlassian.net/browse/OSPR-2602) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

mduboseedx:
@edx/devops Could you give this a look?

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -33,7 +33,7 @@ http {
         # Logging Settings
         ##
 
-        log_format {{ NGINX_LOG_FORMAT_NAME }} '$http_x_forwarded_for - $remote_addr - $remote_user [$time_local]  '
+        log_format p_combined '$http_x_forwarded_for - $remote_addr - $remote_user [$time_local]  '
```
coryleeio: Why are you no longer using NGINX_LOG_FORMAT_NAME, 
It seems to currently be defaulted to p_combined?

blarghmatey: @coryleeio that's because if you override the value of NGINX_LOG_FORMAT_NAAME in this location then two lines below that where `p_combined` is referenced as the default log format, there is no such format defined. The use of the Ansible var in this location has no use as if it is changed from p_combined then the Nginx configuration ends up broken.

natabene:
@coryleeio  Did you have a chance to look at response from @blarghmatey ?

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
coryleeio: I would remove the NGINX_LOG_FORMAT_NAME entirely or change two lines down to:

`access_log {{ nginx_log_dir }}/access.log {{ NGINX_LOG_FORMAT_NAME }};`

You are using the variable in one place but not the other, I would be happy to see the variable removed entirely, or see it used two lines down. I'm happy to merge either of those.

coryleeio (changes requested):

natabene:
@blarghmatey  Please let us know once you have addressed Cory's comments.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
blarghmatey: @coryleeio I'm not sure I understand what you are asking... The Ansible variable in question is not present in this file at all. It is still present in the other template files because it is possible to specify an additional log format in a file located under the `conf.d` directory which can then be referenced by the applications, which is what I am doing for my deployment.

coryleeio: @blarghmatey So in playbooks/roles/nginx/defaults/main.yml


NGINX_LOG_FORMAT_NAME is defaulted to 'p_combined' by this line:

    NGINX_LOG_FORMAT_NAME: 'p_combined' 


The break you are fixing is when you override NGINX_LOG_FORMAT_NAME to somtehing else, and so it breaks on line 44 which currently reads:

    access_log {{ nginx_log_dir }}/access.log p_combined;

because p_combined doesnt exist if you override NGINX_LOG_FORMAT_NAME.

However, if line 44 of playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2 read: 

    access_log {{ nginx_log_dir }}/access.log {{ NGINX_LOG_FORMAT_NAME }};

this would fix the bug and still work for anyone else who is using the variable, unless I misunderstand something.

blarghmatey: @coryleeio it would allow Nginx to run, yes, but it would also force you to still use the same log format as if you never changed that variable. The reason for having that variable is to allow users to specify a new log format and use it in the LMS/CMS/etc.

coryleeio: @blarghmatey Ooh okay right, that makes sense.

One last thing then:
I think we want to leave 
$http_x_forwarded_proto
in place, was it causing you issues?  You removed it from the default format.

coryleeio (changes requested):

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
coryleeio: after holidays bump =]

If you can put $http_x_forwarded_proto back I will merge

coryleeio (approved):
Great thank you for the fix - I'll merge it now

openedx-webhooks:
@blarghmatey 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/configuration&entry.867055334=https://github.com/edx/configuration/pull/4763&entry.1484655318=https://github.com/blarghmatey&entry.752974735=2018-09-07+17:06&entry.1917517419=2019-01-22+18:39&entry.2133058324=Yes) so we can improve your experience in the future.

coryleeio:
Thank you for the contribution @blarghmatey =]

nedbat:
I've cherry-picked this onto Ironwood.

PR #4764, opened 2018-09-07
Adds in prospectus playbook

jlajoie:
Configuration Pull Request
---


playbooks/roles/prospectus/tasks/main.yml
```diff
@@ -0,0 +1,11 @@
+# Copy over the target from the previous build to where it needs to be
+
+- name: Create data folder
+  file:
+    path: "{{ PROSPECTUS_DATA_DIR }}"
+    state: directory
+
+- name: Copy prospectus static files
+  synchronize: # Copy is insanely slow, don't use it
+    src: "{{ PROSPECTUS_SRC_DIR }}" # /godata/pipelines/stage-prospectus/target/public
```
fredsmith: this code isn't going to run on the goagent, it runs on a running instance that's started by your AMI build pipeline.  You'll need to ship these assets somewhere (s3, git) or build them on the AMI.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -0,0 +1,29 @@
+{%- if "prospectus" in nginx_default_sites -%}
+  {%- set default_site = "default_server" -%}
+{%- else -%}
+  {%- set default_site = "" -%}
+{%- endif -%}
+
+server {
+  # Prospectus configuration file for nginx, templated by ansible
+
+  {% if NGINX_PROSPECTUS_PROXY_INTERCEPT_ERRORS %}
```
fredsmith, line 10: you should set defaults for all variables used in your templates. use !null if you want them to evaluate this way.

playbooks/prospectus.yml
```diff
@@ -0,0 +1,25 @@
...
+  roles:
+#    - aws # Might need to add this back in
+    - role: nginx
+      nginx_app_dir: "{{ PROSPECTUS_DATA_DIR }}"
+      nginx_sites:
+      - prospectus
+      nginx_default_sites:
+      - prospectus
+      PROSPECTUS_NGINX_PORT: 8000
+      PROSPECTUS_PROXY_PASS: 'https://prod-edx-mktg-backend.edx.org'
```
fredsmith: this should probably default to something else, and be overridden by an internal variable in the environment, instead of being passed in to the role invocation in the playbook.

fredsmith (changes requested):

playbooks/roles/prospectus/tasks/main.yml
jlajoie: Oh okay good to know, so I should be able to do a step that clones in the repository on the AMI, does the build via npm, then moves those assets over?

playbooks/prospectus.yml
jlajoie: Defaulting to something else because this is production, right? So setting this so something like the stage backend would be okay?

fredsmith: I would default to something useful to the open source community or devstack.  I realize that this code isn't expected to be used much outside of edx.org yet, but maybe default to the LMS?

fredsmith (approved):

PR #4769, opened 2018-09-12
Add Allow-Credentials header so POSTs can work with cross origin

tuchfarber:
@georgebabey was right.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/cors-add-header.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/cors-add-header.j2
@@ -7,6 +7,7 @@
         add_header 'Content-Length' 0;
         return 204;
     }
-    
+
     add_header 'Access-Control-Allow-Origin' $cors_origin always;
     add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS' always;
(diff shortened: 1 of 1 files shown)


tuchfarber:
@georgebabey Wasn't sure if this was supposed to sit in 'OPTIONS' only section or where I have it.

georgebabey (approved):
My understanding is that because `GET`s may also need credentials to be passed, and those don't require the OPTION preflights, you should include the `Access-Control-Allow-Credentials` on all request.

feanil:
This will add it for all the IDAs is that the right intent?  Any security concerns with allowing this everywhere?

georgebabey:
@feanil Yes I believe all IDAs should set this. It is needed to allow SSO via JWT Cookies throughout our applications.

The only reason we didn't need it for the enterprise portal is we are not currently following the JWT Cookies and instead passing authorization bearer tokens, but that doesn't work for SSO scenarios.

tuchfarber:
As I'm reading it, it would only be set in the following two cases:
1) The /static/ directory of all IDAs. There shouldn't be any security issue as there are no APIs, just assets. 
2) IDAs with `edx_django_service_allow_cors_headers` set to True, which right now seems to be Analytics API and Journals. These both also have CORS whitelists, which protects us from the CSRF attack this would otherwise make available.

feanil (approved):

PR #4802, opened 2018-09-26
Allows Journals to accept uploads larger than 1MB

tuchfarber:
1MB is the default Nginx file upload limit. We need journals to be 10MB.

I'm not sure if this is the right way to implement this. 
Possible issues:
 - Should it be done in the playbooks/roles/nginx/templates/edx/app/nginx/sites-available/? This area was harder for me to understand.
 - Should it be done at a different layer? Not entirely sure how many layers lie between the server and the client and if this is on the correct one. I think it's right. 

This PR also contains a small change to a header that I missed earlier. If you need me to break that out into it's own PR, I can, I just figured it'd be easier to deploy once. 

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/allow-larger-uploads.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/allow-larger-uploads.j2
@@ -1,1 +1,0 @@
-client_max_body_size {{ edx_django_service_max_webserver_upload }}M;
--- a/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/allow-larger-uploads.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/allow-larger-uploads.j2
@@ -1,0 +1,1 @@
+client_max_body_size {{ edx_django_service_max_webserver_upload }}M;
--- a/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/cors-add-header.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/cors-add-header.j2
@@ -2,7 +2,6 @@
         add_header 'Access-Control-Allow-Origin' $cors_origin;
         add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
         add_header 'Access-Control-Allow-Headers' 'Authorization';
-        add_header 'Access-Control-Allow-Credentials' true;
         add_header 'Access-Control-Max-Age' 86400;
         add_header 'Content-Type' 'text/plain; charset=utf-8';
         add_header 'Content-Length' 0;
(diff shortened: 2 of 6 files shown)


tuchfarber:
Tested with a sandbox. Even with this setting set, I'm still getting 413 Entity too large error. Not sure where this needs to be set. Any help would greatly appreciated.

tuchfarber:
Moving setting to main app server block instead of app proxy manually seems to fix the issue. Testing now.

tuchfarber:
Current changes work as expected and Travis is green even though it's not updating from yellow in Github. Ready for review and merge.

feanil:
Your ELB sessions aren't sticky right now so this is probably not the right way to do this.  Why not upload directly to s3 like we do from studio?

tuchfarber:
I'm not sure how I would bypass the app server in this case. The upload is done via the the 3rd party Wagtail CMS which saves to the default storage engine, which django-storages sets to the S3 bucket. Is there some way I can change this so we can keep wagtail / the app server in control of everything? I don't know if making ELB sessions sticky is difficult or otherwise a thing we don't want to do.

feanil:
ELB stickiness is generally a thing we don't want to enable because it can lead to other issues and won't completely fix the upload problem.  A sticky ELB will allow packets to all be uploaded to the same server but depending on how slow the connection is and how long a session remains sticky, this won't be sufficient to fix the problem.   For small files this is not an issue but for large files this would become a problem pretty quickly.

10M is not that big so we can solve it with a pretty low sticky session length but it's not a 100% fix.

tuchfarber:
Look into your earlier suggestion and in response to someone suggesting direct to S3 uploads, the main developer said:
> it's not really possible to have full separation between the app server and image files.

With that in mind, I'm not what option would be a 100% fix. I'm okay with a 99% fix for now. Could you point me on how to add stickiness to the ELB (with suggested session length)? I can add that to this PR after. 

Assuming we add that, is the rest of the PR the correct way to do things?

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/cors-add-header.j2
```diff
@@ -2,6 +2,7 @@
         add_header 'Access-Control-Allow-Origin' $cors_origin;
         add_header 'Access-Control-Allow-Methods' 'GET, POST, OPTIONS';
         add_header 'Access-Control-Allow-Headers' 'Authorization';
+        add_header 'Access-Control-Allow-Credentials' true;
```
coryleeio: Can you remove this - put it in a separate PR and put it behind a switch that defaults to 'off'

coryleeio: as it stands it would affect all edx_django_service IDAs, and we need to be mindful about where / when we turn this on.. we also don't want to turn it on for people in the community

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/allow-larger-uploads.j2
```diff
@@ -0,0 +1 @@
+client_max_body_size {{ edx_django_service_max_webserver_upload }}M;
```
coryleeio: I think you can just inline this since it is not used anywhere else, and is specific to the edx_django_service_role

coryleeio (dismissed):

tuchfarber:
Requested changes made. Separate PR for allow-credentials change: https://github.com/edx/configuration/pull/4816

feanil (approved):
You'll also want to update the ELB stickiness for your ELB in terraform.  I suggest nothing higher than 2 minutes to prevent major future problems.  If you think you need to make it higher than that, let's talk an understand what you're trading for longer stickiness.

tuchfarber:
ELB stickiness PR: https://github.com/edx/terraform/pull/1234 (I can open a new ticket for it if need be). Also, @coryleeio I dismissed the outdated review after the fixes were made and approval by Feanil so I could merge.

PR #4813, opened 2018-10-02
Update existing ansible to install latest Android SDK tools & API

nedo:
Configuration Pull Request
---

Make sure that the following steps are done before merging:

  - [ ] A DevOps team member has approved the PR.
  - [ ] Are you adding any new default values that need to be overridden when this change goes live? If so:
    - [ ] Update the appropriate internal repo (be sure to update for all our environments)
    - [ ] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [ ] Add an entry to the CHANGELOG.
  - [ ] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).

PR #4814, opened 2018-10-02
naveed/node - Install nodejs and required packages to necessary for android automation

nedo:
Configuration Pull Request
---


playbooks/roles/jenkins_worker/tasks/node.yml
```diff
@@ -0,0 +1,19 @@
+---
+# Install nodejs and required packages to run android automation
+
+# get node package
+- name: install nodejs
+  shell: "curl -sL https://deb.nodesource.com/setup_{{node_version}} | sudo -E bash -"
+  when: android_worker is defined
+# install nodejs
+- name: install nodejs
+  apt: name=nodejs
```
estute: can you change this part to follow our conventions? https://openedx.atlassian.net/wiki/spaces/OpenOPS/pages/26837527/Ansible+Code+Conventions#AnsibleCodeConventions-Formatting

playbooks/roles/jenkins_worker/defaults/main.yml
```diff
@@ -15,3 +15,5 @@ jenkins_debian_pkgs:
 
 # packer direct download URL
 packer_url: "https://releases.hashicorp.com/packer/0.8.6/packer_0.8.6_linux_amd64.zip"
+
+node_version: "8.x"
```
estute: I am not familiar with this versioning style. I think we should put explicit versions in our configuration. It makes it easier to reason about.

playbooks/roles/jenkins_worker/tasks/node.yml
```diff
@@ -0,0 +1,19 @@
...
+  when: android_worker is defined
+# Install appium
```
estute: Both appium and webdriver are testing dependencies. Therefore, I think they should be placed in the repo that contains the test code. This way, you don't need to rebuild an AMI in case you need to update or add a library.

playbooks/roles/jenkins_worker/tasks/node.yml
```diff
@@ -0,0 +1,19 @@
+---
+# Install nodejs and required packages to run android automation
+
+# get node package
+- name: install nodejs
+  shell: "curl -sL https://deb.nodesource.com/setup_{{node_version}} | sudo -E bash -"
```
estute: In general, it is not advisable to pipe curl to bash (see: https://stackoverflow.com/questions/29382739/why-using-curl-sudo-sh-is-not-advised).

playbooks/roles/jenkins_worker/defaults/main.yml
estute: If it is the case that you cannot specify an exact version, you can look into using nodenv, see https://github.com/edx/edx-platform/blob/master/scripts/jenkins-common.sh#L10

estute (changes requested):
See inline comments

playbooks/roles/jenkins_worker/tasks/node.yml
nedo: Alright, I'll put `appium` & `wd` installation as testing dependency

playbooks/roles/jenkins_worker/defaults/main.yml
nedo: I've put explicit version on Node

playbooks/roles/jenkins_worker/tasks/node.yml
nedo: done

playbooks/roles/jenkins_worker/tasks/node.yml
nedo: I've replaced `curl` with `apt`, please have a look and comment

playbooks/roles/jenkins_worker/defaults/main.yml
nedo: @estute seems like nodejs has `version.x` scheme now, please have a look on their repos, 
 
https://github.com/nodesource/distributions/tree/master/deb
https://nodejs.org/en/download/package-manager/

If you still feels using version in `x` is not fine, we can probably get their `tar` and install these through shell. wdty?

nedo:
Hi @estute I've fixed all points based on the recommendations, please have another pass. 
cc @staubina

estute (approved):
In general, looks good. However, two things:
1) you will nneed to include node.yml in main.yml, similar to this: https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_worker/tasks/main.yml#L12
2) make sure you can build the role on the build-packer-ami job, as I mentioned on my final comment here: https://github.com/edx/configuration/pull/4813

nedo:
@estute I am able to successfully build AMI, can you please have a last pass, before I merge this PR?

cc @staubina

PR #4841, opened 2018-10-12
Compress prospectus images after build

mikix:
Compress total size of all images in prospectus from 1.1M to 316K (71% reduction).

This process *is* technically lossy, but I cannot tell the difference in the quality. Compare for yourself:

edx-masters-aboutgt edx-masters-aboutgt

world_map world_map

harleen harleen

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/prospectus/tasks/main.yml
+++ b/playbooks/roles/prospectus/tasks/main.yml
@@ -153,6 +153,19 @@
     - install
     - install:app-requirements
 
+- name: Install pngquant
+  apt:
+    name: "pngquant"
+  tags:
+    - install
+    - install:system-requirements
+
+- name: Compress images
+  shell: "find public -name '*.png' -exec pngquant --skip-if-larger --quality 50-75 --ext .png --force -- {} \\;"
+  args:
+    chdir: "{{ prospectus_code_dir }}"
+  become_user: "{{ prospectus_user }}"
+
 # Copy over the target from the previous build to where it needs to be
 - name: Create data folder
   file:


playbooks/roles/prospectus/tasks/main.yml
```diff
@@ -153,6 +153,19 @@
...
 
+- name: Install pngquant
+  apt:
+    name: "pngquant"
+  tags:
+    - install
+    - install:system-requirements
+
+- name: Compress images
+  shell: "find public -name '*.png' -exec pngquant --skip-if-larger --quality 50-75 --ext .png --force -- {} \\;"
```
jlajoie, line 164: Just wanted to double check, this just rewrites the files correct? It won't give them new names?

mikix: Correct. pngquant by default uses a new suffix/file. But the "--ext .png" flag tricks it into overwriting the input file.

feanil (approved):
Seems fine but doesn't need to block on a devops review I think.  This is a change in a role for an app that you control.  Happy to provide advice or best practices but don't think it needs to block on devops.

mikix:
Good news, @feanil . Thanks for the look anyway

playbooks/roles/prospectus/tasks/main.yml
```diff
@@ -153,6 +153,19 @@
...
     - install:app-requirements
 
+- name: Install pngquant
+  apt:
+    name: "pngquant"
+  tags:
+    - install
+    - install:system-requirements
+
+- name: Compress images
```
fredsmith, line 163: So, my guess is that this will modify the git checked in images on disk, which means the git state will be dirty, which will cause subsequent checkouts to fail.  I think this process and the goal here are fine, but our build process shouldn't modify the checked in assets within the checked out repo.

fredsmith (dismissed):
Please correct me if I'm wrong about how this will work.  We shouldn't programmatically modify assets within the code_dir as part of the build.

playbooks/roles/prospectus/tasks/main.yml
mikix: The 'public' directory is a git-ignored build directory. ('static' holds the checked-in images)

PR #4856, opened 2018-10-31
Allow all django services to override how often they restart.

feanil:
In particular update the ecommerce service to restart every 3000
requests by default.  This is because we believe there are some issues
with the segment library when run in gevent mode that cause it to take
up more and more CPU over time.

Configuration Pull Request
---


BenjiLee:
I'm looking at NewRelic reports and it looks like we handle around 50k - 100k requests per day. Would 3000 be too frequent? Sounds like you might be looking at a more specific transaction.

feanil:
@BenjiLee that's across all gunicorn workers, so each worker would be doing less than that.  Also about 24hrs is when we first start seeing an impact so doing it more regularly than that makes sense to me.

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+- Role: edx_django_service
+  - Added new overridable variable `edx_django_service_gunicorn_max_requests`
+- Role: ecommerce
+  - Set default max_requests to 6000.(eg. restart gunicorn process every 3000 requests.)
```
BenjiLee: Did you mean to say 6000->3000?

feanil: Fixed.

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -60,6 +60,7 @@ edx_django_service_gunicorn_port: null
 edx_django_service_gunicorn_timeout: 300
 edx_django_service_gunicorn_workers: 2
 edx_django_service_gunicorn_worker_class: 'gevent'
+edx_django_service_gunicorn_max_requests: null
```
BenjiLee, line 63: What's the behaviour what this is set to null?

feanil: The template has a conditional which will evaluate to `False` which should prevent the max_requests config from being set in that file.  Gunicorn will fall-back to its builtin default of not rebooting the gunicorn workers.

robrap: @feanil: I hope [30 second graceful timeout](http://docs.gunicorn.org/en/0.17.2/configure.html#graceful-timeout) is enough. ;)  Hopefully we don't need to restart while discovery is slow or down.

robrap (commented):
@feanil: Thanks for doing this!

feanil:
@robrap @BenjiLee  thanks for the review.  Any chance of getting an approval?

robrap (approved):
@feanil: Thanks!  Do you want to add a note and reference to this PR to the relevant tickets, in case we ever want to see if we can remove this?

BenjiLee (approved):

feanil:
@robrap I think this is good to keep around generally.  It doesn't hurt and deals with a class of issues that we probably don't want to spend too much time on.

PR #4858, opened 2018-10-31
Add JWT refresh API endpoint to LMS nginx config.

douglashall:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -188,6 +188,10 @@ error_page {{ k }} {{ v }};
     try_files $uri @proxy_to_lms_app_api;
   }
 
+  location /login_refresh {
+    try_files $uri @proxy_to_lms_app_api;
+  }
+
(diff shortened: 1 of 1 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
```diff
@@ -188,6 +188,11 @@ error_page {{ k }} {{ v }};
     try_files $uri @proxy_to_lms_app_api;
   }
 
+  # JWT refresh API endpoint
+  location /login_refresh {
```
robrap, line 191: Nit: Comment uses `JWT refresh`, but the path is named `login_refresh`.  Does that mean the path could have been named differently, or just updating the comment, etc.?

Also, some of these start with `No basic auth for ...`.  If you don't want to repeat that, and the name is clear enough, maybe no comments are needed here?

douglashall: I can remove the comment if you think that would be better?

nasthagiri (approved):

PR #4867, opened 2018-11-06
basic it jenkins role

estute:
This sets up the role for provisioning a Jenkins server for IT. The main difference with this instance will be it's auth mechanism (SAML). Aside from that, it is a simplified installation, compared to build Jenkins.

NOTE: until we finish setting up the secrets/credentials for this role, the credentials script is commented out


estute:
cc @wesmason @caplan188

playbooks/roles/jenkins_it/defaults/main.yml
```diff
@@ -0,0 +1,290 @@
...
+    version: '2.4'
+    group: 'org.jvnet.hudson.plugins'
+  - name: 'htmlpublisher'
+    version: '1.16'
+    group: 'org.jenkins-ci.plugins'
+  - name: 'javadoc'
+    version: '1.3'
+    group: 'org.jenkins-ci.plugins'
+  - name: 'job-dsl'
+    version: '1.67'
```
jmbowman: 1.70 is available already; any reason not to start with it?  Should we just start all these plugins at the latest available versions?

playbooks/roles/jenkins_it/defaults/main.yml
```diff
@@ -0,0 +1,290 @@
+it_jenkins_user_uid: 1002
+it_jenkins_group_gid: 1004
+it_jenkins_version: jenkins_2.89.4
```
jmbowman, line 3: Is there a compelling reason not to start out with the latest stable release, given that there aren't any existing jobs on this server?

playbooks/roles/jenkins_it/defaults/main.yml
```diff
@@ -0,0 +1,290 @@
...
+  - name: 'timestamper'
+    version: '1.8.9'
+    group: 'org.jenkins-ci.plugins'
+  - name: 'token-macro'
+    version: '2.3'
+    group: 'org.jenkins-ci.plugins'
+  - name: 'translation'
+    version: '1.16'
+    group: 'org.jenkins-ci.plugins'
+  - name: 'violations'
```
jmbowman, line 225: This plugin has been largely replaced by `warnings` - is there a reason we need to keep it?

playbooks/roles/jenkins_it/defaults/main.yml
estute: One of the goals in having Jenkins configuration be shared between the various instances here at edx is to reduce drift. I would love to use the most recent stable version, but that can have implications on plugin versions, etc. Additionally, all of our groovy scripts are tested with the version mentioned here.

playbooks/roles/jenkins_it/defaults/main.yml
estute: Ah, I had created this PR before you landed the DSL upgrade. I will update the DSL and gradle plugins. As for the rest of them, I am trying to reduce differences between our Jenkins instances, so the jobs can be seeded and run everywhere without issue

playbooks/roles/jenkins_it/defaults/main.yml
michaelyoungstrom: hmm, I'd semi push back on that. I don't necessarily think other teams should be held back version wise just because of build jenkins specific reasons. The testing point is valid, but also kind of easy to check locally for now (and we could go back to testing more than one jenkins version potentially in that repo). I'm okay with whatever you decide but just thinking out loud.

michaelyoungstrom: That being said... I do agree with making it easier to maintain for us. Sort of forgot we are the ones that will continue managing most of it.

estute: How about this: I set up this box using the version I have specified in the PR. Then, we should priortize https://openedx.atlassian.net/browse/TE-2809. Changing the version here should be easy following that. @michaelyoungstrom

jmbowman (approved):
Want to get both this and build Jenkins fully upgraded pretty quickly, but this should be fine for now.

playbooks/roles/jenkins_it/defaults/main.yml
michaelyoungstrom: We're blocked on that ticket so it can't be right away, but yeah it's cool with me

michaelyoungstrom (approved):

playbooks/roles/jenkins_it/defaults/main.yml
estute: how are we blocked?

PR #4868, opened 2018-11-06
Remove basic auth from login page.

douglashall:
Basic auth on the login page is preventing the refresh token flow from working for frontend apps.


georgebabey:
Should just set the flag to false instead of removing the option entirely? I can imagine some openedx installs where basic auth is fine for their usage.

douglashall:
No. I think we want to leave it on for the / location. I could introduce a new flag, but that would defeat the purpose.

georgebabey:
@douglashall Ah I missed this was just for /login

georgebabey (approved):

PR #4872, opened 2018-11-09
TE-2804 Upgrade Job DSL plugin

jmbowman:
Configuration Pull Request
---

Upgrading the Jenkins Job DSL plugin to gain the `lightweight` option for retrieving a Jenkinsfile without cloning the entire repo, plus support for shallow clones.   Also updating the gradle plugin since the new Job DSL version doesn't work with the version we have now.  Finally, made some changes to `seed_job.xml` to match the plugin versions we're now installing.


estute (commented):
@jmbowman did you check https://github.com/jenkinsci/job-dsl-plugin/wiki/Migration? This is helpful for seeing what changes have occurred and what will be added/dropped in future releases

jmbowman:
Yes, already went through that; that's where I noticed that the gradle plugin needed to be upgraded.

estute (approved):
:+1:

PR #4885, opened 2018-11-26
Add defaut writable gradebook URL env variable.

iloveagent57:
https://openedx.atlassian.net/browse/EDUCATOR-3657

This is to point to the LMS at the new Gradebook frontend.
---

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -916,6 +916,9 @@ EDXAPP_MAINTENANCE_BANNER_TEXT: "Sample banner message"
 EDXAPP_PASSWORD_POLICY_COMPLIANCE_ROLLOUT_CONFIG:
   ENFORCE_COMPLIANCE_ON_LOGIN: false
 
+# Needed to link the LMS instructor dashboard to the writable gradebook micro-frontend
+EDXAPP_LMS_WRITABLE_GRADEBOOK_URL: null
+
 #-------- Everything below this line is internal to the role ------------
 
 #Use YAML references (& and *) and hash merge <<: to factor out shared settings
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -916,6 +916,9 @@ EDXAPP_MAINTENANCE_BANNER_TEXT: "Sample banner message"
 EDXAPP_PASSWORD_POLICY_COMPLIANCE_ROLLOUT_CONFIG:
   ENFORCE_COMPLIANCE_ON_LOGIN: false
 
+# Needed to link the LMS instructor dashboard to the writable gradebook micro-frontend
+EDXAPP_LMS_WRITABLE_GRADEBOOK_URL: 'http://localhost:1991'
```
feanil: Will this work with devstacks?  Usually the default here would be whatever you want to have for devstacks.  Maybe this should be `null` by default?  Does the LMS do the right thing if this value is not configured?

iloveagent57: Good question, here's the platform PR: https://github.com/edx/edx-platform/pull/19316

Gradebook can run locally in a container bound to port 1991 and is accessible on the `default_devstack` docker network.  On the platform side, we check if a course waffle flag is on for the given course before doing anything that accesses `WRITABLE_GRADEBOOK_URL`, so if you don't have gradebook running locally, you probably don't have that flag enabled.  On the other hand, we probably want to remove that waffle flag someday, and it might make sense then to have this variable default to `null` and let the developer override it in `private.py`.  I can make that change and update the gradebook README to reflect that strategy.

schenedx (approved):
👍

iloveagent57:
@feanil Does this look good to you now?

coryleeio (approved):
checked that loadtest and edge are good

PR #4897, opened 2018-12-04
Adding tags to splunk web config

coryleeio:


coryleeio:
These changes make it so that running, atleast the web part, is a no-op before I set SPLUNK_PROXY_ENABLED to something else

jdmulloy (approved):

PR #4898, opened 2018-12-05
Restrict ecommerce admin endpoints access to VPN CIDRS (OPS-3138)

nadeemshahzad:
Configuration Pull Request
---


playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/proxy-to-app.j2
```diff
@@ -12,6 +12,8 @@ location ~ ^/({{ edx_django_service_basic_auth_exempted_paths | join('|') }})/ {
 
 {% include "concerns/django_admin_access_from_restricted_cidrs.j2" %}
 
+{% include "concerns/admin_endpoints_access_from_restricted_cidrs.j2" %}
```
coryleeio: Can we more clearly show that this is specific to ecommerce?
At the moment it is hardcoded to block specific endpoints on ecommerce 'courses|coupons|programs|credit|dashboard|management' that could even be harmful if they were accidentally turned on elsewhere.

I wonder if we could combine these into a CIDR restricted endpoints template, and pass in a list of urls and cidrs?

If that isn't easy to do I think it would atleast make sense to change this to be obviously specific to ecommerce.

coryleeio (changes requested):

nadeemshahzad:
As we need to restrict the django admin for ecommerce all environments, so keeping it default and overriding admin URLs for stage environment ( https://github.com/edx/edx-internal/pull/662 )

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -182,5 +182,7 @@ NGINX_EDXAPP_LMS_APP_EXTRA: ""
 
 # List of subnet or IP addressess to allow to access django admin 
 NGINX_DJANGO_ADMIN_ACCESS_CIDRS: []
+NGINX_ADMIN_ACCESS_CIDRS: "{{ NGINX_DJANGO_ADMIN_ACCESS_CIDRS }}"
```
coryleeio: I think you can remove NGINX_DJANGO_ADMIN_ACCESS_CIDRS because it is no longer being read anywhere

coryleeio: After cleaning that up feel free to merge - I don't want to block it another day.

coryleeio (approved):

PR #4899, opened 2018-12-05
Changes required to terminate splunk TLS correctly at the proxy

coryleeio:


jdmulloy (approved):

jdmulloy:
Will this cause issues in a future version of Splunk?

coryleeio:
It is possible.  If it does we can terminate TLS at splunk as a fallback, but hopefully they will make it easier to put it behind a proxy like everyone else.

PR #4906, opened 2018-12-17
Render auth and env json to single yml file

coryleeio:


docker/build/edxapp/ansible_overrides.yml
```diff
@@ -14,7 +14,6 @@ EDXAPP_LMS_GUNICORN_EXTRA_CONF: 'reload = True'
 
 EDXAPP_NO_PREREQ_INSTALL: 0
 COMMON_SSH_PASSWORD_AUTH: "yes"
-EDXAPP_LMS_BASE: "edx.devstack.lms:18000"
```
coryleeio, line 17: This also exists in /home/coryleeio/Workspace/configuration/docker/build/edxapp/ansible_overrides.yml
where it has the same value, so it was removed

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,6 +16,58 @@
     - install:app-configuration
     - edxapp_cfg # Old deprecated tag, will remove when possible
 
+- name: combined lms auth env for yml
+  set_fact:
+    lms_combined_config: '{{lms_env_config|combine(lms_auth_config)}}'
```
coryleeio, line 21: The only key collision between auth and env json in any environment

prod,edge,stage,devstack is DEFAULT_FILE_STORAGE

In all cases, it has the same value in both auth and env.json, and ansible merge will do the right thing with this.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,6 +16,58 @@
...
+  when: '"lms" in service_variants_enabled'
+  tags:
+    - install
+    - install:configuration
+    - install:app-configuration
+    - edxapp_cfg # Old deprecated tag, will remove when possible
+
+- name: combined cms auth env for yml
+  set_fact:
+    cms_combined_config: '{{cms_env_config|combine(cms_auth_config)}}'
```
coryleeio, line 47: The only key collision between auth and env json in any environment

prod,edge,stage,devstack is DEFAULT_FILE_STORAGE

In all cases, it has the same value in both auth and env.json, and ansible merge will do the right thing with this.

coryleeio:
This should render the yml file, but not activate it.  It should work for baked config and work for devstack.
I checked the collisions and found with 1 exception which I fixed that I could do a dumb ansible hash merge on the two and it would work correctly, my PR includes a util for checking top level keys for collisions.  I have additional temporary tooling on my laptop if anyone would like to see.

coryleeio:
sandbox is here:
coryleeio.sandbox.edx.org

has studio.yml and lms.yml

util/check_for_key_collisions/check_for_key_collisions.py
```diff
@@ -0,0 +1,40 @@
...
+    for file_path in files:
+        content = None
+        if ".yml" in file_path or ".yaml" in file_path:
+            stream = file(file_path, 'r')
+            content = yaml.load(stream)
+        elif ".json" in file_path:
+            with open(file_path, "r") as read_file:
+                content = json.load(read_file)
+        for key, value in content.iteritems():
+            if key not in values_for_keys:
```
feanil: Use a [deafultdict](https://docs.python.org/3.7/library/collections.html#collections.defaultdict)

util/check_for_key_collisions/check_for_key_collisions.py
```diff
@@ -0,0 +1,40 @@
...
+import yaml
+import json
+
+@click.command()
+@click.option('--files', '-m', multiple=True)
+def check_for_yaml_key_collisions(files):
+    values_for_keys = {}
+    for file_path in files:
+        content = None
+        if ".yml" in file_path or ".yaml" in file_path:
```
feanil: use [endswith](https://docs.python.org/3.7/library/stdtypes.html?highlight=endswith#str.endswith) which should be safer.

util/check_for_key_collisions/check_for_key_collisions.py
```diff
@@ -0,0 +1,40 @@
...
+            stream = file(file_path, 'r')
+            content = yaml.load(stream)
+        elif ".json" in file_path:
```
feanil: Use [endswith](https://docs.python.org/3.7/library/stdtypes.html?highlight=endswith#str.endswith) which should be safer.

feanil (commented):
Generally looks good, just some small changes in the python script.

docker/build/edxapp/ansible_overrides.yml
feanil: That is the same path as this file.  It looks like it's defined in `docker/build/devstack/ansible_overrides.yml`

Do both of these files get pulled in for edxapp builds?  It seems like this should stay defined here closer to the app.

jdmulloy (approved):

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,6 +16,58 @@
...
+  tags:
+    - install
+    - install:configuration
+    - install:app-configuration
+    - edxapp_cfg # Old deprecated tag, will remove when possible
+
+- name: render lms yml config # meant to replace existing json config eventually
+  template:
+    src: "lms.yml.j2"
+    dest: "{{ edxapp_app_dir }}/lms.yml"
```
feanil: This should live under `{{ COMMON_CFG_DIR }}/lms.yml` instead of in the app directory.  This will make it better match the other applications.

Example of where the Xqueue config is: https://github.com/edx/configuration/blob/cf4d221d384ed396a3008c31487f385544ef08e7/playbooks/roles/xqueue/tasks/main.yml#L109

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,6 +16,58 @@
...
+  tags:
+    - install
+    - install:configuration
+    - install:app-configuration
+    - edxapp_cfg # Old deprecated tag, will remove when possible
+
+- name: render studio yml config # meant to replace existing json config eventually
+  template:
+    src: "studio.yml.j2"
+    dest: "{{ edxapp_app_dir }}/studio.yml"
```
feanil: This should live under `{{ COMMON_CFG_DIR }}/lms.yml` instead of in the app directory.  This will make it better match the other applications.

Example of where the Xqueue config is: https://github.com/edx/configuration/blob/cf4d221d384ed396a3008c31487f385544ef08e7/playbooks/roles/xqueue/tasks/main.yml#L109

feanil (approved):
One last thing, I think this is good to go after that.

playbooks/roles/edxapp/tasks/service_variant_config.yml
coryleeio: Ah good catch - I will make that change

PR #4907, opened 2018-12-18
return the current tasks from the provided queue

syed-awais-ali:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/check_celery_progress/check_celery_progress.py
+++ b/util/jenkins/check_celery_progress/check_celery_progress.py
@@ -208,7 +208,6 @@ def extract_body(task):
         body_dict = json.loads(body.decode("utf-8"))
     except:
         body_dict = {}
-
     return body_dict
 
 
@@ -216,6 +215,7 @@ def generate_info(
     queue_name,
     correlation_id,
     body,
+    running_tasks,
     do_alert,
     first_occurance_time,
     current_time,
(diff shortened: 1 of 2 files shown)


util/jenkins/check_celery_progress/check_celery_progress.py
```diff
@@ -324,7 +324,9 @@ def check_queues(host, port, environment, deploy, default_threshold, queue_thres
         correlation_id = new_state[queue_name]['correlation_id']
         first_occurance_time = new_state[queue_name]['first_occurance_time']
         body = extract_body(queue_first_items[queue_name])
-        redacted_body = {'task': body['task'], 'args': 'REDACTED', 'kwargs': 'REDACTED'}
+        # call to get active tasks from queue
+        active_tasks = get_current_tasks(host, port, queue_first_items[queue_name])
```
coryleeio: Can you modify the signature of generate_info and pass this in as a parameter?

The redacted body is only sent to opsgenie, but isn't printed in the jenkins log, we want the task to be printed in both.

coryleeio: Can you change task to 'next task' and have your thing be currently running tasks? or something to that effect as well?

https://tools-edx-jenkins.edx.org/job/Monitoring/job/check-celery-progress-prod-edx/98005/console

If you check out the output here it shows 'Task' but that is the 'Next task', you are adding 'Currently active tasks'

coryleeio (changes requested):

coryleeio (approved):

PR #4916, opened 2019-01-02
fix celery script

syed-awais-ali:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/check_celery_progress/check_celery_progress.py
+++ b/util/jenkins/check_celery_progress/check_celery_progress.py
@@ -286,7 +286,7 @@ def generate_info(
 @click.option('--opsgenie-api-key', '-k', envvar='OPSGENIE_API_KEY', required=True)
 @click.option('--jenkins-build-url', '-j', envvar='BUILD_URL', required=False)
 def check_queues(host, port, environment, deploy, default_threshold, queue_threshold, opsgenie_api_key, jenkins_build_url):
-
+    redacted_body = ""
     thresholds = dict(queue_threshold)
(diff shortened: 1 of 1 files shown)


util/jenkins/check_celery_progress/check_celery_progress.py
```diff
@@ -326,7 +324,12 @@ def check_queues(host, port, environment, deploy, default_threshold, queue_thres
         first_occurance_time = new_state[queue_name]['first_occurance_time']
         body = extract_body(queue_first_items[queue_name])
         active_tasks = get_current_tasks(host, port, queue_name)
-        redacted_body = {'task': body['task'], 'args': 'REDACTED', 'kwargs': 'REDACTED'}
+        # placed redacted_body code in try/except block to avoid failing since
+        # the task was failing for prod-edx E-D
+        try:
+            redacted_body = {'task': body['task'], 'args': 'REDACTED', 'kwargs': 'REDACTED'}
+        except Exception as e:
+            print("Exception: Could not find {} in body".format(e))
```
feanil: This would only happen when a body has no 'task' key in it right?  What does that mean?  Why would that occur?

jdmulloy: This is catching all exceptions, it should be narrowed to only catch the appropriate exception for a missing key, which I believe is "KeyErro". Also it would be useful to print the entire body so we can better understand the error condition.

jdmulloy (changes requested):

jdmulloy (approved):

PR #4924, opened 2019-01-04
Changes required for xserver to use yaml config instead of json and to work with pipelines

coryleeio:


jdmulloy (approved):
Looks reasonable to me. Can you squash your commits before you merge?

PR #4929, opened 2019-01-07
Removing calls to the blank docker playbook I removed that just ran common on all of our docker images

coryleeio:


coryleeio:
The old docker playbook did nothing but run common, I've removed it, so the lack of docker playbook is causing these to fail

Common runs common_vars, so this can be removed as well

jdmulloy (approved):

PR #4933, opened 2019-01-09
Update node version.

dianakhuang:
Update of https://github.com/edx/configuration/pull/4927
Depends on updating `nodeenv` in `edx-platform`: https://github.com/edx/edx-platform/pull/19567

https://openedx.atlassian.net/browse/ARCH-352

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_django_service/defaults/main.yml
+++ b/playbooks/roles/edx_django_service/defaults/main.yml
@@ -31,7 +31,7 @@ edx_django_service_nodeenv_dir: '{{ edx_django_service_home }}/nodeenvs/{{ edx_d
 edx_django_service_nodeenv_bin: '{{ edx_django_service_nodeenv_dir }}/bin'
 edx_django_service_node_modules_dir: '{{ edx_django_service_code_dir }}/node_modules'
 edx_django_service_node_bin: '{{ edx_django_service_node_modules_dir }}/.bin'
-edx_django_service_node_version: '{{ common_node_version }}'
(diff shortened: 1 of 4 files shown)


playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -31,7 +31,7 @@ edx_django_service_nodeenv_dir: '{{ edx_django_service_home }}/nodeenvs/{{ edx_d
 edx_django_service_nodeenv_bin: '{{ edx_django_service_nodeenv_dir }}/bin'
 edx_django_service_node_modules_dir: '{{ edx_django_service_code_dir }}/node_modules'
 edx_django_service_node_bin: '{{ edx_django_service_node_modules_dir }}/.bin'
-edx_django_service_node_version: '{{ common_node_version }}'
+edx_django_service_node_version: '8.9.3'
```
davidjoy, line 34: Is this to keep edx_django_service_node_version pinned to something older because it breaks if we update it?

dianakhuang: Sorry, I just keep leaving bits and pieces of the plan in different places, and it's not super consolidated. The plan right now is to keep this pinned for now because it's the version that gets used for all the IDAs. I want to take a quick spin through those for possible breaking packages before switching them onto node 12 as well.

adzuci (approved):
:shipit: Make sure to notify dev@edx.org

PR #4937, opened 2019-01-10
SE-614 Adds DOT Application entry for HTTP basic auth

pomegranited:
If HTTP basic auth is enabled, creates a public Application entry with password grant type, using the
HTTP basic auth user and password as the client ID and secret.

As far as we can tell, this issue arises only in this combination of circumstances:

* https://github.com/edx/edx-platform/pull/17633 now requires that the mobile applications use DOT applications in order to access HTTP and Problem blocks in the LMS.
* OpenCraft's haproxy load balancer service uses Basic Authentication credentials to secure the backend servers that it proxies.  So to pass requests through to the backend servers, haproxy adds an `Authorization: 'Basic $HASH'` header to its requests (unless the request already contains this header).
    This basic auth is intended for nginx, but nginx consequently passes it through to the Django request.
* When responding to the `/oauth2/auth_token` POST request, the [django-oauth-toolkit OAuth2Validator checks this basic auth header first](https://github.com/jazzband/django-oauth-toolkit/blob/master/oauth2_provider/oauth2_validators.py#L48). When it can't find an oauth2_provider `Application` entry that matches the basic auth key and secret, the authentication fails.
  This is a change in behaviour from the previous [django-oauth2-provider access token authorization](https://github.com/caffeinehit/django-oauth2-provider/blob/6b5bc0d3ad706d2aaa47fa476f38406cddd01236/provider/oauth2/views.py#L71-L75), which proceeds through to [using the request POST parameters](https://github.com/caffeinehit/django-oauth2-provider/blob/6b5bc0d3ad706d2aaa47fa476f38406cddd01236/provider/oauth2/backends.py#L47-L61) if the [basic Authorization headers don't match an OAuth2 client](https://github.com/caffeinehit/django-oauth2-provider/blob/6b5bc0d3ad706d2aaa47fa476f38406cddd01236/provider/oauth2/backends.py#L19-L44).

Please see https://github.com/edx/edx-platform/pull/19571 for more background information.

**Dependencies**:  https://github.com/edx/edx-platform/pull/19571

**Merge deadline**: Would like to get this merged into Ironwood if possible.

**Sandbox URL & Testing instructions**: See https://github.com/edx/edx-platform/pull/19571

**Reviewers**
- [ ] @symbolist 

Configuration Pull Request
---


openedx-webhooks:
Thanks for the pull request, @pomegranited! I've created [OSPR-2965](https://openedx.atlassian.net/browse/OSPR-2965) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

natabene:
@pomegranited Thank you for your contribution. @edx/devops Can you have a look at this?

symbolist (approved):
- [x] I tested this by verifying the sandbox provisioning created the correct credentials.
- [x] I read through the code
- [x] I checked for accessibility issues. N/A here.
- [x] Includes documentation

pomegranited:
FYI @natabene  Since the dependent https://github.com/edx/edx-platform/pull/19571 is merged now, this PR is ok to merge now too whenever devops are satisfied.  Thanks!

natabene:
@pomegranited Thanks for the heads up!

pomegranited:
@natabene Do you think this could make it into Ironwood too?  CC @nedbat

natabene:
@feanil Could you give this a look? Ned would take it into Ironwood if you can review. Thanks.

pomegranited:
FYI @feanil , I've rebased this PR on [latest master](https://github.com/edx/configuration/commit/4492f3aad492b66577fd2e1b89faf774af3993cf) to resolve conflicts, and squashed commits f38dca5bdf57a798205a78b04ebfcc0e1621a49e and 7743542d9d4a80e0a84d5c902c413bc091a6de44 into df2dd7e3402c8522d03d8d242ba2074935953fb9 to prepare for merge.

Do you think your team will have time to review before Ironwood is cut?  CC @natabene @nedbat

natabene:
@pomegranited  Thanks for reminding, let me see what we can do.

feanil:
@nasthagiri @doctoryes @pwnage101 can you take a look at this?  I'm not sure why the basic auth user needs DOT credentials here...

pomegranited:
@feanil  CC @nasthagiri  @doctoryes @pwnage101 Sorry, I didn't provide context here..  if you could please read the "Background" part of https://github.com/edx/edx-platform/pull/19571 for why we need this?  Thank you!

doctoryes:
I've never seen HTTP basic auth credentials kept in a DOT application before. Is this a standard way of checking these credentials? Or is this implementation non-standard and based on the fact that the mobile app itself also contains a DOT client id used to auth against the LMS?

If this implementation is a non-standard one, using a DOT application for a purpose which it's not designed sets off my alarms. I've read all the background and the previous PRs - but was unable to answer my questions above.

pomegranited:
@doctoryes Thank you for looking at this PR!  I agree it's an odd one, and welcome any other ways for dealing with this issue we're seeing.

I've updated the PR description to provide more detail about our particular use case.  Please let me know if you have further questions, or any suggestions for how we could handle this differently?

doctoryes:
Thanks for the excellent explanation, @pomegranited ! While the [RFC](https://tools.ietf.org/html/rfc6749#section-2.3.1) seems a bit more permissive and in-line with the DOP implementation, the DOT code to which you've linked shows that the basic auth credentials are indeed expected to be contained in a DOT application. So this DOT application creation looks fine to me.

doctoryes (approved):

doctoryes:
@feanil Would you like to review again before merging?

natabene:
@bderusha  @edx/devops If you are ok with this, can you merge or do you need more time to review?

bderusha:
@natabene I will review today

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -203,6 +203,8 @@ EDXAPP_ENABLE_THIRD_PARTY_AUTH: true
 EDXAPP_ENABLE_OAUTH2_PROVIDER: false
 
 EDXAPP_ENABLE_MOBILE_REST_API: false
+EDXAPP_BASIC_AUTH_DOT_APP_NAME: "Basic Auth"
+EDXAPP_BASIC_AUTH_DOT_APP_USERNAME: "login_service_user"
```
bderusha, line 207: This will fail if this user doesn't exist already.  If we do decide to go this route to solve this we should add a step to create this user.

bderusha (changes requested):
@pomegranited I have some questions about the problem this is trying to solve and wonder if other solutions get at the same thing with less configuration.

1.) Can we make the mobile API endpoints to not require basic auth?
2.) Can you strip out the header in NGINX so the application layer never needs to know about basic auth?

Also, does this prevent service users from accessing edx-platform as themselves?  EG will `ecom_worker` or `discovery_service_user` get overridden by this credential?

pomegranited:
@bderusha thanks for your comments and for taking the time to review here.  We agree that this isn't an ideal solution, but since we've got a workaround and we're out of budget for dealing with this issue, I'm going to close this PR.

> 1.) Can we make the mobile API endpoints to not require basic auth?

Yep -- if we put the OAuth client_id into the `Authorization` header [instead of the request POST](https://github.com/edx/edx-app-ios/blob/ea5a634595155ee46ba7ec7b2c10e1cc0b62cb44/Source/OEXAuthentication.m#L103), then we could whitelist these requests in nginx [like we do for XBlock requests](https://github.com/edx/configuration/pull/4943/files).

>  2.) Can you strip out the header in NGINX so the application layer never needs to know about basic auth?

Maybe.

> Also, does this prevent service users from accessing edx-platform as themselves? EG will `ecom_worker` or `discovery_service_user` get overridden by this credential?

Only if they use request POST to send their OAuth credentials, and I don't believe they do (haven't checked).

openedx-webhooks:
@pomegranited Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/configuration&entry.867055334=https://github.com/edx/configuration/pull/4937&entry.1484655318=https://github.com/pomegranited&entry.752974735=2019-01-10+10:01&entry.1917517419=2019-02-22+02:44&entry.2133058324=No) so we can improve your experience in the future.

PR #4947, opened 2019-01-15
Expose sailthru key from ecomworker to ecommerce service

coryleeio:


coryleeio:
This makes it so that ecomworker and ecommerce both read from the same sailthru key, and exposes that key to ecommerce.

playbooks/roles/ecomworker/defaults/main.yml
```diff
@@ -113,8 +115,8 @@ ECOMMERCE_WORKER_SERVICE_CONFIG:
     SAILTHRU_ENROLL_TEMPLATE: '{{ ECOMMERCE_WORKER_SAILTHRU_ENROLL_TEMPLATE }}'
     SAILTHRU_ABANDONED_CART_TEMPLATE: '{{ ECOMMERCE_WORKER_SAILTHRU_ABANDONED_CART_TEMPLATE }}'
     SAILTHRU_ABANDONED_CART_DELAY: '{{ ECOMMERCE_WORKER_SAILTHRU_ABANDONED_CART_DELAY }}'
-    SAILTHRU_KEY: '{{ ECOMMERCE_WORKER_SAILTHRU_KEY }}'
-    SAILTHRU_SECRET: '{{ ECOMMERCE_WORKER_SAILTHRU_SECRET }}'
+    SAILTHRU_KEY: '{{ ECOMMERCE_SAILTHRU_KEY }}'
```
feanil: Why did you make this change? These are two different services and should have different names for these settings.  Even if they do eventually take the same key as input.

feanil (changes requested):
I don't think you should change anything about the ecommerce worker.

feanil (approved):

coryleeio:
We talked and I agreed that we should not combine this setting.

PR #4953, opened 2019-01-17
Add users role with COMMON_USER_INFO to ad_hoc

jdmulloy:
I want to merge this to master first and then cherry-pick onto yonkers-ginko.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/ad_hoc_reporting/meta/main.yml
+++ b/playbooks/roles/ad_hoc_reporting/meta/main.yml
@@ -16,4 +16,7 @@ dependencies:
     user_info: "{{ AD_HOC_REPORTING_USER_INFO }}"
     tags:
       - users
-
+  - role: user
+    user_info: "{{ COMMON_USER_INFO }}"
+    tags:
+      - users


playbooks/roles/ad_hoc_reporting/meta/main.yml
```diff
@@ -16,4 +16,7 @@ dependencies:
     user_info: "{{ AD_HOC_REPORTING_USER_INFO }}"
     tags:
       - users
-
+  - role: user
+    user_info: "{{ COMMON_USER_INFO }}"
```
feanil, line 20: Is this variable specific to yonkers?  If so, do we need to set a default in case it's not defined when running this role in the edx context?

jdmulloy: We use it in edx-internal as well. It has a default defined here https://github.com/edx/configuration/blob/master/playbooks/roles/common_vars/defaults/main.yml#L31

feanil (approved):

PR #4956, opened 2019-01-18
disable splunkforwarder listener (OPS-3682)

nadeemshahzad:
Configuration Pull Request
---


playbooks/roles/splunkforwarder/tasks/main.yml
```diff
@@ -157,3 +157,14 @@
     - inputs
     - outputs
   notify: restart splunkforwarder
+
+- name: Disable splunkforwarder listener
+  blockinfile:
```
jdmulloy: I'm not sure if we should be using blockinfile, but I think the only alternative is to manage the whole file. I'd like a second opinion from @edx/devops.

feanil: +1 to managing the whole file.  Does this file already exist on the machines and have info in it?

jdmulloy: I looked at a machine in stage yesterday. As far as I can tell it's not already managed by Ansible. It has content that I think is generated by the Splunk installer. A conf.d type setup to add an extra file would be nice, but I don't think Splunk has that.

nadeemshahzad: We can update and deploy `server.conf` with our settings, Splunk automatically update and appends required settings in this file when it is restarted.
 https://answers.splunk.com/answers/223905/updating-serverconf-with-chef-how-can-i-generate-p.html

fredsmith (approved):

nadeemshahzad:
@fredsmith @jdmulloy I have made few changes in PR as boot script is changed in new version of splunk forwarder. Can you please review these changes ? thanks

PR #4971, opened 2019-01-25
add jenkins to docker group

estute:
I want to run the Docker cli on IT jenkins, but it needs root. Adding the jenkins user to the docker group should get around that. This creates a wrapper variable, jenkins_common_groups, which defaults to 'jenkins', but can be loaded with more than one group id, if you want. No change should be needed for any of the other Jenkii.


michaelyoungstrom (approved):
Seems fine to me. Extra 👍🏼for using the term "Jenkii"

jmbowman (approved):

PR #4979, opened 2019-01-29
Adds prospectus caching for different file types

jlajoie:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
@@ -20,15 +20,32 @@ server {
     try_files $uri $uri/index.html @backend;
   }
 
+  location /admin/ {
+    add_header 'Cache-Control' 'public, max-age=0, must-revalidate';
+  }
+
+  location /static/ {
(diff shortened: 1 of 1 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -20,15 +20,32 @@ server {
...
+  }
+
+  location @index {
+    add_header 'Cache-Control' 'public, max-age=0, must-revalidate';
+    try_files $uri/index.html @backend;
   }
 
   location @backend {
-      proxy_pass {{ PROSPECTUS_PROXY_PASS }};
+    add_header 'Cache-Control' '';
```
bderusha, line 47: Any reason why we need to set this to empty?

bderusha (approved):
Looks good.  Just one question about the clear header

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
jlajoie: So there isn't a remove_header, so instead this just removes the value we set Cache-Control to for the scenario where we're accessing Drupal after hitting the above location blocks where the Cache-Control is set.

PR #4986, opened 2019-01-31
Add new OAuth2 settings to credentials-related playbooks.

doctoryes:
Configuration Pull Request
---


playbooks/roles/oauth_client_setup/defaults/main.yml
```diff
@@ -36,8 +36,8 @@ oauth_client_setup_oauth2_clients:
     - {
         name: "{{ credentials_service_name | default('None') }}",
         url_root: "{{ CREDENTIALS_URL_ROOT | default('None') }}",
-        id: "{{ CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_KEY | default('None') }}",
-        secret: "{{ CREDENTIALS_SOCIAL_AUTH_EDX_OIDC_SECRET | default('None') }}",
+        id: "{{ CREDENTIALS_SOCIAL_AUTH_EDX_OAUTH2_KEY | default('None') }}",
+        secret: "{{ CREDENTIALS_SOCIAL_AUTH_EDX_OAUTH2_SECRET | default('None') }}",
```
doctoryes: @nasthagiri @pwnage101 Should this be the new client credentials OAuth2 key/secret instead?

doctoryes:
@nasthagiri @pwnage101 I added the new client credentials key/secret everywhere I think it should go here...

playbooks/roles/oauth_client_setup/defaults/main.yml
pwnage101: Under the hood, the ansible task for this role uses the create_oauth2_client management command which creates a DOP client, not a DOT app which this change would actually want.

I'm not yet totally sure what the solution is here, but it seems like we should modify the role to add a new ansible task for creating DOT apps using the create_dot_application management command.

doctoryes: That seems reasonable - I'll look into it.

nasthagiri (approved):

pwnage101 (approved):

PR #4987, opened 2019-02-01
Fixing asg lifecycle watcher by cherry picking region flag from master

coryleeio:


feanil:
@coryleeio this script is used by a bunch of jenkins jobs, does the DSL need to be updated in any way?  Since you're removing the `--profile` flag.

feanil (commented):
Comment above.

coryleeio:
In the most recent run:
https://admin.edx-flatu.org:8080/job/terminate-instances-that-have-been-verified-for-retirement-prod-olivex/444086/console

the job is passing region and not profile - so I believe the dsl is already expecting the new flag

feanil (approved):

PR #4988, opened 2019-02-04
Added file check_table_size.py

syedimranhassan:
Configuration Pull Request
---

### DIFF
--- a/util/check_table_size.py
+++ b/util/check_table_size.py
@@ -1,0 +1,114 @@
+import boto3
+import sys
+import pymysql
+import logging
+import click
+
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+source = boto3.client('rds')
+
+
+def check_table_growth(password, threshold):
+    """
+        Return:
+             Return list all tables that cross threshold limit
+              [
+                  {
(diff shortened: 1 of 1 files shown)


util/check_table_size.py
```diff
@@ -0,0 +1,114 @@
...
+def send_email(table_list):
+    html = ''
+    for table_data in table_list:
+        for key, value in table_data.items():
+            html += '<p>%s</p>' % key
+            html += '<li>' + str(value) + '</li>'
+
+        client = boto3.client('ses')
+        client.send_email(
+            Source='sandbox-notifications@edx.org',
```
jdmulloy, line 76: I don't think this is an appropriate from address, I'm guessing it was copied from another script.

util/check_table_size.py
```diff
@@ -0,0 +1,114 @@
...
+            Destination={
+                'ToAddresses': ["ihassan@edx.org"]
```
jdmulloy, line 78: This should probably be a mailing list and should be passed in via a variable and configurable via a command line option.

jdmulloy (changes requested):

PR #4995, opened 2019-02-05
Define COMMON_LMS_BASE_URL

nasthagiri:
Configuration Pull Request
---


pwnage101 (approved):

doctoryes (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+- common_vars
+  - Added new overridable variable `COMMON_LMS_BASE_URL`.
```
nedbat, line 2: Where do we document what these variables are for?

feanil: What's the reason for adding this instead of using the existing COMMON_OAUTH_BASE_URL?

doctoryes: `COMMON_OAUTH_BASE_URL` was *not* being overridden with a proper value in the internal/secure repos. So we were faced with the decision of defining it *or* defining a more specific, general-purpose URL used in other areas, which is the decision we made. 

@nasthagiri @pwnage101 Do you agree?

@nedbat I'm not sure that we do document all Ansible-based variables? In this case, it's fortunately self-explanatory given the var name.

pwnage101: At the time of creating this new var, we decided against changing the value of COMMON_OAUTH_BASE_URL because we did not know which other consumers there were, and wanted to improve the chance of PR approval.  @coryleeio suggested that truly understanding the impact of changing COMMON_OAUTH_BASE_URL would take some extra time.

I think we are still open to redefining COMMON_OAUTH_BASE_URL if it is safe to do so.

PR #5006, opened 2019-02-11
Remove default oauth2 public urls from the default yml config

pwnage101:
This prevents a default value for SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT
from being loaded into stage/prod.  Thanks to Matt Tuchfarber for
bringing this to our attention.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_django_service/defaults/main.yml
+++ b/playbooks/roles/edx_django_service/defaults/main.yml
@@ -154,7 +154,6 @@ edx_django_service_oidc_logout_url: '{{ COMMON_OAUTH_LOGOUT_URL }}'
 edx_django_service_oidc_issuer: '{{ COMMON_OIDC_ISSUER }}'
 
 edx_django_service_oauth2_url_root: '{{ COMMON_LMS_BASE_URL }}'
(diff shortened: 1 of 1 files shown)


tuchfarber:
If we're removing this here, do we need to put something in `docker/devstack_common_ansible_overrides.yml` so devstack doesn't lose this value?

pwnage101:
@tuchfarber in the last two IDAs that we've migrated off DOP (credentials, ecommerce), neither of their devstacks relied on the ansible defaults to generate the public URL root.  In credentials, it was hard-coded in the `credentials/settings/devstack.py` django settings file, and in ecommerce it will be (not at the time of this writing) hard-coded in the `provision-ecommerce.sh` script in edx/devstack:

https://github.com/edx/devstack/blob/master/provision-ecommerce.sh#L7

since the value should be read from the siteconfiguration model rather than django settings.

tuchfarber:
Works for me. If something doesn't work in devstack that's an easier fix iteration than prod anyhow.

tuchfarber (approved):
LGTM

doctoryes (approved):

PR #5018, opened 2019-02-14
Add Profile Microfrontend URL

nasthagiri:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -929,6 +929,9 @@ EDXAPP_PASSWORD_POLICY_COMPLIANCE_ROLLOUT_CONFIG:
 # Needed to link the LMS instructor dashboard to the writable gradebook micro-frontend
 EDXAPP_LMS_WRITABLE_GRADEBOOK_URL: null
 
+# Needed to link to the new profile micro-frontend.
+EDXAPP_PROFILE_MICROFRONTEND_URL: null
+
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -929,6 +929,11 @@ EDXAPP_PASSWORD_POLICY_COMPLIANCE_ROLLOUT_CONFIG:
 # Needed to link the LMS instructor dashboard to the writable gradebook micro-frontend
 EDXAPP_LMS_WRITABLE_GRADEBOOK_URL: null
 
+# TODO (DEPR-17)
+# Needed to redirect the old profile page to the new profile micro-frontend
+# during a transition period.
+EDXAPP_PROFILE_MICROFRONTEND_URL: null
```
douglashall, line 933: Not sure you need the DEPR comment here. We will still need this setting after the old profile page goes away right?

douglashall (approved):

playbooks/roles/edxapp/defaults/main.yml
nasthagiri: We will as long as Django code needs to link to the profile MFE.  But once we have a fully MFE world, only MFE configs will need the URL, right?  Then again, our backend APIs may continue to provide these URLs.  Unclear.  Ok, I'll remove the DEPR comment from here - we'll keep the one in edx-platform and can revisit at that time.

fredsmith (approved):

PR #5019, opened 2019-02-14
Add CORS static assets in nginx DEVOPS-8273

jdmulloy:
This adds the CORS header if the whitelist is not empty. I think the existing ```add_header Access-Control-Allow-Origin *;``` for fonts might be a vulnerability, but I'm not sure of the risk of removing it for those who haven't set the whitelist. 

There's also a variable called ENABLE_CORS_HEADERS that doesn't seem to be used anywhere.

Configuration Pull Request
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
 - Role: nginx
+  - Added CORS Access-Control-Allow-Origin for static assets.
+  - Replaced wildcard Access-Control-Allow-Origin header for fonts. Make sure you set EDXAPP_CORS_ORIGIN_WHITELIST to include all your domains.
+
+- Role: nginx
   - Modified robots.txt.j2 to accept the Allow rule.
   - Modified robots.txt.j2 to accept either a single string or a list of strings for agent, disallow, and allow.
 
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/static-files.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/static-files.j2
@@ -21,10 +21,8 @@
     # http://www.red-team-design.com/firefox-doesnt-allow-cross-domain-fonts-by-default
     location ~ "/static/(?P<collected>.*\.[0-9a-f]{12}\.(eot|otf|ttf|woff|woff2)$)" {
         add_header "Cache-Control" $cache_header_long_lived always;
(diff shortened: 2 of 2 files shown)


feanil (commented):
Spoke in person,  simplify the logic around the fonts whitelisting to match what we are doing for all static assets.

jdmulloy:
Relevant sections of nginx config on my sandbox.

```
 # The Origin request header indicates where a fetch originates from. It doesn't include any path information,
  # but only the server name (e.g. https://www.example.com).
  # See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin for details.
  #
  # Here we set the value that is included in the Access-Control-Allow-Origin response header. If the origin is one
  # of our known hosts--served via HTTP or HTTPS--we allow for CORS. Otherwise, we set the "null" value, disallowing CORS.
  map $http_origin $cors_origin {
  default "null";
      "~*^https?:\/\/edx\.org$" $http_origin;
      "~*^https?:\/\/www\.edx\.org$" $http_origin;
      "~*^https?:\/\/dev-edx-mktg-edit\.edx\.org$" $http_origin;
      "~*^https?:\/\/sandbox\.edx\.org$" $http_origin;
      "~*^https?:\/\/edx-mktg\.dd$" $http_origin;
      "~*^https?:\/\/edx-mktg\.local$" $http_origin;
      "~*^https?:\/\/https://ecommerce-jdmulloy-cors\.sandbox\.edx\.org$" $http_origin;
    }
```

```
  location ~ ^/static/(?P<file>.*) {
    root /edx/var/edxapp;
    try_files /staticfiles/$file /course_static/$file =404;

    add_header Access-Control-Allow-Origin $cors_origin;

    # return a 403 for static files that shouldn't be
    # in the staticfiles directory
    location ~ ^/static/(?:.*)(?:\.xml|\.json|README.TXT) {
        return 403;
    }

    # http://www.red-team-design.com/firefox-doesnt-allow-cross-domain-fonts-by-default
    location ~ "/static/(?P<collected>.*\.[0-9a-f]{12}\.(eot|otf|ttf|woff|woff2)$)" {
        add_header "Cache-Control" $cache_header_long_lived always;
        add_header Access-Control-Allow-Origin $cors_origin;
        try_files /staticfiles/$collected /course_static/$collected =404;
    }
```

feanil (approved):

PR #5022, opened 2019-02-15
Have discovery set up CORS via django, not nginx

mikix:
The core problem is that our nginx config intercepts OPTIONS http
requests. And we want django to handle them.

By not setting DISCOVERY_ALLOW_CORS_HEADERS, our nginx config no
longer sets up CORS by itself. And by passing the whitelist down
to the application, django-cors-headers can be configured per site.

Configuration Pull Request
---


feanil (commented):
A thing I realized in some other NGINX conversations is that we still use NGINX to serve static assets(JS) so if cors headers are needed for any of those static assets, you'll want to make sure to set the headers at NGINX and in django.  Is that going to be an issue?  If not, I think this looks fine.

mikix:
There shouldn't be any CORS requests for static assets from Discovery. Should be fine.

staubina:
@feanil This is blocking us on the stage Publisher environment now.  Can you help review and merge this one?

fredsmith (approved):

PR #5031, opened 2019-02-25
Script to check RDS slow query logs (OPS-3009)

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/util/check_rds_slow_query_logs.py
+++ b/util/check_rds_slow_query_logs.py
@@ -1,0 +1,103 @@
+import boto3
+import click
+
+
+def get_db_instances(db_engine):
+
+    """ 
+    Returns: 
+          List of provisioned RDS instances
+    """
+
+    if db_engine == "mysql":
+        instances = rds.describe_db_instances()['DBInstances']
+    elif db_engine == "aurora":
+        instances = rds.describe_db_clusters()['DBClusters']
+    return instances
+
+
+def get_db_parameters(db_engine, db_parameter_group, marker):
+    
+    """ 
+    Returns: 
+           The detailed parameter list for a particular DB parameter 
+           group Using marker as pagination token as at max it returns 
+           100 records
+    """
+
+    if db_engine == "mysql":
+        response = rds.describe_db_parameters(
+                       DBParameterGroupName=db_parameter_group, 
+                       Marker=marker)
+    elif db_engine == "aurora":
+        response = rds.describe_db_cluster_parameters(
+                       DBClusterParameterGroupName=db_parameter_group, 
+                       Marker=marker)
+    return response
+
+
+def check_slow_query_logs(db_engine, db_parameter_group):
+
+    slow_log_enabled = False
+
+    marker = ""
+   
+    while True:
+
+        if marker is None:
+            break
+
+        response = get_db_parameters(db_engine, db_parameter_group, marker)
+        marker = response.get('Marker')
+        parameters = response.get('Parameters')
+
+        for param in parameters:
+
+            if 'slow_query_log' in param['ParameterName']:
+
+                if 'ParameterValue' in param and param['ParameterValue'] == '1':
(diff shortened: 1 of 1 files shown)


util/check_rds_slow_query_logs.py
```diff
@@ -0,0 +1,81 @@
+import boto3
+import click
+
+
+def get_client():
```
feanil: You only ever do this once, I don't think this needs to be a separate function.  The indirection doesn't provide any extra value.

util/check_rds_slow_query_logs.py
```diff
@@ -0,0 +1,81 @@
...
+        elif db_engine == "aurora":
+            db_identifier = dbhost['DBClusterIdentifier']
+            db_parameter_group = dbhost['DBClusterParameterGroup']
+
+        slow_query_logs_enabled = check_slow_query_logs(db_engine,db_parameter_group)
+
+        if slow_query_logs_enabled:
+            print("Slow Query logs are enabled for RDS: {0}").format(db_identifier)
+        else:
+            print("Slow Query logs are disabled for RDS: {0}").format(db_identifier)
```
feanil: Do we want the script to fail or to notify when we find engines that don't have slow query logs enabled?

util/check_rds_slow_query_logs.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return slow_log_enabled
+
+@click.command()
+@click.option('--db_engine', help='RDS engine: mysql or aurora', required=True)
+def cli(db_engine):
+
+    dbhosts = get_db_instances(db_engine)
+
+    for dbhost in dbhosts:
```
feanil, line 76: Should the script take a list of known hosts that we don't need slow query logging enabled on?

feanil (commented):
generally looks good.  Can you run pep8 on the script, I noticed spacing issues throughout.

util/check_rds_slow_query_logs.py
nadeemshahzad: Yes, we want to use this script to notify if  slow query logs are disabled for RDS Instances.

feanil: Is the plan to run this on jenkins periodically so we are always checking this?  In that case, we may want to just print a list where slow query logging is disabled rather than interleaving the output of where it's fine and where it's not fine.

util/check_rds_slow_query_logs.py
feanil: @nadeemshahzad checking back in on this question, we don't need to solve it now I guess but it's something to keep in mind.

util/check_rds_slow_query_logs.py
nadeemshahzad: @feanil Yes, It is plan to run this periodically and I have already addressed the comment to print only rds instances for which slow query logging is disabled.

util/check_rds_slow_query_logs.py
```diff
@@ -0,0 +1,103 @@
...
+
+            db_parameter_group = dbhost['DBClusterParameterGroup']
+
+        slow_query_logs_enabled = check_slow_query_logs(db_engine, db_parameter_group)
+
+        if not slow_query_logs_enabled:
+            exit_status = 1
+            slow_query_logs_disabled_rds.append(db_identifier)
+
+    print("Slow query logs are disabled for RDS Instances\n{0}").format(slow_query_logs_disabled_rds)
```
nadeemshahzad: Print list of rds instances for which slow query logging is disabled.

util/check_rds_slow_query_logs.py
nadeemshahzad: @feanil I also addressed this comment to whitelist the rds instances, seems like outdated is not override with my recent changes.

coryleeio (approved):

PR #5034, opened 2019-02-27
Use strings in proctoring config default values

coryleeio:


coryleeio:
Differences in the RENDERED config:

```
docker exec -it edx.devstack.lms cat /edx/app/edxapp/lms.auth.json | grep PROCTORING -C4

Current default config used in devstack:
    "PROCTORING_BACKENDS": {
        "null": {}, 
        "DEFAULT": "null"
    }, 

docker exec -it edx.devstack.lms cat /edx/etc/lms.yml | grep PROCTORING -C4
PROCTORING_BACKENDS:
    null: {}
    DEFAULT: 'null'
```


based on the json output I think it is safe to change the defaults to this, considering that it is overwritten in stage and prod.  So this should only affect devstack and sandboxes, and string values are the currently rendered values.

feanil (approved):

coryleeio:
modified:   common/lib/xmodule/xmodule/course_module.py
get_available_providers()

settings.PROCTORING_BACKENDS


on master:


found settings {u'DEFAULT': u'null', u'null': {}}
found providers [u'null']






on my branch:


found settings {'DEFAULT': 'null', None: {}}
found providers [None]

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -748,8 +748,8 @@ EDXAPP_CREDIT_PROVIDER_SECRET_KEYS: {}
 # Proctoring configuration (redirct URLs and keys shared between systems)
 EDXAPP_PROCTORING_SETTINGS: {}
 EDXAPP_PROCTORING_BACKENDS:
-  DEFAULT: 'null'
-  null: {}
+  DEFAULT: "null"
+  "null": {}
```
feanil, line 754: Add a comment for why this needs to be quoted.

PR #5037, opened 2019-03-05
WIP: Slow query logs export to cloudwatch

syedimranhassan:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/export_slow_logs/export_slow_query_logs.py
+++ b/util/jenkins/export_slow_logs/export_slow_query_logs.py
@@ -1,0 +1,142 @@
+import boto3
+from botocore.exceptions import ClientError
+import sys
+import backoff
+import pymysql
+import time
+import uuid
+
+
+MAX_TRIES = 5
+
+
+class CWBotoWrapper:
+    def __init__(self, **kwargs):
+        self.client = boto3.client("logs", **kwargs)
+
+    @backoff.on_exception(backoff.expo, ClientError, max_tries=MAX_TRIES)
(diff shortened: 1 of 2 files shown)


util/jenkins/export_slow_logs/export_slow_query_logs.py
```diff
@@ -0,0 +1,142 @@
...
+                logStreamName="test",
+                logEvents=cw_logs,
+                sequenceToken=sequencetoken
+            )
+        sequencetoken = response["nextSequenceToken"]
+
+
+@click.command()
+@click.option('--username', required=True, help='Username to speak with RDS')
+@click.option('--password', required=True, help='Password to speak with RDS')
```
coryleeio: https://click.palletsprojects.com/en/5.x/options/#password-prompts

Check out the docs here and make this hidden

Read it from an environment variable on Jenkins, as it is bad practice to type a password as an inline arguement to a program as it can be exposed to other users on the system this way.

https://click.palletsprojects.com/en/5.x/options/#values-from-environment-variables

coryleeio (changes requested):

coryleeio (approved):

util/jenkins/export_slow_logs/export_slow_query_logs.py
syedimranhassan: I was picking them up form env variables but feanil commented on the PR that this should be picked up from credentials binding
https://github.com/edx/jenkins-job-dsl/pull/631#discussion_r265301078

PR #5044, opened 2019-03-06
Feanil/update create rds playbook

feanil:
Configuration Pull Request
---

### DIFF
--- a/playbooks/create_rds.yml
+++ b/playbooks/create_rds.yml
@@ -8,8 +8,9 @@
 # It can read from the same config as create_dbs_and_users.yml and needs this
 # part of that config
 # database_connection:
+#   login_host: "{{ database_name }}......rds.amazonaws.com" # This is pretty predictable for our accounts.
 #   login_user: "root"
-#   login_password: "super-secure-password" #
+#   login_password: "" # max 41 characters, Only printable ASCII characters besides '/', '@', '\"', ' ' may be used.
 # database_name: your-database-name
 # database_size: number of gigabytes (integer)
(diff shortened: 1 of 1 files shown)


playbooks/create_rds.yml
```diff
@@ -8,8 +8,9 @@
 # It can read from the same config as create_dbs_and_users.yml and needs this
 # part of that config
 # database_connection:
+#   login_host: "{{ database_name }}......rds.amazonaws.com" # This is pretty predictable for our accounts.
 #   login_user: "root"
-#   login_password: "super-secure-password" #
+#   login_password: "" # max 41 characters, Only printable ASCII characters besides '/', '@', '\"', ' ' may be used.
```
jdmulloy, line 13: Can we make this just alphanumeric so that selection by double click works? I'm assuming this is what you're getting at here. Maybe we should have a minimum too?

feanil: @jdmulloy this is the requirement from AWS that I just copied here, not my personal recommendation.  I can change this to just be 40 alphanumeric characters if you'd like?  But this was an attempt to capture the current actual requirements for this field.

feanil: Actually on second thought for security I don't really want to limit it to alphanumeric or make that recommendation in a comment here.

jdmulloy: @feanil that does make sense, although selecting the password by double clicking is nice, and it doesn't work with most special characters. I think specifying a minimum of say 25 characters would be good here either way.

 I'll leave it up to you.

jdmulloy (approved):

PR #5045, opened 2019-03-06
Add the configuration values for OAuth2 on registrar service

schenedx:
Configuration Pull Request
---


schenedx:
@edx/masters-neem @edx/devops Please review this PR.
This is to introduce the oauth users into registrar service so API communication can talk to and from this service. See https://openedx.atlassian.net/browse/EDUCATOR-4147

jdmulloy (approved):

PR #5050, opened 2019-03-13
Add the oauth2_public_url_root value to all django IDA role

schenedx:
Configuration Pull Request
---


schenedx:
I verified that this is the missing configuration value we need for registrar. I tested on my local machine with the value of `SOCIAL_AUTH_EDX_OAUTH2_PUBLIC_URL_ROOT` set to `replace-me`, and it had the same problem as we see in stage and prod.

feanil (approved):

PR #5057, opened 2019-03-19
add REGISTRAR_SEGMENT_KEY

jansenk:
Added `REGISTRAR_SEGMENT_KEY` for segment.io event tracking.

[EDUCATOR-4131](https://openedx.atlassian.net/browse/EDUCATOR-4131)

Related:
https://github.com/edx/registrar/pull/39

---

  - [✖️] A DevOps team member has approved the PR if it is code shared across multiple services and you don't own all of the services.
  - [ ☑️ ] Are you adding any new default values that need to be overridden when this change goes live? If so:
    - [ ☑️ ] Update the appropriate internal repo (be sure to update for all our environments)
    - [ ☑️] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [ ☑️ ] Add an entry to the CHANGELOG.
  - [ ✖️ ] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).
  - [ ✖️ ] Think about how this change will affect Open edX operators.  Have you updated the wiki page for the next Open edX release?

### DIFF
--- a/playbooks/roles/registrar/defaults/main.yml
+++ b/playbooks/roles/registrar/defaults/main.yml
@@ -86,6 +86,9 @@ registrar_service_config_overrides:
   LANGUAGE_COOKIE_NAME: '{{ REGISTRAR_LANGUAGE_COOKIE_NAME }}'
   CSRF_COOKIE_SECURE: "{{ REGISTRAR_CSRF_COOKIE_SECURE }}"
 
+# API key for segment.io
+REGISTRAR_SEGMENT_KEY: !!null
+
 # See edx_django_service_automated_users for an example of what this should be
 REGISTRAR_AUTOMATED_USERS: {}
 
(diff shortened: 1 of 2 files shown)


playbooks/roles/registrar/defaults/main.yml
```diff
@@ -86,6 +86,9 @@ registrar_service_config_overrides:
   LANGUAGE_COOKIE_NAME: '{{ REGISTRAR_LANGUAGE_COOKIE_NAME }}'
   CSRF_COOKIE_SECURE: "{{ REGISTRAR_CSRF_COOKIE_SECURE }}"
 
+# API key for segment.io
+REGISTRAR_SEGMENT_KEY: 'segment-key'
```
feanil: Shouldn't the default be `null`?

jansenk: should it? i don't know, i was just following the oauth key stuff above.

feanil: In the registrar code, what decides whether or not to put the segment script onto a page?  If there is not a separate feature flag for enabling/disabling segment then probably the existence of a non-null key, results in the code path trying to execute.  In which case the string will probably cause breakages.  This is just a guess and you should verify how this key is used in your code to figure out what a reasonable default is.

iloveagent57 (approved):

schenedx (approved):
👍

PR #5063, opened 2019-03-21
Remove aws.py

coryleeio:


bderusha (approved):

bderusha:
Are there any timing concerns/dependencies we should note about merging this?

coryleeio:
Nope - we have been running production.py in production for many months now.  Anything using aws.py was potentially still using the json config on disk ( which should have no effect ), this should be safe to remove.

PR #5084, opened 2019-04-16
Estute/te 2881

estute:
Contained in the PR:
* a role for configuring sandboxed python virtualenvs for using/testing codejail for multiple python versions
* a play and packer script for creating a jenkins worker with this codejail environment

I also stopped installing `python-software-properties` when using xenial as a base OS version. It is flaky and sometimes not available. Builds were often, but not always, failing with:
```
00:00:55.448     amazon-ebs: Package python-software-properties is not available, but is referred to by another package.
00:00:55.449     amazon-ebs: This may mean that the package is missing, has been obsoleted, or
00:00:55.450     amazon-ebs: is only available from another source
00:00:55.451     amazon-ebs: However the following packages replace it:
00:00:55.453     amazon-ebs: software-properties-common
```

Successful build of this role here: https://build.testeng.edx.org/job/build-packer-ami/14003/

### DIFF
--- a/playbooks/jenkins_worker_codejail.yml
+++ b/playbooks/jenkins_worker_codejail.yml
@@ -1,0 +1,18 @@
+# Configure a Jenkins worker instance specifically for running tests for
+# CodeJail, which requires specific set up with regards to python execution
+
+- name: Configure instance(s)
+  hosts: jenkins_worker
+  become: True
+  gather_facts: True
+  vars:
+    serial_count: 1
+    codejail_worker: True
+    COMMON_SECURITY_UPDATES: yes
+    SECURITY_UPGRADE_ON_ANSIBLE: true
+  serial: "{{ serial_count }}"
+  roles:
+    - aws
+    - docker-tools
+    - jenkins_worker
+    - codejail
--- a/playbooks/roles/codejail/defaults/main.yml
+++ b/playbooks/roles/codejail/defaults/main.yml
@@ -1,0 +1,12 @@
+---
+codejail_debian_packages:
+  - apparmor-utils
+codejail_python_versions:
+  - python2-7
+  - python3-5
+  - python3-6
+codejail_sandbox_user: 'sandbox'
+codejail_sandbox_group: 'sandbox'
+codejail_sandbox_name_base: 'codejail_sandbox'
+codejail_sandbox_env: '/home/{{ codejail_sandbox_user }}/{{ codejail_sandbox_name_base }}'
+codejail_sandbox_caller: 'ubuntu'
--- a/playbooks/roles/jenkins_worker/defaults/main.yml
+++ b/playbooks/roles/jenkins_worker/defaults/main.yml
@@ -21,3 +21,11 @@ packer_url: "https://releases.hashicorp.com/packer/0.8.6/packer_0.8.6_linux_amd6
 
 nodejs_version: "8"
 ansible_distribution_release: "xenial"
+
+jenkins_worker_python_versions:
+  - python2.7
+  - python2.7-dev
+  - python3.5
+  - python3.5-dev
+  - python3.6
+  - python3.6-dev
--- a/playbooks/roles/jenkins_worker/tasks/python.yml
+++ b/playbooks/roles/jenkins_worker/tasks/python.yml
@@ -1,13 +1,24 @@
 ---
+# Versions of Python newer than 3.5 are not available in the default
+# package index for Ubuntu 16.04. Add the deadsnakes PPA for anything
+# newer
+- name: add deadsnakes PPA for newer Python versions
+  apt_repository:
+    repo: "ppa:fkrull/deadsnakes"
+    update_cache: yes
+  when: ansible_distribution_release == 'xenial'
+
+# Install newer versions of python for testing, but do not set them
+# as the default version
+- name: Install python versions
+  apt:
+    name: '{{ item }}'
+    state: present
+    update_cache: yes
+  with_items: '{{ jenkins_worker_python_versions }}'
 
 # Requests library is required for the github status script.
 - name: Install requests Python library
-  pip: name=requests state=present
-
-# Install Python 3.5. For now, do not set it as the default python
-# version
-- name: Install python 3.5
-  apt:
-    name: 'python3.5-dev'
+  pip:
+    name: requests
     state: present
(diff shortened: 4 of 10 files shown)


playbooks/roles/jenkins_worker/defaults/main.yml
```diff
@@ -21,3 +21,9 @@ packer_url: "https://releases.hashicorp.com/packer/0.8.6/packer_0.8.6_linux_amd6
 
 nodejs_version: "8"
 ansible_distribution_release: "xenial"
+
+python_versions:
+  - python3.5
+  - python3.5-dev
+  - python3.6
+  - python3.6-dev
```
jmbowman, line 31: Nice, does this let us update to 3.6 in https://github.com/edx/jenkins-job-dsl/blob/master/platform/jobs/pipelines/edxPlatformPipelinePr.groovy ?

playbooks/roles/codejail/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+codejail_debian_packages:
+  - apparmor-utils
+python_versions:
+  - 2.7
```
feanil: This is just the version numbers whereas above they are prefixed with `python` why do they need to be in both places?

playbooks/roles/codejail/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
...
+  - 3.6
+sandbox_user: 'sandbox'
```
feanil: The vars for a role should all be prefixed with the role name to prevent naming collisions.

playbooks/jenkins_worker_codejail.yml
```diff
@@ -0,0 +1,23 @@
...
+- name: Configure instance(s)
+  hosts: jenkins_worker
+  become: True
+  gather_facts: True
+  vars:
+    serial_count: 1
+    codejail_worker: True
+    COMMON_SECURITY_UPDATES: yes
+    SECURITY_UPGRADE_ON_ANSIBLE: true
+    python_versions:
```
feanil: Does this list need to exist here?

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -3,11 +3,34 @@
 # Requests library is required for the github status script.
 - name: Install requests Python library
   pip: name=requests state=present
+# tox is needed for running codejail tests
+- name: Install requests Python library
```
jmbowman: Copy-pasta on the name?  Also, are we ok not specifying a version or also depending on tox-battery (to recreate the tox virtualenvs when requirements files change)?

playbooks/roles/codejail/defaults/main.yml
nedbat: Why is 3.5 in the list above, but not here?  I wouldn't think we need 3.5? (Though I'm not opposed to broader testing)

util/install/ansible-bootstrap.sh
```diff
@@ -108,7 +108,7 @@ fi
 
 # Required for add-apt-repository
 apt-get install -y software-properties-common
-if [[ "${SHORT_DIST}" != bionic ]] ;then
+if [[ "${SHORT_DIST}" != bionic ]] && [[ "${SHORT_DIST}" != 'xenial' ]];then
```
nedbat: Weird that one is quoted and one is not.

playbooks/roles/jenkins_worker/defaults/main.yml
estute: yes, it will be available on all jenkins workers

playbooks/roles/jenkins_worker/tasks/python.yml
estute: Correct. I have fixed it and pinned it at a specific (most recent) version. It is the one library needed to bootstrap the jobs. I suppose we could do something like we have done in travis jobs and install tox each time. @jmbowman  What do you think?

util/install/ansible-bootstrap.sh
estute: fixed.

playbooks/roles/codejail/defaults/main.yml
estute: Yeah, good point. Fixed.

playbooks/roles/codejail/defaults/main.yml
estute: I have refactored the roles to call codejail as a `meta` role from jenkins_worker. It is possible that someone else will use the codejail role, so I made it's defaults simple, i.e. 2.7 and 3.6. But since we also need to define these versions in then jenkins_worker role, I think it makes sense to have 3.5 installed as well, as it *may* be useful in testing.

playbooks/jenkins_worker_codejail.yml
estute: It doesn't anymore. Deleted.

playbooks/roles/jenkins_worker/tasks/python.yml
jmbowman: Yeah, I'd kind of rather let the job running tox specify which version to install via a requirements file in the repo.

playbooks/roles/codejail/defaults/main.yml
feanil: Let's have defaults we want and if people need to override them, they can do so.  It's a lot less complex to have the defaults map to what we're doing if they aren't going to be harmful to others.

estute:
I was running into issues when actually executing code via the codejail, which were solved by the following:
* Apparmor requires profile filenames be paths to executables, with `/` substituted by `.`. My paths included the string `python2.7`, which was confusing apparmor. This was fixed by changing the codejail names to be like `python2-7`.
* When invoked with a specific python version, virtualenv only copies that version, and symlinks it to `python` in the virtualenv. However, since the apparmor profile was tied to the path that pointed at `python` and apparmor does not follow symlinks, the codejail was not running safe_exec. This was fixed by invoking virtualenv with `--always-copy` which copies the desired python version into the environment twice, once as `python` and again as `pythonx.x`, without any symlinks.

jmbowman (approved):

nedbat (approved):

PR #5091, opened 2019-04-23
Update ubcpi XBlock to 0.6.5

nedbat:
Configuration Pull Request
---


nedbat:
@kitsook Sorry this took so long.  Future version updates can happen here.

marcotuts:
From looking at the v0.6.5 changelog, it looks like the following items are addressed:

- [Fix] When there are multiple ubcpi components in the same unit, for any questions other than the first one, clicking on the answer text will (1) cause the page to jump to the top; and (2) changed the answer of the first question instead of the intended one.
- [Enhancement] Event logs should show the rationales that student see after their first submission
- [Fix] if I go to the second problem, and then click back to the first problem, it looks like my answers have been deleted. If I hard-refresh the page, I get back to the answer state.

Looks good to me from a product standpoint

marcotuts (approved):

PR #5100, opened 2019-04-26
clean up groovy scripts

estute:
There was some code to do this in the past, but it was removed. Perhaps it was a branch? Either way, this codifies the removal of the init scripts, rather than manually removing them, which can be error prone. This is not an option for local-dev, because it does not use systemd.


michaelyoungstrom:
What's the benefit of this? Don't we want the scripts on the machine to revert any manual settings that might have been made when resetting the application?

estute:
Resetting the application does not make any changes. I think we should move more towards the direction of only changing via an invocation of ansible

michaelyoungstrom (approved):

PR #5143, opened 2019-05-22
Jdmulloy/hermes role template

jdmulloy:
This was pretty busted. Does anyone actually use it?

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/ansible-role-django-ida/tasks/main.yml
+++ b/playbooks/roles/ansible-role-django-ida/tasks/main.yml
@@ -48,7 +48,7 @@
 
 - name: Update role templates
   template:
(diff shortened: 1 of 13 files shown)


adzuci (approved):
I think https://github.com/edx/configuration/pull/5143/commits/956d1fe53ccb56176c047cfd26a4296cb37f2ebd#diff-93d74973b8872c9da94f1baa0a324ef0R69 is needed for Hermes and am a little confused as to why the template path was wrong previously, but the rest fo the changes look like they are purely a rename to avoid the rome_name keyword which seem good to go.  Didn't test this, but feel like it's still ok to merge.

PR #5150, opened 2019-05-23
Specify git key_file instead of using wrapper

jdmulloy:
Configuration Pull Request
---


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -465,6 +465,7 @@ EDXAPP_USE_GIT_IDENTITY: false
 # Paste the contents of the git identity
 # into this var
 EDXAPP_GIT_IDENTITY: !!null
+edxapp_git_key: "{% if EDXAPP_USE_GIT_IDENTITY %}{{ edxapp_git_identity }}{% endif %}"
```
coryleeio: Move to ansible play since we are about to move these into application repo and would prefer there be no logic in it when we do that.

coryleeio (changes requested):

coryleeio (approved):

PR #5156, opened 2019-05-29
Alert on primary keys

syed-awais-ali:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/primary_keys/check_primary_keys.py
+++ b/util/jenkins/primary_keys/check_primary_keys.py
@@ -214,6 +214,9 @@ def get_metrics_and_calcuate_diff(namespace, metric_name, dimension, value, curr
     cosnumed_keys_percentage = 100 - current_consumption
     days_remaining_before_exhaustion = cosnumed_keys_percentage/(current_consumption -
                                                                  last_max_reading)
+    if days_remaining_before_exhaustion < 365:
+        sys.exit(1)
+
 
 
 @click.command()
@@ -233,8 +236,4 @@ def controller(username, password, environment, deployment):
     # get list of all the RDSes across all the regions and deployments
     rds_list = get_rds_from_all_regions()
     table_list = check_primary_keys(rds_list, username, password, environment, deployment)
-    cloudwatch = CwBotoWrapper()
-    response = cloudwatch.list_metrics(Namespace=NAMESPACE,
-                                       MetricName=METRIC_NAME)
-
-    exit(0)
+    sys.exit(0)
--- a/util/jenkins/primary_keys/check_primary_keys.py
+++ b/util/jenkins/primary_keys/check_primary_keys.py
@@ -1,0 +1,240 @@
+import boto3
+from botocore.exceptions import ClientError
+import sys
+import backoff
+import pymysql
+import click
+from datetime import datetime, timedelta
+
+MAX_TRIES = 5
+PERIOD = 360
+UNIT = 'percent'
+
+class EC2BotoWrapper:
+    def __init__(self):
+        self.client = boto3.client("ec2")
+
+    @backoff.on_exception(backoff.expo,
+                          ClientError,
+                          max_tries=MAX_TRIES)
+    def describe_regions(self):
+        return self.client.describe_regions()
+
+
+class CwBotoWrapper():
+    def __init__(self):
+        self.client = boto3.client('cloudwatch')
+
+    @backoff.on_exception(backoff.expo,
+                          ClientError,
+                          max_tries=MAX_TRIES)
+    def list_metrics(self, *args, **kwargs):
+        return self.client.list_metrics(*args, **kwargs)
+
+    @backoff.on_exception(backoff.expo,
+                          ClientError,
+                          max_tries=MAX_TRIES)
+    def put_metric_data(self, *args, **kwargs):
+        return self.client.put_metric_data(*args, **kwargs)
+
+    @backoff.on_exception(backoff.expo,
+                          ClientError,
+                          max_tries=MAX_TRIES)
+    def get_metric_stats(self, *args, **kwargs):
+        return self.client.get_metric_statistics(*args, **kwargs)
+
+
+class RDSBotoWrapper:
+    def __init__(self, **kwargs):
+        self.client = boto3.client("rds", **kwargs)
+
+    @backoff.on_exception(backoff.expo,
+                          ClientError,
+                          max_tries=MAX_TRIES)
+    def describe_db_instances(self):
+        return self.client.describe_db_instances()
+
+
+def get_rds_from_all_regions():
+    """
+    Gets a list of RDS instances across all the regions and deployments in AWS
+
+    :returns:
+    list of all RDS instances across all the regions
+        [
+            {
+                'name': name of RDS,
+                'Endpoint': Endpoint of RDS
+                'Port': Port of RDS
+            }
+        ]
+    name (string)
+    Endpoint (string)
+    Port (string)
+    """
+    client_region = EC2BotoWrapper()
+    rds_list = []
+    try:
+        regions_list = client_region.describe_regions()
+    except ClientError as e:
+        print("Unable to connect to AWS with error :{}".format(e))
+        sys.exit(1)
+    for region in regions_list["Regions"]:
+        client = RDSBotoWrapper(region_name=region["RegionName"])
+        response = client.describe_db_instances()
+        for instance in response.get('DBInstances'):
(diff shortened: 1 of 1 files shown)


util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,211 @@
...
+                metric_data.append({
+                    'MetricName': metric_name,
+                    'Dimensions': [{
+                        "Name": item["name"],
+                        "Value": tables[0]
+                    }],
+                    'Value': tables[6] # percentnage of the usage of primary keys
+                })
+
+                if len(metric_data) > 0:
```
jdmulloy:
```suggestion
        if len(metric_data) > 0:
```
This should be run once after the for item loop.

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,211 @@
...
+                    cloudwatch.put_metric_data(Namespace=namespace, MetricData=metric_data)
```
jdmulloy:
```suggestion
            cloudwatch.put_metric_data(Namespace=namespace, MetricData=metric_data)
```

jdmulloy (changes requested):
I've made a few suggestions. Title says WIP. Is this still in progress or is it ready for final review? Can you provide some sample output?

syed-awais-ali:
Some screenshots attached for your convenience. We want to upload the data to cloudwatch when the job runs to retain the history and to take the diff of the last run and the current run in order to predict when the keys will get exhausted.

Screen Shot 2019-06-13 at 6 25 28 PM
Screen Shot 2019-06-13 at 6 25 15 PM
Screen Shot 2019-06-13 at 6 25 01 PM

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+                  {
+                    "name": "string",
+                    "db": "string",
+                    "table": "string",
+                    "size": "string",
+                  }
+              ]
+    """
+    cloudwatch = CwBotoWrapper()
+    metric_name = 'check_primary_keys'
```
jdmulloy:
```suggestion
    metric_name = 'used_key_space'
```

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+            connection.close()
+
+            for tables in rds_result:
+                metric_data.append({
+                    'MetricName': metric_name,
+                    'Dimensions': [{
+                        "Name": item["name"],
+                        "Value": tables[1]
+                    }],
+                    'Value': tables[6]  # percentage of the usage of primary keys
```
jdmulloy:
```suggestion
                    'Value': tables[6],  # percentage of the usage of primary keys
                    'Unit': 'Percent',
```

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+            ROUND(auto_increment/max_int*100,2) 
+            AS "USED %" FROM 
+            (SELECT table_schema, table_name, column_name, column_type, auto_increment, 
+            pow(2, case data_type when 'tinyint' then 7 when 'smallint' then 15 when 'mediumint' 
+            then 23 when 'int' then 31 when 'bigint' then 63 end+(column_type like '% unsigned'))-1 as 
+            max_int FROM 
+            information_schema.tables t JOIN information_schema.columns c USING 
+            (table_schema,table_name) WHERE t.table_schema not in 
+            ('mysql','information_schema','performance_schema') and 
+            t.table_type = 'base table' and c.extra = 'auto_increment'
```
jdmulloy:
```suggestion
            t.table_type = 'base table' and c.extra like '%auto_increment%' 
```
Maybe use a fuzzy match here in case there are other options in extra.

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+            rds_host_endpoint = item["Endpoint"]
+            rds_port = item["Port"]
+            connection = pymysql.connect(host=rds_host_endpoint,
+                                         port=rds_port,
+                                         user=username,
+                                         password=password)
+            # prepare a cursor object using cursor() method
+            cursor = connection.cursor()
+            # execute SQL query using execute() method.
+            # this query will return the tables with usage in percentage, result is limited to 10
```
jdmulloy, line 127: Why is this limited to 10?

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+            (table_schema,table_name) WHERE t.table_schema not in 
+            ('mysql','information_schema','performance_schema') and 
+            t.table_type = 'base table' and c.extra = 'auto_increment' 
+            AND t.auto_increment IS NOT NULL order by auto_increment desc limit 10) TMP;
+            """)
+            rds_result = cursor.fetchall()
+            cursor.close()
+            connection.close()
+
+            for tables in rds_result:
```
jdmulloy:
```suggestion
            for table in rds_result:
```
I believe each item is a single row representing a single table.

util/jenkins/primary_keys/check_primary_keys.py
jdmulloy:
```suggestion
                    'Value': table[6]  # percentage of the usage of primary keys
```

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+                metric_data.append({
+                    'MetricName': metric_name,
+                    'Dimensions': [{
+                        "Name": item["name"],
+                        "Value": tables[1]
```
jdmulloy:
```suggestion
                        "Value": table[1]
```

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,212 @@
...
+            rds_port = item["Port"]
+            connection = pymysql.connect(host=rds_host_endpoint,
+                                         port=rds_port,
+                                         user=username,
+                                         password=password)
+            # prepare a cursor object using cursor() method
+            cursor = connection.cursor()
+            # execute SQL query using execute() method.
+            # this query will return the tables with usage in percentage, result is limited to 10
+            cursor.execute("""
```
jdmulloy, line 128: I would suggest using this query instead
```
SELECT
    table_schema,
    table_name, 
    column_name,
    column_type,
    auto_increment,
    max_int,
    ROUND(auto_increment/max_int*100,2) AS used_pct
FROM
    (
     SELECT
        table_schema,
        table_name,
        column_name,
        column_type,
        auto_increment,
        pow
            (2,
                case data_type
                when 'tinyint' then 7
                when 'smallint' then 15
                when 'mediumint' then 23
                when 'int' then 31
                when 'bigint' then 63
                end
            +(column_type like '% unsigned'))-1
        as max_int
    FROM
        information_schema.tables t
        JOIN information_schema.columns c 
            USING (table_schema,table_name)
            WHERE t.table_schema not in ('mysql','information_schema','performance_schema')
            AND t.table_type = 'base table' 
            AND c.extra LIKE '%auto_increment%'
            AND t.auto_increment IS NOT NULL
    )
TMP ORDER BY used_pct desc
LIMIT 10;
```

I would have like to use a ```WHERE used_pct > 75``` clause, but that's not possible on calculated columns. Perhaps we should remove the limit and check in python if the value is over a certain threshold. Although that could wait for a second PR after we merge this as this query will get us enough things to put in cloudwatch.

jdmulloy (changes requested):

util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -0,0 +1,240 @@
...
+        EndTime=datetime.now().strftime("%Y, %m, %d"),
+        Period=360,
+        Statistics=[
+            'Maximum',
+        ],
+        Unit='Count'
+    )
+    last_max_reading = res["Datapoints"][0]["Maximum"]
+    cosnumed_keys_percentage = 100 - current_consumption
+    days_remaining_before_exhaustion = cosnumed_keys_percentage/(current_consumption -
```
jdmulloy, line 215: It doesn't look like this get used anywhere, but it seems useful. Is there coming functionally to use this?

jdmulloy (approved):

PR #5159, opened 2019-05-29
base_ami_id overrides ami_id if its set

fredsmith:
the base_ami_id variable is ignored if there is a running, active instance of the service that is being built.  sometimes, we need to override the ami_id.


macdiesel:
I think this is unnecessary as there is a flag you are to use in conjunction with this if you want to use the base_ami_ID even if there is an ASG running. Let me see if I can find it.

fredsmith:
turns out there's a separate variable, `base_ami_id_override` that controls this logic.

macdiesel:
Beat me to it, here is the PR for that work: https://github.com/edx/edx-gomatic/pull/101

PR #5162, opened 2019-05-30
Update node version for prospectus.

dianakhuang:
`common_node_version` is a few LTS versions out of date, and we would like to be more up to date. Previous attempts to update `common_node_version` broke on sandboxes for unclear reasons: https://github.com/edx/configuration/pull/4933
https://github.com/edx/edx-platform/pull/19567


inventhouse (approved):
Yes, please!

PR #5176, opened 2019-06-06
Run newrelic on jenkins ecs containers

michaelyoungstrom:
We run the edxapp playbook to create the docker image that gets run on ECS. I believe adding this override will allow the installation of newrelic on these containers and allow us to collect more data from them: https://github.com/edx/configuration/blob/af4c8d4e879c832089db4850e43d62e80e940095/playbooks/edxapp.yml#L29-L30


jmbowman (approved):

michaelyoungstrom:
This would have required a bit more to it because of the implications of the image being public and needing a newrelic key. Closing because we're actually going to be moving off ECS back to EC2 due to improved performance from new worker types.

PR #5181, opened 2019-06-07
Stop hermes from starting on build boxes

coryleeio:


playbooks/roles/hermes/files/pre_hermes_checks.sh
```diff
@@ -0,0 +1,13 @@
...
+# This script exists because when we build amis we take a snapshot and when we take this snapshot
+# the best practice is to reboot the instance since if you do not do this reboot the instance's
+# file system integrity cannot be guaranteed.
+# Since we monitor hermes, this causes errors that are not a problem to be logged when hermes fails to run correctly
+# on build boxes. 
+# This script is run before hermes is started, preventing it from booting during builds.
+
+
+# This is a hack to return 1 if build box, 0 otherwise
+! aws sts get-caller-identity | grep Arn | grep -i gocd
```
adzuci, line 13: Interesting way to check if it's a gocd box, curious if there is a local state (like a hostname) that could also be grepped or how this logic would play out if there was an AWS API issue, but I suspect that's preemptively solving a problem that we may not have.

adzuci (approved):
:shipit:

playbooks/roles/hermes/files/pre_hermes_checks.sh
coryleeio: I reckon in this case if it fails, it will return a 0, and thus run hermes, which will fail for build boxes, and possibly trigger the alert on the build box.

On a non build box it would return a non zero, get negated and run hermes also, so I think it should work in any case, worst case is we still get the alert on the build box

PR #5189, opened 2019-06-13
BB-184 Modify gunicorn template for more granular gunicorn control

xirdneh:
Configuration Pull Request
---

This PR updates the `workers` formula to improve readability. The formula used to read `(cpus - 1) * mul + mul` where `cpus` is the number of cpus and `mul` is a multiplier which value is set in the yaml variables. If we expand this formula we can simplify it to `cpus * mul`:
```
(cpus - 1) * mul + mul
(mul * cpus) - mul + mul
mul * cpus
```
This PR also adds three more gunicorn config variables:
[threads](http://docs.gunicorn.org/en/stable/settings.html#threads) : This will allow us to use multiple threads for each worker. Making it possible to run more workers and keeping gunicorn memory footprint low.
[worker_class](http://docs.gunicorn.org/en/stable/settings.html#worker-class) : If `threads` are used then it is possible to specify which type of thread class gunicorn will use. By default it will use `gthread` if any number of `threads` if specified.
[max_requests_jitter](http://docs.gunicorn.org/en/stable/settings.html#max-requests-jitter) : This will allow us to specify a max number of requests before gunicorn restart a worker. By doing this we help handle any memory leak happening in the application.

Reasoning:

We have experimented with a different amount of workers as well as max requests before gunicorn restart the workers. Using a number around 14 of workers have helped us better use all the resources in different cases. This is why we suggest adding gunicorn's `thread` and `worker_class` that way other users can run more workers while keeping gunicorn's memory footprint almost the same. Using 14 `sync` workers ( the default type of workers without threads) also works perfectly, hence using `threads` and `worker_class` is completely optional.
We've also added the `max_requests_jitter` configuration so users can specify how many requests will each worker handle before being restarted by gunicorn. This has allowed us to avoid memory leak or high memory consumption by one or multiple workers in different cases. We usually use `max_request` but in the latest gunicorn they've introduced `max_request_jitter` which will make gunicorn use a random number of max request between 0 and this value. This avoids having all threads restart at the same time and make the application sluggish.

**Note:** Once this is approved we have to create another PR for current hawthorn instances if we want to use this across the board. Also, we should use a `EDXAPP_WORKERS` value of 14 as default.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -7,3 +7,4 @@ Make sure that the following steps are done before merging
   - [ ] are you adding any new default values that need to be overridden when this goes live?
     - [ ] Open a ticket (DEVOPS) to make sure that they have been added to secure vars.
     - [ ] Add an entry to the CHANGELOG.
+  - [ ] Are you adding/updating any variables that need to be added/updated to a specific `configuration-secure` repo?
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -453,7 +453,16 @@ EDXAPP_VIRTUAL_UNIVERSITIES: []
 # EDXAPP_WORKERS:
 #   lms: <num workers>
 #   cms: <num workers>
+# EDXAPP_WORKER_THREADS
+#   lms: <num worker threads>
+#   cms: <num worker threads>
 EDXAPP_WORKERS: !!null
+EDXAPP_WORKER_THREADS: !!null
+EDXAPP_WORKER_CLASS: !!null
+# Gunicorn will restart each worker after x amount of
+# requests processed. The x value will be randomly selected between
+# 1 and this value (default 1500)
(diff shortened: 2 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @xirdneh! I've created [OSPR-3658](https://openedx.atlassian.net/browse/OSPR-3658) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

openedx-webhooks:
@xirdneh Even though your pull request wasn’t merged, please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/configuration&entry.867055334=https://github.com/edx/configuration/pull/5189&entry.1484655318=https://github.com/xirdneh&entry.752974735=2019-06-13+14:42&entry.1917517419=2019-06-13+14:43&entry.2133058324=No) so we can improve your experience in the future.

PR #5192, opened 2019-06-13
Installing gh-ost on tools-gp + standalone role/playbook for gh-ost

coryleeio:


playbooks/roles/gh-ost/meta/main.yml
```diff
@@ -0,0 +1,2 @@
+---
```
jdmulloy: I think you can just delete this file since it looks like it's not needed.

playbooks/roles/gh-ost/tasks/main.yml
```diff
@@ -0,0 +1,20 @@
+- name: Check if package is installed and version is correct
+  command:  dpkg -s gh-ost | grep Version | cut -d ':' -f 3 | grep {{ GHOST_VERSION }}
+  register: ghost_correct
+  ignore_errors: True
+
+- name: Remove package if exists
```
fredsmith, line 6: shouldn't this also only fire `when: ghost_correct.rc != 0`?

fredsmith (changes requested):

playbooks/roles/gh-ost/defaults/main.yml
```diff
@@ -0,0 +1,37 @@
...
+# license:    https://github.com/edx/configuration/blob/master/LICENSE.TXT
+#
+# Example play:
+#
+# Rather than being included in the play, this role
+# is included as a dependency by other roles in the meta/main.yml
+# file.  The including role should add the following
+# dependency definition.
+#
+# dependencies:
```
jdmulloy: I think the header form here down can be deleted. You could document the GHOST_VERSION variable if you want.

jdmulloy (dismissed):

playbooks/roles/gh-ost/tasks/main.yml
coryleeio: Updated

fredsmith (approved):
:ship: :it:

jdmulloy (approved):

PR #5193, opened 2019-06-13
Remove service_variant from registrar

jdmulloy:
The only other place I see this service_variant variable used is for edxapp, because it has both LMS and CMS. This seems a bit redundant for any other IDA so I think it makes sense to remove it to simplify things.

The one change I'm not entirely sure about is the ```EDX_REST_API_CLIENT_NAME``` setting. Is there anything that will have an issue if this changes? I couldn't find the old one anywhere.

```
-EDX_REST_API_CLIENT_NAME=edx.{{ w.service_variant }}.core.{{ w.queue }}
+EDX_REST_API_CLIENT_NAME={{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-registrar-worker-{{ w.queue }}
```

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/registrar/defaults/main.yml
+++ b/playbooks/roles/registrar/defaults/main.yml
@@ -146,7 +146,6 @@ REGISTRAR_NEWRELIC_WORKERS_APPNAME: registrar-workers
 worker_django_settings_module: "{{ REGISTRAR_DJANGO_SETTINGS_MODULE }}"
 REGISTRAR_CELERY_WORKERS:
   - queue: '{{ registrar_celery_default_queue }}'
-    service_variant: registrar
     concurrency: 1
     monitor: True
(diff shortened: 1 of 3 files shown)


kdmccormick (approved):
Removing `service_variant`  should be fine.

I have no knowledge on the safety of changing `EDX_REST_API_CLIENT_NAME`. The new value seems reasonable, but I don't know if changing Registrar worker's API user agent will have any unintended consequences. I also don't know what this value ends up being for normal Registrar.

iloveagent57 (approved):
Just seconding what Kyle said.  If the user agent is all that's affected, changing EDX_REST_API_CLIENT_NAME is probably fine.  This is the line we're talking about that changes the user agent: https://github.com/edx/edx-rest-api-client/blob/7ed2a833691d2fdf3a4fbc9189ee5c4443bd892b/edx_rest_api_client/client.py#L27

PR #5195, opened 2019-06-14
[BB-1289] Add support for specifying Mongo authentication database

xitij2000:
Configuration Pull Request
---

This PR allows configuring the [authentication database](https://docs.mongodb.com/manual/core/security-users/#user-authentication-database) for Mongo for edx-platform and cs_comments_service. 

This also requires the following PRs to merge to actually work: 
- https://github.com/edx/edx-platform/pull/20819
- https://github.com/edx/cs_comments_service/pull/275


openedx-webhooks:
Thanks for the pull request, @xitij2000! I've created [OSPR-3661](https://openedx.atlassian.net/browse/OSPR-3661) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -73,6 +73,7 @@ EDXAPP_MONGO_USER: 'edxapp'
 EDXAPP_MONGO_DB_NAME: 'edxapp'
 EDXAPP_MONGO_USE_SSL: False
 EDXAPP_MONGO_REPLICA_SET: ''
+EDXAPP_MONGO_AUTH_DB: ''
```
feanil: Is there a reason we don't use null here and use an empty string instead?

feanil:
Thanks for the PR @xitij2000 I'll queue it up for review.

playbooks/roles/edxapp/defaults/main.yml
xitij2000: @feanil For both the forum and edx-platform an empty string seems to be backwards-compatible, i.e. passing an empty string to auth_source will use the original behaviour. I wasn't so sure about how null would be rendered and picked up in the configuration, so I did not try it. 

There is no other reason though. In fact in the edx-platform PR I'm still coercing the empty string to a None just to be fully in sync with the previous behaviour. 

I will switch this over to null and see how that works out.

xitij2000:
Note: This includes commits from https://github.com/edx/configuration/pull/5196 since without that our sandboxes fail to provision. Once approved I will remove that commit.

playbooks/roles/edxapp/defaults/main.yml
xitij2000: null seems to work fine, so switched it to null.

lgp171188 (approved):
:+1: , the commits to make the sandbox to work should be removed before merge.
- [x] I tested that on sandboxes that this works okay when the `auth_source` parameter is provided via the `EDXAPP_MONGO_AUTH_DB` variable and not provided (which uses the default value matching the current default).
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ NA
- [x] ~~Includes documentation~~ NA

xitij2000:
@feanil  This is now ready for review

feanil:
Sounds good, I'll queue it up.

ormsbee:
@feanil: Could I trouble you to take this one? I don't have much context for the bit about `AWS_GATHER_FACTS`.

feanil:
I'll take a look.

playbooks/roles/forum/defaults/main.yml
```diff
@@ -59,6 +60,7 @@ forum_base_env: &forum_base_env
   SEARCH_SERVER: "{{ FORUM_ELASTICSEARCH_URL }}"
   MONGOHQ_URL: "{{ FORUM_MONGO_URL }}"
   MONGOID_USE_SSL: "{{ FORUM_MONGO_USE_SSL }}"
+  MONGOID_AUTH_SOURCE: "{{ FORUM_MONGO_AUTH_DB }}"
```
feanil, line 63: For EDXAPP, the config gets output to YAML which will convert the `!!null` to None.  But here, the null will get passed through to a jinja template.  @xitij2000 is the output safe in this scenario?  Can you show me a sample of what the output looks like?

feanil (commented):
Just one last question, I think this will be good to merge once deal with that.

playbooks/roles/forum/defaults/main.yml
xitij2000: @feanil Sure! I had tested this with a simple playbook which passes null on to the template and have subsequently also tested this on a live server. 

[This](https://github.com/edx/configuration/blob/cf1b9e6194d11c8715aab1961e1bd1f6b288c854/playbooks/roles/forum/tasks/main.yml#L61-L75) is where it's being written to the database, and [here](https://github.com/edx/configuration/blob/master/playbooks/roles/forum/templates/forum_env.j2) is the template rendering it. Since the template is only outputing something if a value exists, with `FORUM_MONGO_AUTH_DB` set to `!!null` nothing is output at all. If any value is set you see an `export MONGOID_AUTH_SOURCE="value"` line.

feanil: Ah, good point, forgot about that check before output.  Cool.

feanil (approved):

openedx-webhooks:
@xitij2000 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/configuration&entry.867055334=https://github.com/edx/configuration/pull/5195&entry.1484655318=https://github.com/xitij2000&entry.752974735=2019-06-14+13:15&entry.1917517419=2019-07-26+19:14&entry.2133058324=Yes) so we can improve your experience in the future.

xitij2000:
Thanks!

PR #5197, opened 2019-06-14
Celery script cleanup

jdmulloy:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/check_celery_progress/check_celery_progress.py
+++ b/util/jenkins/check_celery_progress/check_celery_progress.py
@@ -24,7 +24,8 @@ DATE_FORMAT = '%Y-%m-%d %H:%M:%S.%f'
 
 
 class RedisWrapper(object):
-    def __init__(self, dev_test_mode=True, *args, **kwargs):
+    def __init__(self, dev_test_mode=None, *args, **kwargs):
+        assert isinstance(dev_test_mode, bool)
         self.dev_test_mode = dev_test_mode
         self.redis = redis.StrictRedis(*args, **kwargs)
 
@@ -95,7 +96,8 @@ class RedisWrapper(object):
 
 
 class CwBotoWrapper(object):
-    def __init__(self, dev_test_mode=True):
+    def __init__(self, dev_test_mode=None):
+        assert isinstance(dev_test_mode, bool)
         self.dev_test_mode = dev_test_mode
         self.client = boto3.client('cloudwatch')
 
@@ -192,7 +194,8 @@ def generate_alert_alias(environment, deploy, queue_name):
 @backoff.on_exception(backoff.expo,
                       (ApiException),
                       max_tries=MAX_TRIES)
-def create_alert(opsgenie_api_key, environment, deploy, queue_name, threshold, info, dev_test_mode=True):
+def create_alert(opsgenie_api_key, environment, deploy, queue_name, threshold, info, dev_test_mode=None):
+    assert isinstance(dev_test_mode, bool)
 
     configuration.api_key['Authorization'] = opsgenie_api_key
     configuration.api_key_prefix['Authorization'] = 'GenieKey'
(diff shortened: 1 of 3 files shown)


util/jenkins/check_celery_progress/check_celery_progress.py
```diff
@@ -22,6 +22,7 @@
 QUEUE_AGE_HASH_NAME = "queue_age_monitoring"
 DATE_FORMAT = '%Y-%m-%d %H:%M:%S.%f'
 
+dev_test_mode = False
```
bderusha: Should this be part of the command interface so you dont need to edit the script to toggle this on/off?

jdmulloy: Just tried to and I can't edit the global variable from the click command function ```check_queues()```. A flag would be nice, but this is better than editing it every time to change the return to pass.

util/jenkins/check_celery_progress/check_celery_progress.py
```diff
@@ -178,43 +192,54 @@ def generate_alert_alias(environment, deploy, queue_name):
 @backoff.on_exception(backoff.expo,
                       (ApiException),
                       max_tries=MAX_TRIES)
-def create_alert(opsgenie_api_key, environment, deploy, queue_name, threshold, info):
+def create_alert(opsgenie_api_key, environment, deploy, queue_name, threshold, info, dev_test_mode=True):
```
coryleeio: I probably wouldn't make this default to true, if you forget to pass in the arg it will not create the alert. It'd be better to get an alert that you didn't mean to create than to have it fail silently.

util/jenkins/check_celery_progress/check_celery_progress.py
```diff
@@ -178,43 +192,54 @@ def generate_alert_alias(environment, deploy, queue_name):
...
+        print('request id: {}'.format(response.request_id))
+        print('took: {}'.format(response.took))
+        print('result: {}'.format(response.result))
 
 
 @backoff.on_exception(backoff.expo,
                       (ApiException),
                       max_tries=MAX_TRIES)
-def close_alert(opsgenie_api_key, environment, deploy, queue_name):
+def close_alert(opsgenie_api_key, environment, deploy, queue_name, dev_test_mode=True):
```
coryleeio: default to false

util/jenkins/check_celery_progress/check_celery_progress.py
```diff
@@ -178,43 +192,54 @@ def generate_alert_alias(environment, deploy, queue_name):
...
-    print('request id: {}'.format(response.request_id))
-    print('took: {}'.format(response.took))
-    print('result: {}'.format(response.result))
+
+    if dev_test_mode:
+        print("Test Mode: would have closed Alert: {}".format(alias))
+    else:
+        print("Closing Alert: {}".format(alias))
+        # Need body=CloseAlertRequest(source="") otherwise OpsGenie API complains that bdoy must be a json object
+        response = AlertApi().close_alert(identifier=alias, identifier_type='alias', body=CloseAlertRequest(source=""))
```
coryleeio, line 232: You did not pass in test_mode here, so it will default to true and not close the alert., this is is what made me think we should change the default to false.

coryleeio (changes requested):
I would consider changing the default to NOT dev mode, OR not defaulting the function parameter and always explicitly specifying it.

util/jenkins/check_celery_progress/check_celery_progress.py
jdmulloy: @coryleeio This is not my close_alert function, it's the AlertApi's close_alert function from the opsgenie module. Their naming is a bit of a pain as it's very generic.

coryleeio: Ah okay cool

coryleeio (approved):

PR #5198, opened 2019-06-14
update registrar_defaults to include home

fredsmith:
registrar_home is used in the dictionary [here](https://github.com/edx/configuration/blob/4292dbad88d1f1a84be055fc8733d9cdc1cd3d44/playbooks/roles/supervisor/defaults/main.yml#L81) but not defined.


kdmccormick (approved):
Looks good, thanks Fred.

PR #5170, opened 2019-06-04
Add wrapper scripts to fix double reload

jdmulloy:
Configuration Pull Request
---

Make sure that the following steps are done before merging:

  - [ ] A DevOps team member has approved the PR if it is code shared across multiple services and you don't own all of the services.
  - [ ] Are you adding any new default values that need to be overridden when this change goes live? If so:
    - [ ] Update the appropriate internal repo (be sure to update for all our environments)
    - [ ] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [ ] Add an entry to the CHANGELOG.
  - [ ] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).
  - [ ] Think about how this change will affect Open edX operators.  Have you updated the wiki page for the next Open edX release?

PR #5212, opened 2019-06-21
make hermes not hup edxapp

fredsmith:


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -963,6 +963,8 @@ EDXAPP_PROGRAM_MANAGER_MICROFRONTEND_URL: null
 
 # Remote config
 EDXAPP_HERMES_ENABLED: "{{ COMMON_HERMES_ENABLED }}"
+HERMES_RELOAD_COMMAND: "date"
```
bderusha, line 966: There is a cms restart bash script.  I'd rather change that directly so we can keep sending HUPs to lms

bderusha (changes requested):

playbooks/roles/edxapp/defaults/main.yml
bderusha: @jdmulloy added them here https://github.com/edx/configuration/pull/5170/files

PR #5215, opened 2019-06-21
Diff files to prevent unnecessary reloading

jdmulloy:
Configuration Pull Request
---

### DIFF
--- a/playbooks/edxapp.yml
+++ b/playbooks/edxapp.yml
@@ -42,7 +42,9 @@
      hermes_service_config:
        - url: '{{ HERMES_REMOTE_FILE_LOCATION }}/{{ COMMON_ENVIRONMENT }}/lms.yml'
          filename: '{{ hermes_download_dir }}/lms.yml'
+         # Must use ! to invert the exit code here because diff has no flags that invert it's exit code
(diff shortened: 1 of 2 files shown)


playbooks/edxapp.yml
```diff
@@ -42,7 +42,7 @@
      hermes_service_config:
        - url: '{{ HERMES_REMOTE_FILE_LOCATION }}/{{ COMMON_ENVIRONMENT }}/lms.yml'
          filename: '{{ hermes_download_dir }}/lms.yml'
-         command:  "sudo /bin/cp {{ hermes_download_dir }}/lms.yml {{ COMMON_CFG_DIR }}/lms.yml && sudo /edx/app/edxapp/reload_lms_config.sh"
+         command: "if ! diff -q {{ hermes_download_dir }}/lms.yml {{ COMMON_CFG_DIR }}/lms.yml > /dev/null; then sudo /bin/cp {{ hermes_download_dir }}/lms.yml {{ COMMON_CFG_DIR }}/lms.yml && sudo /edx/app/edxapp/reload_lms_config.sh; fi"
```
bderusha, line 46: As a best practice we should prefer affirmative conditionals and use the `-s` rather than `not -q`

bderusha (dismissed):

playbooks/edxapp.yml
jdmulloy: Unfortunately using -d instead of -s doesn't change the exit status of diff. I looked, there's no way to make diff invert it's exit code.

coryleeio (approved):

coryleeio:
We would prefer to do this on the Jenkins side - but if we can't do it there I'm okay with this for now.

jdmulloy:
We ended up doing this on the Jenkins side. Closing.

PR #5218, opened 2019-06-24
Fix JWK decode error on sandboxes, attempt #1

kdmccormick:
Update 2: Setting the default value to the empty string also works and seems more elegant.

Update: This PR sets `COMMON_JWT_PUBLIC_SIGNING_JWK_SET` to `'{"keys": []}'` as default instead of `!!null`. Currently it renders as `None` in the case that the setting is not overridden, which is not valid JSON. The optimal scenario would be making it equal `null`, but that doesn't seem like an option given how the setting is currently handled with the Ansible `|string` filter.

-------------------------------------------------

This is to fix an error I was getting on my sandbox:
```
Jun 21 20:14:48 kdmccormick [service_variant=registrar][django.request][env:no_env] ERROR [kdmccorm
Traceback (most recent call last):
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/django/core/handlers/excepti
  response = get_response(request)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/django/core/handlers/base.py
  response = self.process_exception_by_middleware(e, request)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/django/core/handlers/base.py
  response = wrapped_callback(request, *callback_args, **callback_kwargs)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/django/views/decorators/csrf
  return view_func(*args, **kwargs)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/django/views/generic/base.py
  return self.dispatch(request, *args, **kwargs)
File "/edx/app/registrar/registrar/registrar/apps/api/mixins.py", line 66, in dispatch
  response = super().dispatch(*args, **kwargs)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/rest_framework/views.py", li
  response = self.handle_exception(exc)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/rest_framework/views.py", li
  self.raise_uncaught_exception(exc)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/rest_framework/views.py", li
  self.initial(request, *args, **kwargs)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/rest_framework/views.py", li
  self.perform_authentication(request)
File "/edx/app/registrar/venvs/registrar/lib/python3.5/site-packages/rest_framework/views.py", li
  request.user
[...]

PR #5227, opened 2019-06-27
Fix JWK decode error on sandboxes, attempt #2

kdmccormick:
Builds kept failing for strange reasons on https://github.com/edx/configuration/pull/5218, so I'm opening a new PR. See the original PR for context.


kdmccormick:
Closed in favor of #5228

PR #5232, opened 2019-06-27
Fix JWK decode error on sandboxes

kdmccormick:
@edx/masters-neem 

### First commit: JWK decode error
Original PR: #5218 . See that for context.

### Second commit: Ansible workaround
See ansible/ansible#42620. This was stopping configuration from building correctly, as MySQL was not starting. The change in this commit works around the issue.


coryleeio (approved):
I dont know why the service module is doing this - but this fix wont break it.

PR #5234, opened 2019-07-01
Add private key to hermes config (OPS-3836)

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/hermes/defaults/main.yml
+++ b/playbooks/roles/hermes/defaults/main.yml
@@ -27,6 +27,7 @@ hermes_venv_dir: "{{ hermes_app_dir }}/venvs/{{ hermes_service_name }}"
 hermes_venv_bin: "{{ hermes_venv_dir }}/bin"
 hermes_interval: 60
 hermes_manifest_yaml_file_path: "{{ COMMON_CFG_DIR }}/{{ hermes_service_name }}.yml"
+hermes_private_key_file_path: "{{ hermes_app_dir }}/{{ hermes_service_name }}-private-key"
 hermes_environment:
   PATH: "{{ hermes_venv_dir }}/bin:{{ ansible_env.PATH }}"
 #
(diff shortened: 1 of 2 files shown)


playbooks/roles/hermes/defaults/main.yml
```diff
@@ -27,6 +27,7 @@ hermes_venv_dir: "{{ hermes_app_dir }}/venvs/{{ hermes_service_name }}"
 hermes_venv_bin: "{{ hermes_venv_dir }}/bin"
 hermes_interval: 60
 hermes_manifest_yaml_file_path: "{{ COMMON_CFG_DIR }}/{{ hermes_service_name }}.yml"
+hermes_private_key: "{{ hermes_app_dir }}/{{ hermes_service_name }}-private-key"
```
coryleeio: could you append file_path to this variable like the other one so that it is clear the distinction between the file path and the actual key contents?

playbooks/roles/hermes/tasks/main.yml
```diff
@@ -107,6 +107,18 @@
...
     - install:base
 
+- name: install private key
+  copy:
+    content: "{{ HERMES_PRIVATE_KEY }}"
+    dest: "{{ hermes_private_key }}"
+    force: yes
+    owner: "{{ hermes_user }}"
+    mode: "0600"
+  when: HERMES_USE_PRIVATE_KEY
```
coryleeio: You could key this off of HERMES_PRIVATE_KEY being null, then you dont have to remember to set two variables.
HERMES_PRIVATE_KEY
and 
HERMES_USE_PRIVATE_KEY

coryleeio (changes requested):
see other comments

playbooks/roles/hermes/tasks/main.yml
nadeemshahzad: Thanks @coryleeio , I have addressed your comments and updated PR.

coryleeio (approved):

PR #5236, opened 2019-07-01
Update the COMMON_JWT_PUBLIC_SIGNING_JWK_SET to be empty string 

schenedx:
The value COMMON_JWT_PUBLIC_SIGNING_JWK_SET needs to be changed to empty by fault rather than null

Configuration Pull Request
---

  - [ ] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist]


schenedx:
This is not useful anymore.

schenedx:
Try again with different value

schenedx:
No longer useful

PR #5242, opened 2019-07-02
rolling back the warnings ng plugin for build jenkins

estute:
Rolling back one of the plugins in a recent PR (https://github.com/edx/configuration/pull/5233). I was noticing strange behavior w/ pipeline jobs not generating output and began checking if any of the recent plugin upgrades caused this. I am not sure why this is the culprit, but I rolled it back on test jenkins and the problem has gone away. Will put in a ticket to investigate in the future.


estute:
btw @michaelyoungstrom

PR #5250, opened 2019-07-02
Refactor devstack_sqlite_fix role and use it in jenkins_worker

michaelyoungstrom:
Upgrade sqlite in jenkins workers to avoid segfaults. This role existed to solve this problem in devstack, so I made it slightly more general and added it to the jenkins_worker playbook.


bmedx (approved):
I will not pretend to remember how all of this fits together, but I don't see anything that looks wrong. Thanks for getting this done!

estute (approved):
LGTM, as long as the devstack builds still work after changing the tags

PR #5262, opened 2019-07-10
WIP Clean up rendering code

syedimranhassan:
Configuration Pull Request
---


playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -46,18 +34,6 @@
...
-  tags:
-    - install
-    - install:configuration
-    - install:app-configuration
-    - edxapp_cfg # Old deprecated tag, will remove when possible
-    - edxapp_cfg_yaml_only # Used to render the yaml without the json until we remove the json configs
-
 - name: render studio yml config # meant to replace existing json config eventually
   template:
     src: "studio.yml.j2"
```
coryleeio: We should also remove the templates /configuration/playbooks/roles/edxapp/templates/studio.yml.j2
and the render for lms.yml and studio.yml as well.
That template uses cms_combined_config inside of it, the lms one does similar.


However in sandboxes(and in places in the open source community that don't run Hermes to download their lms.yml and studio.yml (for lms/cms) or in edx-django-service role for everything else.  We will need to provide a method of getting a lms.yml/studio.yml or a {my_app}.yml since the application still uses it.  We just want to copy it from internal instead of merging secure and internal.

In the simplest case we can just copy the file from the source machine that is running ansible 
https://docs.ansible.com/ansible/latest/modules/copy_module.html

instead of rendering it from the union of internal and secure like we have done previously.


We should provide a way to optionally download it from a remote url, or S3, as well if it is not too difficult to have the three methods.

Ideally we would support the three methods in the edxapp role and in the edx_django_service role since that will cover all the places you acquire a lms.yml/studio.yml/{my_app}.yml

Let me know if you've any questions regarding this

coryleeio (changes requested):

PR #5276, opened 2019-07-12
Enable Splunk forwarder for Master's integration sandboxes

kdmccormick:
@edx/masters-neem 
@edx/devops 
https://openedx.atlassian.net/browse/EDUCATOR-4451

Depends on https://github.com/edx/sandbox-internal/pull/49

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -425,6 +425,13 @@ veda_pipeline_worker=${video_pipeline:-false}
 veda_encode_worker=${video_encode_worker:-false}
 video_pipeline_integration=${video_pipeline:-false}
 
+# ansible overrides for master's integration environment setup
+if [[ $registrar == "true" ]]; then
+    cat << EOF >> $extra_vars_file
+COMMON_ENABLE_SPLUNKFORWARDER: true
+EOF
+fi
+
 declare -A deploy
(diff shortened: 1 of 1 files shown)


iloveagent57 (approved):
👍

schenedx (approved):
👍

kdmccormick:
Not merging this yet; I think I also need to do a PR against edx/sandbox-internal to add the Splunk config values. EDIT: Updated the descrpition with said PR.

util/jenkins/ansible-provision.sh
```diff
@@ -425,6 +425,13 @@ veda_pipeline_worker=${video_pipeline:-false}
 veda_encode_worker=${video_encode_worker:-false}
 video_pipeline_integration=${video_pipeline:-false}
 
+# ansible overrides for master's integration environment setup
+if [[ $registrar == "true" ]]; then
```
jdmulloy, line 429: What happens if someone creates a regular (i.e. non-integration) sandbox with registrar?

kdmccormick: Currently we've been running under the assumption that every sandbox with registrar should behave like an integration environment.

If that's a flawed assumption, we could move these behaviors under an `$masters_integration_env` option.

jdmulloy: @kdmccormick I think that might be a good idea. There should probably be some sort of master switch to say this is an integration environment box and not a regular sandbox. At some point someone will probably want registrar on a normal sandbox.

kdmccormick: Makes sense. I'll add a ticket to our backlog to add that switch.

PR #5260, opened 2019-07-09
Create ansible setup for Program Manager to be in sandbox build

schenedx:
Configuration Pull Request
Add Program Manager MFE
---

Make sure that the following steps are done before merging:

  - [ ] A DevOps team member has approved the PR if it is code shared across multiple services and you don't own all of the services.
  - [x] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).

PR #5278, opened 2019-07-12
Create Ansible role to add learner-portal MFE into sandbox builds

Tj-Tracy:
Configuration Pull Request
---

As a note - this work is to support the master's team getting a workable sandbox. This PR should be very similar to https://github.com/edx/configuration/pull/5260


playbooks/learner_portal.yml
```diff
@@ -0,0 +1,21 @@
...
+  gather_facts: True
+  vars:
+    ENABLE_NEWRELIC: False
+    CLUSTER_NAME: 'learner-portal'
+    LEARNER_PORTAL_ENABLED: True
+    LEARNER_PORTAL_SANDBOX_BUILD: False
+  roles:
+    - aws
+    - role: nginx
+      nginx_app_dir: "/etc/nginx"
```
schenedx: I don't think we need this specification. It should just be the default.

playbooks/edx_continuous_integration.yml
```diff
@@ -37,6 +37,8 @@
       when: REGISTRAR_ENABLED
     - role: journals
       when: JOURNALS_ENABLED
+    - role: learner_portal
```
schenedx, line 42: I mentioned this before, but I will mention it here. I believe you need to add `learner_portal` role into the nginx service list after line 14 for learner-portal to be served out of the box on sandbox builds.

playbooks/roles/learner_portal/tasks/main.yml
```diff
@@ -0,0 +1,78 @@
...
+  git:
+    dest: "{{ learner_portal_code_dir }}"
+    repo: "{{ learner_portal_repo }}"
+    version: "{{ learner_portal_version }}"
+    accept_hostkey: yes
+  become_user: "{{ learner_portal_user }}"
+  register: learner_portal_checkout
+
+- name: Install nodeenv
+  command: apt install nodeenv
```
schenedx: You can just use the `apt` module to do this. It's super easy

schenedx (changes requested):
Some minor feedback. I think this is close.

playbooks/edx_continuous_integration.yml
Tj-Tracy: Whoops, I thought I had made this change. Thanks for pointing it out. Though the conversation wasn't updated, I've made the change.

schenedx (approved):
👍

PR #5279, opened 2019-07-12
remove quotes from sudoers file

fredsmith:
quotes are never passed into sudo because the shell intercepts them, which means the sudoers file should have commands without quotes.


adzuci (approved):
:shipit:

PR #5288, opened 2019-07-16
Fix learner portal build

Tj-Tracy:
This fixes the `learner-portal` project build for the sandbox. Somehow, two typos got by without throwing a single error, and I've added the extra task to make sure `npm` properly links to node and builds the right binaries.


schenedx (approved):
👍

PR #5306, opened 2019-07-19
More primary keys improvements

jdmulloy:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/primary_keys/check_primary_keys.py
+++ b/util/jenkins/primary_keys/check_primary_keys.py
@@ -247,7 +247,7 @@ def check_primary_keys(rds_list, username, password, environment, deploy):
                         'MetricName': metric_name,
                         'Dimensions': [{
                             "Name": rds_instance["name"],
-                            "Value": table_name
+                            "Value": table_name_combined
                         }],
                         'Value': table_percent,  # percentage of the usage of primary keys
                         'Unit': UNIT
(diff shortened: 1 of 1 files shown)


util/jenkins/primary_keys/check_primary_keys.py
```diff
@@ -247,20 +247,37 @@ def check_primary_keys(rds_list, username, password, environment, deploy):
...
                         }],
                         'Value': table_percent,  # percentage of the usage of primary keys
                         'Unit': UNIT
                     })
                     table_data["database_name"] = rds_instance['name']
                     table_data["table_name"] = table_name_combined
                     table_data["percentage_of_PKs_consumed"] = table_percent
+                    remaining_days_table_name = table_name_combined
+                    # Hack to transition to metric names with db prepended
+                    if table_name == "courseware_studentmodule" and rds_instance["name"] in [
```
syed-awais-ali, line 260: Is this just for testing?  This can also be put above where we are checking  `> 70` condition as these are the only DBs that will fall into that condition and `if, else` checks can be changed accordingly?

syed-awais-ali (approved):
If the conditions put in is a temporary change (for testing) and may be removed later.

util/jenkins/primary_keys/check_primary_keys.py
jdmulloy: @syed-awais-ali This is because I'm changing the metric name from "table" to "database.table". Since we're using the old data in CloudWatch to calculate the number of remaining days, the script will no longer be able to get the historical data. This is temporary code to track the data in CloudWatch under both the old and new metric names for the table where the metrics already exist with the old name.

syed-awais-ali: @jdmulloy yeah okay. Got it. Thanks

PR #5308, opened 2019-07-22
Move JWT signature stuff into individual roles, only do it for sandboxes

iloveagent57:
Still iterating on https://openedx.atlassian.net/browse/EDUCATOR-4498
The problem I'm solving for now: running the `/edx/bin/update` script on a sandbox will blow away any existing config in `/edx/etc/[service].yml`, because that script doesn't call the `masters_sandbox.yml` playbook.  Rather than always running that playbook, I'd like to have this JWT signature-generation stuff only happen (for the LMS and registrar plays) when we're in the context of a registrar sandbox, i.e. when `REGISTRAR_SANDBOX_BUILD` is true.  So this PR basically splits out my previous additions to `masters_sandbox.yml` into edxapp's `deploy.yml` and registrar's `main.yml` tasks file, conditional on `REGISTRAR_SANDBOX_BUILD`.


schenedx (commented):
I am fine with this general approach. And I am wondering if you can turn all that steps for setting up the jwt in LMS as its own role. Then you can use conditional role dependency on edxapp? See https://edunham.net/2015/09/10/ansible_conditional_role_dependencies.html

schenedx (approved):
👍 Nice work!

playbooks/roles/registrar/tasks/main.yml
```diff
@@ -45,3 +45,14 @@
   tags:
     - install
     - install:configuration
+
+- name: Include JWT signature settings in registrar.yml config file
+  include_role:
+    name: jwt_signature
+  when: REGISTRAR_SANDBOX_BUILD
```
kdmccormick: `{{ REGISTRAR_SANDBOX_BUILD }}`

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -138,3 +138,14 @@
 - include: deploy.yml
   tags:
     - deploy
+
+- name: Include JWT signature settings in lms.yml config file
+  include_role:
+    name: jwt_signature
+  when: EDXAPP_SANDBOX_BUILD
```
kdmccormick: `{{ EDXAPP_SANDBOX_BUILD }}` maybe?

kdmccormick (approved):
Nice :+1:  Approving with the recommendation that you build Docker images for both edxapp and registrar, and make sure that this role wasn't run on them (to ensure that your conditional logic worked correctly).

iloveagent57:
Cool, built a new docker image, spun it up, looked in `/edx/etc/registrar.yml`, didn't have any JWT signature stuff inside the JWT_AUTH dictionary.  Ditto for edxapp/LMS.

schenedx (approved):
👍

kdmccormick (approved):
Nice.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -18,7 +18,7 @@
 
 - name: combined lms auth env for yml
   set_fact:
-    lms_combined_config: '{{lms_env_config|combine(lms_auth_config)}}'
+    app_combined_config: '{{lms_env_config|combine(lms_auth_config)}}'
```
coryleeio: I'd prefer to keep this as lms_combined_config atm because there is also a cms_combined_config. 

We are deleting this in the next few weeks as well if it makes you feel better.

coryleeio: (Unless there is another reason for the change that i'm not aware of)

playbooks/edxapp.yml
```diff
@@ -5,6 +5,7 @@
   vars:
     serial_count: 1
     CLUSTER_NAME: 'edxapp'
+    CAN_GENERATE_NEW_JWT_SIGNATURE: True
```
coryleeio, line 8: This will run everywhere, not just for sandboxes - can we set it to true in internal instead of using the default value to roll it out?

playbooks/roles/jwt_signature/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+    chdir: "{{ edxapp_code_dir }}"
+  when: CAN_GENERATE_NEW_JWT_SIGNATURE
+
+- name: fetch JWT signature settings from host
+  fetch:
+    src: "{{ jwt_signature_file }}"
+    # this will save into /tmp/{{ inventory_hostname }}/{{ jwt_signature_file }} on host
+    dest: /tmp
+
+- name: read JWT signature settings
```
coryleeio, line 15: This is fine for sandboxes, but it will be blown away by remote config at the moment, is that okay?

Why can we not just generate one of these and set it in sandbox-internal / sandbox-secure?

playbooks/roles/edxapp/tasks/service_variant_config.yml
coryleeio: That deletion may cause problems for you doing this also, unclear.  We should probably talk a bit more about this

playbooks/roles/jwt_signature/defaults/main.yml
```diff
@@ -0,0 +1,31 @@
...
+app_config_template: roles/edxapp/templates/lms.yml.j2
+
+# which user and group owns the app config file, with what perms
+app_config_owner: edxapp
+app_config_group: www-data
+app_config_mode: 0640
+
+# Generating the JWT signature settings should only occur once,
+# by the edxapp play, and that play should set this to true.
+CAN_GENERATE_NEW_JWT_SIGNATURE: False
```
coryleeio, line 32: Update runs the edxapp play so i think this isn't true.  ATM it's going to occur every time you update - perhaps we just do it whenever the play is run?

coryleeio (changes requested):

playbooks/roles/jwt_signature/defaults/main.yml
iloveagent57: Continuing conversation from slack here: that command basically appends a new key id to the list of public key ids it will accept (and presumably does something with the private key to make sure any of those public keys will work), so I can't just replace the existing keys

coryleeio: I mostly wanted the comment to be more clear, it doesn't just run once. Maybe just remove that part? It gets run whenever the edxapp play runs

coryleeio (approved):

PR #5309, opened 2019-07-23
PROD-440: Upgrading requests and paramiko dependency to fix security vulnerabilities

jinder1s:
Sustaining and Escalation has received few tickets to fix security vulnerabilities in our system, we are slowly working though them. Here I've changed the pinning on requests and paramiko in requirements/base.in file. From what I could tell this shouldn't break everything, but please verify!

Large ticket explaining task: [here](https://openedx.atlassian.net/secure/RapidBoard.jspa?rapidView=537&modal=detail&selectedIssue=PROD-440)
Ticket with paramiko security issue: [here](https://openedx.atlassian.net/browse/PROD-330)
Ticket with requests security issue: [here](https://openedx.atlassian.net/browse/PROD-293)

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -12,7 +12,7 @@ Jinja2==2.8
 MarkupSafe==1.0
 MySQL-python==1.2.5                 # Needed for the mysql_db module
 networkx==1.11
-paramiko==2.4.1
+paramiko<2.5
 pathlib2==2.3.0
 prettytable==0.7.2
 pycrypto==2.6.1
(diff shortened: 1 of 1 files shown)


jmbowman:
Hi, sorry for the late response (just got back from paternity leave).  Question and comment:

* Any reason why the paramiko version needs to be capped?  My normal preference would be to just unpin it to make future upgrades earlier.
* You need to run `make upgrade` to update the actual requirements files to reflect these changes.

PR #5313, opened 2019-07-29
Get splunk backup dir as parameter

syedimranhassan:
Configuration Pull Request
---

### DIFF
--- a/playbooks/splunk_config_backup.yml
+++ b/playbooks/splunk_config_backup.yml
@@ -1,6 +1,5 @@
 - name: Backup splunk configurations
   hosts: all
-  become: True
   vars:
     splunk_config_dir: "/opt/splunk/etc"
     splunk_host: "{{ splunk_host_id }}"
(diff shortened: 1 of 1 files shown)


playbooks/splunk_config_backup.yml
```diff
@@ -1,20 +1,17 @@
...
     splunk_config_archive: "{{ splunk_host }}-{{ date }}.tar.gz"
-    splunk_s3_backup_tempdir: "/var/tmp"
+    splunk_s3_backup_tempdir: "{{ splunk_backup_dir }}"
     splunk_s3_bucket: "{{ splunk_s3_backups_bucket }}"
   tasks:
     - set_fact:
         date: "{{ lookup('pipe', 'date +%Y-%m-%dT%H%M') }}"
     - name: archive splunk configuration dir
-      command: tar -cpzf  "{{ splunk_s3_backup_tempdir }}/{{ splunk_config_archive }}" {{ splunk_config_dir }}
+      command: sudo tar -cpzf  "{{ splunk_s3_backup_tempdir }}/{{ splunk_config_archive }}" {{ splunk_config_dir }}
```
adzuci, line 13: I like that we're not using become, but am curious how hard it would be to further protect Splunk considering how many people have access to run things on Tools jenkins.  

I think that we could manually update the `etc/sudoers` file to include `tar -cpzf *`, `aws s3 cp` & `rm *` and update the Splunk creation role to have a sudoers template like we do for https://github.com/edx/configuration/blob/master/playbooks/roles/user/templates/restricted.sudoers.conf.j2.

Unfortunately the https://github.com/edx/configuration/tree/master/playbooks/roles/splunk-server role can not be deployed, so it's essentially just a way to document how we build splunk instances incase we need to deploy a new one.

Semi-related, did you create the splunkbackup user with a playbook?  I don't seem to have an account on that box and would like to create one for myself.

adzuci (approved):
Requested some changes, but if they don't make sense let me know.

playbooks/splunk_config_backup.yml
syedimranhassan: Thanks @adzuci , I have already added new ```splunkbackup``` user on tools splunk machine and  file ```etc/sudoers.d/splunkback``` with required permissions to run this play.

adzuci: Awesome!  I must have been looking in the wrong place, thanks for taking care of this ticket!

PR #5316, opened 2019-07-29
Add zoom xblock

davestgermain:
This installs the Zoom XBlock. It's not in edx-platform's requirements because it's not ready to be made available to the OpenEdx community.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -520,6 +520,8 @@ EDXAPP_PRIVATE_REQUIREMENTS:
       extra_args: -e
     - name: git+https://github.com/open-craft/xblock-activetable.git@e933d41bb86a8d50fb878787ca680165a092a6d5#egg=xblock-activetable
       extra_args: -e
+    - name: git+https://github.com/edx/edx-zoom.git@ffd44b69735a670e889d14eaf4cd19fefb582e92#egg=edx-zoom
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -520,6 +520,7 @@ EDXAPP_PRIVATE_REQUIREMENTS:
       extra_args: -e
     - name: git+https://github.com/open-craft/xblock-activetable.git@e933d41bb86a8d50fb878787ca680165a092a6d5#egg=xblock-activetable
       extra_args: -e
+    - name: git+https://github.com/edx/edx-zoom.git@ffd44b69735a670e889d14eaf4cd19fefb582e92#egg=edx-zoom
```
bderusha, line 523: This need the `extra_args: -e` line as well

bderusha (changes requested):

playbooks/roles/edxapp/defaults/main.yml
davestgermain: I didn't think that was necessary since that installs the package in "editable" mode (which is not relevant on prod), but I just added it.

bderusha: edX best practice has been to use this flag to prevent weird caching issues IIRC.  Could find docs on it though. _shrug_

bderusha (approved):

PR #5320, opened 2019-07-30
Upgrade Jenkins LTS version to 2.176.2

michaelyoungstrom:
- Upgrade our Jenkins version
- Stop running the shutdown CLI script, as it's no longer relevant
- Plugin bumps

### DIFF
--- a/playbooks/roles/jenkins_build/defaults/main.yml
+++ b/playbooks/roles/jenkins_build/defaults/main.yml
@@ -1,6 +1,6 @@
 build_jenkins_user_uid: 1002
 build_jenkins_group_gid: 1004
-build_jenkins_version: jenkins_2.164.3
+build_jenkins_version: jenkins_2.176.2
 build_jenkins_jvm_args: '-Djava.awt.headless=true -Xmx16384m -DsessionTimeout=60'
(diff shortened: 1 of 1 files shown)


estute (approved):

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -202,7 +201,7 @@ build_jenkins_plugins_list:
     version: '1.0'
     group: 'org.jenkins-ci.plugins'
   - name: 'script-security'
-    version: '1.58'
+    version: '1.62'
```
estute, line 204: ^ this is the only one I am *slightly* concerned about. Is it going to break whitelisting of scripts?

estute (approved):
one question, but otherwise, LGTM

playbooks/roles/jenkins_build/defaults/main.yml
michaelyoungstrom: I don't think so. The general trend (this jump included) is that they add more and more things to their generic whitelist. I don't see anything in the changelog that makes me suspicous of things breaking.

PR #5326, opened 2019-08-01
Increase LMS client_max_body_size to 10MB.

waheedahmed:
On latest devices with higher resolution webcams, images are larger and
increasing request size.

PROD-499

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -65,13 +65,13 @@ server {
 
   {% if NGINX_EDXAPP_ENABLE_S3_MAINTENANCE %}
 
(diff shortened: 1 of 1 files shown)


fredsmith (approved):

feanil:
@fredsmith @waheedahmed if there isn't a way to transmit lower resolution video, then we'll probably want to up the cap on the max number of lms workers a bit since we will be increasing the amount of time spent per request if we're transmitting larger amounts of data per request.

waheedahmed:
@feanil I have reduced the resolution for webcams but the problem comes when learner submit the photos using mobile devices from select input field and select an already taken larger size photo with main camera of the device. I don't think there will be many requests coming from mobile devices because only few devices show select input field instead of webcam e.g. iPhones.

feanil:
Cool, thanks for the context.  That would have been good to put in the commit message or as a comment in the NGINX profile as well.

PR #5330, opened 2019-08-05
Allow ssh access to users (OPS-4232)

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/user/tasks/main.yml
+++ b/playbooks/roles/user/tasks/main.yml
@@ -225,3 +225,17 @@
   with_nested:
     - "{{ user_info }}"
     - "{{ user_rbash_links }}"
+
+- name: Allow users to ssh
+  lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: "#users_allowed_to_ssh {{ vars['user_info'] }}"
+    line: "AllowUsers {{ (( user_info | map(attribute='name') | list ) + [ ansible_ssh_user ]) | join(' ') }} #users_allowed_to_ssh {{ vars['user_info'] }}"
+  when: user_info|length > 0
+  register: users_ssh_access
+
+- name: restart ssh
+  service:
+    name: ssh
+    state: restarted
+  when: users_ssh_access.changed
--- a/playbooks/roles/automated/tasks/main.yml
+++ b/playbooks/roles/automated/tasks/main.yml
@@ -89,3 +89,22 @@
     owner: "{{ item.key }}"
     group: "{{ item.key }}"
   with_dict: "{{ AUTOMATED_USERS }}"
+
+- name: Are we in a Docker container
+  shell: echo $(egrep -q 'docker' /proc/self/cgroup && echo 'yes' || echo 'no')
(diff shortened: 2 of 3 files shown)


playbooks/roles/user/tasks/main.yml
```diff
@@ -225,3 +225,17 @@
   with_nested:
     - "{{ user_info }}"
     - "{{ user_rbash_links }}"
+
+- name: Allow user to login remotely
```
coryleeio: You have to add the Deny line as well or it doesn't work.

You need to deny *, then explicitly allow the people in the list after the deny

coryleeio (changes requested):

playbooks/roles/user/tasks/main.yml
nadeemshahzad: @coryleeio ```DenyUsers *``` will disallow login for all and not working here. ```AllowUsers``` will make sure that login is allowed only for user names provided in list and will deny all other users.

coryleeio: You are correct, as per the documentation here:
https://www.freebsd.org/cgi/man.cgi?query=sshd_config&sektion=5

coryleeio (approved):

playbooks/roles/user/tasks/main.yml
```diff
@@ -225,3 +225,17 @@
...
+  lineinfile:
+    dest: /etc/ssh/sshd_config
+    regexp: '^AllowUsers'
+    line: AllowUsers {{ USERS_ALLOWED_TO_SSH.split(',')| join(' ') }}
+  when: USERS_ALLOWED_TO_SSH is defined
+  register: users_ssh_access
+
```
coryleeio, line 236: Can you derive USERS_ALLOWED_TO_SSH from the existing list so that we dont have to remember to add USERS_ALLOWED_TO_SSH when people update their version of configuration?

It should be possible like so:

`{{ ['x','y'] | map('extract', {'x': 42, 'y': 31}) | list }}
results in 
https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#extracting-values-from-containers`

coryleeio (changes requested):

playbooks/roles/user/tasks/main.yml
coryleeio: I think you can use the map method even with the ubuntu user if you add him and set github:false on him. Will that make it so that we can use the map method?

nadeemshahzad: Thanks @coryleeio , I have updated PR to dervied users from existing vars and closed the internal PR also added ```AUTOMATED_USERS``` to list as we need these to run the management commands from jenkins.

coryleeio (approved):

mahyard:
After a native installation I lost access to my VM because of this line at the end of sshd_config:
```
AllowUsers automated_user #automated_users_allowed_to_ssh
```
It would make trouble for people in the future.
@nadeemshahzad @coryleeio

PR #5331, opened 2019-08-05
Include Greenhouse token in prospectus build (for careers page).

dianakhuang:
Configuration Pull Request
---


playbooks/roles/prospectus/templates/.env.environment.j2
```diff
@@ -5,3 +5,5 @@ OAUTH_SECRET={{ PROSPECTUS_OAUTH_SECRET }}
 
 NEW_RELIC_LICENSE_KEY={{ PROSPECTUS_NEW_RELIC_LICENSE_KEY }}
 NEW_RELIC_APP_ID={{ PROSPECTUS_NEW_RELIC_APP_ID }}
```
coryleeio, line 7: You will also need to set the values here:
https://github.com/edx/configuration/blob/master/playbooks/roles/prospectus/defaults/main.yml#L15-L17

coryleeio (changes requested):

playbooks/roles/prospectus/templates/.env.environment.j2
coryleeio: @dianakhuang  If you lastpass share me the key I will add it for you to secure

dianakhuang: @AlasdairSwan can you get DevOps the API key?

playbooks/roles/prospectus/defaults/main.yml
```diff
@@ -19,6 +19,8 @@ PROSPECTUS_OAUTH_SECRET: !!null
 PROSPECTUS_NEW_RELIC_LICENSE_KEY: 'fake-key'
 PROSPECTUS_NEW_RELIC_APP_ID: 'fake-id'
 
+GATSBY_GREENHOUSE_API_KEY: 'fake-key'
```
coryleeio: this should be PROSPECTUS_NEW_RELIC_APP_ID
not GATSBY_GREENHOUSE_API_KEY

coryleeio (changes requested):

coryleeio:
@AlasdairSwan @dianakhuang Same key for stage vs prod?

AlasdairSwan:
@coryleeio yes please.

playbooks/roles/prospectus/defaults/main.yml
dianakhuang: I believe I did screw this up, but I fixed it to be `GATSBY_GREENHOUSE_API_TOKEN` instead.

playbooks/roles/prospectus/defaults/main.yml
```diff
@@ -19,6 +19,8 @@ PROSPECTUS_OAUTH_SECRET: !!null
 PROSPECTUS_NEW_RELIC_LICENSE_KEY: 'fake-key'
 PROSPECTUS_NEW_RELIC_APP_ID: 'fake-id'
 
+GATSBY_GREENHOUSE_API_TOKEN: 'fake-key'
```
coryleeio:
```suggestion
PROSPECTUS_GREENHOUSE_API_TOKEN: 'fake-key'
```

coryleeio (changes requested):

playbooks/roles/prospectus/defaults/main.yml
coryleeio: This results in the same behavior and is consistent with the application specific values above it.

playbooks/roles/prospectus/templates/.env.environment.j2
coryleeio: I have the key

playbooks/roles/prospectus/defaults/main.yml
dianakhuang: Unfortunately, some of these values have to start with `GATSBY` for uh, gatsby-specific reasons. We can switch the key to `GATSBY_PROSPECTUS_GREENHOUSE_API_TOKEN` if absolutely necessary, but I'm not a big fan of making it that long.

coryleeio (approved):

PR #5334, opened 2019-08-05
Dockerfile fix 404, debian mirror removed

hunytalk:
Configuration Pull Request
Related Ticket: https://openedx.atlassian.net/browse/DE-1603
All images related to hadoop / spark have been failing due to one of the mirrors ("jessie") removed from debian. ( https://lists.debian.org/debian-devel-announce/2019/03/msg00006.html )

### DIFF
--- a/docker/build/analytics_pipeline_hadoop_datanode/Dockerfile
+++ b/docker/build/analytics_pipeline_hadoop_datanode/Dockerfile
@@ -1,15 +1,20 @@
 FROM uhopper/hadoop:2.7.2
 LABEL maintainer="edxops"
 
+RUN echo "deb http://deb.debian.org/debian buster-backports main" > /etc/apt/sources.list.d/backports.list
+
+RUN \
+echo "deb http://mirrors.linode.com/debian/ stretch main" > /etc/apt/sources.list && \
+echo "deb-src http://mirrors.linode.com/debian/ stretch main" >> /etc/apt/sources.list && \
+echo "deb http://mirrors.linode.com/debian-security/ stretch/updates main" >> /etc/apt/sources.list && \
+echo "deb-src http://mirrors.linode.com/debian-security/ stretch/updates main" >> /etc/apt/sources.list && \
+echo "deb http://mirrors.linode.com/debian/ stretch-updates main" >> /etc/apt/sources.list && \
+echo "deb-src http://mirrors.linode.com/debian/ stretch-updates main" >> /etc/apt/sources.list
+
 ENV HDFS_CONF_dfs_datanode_data_dir=file:///hadoop/dfs/data \
     MYSQL_VERSION=5.6 \
     DEBIAN_FRONTEND=noninteractive
 WORKDIR /tmp
-RUN \
-  echo "deb http://ftp.de.debian.org/debian/ stretch main non-free contrib\n" > /etc/apt/sources.list && \
-  echo "deb-src http://ftp.de.debian.org/debian/ stretch main non-free contrib\n" >> /etc/apt/sources.list && \
-  echo "deb http://security.debian.org/ stretch/updates main contrib non-free\n" >> /etc/apt/sources.list && \
-  echo "deb-src http://security.debian.org/ stretch/updates main contrib non-free" >> /etc/apt/sources.list
 
 RUN apt-get -y update
 RUN apt-get -yqq install apt-transport-https lsb-release ca-certificates gnupg2
--- a/docker/build/analytics_pipeline_spark_master/Dockerfile
+++ b/docker/build/analytics_pipeline_spark_master/Dockerfile
@@ -1,6 +1,9 @@
 FROM bde2020/spark-base:2.1.0-hadoop2.7
 LABEL maintainer="edxops"
 
+RUN echo "deb [check-valid-until=no] http://archive.debian.org/debian jessie-backports main" > /etc/apt/sources.list.d/jessie-backports.list
+RUN sed -i '/deb http:\/\/deb.debian.org\/debian jessie-updates main/d' /etc/apt/sources.list
+
 ADD docker/build/analytics_pipeline_spark_master/master.sh /
 ENV SPARK_MASTER_PORT=7077 \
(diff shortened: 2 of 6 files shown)


docker/build/analytics_pipeline_spark_master/Dockerfile
```diff
@@ -27,7 +30,18 @@ ENV SPARK_MASTER_PORT=7077 \
...
     YARN_CONF_yarn_resourcemanager_address=resourcemanager:8032 \
     YARN_CONF_yarn_resourcemanager_scheduler_address=resourcemanager:8030 \
-    YARN_CONF_yarn_resourcemanager_resource__tracker_address=resourcemanager:8031
+    YARN_CONF_yarn_resourcemanager_resource__tracker_address=resourcemanager:8031 \
+    SPARK_VERSION=2.1.0 \
+    HADOOP_VERSION=2.7
+
+RUN apt-get -y update && apt-get -y install wget
+
+# We are downloading spark again as base spark image (here bde2020/spark-base) has been removed.
```
brianhw: I'm confused.   If bde2020/spark-base has been removed, do we need to change line 1?  And would something similar need to be done in analytics_pipeline_spark_worker as well?

macdiesel (approved):

docker/build/analytics_pipeline_hadoop_datanode/Dockerfile
```diff
@@ -1,15 +1,13 @@
 FROM uhopper/hadoop:2.7.2
 LABEL maintainer="edxops"
 
+FROM debian:stable
```
pwnage101: How come you introduce another `FROM` statement without copying any build artifacts from the previous stage?  In fact, it doesn't appear that the first `FROM` has any impact at all since there are no `RUN` statements before the second `FROM`.  Can you elaborate a bit more on this?

pwnage101: more details about multi-stage builds: https://docs.docker.com/develop/develop-images/multistage-build/

docker/build/analytics_pipeline_spark_master/Dockerfile
hunytalk: Sorry maybe i made it confusing. I meant to say that spark is not available at the given path here.
I didn't change anything in bde2020/spark-base, just downloaded spark now. 
Taking a look at spark worker now.

docker/build/analytics_pipeline_hadoop_datanode/Dockerfile
hunytalk: Removed multi stage builds now

docker/build/analytics_pipeline_spark_master/Dockerfile
hunytalk: Spark worker is working fine locally and in test as well. I don't think there will be a need to in analytics_pipeline_spark_worker

docker/build/analytics_pipeline_spark_master/Dockerfile
```diff
@@ -27,9 +30,20 @@ ENV SPARK_MASTER_PORT=7077 \
...
+    HADOOP_VERSION=2.7
+
+RUN apt-get -o Acquire::Check-Valid-Until=false update && apt-get -y install wget
+
+# We are downloading spark again as base spark image (here bde2020/spark-base) has been modified and not available at this path.
+RUN wget https://archive.apache.org/dist/spark/spark-${SPARK_VERSION}/spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz \
+      && tar -xvzf spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz \
+      && mv spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION} spark \
+      && rm spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz \
+      && cd /
```
brianhw: Sorry, this part still confuses me.   Best I can tell from the BDE images, there should be a spark instance already installed.   At least, looking at https://github.com/big-data-europe/docker-spark/blob/2.1.0-hadoop2.7/base/Dockerfile.  Is the issue that the definition in the base docker image's docker file is doing a wget?   I thought that would have already been done at the time that the image was built, two years ago.   (That is, the wget on http://d3kbcqa49mib13.cloudfront.net/spark-${SPARK_VERSION}-bin-hadoop${HADOOP_VERSION}.tgz .)

docker/build/analytics_pipeline_spark_master/Dockerfile
```diff
@@ -27,9 +30,20 @@ ENV SPARK_MASTER_PORT=7077 \
     YARN_CONF_yarn_timeline___service_hostname=historyserver \
     YARN_CONF_yarn_resourcemanager_address=resourcemanager:8032 \
     YARN_CONF_yarn_resourcemanager_scheduler_address=resourcemanager:8030 \
-    YARN_CONF_yarn_resourcemanager_resource__tracker_address=resourcemanager:8031
+    YARN_CONF_yarn_resourcemanager_resource__tracker_address=resourcemanager:8031 \
+    SPARK_VERSION=2.1.0 \
+    HADOOP_VERSION=2.7
```
brianhw: Are these two lines necessary?   They are presumably defined in the ENV of the base image -- doesn't that carry over?  It would be nice if we only have to worry about the versions when defining the base image (i.e. 2.1.0-hadooop2.7 or whatever).

docker/build/analytics_pipeline_spark_master/Dockerfile
hunytalk: Initially i tested locally and were unable to find spark configuration files. but as you mentioned it, it's working fine now locally without having to download spark again. I'll commit new changes

coryleeio (approved):
There is no need for devops to review this I think. LGTM so long as data eng is comfortable with the sources. I'll leave it to them to make the call.

brianhw (approved):
Thanks for the changes.   LGTM, assuming it works. : )

hunytalk:
Yes locally these changes are working fine.

PR #5345, opened 2019-08-09
Install asym-crypto-yaml into go-agent (OPS-4261)

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/docker/build/go-agent/Dockerfile
+++ b/docker/build/go-agent/Dockerfile
@@ -85,6 +85,10 @@ RUN apt-get update && apt-get install zlib1g-dev wget lsb-core -qy && cd /opt &&
 # Install AWS command-line interface - for AWS operations in a go-agent task.
 RUN pip install 'awscli>=1.11.58'
 
+# Install pip on python3.6 to install asym-crypto-yaml which requires python >= 3.6
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+python3.6 get-pip.py
+
 # !!!!NOTICE!!!! ---- Runner of this pipeline take heed!! You must replace go_github_key.pem with the REAL key material
 # that can checkout private github repositories used as pipeline materials. The key material here is faked and is only
 # used to pass CI!


docker/build/go-agent/Dockerfile
```diff
@@ -85,6 +85,13 @@ RUN apt-get update && apt-get install zlib1g-dev wget lsb-core -qy && cd /opt &&
 # Install AWS command-line interface - for AWS operations in a go-agent task.
 RUN pip install 'awscli>=1.11.58'
 
+# Install pip on python3.6
+curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
+python3.6 get-pip.py
+
+# Install asym-crypto-yaml to decrypt encrypted yaml config files
+pip3.6 install asym-crypto-yaml
```
coryleeio: I think you just want to add this to an appropriate .in file the configuration/requirements and run make upgrade

coryleeio (changes requested):

docker/build/go-agent/Dockerfile
nadeemshahzad: @coryleeio I try to run make upgrade after adding asym-crypto-yaml to new ini file, but it failed as we are using pip-compile with python2.7 for all packages and asym-crypto-yaml requires python >=3.6. 
```pip-compile --upgrade -o requirements3.6.txt requirements/asym-crypto-yaml.in
Could not find a version that matches asym-crypto-yaml (from -r requirements/asym-crypto-yaml.in (line 2))
No versions found
Was https://pypi.python.org/simple reachable?
```
Can I add this requirement to tubular in new requirements3.6.txt file and add new task in pipeline to install this using pip3.6 or pre-build go-agent having this requirement installed ?

coryleeio: Yeah I suppose the extra requirements3.txt file would be better than this as otherwise the configuration playbook calls out to something that may or may not actually be there.  Atleast this way the play would install it even if it is less than ideal.

Another option would be adding python 2.7 to tox and verifying outside of docker in a normal venv that the tests pass  in python2 for asym-crypto-yaml.  If they do we *might* be able to relax the version constraint.  I would do one of those things.. Given that python 2 is almost not supported lets go with the requirements3.txt idea I suppose. Thanks!

coryleeio (changes requested):

nadeemshahzad:
@coryleeio I have updated tox and setup.py to run tests with python 2.7 and python 3.5 but tests are failing.  I have updated configuration PR ( https://github.com/edx/configuration/pull/5346 ) to add new requirement file ```asym-crypto-yaml.txt``` that we can use to install asym-crypto-yaml in pipelines using pip3.6.

coryleeio (approved):

PR #5346, opened 2019-08-09
decrypt and copy app configs from internal

nadeemshahzad:
Configuration Pull Request
---


playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -93,12 +93,23 @@
...
+    - install:app-configuration
+
+- name: Copy app config file
+  copy:
+    src: "{{ UNENCRYPTED_CFG_DIR }}/{{ edx_service_name }}.yml"
+    dest: "{{ COMMON_CFG_DIR }}/{{ edx_service_name }}.yml"
+    mode: 0640
+  become: true
+  no_log: "{{ COMMON_CONFIG_NO_LOGGING }}"
+  when: REMOTE_CONFIG_ENABLED
```
coryleeio: Make a new variable instead of REMOTE_CONFIG_ENABLED to enable the copy.

The reason is that not everyone will be using encrypted configs, so you might want to copy without doing the decrypt.
Technically if you dont provide a key it would work, but we dont want people to have to think about it if they are not using encrypted configs.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,65 +16,62 @@
...
+    src: "{{ UNENCRYPTED_CFG_DIR }}/lms.yml"
+    dest: "{{ COMMON_CFG_DIR }}/lms.yml"
+    owner: "{{ edxapp_user }}"
+    group: "{{ common_web_group }}"
+    mode: 0640
+  become: true
   no_log: "{{ COMMON_CONFIG_NO_LOGGING }}"
-  when: '"cms" in service_variants_enabled'
+  with_items: "{{ service_variants_enabled }}"
+  when: '"lms" in service_variants_enabled REMOTE_CONFIG_ENABLED'
```
coryleeio: Make a new variable instead of REMOTE_CONFIG_ENABLED to enable the copy.

The reason is that not everyone will be using encrypted configs, so you might want to copy without doing the decrypt.
Technically if you dont provide a key it would work, but we dont want people to have to think about it if they are not using encrypted configs.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,65 +16,62 @@
...
+    src: "{{ UNENCRYPTED_CFG_DIR }}/cms.yml"
+    dest: "{{ COMMON_CFG_DIR }}/cms.yml"
     owner: "{{ edxapp_user }}"
     group: "{{ common_web_group }}"
     mode: 0640
   become: true
   no_log: "{{ COMMON_CONFIG_NO_LOGGING }}"
   with_items: "{{ service_variants_enabled }}"
-  when: '"cms" in service_variants_enabled'
+  when: '"cms" in service_variants_enabled REMOTE_CONFIG_ENABLED'
```
coryleeio: Make a new variable instead of REMOTE_CONFIG_ENABLED to enable the copy.

The reason is that not everyone will be using encrypted configs, so you might want to copy without doing the decrypt.
Technically if you dont provide a key it would work, but we dont want people to have to think about it if they are not using encrypted configs.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,65 +16,62 @@
...
-    owner: "{{ edxapp_user }}"
-    group: "{{ common_web_group }}"
-    mode: 0640
-  become: true
-  no_log: "{{ COMMON_CONFIG_NO_LOGGING }}"
+- name: Decrypt cms config
+  local_action: command asym_crypto_yaml decrypt-encrypted-yaml --secrets_file_path {{ ENCRYPTED_CFG_DIR }}/cms.yml --private_key_path {{ REMOTE_CONFIG_PRIVATE_KEY }} --outfile_path {{ UNENCRYPTED_CFG_DIR }}/cms.yml
   with_items: "{{ service_variants_enabled }}"
-  when: '"lms" in service_variants_enabled'
+  when: '"cms" in service_variants_enabled and REMOTE_CONFIG_ENABLED'
```
coryleeio: Make a new variable instead of REMOTE_CONFIG_ENABLED to enable the copy separately from the decrypt.

The reason is that not everyone will be using encrypted configs, so you might want to copy without doing the decrypt.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,65 +16,62 @@
...
-  template:
-    src: "lms.yml.j2"
-    dest: "{{ COMMON_CFG_DIR }}/lms.yml"
-    owner: "{{ edxapp_user }}"
-    group: "{{ common_web_group }}"
-    mode: 0640
-  become: true
-  no_log: "{{ COMMON_CONFIG_NO_LOGGING }}"
+- name: Decrypt cms config
+  local_action: command asym_crypto_yaml decrypt-encrypted-yaml --secrets_file_path {{ ENCRYPTED_CFG_DIR }}/cms.yml --private_key_path {{ REMOTE_CONFIG_PRIVATE_KEY }} --outfile_path {{ UNENCRYPTED_CFG_DIR }}/cms.yml
```
coryleeio: these should be studio.yml

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,65 +16,62 @@
...
     - install:app-configuration
     - edxapp_cfg # Old deprecated tag, will remove when possible
-    - edxapp_cfg_yaml_only # Used to render the yaml without the json until we remove the json configs
 
-- name: combined cms auth env for yml
-  set_fact:
-    cms_combined_config: '{{cms_env_config|combine(cms_auth_config)}}'
+- name: Copy lms config file
+  copy:
+    src: "{{ UNENCRYPTED_CFG_DIR }}/lms.yml"
```
coryleeio: Since the copy may happen for people that are not using encryption can you replace UNENCRYPTED_CFG_DIR with something like LOCAL_CONFIG_FILE

and just set LOCAL_CONFIG_FILE = {{ UNENCRYPTED_CFG_DIR }}/lms.yml


that away people that don't want to use asym-crypto-yaml remote config can just specify where the yml is on their local system to upload without having to have the exact file structure as we have to get it working.

playbooks/roles/edxapp/tasks/service_variant_config.yml
```diff
@@ -16,65 +16,62 @@
...
-    - edxapp_cfg_yaml_only # Used to render the yaml without the json until we remove the json configs
 
-- name: render studio yml config # meant to replace existing json config eventually
-  template:
-    src: "studio.yml.j2"
-    dest: "{{ COMMON_CFG_DIR }}/studio.yml"
+- name: Copy cms config file
+  copy:
+    src: "{{ UNENCRYPTED_CFG_DIR }}/cms.yml"
+    dest: "{{ COMMON_CFG_DIR }}/cms.yml"
```
coryleeio: studio.yml here also

coryleeio (changes requested):
Looks great - I left a few comments, but it's really close.

coryleeio (approved):

coryleeio:
I approved this - tests are failing though, I think if you deactivate the decryption by default it might make it easier to get tests passing.

coryleeio:
I think it's reasonable to default it to off since it's not the default behavior in the plays or for the open source community typically.

coryleeio (approved):

PR #5355, opened 2019-08-14
Upgraded Zoom XBlock

davestgermain:


davestgermain:
@bderusha I think edx/app-permissions#961 failed because one of the model permissions doesn't exist yet. This PR adds that model.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -525,7 +525,7 @@ EDXAPP_PRIVATE_REQUIREMENTS:
       extra_args: -e
     - name: git+https://github.com/open-craft/xblock-activetable.git@e933d41bb86a8d50fb878787ca680165a092a6d5#egg=xblock-activetable
       extra_args: -e
-    - name: git+https://github.com/edx/edx-zoom.git@ffd44b69735a670e889d14eaf4cd19fefb582e92#egg=edx-zoom
+    - name: git+https://github.com/edx/edx-zoom.git@7feb9972c4c689acf00bf9f1a0b77d0f21b52d9a#egg=edx-zoom
```
bderusha, line 528: Another option is to tag commits in this repo with versions and reference the tag name.  `edx-zoom.git@1.0.1#egg=edx-zoom`  Might be easier to track and reason about differences between the xblock and what we are running in production

bderusha (approved):

PR #5358, opened 2019-08-16
create multiple virtualenvs for jenkins workers

estute:
Configuration Pull Request
---

-----
We already install multiple versions of Python onto Jenkins workers to be used by tox, but we only create a python 2.7 virtualenv. This is used by the scripts prior to tox invocation (i.e. worker spin up, node-env creation, etc). This PR will create a virtualenv (and tar it) for each installed version of python.

Requires that https://github.com/edx/edx-platform/pull/21353 be merged.

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
+++ b/playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
@@ -23,9 +23,8 @@
   become_user: "{{ jenkins_user }}"
 
 # Combine testing and django requirements files for single virtualenv invocation
-- name: cat
-  command: "cat shallow-clone/requirements/edx/testing.txt shallow-clone/requirements/edx/django.txt shallow-clone/requirements/edx/jenkins.txt"
-  args:
+- name: Combine requirements files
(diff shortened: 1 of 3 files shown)


playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
```diff
@@ -13,27 +13,44 @@
...
+  with_items: "{{ jenkins_worker_python_versions }}"
+  become_user: "{{ jenkins_user }}"
+
+# Temporary fix: make a copy of the python 2.7 virtualenv with a
+# generic name to avoid breaking older PRs
+- name: Make copy of py2.7 virtualenv tarball
+  command: "cp edx-venv_clean-2.7.tar.gz edx-venv_clean.tar.gz"
   args:
     chdir: "{{ jenkins_home }}"
+  with_items: "{{ jenkins_worker_python_versions }}"
```
jmbowman: This task only needs to be run once for the 2.7 virtualenv, doesn't it?

feanil: Won't we want 3.5 venvs to pull down for faster requirements installing?

jmbowman: Yes, but that's the task above; this one is just to keep a copy of the 2.7 virtualenv under the original name.

estute: my mistake. Fixing.

jmbowman (approved):

PR #5362, opened 2019-08-21
add ls step

rgraber:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/prospectus/tasks/main.yml
+++ b/playbooks/roles/prospectus/tasks/main.yml
@@ -134,6 +134,25 @@
     - install
     - install:app-requirements
 
+#show installed packages for debugging
+- name: show installed packages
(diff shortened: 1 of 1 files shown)


adzuci (approved):
:shipit:

playbooks/roles/prospectus/tasks/main.yml
```diff
@@ -134,7 +134,26 @@
...
+  become_user: "{{ prospectus_user }}"
+  tags:
+    - install
+    - install:app-requirements
+
+- name: print installed package output
+  debug: msg="{{ npm_ls_out.stdout }}"
+  when: npm_ls_out.stdout is defined
+  tags:
+    - "install"
```
adzuci, line 153: I missed that these tags are quoted and may not need to be, but I don't think that would cause the current failure mode.

PR #5390, opened 2019-09-03
enable decrypt and copy config sandbox (OPS-4261)

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -94,7 +94,7 @@ program_manager="false"
 if [[ $edx_internal == "true" ]]; then
     # if this is a an edx server include
     # the secret var file
-    extra_var_arg="-e@${sandbox_internal_vars_file} -e@${sandbox_secure_vars_file} -e@${extra_vars_file}"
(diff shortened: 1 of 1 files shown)


coryleeio (approved):
This should be fine to go ahead and merge - before closing this though, can you make it so that the jenkins job can be parameterized with which yml file to decrypt from the internal repo?  I think it is safe to assume the key is the same for now.

The reason for this is that some of our long lived sandboxes are essentially working around having to use the main config.yml and could do a lot less if they just used a different config.yml so I feel like it might be worth enabling them to do this for the future.

coryleeio:
Feel free to do the above as a separate PR

nadeemshahzad:
Thanks @coryleeio I have made PR (https://github.com/edx/jenkins-job-dsl/pull/784) to add parameters  in sandbox job and made relevant changes in this PR to enable decrypt and copy config for apps.

coryleeio:
@nadeemshahzad Can you make it possible for the sandbox job to change their yaml file that they upload from internal?

For instance, instead of sandbox-internal/sandbox-remote-config/notes.yml someone might want to do the following to test a new feature:
sandbox-internal/sandbox-remote-config/sally-notes.yml

coryleeio:
The default should be set to the normal .yml in sandbox-internal though

PR #5414, opened 2019-09-23
Clean up rendering code from configuration repo

syedimranhassan:
Configuration Pull Request
---


syedimranhassan:
I have made separate commit for each application.
For time being I have commented the ansible tasks which required lms/cms auth and env variables. So, it's easy for review and uncomment them if we don't want to remove them.

coryleeio (changes requested):
This looks good - but I think you need to still remove edx_django_service role rendering.

I think you should separate this into two PRs as well, one for lms/cms

and one for all the django IDAs

Because all the IDAs are coupled together, but the platform can be done separately, and it's the largest scarriest one.

coryleeio:
Go ahead and remove the commented renders as well, we can get them from git history if needed

coryleeio:
We should hold off on merging the config PR until the defaults have successfully shipped to production

coryleeio (approved):
This looks good - similar to the lms/cms PR we will want to make sure to push the defaults into the applications and ship them to production before merging the config which removes the defaults

Thank you for splitting them out.

syedimranhassan:
> This looks good - but I think you need to still remove edx_django_service role rendering.
> 
> I think you should separate this into two PRs as well, one for lms/cms
> 
> and one for all the django IDAs
> 
> Because all the IDAs are coupled together, but the platform can be done separately, and it's the largest scarriest one.

@coryleeio I think [veda](https://github.com/edx/configuration/blob/cf4d221d384ed396a3008c31487f385544ef08e7/playbooks/roles/veda_web_frontend/meta/main.yml#L38), blockstore using edx_django_service and we don't have remote config file for them. So, we should made remote config file for them before removing edx_django_service role rendering.

coryleeio:
I would avoid touching veda - is it possible to just add the render that we are removing their play so that we don't have to make it work with remote config?

coryleeio (approved):
I think we are fine to move this forward, it might be worth considering breaking this into different PRs, I don't think you need to get approval on each one, but that away if one of them breaks they can be rolled back independently without needing to roll back each application.  Up to you though, I'd also be fine with merging it and just rolling it all back if it breaks until we land it, as I expect it should be fine given that we have migrated the defaults and aren't using this config anymore?

coryleeio:
Actually yeah - lets try to land it given that you have a successful sandbox build, if it breaks and is difficult to land we will revert and apply them independently

bradenmacdonald:
@syedimranhassan the ansible scripts we use to deploy an Open edX instance with blockstore has started failing with this error:

```
TASK [edx_django_service : migrate database] ************************************************************
...
File \"/edx/app/blockstore/blockstore/blockstore/settings/production.py\", line 22, in <module>", "    with open(CONFIG_FILE) as f:", "FileNotFoundError: [Errno 2]
No such file or directory: '/edx/etc/blockstore.yml'
```

I think it's related to this PR, but it might be some other change. If it is this, is there some guidance on what variables need to be updated to work with this change?

coryleeio:
@bradenmacdonald Hey Braden, sorry about the lack of communication regarding this.  The idea is that we are hoping to be less prescriptive about how we do the configuration and that you don't need to pass in a secure and an internal config anymore, you just pass in the final product, and you can choose to merge or not merge your config or manage it however you see fit.

The idea is you should take one of your previously rendered configs and pass it in at build time:
Here is the ansible that handles copying the config:
https://github.com/edx/configuration/blob/2ff87c992464f1600a9a83ac585068fc60b0959b/playbooks/roles/edx_service/tasks/main.yml#L119-L129

So you would just need to set 
edx_service_copy_config_enabled=true
edx_service_local_config_file=some_local_file


Let me know if you still have issues regarding this.

bradenmacdonald:
@coryleeio Thanks for the info. That sounds nice and simple. We will probably have to change our tooling before I can test it out though, because we have a deployment system that generates a single consolidated set of variables and passes them in to ansible, but it doesn't currently have a way to also generate extra files like needed for `edx_service_local_config_file`. Or perhaps we'll look at adding a second option of setting the config file contents from an ansible/yaml variable, which would be easier for us to work with.

giovannicimolin:
@coryleeio I'm working on getting the master builds working again now, but it's being somewhat difficult to make it work given our current deployment system.
Since this change (removal of rendering from configuration) didn't go through any deprecation process, can we have a isolated role, that can be enabled from `edxapp` to render the configuration files? 
This will serve as a middle step in the migration to rendering the files before passing them to the playbooks. Do you think this is an acceptable solution?

bradenmacdonald:
@coryleeio Following up on what @giovannicimolin said, after discussing this with the team, I'm not sure these changes are going to be workable for us.

Consider that we maintain dozens and dozens of separate Open edX instances, mostly running the latest named release, but some running master and some occasionally running one release behind.

We can't just "take one of your previously rendered configs" because that doesn't apply to new instances which we spin up all the time. So we'd have to develop some logic for producing those config files from scratch, using a set of core variables - but this is exactly what `edx/configuration` was previously doing for us. We'd also have to maintain versions of those templates for each named release and for master, but again that was previously done by `edx/configuration` itself being versioned in step with the platform.

As a specific example, consider the SSO setup for each of the IDAs. When provisioning a new instance, it's necessary to use the platform API to generate a service user and OAuth credentials for each service, and then write some unique values like
```
SOCIAL_AUTH_EDX_OAUTH2_ISSUER: http://www.the.lms.example.com
SOCIAL_AUTH_EDX_OAUTH2_KEY: ecommerce-sso-key
SOCIAL_AUTH_EDX_OAUTH2_LOGOUT_URL: http://www.the.lms.example.com/logout
SOCIAL_AUTH_EDX_OAUTH2_SECRET: ecommerce-sso-secret
SOCIAL_AUTH_EDX_OAUTH2_URL_ROOT: http://www.the.lms.example.com
```
into that IDA's configuration file.

Previously, that was all done automatically and we have only to set a single ansible variable, the LMS domain name.

It doesn't much matter to us if the logic and templates for doing that sort of setup is in `edx/configuration` or in each IDA's own repository, but I don't think it makes sense for us to create a whole separate system of tooling just to generate such configs for each deployment and have separate versions of that for each named release.

Can we keep some mechanism for auto-generating the necessary parts of the configuration and then just allow merging in any overrides that people want?

I may be missing something here though; happy to hear your advice.

pomegranited:
@coryleeio @syedimranhassan CC @natabene  @bradenmacdonald  Has there been any further discussion or ideas on this topic?  It's a big deal for us, so we really need to find a compromise here.  It would also help to have visibility into the larger plan.

I can definitely see the benefits of removing all the service-specific variables out of the configuration repo.  But we still need to be able to easily configure hundreds of separate deployments, and don't want to create separate secure forks for each of them.  For example, for ecommerce, it would be enough for us to be able to set [`ecommerce_service_config_overrides`](https://github.com/edx/configuration/pull/5414/files#diff-dd4f3df8a0026e7f194781ec48f0338eL197) and have continue to be passed through to `edx_django_service_config_overrides` when generating the `ecommerce.yml` file.  Is there a way to retain this functionality, while still reducing maintenance on these playbooks?

bderusha:
@pomegranited @bradenmacdonald @giovannicimolin Sorry for the delays here.  Lots of moving parts around here in the past few weeks.  

I have put up a PR to revert this commit and the LMS/CMS rendered config PR here https://github.com/edx/configuration/pull/5723

I will be putting together an OEP to better communicate where edX is going operationally.  We can move the discussion that probably should have happened before these PRs there.

bradenmacdonald:
Sounds good, thanks for the update @bderusha!

pomegranited:
Thank you @bderusha !

giovannicimolin:
Thanks! This helps a lot @bderusha!

PR #5415, opened 2019-09-23
Run xqueue on python3

zainab-amir:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/xqueue/defaults/main.yml
+++ b/playbooks/roles/xqueue/defaults/main.yml
@@ -181,3 +181,6 @@ xqueue_debian_pkgs:
   - python-mysqldb
   # Needed to be able to build wheel for mysqlclient
   - python3-dev
+
+# flag to run xqueue on python3
+xqueue_use_python3: true
--- a/playbooks/roles/xqueue/tasks/main.yml
+++ b/playbooks/roles/xqueue/tasks/main.yml
@@ -1,4 +1,51 @@
 ---
+######## BEGIN PYTHON3 ########
+- name: install python3
+  apt:
+    name: "{{ item }}"
+  when: xqueue_use_python3
+  with_items:
+    - python3-pip
+    - python3-dev
+  tags:
+    - install
+    - install:system-requirements
+
+- name: build virtualenv with python3
+  command: "virtualenv --python=python3 {{ xqueue_venv_dir }}"
+  args:
+    creates: "{{ xqueue_venv_dir }}/bin/pip"
+  become_user: "{{ xqueue_user }}"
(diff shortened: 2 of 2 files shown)


feanil (changes requested):
You should not make this change in the edx_service role, which is meant to be a role that is shared by services that might not be python.  You should instead add this to the xqueue role and conditionally create a python3 xqueue venv on a python 2 xqueue venv.

I would look to making your change here: https://github.com/edx/configuration/blob/master/playbooks/roles/xqueue/tasks/main.yml

Ideally long term, xqueue would be using the edx_django_service role as its base instead of just the edx_service role but I think that doing that is too much work to get this over the line.

feanil (approved):

bderusha:
@zainab-amir based on an issue we ran into in another service you may also need to make the `install pip requirements` also switch between `pip` and `pip3` conditionally by adding the line `executable=pip3` to the requirements install

https://github.com/edx/configuration/pull/5415/files#diff-49cfef58ebeac4f9f324b8fdf8aee7afL66-L88

bderusha:
```
  name: install python3 requirements
  pip: >
    requirements="{{ item }}"
    virtualenv="{{ venv_dir }}"
    state=present
    executable=pip3
  sudo_user: "{{ user }}"
  when: use_python3
  with_items:
    - "{{ prerequirements_file }}"
    - "{{ requirements_file }}"
```

syed-awais-ali:
> ```
>   name: install python3 requirements
>   pip: >
>     requirements="{{ item }}"
>     virtualenv="{{ venv_dir }}"
>     state=present
>     executable=pip3
>   sudo_user: "{{ user }}"
>   when: use_python3
>   with_items:
>     - "{{ prerequirements_file }}"
>     - "{{ requirements_file }}"
> ```

It would lead to an error stating
`"msg": "parameters are mutually exclusive: ['executable', 'virtualenv']"`

The correct way is to remove the `executable` bit from this and add

`virtualenv_python="python{{ VENV_PYTHON_VERSION }}"`

where `VENV_PYTHON_VERSION` can be defined in the defaults.

```
- name: install python requirements
  pip: >
    requirements="{{ item }}"
    virtualenv="{{ venv_dir }}"
    virtualenv_python="python{{ VENV_PYTHON_VERSION }}"
    state=present
  sudo_user: "{{ user }}"
  with_items:
    - "{{ requirements_file }}"
```

OR an alternate way could be of what we have done in `edx_django_service` by removing the pip bit and adding the command and giving it the requirements file as the item (if there are multiple files to loop over)

feanil (approved):

PR #5418, opened 2019-09-24
WIP: Refactor python installs from jenkins_worker to jenkins_common

bmedx:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_common/tasks/main.yml
+++ b/playbooks/roles/jenkins_common/tasks/main.yml
@@ -1,21 +1,12 @@
 ---
+- include: python.yml
+
 - name: Install jenkins specific system packages
   apt:
     name: '{{ item }}'
     state: present
     update_cache: yes
   with_items: '{{ jenkins_common_debian_pkgs }}'
-  tags:
-    - jenkins
-    - install
-    - install:system-requirements
-
-- name: Install specific versions of python
-  apt:
-    name: '{{ item }}'
-    state: present
-    update_cache: yes
-  with_items: '{{ jenkins_common_python_versions }}'
   tags:
     - jenkins
     - install
--- a/playbooks/roles/jenkins_common/tasks/python.yml
+++ b/playbooks/roles/jenkins_common/tasks/python.yml
@@ -4,7 +4,7 @@
 # newer
 - name: add deadsnakes PPA for newer Python versions
   apt_repository:
-    repo: "ppa:fkrull/deadsnakes"
+    repo: "ppa:deadsnakes/ppa"
     update_cache: yes
   when: ansible_distribution_release == 'xenial'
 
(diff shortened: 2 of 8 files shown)


playbooks/roles/jenkins_common/tasks/main.yml
```diff
@@ -10,17 +12,6 @@
     - install
     - install:system-requirements
 
-- name: Install specific versions of python
-  apt:
-    name: '{{ item }}'
-    state: present
-    update_cache: yes
-  with_items: '{{ jenkins_common_python_versions }}'
-  tags:
```
bmedx, line 19: These tags are lost in the current set of changes, this is probably bad.

playbooks/roles/jenkins_common/tasks/python.yml
```diff
@@ -4,7 +4,7 @@
 # newer
 - name: add deadsnakes PPA for newer Python versions
   apt_repository:
-    repo: "ppa:fkrull/deadsnakes"
+    repo: "ppa:deadsnakes/ppa"
```
bmedx, line 7: This is supposedly the best PPA to use for deadsnakes these days.

playbooks/roles/jenkins_worker/tasks/main.yml
```diff
@@ -7,9 +7,9 @@
 # All jenkins workers
 - include: packer.yml
 - include: system.yml
-- include: python.yml
+- include: python_requests.yml
```
bmedx, line 10: I made an assumption here that `jenkins_worker` inherits from `jenkins_common` and will have those tasks run before this, I have no idea if that's actually true.

estute: It is not true. `jenkins_common` is the base role inherited by any play that sets up a jenkins master instance, i.e. `jenkins_build`. `jenkins_worker` is unrelated and is used to provision an ec2 box to be able to run as a jenkins worker. This means installing the jdk, docker, and in the case of running tests for the edx-platform, a pre-populated virtualenv.

playbooks/roles/jenkins_common/tasks/main.yml
estute: Can you conditionally `include` on the presence of tags?

playbooks/roles/jenkins_common/tasks/python.yml
estute: You may want to check w/ @feanil just to make sure this doesn't have implications on his team's python upgrade work

playbooks/roles/jenkins_worker/tasks/main.yml
bmedx: Hmm. So I guess that breaks this PR. Is there a better place to put common Python install code than where I stuck it in `jenkins_common`?

PR #5446, opened 2019-10-16
Fix the job definition for the jobs which app-permissions kicks off

matthugs:
Issue has caused e.g. this failure https://tools-edx-jenkins.edx.org/job/App-Permissions/job/app-permissions-runner-prod-edx/307/console


adzuci (approved):
I'm not sure where `bool` comes from, but it seems to have been there previously so maybe this will make the job run?  It likely needs to be changed further though.

PR #5450, opened 2019-10-21
add tasks to replace host to sandbox dns name in app config files

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_service/tasks/main.yml
+++ b/playbooks/roles/edx_service/tasks/main.yml
@@ -115,6 +115,20 @@
     - install:configuration
     - install:app-configuration
 
+- name: Replace deploy host to sandbox dns name
+  replace:
+    path: "{{ UNENCRYPTED_CFG_DIR }}/{{ edx_service_name }}.yml"
+    regexp: 'deploy_host'
+    replace: "{{ COMMON_DEPLOY_HOSTNAME }}"
+  when: edx_service_decrypt_config_enabled and COMMON_SANDBOX_BUILD
+  no_log: "{{ COMMON_CONFIG_NO_LOGGING }}"
+  become: false
+  delegate_to: localhost
+  tags:
+    - install
+    - install:configuration
+    - install:app-configuration
+
 - name: Copy app config file
   copy:
     src: "{{ edx_service_local_config_file }}"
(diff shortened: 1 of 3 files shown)


playbooks/roles/edx_service/tasks/main.yml
```diff
@@ -115,6 +115,18 @@
     - install:configuration
     - install:app-configuration
 
+- name: Replace deploy host to sandbox dns name
+  replace:
+    path: "{{ UNENCRYPTED_CFG_DIR }}/{{ edx_service_name }}.yml"
+    regexp: 'deploy_host'
+    replace: "{{ COMMON_DEPLOY_HOSTNAME }}"
+  when: edx_service_decrypt_config_enabled and COMMON_DEPLOY_HOSTNAME|length > 0
```
coryleeio: Can you add an explicit toggle to this and default it to false
I don't want someone to set COMMON_DEPLOY_HOSTNAME and have their yaml modified, especially given this is a hack for sandboxes.

Alternatively you could change 'COMMON_DEPLOY_HOSTNAME' to be more explicitly obvious that it will make changes to your yaml.  Maybe by appending 'OVERRIDE' or even 'HACK'.  I don't want people to think that this is a value they should set in production and to wind up having odd behaviour because they set it.

coryleeio (changes requested):

playbooks/roles/edx_service/tasks/main.yml
nadeemshahzad: Used ```COMMON_SANDBOX_BUILD```  which is true in ansible provision script. https://github.com/edx/configuration/blob/270149c67cfc3a529c8c7ee81de25ad043cda082/util/jenkins/ansible-provision.sh#L223

coryleeio (approved):
Looks fine to me, I don't see you passing in COMMON_SANDBOX_BUILD though

nadeemshahzad:
@coryleeio it is already defined here in ansible-privsion script (https://github.com/edx/configuration/blob/e693ea8d6f98c0b1184d7de5f72d95de41d538d2/util/jenkins/ansible-provision.sh#L223 ) that we are using  to build sandboxes  and passed in extra vars.

PR #5454, opened 2019-10-22
Bump labxchange-xblocks version on edx.org

bradenmacdonald:
Pulls in a minor change to the LabXchange Image XBlock: https://github.com/open-craft/labxchange-xblocks/pull/3

Only used on edx prod/stage.


ormsbee (approved):

ormsbee (approved):

openedx-webhooks:
Although this pull request is already merged, I've created [OSPR-5394](https://openedx.atlassian.net/browse/OSPR-5394) so that we can track it in Jira.

There is nothing you have to do. No action is needed from your side.  Thanks again for your contribution.

PR #5456, opened 2019-10-22
Update Django service node version.

dianakhuang:
Configuration Pull Request
---


dianakhuang:
It looks like the main thing that needs to happen is that node-sass needs to be upgraded to 4.12+ to support Node 12.

dianakhuang:
I will attempt to merge this update independent of credentials and discovery and open separate PRs for updating those IDAs separately.

adzuci (approved):
:shipit: If we forgot to override it anywhere this may cause some issues, but I think they'll be obvious.

Might be helpful to list the edX things that use this and the places where it's overridden to let other people know when you merge this if you want to be comprehensive about this.

PR #5465, opened 2019-10-23
Refactor node versions for discovery and credentials.

dianakhuang:
Refactoring out the node versions for discovery and credentials so that they can be updated separately.

This is not a complete refactor, but it does separate out two of the individual IDAs so that https://github.com/edx/configuration/pull/5456 can be merged separately.

### DIFF
--- a/playbooks/roles/credentials/defaults/main.yml
+++ b/playbooks/roles/credentials/defaults/main.yml
@@ -21,6 +21,7 @@ credentials_environment:
   CREDENTIALS_CFG: '{{ COMMON_CFG_DIR }}/{{ credentials_service_name }}.yml'
 
 credentials_gunicorn_port: 8150
+credentials_node_version: '8.9.3'
 
 #
 # OS packages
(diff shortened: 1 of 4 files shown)


robrap (commented):
In general, this PR looks good to me.

I think it would be useful to have @coryleeio or someone on devops look at this.  I am not sure if it makes sense to update a default in https://github.com/edx/configuration/pull/5456, or if that should be 'REPLACE_ME' (until the config work is complete), and becomes clear that all apps should define their own value.  Basically, I don't see when and where a shared value is ever a benefit here.

dianakhuang:
I am currently of the opinion of slowly splitting out the individual IDAs when relevant, especially since some of them don't care about node at all (see: analytics api). I can stand to be convinced, though.

robrap (approved):
👍 I don’t think there is anything wrong with this PR, but I don’t know how to verify that. That was the main reason for stating you may want a second pair of eyes, but I’ll leave that up to you.

Regarding the other PR, that too is up to you. I think the ultimate plan is to no longer have shared config settings across apps, but Devops is already working on that, so feel free to do what makes sense to you, including discussing with devops or not. 🙂

coryleeio (approved):

PR #5479, opened 2019-10-29
Decouple applications that do not have remote config from edx_service's ansible config render

syedimranhassan:
Configuration Pull Request
---


syedimranhassan:
@coryleeio FYI

- There is one role (veda_web_frontend) that had dependencies on edx_django_service and don't have remote config enabled. So, I have moved edx_django_service (variables, dependencies, tasks and templates) to veda_web_frontend role to remove it's dependencies on edx_django_service role.

- For edx_service role there were multiple roles that had dependencies on edx_service role. So, I made a separate role (edx_rendering_service) to remove their dependencies on edx_service.

**I have set new role name edx_renering_service. You may suggest any other if it doesn't sounds good.**

coryleeio (changes requested):
This seems reasonable - is it possible add edx_service as a meta role of  from edx_rendering_service to avoid the duplication?

Could you also rename edx_rendering_service to edx_service_with_rendered_config?

I don't want someone to think this service is actually doing any rendering.

coryleeio:
I am wondering if we could avoid duplicating all of the edx_django_service role specific stuff into the veda_frontend role by creating a  edx_django_service_role_with_rendered_config role which calls edx_django_service and appends the render.

Then we wouldn't need to decouple veda from edx_django_service to edx_service, as that is a much larger change.

syedimranhassan:
@coryleeio I have added new role **edx_django_service_with_rendered_config** which calls **edx_service_with_rendered_config** and this calls **edx_service**
The flow will be this
**edx_django_service_with_rendered_config** (contains edx_django logic) --> **edx_service_with_rendered_config** (contains rendering logic only and calls edx_service) --> **edx_service** (perform further actions)

coryleeio (approved):

coryleeio:
This looks right thanks!

PR #5492, opened 2019-11-05
Fix zombie workers

jdmulloy:
Having this set to 5 days is causing us a lot of pain with workers not shutting down. Looking at the history it was put in place to prevent grading tasks from getting killed. I suspect the issue might not exist any more.

Configuration Pull Request
---


adzuci (approved):
:shipit: 1 hour sounds good, maybe we can make the Jenkins job not fail for an hour to sync these things up too?  cc @feanil 

Context: https://github.com/edx/configuration/pull/1410/files

kdmccormick:
@jdmulloy what does this setting control? is it how long a task is allowed to run before getting killed?

If so, that's fine for Registrar, because none of its tasks should take longer than a minute.

I don't know about grading tasks, though; @iloveagent57 might know more.

jdmulloy:
@kdmccormick This only comes into play when shutting down a server. When we shutdown supervisro sends the celery worker process a sigterm and if it still hasn't shut down after ```stopwiatsecs``` it sends a sigkill. We should check how long your tasks run. edx-platform actually has some processes that can run for 19+ hours. I figured this setting was just copied when the registrar workers were created and it may not be the setting you want.

kdmccormick:
@jdmulloy got it. yeah, Registrar tasks shouldn't last longer than a minute, and also they will not corrupt anything if they are killed (they are just aggregating the results of several individually-safe LMS API calls), so definitely feel free to lower Registrar's `stopwaitsec`.

It's just that I can't say anything about grading tasks, as they aren't related to Registrar.

fredsmith (changes requested):
I think this will cause issues with long running grading tasks.  I think we've resolved the failure we were seeing in monitoring without having to do this.

kdmccormick:
@fredsmith @jdmulloy is this safe to close?

fredsmith:
Closing this as we're no longer being impacted by long running tasks

PR #5530, opened 2019-11-21
add static site routing to prospectus

fredsmith:
AKA:  [You got conductor in my prospectus](https://www.youtube.com/watch?v=O7oD_oX-Gio)


adzuci:
I feel like you have a better memory of memes and youtube than I do.

Also, I don't know what I don't know about this, the URLs in https://github.com/edx/edx-internal/pull/1278/files look reasonable, but I don't quite know how I would test if this code change does what we expect.

fredsmith:
@adzuci that's why I'm only shipping to stage.

tuchfarber (approved):
I think some assets might not load correctly due to routing, but it's just for stage, so I'm not super concerned. Conductor ego aside, I think putting routing things in a place called prospectus isn't the right move. Prospectus should be one of the many items that routing knows about, it shouldn't own the routing. 

So I'm approving this as a way for us to test in stage, but I wouldn't approve a similar PR to prod.

fredsmith:
@tuchfarber unfortunately, since we're co-opting prospectus's namespace we kind of have to play in their namespace.  They can't give us a set of routes to push into conductor.  

I guess the other option is to add a "fallback" service to conductor, and jam conductor between users and prospectus, but I don't think we'd get buy-in from the website team for that.

tuchfarber:
I guess my point is less about using conductor vs prospectus, but more about the routing layer living inside prospectus. If there's a box that listens to `www.edx.org` and we want to route other things on that box, then we should make that box just about routing and not have it also be the marketing site. That's why I'm fine with this as a POC, but ultimately I think prospectus and it's associated routing should be split up. 

That project would be easier when we didn't have to `try_files` and deal with prospectus + drupal, but I'm not sure if there's an ETA on drupal removal yet. 

(Begin ad-hoc idea: ) If we didn't want to wait on drupal removal to start building, we could also point a temp DNS name at a router laying, make sure everything works including prospectus routing, then flip `www` over to it when we're ready.

PR #5544, opened 2019-12-02
Update xblocks to support python3.

feanil:


jmbowman (approved):
I think this just leaves `xblock-problem-builder` left to be updated?  Current status at https://github.com/open-craft/problem-builder/pull/252

PR #5550, opened 2019-12-04
Increase upload limit for Blockstore to 10M

ormsbee:
@feanil: The Blockstore upload PR as discussed a week ago. Is that all I have to change?


feanil (approved):
Yep, that should be it.

PR #5551, opened 2019-12-05
SE-1787 Add mongo backups

samuelallan72:
This adds support for automated backups on a mongo db node to an EBS volume.

This has been used successfully on a client's mongo cluster. See https://github.com/edx-olive/configuration/pull/14 The way it works is to have a mongo node designated as the `MONGO_BACKUP_NODE`. On this node is a cron job that periodically launches the backup script. This backup script does roughly the following:

- mounts an ebs volume
- deletes all the files
- dumps the mongo database with mongodump
- unmounts the ebs volume
- creates a snapshot of the volume
- prunes old snapshots of the volume

This allows for flexible, consistent backups, versioned with ebs snapshots on aws.

This PR differs from the revision used on the client in that it is disabled by default, and backup related tasks will only run if `MONGO_BACKUP_ENABLED` is true *and* the current host is the `MONGO_BACKUP_NODE`.

**JIRA tickets**: [OSPR-3954](https://openedx.atlassian.net/browse/OSPR-3954)

**Dependencies**: None

**Merge deadline**: None

**Testing instructions**:

- Note that steps 1 to 3 were thoroughly tested in the context of  https://github.com/edx-olive/configuration/pull/14. This PR only differs in terms of extra variables and conditionals to disable this functionality.

1. set appropriate vars for the `MONGO_BACKUP_*` vars found in
   `playbooks/roles/mongo_3_2/defaults/main.yml`
2. deploy a mongo node on aws
3. verify that the backup script runs on the `MONGO_BACKUP_NODE` instance.
4. set `MONGO_BACKUP_ENABLED` to false
5. deploy a mongo node
6. verify that none of the backup tasks from `playbooks/roles/mongo_3_2/tasks/main.yml` were run, and that the deployed mongo_logrotate.j2 template does not include the block for the mongo-backup.log.

See https://github.com/edx-olive/configuration/pull/17, where we show the results of live tests. Note that this was against a different branch than master since it was required by the client.

**Author notes**:

**Reviewers**
- [x] @itsjeyd
- [ ] edX reviewer[s] TBD

### DIFF
--- a/playbooks/roles/mongo_3_2/defaults/main.yml
+++ b/playbooks/roles/mongo_3_2/defaults/main.yml
@@ -107,4 +107,3 @@ mongo_backup_cron:
   day: '*'
   month: '*'
   weekday: '*'
-is_backup_node: false
--- a/playbooks/roles/mongo_3_2/defaults/main.yml
+++ b/playbooks/roles/mongo_3_2/defaults/main.yml
@@ -107,3 +107,4 @@ mongo_backup_cron:
   day: '*'
   month: '*'
   weekday: '*'
+is_backup_node: false
--- a/playbooks/roles/mongo_3_2/defaults/main.yml
+++ b/playbooks/roles/mongo_3_2/defaults/main.yml
@@ -86,3 +86,20 @@ mongo_dbpath: "{{ mongo_data_dir }}/mongodb"
 mongo_enable_journal: true
 
 MONGO_LOG_SERVERSTATUS: true
+
+# backups vars. These must all be set according to your mongo instance setup
+MONGO_BACKUP_EBS_VOLUME_DEVICE: ""
+MONGO_BACKUP_EBS_VOLUME_ID: ""
+MONGO_BACKUP_VOLUME_MOUNT_PATH: "/mnt/mongo-backup"
+MONGO_BACKUP_NODE: "" # note: most likely the ip address of the instance on which to perform the backups
+MONGO_BACKUP_AUTH_DATABASE: ""
+MONGO_BACKUP_SNAPSHOT_DESC: "mongo-backup"
+MONGO_BACKUP_PRUNE_OLDER_THAN_DATE: ""  # passed to `date -d`; should be a relative date like "-30days"
+MONGO_BACKUP_SNITCH_URL: "" # Optional URL that will be used to ping a monitoring service (such as Dead Man's Snitch) upon successful completion of a backup.
+mongo_backup_script_path: "/usr/local/sbin/backup-mongo.sh"
+mongo_backup_cron:
+  minute: '12'
+  hour: '*/12'
+  day: '*'
+  month: '*'
+  weekday: '*'
--- a/playbooks/roles/mongo_3_2/defaults/main.yml
+++ b/playbooks/roles/mongo_3_2/defaults/main.yml
@@ -107,3 +107,7 @@ mongo_backup_cron:
   day: '*'
   month: '*'
   weekday: '*'
+
+# Internal variable set to true dynamically if backups enabled and playbook running on MONGO_BACKUP_NODE. Do not
+# manually override.
+is_backup_node: false
--- a/playbooks/roles/mongo_3_2/defaults/main.yml
+++ b/playbooks/roles/mongo_3_2/defaults/main.yml
@@ -87,15 +87,19 @@ mongo_enable_journal: true
 
 MONGO_LOG_SERVERSTATUS: true
 
-# backups vars. These must all be set according to your mongo instance setup
+# Vars for configuring a mongo backup node. If enabled, this node will be provisioned with a script that uses mongodump
+# to backup the database to an ebs volume at a period set by mongo_backup_cron.
+# Set MONGO_BACKUP_ENABLED to true to enable. If enabled, all the other MONGO_BACKUP_ vars must be set according to your
+# setup.
+MONGO_BACKUP_ENABLED: false
+MONGO_BACKUP_NODE: "" # note: most likely the ip address of the instance on which to perform the backups
 MONGO_BACKUP_EBS_VOLUME_DEVICE: ""
 MONGO_BACKUP_EBS_VOLUME_ID: ""
-MONGO_BACKUP_VOLUME_MOUNT_PATH: "/mnt/mongo-backup"
-MONGO_BACKUP_NODE: "" # note: most likely the ip address of the instance on which to perform the backups
 MONGO_BACKUP_AUTH_DATABASE: ""
-MONGO_BACKUP_SNAPSHOT_DESC: "mongo-backup"
 MONGO_BACKUP_PRUNE_OLDER_THAN_DATE: ""  # passed to `date -d`; should be a relative date like "-30days"
 MONGO_BACKUP_SNITCH_URL: "" # Optional URL that will be used to ping a monitoring service (such as Dead Man's Snitch) upon successful completion of a backup.
+MONGO_BACKUP_VOLUME_MOUNT_PATH: "/mnt/mongo-backup"
+MONGO_BACKUP_SNAPSHOT_DESC: "mongo-backup"
 mongo_backup_script_path: "/usr/local/sbin/backup-mongo.sh"
 mongo_backup_cron:
   minute: '12'
--- a/playbooks/roles/mongo_3_2/tasks/main.yml
+++ b/playbooks/roles/mongo_3_2/tasks/main.yml
@@ -172,19 +172,6 @@
     - "manage:db-replication"
     - "mongodb_key"
 
-- name: install prereqs for backup script
-  apt:
-    pkg: "{{ item }}"
-    state: present
-    update_cache: yes
-  with_items:
-    - jq
-  tags:
-    - "backup:mongo"
-    - "install"
-    - "install:app-requirements"
-    - "mongo_packages"
-
 # If skip_replica_set is true, this template will not contain a replica set stanza
 # because of the fact above.
 - name: copy configuration template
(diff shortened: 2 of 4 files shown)


openedx-webhooks:
Thanks for the pull request, @swalladge! I've created [OSPR-3954](https://openedx.atlassian.net/browse/OSPR-3954) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

playbooks/roles/mongo_3_2/tasks/main.yml
```diff
@@ -172,6 +172,20 @@
...
     - "mongodb_key"
 
+- name: install prereqs for backup script
+  apt:
+    pkg: "{{ item }}"
+    state: present
+    update_cache: yes
+  with_items:
+    - jq
+  when: "'{{ ansible_default_ipv4.address|default(ansible_all_ipv4_addresses[0]) }}' == '{{ MONGO_BACKUP_NODE }}'"
```
itsjeyd: @swalladge At first glance this is a bit of a mouthful :) I think it would be worth adding an inline comment that explains or paraphrases this check in regular English.

Also, IIUC the changes from 5a45464dbb46c5e76bbadee953a7074ad2370fac (which this line belongs to) don't make setting up a backup node optional? (**EDIT**: This line just checks whether the current server is the designated `MONGO_BACKUP_NODE`, right?) We should probably do that by introducing a separate variable (called something like `MONGO_PROVISION_BACKUP_NODE`) that defaults to `False`, to allow people to choose whether they want to use the backup solution from this role or a different one, and check the value of that variable here.

If we had such a variable, would we still need the check above? (Genuine question. Things were working fine without it when deploying https://github.com/edx-olive/configuration/pull/14.)

itsjeyd (changes requested):
@swalladge Thanks for the PR! I think we should make the backup node functionality opt-in (cf. below). Other than that the changes look good to me -- for the most part they were already reviewed and tested via https://github.com/edx-olive/configuration/pull/14. (Could you please add a reference to that PR to the PR description?)

natabene:
@swalladge Thank you for your contribution, please let me know once it is ready for our review.

playbooks/roles/mongo_3_2/tasks/main.yml
samuelallan72: > If we had such a variable, would we still need the check above?

Technically, no - the backup script already does it's own check to make sure it's running on the correct node. However, it could be nice just so that the backup script items are only pushed to the mongo node.

Also adding that check to all the things, effectively makes it opt in. If `MONGO_BACKUP_NODE` is empty, then all those checks will fail and nothing will be run.

Do you think it would be cleaner to add a separate variable solely for enabling/disabling the extra backup entries?

itsjeyd: @swalladge Thanks for the details.

> Do you think it would be cleaner to add a separate variable solely for enabling/disabling the extra backup entries?

It might be a matter of personal preference, but I think I do 😄 

The thing that doesn't feel quite right (to me) with the current approach is that it's using a single check to answer two questions (which are related but still different):

1. Is the backups feature enabled?
1. Is the current server the designated `MONGO_BACKUP_NODE`?

A tangible effect of this is that if I were looking at the list of variables related to the backups feature, it wouldn't be clear to me (at least not immediately/without further research) what I would need to do if I didn't want to provision a backup node.

This would be different if there were a separate "feature-flag" variable for enabling the backups feature (and if [this comment](https://github.com/edx/configuration/pull/5551/files#diff-772269a3d5f75dd1b0d2f0ba5543c6e0R90) were modified to indicate that the variables below it will only need to be set if the backups feature is enabled).

Does that make sense?

>> If we had such a variable, would we still need the check above?
>
> Technically, no - the backup script already does it's own check to make sure it's running on the correct node. However, it could be nice just so that the backup script items are only pushed to the mongo node.

Yep, OK to keep that check then.

samuelallan72: Yep that's fair.

samuelallan72: @itsjeyd  variable added. The only thing that currently will be deployed if MONGO_BACKUP_ENABLED is false is the logrotate entry (I couldn't get it working with the ansible variables and debugging that is a nightmare) - I can try again if time isn't an issue, but this appears low impact (logrotate shouldn't do anything extra if the backup log files aren't created in the first place?).

I tested the current setup for syntax and basic conditionals working, but this will definitely need double checking since ansible variables are a pain to get right.

playbooks/roles/mongo_3_2/defaults/main.yml
```diff
@@ -86,3 +86,24 @@ mongo_dbpath: "{{ mongo_data_dir }}/mongodb"
 mongo_enable_journal: true
 
 MONGO_LOG_SERVERSTATUS: true
+
+# Vars for configuring a mongo backup node. If enabled, this node will be provisioned with a script that uses mongodump
+# to backup the database to an ebs volume at a period set by mongo_backup_cron.
+# Set MONGO_BACKUP_ENABLED to true to enable. If enabled, all the other MONGO_BACKUP_ vars must be set according to your
+# setup.
```
itsjeyd, line 93: @swalladge Thanks for adding this info! 🙂

playbooks/roles/mongo_3_2/tasks/main.yml
```diff
@@ -187,15 +187,78 @@
     - "manage:db-replication"
     - "update_mongod_conf"
 
+# this sets the enable_mongo_backups var if mongo backups are enabled AND we're currently running against the designated
+# mongo backup node.
```
itsjeyd: @swalladge Thanks for documenting the purpose of this variable :)

Nit: `this` should be `This`.

I'm wondering if there is a way to rename `enable_mongo_backups` that makes the comment above seem less redundant. How about something like `is_backup_node` or `configure_backup_script`?

Also, as it's currently written this comment mostly focuses on the *what* rather than the *why* of the task below. What do you think about the following version:

```
# This sets the is_backup_node/configure_backup_script var by checking whether 
# mongo backups are enabled AND we're currently running against the designated mongo backup node.
# This allows backup-related tasks below to determine whether or not they should on the current mongo node.
```

playbooks/roles/mongo_3_2/tasks/main.yml
```diff
@@ -187,15 +187,78 @@
...
+- name: check mongo backups enabled
```
itsjeyd: @swalladge Building on the suggestions from the previous comment, this could be `Determine if backup-related tasks should be run`.

playbooks/roles/mongo_3_2/tasks/main.yml
itsjeyd: @swalladge Thanks for the updates!

> The only thing that currently will be deployed if MONGO_BACKUP_ENABLED is false is the logrotate entry ...

Not completely sure if the following would work but have you tried setting a default for `include_backup_log` in playbooks/roles/mongo_3_2/defaults/main.yml and then updating it in playbooks/roles/mongo_3_2/tasks/main.yml (instead of defining it for the first time via `install logrotate configuration`)?

It would be great to exclude the backup-specific logrotate configuration from any node that's not a backup node but you're right that we shouldn't spend too much time on this if it proves difficult. (We don't have unlimited budget to spend on this task.)

itsjeyd (changes requested):
@swalladge This is mostly ready for testing :) See my comments below for a couple small suggestions to apply before moving forward with that.

As I mentioned [here](https://github.com/edx/configuration/pull/5551#discussion_r355967615), if we could find a way to exclude the logrotate config from nodes that aren't backup nodes, that would be great but we should generally limit the amount of time spent researching this.

samuelallan72:
@itsjeyd comments addressed. :)

playbooks/roles/mongo_3_2/tasks/main.yml
samuelallan72: Turns out that the problem was that ansible wasn't actually evaluating that expression where it was defined, so it was evaluating on the fly in the `when` clauses, but the template must have been intepreting it as a string. Now fixed. :)

itsjeyd: @swalladge Ah, OK :) Thanks for figuring that out.

It looks like we'll need to set a default for `is_backup_node` (or update otherwise update the implementation to make sure that variable is always defined) after all -- some of the tests are failing with:

```
TASK [mongo_3_2 : install logrotate configuration] *****************************

fatal: [127.0.0.1]: FAILED! => {"changed": false, "msg": "AnsibleUndefinedVariable: 'is_backup_node' is undefined"}
```

itsjeyd:
@swalladge Thanks! The changes look good. A couple tests will need to be fixed (cf. https://github.com/edx/configuration/pull/5551#discussion_r356480003) but other than that this is ready for verification.

samuelallan72:
@itsjeyd tests are green and test instructions updated in the PR description. :smile:

itsjeyd (approved):
@swalladge This is ready to go :+1:

- [x] I tested this: (via https://github.com/edx-olive/configuration/pull/17, cf. https://github.com/edx-olive/configuration/pull/17#pullrequestreview-332445445)
- [x] I read through the code
- [x] ~~I checked for accessibility issues~~ N/A
- [x] Includes documentation

samuelallan72:
@natabene This is ready for edX review :smile:

natabene:
@swalladge Thanks for letting me know. I will queue it.

fredsmith (approved):
this looks good to me

openedx-webhooks:
@swalladge 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/configuration&entry.867055334=https://github.com/edx/configuration/pull/5551&entry.1484655318=https://github.com/swalladge&entry.752974735=2019-12-05+04:25&entry.1917517419=2020-01-24+20:22&entry.2133058324=Yes) so we can improve your experience in the future.

samuelallan72:
Oh sorry, I just realized the commits weren't squashed. :/

PR #5555, opened 2019-12-05
Upgrade and install build jenkins plugins

jdmulloy:
Configuration Pull Request
---


adzuci (approved):
:shipit:

playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -138,6 +138,9 @@ build_jenkins_plugins_list:
   - name: 'javadoc'
     version: '1.3'
     group: 'org.jenkins-ci.plugins'
+  - name: 'jobConfigHistory'
+    version: '2.24'
+    group: 'org.jenkins-ci.plugins'
```
estute, line 143: ^ Why are we installing this plugin? We used to have it installed, but decided on using the job dsl plugin instead

jdmulloy: @estute I want it so I can see changes to jobs. A while ago I looked at jobs and they were showing up as changed since seeding and I had no way of knowing what the changes were. There's also an issue with jobs not always linking to the seed job, but I suspect that's because the one seed job seeds all the jobs based on input so it gets disconnected. I'm not sure I like the manual seed process with the one job. On tools Jenkins it automatically seeds when the repo is updated and I can always trace a job back to its seed.

estute: That's fair.

PR #5561, opened 2019-12-09
Configure prospectus basic auth with separate username and password, …

reillz10:
This pull request is a POC to configure a custom username/password for prospectus preview since this is something that will be shared with people external to edX. We developing this as a temporary solution and might sunset it as early as 1/31/2020. We have a thumbs up from the folks in the #security-engineering channel to move forward with this plan for the time being.

This PR will be merged after: 
https://github.com/edx/edx-internal/pull/1326


adzuci (approved):
This looks correct to me, but I welcome someone else taking a quick look and want to mention that prospectus doesn't currently auth users against the LMS like our other IDAs do which makes it difficult to use groups for the purpose of exposing a preview endpoint and that this work is intended to be temporary and prove out a replacement for an existing drupal feature.

reillz10:
> This looks correct to me, but I welcome someone else taking a quick look and want to mention that prospectus doesn't currently auth users against the LMS like our other IDAs do which makes it difficult to use groups for the purpose of exposing a preview endpoint and that this work is intended to be temporary and prove out a replacement for an existing drupal feature.

Thats right. It doesn't currently auth against LMS or Discovery but it will need to in order to stand up a dynamic preview feature that we will be building. Basically, we will be authenticating against disco in order to pull down fresh new unpublished pages live https://openedx.atlassian.net/jira/software/projects/WEBSITE/boards/562/backlog?selectedIssue=WEBSITE-306&text=dynamic

feanil (commented):
This will work but a simpler way to do this without adding prospectus specific tasks and vars here is to just pass in the `PROSPECTUS_PREVIEW_NGINX_USERS` as the value of `NGINX_USERS` and setting `NGINX_CREATE_HTPASSWD_FILE` to `true` in the prospectus `meta/main.yml`

This will let you put the prospectus related defaults in the prospectus role.

See https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-dependencies for more info on how that would work.

This is not meant to block this work, I know this is temporary so feel free to merge knowing that we'll be removing it in a few weeks but just FYI for the future.

reillz10:
> This will work but a simpler way to do this without adding prospectus specific tasks and vars here is to just pass in the `PROSPECTUS_PREVIEW_NGINX_USERS` as the value of `NGINX_USERS` and setting `NGINX_CREATE_HTPASSWD_FILE` to `true` in the prospectus `meta/main.yml`
> 
> This will let you put the prospectus related defaults in the prospectus role.
> 
> See https://docs.ansible.com/ansible/latest/user_guide/playbooks_reuse_roles.html#role-dependencies for more info on how that would work.
> 
> This is not meant to block this work, I know this is temporary so feel free to merge knowing that we'll be removing it in a few weeks but just FYI for the future.

Thank you @feanil! I appreciate your help and the advise. I'll be sure to keep that in mind for the next one.

PR #5562, opened 2019-12-10
Turns out that port_in_redirect off needs to be added to any new loca…

reillz10:
…tions if you don't want the port to show up in a redirect.


rgraber (approved):
LGTM

albemarle (approved):

PR #5583, opened 2019-12-19
Build edx-platform assets to both local disk and S3 as part of transi…

cpennington:
…tioning to S3 backed assets

Configuration Pull Request
---


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1716,3 +1716,5 @@ EDXAPP_ENABLE_DJANGO_ADMIN_RESTRICTION: false
 
 EDXAPP_LMS_LOCAL_CONFIG_FILE: "{{ UNENCRYPTED_CFG_DIR }}/lms.yml"
 EDXAPP_CMS_LOCAL_CONFIG_FILE: "{{ UNENCRYPTED_CFG_DIR }}/studio.yml"
+
+edxapp_staticfiles_storage_overrides: !!null
```
adzuci, line 1720: Could/should this be uppercase?

adzuci (approved):
:shipit:

playbooks/roles/edxapp/defaults/main.yml
cpennington: I'd argue no, because it isn't an externally exposed variable... But that's a bit tenuous. We can uppercase it later, if we want.

PR #5595, opened 2020-01-03
Automated user should not be added to ssh allow users, as it could be a real user, and is causing openedx users to be locked out when doing native installs.

coryleeio:


nedbat:
These two commits fixed the problem for Open edX installs, thanks.

nedbat (approved):

PR #5607, opened 2020-01-09
Fix JWT signing data setup for sandboxes

matthugs:
- make sure sandbox-hood is passed through correctly to all the idas
- move JWT role invocation down a layer of ansible abstraction, so we
  can be more sure that it will be generated just after the edxapp
  role

JIRA:EDUCATOR-4812


matthugs:
Along with https://github.com/edx/sandbox-internal/pull/70, I built two sandboxes successfully: https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/26413/ and https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/26417/. I don't think this runs the full gamut of the testing checklist on confluence, but my hope is that it'll be enough.

adzuci (approved):
:shipit:

PR #5624, opened 2020-01-30
[PROD-1189] -  Update to ubcpi-xblock package.

saadyousafarbi:
Configuration Pull Request
---
Update UBCPI version after package change.

### Description
UBCPI (UBC Peer Instruction Tool for edX) is third-party package used by EDX for Peer instruction block. A recent ticket https://openedx.atlassian.net/browse/PROD-1173 was reported in which the display name was not being updated while editing this block. 
Upon investigation, it was noted that this third-party package expected a value (i.e. corrent_rationale)  at all times when updating/editing this block. In the reported case, the field Explanation which is OPTIONAL (responsible for correct_rationale value) was empty. As a result, whenever any field was updated on the block with this empty value, the block was not updated. 
This scenario occurs when we have a peer instruction block with the optional field (because n/a correct answer selected) and the course is exported and then imported back in.

### Sandbox
https://studio-saadyousafarbi.sandbox.edx.org/course/course-v1:edX+DemoX+Demo_Course

### Reviewers
- [ ] @awaisdar001 
- [ ] @DawoudSheraz 

#### UBCPI Pull Request: https://github.com/edx/ubcpi/pull/1

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -526,7 +526,7 @@ EDXAPP_PRIVATE_REQUIREMENTS:
       extra_args: -e
     # Peer instruction XBlock
     # Need it from github until we can land https://github.com/ubc/ubcpi/pull/167 upstream.
-    - name: git+https://github.com/edx/ubcpi.git@3c4b2cdc9f595ab8cdb436f559b56f36638313b6#egg=ubcpi-xblock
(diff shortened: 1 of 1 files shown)


DawoudSheraz (commented):
Please update the following:
1. PR title
2. PR description, describing why are we updating ubcpi xblock
   - Add a link to your merged PR

adzuci (approved):
:shipit:

DawoudSheraz (approved):

PR #5628, opened 2020-01-31
Add Free Text Response XBlock

ormsbee:
The fourth of four Stanford developed XBlocks we're enabling.


abutterworth (approved):

itsjeyd:
@ormsbee Great to see this XBlock being enabled by default :)

Do you happen to know if there is a specific version of the XBlock that would be compatible with Ginkgo?

PR #5631, opened 2020-02-05
Add back registrar configs which trigger Oauth backend setup

matthugs:
This partially reverts some changes in 00e2d127df81837ad1bd2e7d0de09d114be3a8e1 to ensure that the `oauth_client_setup` role sets up configurations which would enable registrar to request auth tokens from the LMS. (See https://github.com/edx/configuration/blob/master/playbooks/roles/oauth_client_setup/defaults/main.yml and https://github.com/edx/configuration/blob/master/playbooks/roles/oauth_client_setup/tasks/main.yml to understand why these variables are necessary to trigger this setup.)

We should discuss whether a more thorough revert is necessary/how to avoid this issue without blocking @syedimranhassan's current work. But I'm putting up this PR because I'd like to be unblocked building master's integration sandboxes in the interim.

JIRA:EDUCATOR-4865

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/registrar/defaults/main.yml
+++ b/playbooks/roles/registrar/defaults/main.yml
@@ -72,6 +72,10 @@ REGISTRAR_VERSION: 'master'
 REGISTRAR_GUNICORN_EXTRA: ''
 
 # Used to automatically configure OAuth2 Client
+REGISTRAR_SOCIAL_AUTH_EDX_OAUTH2_KEY: 'registrar-sso-key'
+REGISTRAR_SOCIAL_AUTH_EDX_OAUTH2_SECRET: 'registrar-sso-secret'
+REGISTRAR_BACKEND_SERVICE_EDX_OAUTH2_KEY: 'registrar-backend-service-key'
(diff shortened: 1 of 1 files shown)


playbooks/roles/registrar/defaults/main.yml
```diff
@@ -72,6 +72,10 @@ REGISTRAR_VERSION: 'master'
 REGISTRAR_GUNICORN_EXTRA: ''
 
 # Used to automatically configure OAuth2 Client
+REGISTRAR_SOCIAL_AUTH_EDX_OAUTH2_KEY: 'registrar-sso-key'
```
adzuci, line 75: Just curious, what does registrar do with social oauth stuff?

adzuci (approved):
:shipit:

syedimranhassan (approved):
LGTM

playbooks/roles/registrar/defaults/main.yml
matthugs: Dunno, the backend service is the operative blocker for us

But I suspect logging in to, say, django admin may be downstream of this

PR #5635, opened 2020-02-07
Remove OIDC values and DOP references.

dianakhuang:
https://openedx.atlassian.net/browse/BOM-1215

Internal PR: https://github.com/edx/edx-internal/pull/1497
Edge PR: https://github.com/edx/edge-internal/pull/201
Sandbox PR: https://github.com/edx/sandbox-internal/pull/75

Configuration Pull Request
---


playbooks/roles/edx_django_service_with_rendered_config/defaults/main.yml
```diff
@@ -191,15 +184,6 @@ edx_django_service_with_rendered_config_service_config_default:
   CACHES: '{{ edx_django_service_with_rendered_config_caches }}'
   DATABASES: '{{ edx_django_service_with_rendered_config_databases }}'
 
-  SOCIAL_AUTH_EDX_OIDC_KEY: '{{ edx_django_service_with_rendered_config_social_auth_edx_oidc_key }}'
-  SOCIAL_AUTH_EDX_OIDC_SECRET: '{{ edx_django_service_with_rendered_config_social_auth_edx_oidc_secret }}'
-  SOCIAL_AUTH_EDX_OIDC_ID_TOKEN_DECRYPTION_KEY: '{{ edx_django_service_with_rendered_config_social_auth_edx_oidc_secret }}'
-  SOCIAL_AUTH_EDX_OIDC_URL_ROOT: '{{ edx_django_service_with_rendered_config_oidc_url_root }}'
-  SOCIAL_AUTH_EDX_OIDC_PUBLIC_URL_ROOT: '{{ edx_django_service_with_rendered_config_oidc_public_url_root }}'
-  SOCIAL_AUTH_REDIRECT_IS_HTTPS: '{{ edx_django_service_with_rendered_config_social_auth_redirect_is_https }}'
```
timmc-edx, line 199: Is this line removal intentional? It's in the middle of a block of OIDC settings (inexplicably) and I don't see it removed in other files.

timmc-edx (approved):

playbooks/roles/edx_django_service_with_rendered_config/defaults/main.yml
schenedx: I think this line should be removed, since it belongs to the same social_auth library right? That means everywhere else this line was not removed, that should be updated to be removed as well?

dianakhuang: It's kind of messy, but this is a setting that seems to be used by the underlying `social-django` library that we're still using and not the `django-oauth2-provider` library that we're trying to remove.

feanil: So it sounds like this line should not be removed?

timmc-edx: It's re-added a few lines down.

feanil (approved):

PR #5645, opened 2020-02-11
BOM-1241 - Add watchman to devstack.

feanil:
Watchman is used by runserver in Django 2.2.  It will be installed by
default in newer versions of ubuntu but needs to be install manually for
ubuntu 16.04(xenial).

We choose to install it from a PPA instead of from source to not make
the build more complex.  The PPA is active and backed by a linux
news company.

### DIFF
--- a/docker/build/xenial-common/Dockerfile
+++ b/docker/build/xenial-common/Dockerfile
@@ -1,3 +1,4 @@
+ARG BASE_IMAGE_TAG=latest
 FROM ubuntu:xenial
 LABEL maintainer="edxops"
 
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -619,8 +619,6 @@ edxapp_requirements_with_github_urls:
   - "{{ base_requirements_file }}"
   - "{{ sandbox_base_requirements }}"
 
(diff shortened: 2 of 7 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -619,8 +619,6 @@ edxapp_requirements_with_github_urls:
   - "{{ base_requirements_file }}"
   - "{{ sandbox_base_requirements }}"
 
-edxapp_chrislea_ppa: "ppa:chris-lea/node.js"
```
feanil, line 622: Note: This is here because while I was looking at where to put the new logic, I noticed that this variable is never referenced. The code that references was removed a while ago I think.

docker/build/xenial-common/Dockerfile
```diff
@@ -1,3 +1,4 @@
+ARG BASE_IMAGE_TAG=latest
```
dianakhuang, line 1: This isn't really a review question but an informational question. Why did this need to get added?

dianakhuang (approved):
Seems good to me.

docker/build/xenial-common/Dockerfile
feanil: Some of the make logic to build the docker files locally looks for this tag via grep and extracts it to use in other places.

PR #5653, opened 2020-02-14
Presupervisor check is failing because the service name for workers is X-worker, but the user is just X. Setting it to the play name instead.

coryleeio:


jdmulloy:
@coryleeio which app(s) is this failing for?

coryleeio:
@jdmulloy you can see the failure on stage registrar

coryleeio:
if you check the status of supervisor it is failing bc of the presupervisor check

coryleeio:
it is failing bc there is no registrar_worker user

coryleeio:
app code owner is named 'registrar'

jdmulloy (approved):

PR #5652, opened 2020-02-13
Add more conditionals for applying 'aws' role

markbreedlove:
**Background**: The task `aws : Install aws python packages` is applied even when we have `COMMON_ENABLE_AWS_ROLE: False` in our Ansible vars.

This PR will add more conditionals for applying the `aws` role, as is done already in
some files, like [playbooks/edxapp.yml](https://github.com/edx/configuration/blob/8afeda16e07f4ed2fd9b0283945a3f650243c616/playbooks/edxapp.yml#L11-L12).

The checkboxes below are from the PR template. I don't think that any of them apply.

I'm happy to redo this PR against the master branch, or another branch if that's more appropriate.

### PR template checkboxes

Make sure that the following steps are done before merging:

  - [ ] A DevOps team member has approved the PR if it is code shared across multiple services and you don't own all of the services.
  - [ ] Are you adding any new default values that need to be overridden when this change goes live? If so:
    - [ ] Update the appropriate internal repo (be sure to update for all our environments)
    - [ ] If you are updating a secure value rather than an internal one, file a DEVOPS ticket with details.
    - [ ] Add an entry to the CHANGELOG.
  - [ ] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/display/EdxOps/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).
  - [ ] Think about how this change will affect Open edX operators.  Have you updated the wiki page for the next Open edX release?

PR #5656, opened 2020-02-19
use debian s3cmd package everywhere

fredsmith:


nedbat:
@fredsmith Thanks.  We think this fixes the same problem as https://github.com/edx/configuration/pull/5652 ?

fredsmith:
@nedbat both are useful independently and together.  I want to keep both changes, although I think either of them would have fixed the problem.

nedbat:
@fredsmith help me out: here's the state I want to get to: 1) we think the problem is fixed, and 2) the problem is fixed the same way on both master and open-release/juniper.alpha1.

We have three options:
1. Merge this PR onto master, cherry-pick it onto juniper.alpha1, and revert #5652.
2. Merge this PR onto master, cherry-pick it onto juniper.alpha1, and also cherry-pick #5652 onto master.
3. Close this PR, and cherry-pick #5652 onto master.

Which do you recommend?

jdmulloy (approved):

nedbat:
@fredsmith Do you have a recommendation?

fredsmith:
@nedbat I would cherry pick both fixes.

nedbat:
I've cherry-picked this PR onto open-release/juniper.alpha1

PR #5690, opened 2020-03-09
Revert "Remove create_user playbook, it breaks sshd config"

fredsmith:
Removing this playbook broke sandboxes.  

Reverts edx/configuration#5667


coryleeio (approved):
You're welcome to merge this if it unblocks someone, but because of the user role setting allowed users in the SSH config it is going to lock out all other users when you add a user.

PR #5705, opened 2020-03-20
Diana/make config upgrade

dianakhuang:
For this PR, we are mostly cleaning up the tasks used to set up `jenkins_worker`s and make them compatible with our Python 3 future. 

Things this does:
* Removed installation of edx-platform requirements under Python 2.7
* Installs node packages without attempting to use nodeenv
* Removes attempts to install and compile `pysqlite`, since `pysqlite` is now included in Python 3
* Installs aws requirements in a virtualenv

Configuration Pull Request
---


playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
```diff
@@ -36,7 +36,7 @@
     virtualenv: "{{ jenkins_home }}/edx-venv-{{ item }}/edx-venv"
     virtualenv_command: virtualenv
     virtualenv_python: "python{{ item }}"
-  with_items: "{{ jenkins_worker_python_versions }}"
+  with_items: "{{ edx_platform_python_versions }}"
```
jmbowman, line 39: There's one more usage of `jenkins_worker_python_versions` above (to create the virtualenv directories) that looks like it can be switched to the new variable.

playbooks/roles/jenkins_worker/templates/jenkins_nodeenv.j2
```diff
@@ -5,5 +5,4 @@
 # installs will be cached and used as a baseline for installations
 # in future test runs.
 cd {{ jenkins_home }}/shallow-clone
-source scripts/jenkins-common.sh
```
jmbowman, line 8: Do we even need this template anymore?  I thought we stopped using nodeenv in Jenkins.

playbooks/roles/aws/tasks/main.yml
```diff
@@ -65,6 +65,7 @@
 - name: Install aws python packages
   pip:
     requirements: "{{ aws_requirements }}"
+    virtualenv: "{{ aws_venv_dir }}"
```
jmbowman, line 68: Shouldn't `virtualenv_python` also be set to tell it which Python version to use?

playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
dianakhuang: It possibly can be, but I thought it might be better to limit the impact of this and have the virtualenv created in case something else needs to use it.

playbooks/roles/jenkins_worker/templates/jenkins_nodeenv.j2
dianakhuang: I think we do because it installs the platform's node packages. I'll update the comment to reflect that.

playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
jmbowman: This is the task that creates the virtualenvs, the one above just creates empty directories to put them in (which I suspect isn't needed anymore, but may have been with a previous version of Ansible).

jmbowman (approved):

playbooks/roles/jenkins_worker/tasks/python_platform_worker.yml
dianakhuang: Ah, okay. I switched it for now and will test the updated branch.

PR #5712, opened 2020-03-27
Page_size=20 on CatalogIntegration for Masters Sandboxes

schenedx:
Update the page_size parameter of the catalog_integration on the Masters sandboxes to be 20 instead of default of 100.
The reason we want to limit the page size to be 20 is because If found an issue where when we make the course_runs api call, the page size 100 would cause 404 on second or third page call to discovery service. Discovery service have default page size of 20 on that endpoint.

@edx/masters-devs Please help review.

Configuration Pull Request
---

### DIFF
--- a/playbooks/masters_sandbox.yml
+++ b/playbooks/masters_sandbox.yml
@@ -45,7 +45,7 @@
     - name: create LMS catalog integration
       shell: >
         . {{ edxapp_env_path }} && {{ edxapp_venv_dir }}/bin/python manage.py lms create_catalog_integrations --enabled --internal_api_url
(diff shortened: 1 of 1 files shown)


kdmccormick (commented):
I'm not sure I understand how changing the page size would fix a 404. Is this a bug in Discovery?

schenedx:
@kdmccormick Fair question. I can dig a bit more to see if we indeed have bugs on discovery service API

matthugs (approved):
either way, I think this is a fine configuration change for Master's sandboxes

kdmccormick:
Yeah, don't take my question as blocking. I didn't want to approve because I don't understand the change, but I also have no qualms with this going out.

PR #5732, opened 2020-04-08
Docker shim INFRA-2

jdmulloy:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edx_django_service/defaults/main.yml
+++ b/playbooks/roles/edx_django_service/defaults/main.yml
@@ -23,6 +23,8 @@ edx_django_service_version: 'master'
 edx_django_service_git_identity: null
 edx_django_service_django_settings_module: null
 
+edx_django_service_app_config_file: "{{ COMMON_CFG_DIR  }}/{{ edx_django_service_name }}.yml"
+
 edx_django_service_code_dir: '{{ edx_django_service_home }}/{{ edx_django_service_name }}'
 edx_django_service_venv_dir: '{{ edx_django_service_home }}/venvs/{{ edx_django_service_name }}'
 edx_django_service_venv_bin_dir: '{{ edx_django_service_venv_dir }}/bin'
@@ -263,3 +265,8 @@ edx_django_service_celery_heartbeat_enabled: true
 edx_django_service_lang: 'en_US.UTF-8'
 edx_django_service_default_stopwaitsecs: 432000
 edx_django_service_workers_supervisor_conf: '{{ edx_django_service_name }}-workers.conf'
+
+edx_django_service_enable_experimental_docker_shim: false
+# Default to blank if edx_django_service_enable_experimental_docker_shim is false
+edx_django_service_docker_run: ""
+edx_django_service_docker_image_name: 'openedx/{{ edx_django_service_name }}'
--- a/playbooks/roles/edx_django_service/meta/main.yml
+++ b/playbooks/roles/edx_django_service/meta/main.yml
@@ -1,6 +1,12 @@
 ---
 dependencies:
   - common
+  - role: docker-tools
+    when: edx_django_service_enable_experimental_docker_shim
+    docker_users:
+      - "www-data"
+      - "{{ supervisor_user }}"
+      - "{{ edx_django_service_user }}"
   - role: supervisor
     supervisor_spec:
       - service: "{{ edx_django_service_name }}"
--- a/playbooks/roles/edx_django_service/tasks/main.yml
+++ b/playbooks/roles/edx_django_service/tasks/main.yml
@@ -21,11 +21,42 @@
 #
 #
 
+- name: Get service user numeric id for docker
+  command: "id -u {{ edx_django_service_user }}"
+  register: edx_django_service_user_id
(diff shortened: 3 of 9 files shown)


playbooks/roles/enterprise_catalog/meta/main.yml
```diff
@@ -12,6 +12,7 @@
 
 dependencies:
   - role: edx_django_service
+    EDX_DJANGO_SERVICE_ENABLE_DOCKER_SHIM: true
```
bderusha: NOTE: We should not merge this with this flag set to `true` without consulting the team that own it.

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -239,12 +263,14 @@
...
     dest: "{{ COMMON_BIN_DIR }}/{{ item.split('.')[0] }}.{{ edx_django_service_name }}"
     state: link
+  when: not EDX_DJANGO_SERVICE_ENABLE_DOCKER_SHIM
   with_items:
     - manage.py
   tags:
     - install
     - install:app-requirements
 
+# TODO: How to handle this in docker build?
```
bderusha: I agree with @coryleeio 's comment made out-of-band 

>  run the make command inside the container with exec when the flag is enabled

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -132,3 +132,5 @@ edx_django_service_migration_check_services: "{{ edx_django_service_name }}"
 EDX_DJANGO_SERVICE_ENABLE_ADMIN_URLS_RESTRICTION: false
 
 EDX_DJANGO_SERVICE_ADMIN_URLS: []
+
+EDX_DJANGO_SERVICE_ENABLE_DOCKER_SHIM: false
```
bderusha: I don't love the term shim here.  Running a docker container on EC2 is a valid approach.  It's a shim for us as we target k8s, but I would rather drop it from the variable name

adzuci: Maybe `ENABLE_SINGLE_CONTAINER_MODE`?

Even though it's a valid approach it isn't widely suggested and I do like the idea of discouraging it's use.
https://docs.docker.com/get-started/orchestration
image

playbooks/roles/edx_django_service/templates/edx/app/app/app.sh.j2
```diff
@@ -2,6 +2,10 @@
 
 # {{ ansible_managed }}
 
+{% if EDX_DJANGO_SERVICE_ENABLE_DOCKER_SHIM %}
+cd ..
+docker-compose up
```
coryleeio: may need to exec here

playbooks/roles/edx_django_service/meta/main.yml
```diff
@@ -1,6 +1,10 @@
 ---
 dependencies:
   - common
+  - role: docker-tools
+    docker_users:
+      - '{{ edx_django_service_automated_users }}'
+      - "{{ edx_django_service_user }}"
   - role: supervisor
     supervisor_spec:
       - service: "{{ edx_django_service_name }}"
```
coryleeio, line 10: I note that just below this it refers to python.service_name, is that still needed?

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -21,19 +21,39 @@
...
+  command: "id -u {{ edx_django_service_user }}"
+  register: edx_django_service_user_id
+  when: edx_django_service_enable_experimental_docker_shim
+  tags:
+    - install
+    - install:system-requirements
+
+- name: Docker build
+  when: edx_django_service_enable_experimental_docker_shim
+  command: make docker_build
```
coryleeio: I wonder if we should force them to implement this in a make file or just call "docker build ."

playbooks/roles/edx_django_service/defaults/main.yml
```diff
@@ -263,3 +265,8 @@ edx_django_service_celery_heartbeat_enabled: true
 edx_django_service_lang: 'en_US.UTF-8'
 edx_django_service_default_stopwaitsecs: 432000
 edx_django_service_workers_supervisor_conf: '{{ edx_django_service_name }}-workers.conf'
+
+edx_django_service_enable_experimental_docker_shim: false
+# Default to blank if edx_django_service_enable_experimental_docker_shim is false
+edx_django_service_docker_run: ""
```
coryleeio, line 271: run_command might be more clear

coryleeio (approved):

playbooks/roles/edx_django_service/tasks/main.yml
jdmulloy: @coryleeio Good thought. The playbook already assumes ```make static``` and ```make migrations```. This is ```make docker_build``` for enterprise catalog. Should I hard code this in Ansible or should I let the app decide what it should do in the Makefile? The playbook does assume that an image named :latest and :latest-newrelic gets built. I suppose I could just build the one we need based on if new relic is enabled or not.

coryleeio (approved):

PR #5738, opened 2020-04-14
Move worker scripts into edx_django_service

jdmulloy:
This moves the worker supervisor conifg files/scripts into the common
edx_django_service role and standardizes deployment of IDAs with celery
workers. It also fixes an issue with all workers for an IDA having the
same app name in NewRelic, regardless of environment.

Configuration Pull Request
---


playbooks/roles/edx_django_service/templates/edx/app/supervisor/conf.d.available/app-workers.conf.j2
```diff
@@ -0,0 +1,22 @@
+{% for w in edx_django_service_workers %}
+[program:{{ w.queue }}_{{ w.concurrency }}]
+
+environment={% if COMMON_ENABLE_NEWRELIC_APP %}NEW_RELIC_APP_NAME={{ edx_django_service_newrelic_appname }}-workers,NEW_RELIC_DISTRIBUTED_TRACING_ENABLED={{ edx_django_service_enable_newrelic_distributed_tracing }},NEW_RELIC_LICENSE_KEY={{ NEWRELIC_LICENSE_KEY }},{% endif -%}CONCURRENCY={{ w.concurrency }},LOGLEVEL=info,DJANGO_SETTINGS_MODULE={{ worker_django_settings_module }},LANG={{ edx_django_service_lang }},PYTHONPATH={{ edx_django_service_code_dir }},BOTO_CONFIG="{{ edx_django_service_home }}/.boto,EDX_REST_API_CLIENT_NAME={{ COMMON_ENVIRONMENT }}-{{ COMMON_DEPLOYMENT }}-{{ edx_django_service_name }}-worker-{{ w.queue }},"
```
jdmulloy, line 4: The LANG setting was set for registrar workers but not enterprise_catalog. UTF-8 seems like a reasonable default for everything. If it's not I can do more work to make it only on registrar.

jdmulloy: Setting EDX_REST_API_CLIENT_NAME here will change the client name for enterprise_catalog workers from
```stage-edx-enterprise-catalog-worker-enterprise_catalog.default```
to
```stage-edx-enterprise_catalog-worker-enterprise_catalog.default```

I don't think this will cause problems. I think it's just a session identifier that can be anything and not an ID that needs to be setup on the LMS. Is this correct?

fredsmith (approved):
nice!

playbooks/roles/edx_django_service/templates/edx/app/supervisor/conf.d.available/app-workers.conf.j2
christopappas: UTF-8 is the standard encoding for python, so that is fine

christopappas:
I don't know quite enough about our playbooks yet to be a great reviewer, but this seems sensible. if it works, 👍

jdmulloy:
@christopappas @chavesj The one complication is for enterprise_catalog, the services tag for the instances needs to be updated, because now the config code will be looking for ```enterprise_catalog-workers``` instead of ```enterprise_catalog_worker```.

Thankfully [pre_supervisor_checks.py](https://github.com/edx/configuration/blob/master/playbooks/roles/supervisor/files/pre_supervisor_checks.py#L42) can handle comma separated lists, so I can make it support both during the switchover.

jdmulloy:
> @christopappas @chavesj The one complication is for enterprise_catalog, the services tag for the instances needs to be updated, because now the config code will be looking for `enterprise_catalog-workers` instead of `enterprise_catalog_worker`.
> 
> Thankfully [pre_supervisor_checks.py](https://github.com/edx/configuration/blob/master/playbooks/roles/supervisor/files/pre_supervisor_checks.py#L42) can handle comma separated lists, so I can make it support both during the switchover.

Actually this doesn't work, because pre_supervisor fails if a config file in the list doesn't exist. I'll need to figure out another solution. I could do a manual deployment to get past this, or I could add an override to the file name, but that feels like a hack.

kdmccormick:
Thanks for the ping; sorry I didn't get to look through this. By cursory glance, it looks good to me.

PR #5739, opened 2020-04-14
No longer deploy private requirements by default

robrap:
The private requirements of edx.org will no longer
be installed by default to docker devstack.

Configuration Pull Request
---


docker/build/edxapp/ansible_overrides.yml
```diff
@@ -20,7 +20,7 @@ EDXAPP_LMS_BASE_SCHEME: http
 COMMON_SECURITY_UPDATES: true
 SECURITY_UPGRADE_ON_ANSIBLE: true
 
-EDXAPP_INSTALL_PRIVATE_REQUIREMENTS: true
+EDXAPP_INSTALL_PRIVATE_REQUIREMENTS: false
```
robrap, line 23: **Reviewers**: I'm guessing this is wrong, since I only want this to affect devstack, but not sure???
Do we want to instead set `EDXAPP_PRIVATE_REQUIREMENTS = []` for devstack?  Where should that be done?  In [lms/envs/devstack.py](https://github.com/edx/edx-platform/blob/master/lms/envs/devstack.py) and cms?

robrap:
Closing for now.  I will reopen based on what the right approach would be.

docker/build/edxapp/ansible_overrides.yml
jmbowman: This should be fine; we're only using the containers built by the Dockerfile here for devstack, I think Tutor uses its own Dockerfiles, and future containers for production use in Kubernetes wouldn't be using Ansible.

jmbowman (approved):

robrap:
Thanks @jmbowman.  I'll merge in the morning.

PR #5749, opened 2020-04-17
Make the change to fix the masters sandbox build failures

schenedx:
Move the analytics_api into python3 now since that repo is ready for it.
---

@edx/masters-devs-cosmonauts FYI


playbooks/roles/analytics_api/meta/main.yml
```diff
@@ -39,7 +39,6 @@ dependencies:
     edx_django_service_django_settings_module: '{{ ANALYTICS_API_DJANGO_SETTINGS_MODULE }}'
     edx_django_service_environment_extra: '{{ analytics_api_environment }}'
     edx_django_service_secret_key: '{{ ANALYTICS_API_SECRET_KEY }}'
-    edx_django_service_use_python3: false
```
coryleeio, line 42: I'm not sure if it is set to true by default, I would explicitly set to true instead of deleting the line

schenedx: @coryleeio It's set to `True` by default. I tested this change and my sandbox build on this branch is green.

coryleeio (approved):

PR #5767, opened 2020-04-29
Pin pip for sandboxes.

feanil:
Using a newer version of pip for sandboxes seems to be causing a
permission denied error when trying to to install sandbox-packages and
symmath from common/lib.  Putting this in place to work around that so
that production deployes aren't blocked on the issue.

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -260,6 +260,32 @@
     - install
     - install:app-requirements
 
+- name: Create the virtualenv to install the Python sandbox requirements
+  command: "virtualenv {{ edxapp_sandbox_venv_dir }} -p python2.7"
+  args:
+    chdir: "{{ edxapp_code_dir }}"
+    creates: "{{ edxapp_sandbox_venv_dir }}/bin/pip"
+  become_user: "{{ edxapp_sandbox_user }}"
+  environment: "{{ edxapp_environment }}"
+  when: EDXAPP_PYTHON_SANDBOX
+  tags:
+    - edxapp-sandbox
+    - install
+    - install:app-requirements
+
+- name: Pin pip to a specific version.
+  # Not pinning to the same version as everything else because sandboxes are still python 2.7
+  command: "{{ edxapp_sandbox_venv_dir }}/bin/pip install pip==20.0.2"
+  args:
+    chdir: "{{ edxapp_code_dir }}"
+  become_user: "{{ edxapp_sandbox_user }}"
+  environment: "{{ edxapp_environment }}"
+  when: EDXAPP_PYTHON_SANDBOX
+  tags:
+    - edxapp-sandbox
+    - install
+    - install:app-requirements
+
 - name: code sandbox | Install base sandbox requirements and create sandbox virtualenv
   pip:
     chdir: "{{ edxapp_code_dir }}"


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -260,6 +260,30 @@
...
     - install:app-requirements
 
+- name: Create the virtualenv to install the Python sandbox requirements
+  command: "virtualenv {{ edxapp_sandbox_venv_dir }} -p python2.7"
+  args:
+    chdir: "{{ edxapp_code_dir }}"
+    creates: "{{ edxapp_sandbox_venv_dir }}/bin/pip"
+  become_user: "{{ edxapp_sandbox_user }}"
+  environment: "{{ edxapp_environment }}"
+  tags:
```
jmbowman, line 271: Should these also have `when: EDXAPP_PYTHON_SANDBOX` like the surrounding tasks?

playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -260,6 +260,30 @@
...
+    - edxapp-sandbox
+    - install
+    - install:app-requirements
+
+- name: Pin pip to a specific version.
+  # Not pinning to the same version as everything else because sandboxes are still python 2.7
+  command: "{{ edxapp_sandbox_venv_dir }}/bin/pip install pip==20.0.2"
```
jmbowman, line 278: We're actually using 19.3.1 elsewhere, although we may want to stick with what you have here since it seemed to be working and we're about to upgrade to it in other places.

playbooks/roles/edxapp/tasks/deploy.yml
feanil: Good catch, will update.

playbooks/roles/edxapp/tasks/deploy.yml
feanil: yea, I wasn't sure if I wanted this to be tracking common pip version or not, thinking more about it, that probably makes it so we test this along with everything else, I was just concerned about pip dropping 2.7 support before we have a chance of pulling the sandbox to python 3.x

timmc-edx: Any value in using the [pip module](https://docs.ansible.com/ansible/latest/modules/pip_module.html) here?

feanil: If I recall, it didn't play nicely when we tried to update pip with pip,  This is mostly a copy paste of the same pattern we use for the edx-platform virtual environments.

jmbowman (approved):

lgp171188:
@feanil, @jmbowman, can this fix be backported to the `open-release/ironwood.master` branch as well since even that is affected? CC @nedbat

nedbat:
It cherry-picks cleanly onto ironwood. Ironwood had a common pip version of 9.0.3, while master has 19.3.1.  Is it OK to use as-is (with the sandbox pip version of 20.0.2)?

lgp171188:
@nedbat, we tested the patch on Ironwood by spawning a sandbox with these changes and it (the provisioning) works okay. Even before this issue happened, the version of `pip` installed in the `edxapp-sandbox` virtualenv environment was `20.0.2`. I picked this up from a sandbox created recently, before this issue happened.

nedbat:
I've cherry-picked this onto open-release/ironwood.master

lgp171188:
@nedbat, thanks for taking care of this so quickly! A very minor nit in the cherry-pick is that [this comment](https://github.com/edx/configuration/blob/open-release/ironwood.master/playbooks/roles/edxapp/tasks/deploy.yml#L285) is not appropriate for Ironwood :)

nedbat:
I can live with that. :)

PR #5768, opened 2020-04-29
Updating build jenkins plugins and war version

coryleeio:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_build/defaults/main.yml
+++ b/playbooks/roles/jenkins_build/defaults/main.yml
@@ -1,6 +1,6 @@
 build_jenkins_user_uid: 1002
 build_jenkins_group_gid: 1004
-build_jenkins_version: jenkins_2.222.1
+build_jenkins_version: jenkins_2.222.3
 build_jenkins_jvm_args: '-Djava.awt.headless=true -Xmx16384m -DsessionTimeout=60'
 
 build_jenkins_python_versions:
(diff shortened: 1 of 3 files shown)


playbooks/roles/jenkins_build/defaults/main.yml
```diff
@@ -1,14 +1,13 @@
...
-build_jenkins_version: jenkins_2.222.1
+build_jenkins_version: jenkins_2.222.3
 build_jenkins_jvm_args: '-Djava.awt.headless=true -Xmx16384m -DsessionTimeout=60'
 
 build_jenkins_python_versions:
   - python3.5-dev
 
 build_jenkins_configuration_scripts:
   - 1addJarsToClasspath.groovy
-  - 2checkInstalledPlugins.groovy
```
adzuci, line 11: Is there a way to have this default to include this check and be able to remove it for local dev easily?

adzuci (approved):
LGTM, I'm hopeful that removing that check won't cause us to not update plugins, but agree that it does slow down local dev.

playbooks/roles/jenkins_build/defaults/main.yml
coryleeio: I dont think it is a meaningful check after updating the ansible to purge that directory on run, I'm fairly sure that is why it was added in the first place

coryleeio: my change should make the plugins directory idempotent

PR #5771, opened 2020-04-30
enable devstack paver autocomplete

robrap:
Configuration Pull Request
---


playbooks/roles/edxapp/templates/devstack.sh.j2
```diff
@@ -20,6 +20,7 @@ case $COMMAND in
 
         . {{ edxapp_venv_bin }}/activate
         cd {{ edxapp_code_dir }}
+        source /edx/app/edx_ansible/edx_ansible/playbooks/roles/local_dev/files/paver_autocomplete.sh
```
robrap: @feanil: I'm not sure where this command belongs in this file, and what the different commands are for. Hoping this might be for the shell?

robrap (commented):
@feanil: I know devops wants new changes to go into docker files in IDAs.  I don't know how far away that is for edx-platform, but I would argue this change will make it less likely that the autocomplete file will get list forever, and more likely that we can transition it to edx-platform along with the docker file.  Thoughts?

playbooks/roles/edxapp/templates/devstack.sh.j2
robrap: @feanil: Oops.  Also, in my notes you wanted this script to move to edx-platform, so I will also need to do that in a separate PR first.  I wonder if it will be safe to remove from here as well?  I'll search and see.

robrap: Added in this PR: https://github.com/edx/edx-platform/pull/23868

playbooks/roles/edxapp/templates/devstack.sh.j2
```diff
@@ -20,6 +20,7 @@ case $COMMAND in
 
         . {{ edxapp_venv_bin }}/activate
         cd {{ edxapp_code_dir }}
+        source scripts/paver_autocomplete.sh
```
robrap: Note: I'm still not sure if this is in the right place in this file.  Help please.

timmc-edx:
Would it make sense to delete `paver_autocomplete.sh` from this repo in the same PR, after https://github.com/edx/edx-platform/pull/23868 is merged? Or does there need to be an overlap period?

robrap:
Thanks @timmc-edx.  I added clean-up to this PR as well.  I imagine it makes sense to do everything at once, once changes are merged to edx-platform.

playbooks/roles/edxapp/templates/devstack.sh.j2
robrap: Turns out this doesn't work. 
1. It belongs under 'open', and not 'exec'.
2. Even under 'open', it isn't working.  It works in .bashrc.  Is there something that can be done to get it to work from here?

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -48,6 +48,15 @@
...
     - devstack:install
 
+- name: add paver autocomplete to bashrc
+  lineinfile:
+    dest: /root/.bashrc
+    line: "source scripts/paver_autocomplete.sh"
+  when: devstack is defined and devstack
+  tags:
+  - devstack
+  - devstack:install
```
robrap, line 58: **Reviewer Questions:** 
1. Should I keep `ignore_errors` from original command?
2. I copy/pasted these tags.  I don't actually know whether they are appropriate. :)

playbooks/roles/edxapp/templates/devstack.sh.j2
robrap: Moved to bashrc after discussion.

playbooks/roles/edxapp/tasks/main.yml
feanil: I think you don't want to ignore errors, and the tags look appropriate.

playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -48,6 +48,15 @@
     - devstack
     - devstack:install
 
+- name: add paver autocomplete to bashrc
+  lineinfile:
+    dest: /root/.bashrc
+    line: "source scripts/paver_autocomplete.sh"
```
feanil: Instead of using a relative path here, I recommend just using the absolute path.  `{{ edxapp_code_dir }}/scripts/paver_autocomplete.sh`

robrap: Done

feanil (approved):
Tested locally, looks good.  It looks like you'll want to merge the edx-platform changes first to prevent this error at lms-shell startup:

```
❯ docker run -it 7230364fca61 /bin/bash
bash: /edx/app/edxapp/edx-platform/scripts/paver_autocomplete.sh: No such file or directory
root@9fff9022fc70:/edx/app/edx_ansible/edx_ansible/docker/plays# 
[0] 0:..form/lms/envs- 1:docker*                      "root@9fff9022fc70: /e" 13:05 05-May-20
```

PR #5774, opened 2020-05-05
Upgrade setuptools, pip, and virtualenv

jmbowman:
Configuration Pull Request
---

The versions of setuptools, pip, and virtualenv we specify here are getting a little dated, which is already causing incompatibility with recent pip-tools releases and has caused at least one installation failure during Juniper testing.  We'd like to upgrade these before Juniper is released.  Points worth noting:

* virtualenv 20+ changed the default behavior to symlinking instead of copying the Python binary, which doesn't play nicely with our current secondary symlinks to that that we use in a few places.  Sticking with 16.7.9 for now, maybe the new service containers won't even need those symlinks so not wasting time working around the issue.
* setuptools 46.1.3 is the latest version as of this writing, but 44.1.0 is the latest release that still supports Python 2.7.  For now I'm just setting us up to use 44.1.0 everywhere, since there are still a few corners where Ansible and such are running 2.7 (those will be fixed in later PRs).
* pip 20.0.2 works with all recent versions of pip-tools, and avoids some breaking changes in 20.1 that we need more time to address.
* virtualenvwrapper was being installed but seemed unused (I can't find anything outside of README files in the `edx` or `edx-ops` organizations that calls `mkvirtualenv` or `workon`), and there isn't a release yet that works with virtualenv>16.7.9.

Successful sandbox build can be found at https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/29943/ .  I'm kicking off another one with more services enabled just to be doubly sure this doesn't break anything.


jmbowman:
Sandbox builds that had problems:

https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/29887/
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/29889/
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/29910/
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/29912/

Edit: these were started using "full_edx_installation", which is no longer the default and had apparently stopped working months ago, but was still specified as the setting to use first in https://openedx.atlassian.net/wiki/spaces/SRE/pages/28967861/Ops+Ansible+Testing+Checklist linked from the PR template.  That page has since been updated accordingly.

jmbowman:
virtualenv 20+ is a rewrite which among other things changes the default behavior on Ubuntu from copying the python executable to making it a symlink.  This is normally fine, but symlinks to that symlink (such as `/edx/bin/python.edxap`) aren't working, at least the way they're currently set up.  Postponing that upgrade until later, since we may not even want that extra symlink in future service containers.

Example of the problem: https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/29918/console

feanil (approved):

timmc-edx (approved):

nedbat:
The Open edX native installation completed successfully on this branch. :+1:

coryleeio (approved):

PR #5775, opened 2020-05-05
Enable JWT settings for Open edX installs CRI-173

nedbat:
This also splits COMMON_SANDBOX_BUILD into its two components: SANDBOX_CONFIG and CONFIGURE_JWTS.

Configuration Pull Request
---

### DIFF
--- a/playbooks/openedx_native.yml
+++ b/playbooks/openedx_native.yml
@@ -27,6 +27,7 @@
     COMMON_ENABLE_AWS_ROLE: false
     ecommerce_create_demo_data: true
     credentials_create_demo_data: true
+    CONFIGURE_JWTS: true
     SANDBOX_ENABLE_BLOCKSTORE: false
     SANDBOX_ENABLE_DISCOVERY: true
     SANDBOX_ENABLE_ECOMMERCE: true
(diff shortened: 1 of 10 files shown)


playbooks/openedx_native.yml
```diff
@@ -27,6 +27,8 @@
     COMMON_ENABLE_AWS_ROLE: false
     ecommerce_create_demo_data: true
     credentials_create_demo_data: true
+    # We are not building a sandbox, but this turns on jwt configuration:
+    COMMON_SANDBOX_BUILD: true
```
adzuci: Could we rename this and split it out from the sandbox CNAME bits rather than merge this as is?

nedbat: I am fine with splitting it, but could use some guidance.

adzuci (approved):
LGTM

adzuci:
@matthugs Will this impact existing registrar sandboxes?

matthugs (approved):
I don't anticipate an adverse impact on master's sandbox builds, since ansible-provision.sh is also our main entry point into this labyrinth, and Ned's addressed that. Thanks for the tag @adzuci

PR #5778, opened 2020-05-11
this is fine

fredsmith:
This is breaking builds because digicert is serving their intermediate certs off a server which is, itself, not serving a complete certificate chain.


matthugs (approved):
😭

PR #5779, opened 2020-05-11
INFRA-144 | docker-shim for management commands

christopappas:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/automated/tasks/main.yml
+++ b/playbooks/roles/automated/tasks/main.yml
@@ -108,3 +108,7 @@
     name: ssh
     state: restarted
   when: users_ssh_access.changed
+
+- name: Add management command users to docker group
+  command: "for user in $(AUTOMATED_USERS.keys()) | list | join(' ')); do sudo usermod -aG docker "$user"; done"
+  when: ENABLE_DOCKER_ACCESS_FOR_AUTOMATED_USERS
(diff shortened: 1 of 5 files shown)


adzuci (approved):

christopappas:
The assumption this makes is that deployments always happen on a fresh filesystem. That assumption is wrong, then I'd also need to add a case for removing the flag if the variable is false

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -407,3 +407,7 @@
      app_config_owner: root
      app_config_group: root
      app_config_mode: 0644
+
+- name: Create docker shim flag file
+  command: "touch /edx/etc/docker_shim_enabled"
```
coryleeio, line 412: Now that we have the tag we could use 'docker exists and tag is present to avoid using the flag, but I don't think it is worth blocking on

playbooks/roles/automated/tasks/main.yml
```diff
@@ -108,3 +108,7 @@
     name: ssh
     state: restarted
   when: users_ssh_access.changed
+
+- name: Add management command users to docker group
+  command: 'for user in $(AUTOMATED_USERS.keys()); do sudo usermod -aG docker "$user"; done'
```
coryleeio: Ansible can do loops with 'with_items'

https://docs.ansible.com/ansible/2.4/playbooks_loops.html
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/tasks/main.yml#L191-L197

I would use the native syntax for this, then I think it's good to go

coryleeio (approved):

PR #5780, opened 2020-05-12
copy jenkins data engineering into new role

estute:
Create a new play/role for setting up a Jenkins server for analytics. I am temporarily suffixing this with "_new", with the intent to port everything over to this, delete the old one, and then rename everything, like this role.

----


HassanJaveed84 (approved):

playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
```diff
@@ -0,0 +1,249 @@
...
+  - name: 'xunit'
+    version: '1.93'
+    group: 'org.jenkins-ci.plugins'
+
+# ghprb
+de_jenkins_ghprb_white_list_phrase: '.*[Aa]dd\W+to\W+whitelist.*'
+de_jenkins_ghprb_ok_phrase: '.*ok\W+to\W+test.*'
+de_jenkins_ghprb_retest_phrase: '.*jenkins\W+run\W+all.*'
+de_jenkins_ghprb_skip_phrase: '.*\[[Ss]kip\W+ci\].*'
+de_jenkins_ghprb_cron_schedule: 'H/5 * * * *'
```
pwnage101, line 235: Looks like there are configurations for ghprb, but I don't see ghprb in the plugins list above.  Is that expected?

pwnage101 (approved):

playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
estute: This was just a copy of the existing de role (which I think copied from build jenkins, which does have the ghprb installed). It is not currently installed on analytics jenkins (hence all of this work). But I am going to install it, so this will be used, eventually.

PR #5786, opened 2020-05-19
Auto-generate more secrets for Open edX installs CRI-199

nedbat:
Some settings were not included in the auto-generated of secure settings.

Also, a minor tweak to how we decode error messages if Ansible fails.


util/install/native.sh
```diff
@@ -174,7 +174,7 @@ if [[ $ansible_status -ne 0 ]]; then
     echo " "
     echo "Decoded error:"
     # Find the FAILED line before the "to retry," line, and decode it.
-    awk '/to retry,/{if (bad) print bad} /FAILED/{bad=$0}' $log_file | python3 /var/tmp/configuration/util/ansible_msg.py
+    awk '/to +retry,/{if (bad) print bad} /FAILED/{bad=$0}' $log_file | python3 /var/tmp/configuration/util/ansible_msg.py
```
adzuci, line 177: What did this do without the +? https://www.grymoire.com/Unix/Awk.html

adzuci (approved):
LGTM

util/install/native.sh
nedbat: I tried to describe it in the commit message for that change:

> If we run this script with -x, then the commands are echoed as they are
executed.  This will echo the awk line, which means awk will find its
own command line in the log file, and print the bad line twice, which
will make the JSON unparsable.
> 
> By changing the pattern to /to +retry,/ we still find the log line we
want, but the pattern doesn't match itself, so it still works when run
with -x.

adzuci: TIL: 
image

Thank you for the detailed commit message!  That makes sense.

PR #5787, opened 2020-05-19
Estute/update de jenkins plugins

estute:
Updating the "new" DE Jenkins. There are a few things here:
* update the core version of jenkins to match build jenkins (v 2.222.3). That server is tested and live, so it should be a good migration target.
* remove some unused plugins that had current security issues. If we choose to install them at a later date, we can pick an appropriate version.
* bump the rest of the plugin versions to match build jenkins, as it has been vetted and is in use.
* Jenkins now writes config files using XML 1.1, so I updated the template files used in this role.
* Stop running the `removeCLI` script. This was a done to patch a security issue which has long since passed (https://github.com/jenkinsci-cert/SECURITY-218) and is now just cruft.
* needed to remove an abandoned `description` field in the seed job dsl. It was causing issues.

I have checked that these plugin versions work with our jenkins instance in a docker container. The scripts run and I can navigate Jenkins. As we port jobs, we might encounter issues, but since we will be porting in chunks, the effects should be minimal.


pwnage101:
Is adding ghprb part of this?

bmedx (approved):
LGTM

pwnage101 (approved):
LGTM, assuming ghprb is part of the plan, still.

estute:
@pwnage101 i had it in a separate pr (prior to doing this upgrade work), but yeah, makes sense to throw it in.

PR #5791, opened 2020-05-20
Install Python 3.8 in Jenkins workers and xenial-common

jmbowman:
Make some changes to facilitate testing our services under Python 3.8:

* Add the deadsnakes PPA to the xenial-common images and install python3.8-dev and python3.8-distutils from it.  This increases the image size from 1.04 GB to 1.06 GB, which I think we can live with.
* Switch jenkins-worker from the legacy deadsnakes PPA that only has up to Python 3.6 to the current one that has through 3.9
* Install Python 3.8 on jenkins_worker, including distutils which is now a separate package (and needed for virtualenv creation to succeed)
* Create an edx-platform virtualenv for Python 3.8 also, we already have Jenkins test jobs that need it
* Try to fix packer AMI builds by specifying the correct Python version to use when testing the image.  Without this, it doesn't use a virtualenv and blows up on a pip/pip-tools version incompatibility ([example](https://build.testeng.edx.org/job/build-packer-ami/14752/consoleFull))

### DIFF
--- a/playbooks/roles/jenkins_worker/tasks/python.yml
+++ b/playbooks/roles/jenkins_worker/tasks/python.yml
@@ -4,7 +4,7 @@
 # newer
 - name: add deadsnakes PPA for newer Python versions
   apt_repository:
-    repo: "ppa:fkrull/deadsnakes"
+    repo: "ppa:deadsnakes/ppa"
     update_cache: yes
   when: ansible_distribution_release == 'xenial'
 
@@ -24,6 +24,14 @@
     state: present
     update_cache: yes
   with_items: '{{ jenkins_worker_python_versions }}'
+
+# Install 'distutils' packages for each installed version of python which has one
+- name: Install python distutils packages
+  apt:
+    name: 'python{{ item }}-distutils'
+    state: present
+    update_cache: yes
+  with_items: '{{ jenkins_worker_distutils_versions }}'
 
 # Requests library is required for the github status script.
 - name: Install requests Python library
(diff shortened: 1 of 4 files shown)


feanil (approved):

playbooks/roles/test_build_server/files/test-development-environment.sh
```diff
@@ -18,6 +18,7 @@ cd edx-platform-clone
 
 # This will run all of the setup it usually runs, but none of the
 # tests because TEST_SUITE isn't defined.
+export PYTHON_VERSION=3.5
```
feanil, line 21: Not blocking but why is this needed?

jmbowman: This was the last bullet point in the description; packer builds have been failing for a couple of weeks because of a version conflict triggered because we weren't setting this to tell them which virtualenv tarball to use.

playbooks/roles/jenkins_worker/tasks/python.yml
```diff
@@ -4,7 +4,7 @@
 # newer
 - name: add deadsnakes PPA for newer Python versions
   apt_repository:
-    repo: "ppa:fkrull/deadsnakes"
+    repo: "ppa:deadsnakes/ppa"
```
adzuci, line 7: Saw that this is needed in a few places, good catch.

adzuci (approved):
:shipit:

PR #5800, opened 2020-05-27
ARCHBOM-1239 - Be able to disable gunicorn instrumentation.

feanil:
Make newrelic monitoring of gunicorn configurable.  We default it to
true which is what it is now by default.  This allows us to configure it
to be false in the future on a per environment basis as we roll it out.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -1613,6 +1613,9 @@ edxapp_cms_gunicorn_host: 127.0.0.1
 edxapp_lms_gunicorn_port: 8000
 edxapp_lms_gunicorn_host: 127.0.0.1
 
+# used in newrelic.ini to enable/disable monitorig of gunicorn
+edxapp_newrelic_monitor_gunicorn: true
+
 # These vars are for creating the application json config
 # files.  There are two for each service that uses the
 # 'edx-platform' code.  Defining them will create the upstart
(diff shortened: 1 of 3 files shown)


playbooks/roles/edxapp/templates/newrelic.ini.j2
```diff
@@ -27,3 +27,13 @@
 # `course_id`.
 #
 browser_monitoring.attributes.enabled=true
+
+
+# See https://docs.newrelic.com/docs/agents/python-agent/configuration/python-agent-configuration#builtin-instrumentation<Paste>
```
robrap: Remove `<Paste>` from end of URL.

robrap (commented):
Thanks.
1. Reminder to update changelog.
2. Where/when will `edxapp_newrelic_monitor_gunicorn` be set to `false` for Stage?  Does that happen inside or outside of configuration?

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1613,6 +1613,9 @@ edxapp_cms_gunicorn_host: 127.0.0.1
 edxapp_lms_gunicorn_port: 8000
 edxapp_lms_gunicorn_host: 127.0.0.1
 
+# used in newrelic.ini to enable/disable monitorig of gunicorn
```
robrap, line 1616: typo: monitorig

robrap (approved):
Minor typo.  I'll also test before we merge.

feanil:
Closing because we're going a different route.

PR #5816, opened 2020-06-09
Feanil/update including sandbox 2

feanil:


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1649,8 +1650,7 @@ base_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/base.txt"
 django_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/django.txt"
 openstack_requirements_file: "{{ edxapp_code_dir }}/requirements/edx/openstack.txt"
 
-sandbox_base_requirements: "{{ edxapp_code_dir }}/requirements/edx-sandbox/base.txt"
-sandbox_py35_requirements: "{{ edxapp_code_dir }}/requirements/edx-sandbox/py35.txt"
+sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/{% if edxapp_sandbox_python_version == 'python2.7' %}base.txt{% else %}py35.txt{% endif %}"
```
robrap, line 1653: 1. Could you explain this transition and the relationship between `sandbox_base_requirements`, `sandbox_py35_requirements`, `base.txt` and `py35.txt`?
2. How are you able to remove `sandbox_py35_requirements` with no other changes?  Was it never used?

robrap (commented):
Just questions at this point.  Thanks.

playbooks/roles/edxapp/defaults/main.yml
feanil: ah, good catch @robrap need to update sandbox requirements in one more place.  I'll make a followup pr.

PR #5819, opened 2020-06-10
Fix ecomworker Docker image builds ARCHBOM-1270

jmbowman:
This Docker image hasn't been built in 3 years, but is still used in the ecommerce-worker reop's Travis tests.  Using this outdated image is now a blocker for merging Python 3.8 support, since it doesn't have Python 3.8 installed to test with (whereas the latest xenial-common images do).  Updated the Dockerfile to work with the current tooling for image builds.

### DIFF
--- a/docker/build/ecomworker/Dockerfile
+++ b/docker/build/ecomworker/Dockerfile
@@ -7,7 +7,8 @@
 # This allows the dockerfile to update /edx/app/edx_ansible/edx_ansible
 # with the currently checked-out configuration repo.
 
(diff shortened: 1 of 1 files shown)


timmc-edx (approved):
Seems about right!

Should there be a changelog entry?

jmbowman:
I only see one previous changelog entry for a change in the Docker image builders, and that was a pretty significant one for edx-platform; I'll skip the entry for "fixed this thing that hasn't worked in years", since it seems unlikely to be relevant for many people.

PR #5820, opened 2020-06-10
update sandbox env to python 3 on Juniper

feanil:


nedbat:
Does this mean that all courses have to have their Python code updated to Python 3?

feanil:
Yes.

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1649,7 +1650,7 @@ base_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/base.txt"
 django_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/django.txt"
 openstack_requirements_file: "{{ edxapp_code_dir }}/requirements/edx/openstack.txt"
 
-sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/base.txt"
+sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/{% if edxapp_sandbox_python_version == 'python2.7' %}base.txt{% else %}py35.txt{% endif %}"
```
nedbat, line 1653: It might have made more sense to rename base to py27, and let base be the "right default", especially since we will be moving to 3.8.

nedbat (approved):

PR #5822, opened 2020-06-12
Fix pip version in jenkins_worker virtualenvs ARCHBOM-1277

jmbowman:
The latest failure in jenkins_worker AMI builds was that a pip version compatible with the current version of pip-tools in edx-platform wasn't being installed, so pip-sync would fail when updating dependencies in those virtualenvs.  This change creates the virtualenv and installs the correct versions of pip, setuptools, etc. before installing the rest of the dependencies.

The build at https://build.testeng.edx.org/job/build-packer-ami/14779 from this branch succeeded.


robrap (approved):
Nothing stands out at me as a problem, but a successful build will probably tell you more than my review. :)

PR #5828, opened 2020-06-22
WIP: Remove /var/lib/apt/lists/* before apt-get-update

adzuci:
Configuration Pull Request
---

### DIFF
--- a/util/install/ansible-bootstrap.sh
+++ b/util/install/ansible-bootstrap.sh
@@ -92,6 +92,7 @@ fi
 EDX_PPA="deb http://ppa.edx.org ${SHORT_DIST} main"
 
 # Upgrade the OS
+sudo rm -r /var/lib/apt/lists/* -vf
 apt-get update -y
 
 # To apt-key update in bionic, gnupg is needed.
@@ -122,7 +123,6 @@ if [[ "${SHORT_DIST}" != bionic ]] ;then
   add-apt-repository -y "${EDX_PPA}"
 fi
 
-
 # Install python 2.7 latest, git and other common requirements
 # NOTE: This will install the latest version of python 2.7 and
 # which may differ from what is pinned in virtualenvironments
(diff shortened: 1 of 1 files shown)


util/install/ansible-bootstrap.sh
```diff
@@ -92,6 +92,7 @@ fi
 EDX_PPA="deb http://ppa.edx.org ${SHORT_DIST} main"
 
 # Upgrade the OS
+sudo rm /var/lib/apt/lists/* -vf
```
jmbowman: Looks like it needs to be `sudo rm -r ...`: https://build.testeng.edx.org/job/build-packer-ami/14791/console

util/install/ansible-bootstrap.sh
```diff
@@ -92,6 +92,7 @@ fi
 EDX_PPA="deb http://ppa.edx.org ${SHORT_DIST} main"
 
 # Upgrade the OS
+sudo rm -r /var/lib/apt/lists/* -vf
```
coryleeio, line 95: I wouldn't sudo this, given the other lines in here you can probably assume it's already been sudoed or run as root

coryleeio: Ah Jeremy suggested you sudo it, so I just made a catch 22 for you I'm worried sudo wont be installed in some contexts, and i'm worried about side effects if someone in the community runs this

coryleeio: I would make this build from scratch or use a new linux ami of the correct version
It is either using an old AWS ami or using our last build AMI and eventually becoming out of date, I would recommend fixing it that way if you can then this issue shouldn't occur.

coryleeio (approved):
Thumbed with caveats

util/install/ansible-bootstrap.sh
adzuci: Working on looking at how to update the AMI.

adzuci: Based on https://github.com/edx/jenkins-job-dsl/blame/master/testeng/resources/build-packer-ami.sh#L6 I think at least this packer build does build from scratch from the latest 16.04 image.

We may want to file a ticket to update build jenkins to 18.04 or 10.04 too.

PR #5833, opened 2020-06-25
Add the ability to deploy credentials on python 3.8

mraarif:
`credentials` is already running tests against python 3.8 in CI, this PR intends to get the pipeline ready for credentials deployment on python 3.8.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/credentials/meta/main.yml
+++ b/playbooks/roles/credentials/meta/main.yml
@@ -12,6 +12,7 @@
 #
 dependencies:
   - role: edx_django_service
+    edx_django_service_use_python38: true
     edx_django_service_version: '{{ CREDENTIALS_VERSION }}'
     edx_django_service_name: '{{ credentials_service_name }}'
     edx_django_service_config_overrides: '{{ credentials_service_config_overrides }}'
--- a/playbooks/roles/edx_django_service/tasks/main.yml
+++ b/playbooks/roles/edx_django_service/tasks/main.yml
@@ -61,6 +61,21 @@
     - install
     - install:configuration
 
+- name: add deadsnakes repo
+  apt_repository:
+      repo: ppa:deadsnakes/ppa
+  when: edx_django_service_use_python38 and not edx_django_service_enable_experimental_docker_shim
+
+- name: install python3.8
+  apt:
+    pkg:
+      - python3.8-dev
+      - python3.8-distutils
+  when: edx_django_service_use_python38 and not edx_django_service_enable_experimental_docker_shim
+  tags:
+    - install
+    - install:system-requirements
+
 - name: install python3
   apt:
     name: "{{ item }}"
@@ -72,12 +87,22 @@
     - install
     - install:system-requirements
 
+- name: build virtualenv with python3.8
+  command: "virtualenv --python=python3.8 {{ edx_django_service_venv_dir }}"
+  args:
+    creates: "{{ edx_django_service_venv_dir }}/bin/pip"
+  become_user: "{{ edx_django_service_user }}"
+  when: edx_django_service_use_python38 and not edx_django_service_enable_experimental_docker_shim
+  tags:
+    - install
+    - install:system-requirements
+
 - name: build virtualenv with python3
   command: "virtualenv --python=python3 {{ edx_django_service_venv_dir }}"
   args:
     creates: "{{ edx_django_service_venv_dir }}/bin/pip"
   become_user: "{{ edx_django_service_user }}"
-  when: edx_django_service_use_python3 and not edx_django_service_enable_experimental_docker_shim
+  when: edx_django_service_use_python3 and not edx_django_service_use_python38 and not edx_django_service_enable_experimental_docker_shim
   tags:
     - install
     - install:system-requirements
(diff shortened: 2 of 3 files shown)


playbooks/roles/credentials/meta/main.yml
```diff
@@ -12,6 +12,7 @@
 #
 dependencies:
   - role: edx_django_service
+    edx_django_service_use_python38: true
```
syedimranhassan, line 15: This variable needs to be defined in edx_django_service role defaults.
https://github.com/edx/configuration/blob/master/playbooks/roles/edx_django_service/defaults/main.yml

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -72,6 +79,17 @@
     - install
     - install:system-requirements
 
+- name: build virtualenv with python3.8
+  command: "virtualenv --python=python3.8 {{ edx_django_service_venv_dir }}"
+  args:
+    creates: "{{ edx_django_service_venv_dir }}/bin/pip"
+  become_user: "{{ edx_django_service_user }}"
+  when: edx_django_service_use_python38 and not edx_django_service_enable_experimental_docker_shim
```
syedimranhassan, line 95: Update condition for py2.7 and py3 virtual environment to not install when this variable is set.
You can see a similar example in **build virtualenv with python2.7** task bellow.

syedimranhassan (changes requested):

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -72,12 +79,23 @@
     - install
     - install:system-requirements
 
+- name: build virtualenv with python3.8
```
morenol, line 90: I think that we can reuse the 

- name: build virtualenv with python3, task by defining something like 

`edx_django_service_python_version` and use that there, we can also use that instead of `edx_django_service_use_python38`

playbooks/roles/edx_django_service/tasks/main.yml
```diff
@@ -61,6 +61,13 @@
     - install
     - install:configuration
 
+- name: install python3.8
+  command: apt-add-repository -y ppa:deadsnakes/ppa | apt-get update && apt-get install -y python3.8-dev python3.8-distutils
```
morenol: I think that we should use here

 https://docs.ansible.com/ansible/latest/modules/apt_repository_module.html

And then we can install these packages using https://docs.ansible.com/ansible/latest/modules/apt_module.html just like in the install python3 task that is above

playbooks/roles/edx_django_service/tasks/main.yml
mraarif: currently, there are 3 different python versions we've to deal with, you're suggesting we check for the exact python version while building a virtualenv? (edx_django_service_python_version==3.8)

morenol: Is there a service (that is installed with edx_django_service role) that is still using python2.7? In that case I think that the change that you made is Ok, otherwise I think that we should remove the edx_django_service_use_python3 and instead use something like edx_django_service_python_version as you said.

mraarif: there are a few playbooks that have `edx_django_service_use_python3` set to `false` implying that they use python 2.7

mraarif:
I was able to successfully create a sandbox ([here](https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/30575/)) using this branch.

morenol (approved):
This looks good to me, maybe we should check the ubuntu version used in order to decide if we need to add the ppa:deadsnakes/ppa repository. But since we are not working with focal fossa, I think that this is fine.

nadeemshahzad (approved):

morenol:
@mraarif I think that the changes made in playbooks/roles/edx_django_service/* can be made in a separate PR, and keep in this one only the change to the playbooks/roles/credentials/meta/main.yml

So we can merge that other PR quickly and we can reuse these changes for the other services, What do you think @awais786 ?

mraarif:
> @mraarif I think that the changes made in playbooks/roles/edx_django_service/* can be made in a separate PR, and keep in this one only the change to the playbooks/roles/credentials/meta/main.yml
> 
> So we can merge that other PR quickly and we can reuse these changes for the other services, What do you think @awais786 ?

@morenol I think none of the other services are deployed on python 3.8 yet, and as per @jmbowman team owning credentials should be on-board with this change first.

morenol:
> > @mraarif I think that the changes made in playbooks/roles/edx_django_service/* can be made in a separate PR, and keep in this one only the change to the playbooks/roles/credentials/meta/main.yml
> > 
> > So we can merge that other PR quickly and we can reuse these changes for the other services, What do you think @awais786 ?
> 
> @morenol I think none of the other services are deployed on python 3.8 yet, and as per @jmbowman team owning credentials should be on-board with this change first.


Yes, but I mean, in order to create a PR to change python3.8 for other service, let's say course discovery we would need to repeat the changes that you made in edx_django_service. I think that these changes themselves are OK and they would not affect any of the other services since python3.5 is still the default configuration

jmbowman:
@mraarif Are you planning to do any further refactoring of this PR, or go with it as is?

@edx/masters-devs Could one of you review this and start making any plans needed to roll out the upgrade?

mraarif:
@jmbowman as there aren't any services currently blocked because of python 3.8 (i.e these config changes), I think we should go ahead with this and see how credentials deployment goes

morenol:
Hi @edx/master-devs This is a kindly remainder for this PR. Do you have any updates?

matthugs (approved):
I think testing this change against staging would be a good next step.

Approving on behalf of @edx/masters-devs

syedimranhassan (approved):

matthugs:
deployed to production credentials 29 Jul 2020 at 11:35:05

PR #5842, opened 2020-06-30
ARCHBOM-1326 - Update oppia xblock to point back upstream.

feanil:
Oppia has merged our PR to allow them to run on python 3 so we're updating
the pointer to the merge commit upstream.


timmc-edx (approved):
Cool! Verified that commits have identical content.

PR #5864, opened 2020-07-13
BOM-1872 : Change edxapp configuration to use edx-zoom pypi release

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-1872

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -532,7 +532,7 @@ EDXAPP_PRIVATE_REQUIREMENTS:
       extra_args: -e
     - name: git+https://github.com/open-craft/xblock-activetable.git@013003aa3ce28f0ae03b8227dc3a6daa4e19997d#egg=xblock-activetable
       extra_args: -e
-    - name: edx-zoom
+    - name: edx-zoom==1.8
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -532,8 +532,7 @@ EDXAPP_PRIVATE_REQUIREMENTS:
       extra_args: -e
     - name: git+https://github.com/open-craft/xblock-activetable.git@013003aa3ce28f0ae03b8227dc3a6daa4e19997d#egg=xblock-activetable
       extra_args: -e
-    - name: git+https://github.com/edx/edx-zoom.git@37c323ae93265937bf60abb92657318efeec96c5#egg=edx-zoom
-      extra_args: -e
+    - name: edx-zoom
```
jmbowman: It looks like the current PyPI release predates Django 2 support; I presume a new release will get made before this is merged?  Also, we should probably pin to a specific version (like xblock-problem-builder above in this file).

mzulqarnain1: Got it. I'll release a newer version and pin it.

mzulqarnain1: Done.

jmbowman (approved):

PR #5867, opened 2020-07-15
[BD-6] Use pip3 in the Makefile

morenol:
Use pip3 in the Makefile in order to force the usage of python3

## Reviewers
- [ ] @awais786 
- [ ] @ericfab179

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,13 +17,13 @@ main.help:
 	@echo ''
 
 requirements:
-	pip install -qr pre-requirements.txt --exists-action w
-	pip install -qr requirements.txt --exists-action w
+	pip3 install -qr pre-requirements.txt --exists-action w
+	pip3 install -qr requirements.txt --exists-action w
 
 upgrade: export CUSTOM_COMPILE_COMMAND=make upgrade
 upgrade: ## update the pip requirements files to use the latest releases satisfying our constraints
-	pip install -qr pre-requirements.txt --exists-action w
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [BLENDED-450](https://openedx.atlassian.net/browse/BLENDED-450) to keep track of it in Jira.  More details are on the [BD-06](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1594000004/BD-06+Python+3.8+Upgrade) project page.

awais786 (approved):

nedbat:
I think i see the problem in the native instructions that get some stuff from master ("sudo bash" should be "sudo -E bash").

What problem was this originally solving? We should be able to make changes on master.

morenol:
> What problem was this originally solving?

Not a problem in specific, I am just trying to remove the usage of python2.7

nedbat:
These commands should all be running in a virtualenv, so "pip" does refer to Python 3, no?

morenol:
> These commands should all be running in a virtualenv, so "pip" does refer to Python 3, no?

Yes, if the virtualenv is created with python3. Currently it is being created with python2.7 

https://github.com/edx/configuration/blob/4c8a5e98370270cb231339aebc3d7bca2e3a8763/util/install/ansible-bootstrap.sh#L144

nedbat:
That is something to fix!

PR #5876, opened 2020-07-20
[BD-6] Remove duplicated dict key (when).

morenol:
`when` is duplicated in this task, it is using the valie defined in the last line of the task.

 ```yaml
 when: edx_django_service_npm_version is defined and not edx_django_service_enable_experimental_docker_shim
```

So we can remove this line. 

This is the warning in the travis tests:
```
 [WARNING]: While constructing a mapping from /home/travis/build/edx/configurat
ion/playbooks/roles/edx_django_service/tasks/main.yml, line 125, column 3,
found a duplicate dict key (when). Using last defined value only.
```

## Reviewers
- [ ] @awais786 
- [ ] @ericfab179

### DIFF
--- a/playbooks/roles/edx_django_service/tasks/main.yml
+++ b/playbooks/roles/edx_django_service/tasks/main.yml
@@ -125,7 +125,6 @@
 - name: upgrade npm
   command: "npm install -g npm@{{ edx_django_service_npm_version }}"
   become_user: "{{ edx_django_service_user }}"
-  when: not edx_django_service_enable_experimental_docker_shim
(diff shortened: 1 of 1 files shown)


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [BLENDED-476](https://openedx.atlassian.net/browse/BLENDED-476) to keep track of it in Jira.  More details are on the [BD-06](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1594000004/BD-06+Python+3.8+Upgrade) project page.

awais786 (approved):

PR #5881, opened 2020-07-20
[BD-6]Ansible 2.8

morenol:
Upgrade ansible to 2.8.

I checked the [porting guide](https://docs.ansible.com/ansible/latest/porting_guides/porting_guide_2.8.html) for ansible 2.8.  And it seems to me that there is not issues with this upgrade.

After this change the python3.8 tests are passing, though I think that we should use at least ansible 2.9.5 since that is the first ansible release with python3.8 in its PyPi classifiers.

## Reviewers
- [ ] @awais786 
- [ ] @ericfab179

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,3 @@ script:
   - docker --version
   - make --version
   - travis_wait 90 make --keep-going $MAKE_TARGET SHARD=$SHARD SHARDS=$SHARDS
-
-matrix:
-  allow_failures:
-    - python: 3.8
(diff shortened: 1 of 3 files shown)


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [BLENDED-485](https://openedx.atlassian.net/browse/BLENDED-485) to keep track of it in Jira.  More details are on the [BD-06](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1594000004/BD-06+Python+3.8+Upgrade) project page.

PR #5901, opened 2020-07-30
[BD-6] Add configuration to allow the installation of insights with python3.8

morenol:
Add INSIGHTS_PYTHON_VERSION ansible configuration in order to allow the installation of insights with python3.8.

Based in the changes of https://github.com/edx/configuration/pull/5833
## Reviewers
- [ ] @awais786

### DIFF
--- a/playbooks/roles/insights/tasks/main.yml
+++ b/playbooks/roles/insights/tasks/main.yml
@@ -32,13 +32,31 @@
     - install
     - install:configuration
 
+- name: add deadsnakes repo
+  apt_repository:
+      repo: ppa:deadsnakes/ppa
+  when: INSIGHTS_PYTHON_VERSION == "3.8" and ansible_distribution_release != 'focal'
+  tags:
+    - install
+    - install:system-requirements
+
+- name: install python3.8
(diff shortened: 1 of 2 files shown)


openedx-webhooks:
Thanks for the pull request, @morenol! I've created [BLENDED-525](https://openedx.atlassian.net/browse/BLENDED-525) to keep track of it in Jira.   More details are on the [BD-06](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1594000004/BD-06+Python+3.8+Upgrade) project page.


playbooks/roles/insights/tasks/main.yml
```diff
@@ -32,13 +32,31 @@
     - install
     - install:configuration
 
+- name: add deadsnakes repo
```
mzulqarnain1, line 35: We don't need to add it here separately, Insights is using `edx_service` as it's base and we're already adding it there, you just need to turn on the flag `edx_service_use_python38`. I already have a PR up for that https://github.com/edx/configuration/pull/5975/files
It's in review by the owning team.

PR #5919, opened 2020-08-06
Revert "WS-1119 (take 2): running tests on prospectus stage build"

srwang:
Having this ready in case edx/configuration#5917 breaks or slows down build significantly


adzuci (approved):
Thumbing this in case you need it.

PR #5932, opened 2020-08-11
Revert "WS-1119 (try four-fix): add docker role for Prospectus tests"

srwang:
Reverts edx/configuration#5930


adzuci (approved):
Could probably do these reverts in one PR or fix forward by removing integration tests but keeping unit tests.

PR #5953, opened 2020-08-20
Improve sandboxing.

feanil:
- Add a test to ensure that sandbox is being enforced correctly.
- Also Enforce apparmor on all binaries in sandbox python env.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/tasks/python_sandbox_env.yml
+++ b/playbooks/roles/edxapp/tasks/python_sandbox_env.yml
@@ -91,3 +91,21 @@
   command: /usr/sbin/{{ edxapp_aa_command }} /etc/apparmor.d/code.sandbox
   tags:
     - edxapp-sandbox
+
+- name: code sandbox | test enforcement 1
+  command: "{{ edxapp_sandbox_venv_dir }}/bin/python -c \"import os; os.listdir('/')\""
+  register: sandbox_test1
+  failed_when: "'PermissionError:' not in sandbox_test1.stderr"
+  when: EDXAPP_SANDBOX_ENFORCE
+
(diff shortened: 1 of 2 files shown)


feanil:
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/32109/consoleFull  if you search for `test enforcement` you'll see the runs of these tests.

fredsmith (approved):
nice!

timmc-edx (approved):
Looks great!

Should there also be a positive case to show that apparmor is not overly strict?

feanil:
@timmc-edx were you thinking something like basic math?  given that the possible positive cases are quite large, and we have good mechanisims for knowing if we've tripped them(alerting, instructor reports)  I didn't add one.  But if you have a minimal case in mind, I'm happy to put it in here.

timmc-edx:
Yeah, was thinking math or other trivial execution. But I'm not sure how hard it would be to check that it worked as opposed to being silently ignored. :-P Maybe this isn't the right place for that sort of test.

nedbat:
I like the idea of a positive test. Maybe I'm missing something, wouldn't it be as simple as:
```
- name: code sandbox | test enforcement
  command: "{{ edxapp_sandbox_venv_dir }}/bin/python -c \"print(7*11*13*17)\""
  register: sandbox_test3
  failed_when: "'17017' not in sandbox_test3.stdout"
  when: EDXAPP_SANDBOX_ENFORCE
```
Why would it be silently ignored?

timmc-edx:
Ah, I didn't know we could write to stdout here -- but that makes sense. :-)

PR #5968, opened 2020-08-27
update the edx-notes-api config to use python 3.8

mraarif:
Configuration Pull Request
---
Relevant JIRA issue here: https://openedx.atlassian.net/browse/BOM-1965

### DIFF
--- a/playbooks/roles/edx_notes_api/defaults/main.yml
+++ b/playbooks/roles/edx_notes_api/defaults/main.yml
@@ -142,5 +142,6 @@ edx_notes_api_debian_pkgs:
   - python-mysqldb
   - libssl-dev # needed for mysqlclient python library
   - python3-dev
+  - python3.8-dev
 
 edx_notes_api_redhat_pkgs: []
--- a/playbooks/roles/edx_notes_api/tasks/main.yml
+++ b/playbooks/roles/edx_notes_api/tasks/main.yml
@@ -48,7 +48,7 @@
     virtualenv: "{{ edx_notes_api_home }}/venvs/{{ edx_notes_api_service_name }}"
     state: present
     extra_args: "--exists-action w"
-    virtualenv_python: 'python3.5'
+    virtualenv_python: 'python3.8'
   become_user: "{{ edx_notes_api_user }}"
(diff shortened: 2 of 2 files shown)


mraarif:
I've created a sandbox: https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/32296/
Using this branch and the sandbox build seems to be successful

jdmulloy (approved):
@mraarif This seems fine to me. It's good to update this but I don't think this will affect the python version for stage and prod since those are now deployed in a docker container on kubernetes. The Dockerfile in the notes repo also needs to be updated if you haven't already done that.

Thanks

mraarif:
@jdmulloy the [Dockerfile](https://github.com/edx/edx-notes-api/blob/master/Dockerfile) in edx-notes-api  seems to have python 3.8 installation already in place
let me know if there's something else that needs to be done.

mraarif:
@jmbowman please tag the owning team here.

jmbowman:
The Dockerfile installs packages using `pip3 --install`, which uses the system default Python (which on Ubuntu 18.04 is 3.6).  So yes, that'll have to change to something like `python3.8 -m pip install`.  Ditto for the `newrelic` installation near the bottom of the file.

And although programs-masters now officially owns edx-notes-api, I think SRE is still handling much of the operations until they can finish adopting ownership.  So I think @jdmulloy 's approval is the one that counts here.

mraarif:
I created a PR ([edx-notes-api/#196](https://github.com/edx/edx-notes-api/pull/196)) to update the Dockerfile,
thanks for the catch @jmbowman

playbooks/roles/edx_notes_api/defaults/main.yml
```diff
@@ -142,5 +142,6 @@ edx_notes_api_debian_pkgs:
   - python-mysqldb
   - libssl-dev # needed for mysqlclient python library
   - python3-dev
+  - python3.8-dev
```
mzulqarnain1, line 145: Since note-api uses edx_service as a base , we should also turn on `edx_service_use_python38` flag here to add deadsnakes repo like its done [here](https://github.com/edx/configuration/pull/5975/files)

PR #5974, opened 2020-08-30
add support for Ubuntu 20.04 (PSRE-203)

nadeemshahzad:
Configuration Pull Request
---


playbooks/roles/common/tasks/main.yml
```diff
@@ -106,6 +104,19 @@
     update_cache: yes
   when: ansible_distribution in common_debian_variants
 
+# Pip for Python2.7 is not included in the Ubuntu 20.04 repositories
```
fredsmith: why are we installing python 2.7?

nadeemshahzad: @fredsmith Installing python 2.7 as Ansible tries to use python 2 by default and can not even make it past the Gathering facts task so we will not able to ```use ansible_python_interpreter: /usr/bin/python3``` in our configuration for Ubuntu 20.04. Ansible 2.8 was shipped with interpreter discovery (https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst#major-changes ), we can update the ansible in configuration with 2.8 release and can completely get rid of python 2  on Ubuntu 20.04.

playbooks/roles/mysql/tasks/mysql.yml
```diff
@@ -33,19 +33,62 @@
...
+  apt_repository:
+    repo: "{{ MYSQL_REPO }}"
+    state: present
+  when: ansible_distribution_release == 'focal'
+
+# Installing mysql-5.7-client, mysql-5.7-community-server
+# and mysql-5.7-server in separate tasks to resolve dependencies
+- name: install mysql-5.7-client
+  apt:
+    name: mysql-client=5.7.31-1ubuntu18.04
```
adzuci: It looks like you define `{{ mysql_pkgs }}` for mysql 5.6 in defaults, but hard code 5.7 here, any reason for that? https://github.com/edx/configuration/pull/5974/files#diff-64d565c7641d1d6d0f05105b2ab64fe7R44

playbooks/roles/nginx/defaults/main.yml
```diff
@@ -128,7 +128,7 @@ nginx_user: root
 nginx_htpasswd_file: "{{ nginx_app_dir }}/nginx.htpasswd"
 nginx_default_sites: []
 nginx_debian_pkgs:
-  - python-passlib
+    - python3-passlib
```
adzuci: I don't think you needed to add two spaces in front of this.

playbooks/roles/notifier/tasks/deploy.yml
```diff
@@ -49,6 +49,7 @@
   pip:
     requirements: "{{ NOTIFIER_REQUIREMENTS_FILE }}"
     virtualenv: "{{ NOTIFIER_VENV_DIR }}"
+    virtualenv_python: "{{ notifier_python_version }}"
```
adzuci: Can we just get rid of notifier?

adzuci (approved):
I left a few comments, but I think this should be ok to merge from our perspective.  I am curious if OpenCraft may be using the bootstrap script for instances that need Python 2.7 to be explicitly installed.

cc @bradenmacdonald do you feel that this change will impact you guys and would it make sense to get a review from someone on your team before we merge this change into master?

playbooks/roles/mysql/tasks/mysql.yml
nadeemshahzad: Updated to use vars for mysql 5.7 pkgs

playbooks/roles/nginx/defaults/main.yml
nadeemshahzad: fixed

playbooks/roles/notifier/tasks/deploy.yml
nadeemshahzad: Removed notifier from sandbox builds

nadeemshahzad:
Ansible 2.5 and above work with python 3 and from release 2.8 it can correctly discover the default python interpreter on remote hosts ( https://github.com/ansible/ansible/blob/stable-2.8/changelogs/CHANGELOG-v2.8.rst#major-changes ). We have upgraded the ansible to 2.8.18 (https://github.com/edx/configuration/pull/5986/files) and now we can safely remove python2.7 packages for Ubuntu xenial and above.  I have successfully build the sandboxes for Ubuntu xenial and Bionic. 
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/32757/
https://tools-edx-jenkins.edx.org/job/Sandboxes/job/CreateSandbox/32759/

symbolist:
@adzuci @nadeemshahzad Thanks for the heads up! @bradenmacdonald asked me to take a quick look here. As long as support for Ubuntu 18.04 is not being dropped, it should be perfectly fine (we can resolve any small issues that come up quickly). From a scan it does not look like this PR does drop that?

PR #6009, opened 2020-09-14
WS-1232: Dry run of turning off the drupal fall back

reillz10:
Configuration Pull Request
---


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -31,28 +31,6 @@ server {
...
-  location /admin/ {
-    add_header 'Cache-Control' 'public, max-age=0, must-revalidate';
-    try_files $uri @drupal;
-  }
-
-  location /static/ {
-    add_header 'Cache-Control' 'public, max-age=31536000, immutable';
-  }
-
-  location /es/bio/ {
```
rgraber: we need to keep these blocks in for the dynamic bio pages to work

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -82,23 +60,12 @@ server {
   {% endfor %}
 
   location / {
-    location ~ \.(js|css)$ {
-      add_header 'Cache-Control' 'public, max-age=31536000, immutable';
```
rgraber: i think we need to keep the caching for js/css files in

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -122,16 +89,4 @@ server {
     }
 
   {% endfor %}
-
-  location @index {
-    add_header 'Cache-Control' 'public, max-age=0, must-revalidate';
-    try_files $uri/index.html @drupal;
```
rgraber: We need the index.html part of this because that's where the actual generated html lives

rgraber (changes requested):
We should just take out the drupal parts of things, not all the try_files

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
reillz10: ahh good to know. I'll put them back in. Thanks @rgraber!

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
reillz10: But not as a `try_files` right? This is how the learnerPortal MFE does it: https://github.com/edx/configuration/blob/master/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/learner_portal.j2#L12

rgraber: oh interesting. i guess not as a try_files. I would leave in the cache part though

rgraber (changes requested):
I think we should leave in all the Cache-Control headers. I don't think they're just for drupal

rgraber (commented):
I hate to make things more complicated, but I think we should put these changes under a feature flag so we can test on stage without pausing the pipeline

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
reillz10: Got it! Those are back in now

rgraber (approved):
LGTM

srwang (approved):

PR #6045, opened 2020-09-25
update the configuration to use python 3.8 for edx-platform

mraarif:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/tasks/main.yml
+++ b/playbooks/roles/edxapp/tasks/main.yml
@@ -152,6 +152,19 @@
     - install
     - install:base
 
+- name: add deadsnakes repo
+  apt_repository:
+      repo: ppa:deadsnakes/ppa
+
+- name: install python3.8
+  apt:
+    pkg:
+      - python3.8-dev
+      - python3.8-distutils
+  tags:
+    - install
+    - install:system-requirements
+
 - name: create log directories for service variants
   file:
     path: "{{ edxapp_log_dir }}/{{ item }}"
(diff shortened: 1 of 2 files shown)


playbooks/roles/edxapp/tasks/main.yml
```diff
@@ -152,6 +152,19 @@
     - install
     - install:base
 
+- name: add deadsnakes repo
+  apt_repository:
+      repo: ppa:deadsnakes/ppa
```
morenol, line 157: I think that we should add this repo if we are not in ubuntu focal, we could add that condition to this task.

morenol: I am sorry, I just noticed that this is a draft

mraarif:
I've filed an SRE ticket here: https://openedx.atlassian.net/browse/DOS-1555 to review these changes

mraarif:
there's a sandbox here: https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/33291/ that built successfully using this branch

playbooks/roles/edxapp/tasks/main.yml
jdmulloy: I think just adding deadsnakes all the time is easier. If we use the Ubnuntu repo the package name is ```python3```. Focal will likely be stuck on 3.8 forever and we'll be adding deadsnakes to get newer python versions, so I think it's simpler to just always install it and install the packages for the specific versions of python we want.

jdmulloy (approved):

adzuci (approved):
LGTM

PR #6064, opened 2020-10-07
WS-1406 part 1: add htaccess to secure-preview

rgraber:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
@@ -61,6 +61,22 @@ server {
     port_in_redirect off;
   }
 
+  location /secure-preview/ {
+    {% if PROSPECTUS_PREVIEW_ENABLE_BASIC_AUTH|bool %}
+      auth_basic            "Restricted";
+      auth_basic_user_file  {{ nginx_htpasswd_file }};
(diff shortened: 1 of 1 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -61,6 +61,22 @@ server {
...
 
+  location /secure-preview/ {
+    {% if PROSPECTUS_PREVIEW_ENABLE_BASIC_AUTH|bool %}
+      auth_basic            "Restricted";
+      auth_basic_user_file  {{ nginx_htpasswd_file }};
+    {% endif %}
+    port_in_redirect off;
+  }
+
+  location /es/secure-preview/ {
```
albemarle, line 72: Oh, hmm, I would have assumed the preview part of the URL goes first.  Is this the pattern for how we do es preview now?  If so we need to update our wiki for the PCs.

albemarle (approved):

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
rgraber: This is indeed the correct pattern. Good catch on updating the wiki.

PR #6065, opened 2020-10-08
Install mysql 5.7 on sandbox

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/mysql/defaults/main.yml
+++ b/playbooks/roles/mysql/defaults/main.yml
@@ -12,7 +12,7 @@ mysql_release_specific_debian_pkgs:
     - python3-mysqldb
 mysql_debian_pkgs: "{{ mysql_debian_pkgs_default + mysql_release_specific_debian_pkgs[ansible_distribution_release] }}"
 
-mysql_server_pkg: "mysql-server-5.7"
+mysql_server_pkg: "{{ 'mysql-server-5.7' if mysql_server_version_5_7 is defined and (mysql_server_version_5_7 | bool) else 'mysql-server-5.6' }}"
 mysql_server_5_7_pkg: "mysql-server=5.7.31-1ubuntu18.04"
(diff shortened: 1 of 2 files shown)


playbooks/roles/mysql/defaults/main.yml
```diff
@@ -12,13 +12,7 @@ mysql_release_specific_debian_pkgs:
...
 mysql_debian_pkgs: "{{ mysql_debian_pkgs_default + mysql_release_specific_debian_pkgs[ansible_distribution_release] }}"
 
-mysql_server_release_specific_pkg:
-  xenial:
-    - "mysql-server-5.6"
-  bionic:
-    - "mysql-server-5.7"
-
-mysql_server_pkg: "{{ mysql_server_release_specific_pkg[ansible_distribution_release] }}"
+mysql_server_pkg: "mysql-server-5.7"
```
adzuci: @nadeemshahzad is there any way to have both 5.6 and 5.7 running on sandboxes at the same time and do you know which Jenkins jobs rely on this role to install mysql on Jenkins and if we need to update MySQL on Jenkins? cc @fredsmith & @nedbat

adzuci (commented):
Not sure if we can merge this yet since some services have not been tested in devstack on MySQL 5.6 (mostly worried about the timestamp thing) and this role is used by things besides sandboxes like Jenkins, but once we sort out that all services run on MySQL 5.7 in devstack and sort out the Jenkins thing I think we may be able to merge it.

mikix (approved):
I mean change itself makes sense to me. Just a matter of when it's safe to do.

playbooks/roles/mysql/defaults/main.yml
nadeemshahzad: @adzuci I have updated PR to install MySQL 5.7 only to sandbox so this will not impact any other install of MySQL on jenkins.

adzuci (approved):

PR #6069, opened 2020-10-08
Create a new ES7 variable for forums.

dianakhuang:
Creates a new variable in ansible for forums. A corresponding PR in edx-internal is also being created.


playbooks/roles/forum/defaults/main.yml
```diff
@@ -61,6 +63,7 @@ forum_base_env: &forum_base_env
   SINATRA_ENV: "{{ FORUM_SINATRA_ENV }}"
   API_KEY: "{{ FORUM_API_KEY }}"
   SEARCH_SERVER: "{{ FORUM_ELASTICSEARCH_URL }}"
+  SEARCH_SERVER_ES7: "{{ FORUM_ELASTICSEARCH_URL_ES7 }}"
```
davidjoy, line 66: I think `SEARCH_SERVER_ES7` may need to be added to `devstack_forum_env` section as well, but with the `http://edx.devstack.elasticsearch7:9200/` URL instead.

davidjoy (changes requested):

playbooks/roles/forum/defaults/main.yml
dianakhuang: Thanks for catching that. Updated.

playbooks/roles/forum/defaults/main.yml
```diff
@@ -35,6 +35,8 @@ FORUM_ELASTICSEARCH_PORT: "9200"
 FORUM_ELASTICSEARCH_PROTOCOL: "http"
 FORUM_ELASTICSEARCH_URL: "{{ FORUM_ELASTICSEARCH_PROTOCOL }}://{{ FORUM_ELASTICSEARCH_HOST }}:{{ FORUM_ELASTICSEARCH_PORT }}"
 
+FORUM_ELASTICSEARCH_URL_ES7: "http://localhost:9200"
```
robrap: @dianakhuang: Is there a reason we collapsed host/port/protocol to URL? Note that we don't have a test here to match the old ES: https://github.com/edx/configuration/blob/b314385f3e468d62c05f278507138ef8cbc2ff84/playbooks/roles/forum/defaults/main.yml#L106-L107  Maybe this should be `FORUM_ELASTICSEARCH_HOST_ES7` instead?

dianakhuang: I thought it was unnecessarily complex to separate it out, but if we need to do that, we can do that.

playbooks/roles/forum/defaults/main.yml
```diff
@@ -101,3 +105,4 @@ FORUM_VERSION: "master"
 #
 forum_services:
   - {service: "elasticsearch", host: "{{ FORUM_ELASTICSEARCH_HOST }}", port: "{{ FORUM_ELASTICSEARCH_PORT }}"}
+  - {service: "elasticsearch7", host: "{{ FORUM_ELASTICSEARCH_HOST_ES7 }}", port: "{{ FORUM_ELASTICSEARCH_PORT }}"}
```
robrap, line 108: @feanil: Does this service name need to match something, and could this cause an issue?

feanil: It's only used to test connectivity to that service, so it should be safe to add a second item.

https://github.com/edx/configuration/blob/master/playbooks/roles/forum/tasks/test.yml#L7

robrap: Great.  Thanks.

robrap (approved):
@dianakhuang: It probably makes sense for you to give a thumb on this comment, or for someone else to approve my change.  I will add comments to the other PRs to update to use `FORUM_ELASTICSEARCH_HOST_ES7`.

dianakhuang:
👍 from me on these changes

robrap (approved):

davidjoy (approved):

PR #6075, opened 2020-10-15
WS-1395: prospectus preview 404 configurations

srwang:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
@@ -179,7 +179,7 @@ server {
   }
 
   location /preview/ {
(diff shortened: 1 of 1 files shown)


rgraber (commented):
I don't think we need the try_files here. I think it'd be worth testing locally to see what happens if we get rid of it.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -165,19 +169,13 @@ server {
   }
 
   location /preview/ {
-    {% if PROSPECTUS_PREVIEW_ENABLE_BASIC_AUTH|bool %}
```
rgraber: can you take out this change? i worry it will be confusing for someone looking at git history since it's in code we don't actually use (the if-drupal-is-not-deprecated-block)

rgraber (changes requested):

rgraber (approved):

PR #6093, opened 2020-10-26
Pin forums to an earlier, pre-ES7 version until we get all our services using ES7.

dianakhuang:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/forum/defaults/main.yml
+++ b/playbooks/roles/forum/defaults/main.yml
@@ -92,7 +92,8 @@ devstack_forum_env:
 forum_user: "forum"
 FORUM_RUBY_VERSION: "2.5.7"
 forum_source_repo: "https://github.com/edx/cs_comments_service.git"
-FORUM_VERSION: "master"
+# TODO: This is pinned to an earlier version of forums until we get native and other installations using ES7.
+FORUM_VERSION: "609eef0fa2f4c69dba92ce2ef98c557b876e77b9"
 
 #
 # test config


playbooks/roles/forum/defaults/main.yml
```diff
@@ -92,7 +92,8 @@ devstack_forum_env:
 forum_user: "forum"
 FORUM_RUBY_VERSION: "2.5.7"
 forum_source_repo: "https://github.com/edx/cs_comments_service.git"
-FORUM_VERSION: "master"
+# TODO: This is pinned to an earlier version of forums until we get native and other installations using ES7.
+FORUM_VERSION: "609eef0fa2f4c69dba92ce2ef98c557b876e77b9"
```
nedbat, line 96: When do we expect to get to ES7?

nedbat: Hmm, I tried this branch, and got the same failure, though I have to dig to see if ended up pinned properly.

nedbat: Oh: it didn't work because util/install/native.sh sets FORUM_VERSION to the release you are trying to install.  We need a different way to do this.

nedbat: @dianakhuang ^^

nedbat (changes requested):

playbooks/roles/forum/defaults/main.yml
dianakhuang: Hmm, I was wondering about that. I thought that if you didn't pass in a release that it would try to install the defaults. What version are you trying to install? `master`?

nedbat: Right, it would, but I always make a branch using the release-tagging tools, and then install from the branch.   This gives a better approximation to a real community installation.

mavidser:
Hi @dianakhuang - checking if there's any existing timeline for merging this PR :)

nedbat:
@mavidser We're working on how to properly get the version of forums that will work for Koa.

mavidser:
Hi @nedbat - checking if pinning the forum to pre-ES7 is still an option here (we can attempt a PR for this if that's still a valid solution) as the master branch [is still failing](https://groups.google.com/g/open-edx-btr-notifications/c/Qy68eTORzto) to provision correctly right now.

Though of course upgrading the repo to use ES7 would be much more cleaner, but I'm not sure what the timeline for that is. edx-platform seems to have ES7 support now (edx/edx-platform#25596), and going from the [devstack dependencies](https://github.com/edx/devstack/blob/33c792693d14188ba43673ffcdf2877e1259e725/docker-compose.yml#L201), only discovery remains on 1.5 - which did upgrade to use ES7 but reverted due to some issues (edx/course-discovery#2833)

nedbat:
@mavidser For the Koa branches, we are sticking with ES1.5, if that is what you are asking?  Master branches should be on ES7. /cc @dianakhuang

mavidser:
@nedbat @dianakhuang - Ah. I was mainly exploring ways to get the master configuration working again (as provisioning the stack from master is failing right now - forum needs ES7, `configuration` sets up ES1.5).

> Master branches should be on ES7. 

I assume this means that we should wait while this repo gets updated to install ES7 instead? (as opposed to the pinning solution in this PR)

nedbat:
I think that is right: master will not install cleanly on one machine right now, because we are mid-stream with the ES upgrade.

PR #6098, opened 2020-10-27
Removed all references to the notifier service DEPR-110

jmbowman:
Removing all references to the deprecated notifier service.

---


playbooks/roles/rabbitmq/templates/etc/logrotate.d/rabbitmq.j2
```diff
@@ -1,5 +1,5 @@
 # We want to hit the top level queues and any vhost queues
-# such as notifier and fulfillment
+# such as fulfillment
 {{ rabbitmq_log_dir }}/*.log {{ rabbitmq_log_dir }}/*/*.log {
```
robrap, line 3: Maybe a future DEPR, but why do we still have rabbitmq code?

util/jenkins/update_celery_monitoring/create_celery_dashboard.py
```diff
@@ -103,8 +103,8 @@ def generate_dashboard(environment, deploy):
     width = 24
     y_cord = 0
     region = "us-east-1"
-    right_axis_items=["edx.lms.core.ace", "edx.lms.core.background_process", "notifier.default"]
-    right_axis_items_age=["notifier.default"]
+    right_axis_items=["edx.lms.core.ace", "edx.lms.core.background_process"]
+    right_axis_items_age=[]
```
robrap, line 107: Seems odd now, but I guess it is correct.  Not sure if it should be deleted, so I understand why you left it, but maybe a comment about its history would help someone coming back to this?  Just a non-blocking thought.

robrap (approved):

playbooks/roles/rabbitmq/templates/etc/logrotate.d/rabbitmq.j2
jmbowman: I think a fair number of Open edX installations still use RabbitMQ; we didn't force the community to switch to Redis as celery broker when we did.

util/jenkins/update_celery_monitoring/create_celery_dashboard.py
jmbowman: Yeah, I was just playing it safe with this since I wasn't too familiar with the code.  Maybe @jdmulloy could speak to how useful (or not) it is to keep this variable.

PR #6083, opened 2020-10-20
BOM-2015 : Unpin virtualenv 

mzulqarnain1:
Relevant JIRA : https://openedx.atlassian.net/browse/BOM-2015

PR #6121, opened 2020-11-04
Remove symlinks to virtualenv binaries BOM-2015

jmbowman:
Symlinks to binaries in the virtualenv bin dir will stop working once we upgrade virtualenv due to changes in how it sets up the virtual environments (using some symlinks itself).  We had to revert the last upgrade attempt because we missed some of these, so trying to update them first this time.  This PR:

* Changes all use of such symlinks to use the link's target directly instead
* Stops creating these symlinks, since they'll stop working soon anyway

If we can get this deployed without incident, we should be clear to re-attempt the reverted https://github.com/edx/configuration/pull/6083 (after applying the configparser and zipp pins from https://github.com/edx/configuration/pull/6115 and https://github.com/edx/configuration/pull/6116).

More details on why such symlinks stopped working in virtualenv 20+:

* https://github.com/pypa/virtualenv/issues/1599
* https://github.com/pypa/virtualenv/issues/1773

---

### DIFF
--- a/playbooks/continuous_delivery/rollback_migrations.yml
+++ b/playbooks/continuous_delivery/rollback_migrations.yml
@@ -44,7 +44,7 @@
   become: true
 
   vars:
-    COMMAND_PREFIX: ". {{ APPLICATION_PATH }}/{{ APPLICATION_NAME }}_env; DB_MIGRATION_USER={{ DB_MIGRATION_USER }} DB_MIGRATION_PASS='{{ DB_MIGRATION_PASS }}' /edx/bin/python.{{ APPLICATION_NAME }} /edx/bin/manage.{{ APPLICATION_NAME }} "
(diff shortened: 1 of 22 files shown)


jmbowman:
@mzulqarnain1 Could you try this in a sandbox to see if it causes any problems?

mzulqarnain1 (approved):

mzulqarnain1:
@jmbowman Sandbox made successfully at https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/34465/

fredsmith (approved):

jmbowman:
@mzulqarnain1 We should be clear to re-attempt the changes from the 3 PRs mentioned in the description above, please prepare a new PR for that when you get a chance.

mzulqarnain1:
Sure @jmbowman

PR #6122, opened 2020-11-04
Remove rendering of obsolete json files

jdmulloy:
Configuration Pull Request
---


adzuci (approved):
There were some notes in OPS-3772 about a ticket being created that may have involved cleaning this up, but I think it can be removed.  If it turns out it was in use elsewhere we can likely fix forward.

adzuci:
@jdmulloy, I un-reviewed since I see @coryleeio found https://github.com/edx/configuration/pull/5414#issuecomment-588398918 which has more context that I don't entirely understand.  Could be worth asking someone from @bradenmacdonald's team to take a look.

jdmulloy:
@coryleeio @adzuci @bradenmacdonald 

I changed it so that it can be turned back on with a flag in case anyone needs it. Is this ok? Edxapp shouldn't even be reading these files any more, but it could be used if for some reason someone is using master of configuration with an older relase or edxapp.

adzuci (approved):
I like the flag idea.  Ship it.

coryleeio (approved):

bradenmacdonald:
@adzuci @jdmulloy Thanks for the ping. From looking at https://github.com/edx/configuration/blob/b36e34c0d9a4156c990b12167c668efcbb251353/playbooks/roles/edxapp/tasks/service_variant_config.yml#L20-L48 it seems that this code is still rendering the .yml files, but not the old .json files right? If so it should be totally fine with us, flag or not. As mentioned, edxapp doesn't even read the JSON.

Having the configuration end up only in a `.yml` file is fine with us, we just didn't want to remove from edx/configuration the code that creates the application [yml or json] config file(s) from ansible vars.

PR #6124, opened 2020-11-04
Really link pip to pip3

adzuci:


julianajlk (approved):

julianajlk:
This will point pip to pip 3, correct? The syntax is the later would create a link to pip3

PR #6131, opened 2020-11-06
BOM-2015 : Unpin virtualenv to latest version

mzulqarnain1:
Second attempt at unpinning virtualenv after #6083 was reverted. Now that #6121 is merged we are clear to attempt it again.


mzulqarnain1:
Sandbox made successfully at https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/34494/

util/install/native.sh
```diff
@@ -104,7 +104,7 @@ sudo apt-get install -y build-essential software-properties-common curl git-core
 sudo apt-get remove -y python-yaml
 sudo pip install --upgrade pip==20.0.2
 sudo pip install --upgrade setuptools==44.1.0
-sudo -H pip install --upgrade virtualenv==16.7.10
+sudo -H pip install --upgrade virtualenv==20.1.0
```
jmbowman, line 107: @nedbat I'm not sure if this will hit the same dependency resolution issue under Python 2.7 that forced us to pin `configparser` and `zipp` above in `playbooks/roles/common_vars/defaults/main.yml`.  Do you think we should pre-emptively pin them, or wait to see if it's a problem after this merges?

nedbat: For Koa, we are changing these lines to "pip3", so I don't think 2.7 is a concern any more.

jmbowman (approved):

kdmccormick:
We had to revert this again because instructor tasks weren't getting registered: #6141

PR #6146, opened 2020-11-12
Unpin virtualenv to latest version

mzulqarnain1:
Reverts edx/configuration#6141 which reverted #6137, which reverted #6135, which reverted #6131

Don't merge, Testing in progress


iamsobanjaved:
I have tested and verified that `Instructor Tasks` are running successfully on the sandbox created with this change

**Sandbox:** https://venvunpintesting.sandbox.edx.org/
**Flower:** http://100.26.141.67:5555/

image

mzulqarnain1:
FYI, @jmbowman and @kdmccormick

jmbowman:
Is there anything different here than the last time we merged?  If not, we may want to hold off until we make some progress on figuring out how to log the actual cause of task registration failures, since the problem has been intermittent even between different workers in the same deployment.

jmbowman:
The unregistered celery tasks issue was resolved in https://github.com/edx/edx-platform/pull/25840 , so we should be clear to try merging this again.

UsamaSadiq (approved):

jmbowman:
@nedbat Do you want to backport this to Koa, in case there's another broken pip or setuptools release before Lilac comes out?

PR #6150, opened 2020-11-16
ftgfi - migrate travis to github actions

fredsmith:


coryleeio (approved):

.github/workflows/playbook-test.yml
```diff
@@ -0,0 +1,30 @@
...
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
```
adzuci, line 12: Could we consider pegging this to `ubuntu-focal` or would we rather it not be pegged?

adzuci (approved):

PR #6184, opened 2020-12-01
docs: delete AUTHORS file

kdmccormick:
We have been doing this in every edX repo, as they've
fallen out of date and are no longer a useful rep-
resentation of who's worked on the project.

If you want to see a list of contributors, we
recommend deriving one from the Git history.


kdmccormick:
Dropping this PR because this repo will be removed eventually, anyway.

PR #6193, opened 2020-12-04
add basic newrelic logging support

fredsmith:


adzuci (approved):

playbooks/roles/newrelic_infrastructure/templates/etc/newrelic-infra/logging.d/logs.j2
```diff
@@ -0,0 +1 @@
+{{ NEWRELIC_LOGS | to_nice_yaml }}
```
adzuci: I had to look up what `to_nice_yaml` actually did: https://docs.ansible.com/ansible/latest/user_guide/playbooks_filters.html#formatting-data-yaml-and-json

adzuci (approved):

playbooks/roles/newrelic_infrastructure/templates/etc/newrelic-infra/logging.d/logs.j2
adzuci: I could also see us wanting to make this j2 template a little more detailed so that the complicated bit's don't need to live in each services edx-internal config: https://docs.newrelic.com/docs/logs/enable-log-management-new-relic/enable-log-monitoring-new-relic/forward-your-logs-using-infrastructure-agent#running-modes

image

playbooks/roles/newrelic_infrastructure/tasks/main.yml
```diff
@@ -114,6 +114,17 @@
     group: root
     mode: u=rwx,g=r,o=r
 
+- name: configure logging
+  template:
+    src: etc/newrelic-infra/logging.d/logs.j2
```
adzuci: Small nit, I know we're not consistent about this, but I prefer to name templates with the file extension, e.g. logs.yml.j2

playbooks/roles/newrelic_infrastructure/templates/etc/newrelic-infra/logging.d/logs.j2
fredsmith: so..  I want this PR to be simple.  I was thinking a global config that ships to all instances, since newrelic_infrastructure isn't included per-service.  individual services can also drop specific files into this directory if they want.

fredsmith: since "extra" logs defined in services won't be captured because the files don't exist, we can just ship the same config to every service.  We should therefore avoid blanket globs.

PR #6205, opened 2020-12-10
Adding new code for virtualenv management

christopappas:
# Configuration Pull Request

The idea here is that instead of using shiningpanda, you can do the following when you need virtualenvs in jenkins:
```
. /edx/var/jenkins/jobvenvs/virtualenv_tools.sh
# create a venv with its location stored in variable: venvpath
create_virtualenv --python=python3.8 --clear
. "$venvpath/bin/activate"

pip install blah
```
This avoids:
1. using shiningpanda, which does not support python 3.8
2. using virtualenvwrapper (which is used by shining panda underneath the hood and I am avoiding ENV var collisions while we phase shining panda out)
3. using regular virtualenv and hitting an issue where the shebang line generated is so long the OS cannot parse it
---


playbooks/roles/jenkins_common/defaults/main.yml
```diff
@@ -11,6 +11,7 @@ jenkins_common_protocol_https: true
 # When checking if Jenkins is finished initializing, expect a 200 as it should
 # be publicly available
 jenkins_common_ready_status_code: 200
+jenkins_common_job_venv_dir: ''
```
adzuci: do you know which Jenkins instances use jenkins_common?

adzuci (commented):
I'm not very familiar with what uses this role, if it is unlikely to be used by or brak build jenkins, then I think it is ok to put in the common defaults and tasks files.  cc @estute in case he remembers if build Jenkins or other Jenkins instances use this role.

fredsmith (approved):
:+1:

playbooks/roles/jenkins_common/defaults/main.yml
christopappas: ended up moving this stuff to jenkins_master and ran ansible against tools_jenkins.yml playbook and it got picked up (ideally having this in jenkins_master will also let this get picked up by analytics jenkins for their jobs too (in case they need to deprecate shiningpanda)

PR #6209, opened 2020-12-17
Improve scripts handing of Aurora

jdmulloy:
* Check both cluster and instance level parameter groups
* Remove the need to specify DB type, which was inaccurate anyways.
  It was really toggling between instance and cluster level parameter
  groups

Configuration Pull Request
---

### DIFF
--- a/util/check_rds_slow_query_logs/check_rds_slow_query_logs.py
+++ b/util/check_rds_slow_query_logs/check_rds_slow_query_logs.py
@@ -4,59 +4,58 @@ import boto3
 import click
 
 
-def get_db_instances(db_engine):
+def get_db_instances():
 
-    """ 
-    Returns: 
+    """
+    Returns:
           List of provisioned RDS instances
     """
+    return rds.describe_db_instances()['DBInstances']
 
-    if db_engine == "mysql":
-        instances = rds.describe_db_instances()['DBInstances']
-    elif db_engine == "aurora":
-        instances = rds.describe_db_clusters()['DBClusters']
-    return instances
+def get_db_clusters():
 
+    """
+    Returns:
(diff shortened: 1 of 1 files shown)


adzuci (commented):
Not sure why this isn't working, based on https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.describe_db_cluster_parameters vs https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.describe_db_parameters it looks like both fuctions have the same requirements and one only works for Aurora clusters.


Could you specify the `
DBParameterGroupName=
and
DBClusterParameterGroupName=
?

adzuci (approved):

jdmulloy:
> Not sure why this isn't working, based on https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.describe_db_cluster_parameters vs https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/rds.html#RDS.Client.describe_db_parameters it looks like both fuctions have the same requirements and one only works for Aurora clusters.
> 
> Could you specify the `
> DBParameterGroupName=
> and
> DBClusterParameterGroupName=
> ?

I'm a bit confused what you're asking. It is a bit confusing. I think the confusion is that the cluster level parameter group is only for Aurora, but the instance level ones apply to both Aurora and non-aurora, although the same one can't be used for both, each paramter group object has a type, like "mysql5.6", or "aurora" (Which really means Aurora MySQL 5.6), or "mysql5.7", or "aurora-mysql-5.7". But it's the same type of object, so in the API to look them up it's the same call.

PR #6239, opened 2021-01-12
discovery-worker config

muhammad-ammar:
Configuration Pull Request
---


adzuci (approved):
LGTM

playbooks/roles/edx_django_service/templates/edx/app/supervisor/conf.d.available/app-workers.conf.j2
```diff
@@ -34,7 +34,12 @@ directory={{ edx_django_service_code_dir }}
 stdout_logfile={{ supervisor_log_dir }}/%(program_name)s-stdout.log
 stderr_logfile={{ supervisor_log_dir }}/%(program_name)s-stderr.log
 
-command={{ edx_django_service_home }}/{{ edx_django_service_name }}-workers.sh worker -A {{ edx_django_service_name }} --app {{ edx_django_service_name }}.celery:app --loglevel=info --queue={{ w.queue }} --hostname={{ edx_django_service_name }}.{{ w.queue }}.%%h --concurrency={{ w.concurrency }} {{ '--maxtasksperchild ' + w.max_tasks_per_child|string if w.max_tasks_per_child is defined else '' }} {{ '--without-heartbeat' if not edx_django_service_celery_heartbeat_enabled|bool else '' }}
+{% set worker_app_name = edx_django_service_name %}
```
adzuci, line 37: I'm not sure if I understand why we need this if statement to map discovery to course_discovery here, could you explain how it works?

muhammad-ammar: @adzuci This is needed because the `app-workers.conf.j2` expects that value of `edx_django_service_name` is same as the directory name of the root in a particular repo. In case of course discovery, `edx_django_service_name` is set to `discovery` but in the course discovery repo the directory name is `course-discovery`. This is not the case with other repos like` registrar`, `enterprise_catalog` etc where the value of `edx_django_service_name` and directory name is same.

PR #6246, opened 2021-01-15
Bump xblock-problem-builder to 4.1.9 for deprecated import fix

kdmccormick:
```
The old version, 4.0.0, referenced an edx-platform path that
is deprecated and will soon be unsupported.
```

Package diff: https://github.com/open-craft/problem-builder/compare/v4.0.0...v4.1.9

Here is the specific change we need for the import fix: https://github.com/open-craft/problem-builder/commit/b7f7a047b85218876d97fb0f336d9c2ea92e9686

(we only need v4.1.8 for the import fix, but it seemed worth going all the way up to the latest version while we're upgrading anyway)

This is related to the [the impending removal of support for deprecated edx-platform import paths](https://github.com/edx/edx-platform/pull/25932).


kdmccormick:
Hey @jmbowman , could you either review this or point me to someone who might be suited to review it?

feanil (approved):

PR #6259, opened 2021-01-21
Add role to cluster task for create_rds playbook.

HassanJaveed84:
Configuration Pull Request
---

### DIFF
--- a/playbooks/create_rds.yml
+++ b/playbooks/create_rds.yml
@@ -72,5 +72,16 @@
     retries: 20
     delay: 60
 
+  - name: Add role to DB cluster
+    command: "aws rds add-role-to-db-cluster
+              --db-cluster-identifier {{ cluster_name }}
+              --role-arn {{ cluster_role_arn }}
+              "
+    register: add_role_result
+    failed_when: >
+      add_role_result.rc !=0 and ('DBClusterRoleAlreadyExists' not in add_role_result.stderr)
+    changed_when: "add_role_result.rc == 0"
+    when: cluster_name is defined and cluster_role_arn is defined
+
 - include: create_db_and_users.yml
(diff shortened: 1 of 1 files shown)


playbooks/create_rds.yml
```diff
@@ -72,5 +72,14 @@
...
     delay: 60
 
+  - name: Add role to DB cluster
+    command: "aws rds add-role-to-db-cluster
+              --db-cluster-identifier {{ cluster_name }}
+              --role-arn {{ cluster_role_arn }}
+              "
+    register: result
+    changed_when: "result.rc == 0"
+    when: cluster_name is defined and cluster_role_arn is defined
```
adzuci, line 84: After we run this the first time, should we add `and ('DBInstanceAlreadyExists' not in result.stderr)` to it or can we confirm that you think this command is idempotent?

adzuci (approved):
:shipit:

playbooks/create_rds.yml
HassanJaveed84: I can check that.

playbooks/create_rds.yml
```diff
@@ -72,5 +72,16 @@
...
     delay: 60
 
+  - name: Add role to DB cluster
+    command: "aws rds add-role-to-db-cluster
+              --db-cluster-identifier {{ cluster_name }}
+              --role-arn {{ cluster_role_arn }}
+              "
+    register: add_role_result
+    failed_when: >
+      add_role_result.rc !=0 and ('DBClusterRoleAlreadyExists' not in add_role_result.stderr)
```
HassanJaveed84, line 82: @adzuci I've added this so it will not fail if the role already exists.

adzuci (approved):

PR #6275, opened 2021-02-02
Adding Admin Portal and Learner Portal Enterprise MFEs to the sandbox

edx-abolger:
(per confluence instructions)

See related: https://github.com/edx/jenkins-job-dsl/pull/1289

Configuration Pull Request
---


playbooks/enterprise_catalog.yml
```diff
@@ -1,10 +1,12 @@
-- name: Deploy edX designer
+- name: Deploy edX enterprise catalog
```
edx-abolger, line 1: @adzuci  I am actually not at all sure this is needed anymore?

util/jenkins/ansible-provision.sh
```diff
@@ -307,6 +324,10 @@ PROSPECTUS_VERSION: $prospectus_version
 PROSPECTUS_ENABLED: $prospectus
 PROSPECTUS_SANDBOX_BUILD: True
```
edx-abolger: @adzuci I may need to take this part out as well given the changes in how the catalog was added?

CHANGELOG.md
```diff
@@ -19,6 +19,9 @@ Add any new changes to the top (right below this line).
     - Role: ecommerce
       - Added new configuration variable ECOMMERCE_EXTRA_CONFIG_OVERRIDES, which will allow override any ecommerce settings.
 
+ - 2021-02-11
```
adzuci: Thank you for adding this to the changelog!

playbooks/enterprise_catalog.yml
adzuci: Thanks for the copy pasta fix.  It may be helpful for Nadeem to take a look through this logic after it merges for extraneous code, but I thought that this playbook would be pulled in via https://github.com/edx/configuration/pull/6340/files, do you feel that there is duplicate logic elsewhere?

util/jenkins/ansible-provision.sh
```diff
@@ -307,6 +324,10 @@ PROSPECTUS_VERSION: $prospectus_version
...
 
+ENTERPRISE_CATALOG_VERSION: $enterprise_catalog_version
+ENTERPRISE_CATALOG_ENABLED: $enterprise_catalog
+ENTERPRISE_CATALOG_SANDBOX_BUILD: True
+
```
adzuci:
```suggestion
```

util/jenkins/ansible-provision.sh
adzuci: Yes, I think these three lines should be removed since they are covered by https://github.com/edx/configuration/pull/6275/files#diff-a119de766174cef876a3fc1d0f98540f6941bf7017feea5a268b61a116a8ff7eR365-R370

playbooks/enterprise_catalog.yml
```diff
@@ -1,10 +1,12 @@
...
   hosts: all
   become: True
   gather_facts: True
   vars:
     ENABLE_NEWRELIC: True
     CLUSTER_NAME: 'enterprise_catalog'
+    ENTERPRISE_CATALOG_SANDBOX_BUILD: False
```
adzuci: I can't find a use of this var in a when statement and see that we don't have it in the licence manager bit, so I believe it can be removed here and in the ansible-provision.sh file.

This is an example of a task where SANDBOX_BUILD variables used to be used in a when statement: https://github.com/edx/configuration/commit/f9f7fd8d831cd447007df3741bf141ad016590e7#diff-5b0009fde4f98a4a6d03f2b21b77cf9afa64ddb43488a7e83262fb6b4e4748e5L110

adzuci (approved):
:shipit: (after removing at least those three lines) I also added a few comments and think some cleanup may be warranted in a future PR.

adzuci:
*It may be worth rebasing this PR and building a sandbox off of the branch before merging.

playbooks/enterprise_catalog.yml
edx-abolger: By 'this' I meant my changes, not the playbook entirely.

PR #6287, opened 2021-02-09
Verified user will now be enrolled in verified mode

Dillon-Dumesnil:
Previously, the verified test user would be enrolled in the audit
mode in the demo course which made things difficult if you wanted to
test logic specific to verified learners. Now the verified learner
will be in the verified enrollment track from the start

Unclear if desired or not, but I have confirmed that this will not have any effect on existing users. So for the already created Verified user, they will remain in the Audit mode. On the next full provision (creating new data), the user will be created with the Verified enrollment (note: this is not to say that you should do that. Provisioning is expensive. Cheaper ways to get a verified learner exist via database updates or `update_enrollment` functions).

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/demo/defaults/main.yml
+++ b/playbooks/roles/demo/defaults/main.yml
@@ -25,22 +25,26 @@ demo_test_users:
     hashed_password: "{{ demo_hashed_password }}"
     is_staff: false
     is_superuser: false
+    mode: audit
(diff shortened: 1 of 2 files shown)


mikix (approved):

MatthewPiatetsky (approved):
seems good though I haven't worked in this specific code before, but this change sounds great!

PR #6292, opened 2021-02-16
Revert "Verified user will now be enrolled in verified mode"

Dillon-Dumesnil:
Reverts edx/configuration#6287

Sandbox builds began failing after this merged because the verified mode could not be found on the Demo course. Will investigate at a later point.

### DIFF
--- a/playbooks/roles/demo/defaults/main.yml
+++ b/playbooks/roles/demo/defaults/main.yml
@@ -25,26 +25,22 @@ demo_test_users:
     hashed_password: "{{ demo_hashed_password }}"
     is_staff: false
     is_superuser: false
-    mode: audit
   - email: 'audit@example.com'
     username: audit
     hashed_password: "{{ demo_hashed_password }}"
     is_staff: false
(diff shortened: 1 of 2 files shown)


MatthewPiatetsky (approved):

Dillon-Dumesnil:
I did some investigation into this and at this point, I do not think it is a reasonable lift to try and add in this ability. Ecommerce provisioning is what creates the Verified mode in the LMS. As such, the Ecommerce provisioning would need to happen prior to the test users enrolling. BUT the Ecommerce provisioning assumes the LMS is already provisioned with the course (the course and users in the LMS are all provisioned together). So then you get into a bit of a cycle if you want the verified learner. There is a management command to update enrollments so I think that will just need to be an acceptable solution for now. There does not seem to be any management command for adding in a Verified mode directly in the LMS (which after reflection might be because paid modes want/need to have associated product information about them so that's why ecommerce owns the creation).

Hope someone else is able to figure this out however!

PR #6293, opened 2021-02-19
Switch LMS urls to use localhost:18000 instead of edx.devstack.lms

jinder1s:
Setting the urls to edx.devstack.lms causes issues with redirects from frontend.

These settings are already set to localhost:18000 in edx-platform/lms/envs/devstack.py.
This just does it such that the value changes in edx-platform/lms/envs/production.py.

Configuration Pull Request
---

### DIFF
--- a/docker/build/edxapp/lms.yml
+++ b/docker/build/edxapp/lms.yml
@@ -346,7 +346,7 @@ JWT_PRIVATE_SIGNING_KEY: null
 LANGUAGE_CODE: en
 LANGUAGE_COOKIE: openedx-language-preference
 LEARNER_PORTAL_URL_ROOT: https://learner-portal-edx.devstack.lms:18000
-LMS_BASE: http://localhost:18000
+LMS_BASE: localhost:18000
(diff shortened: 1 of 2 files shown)


timmc-edx (approved):

docker/build/edxapp/lms.yml
```diff
@@ -346,9 +346,9 @@ JWT_PRIVATE_SIGNING_KEY: null
 LANGUAGE_CODE: en
 LANGUAGE_COOKIE: openedx-language-preference
 LEARNER_PORTAL_URL_ROOT: https://learner-portal-edx.devstack.lms:18000
-LMS_BASE: edx.devstack.lms:18000
-LMS_INTERNAL_ROOT_URL: http://edx.devstack.lms:18000
-LMS_ROOT_URL: http://edx.devstack.lms:18000
+LMS_BASE: http://localhost:18000
```
jmbowman: The previous value didn't include "http://"; should that still be omitted?

jinder1s: Yep, it does. I will remove that now

PR #6297, opened 2021-02-22
Revert "Add role to cluster task for create_rds playbook."

fredsmith:
Reverts edx/configuration#6259

This doesn't work and isn't the right place to capture this anyway.

### DIFF
--- a/playbooks/create_rds.yml
+++ b/playbooks/create_rds.yml
@@ -72,16 +72,5 @@
     retries: 20
     delay: 60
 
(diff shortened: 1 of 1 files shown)


adzuci (approved):
😿 We may want to figure out how to add something that accomplishes the original intent back in later.

fredsmith:
I'm going to leave this, but this whole playbook due to data inconsistency in our secure repo and either needs some defaults, or needs to be trashed.

PR #6314, opened 2021-03-10
add prospectus global 404 no-cache

fredsmith:


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -53,7 +61,6 @@ server {
       auth_basic_user_file  {{ nginx_htpasswd_file }};
     {% endif %}
     port_in_redirect off;
-    error_page 404 /404.html;
```
rgraber, line 56: this we need for gatsby!reasons

rgraber (dismissed):
I think we need the `error_page 404 /404.html` directives because of how gatsby does that page. the cache control thing i don't know about

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -105,7 +109,6 @@ server {
 
     add_header 'Cache-Control' 'public, max-age=0, must-revalidate';
     try_files $uri $uri/index.html =404;
-    error_page 404 /404.html;
   }
 
   # PROSPECTUS_STATIC_SITES will be a list of dictionaries which have a:
```
tuchfarber, line 114: If we're in the spirit of ripping things out, we should probably rank this copy/pasted test code.

fredsmith: I just merged that last week!

fredsmith: Sorry, I mean, I just merged code that configures that last week.  https://github.com/edx/edx-internal/pull/3053

adzuci (approved):

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
tuchfarber: But like... that regex was specific to the 2-layer learner portal MFE. I thought it was a test from a while ago that we abandoned. I have feelings about jamming this into prospectus, but it doesn't super matter as a test.

albemarle (approved):

PR #6315, opened 2021-03-11
Upgrade pip to 20.1.1

mzulqarnain1:
More details on https://openedx.atlassian.net/browse/BOM-2247

Sandbox configured successfully for all services here https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/38191/


jmbowman (commented):
Shouldn't `pre-requirements.txt` also be updated?  Also, see my note on the ticket about a handful of repos that still use pip-tools 4.x.

mzulqarnain1:
> Shouldn't `pre-requirements.txt` also be updated? Also, see my note on the ticket about a handful of repos that still use pip-tools 4.x.

Done.

jmbowman (approved):
Looks good as long as those repos using pip-tools 4.x have been updated.

PR #6324, opened 2021-03-22
remove unused confusing drupal fallback config

fredsmith:


adzuci (approved):
I like that this will reduce some confusion.  I am curious though, I don't think prospectus and sandboxes get along, but would this might make it harder for people to test things?

albemarle (approved):

fredsmith:
> I like that this will reduce some confusion. I am curious though, I don't think prospectus and sandboxes get along, but would this might make it harder for people to test things?

The only way the two configs were different were because of the old, no longer used, drupal config.  This shouldn't make a difference to sandboxes, who also have the same fallback.

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -78,6 +77,10 @@ server {
     add_header Cache-Control "no-store, max-age=0";
   }
 
+  location /images/ {
```
adzuci, line 80: Why is this being added in this PR?

fredsmith: because that was the thing that Joe was trying to make work, but couldn't because it was in the wrong block.

PR #6325, opened 2021-03-23
cache json, logos and icons more at the edge

fredsmith:


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -103,6 +107,16 @@ server {
       add_header 'Cache-Control' 'public, max-age=31536000, immutable';
     }
 
+  location / {
+    location ~ data\.json$ {
+      add_header 'Cache-Control' 'public, max-age=3600';
```
albemarle: No, data.json is used for rehydrating the page, I think we definitely don't want this.

albemarle (changes requested):
I don't think these are a good idea.  Let's discuss?

albemarle (approved):

PR #6342, opened 2021-03-30
allow edxapp to connect to TLS 1.0 resources

fredsmith:


adzuci (approved):
LGTM, we should carefully test this i stage.

PR #6350, opened 2021-04-07
docs: remove venv activation instructions

nedbat:
If one runs the Native Installation instructions, which includes running
ansible-bootstrap.sh, then activates the bootstrapped ansible
virtualenv, then proceeds to run native.sh, this will result in a
clobbered python3-pkg-resources package, necessitating its reinstallation.

Thus, we remove these instructions from the output of
ansible-bootstrap.sh.

(cherry picked from commit 3cc14096239e03b9a2ccc4a1f4232d71892c41a7)

Configuration Pull Request
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,13 @@
 # Changelog
-All notable changes to this project will be documented in this file.
-Add any new changes to the top(right below this line).
 
- - 2021-07-03
+All notable changes to this project will be documented in this file.
+Add any new changes to the top (right below this line).
+
+ - 2021-03-08
+    - Remove instruction from ansile-bootstrap.sh that instructed people to activate
+      the virtualenv.  This was incorrect for community installations.
+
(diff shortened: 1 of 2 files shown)


nedbat:
@fredsmith Thanks for the pre-review :)

@bmedx there was some thinking that Data Eng might still use this script? Is it OK to remove this message?

bmedx:
I'm unfamiliar with it, but I'll ask the team.

pwnage101:
I grepped my local checkouts of DE repos and didn't find any explicit calls to ansible-bootstrap.sh.  Also, the only time DE does any manual running of ansible is to update our jenkins config, which doesn't involve ansible-bootstrap.sh to my knowledge.

nedbat:
@pwnage101 so is that "ok to delete"? :)

pwnage101 (approved):
I was giving more team members more time to respond, but I feel fairly confident this is going to be okay.

PR #6385, opened 2021-04-26
fix: change old jenkins timestamp patterns to use ZZ

pwnage101:
Old jenkins uses an older version of Java's timestamp pattern format.
X or XX was introduced later (available in new jenkins).  I'm not really
sure if this is a Jenkins limitation or a Java JRE limitation.


pwnage101:
I already made the change on the jenkins configuration manually.

brianhw (approved):
LGTM.

PR #6386, opened 2021-04-26
fix: native.sh should exit with ansible's status

nedbat:
This script was always exiting with a success status, which made it
difficult to use it in a CI setting.  Now we exit with the status of the
ansible command.


nedbat:
https://github.com/edx/configuration/pull/6383#pullrequestreview-645189137 instead

PR #6389, opened 2021-04-28
feat: Install python3.8 on New DE Jenkins

jazibhumayun:
Python3.8 is not installed in New DE Jenkins, I have encountered some packages are not availble in < python3.8 in order to deploy prefect flows, so installing python3.8

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
+++ b/playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
@@ -36,6 +36,7 @@ de_jenkins_snap_pkgs:
 de_jenkins_python_versions:
   - python3.5
   - python3.7
+  - python3.8
 
 # Jenkins aliases to installed Python binaries
 de_jenkins_python_installations:


jazibhumayun:
@estute 
I have checked inside `/usr/bin/ `and there isn't python3.8 is available in New DE Jenkins.
The original error I was getting with python3.7 is `ImportError: cannot import name 'TypedDict' from 'typing' (/usr/lib/python3.7/typing.py)`

playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
```diff
@@ -36,6 +36,7 @@ de_jenkins_snap_pkgs:
 de_jenkins_python_versions:
   - python3.5
   - python3.7
+  - python3.8
```
estute, line 39: should we include python3.9 while we are at it? If it's easy, it would save some time in the future

estute (approved):

playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
jazibhumayun: Well I was able to run inside docker based testing environment. But unfortunately this python3.8 was not installed not even python3.7 I think these type of changes are reflected in docker based environment

jazibhumayun: I feel like adding python3.9 should work, that should be the only change required

jazibhumayun: Because whatever we will pass to this list will ultimately go into the following jenkins_common taks and gets installed using apt.
https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_common/tasks/main.yml#L40

jazibhumayun: Here is the new PR @estute 
https://github.com/edx/configuration/pull/6391

PR #6401, opened 2021-05-11
fix: better decoding of ansible errors

nedbat:
Some failed log lines say "FAILED", but some don't, depending on whether
the task is loop or not.  Sometimes an earlier FAILED message is
actually ignored, but is found by the old check.

Look for the leading "failed:" or "fatal:" lable instead to always get
the actual error that stopped the playbook.

Configuration Pull Request
---

### DIFF
--- a/util/install/native.sh
+++ b/util/install/native.sh
@@ -174,10 +174,8 @@ if [[ $ansible_status -ne 0 ]]; then
     echo "------------------------------------------------------------"
     echo " "
     echo "Decoded error:"
-    # Find the FAILED line before the "NO MORE HOSTS" line, and decode it.
-    # The plusses in the regex are because if I run this with -x, the awk line
-    # will be added to the log, and the regex would find itself if it didn't have plusses.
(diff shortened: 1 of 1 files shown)


nedbat:
@regisb This fixes the problem of finding the wrong error to decode.  Look ok?

regisb:
I honestly have no idea what's the impact of this change... My edx/configuration and awk skills are too poor. I trust you to do the right thing, though. And if this is not correct we'll find out sooner or later thanks to CI. Would you like to merge this to koa.master first, such that it's picked up by CI?

nedbat:
@BbrSofiane @cmltaWt0 what do you think?

stvstnfrd:
@nedbat The AWK here looks reasonable; I'm trying to spin up a VM to test this.
Will it be "easy" for me to reproduce _any_ of these failures?

nedbat:
@edx/site-reliability-engineering Anyone have a moment for an awk review?

nedbat:
I've cherry-picked this to Lilac.

PR #6414, opened 2021-05-20
build: add automerge workflow for open-release Maintainers

stvstnfrd:
Story
=====
As a core contributor and maintainer of the build-test-release group,
we want to be able to update named-release branches
without edx intervention
so we can lead in community-driven development of Named Releases.

As a core contributor's champion,
we want to delegate control of the named release branches,
without granting wider access to the repository
so we can empower community-driven development of Named Releases.

Behavior
========
This workflow only runs on PRs against a Named Release branch,
`open-release/*.master`. If the PR is from a user in the specified
maintainers group, we will automerge the PR once triggered by the
developer (by adding `[automerge]` to the PR description).
The workflow should also tag the community-engineering team, for the
time being.

### DIFF
--- a/.github/workflows/pr-automerge-open-release.yml
+++ b/.github/workflows/pr-automerge-open-release.yml
@@ -1,0 +1,86 @@
+# For non-draft changes to Named Release branches:
+# - Check if the user belongs to a maintainers team. If so,
+#   - approve the pull request
+#   - tag community-engineering
+#   - if it also has '[automerge]' in the PR body,
+#     label it as 'automerge', then merge it.
+---
+name: automerge BTR open-release PRs
+on:
+  pull_request_target:
+    branches:
+    - open-release/*.master
+    types:
+    - labeled
+    - unlabeled
+    - synchronize
+    - opened
+    - edited
+    - ready_for_review
+    - reopened
+    - unlocked
+  pull_request_review:
+    branches:
+    - open-release/*.master
+    types:
+    - submitted
+  check_suite:
+    branches:
+    - open-release/*.master
+    types:
+    - completed
+  status:
+    branches:
+    - open-release/*.master
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+    - run: echo "${{ toJSON(github) }}"
+    - run: echo "TEAM_CONTRIBUTORS=ceng-test" >> "${GITHUB_ENV}"
+    - run: echo "TEAM_CHAMPIONS=openedx/ceng-test" >> "${GITHUB_ENV}"
+    - run: echo 'TEAM_CHAMPIONS_ARRAY=["stvstnfrd"]' >> "${GITHUB_ENV}"
+    - name: lookup teams
+      id: teams
+      if: ${{ !github.event.pull_request.draft }}
+      uses: tspascoal/get-user-teams-membership@v1
+      with:
+        username: "${{ github.actor }}"
+        team: ${{ env.TEAM_CONTRIBUTORS }}
+        GITHUB_TOKEN: "${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}"
+    - name: tag team on all Named Release PRs
+      if: ${{ steps.teams.outputs.isTeamMember && github.event.action == 'opened' }}
+      run: |
+        curl \
+          -X POST \
+          -H 'Accept: application/vnd.github.v3+json' \
+          -H "authorization: Bearer ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}" \
+          https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/requested_reviewers \
+          -d '{"reviewers": ${{ env.TEAM_CHAMPIONS_ARRAY }}}' \
+        ;
+    - name: approve PR
+      if: ${{ !github.event.pull_request.draft && steps.teams.outputs.isTeamMember && github.event.action == 'opened' }}
+      uses: andrewmusgrave/automatic-pull-request-review@0.0.5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        event: APPROVE
+        body: |
+          :+1:
+
+          When you're ready to merge, edit the PR description and add this to it: `[automerge]`;
+          we'll handle the rest.
+          CC: @${{ env.TEAM_CHAMPIONS }}
+    - name: label PR as auto-mergeable
+      if: ${{ !github.event.pull_request.draft && steps.teams.outputs.isTeamMember && contains(github.event.pull_request.body, '[automerge]') }}
+      uses: andymckay/labeler@978f846c4ca6299fd136f465b42c5e87aca28cac
+      with:
+        add-labels: 'automerge'
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: automerge
+      uses: "pascalgn/automerge-action@v0.13.1"
+      env:
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+        MERGE_COMMIT_MESSAGE: |
+          merge: PR #{pullRequest.number}: {pullRequest.title}
+
+          {pullRequest.body}
--- a/.github/workflows/pr-automerge-open-release.yml
+++ b/.github/workflows/pr-automerge-open-release.yml
@@ -1,8 +1,8 @@
 # For non-draft changes to Named Release branches:
-# - Check if the user belongs to a maintainers team. If so,
-#   - approve the pull request
-#   - tag community-engineering
-#   - if it also has '[automerge]' in the PR body,
+# - Always tag community-engineering (for now).
+# - Check if the user belongs to a maintainers team.
+# - If so, approve the pull request.
+#   - If it also has '[automerge]' in the PR body,
 #     label it as 'automerge', then merge it.
 # Required organization secrets
 # - CC_GITHUB_TOKEN=...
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    types:
+    - completed
+  status:
+    branches:
+    - open-release/*.master
+jobs:
+  automerge:
+    runs-on: ubuntu-latest
+    steps:
+    - run: echo "${{ toJSON(github) }}"
```
stvstnfrd, line 34: TODO: Remove prior to merge.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    - run: echo "TEAM_CONTRIBUTORS=ceng-test" >> "${GITHUB_ENV}"
```
stvstnfrd: TODO: What group should we use for BTR maintainers? Is there an existing one already or shall we create one?

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    - run: echo "TEAM_CHAMPIONS=openedx/ceng-test" >> "${GITHUB_ENV}"
```
stvstnfrd: TODO: Change this to community-engineering prior to merge.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    - run: echo 'TEAM_CHAMPIONS_ARRAY=["stvstnfrd"]' >> "${GITHUB_ENV}"
```
stvstnfrd: ~TODO: Add the entire team here (yes, individually for now, see notes below re:reasoning).~

Done.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: lookup teams
+      id: teams
+      if: ${{ !github.event.pull_request.draft }}
```
stvstnfrd: Always check membership for non-drafts.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      uses: tspascoal/get-user-teams-membership@v1
+      with:
+        username: "${{ github.actor }}"
+        team: ${{ env.TEAM_CONTRIBUTORS }}
+        GITHUB_TOKEN: "${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}"
```
stvstnfrd: We need to try using an available PAT (personal access token) [1] as the default `GITHUB_TOKEN` does _not_ seem to work properly for team-based API calls. All other steps work w/ the default provided `GITHUB_TOKEN`, but this step (team lookup) and the next (tagging `team_reviewers`) don't work. This happened both with Marketplace actions, as well as with hitting the API directly. 

I could not find good documentation to account for this, so it's possible I was doing something incorrectly, though I did iterate quite a bit :shrug: 

- [1] https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token

stvstnfrd: TODO: We should probably rename `GH_TOKEN` to something more specific and then set it as an organization secret, instead of a per-repo secret.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: tag team on all Named Release PRs
+      if: ${{ steps.teams.outputs.isTeamMember && github.event.action == 'opened' }}
+      run: |
+        curl \
+          -X POST \
+          -H 'Accept: application/vnd.github.v3+json' \
+          -H "authorization: Bearer ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}" \
+          https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/requested_reviewers \
+          -d '{"reviewers": ${{ env.TEAM_CHAMPIONS_ARRAY }}}' \
```
stvstnfrd: Note: The API docs [1] say this should also take a `team_reviewers` array, but I couldn't get this working. I tried both with and without the organization prefix `openedx/*`. Sometimes I got an error, others it gave me a valid response, but never actually updated the reviewers list :shrug: 

I'd prefer to have this tag the team, but deferred to just tagging us all individually for now.

- [1] https://docs.github.com/en/rest/reference/pulls#request-reviewers-for-a-pull-request

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      uses: andrewmusgrave/automatic-pull-request-review@0.0.5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        event: APPROVE
+        body: |
+          :+1:
+
+          When you're ready to merge, edit the PR description and add this to it: `[automerge]`;
+          we'll handle the rest.
+          CC: @${{ env.TEAM_CHAMPIONS }}
```
stvstnfrd: Caveat: This doesn't work as well as I'd expect. It _does_ enter the proper text in the comment body (`@teamname`), but for some reason it doesn't register as a linked team/user name, so it doesn't actually notify the group explicitly :shrug: 

If I had got this working (of if we think this is good enough for now), we could (have) drop(ped) the individual review request(s) above. With this (assuming you're subscribed to the repo, and we're all auto-subscribed to everything), you can flag emails by filtering emails like `@openedx/community-engineering` but _not_ w/ `CC: team_review@noreply.github.com`.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      run: |
+        curl \
+          -X POST \
+          -H 'Accept: application/vnd.github.v3+json' \
+          -H "authorization: Bearer ${{ secrets.GH_TOKEN || secrets.GITHUB_TOKEN }}" \
+          https://api.github.com/repos/${{ github.repository }}/pulls/${{ github.event.pull_request.number }}/requested_reviewers \
+          -d '{"reviewers": ${{ env.TEAM_CHAMPIONS_ARRAY }}}' \
+        ;
+    - name: approve PR
+      if: ${{ !github.event.pull_request.draft && steps.teams.outputs.isTeamMember && github.event.action == 'opened' }}
```
stvstnfrd: Go ahead and :+1:  the PR immediately and provide feedback/instructions, but we'll hold off on actually merging until flagged by the developer.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        event: APPROVE
+        body: |
+          :+1:
+
+          When you're ready to merge, edit the PR description and add this to it: `[automerge]`;
+          we'll handle the rest.
+          CC: @${{ env.TEAM_CHAMPIONS }}
+    - name: label PR as auto-mergeable
+      if: ${{ !github.event.pull_request.draft && steps.teams.outputs.isTeamMember && contains(github.event.pull_request.body, '[automerge]') }}
```
stvstnfrd: Wait until the body contains `[automerge]`, to require some manual intervention/intent.
BTW: I went with `body` instead of `title` because some email clients hiccup on threading if the subject line changes mid-PR (`Re: <PR-TITLE>`)

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+      uses: andymckay/labeler@978f846c4ca6299fd136f465b42c5e87aca28cac
+      with:
+        add-labels: 'automerge'
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+    - name: automerge
```
stvstnfrd, line 79: This only ever executes iff the the `automerge` label has been applied.

.github/workflows/pr-automerge-open-release.yml
```diff
@@ -0,0 +1,86 @@
...
+    - name: approve PR
+      if: ${{ !github.event.pull_request.draft && steps.teams.outputs.isTeamMember && github.event.action == 'opened' }}
+      uses: andrewmusgrave/automatic-pull-request-review@0.0.5
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        event: APPROVE
+        body: |
+          :+1:
+
+          When you're ready to merge, edit the PR description and add this to it: `[automerge]`;
```
stvstnfrd, line 64: Add a +1 that explains how to self-merge.

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: TODO: Should this be an org-secret instead, to avoid setting it up for every repo?

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: TODO: Should this be an org-secret instead, to avoid setting it up for every repo?

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: ~TODO: Should this be an org-secret instead, to avoid setting it up for every repo?~

Done.

stvstnfrd: TODO: Remove this if the "half" at-mention below is enough.

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: TODO: Create a token on an actual bot account.

I've been using a _personal_ personal token for testing, but we should either add a token to an existing bot account for use or create a new one.

Is there an existing bot account we should use? `openedx-release-bot`?

stvstnfrd (commented):
Pre-review

.github/workflows/pr-automerge-open-release.yml
sarina: I think this should be an org-secret, but can we wait to do that as follow-on work?

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: Done.

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: Moved to secrets.

.github/workflows/pr-automerge-open-release.yml
stvstnfrd: Done.

stvstnfrd:
@edx/community-engineering code-wise, I think this is ready to go.

The outstanding bits relate to Github (Organization) Secrets:
- set the secrets, as described in the comments, for the `edx` organization (done on `openedx` for testing)
- decide what values to use for:
  - the BTR team used to decide membership
    - @nedbat @sarina is there an existing team for this or should I create a new one?
  - the bot account used for `CC_GITHUB_TOKEN`
    - we need to generate and specify an access token from an edx bot account
      - is there a pre-existing bot account we should use? I know there are PR bots in existence, but I don't have access, so hard to verify options

sarina:
@stvstnfrd I don't think there's an existing group; for the bot accout I'm not sure who owns our existing bots but my guess is reaching out in `#arch-bom` would yield results

PR #6444, opened 2021-06-14
fix:pass celery always_eager flag as it is consumed by celery4

matthugs:
Based on
https://github.com/celery/celery/blob/04d921b3fbcd9cc42285edeb916c41df507fe691/docs/history/whatsnew-4.0.rst
this and several other celery settings changed in the course of our
py3.8 upgrade. This is one which matters a great deal for
sandboxes (which shouldn't rely on celery for the execution of asynch
tasks). [There are some more which I'll follow up with](https://github.com/edx/configuration/pull/6445) after I
understand better why their apparent lack isn't impacting production
builds' operation of celery.

  - [x] ~A DevOps team member has approved the PR if it is code shared across multiple services and you don't own all of the services.~ We do own all the services


alangsto (approved):
👍

PR #6445, opened 2021-06-14
fix: follow-up from #6444

matthugs:
Follow-up from https://github.com/edx/configuration/pull/6444

  - [x] ~A DevOps team member has approved the PR if it is code shared across multiple services and you don't own all of the services.~ we own all the services


matthugs:
I'm more confident that this'll be fine because these values were being applied from here: https://github.com/edx/registrar/commit/96791b9e077eb7115036c397f30b9000f81d1350#diff-ed30bf02132d33e1f8187babfd59d7cbda05c2885d79334eb382722bb953423eR129

They should've also been changed here & in remote-config. Let me update the remote config values first before trying to merge this.

bseverino (approved):

PR #6458, opened 2021-06-25
fix: Suppress logging when generating EC2 keypair (was logging private key)

timmc-edx:

  - [ ] Consider whether this should be backported to Lilac
  - [x] N/A: Are you adding any new default values that need to be overridden when this change goes live? If so [...]

### DIFF
--- a/playbooks/continuous_delivery/launch_instance.yml
+++ b/playbooks/continuous_delivery/launch_instance.yml
@@ -46,6 +46,7 @@
       name: "{{ automation_prefix }} {{ unique_key_name.stdout }}"
       region: "{{ ec2_region }}"
     register: ssh_key_register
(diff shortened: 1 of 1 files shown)


playbooks/continuous_delivery/launch_instance.yml
```diff
@@ -46,6 +46,7 @@
       name: "{{ automation_prefix }} {{ unique_key_name.stdout }}"
       region: "{{ ec2_region }}"
     register: ssh_key_register
+    no_log: True
```
adzuci, line 49: I always forget if it should be true or True, but both are truthy anyways and I think True is correct.

adzuci (approved):
:shipit:

PR #6491, opened 2021-07-28
BOM-2715: Install py38 requirements in edxapp-sandbox

UsamaSadiq:
**Issue:** [BOM-2715](https://openedx.atlassian.net/browse/BOM-2715)

### Description
- Use `requirements/edx-sandbox/py38.txt` as the sandbox requirements instead of `py35.txt`.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -1690,7 +1690,7 @@ base_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/base.txt"
 django_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/django.txt"
 openstack_requirements_file: "{{ edxapp_code_dir }}/requirements/edx/openstack.txt"
 
-sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/py38.txt"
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -1690,7 +1690,7 @@ base_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/base.txt"
 django_requirements_file:   "{{ edxapp_code_dir }}/requirements/edx/django.txt"
 openstack_requirements_file: "{{ edxapp_code_dir }}/requirements/edx/openstack.txt"
 
-sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/{% if edxapp_sandbox_python_version == 'python2.7' %}base.txt{% else %}py35.txt{% endif %}"
+sandbox_base_requirements:  "{{ edxapp_code_dir }}/requirements/edx-sandbox/py38.txt"
```
jmbowman: Should we instead just update the condition to distinguish between 3.5 and 3.8 instead?  That would let us control the version just with remote config changes, without needing to deploy or roll back configuration repo changes.

UsamaSadiq: This check will probably be unnecessary after we complete the upgrade to `Python38` but looks helpful for now. I've updated the condition as per suggestion and we can update it later on after we've successfully upgraded the codejail environment to `Python3.8`.

jmbowman (approved):

PR #6493, opened 2021-07-29
fix: Add jenkins user in docker group

jazibhumayun:
In order to run and build docker images user that runs docker should be added in docker group that mapping was missing. DENG-934

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_data_engineering_new/meta/main.yml
+++ b/playbooks/roles/jenkins_data_engineering_new/meta/main.yml
@@ -51,3 +51,8 @@ dependencies:
     jenkins_common_python_installations: '{{ de_jenkins_python_installations }}'
     jenkins_common_snap_pkgs: '{{ de_jenkins_snap_pkgs }}'
     jenkins_common_timestamper_system_clock_format: '{{ de_jenkins_timestamper_system_time }}'
+
+  # Needed to be able to build docker images. Used by Docker Image Builder Jobs.
+  - role: docker-tools
+    docker_users:
+        - '{{ jenkins_user }}'


estute (commented):
I am pretty sure we already do this. See these links:
https://github.com/edx/configuration/blob/master/playbooks/jenkins_data_engineering_new.yml#L36

https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_data_engineering_new/defaults/main.yml#L10

https://github.com/edx/configuration/blob/56b32423efc3b29309f4d6e475dd296e13c55cad/playbooks/roles/jenkins_data_engineering_new/tasks/system.yml#L9

This pattern was copied from the build jenkins worker, which is able to use docker without the change you are proposing. Maybe something changed?

adzuci (approved):
:shipit:

jazibhumayun:
> I am pretty sure we already do this. See these links:
> https://github.com/edx/configuration/blob/master/playbooks/jenkins_data_engineering_new.yml#L36
> 
> https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_data_engineering_new/defaults/main.yml#L10
> 
> https://github.com/edx/configuration/blob/56b32423efc3b29309f4d6e475dd296e13c55cad/playbooks/roles/jenkins_data_engineering_new/tasks/system.yml#L9
> 
> This pattern was copied from the build jenkins worker, which is able to use docker without the change you are proposing. Maybe something changed?

I took reference from here: [https://github.com/edx/configuration/blob/master/playbooks/roles/tools_jenkins/meta/main.yml#L15-L17](https://github.com/edx/configuration/blob/master/playbooks/roles/tools_jenkins/meta/main.yml#L15-L17)

It seems both of these things are trying to do same i.e. "adding jenkins user in docker".
But what's currently in master right now in jenkins analytics play does not add jenkins user in docker group so I thought pass list of users to docker-tools role: [https://github.com/edx/configuration/blob/master/playbooks/roles/docker-tools/tasks/main.yml#L11-L19](https://github.com/edx/configuration/blob/master/playbooks/roles/docker-tools/tasks/main.yml#L11-L19)

One thing I also noticed [https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_data_engineering_new/defaults/main.yml#L8-L10](https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_data_engineering_new/defaults/main.yml#L8-L10)
`jenkins_group` variable is defined twice here.

jazibhumayun:
one reason why it is not working properly can be because `jenkins` installation happens first and at that time docker group is not available, later on it installs docker creates docker group but as the `jenkins` play ran already before even docker was installed.

jazibhumayun:
@estute can you also point me where exactly (which lines) `jenkins_groups` is used in ansible task ?
I searched in [https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_data_engineering_new/tasks/main.yml](https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_data_engineering_new/tasks/main.yml) and [https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_common/tasks/main.yml](https://github.com/edx/configuration/blob/master/playbooks/roles/jenkins_common/tasks/main.yml) but couldn't find it is these files

PR #6506, opened 2021-08-10
fix: Add retry logic to add apt repository tasks

nadeemshahzad:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/common/tasks/main.yml
+++ b/playbooks/roles/common/tasks/main.yml
@@ -59,11 +59,21 @@
 - name: Add git apt repository
   apt_repository:
     repo: "{{ common_git_ppa }}"
+    update_cache: yes
+  register: add_repo
+  until: add_repo|success
+  retries: 10
+  delay: 5
   when: ansible_distribution in common_debian_variants
 
 - name: Add ppa for watchman package
   apt_repository:
     repo: "ppa:linuxuprising/apps"
+    update_cache: yes
+  register: add_repo
+  until: add_repo|success
+  retries: 10
+  delay: 5
   when: >
       ansible_distribution in common_debian_variants and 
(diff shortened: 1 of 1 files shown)


playbooks/roles/common/tasks/main.yml
```diff
@@ -100,6 +105,11 @@
 - name: add deadsnakes repository
   apt_repository:
     repo: "ppa:deadsnakes/ppa"
+    update_cache: yes
```
adzuci, line 113: Can you also add this here? https://github.com/edx/configuration/pull/6506/files#diff-6b9086567f8ddca7d54ce7bf6b8e89b7974a1b65edefb76f3bd2087d281c4b7eR71

playbooks/roles/common/tasks/main.yml
```diff
@@ -59,6 +59,11 @@
 - name: Add git apt repository
   apt_repository:
     repo: "{{ common_git_ppa }}"
+    update_cache: yes
```
adzuci, line 62: This will be pretty slow right?  Do you know if this was happening before?

NIXKnight: According to the module docs by default `update_cache` is set to `yes`. I believe that adding the parameter explicitly will cause no harm.

adzuci: Nvm, this was already yes and you are just making explicit

adzuci (approved):
:shipit:

playbooks/roles/common/tasks/main.yml
matthugs: I'm not sure that's true.

Screen Shot 2021-08-12 at 1 15 10 PM
from https://docs.ansible.com/ansible/2.8/modules/apt_module.html?highlight=apt

PR #6507, opened 2021-08-10
Revert "feat: Move prospectus to install Node directly."

dianakhuang:
Reverts edx/configuration#6505

Broken build example: https://gocd.tools.edx.org/go/tab/build/detail/stage-prospectus/1012/run_play/1/stage_edx


adzuci (approved):
:shipit:

matthugs (approved):

PR #6508, opened 2021-08-11
feat: Added helm checker script

syedimranhassan:
Configuration Pull Request
---

### DIFF
--- a/util/jenkins/helm_update_checker/helm_update_checker.py
+++ b/util/jenkins/helm_update_checker/helm_update_checker.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python
+import subprocess
+import yaml
+import sys
+import logging
+import click
+import os
+import json
+import boto3
+import backoff
+from botocore.exceptions import ClientError
+
+
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig()
+global_list = []
+MAX_TRIES = 5
+
+
+class SESBotoWrapper:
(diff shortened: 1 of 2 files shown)


util/jenkins/helm_update_checker/helm_update_checker.py
```diff
@@ -0,0 +1,94 @@
...
+import sys
+import logging
+import click
+
+LOGGER = logging.getLogger(__name__)
+logging.basicConfig()
+global_list = []
+
+"""
+This function except a filename.
```
christopappas: Did you mean "accepts" and not "except"?

Also, this comment should be under `def parse_yaml(file_name):`

util/jenkins/helm_update_checker/helm_update_checker.py
```diff
@@ -0,0 +1,94 @@
...
+
+def compare_version(current_version, latest_version):
+    if current_version == latest_version:
+        return True
+    return False
+
+
+@click.command()
+@click.option('--file-name', required=True, help='Filename which have helm chart details.')
+def controller(file_name):
```
christopappas: do we need to update the yaml file every time we add a new helm based k8s app to edx-internal? or is there a way for this to be generalized so that we don't need to hardcode the names in a config file? in other words, is it possible for the python code to look at files in the directory and determine which helm apps are present, and if something was missing, it automatically adds it to the config file?

arbabkhalil (approved):

PR #6526, opened 2021-08-24
Chore: Python requirements update (PSRE-824)

nadeemshahzad:
Configuration Pull Request
---


NIXKnight (approved):

.github/workflows/playbook-test.yml
```diff
@@ -13,7 +13,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.5, 3.8]
+        python-version: [3.8]
```
adzuci, line 16: +1 to removing 3.5 tests, though I wonder if we should also add a change log stating that our ansible playbooks no longer support 3.5 after Maple? 
(also checked other apps to which are only testing 3.8: https://github.com/edx/edx-notes-api/blob/master/.github/workflows/ci.yml#L16)

doctoryes:
@nadeemshahzad This PR seems to be the cause for the recent blockstore deployment failure here:
https://gocd.tools.edx.org/go/tab/build/detail/stage-blockstore/27/run_play/1/stage_edx
which seems to still install the AWS requirements with pip2?

doctoryes:
Nevermind - the blockstore pipeline was using Ubuntu 18.04 to build the AMI instead of 20.04 - fixed here:
https://github.com/edx/edx-internal/pull/5388

PR #6532, opened 2021-08-25
feat: Studio uses oauth2 to login

jinder1s:
Settings to enable oauth2 login through lms for studio in sandbox.

### DIFF
--- a/playbooks/roles/edxapp/defaults/main.yml
+++ b/playbooks/roles/edxapp/defaults/main.yml
@@ -228,10 +228,11 @@ EDXAPP_THIRD_PARTY_AUTH_BACKENDS:
   - common.djangoapps.third_party_auth.saml.SAMLAuthBackend
   - common.djangoapps.third_party_auth.lti.LTIAuthBackend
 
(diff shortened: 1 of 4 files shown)


dianakhuang (approved):

playbooks/roles/edxapp/defaults/main.yml
```diff
@@ -228,6 +228,12 @@ EDXAPP_THIRD_PARTY_AUTH_BACKENDS:
   - common.djangoapps.third_party_auth.saml.SAMLAuthBackend
   - common.djangoapps.third_party_auth.lti.LTIAuthBackend
 
+EDXAPP_CMS_SERVICE_NAME: 'edxapp-cms'
```
timmc-edx, line 231: How about `studio` rather than `edxapp-cms` for these names? Would match devstack.

jinder1s: I'm following the convention in configuration repo that uses "edxapp-lms". Essentially, when in rome, do as the romans do

timmc-edx: Fair enough! Better to match the local than the remote, probably.

PR #6547, opened 2021-09-13
Feat: Add group and user management tasks for other IDAs

NIXKnight:
Configuration Pull Request
---

### DIFF
--- a/playbooks/manage_edxapp_users_and_groups.yml
+++ b/playbooks/manage_edxapp_users_and_groups.yml
@@ -115,7 +115,7 @@
     ignore_user_creation_errors: no
     deployment_settings: "{{ EDXAPP_SETTINGS | default('production') }}"
     group_environment: ""  # By default, create groups for all envs (for backwards compatibility).
-    service: ""
+    service: "" # Used to display the service name during execution with *-ida tags. Set using ansible-playbook -e.
   vars_files:
     - roles/common_vars/defaults/main.yml
   tasks:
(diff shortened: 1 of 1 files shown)


playbooks/manage_edxapp_users_and_groups.yml
```diff
@@ -115,6 +115,7 @@
     ignore_user_creation_errors: no
     deployment_settings: "{{ EDXAPP_SETTINGS | default('production') }}"
     group_environment: ""  # By default, create groups for all envs (for backwards compatibility).
+    service: ""
   vars_files:
```
adzuci, line 119: Could we add a comment mentioning what this will do if it's left blank?  (Or what it won't do or what the example options are)

playbooks/manage_edxapp_users_and_groups.yml
```diff
@@ -186,3 +187,58 @@
...
+      with_items: "{{ django_groups }}"
+      when: (not group_environment) or group_environment in item.environments
+      become: true
+      become_user: "{{ common_web_user }}"
+
+    - name: Manage active {{ service }} users
+      tags:
+        - manage-active-users-ida
+      shell: >
+        . {{env_path}} && {{ python_path }} {{ manage_path }} --settings={{ deployment_settings }}
```
adzuci, line 208: Wow this is a lot of if blocks.

@estute This was similar to what I was thinking of for another script we were working on.  That other script isn't in an Ansible playbook though so I don't actually think we could have done this unfortunately.

adzuci (approved):
LGTM

playbooks/manage_edxapp_users_and_groups.yml
NIXKnight: Added a comment

PR #6612, opened 2021-11-11
feat: tableau DE ansible role 

jazibhumayun:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/tableau_de/meta/main.yml
+++ b/playbooks/roles/tableau_de/meta/main.yml
@@ -1,0 +1,3 @@
+---
+dependencies:
+  - aws_cloudwatch_agent
--- a/playbooks/roles/tableau_de/tasks/main.yml
+++ b/playbooks/roles/tableau_de/tasks/main.yml
@@ -1,0 +1,59 @@
+---
+
+- name: Install AWS CLI
+  apt:
+    name: awscli
+    state: present
+  tags:
+    - install
+    - install:base
+
+- name: Add OS default user in tableau group
+  shell: "usermod -a -G {{ tableau_user_group }} {{ os_default_user }}"
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Enable tableau user shell
+  shell: "usermod -s /bin/bash {{ tableau_user }}"
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Create directory for Postgre driver if it does not exist
+  file:
+    path: /opt/tableau/tableau_driver/jdbc
+    state: directory
+    mode: '0777'
+  tags:
+    - install
+    - install:base
+
+- name: Download Postgre driver jar file
+  get_url:
+    url: "{{postgre_driver_endpoint}}/postgresql-{{postgre_driver_version}}.jar"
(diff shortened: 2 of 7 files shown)


playbooks/roles/tableau_de/meta/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+# dependencies:
```
estute: Should we include this if there are no dependencies?

playbooks/roles/tableau_de/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+
+- name: Install Snowflake odbc driver
+  dpkg_selections:
+    name: snowflake-odbc-{{snowflake_driver_version}}.x86_64.deb
+    selection: install
+  tags:
+    - install
+    - install:base
+
+- name: Install Collectd to use with AWS CloudWatch Agent
```
estute: I think it makes sense to separate the Cloudwatch Agent installation/configuration into a separate role. That way, we can install it onto other pieces of infrastructure. I think you could create a separate role and call it in via a `meta` file, overriding default values with any specific to this Tableau instance. What do you think?

playbooks/roles/tableau_de/tasks/main.yml
```diff
@@ -0,0 +1,120 @@
...
+  apt: pkg=collectd={{ collectd_version }}
+       install_recommends={{ collectd_install_recommends }}
+       state=present
+  tags:
+    - install
+    - install:base
+
+- name: Install xz-utils (required when using deb parameter of apt module)
+  package:
```
estute: Why are you using `package` here, but using `apt` elsewhere in this file?

estute (commented):
In general looks good, although I think it would make sense to split out the AWS stuff into a separate role. Let me know what you think and we can either do that or I'll thumb it.

playbooks/roles/tableau_de/tasks/main.yml
```diff
@@ -0,0 +1,59 @@
+---
+
+- name: Install AWS CLI
+  apt:
+    name: awscli
+    state: present
+  tags:
+    - install
+    - install:base
+
```
jazibhumayun, line 10: I kept the AWS CLI installation in tableau_de because it is not related to AWS Cloudwatch agent.

playbooks/roles/tableau_de/tasks/main.yml
jazibhumayun: I liked your suggestion. I have created separated role for Cloudwatch agent installation. This role should be useful when we want to install Cloudwatch agent in other servers.

playbooks/roles/tableau_de/tasks/main.yml
jazibhumayun: Ohh yes good catch. It was useful when I was testing it for `CentOS` since the play is written for `Ubuntu` so we can be specify apt.

jazibhumayun:
Its ready for second review @estute

playbooks/roles/tableau_de/tasks/main.yml
estute: Oh, that's right, it's used by the backups. Thanks

estute (approved):
Thanks for refactoring it. I imagine at some point we may want to parameterize the metrics collection/aggregation intervals, but that's for another time.

PR #6637, opened 2021-12-14
Add retry logic to AMI registration

marekwrobel:
We noticed that sometimes the AMI creation takes some longer time and stays in 'pending' till finally times out. 

Instead of having to manually trigger the build stage to retry, I added a retry logic for that. 

Configuration Pull Request
---

### DIFF
--- a/playbooks/continuous_delivery/create_ami.yml
+++ b/playbooks/continuous_delivery/create_ami.yml
@@ -70,6 +70,9 @@
                 'deployment':'{{ deployment }}'
             }"
     register: ami_register
+    retries: 5
+    delay: 10
+    until: ami_register is succeeded
 
   - name: Add any tags that are on the instance to the AMI
     ec2_tag:


adzuci (approved):
I believe this looks good to me, it's hard for me to test locally, so I think we should merge it, test a deploy in another service and then revert if needed.  What do you think @arbabkhalil ?

arbabkhalil (approved):
LGTM

playbooks/continuous_delivery/create_ami.yml
```diff
@@ -70,6 +70,9 @@
                 'deployment':'{{ deployment }}'
             }"
     register: ami_register
+    retries: 5
```
adzuci, line 73: Hmm, while thinking about this I realized there is a chance we will hit the max retries and then try to make a new image instead of just waiting.

New Ideas:
1: Based on https://github.com/ansible-collections/amazon.aws/blob/main/plugins/modules/ec2_ami.py#L434 max_attempts is wait_timeout / 15 seconds, but we don't know how often AWS calls are made and the current timeout is 90 minutes which makes it seem odd that the task fails after ~60 minutes.

If that's true we could split this task into two and have the second task just wait for the image to go out of pending, then the retry logic would only be needed on the second task.

2: We could just bump this to 7200 and see if the next max retry error is still at ~60 minutes: https://github.com/edx/configuration/blame/master/playbooks/continuous_delivery/create_ami.yml#L40

adzuci:
Lets just try this and see if we get a new error in the next few weeks

PR #6644, opened 2021-12-22
Revert "Add retry logic to AMI registration"

marekwrobel:
Reverts edx/configuration#6637

It appeared that the retry logic causes the new AMI to be created with the same ID, which fails. For now we will need to manually re-trigger that build stage so that the new AMI is created with a fresh ID.


adzuci (approved):
LGTM, there should be a way to split this task in two and have part one not wait and part two retry while it waits for the instance to be ready.

Alternatively we can bump the wait timeout to 120 minutes.

PR #6649, opened 2022-01-05
fix: Remove an extraneous `-A ecommerce_worker` from the ecomworker startup script

iloveagent57:
This was preventing the celery worker process from starting, because it overrode
the correct `--app ecommmerce_worker.celery_app:app` option.  `-A ecommerce_worker`
tells celery to look for an app named "celery" inside ecommerce_worker, and no
app with that name exists.
ENT-5285

The logged symptom of what's being fixed here:
```
$ head /edx/var/log/supervisor/ecomworker-stderr.log
Error: 
Unable to load celery application.
Module 'ecommerce_worker' has no attribute 'celery'

Error: 
Unable to load celery application.
Module 'ecommerce_worker' has no attribute 'celery'

Error: 
Unable to load celery application.
```

How did this ever work?
This commit https://github.com/openedx/configuration/commit/b8ab26c654a54d754cd01448f72f8b783c95ef6c changed the order of the `-A` and `--app` options, but no deploy for ecomworker went out for that commit until yesterday.  So we were basically just lucky before that the order of the competing options had the correct app name last.

### DIFF
--- a/playbooks/roles/ecomworker/templates/edx/app/ecomworker/ecomworker.sh.j2
+++ b/playbooks/roles/ecomworker/templates/edx/app/ecomworker/ecomworker.sh.j2
@@ -17,4 +17,4 @@ export NEW_RELIC_LICENSE_KEY='{{ NEWRELIC_LICENSE_KEY }}'
 
 source {{ ecommerce_worker_home }}/{{ ecommerce_worker_service_name }}_env
 # We exec so that celery is the child of supervisor and can be managed properly
-exec {{ executable }} --app ecommerce_worker.celery_app:app worker -A ecommerce_worker --concurrency={{ ECOMMERCE_WORKER_CONCURRENCY }} --loglevel=info --hostname=ecomworker.%%h --queue=ecommerce.fulfillment,ecommerce.email_marketing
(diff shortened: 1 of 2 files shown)


playbooks/roles/ecomworker/templates/edx/app/ecomworker/ecomworker.sh.j2
```diff
@@ -17,4 +17,4 @@ export NEW_RELIC_LICENSE_KEY='{{ NEWRELIC_LICENSE_KEY }}'
 
 source {{ ecommerce_worker_home }}/{{ ecommerce_worker_service_name }}_env
 # We exec so that celery is the child of supervisor and can be managed properly
-exec {{ executable }} --app ecommerce_worker.celery_app:app worker -A ecommerce_worker --concurrency={{ ECOMMERCE_WORKER_CONCURRENCY }} --loglevel=info --hostname=ecomworker.%%h --queue=ecommerce.fulfillment,ecommerce.email_marketing
+exec {{ executable }} --app ecommerce_worker.celery_app:app worker --concurrency={{ ECOMMERCE_WORKER_CONCURRENCY }} --loglevel=info --hostname=ecomworker.%%h --queue=ecommerce.fulfillment,ecommerce.email_marketing
```
christopappas: i think this might need to be `-A ecommerce_worker worker --app ecommerce_worker.celery_app:app` (based on the deleted line from https://github.com/openedx/configuration/commit/b8ab26c654a54d754cd01448f72f8b783c95ef6c#diff-8ab0393c8aa3da6ca19d9e3f790279439408de4227828a5a6eb0a68d3664c743)

christopappas (approved):

PR #6650, opened 2022-01-06
fix: remove edx-zoom dependency

schenedx:
Remove the edx-zoom installation from edx-platform because the xblock is being deprecated.
All the courses that uses edx-zoom was edited and the units contain the xblock was removed from course content tree.


oel8man (approved):
reviewed

alangsto (approved):

PR #6654, opened 2022-01-09
chore: install ws-cleanup plugin

jazibhumayun:
Configuration Pull Request
---


jazibhumayun:
Ran the plan plugin upgrade job log seems fine 
[https://build.testeng.edx.org/job/plan-plugin-upgrade-path/18/console](https://build.testeng.edx.org/job/plan-plugin-upgrade-path/18/console)

playbooks/roles/jenkins_data_engineering_new/defaults/main.yml
```diff
@@ -267,6 +267,9 @@ de_jenkins_plugins_list:
   - name: 'workflow-step-api'
     version: '2.21'
     group: 'org.jenkins-ci.plugins.workflow'
+  - name: 'ws-cleanup'
+    version: '0.39'
```
HassanJaveed84, line 271: Can we go with the latest v0.40 ?

jazibhumayun: I think we can't until we upgrade jenkins version
I checked the jenkins version in pom.xml
latest v0.40 requires jenkins v2.235.5 https://github.com/jenkinsci/ws-cleanup-plugin/blob/ws-cleanup-0.40/pom.xml#L47

Our jenkins version is 2.222.3

jazibhumayun: v0.39 requires 2.164.3 or above https://github.com/jenkinsci/ws-cleanup-plugin/blob/ws-cleanup-0.39/pom.xml#L47
so we this should with our current version of jenkins

HassanJaveed84: Thanks for the explanation.

HassanJaveed84 (approved):

PR #6666, opened 2022-01-19
feat: update celery flower (PSRE-1267)

nadeemshahzad:
This PR upgrades our Flower Dockerfile to use focal instead of xenial.  It also upgrades `redis` from 3.2.0 to 4.1.1, `celery` from 4.4.7 to 5.2.3 and `flower` from 0.9.5 to 1.0.0.

### DIFF
--- a/docker/build/flower/Dockerfile
+++ b/docker/build/flower/Dockerfile
@@ -1,10 +1,10 @@
-FROM ubuntu:xenial
+FROM ubuntu:focal
 
 # Update and get pip.
 RUN apt-get update && apt-get install -y python3-pip
 
 # Install the required packages
-RUN pip3 install --no-cache-dir redis==3.2.0 celery==4.4.7 flower==0.9.5
+RUN pip3 install --no-cache-dir celery==5.2.3 https://github.com/mher/flower/zipball/v1.0.0#egg=flower redis==4.1.1
 
 # PYTHONUNBUFFERED: Force stdin, stdout and stderr to be totally unbuffered. (equivalent to `python -u`)
 # PYTHONHASHSEED: Enable hash randomization (equivalent to `python -R`)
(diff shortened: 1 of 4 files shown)


docker/build/flower/Dockerfile
```diff
@@ -22,4 +22,4 @@ USER nobody
 # Mount a config here if you want to enable OAuth etc
 ADD docker/build/flower/flowerconfig.py /flowerconfig.py
 
-ENTRYPOINT [ "flower" ]
+ENTRYPOINT [ "celery" ]
```
adzuci, line 25: Why is the entrypoint now celery?

adzuci:
Could you add a description to your PR mentioning the focal upgrade and that flower v1 is not currently available via pip directly so we are temporarily using an egg?

Could also be worth a comment in the dockerfile saying we should remove the GitHub link in the future if possible.

nadeemshahzad:
thanks, @adzuci, flower version `1.0.0` is available via pip. I've updated the PR to install flower through pip.

docker/build/flower/Dockerfile
nadeemshahzad: In the new releases of flower, we need to invoke `flower` via `celery`. It is mentioned in the doc that this change is effective from version `1.0.1` (  https://flower.readthedocs.io/en/latest/install.html#usage ) but seems like change is also there for `1.0.0` as I am not able to use `flower` command to launch its service.

adzuci: Ah, I see, thanks for the docs link.

adzuci (approved):
:shipit: If there is a good way to test locally, it might be good to add a comment to the PR to explain how to test these changes.

nadeemshahzad:
We've README.txt to run and test this locally https://github.com/openedx/configuration/blob/master/docker/build/flower/README.txt

PR #6667, opened 2022-01-20
chore: add stop instance task to create_ami play

nadeemshahzad:
Creating the AMI while OS is not running will speed up the AMI creation time also provide a consistent snapshot.
I’ve done testing and AMI setup from a running instance (50Gig EBS volume size) took ~30 minutes while AMI setup, while the instance is stopped, took ~17 minutes to complete.

Configuration Pull Request
---

### DIFF
--- a/playbooks/continuous_delivery/create_ami.yml
+++ b/playbooks/continuous_delivery/create_ami.yml
@@ -37,6 +37,7 @@
   vars:
     ec2_region: us-east-1
     ami_wait: yes
+    stop_wait: yes
     ami_creation_timeout: 5400
     no_reboot: no
     artifact_path: /tmp/ansible-runtime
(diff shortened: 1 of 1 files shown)


playbooks/continuous_delivery/create_ami.yml
```diff
@@ -37,6 +37,7 @@
   vars:
     ec2_region: us-east-1
     ami_wait: yes
+    stop_wait: yes
```
adzuci, line 40: Why does this need to be added?

adzuci (approved):
🎉

playbooks/continuous_delivery/create_ami.yml
nadeemshahzad: I'm using var value in `wait: "{{ stop_wait }}"` in task.

PR #6680, opened 2022-02-02
chore: WS-2491 Greenhouse token and drupal cleanup

marekwrobel:
Simple cleanup - removes the greenhouse token and PROSPECTUS_PROXY_PASS (this was used when we were on Drupal).

Configuration Pull Request
---


playbooks/roles/prospectus/defaults/main.yml
```diff
@@ -19,15 +19,13 @@ PROSPECTUS_OAUTH_SECRET: !!null
...
 
-PROSPECTUS_GREENHOUSE_API_TOKEN: 'fake-key'
 PROSPECTUS_CONTENTFUL_SPACE_ID: 'fake-key'
 PROSPECTUS_CONTENTFUL_ACCESS_TOKEN: 'fake-key'
 PROSPECTUS_SEGMENT_WRITE_KEY: 'fake-key'
 PROSPECTUS_ALGOLIA_BROWSE_KEY: 'fake-key'
 
 # nginx vars
 PROSPECTUS_DATA_DIR: '/edx/var/prospectus'
-PROSPECTUS_PROXY_PASS: 'http://localhost:8080'
```
albemarle, line 30: Why is this being taken out?

albemarle (changes requested):
See comments.

playbooks/roles/prospectus/defaults/main.yml
marekwrobel: As per ticket description, PROSPECTUS_PROXY_PASS  was used when we were on Drupal and should also be cleaned up.

jesseday (approved):

playbooks/roles/prospectus/defaults/main.yml
albemarle: 🤦‍♀️  Right.  :-)

albemarle (approved):

PR #6689, opened 2022-02-24
chore: update firefox and chrome versions

dianekaplan:
We have an ecommerce testing failing in selenium because it doesn't recognize the relatively new globalThis.  Based on research in [REV-2624](https://openedx.atlassian.net/browse/REV-2624), we need to update the browser versions used by these tests. 

I'm following and updating the steps [here](https://openedx.atlassian.net/wiki/spaces/TE/pages/814022796/How+to+upgrade+browser+versions), and comparing with previous update [here](https://github.com/openedx/configuration/commit/9882a1d59eba947941534481e3e2ffc6eff31de9)

Configuration Pull Request
---

  -  [ ] Other testing: **we need to identify an appropriate sanity check that new docker containers work**. I'd planned to do the sanity check in step 3 [here](https://openedx.atlassian.net/wiki/spaces/TE/pages/814022796/How+to+upgrade+browser+versions), but those steps (from 2018) don't work on master as a baseline, so I have a ticket here to find out what a suitable baseline check would be: https://2u-internal.atlassian.net/browse/PSRE-1430

  - [n/a] Are you adding any new default values that need to be overridden when this change goes live? 
  - [n/a] If you are making a complicated change, have you performed the proper testing specified on the [Ops Ansible Testing Checklist](https://openedx.atlassian.net/wiki/spaces/SRE/pages/28967861/Ops+Ansible+Testing+Checklist)?  Adding a new variable does not require the full list (although testing on a sandbox is a great idea to ensure it links with your downstream code changes).
  -   [n/a] Think about how this change will affect Open edX operators.  Have you updated the wiki page for the next Open edX release?

### DIFF
--- a/docker/build/chrome/Dockerfile
+++ b/docker/build/chrome/Dockerfile
@@ -1,4 +1,4 @@
-FROM selenium/standalone-chrome-debug:3.14.0-arsenic
+FROM selenium/standalone-chrome-debug:4.1.2-20220217
 LABEL maintainer="edxops"
 
 USER root
--- a/docker/build/firefox/Dockerfile
+++ b/docker/build/firefox/Dockerfile
@@ -1,4 +1,4 @@
-FROM selenium/standalone-firefox-debug:3.14.0-arsenic
+FROM selenium/standalone-firefox-debug:4.1.2-20220217
 LABEL maintainer="edxops"
 
 USER root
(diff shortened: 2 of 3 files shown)


docker/build/chrome/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM selenium/standalone-chrome-debug:3.14.0-arsenic
+FROM selenium/standalone-chrome-debug:4.1.2-20220217
 LABEL maintainer="edxops"
```
dianekaplan, line 2: I wasn't sure if there are any guidelines for which new selenium version to use; I used the latest [here](https://github.com/SeleniumHQ/docker-selenium/releases)

nadeemshahzad (approved):

PR #6692, opened 2022-03-01
chore!: Upgraded to Node 16 and NPM 8

Jawayria:
BREAKING CHANGE: Node and NPM versions upgraded

JIRA: https://openedx.atlassian.net/browse/BOM-3335

Configuration Pull Request
---


playbooks/roles/mfe/tasks/main.yml
```diff
@@ -85,7 +85,7 @@
     - install:app-requirements
 
 - name: install npm overrides
-  shell: "npm install {{ item }} --no-save"
+  shell: "npm install {{ item }} --include=dev --no-save"
```
iamsobanjaved, line 88: As we are setting `NODE_ENV=production` in env vars, so if we install npm packages from `MFE_NPM_OVERRIDES` list without `--include=dev`, it uninstalls dev requirements which is causing build failure in node 16.

- Also changing `--dev` with `--include=dev` as earlier one is deprecated

iamsobanjaved (approved):
Yes, we can merge this PR as this will be only effective for MFEs on the sandbox, and all MFEs available on sandbox are already node 16 compatible. 👍

nadeemshahzad (approved):

PR #6693, opened 2022-03-02
Add playbook to configure a dbt docs server

AbdouSeck:
Configuration Pull Request
---

### DIFF
--- a/playbooks/deploy_nginx_for_dbt_docs.yml
+++ b/playbooks/deploy_nginx_for_dbt_docs.yml
@@ -1,0 +1,6 @@
+- name: Deploy dbt docs server
+  hosts: all
+  become: True
+  gather_facts: True
+  roles:
+   - dbt_docs_nginx
--- a/playbooks/deploy_nginx_for_dbt_docs.yml
+++ b/playbooks/deploy_nginx_for_dbt_docs.yml
@@ -3,4 +3,7 @@
   become: True
   gather_facts: True
   roles:
-   - dbt_docs_nginx
+    - aws
+    - nginx
+    - aws_cloudwatch_agent
+    - dbt_docs_nginx
--- a/playbooks/deploy_nginx_for_dbt_docs.yml
+++ b/playbooks/deploy_nginx_for_dbt_docs.yml
@@ -1,4 +1,4 @@
-- name: Deploy dbt docs server
+- name: Deploy dbt-docs server
   hosts: all
   become: True
   gather_facts: True
--- a/playbooks/roles/aws_cloudwatch_agent/templates/amazon-cloudwatch-agent.json
+++ b/playbooks/roles/aws_cloudwatch_agent/templates/amazon-cloudwatch-agent.json
@@ -4,6 +4,7 @@
 		"run_as_user": "root"
 	},
 	"metrics": {
+		"namespace": "{{ cloudwatch_namespace }}",
 		"append_dimensions": {
 			"InstanceId": "${aws:InstanceId}"
 		},
--- a/playbooks/roles/dbt_docs_nginx/meta/main.yml
+++ b/playbooks/roles/dbt_docs_nginx/meta/main.yml
@@ -1,0 +1,3 @@
+---
+dependencies:
+  - aws_cloudwatch_agent
--- a/playbooks/roles/dbt_docs_nginx/meta/main.yml
+++ b/playbooks/roles/dbt_docs_nginx/meta/main.yml
@@ -1,5 +1,0 @@
----
-dependencies:
-  - aws
-  - nginx
-  - aws_cloudwatch_agent
--- a/playbooks/roles/dbt_docs_nginx/meta/main.yml
+++ b/playbooks/roles/dbt_docs_nginx/meta/main.yml
@@ -1,3 +1,5 @@
 ---
 dependencies:
+  - aws
+  - nginx
   - aws_cloudwatch_agent
--- a/playbooks/roles/dbt_docs_nginx/tasks/main.yml
+++ b/playbooks/roles/dbt_docs_nginx/tasks/main.yml
@@ -1,0 +1,87 @@
+---
+
+- name: Update the cache and update all packages
+  apt:
+    name: "*"
+    state: latest
+    update_cache: yes
+  tags:
+    - install
+    - install:base
+
+- name: Install AWS CLI
+  apt:
+    name: awscli
+    state: present
+  tags:
+    - install
+    - install:base
+
+- name: Install and start nginx
+  apt:
+    name: nginx
+    state: present
+  tags:
+    - install
+    - install:base
+
+- name: Copy dbt docs html files from S3 to the local machine
+  shell: 'aws s3 sync s3://edx-dbt-docs/ /usr/share/nginx/html/ --delete --include "*"'
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Make a Server Configuration Directory
+  file:
+    path: /home/server-config/
+    state: directory
+    mode: 0755
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Copy dbt docs nginx server configurations from S3
+  shell: 'aws s3 cp s3://edx-dbt-docs/server-config/ /home/server-config/ --recursive'
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Copy the nginx configuration file from S3 to the local nginx config directory
+  copy:
+    src: /home/server-config/nginx.conf
+    dest: /etc/nginx/nginx.conf
+    remote_src: yes
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Make the dbt-files-sync.sh script executable
+  file:
+    path: /home/server-config/dbt-files-sync.sh
+    mode: "a+x"
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Set up cron job to update the html files from S3
+  ansible.builtin.cron:
+    name: Run the /home/server-config/dbt-files-sync.sh script
+    minute: 0
+    hour: 23
+    job: /home/server-config/dbt-files-sync.sh
+
+- name: Restart nginx and enable it on reboot
+  ansible.builtin.service:
+    name: nginx
+    state: restarted
+    enabled: yes
--- a/playbooks/roles/dbt_docs_nginx/tasks/main.yml
+++ b/playbooks/roles/dbt_docs_nginx/tasks/main.yml
@@ -9,18 +9,10 @@
     - install
     - install:base
 
-- name: Install AWS CLI
+- name: Uninstall AWS CLI from the package manager
   apt:
     name: awscli
-    state: present
-  tags:
-    - install
-    - install:base
-
-- name: Install and start nginx
-  apt:
-    name: nginx
-    state: present
+    state: absent
   tags:
     - install
     - install:base
--- a/playbooks/roles/dbt_docs_nginx/tasks/main.yml
+++ b/playbooks/roles/dbt_docs_nginx/tasks/main.yml
@@ -15,7 +15,7 @@
     state: absent
   tags:
     - install
-    - install:base
+    - install:aws
 
 - name: Copy dbt-docs html files from S3 to the local machine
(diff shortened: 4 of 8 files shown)


playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+  tags:
+    - install
+    - install:base
+
+- name: Set up cron job to update the html files from S3
+  ansible.builtin.cron:
+    name: Run the /home/server-config/dbt-files-sync.sh script
+    minute: 0
+    hour: 23
+    job: /home/server-config/dbt-files-sync.sh
```
jazibhumayun: should we add tags on last two tasks as well ? As you have used tags on all the above tasks

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,87 @@
...
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Set up cron job to update the html files from S3
+  ansible.builtin.cron:
+    name: Run the /home/server-config/dbt-files-sync.sh script
+    minute: 0
+    hour: 23
```
jazibhumayun: This job  http://jenkins-new.analytics.edx.org/view/dbt/job/dbt-docs/ runs every 6 hours. I would recommend to update the cron to run every 6 hours. We can use `30 */6 * * *` so it will run “At 30 minute past every 6th hour.”

playbooks/roles/aws_cloudwatch_agent/templates/amazon-cloudwatch-agent.json
```diff
@@ -4,6 +4,7 @@
 		"run_as_user": "root"
 	},
 	"metrics": {
+		"namespace": "{{ cloudwatch_namespace }}",
```
jazibhumayun, line 7: great! thanks for making it dynamic namespace

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: The update has been made.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: The update has been made

playbooks/roles/aws_cloudwatch_agent/templates/amazon-cloudwatch-agent.json
jazibhumayun: This should be via parameters file. Can you share pull request for https://github.com/edx-ops/analytics-secure ?

jazibhumayun (approved):

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: Install AWS CLI
+  apt:
+    name: awscli
+    state: present
+  tags:
+    - install
+    - install:base
+
+- name: Install and start nginx
```
estute: We already have a role for nginx, which seems to be more complete. It would probably be a good idea to use that instead to get the benefit of the SRE team's experience? See: https://github.com/openedx/configuration/tree/master/playbooks/roles/nginx

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+
+- name: Copy dbt docs nginx server configurations from S3
+  shell: 'aws s3 cp s3://edx-dbt-docs/server-config/ /home/server-config/ --recursive'
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Copy the nginx configuration file from S3 to the local nginx config directory
```
estute: What is the benefit of keeping these configs in s3 and separate from this role? Often times we include a template config file

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+    path: /home/server-config/
+    state: directory
+    mode: 0755
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Copy dbt docs nginx server configurations from S3
```
estute: Can you explain the difference between this config file and the one in the next step?

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,93 @@
...
+    remote_src: yes
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:base
+
+- name: Make the dbt-files-sync.sh script executable
+  file:
+    path: /home/server-config/dbt-files-sync.sh
```
estute: I see. This comes from the s3 bucket. We should either include this script in this role or put it into version control, depending on the nature of it.

playbooks/deploy_nginx_for_dbt_docs.yml
```diff
@@ -0,0 +1,6 @@
+- name: Deploy dbt docs server
```
estute: nit: I feel like the name of this file should be `dbt-docs` or something like that, as it represents the whole setup of the server, which can include more than nginx.

playbooks/deploy_nginx_for_dbt_docs.yml
```diff
@@ -0,0 +1,6 @@
...
+  hosts: all
+  become: True
+  gather_facts: True
+  roles:
```
estute, line 5: Do we want to include the `aws` role as well?

estute (changes requested):
See inline comments

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: That role is doing a bunch more than was needed. I would consider using it once I have an idea how to provide some of the extra args it's templated.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: I do not know the answer to that. I was working with a shell script that @jazibhumayun had in terraform as `user_data`. In that script, html and config files were copied from S3. I would be open to having them here.

AbdouSeck: @estute  Is that a path we want to take now?

playbooks/deploy_nginx_for_dbt_docs.yml
AbdouSeck: This is also related to what I mentioned about the `nginx` role. There is quite a bit going on there that we may not need here. Happy to look into incorporating tho.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
estute: In general, I think it is a good idea to place all of the configs in one place. We don't really have the ability to review and audit the history of an s3 bucket. However, if the configs in question are sensitive, then we store them separately

playbooks/deploy_nginx_for_dbt_docs.yml
estute: I think it would be a good idea to look into it. Part of the benefit of using the `aws` and `nginx` roles is that we can benefit from other teams working on updates, patches and fixes. @jazibhumayun what do you think?

AbdouSeck: @estute the `aws` and `nginx` are being used now. One thing to note is the `aws` role installs an older version of the CLI tool. It works just fine, though.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: @estute the latest changes should take care of this.

playbooks/roles/dbt_docs_nginx/meta/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+dependencies:
```
estute: nit: In general, we treat the playbook as the top level representation of the system we want to build, and roles represent individual components. In this case, I think `aws`  and `aws_cloudwatch_agent` aren't dependencies of the dbt service we are hosting, so they should probably be moved to the playbook. See: https://github.com/openedx/configuration/blob/master/playbooks/ecomworker.yml

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+- name: Update the cache and update all packages
+  apt:
+    name: "*"
+    state: latest
+    update_cache: yes
+  tags:
+    - install
+    - install:base
+
+- name: Uninstall AWS CLI from the package manager
```
estute: Why are we doing this? I'm guessing we install it in another way and don't want there to be conflicts. If so, can you call it out explicitly in a comment?

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+  template:
+    src: "nginx/nginx.conf.j2"
+    dest: /etc/nginx/nginx.conf
+  become: yes
+  become_method: sudo
+  tags:
+    - install
+    - install:nginx
+
+- name: Make the dbt-files-sync.sh script executable
```
estute: I'm not sure this file is present on the system. If it doesn't contain sensitive values, can we include it in this PR as a template or a file (see: https://github.com/openedx/configuration/tree/master/playbooks/roles/jenkins_common/files)

playbooks/roles/dbt_docs_nginx/tasks/main.yml
```diff
@@ -0,0 +1,76 @@
...
+
+- name: Uninstall AWS CLI from the package manager
+  apt:
+    name: awscli
+    state: absent
+  tags:
+    - install
+    - install:aws
+
+- name: Copy dbt-docs html files from S3 to the local machine
```
estute, line 20: How is this different than the script that gets run on the cron job?

estute (changes requested):
See inline questions

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: This copies the files from S3 once. The cron piece is needed to make sure that the files get copied periodically. The cron script uses the same command.

playbooks/roles/dbt_docs_nginx/templates/nginx/nginx.conf.j2
```diff
@@ -0,0 +1,65 @@
+user www-data;
```
estute, line 1: It might make sense to use https://github.com/openedx/configuration/blob/master/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2. You could take the values we use in this and pass it to that template. That way, we get the benefit of a config file that many engineers have had eyes on, but it might not suit this purpose. Just an idea.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: @estute this has been updated with `template`.

playbooks/roles/dbt_docs_nginx/meta/main.yml
AbdouSeck: Those roles are no longer listed as dependencies.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
AbdouSeck: The name has been updated to reflect the intent for this task.

playbooks/roles/dbt_docs_nginx/templates/nginx/nginx.conf.j2
AbdouSeck: Our config file is a lot simpler and has a `server_name` attribute that the `nginx` role's config file does not have. I doubt that I should be updating that template file with that attribute.

playbooks/roles/dbt_docs_nginx/tasks/main.yml
estute: Great, thanks

estute (approved):

PR #6696, opened 2022-03-04
feat: jenkins update Automation

oel8man:
De[script](https://github.com/edx/edx-internal/pull/6222)ion
As an owner of Jenkins I know that we are able to use an automated script to update it so that we can ensure it is on a recent version.

### DIFF
--- a/playbooks/roles/aws/tasks/main.yml
+++ b/playbooks/roles/aws/tasks/main.yml
@@ -53,7 +53,8 @@
     state: present
     update_cache: yes
   register: install_pkgs
-  until: install_pkgs|success
+  until: install_pkgs is succeeded
+    ###until: install_pkgs|success
   retries: 10
   delay: 5
   when: ansible_distribution in common_debian_variants
--- a/playbooks/roles/neo4j/meta/main.yml
+++ b/playbooks/roles/neo4j/meta/main.yml
@@ -2,7 +2,7 @@
 dependencies:
   - common
   - role: oraclejdk
-    ORACLEJDK_VERSION: "8u131"
-    oraclejdk_base: "jdk1.8.0_131"
-    oraclejdk_build: "b11"
-    oraclejdk_link: "/usr/lib/jvm/java-8-oracle"
+    ORACLEJDK_VERSION: "11.0.13"
+    oraclejdk_base: "jdk-11.0.13"
+    oraclejdk_build: "13"
+    oraclejdk_link: "/usr/lib/jvm/jdk-11.0.13"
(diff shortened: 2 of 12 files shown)


adzuci (changes requested):
Hey Abdoul, I think you copied the edX Jira ticket to the description, but I don't think that's appropriate for a public repo like this.  Could you look at some other PRs for examples and add a concise blurb about what this PR is changing + a link to the docs that explain the is successful syntax?

Also, I think we use the |success syntax a bunch elsewhere, is that something that doesn't work for Ansible 2.8?  Or do you know if we'll need to update other roles?

playbooks/roles/aws/tasks/main.yml
```diff
@@ -53,7 +53,8 @@
     state: present
     update_cache: yes
   register: install_pkgs
-  until: install_pkgs|success
+  until: install_pkgs is succeeded
```
ohnickmoy, line 56: did flags like these change syntax because of upgrade?

playbooks/roles/python/tasks/main.yml
```diff
@@ -13,8 +13,8 @@
   retries: 10
   delay: 10
 
-- name: Install packages
-  raw: "apt-get install -qq {{ item }}"
-  with_items: "{{ python_packages }}"
-  retries: 10
-  delay: 10
+#- name: Install packages
```
ohnickmoy: what was the purpose of commenting this out?

ohnickmoy (approved):

playbooks/roles/aws/tasks/main.yml
oel8man: yes

playbooks/roles/python/tasks/main.yml
oel8man: It was failing the python minimal install. I reverted back

playbooks/roles/aws/tasks/main.yml
adzuci: What got upgraded from what to what to cause this?

playbooks/roles/neo4j/meta/main.yml
```diff
@@ -2,7 +2,7 @@
 dependencies:
   - common
   - role: oraclejdk
-    ORACLEJDK_VERSION: "8u131"
-    oraclejdk_base: "jdk1.8.0_131"
-    oraclejdk_build: "b11"
-    oraclejdk_link: "/usr/lib/jvm/java-8-oracle"
+    ORACLEJDK_VERSION: "11.0.13"
```
adzuci, line 5: @oel8man Does Jenkins install this role?  I think this is used in coursegraph, so we shouldn't changes it without asking TNL for a review, cc @kenclary

adzuci (changes requested):
@oel8man Could you add a bulleted list of the changes that this PR is making and what roles you touched in your description rather "As an owner of Jenkins I know that we are able to use an automated script to update it so that we can ensure it is on a recent version."

PR #6699, opened 2022-03-08
feat: update flowerconfig to add basic_auth support (PSRE-1441)

nadeemshahzad:
The purpose of this PR is to add support in flower config to use basic access authentication to restrict flower access via usernames and passwords. In sandbox builds we have different URLs and we can not use oAuth to limit sandboxes so alternatively, we can use basic auth here.


playbooks/roles/flower/defaults/main.yml
```diff
@@ -13,6 +13,7 @@ FLOWER_OAUTH2_KEY: "A Client ID from Google's OAUTH2 provider"
 FLOWER_OAUTH2_SECRET: "A Client Secret from Google's OAUTH2 provider"
 FLOWER_OAUTH2_REDIRECT: "A URL registered with Google's OAUTH2 provider"
 FLOWER_AUTH_REGEX: ".*@example.com" # Can be blank to disable auth
+FLOWER_BASIC_AUTH: []
```
adzuci, line 18: Could you add a comment here to differentiate this variable from the one above it and make it clearer how to enable or disable basic auth?

adzuci (approved):
Looks good to me!  Thanks for the quick fix Nadeem!

It may be helpful to add a PR description as well.

PR #6700, opened 2022-03-08
feat: npm installation doesn't perform known host checking

matthugs:
This PR was motivated by timeout errors in sandbox builds, where the last step ansible had attempted to execute was `[edxapp: install node dependencies]`. On investigation it looks like this was the result of new JS code we were trying to build against needing to communicate with github over ssh in a way we weren't previously dependent on, likely the result of newer/more up-to-date JS dependencies. This PR removes this blocker, and changes our edxapp deployments to use `npm ci` rather than `npm install` when pulling in JS/frontend dependencies. This last should be a build reliability improvement.

Successful sandbox build using this config branch: https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/45283/console

Configuration Pull Request


adzuci (approved):
:shipit: Thank you!

PR #6709, opened 2022-03-30
fix: CORS issue for CSRF token API

waheedahmed:
If basic auth is enabled on the sandbox, CSRF token endpoint is resolving to location / and setting auth_basic to `Restricted` for this endpoint. Fixed by defining its own rule.

**Reproduction steps:** 
1. Open network tab in browser inspector tool and visit https://authn-wahmed.sandbox.edx.org/login
2. Note that CSRF token endpoint failed with CORS error:
Screenshot 2022-03-30 at 12 52 07 PM

**Sandbox with fix:** https://authn-csrf-api.sandbox.edx.org/login
1. Follow the above steps for this sandbox and notice that there is no error on CSRF token endpoint
Screenshot 2022-03-30 at 1 00 25 PM

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2
@@ -199,6 +199,11 @@ error_page {{ k }} {{ v }};
     try_files $uri @proxy_to_lms_app_api;
   }
 
+  # CSRF token API
+  location /csrf/api {
+    try_files $uri @proxy_to_lms_app_api;
+  }
+
   #enterprise API
   location /enterprise/api {
     try_files $uri @proxy_to_lms_app_api;


nadeemshahzad (approved):

timmc-edx:
Does this change remove the HTTP Auth protection for the CSRF endpoint in sandboxes? I don't think that's a problem, but I'm not sure why it's needed as opposed to making sure that the existing HTTP Auth is applied. Can you expand on this?

waheedahmed:
@timmc-edx yes this does remove the HTTP basic auth on this endpoint but I don't think this is an issue since we are already doing this for the other LMS API endpoints e.g. https://github.com/openedx/configuration/blob/2667d8984404dc8e1b1279a8e3b7c6c344acc9dd/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/lms.j2#L207-L225

waheedahmed:
Also, since these endpoints are called by the MFEs I don't think we can keep them behind the basic auth rule, if so how we would handle the basic auth from the frontends?

timmc-edx:
Yeah, I guess there's an issue with CORS here -- we're probably not setting the right CORS headers to allow the browser to send basic auth headers to these endpoints.

PR #6713, opened 2022-04-07
feat: add conductor support for newer versions of gatsby

matthugs:
Newer versions of gatsby appear to rely on files it generates in the
page-data directory, which didn't match our previous regular
expression.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/conductor.j2
+++ b/playbooks/roles/nginx/templates/edx/app/nginx/sites-available/conductor.j2
@@ -55,6 +55,11 @@ server {
         proxy_pass {{ static_site.proxied_path }}/$1$2;
     }
 
+    # Matches: <router>/<organization>/page-data/[.../]<file.ext>
+    location ~ ^/{{ static_site.router_path }}/page-data/((?:\w+\/+)*)([\w\-\.]+\.[\w\-\.]+) {
+        proxy_pass {{ static_site.proxied_path }}/page-data/$1$2;
+    }
+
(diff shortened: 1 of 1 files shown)


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/conductor.j2
```diff
@@ -55,6 +55,11 @@ server {
         proxy_pass {{ static_site.proxied_path }}/$1$2;
     }
 
+    # Matches: <router>/<organization>/page-data/[.../]<file.ext>
+    location ~ ^/{{ static_site.router_path }}/page-data/((?:\w+\/+)*)([\w\-\.]+\.[\w\-\.]+) {
```
schenedx, line 59: Do we need some wild card before `page-data` path?

schenedx (approved):
Try this?

playbooks/roles/nginx/templates/edx/app/nginx/sites-available/conductor.j2
matthugs: I don't think so; it looks like gatsby outputs this file folder in the root of its build output every time.
Screen Shot 2022-04-07 at 3 30 53 PM

Screen Shot 2022-04-07 at 3 31 00 PM

matthugs: So you must be thinking ahead towards making it less brittle in the future, right? If we're doing that I think it'll look a lot simpler if we just update the folder regex to accept hyphens as readily as it accepts letters, an idea we rejected on the call.

PR #6732, opened 2022-04-19
fix: Mark all directories as safe for git on devstack

timmc-edx:
When a user runs `make requirements` inside devstack's `lms-shell` (or tries to provision), pip pip pulls some from dependencies from git URLs and then tries to use git commands against those directories in the virtualenv. However, in devstack, the owner of those repo directories doesn't match the current user (root), and git versions 2.35.2 will refuse to read repositories from a different user (since this could allow code execution).

This change tells git to consider all directories safe, at least on devstack. (Listing out specific directories to consider safe isn't really feasible, since the set of git-based dependencies changes on a regular basis.)

ref: ARCHBOM-2096

----

### DIFF
--- a/playbooks/roles/git_clone/tasks/main.yml
+++ b/playbooks/roles/git_clone/tasks/main.yml
@@ -28,6 +28,7 @@
 - name: Mark all directories as safe for git on devstack
   shell: "git config --global --add safe.directory '*'"
   become: true
+  when: "({{ devstack | default(False) }} or {{ edx_django_service_is_devstack | default(False) }})"
   tags:
     - devstack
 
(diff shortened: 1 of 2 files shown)


playbooks/roles/git_clone/tasks/main.yml
```diff
@@ -121,6 +121,17 @@
...
+# On devstack, tell Git that repos owner by other users are safe.
+# This is necessary in git 2.35.2 and higher. Devstack uses a mix of
+# root and edxapp and git+https pip dependencies end up cloning repos
+# into an open-ended set of directories, so our best bet is to just
+# say every dir on devstack is safe.
+- name: Mark all directories as safe for git on devstack
+  shell: "git config --global --add safe.directory '*'"
+  become: true
+  tags:
+    - devstack
```
timmc-edx: Author note: I'm not sure if this is the correct way to scope this to just devstack,

timmc-edx:
In chat, @nadeemshahzad pointed out that edxapp doesn't actually bring in the `git_clone` role! May have to duplicate this task.

thomty (approved):
Seems like a straightforward way to run this for now.

PR #6745, opened 2022-04-26
fix: Mark all directories as safe for git on devstack (#6732)

timmc-edx:
When a user runs make requirements inside devstack's lms-shell (or
tries to provision), pip pulls some from dependencies from git URLs
and then tries to use git commands against those directories in the
virtualenv. However, in devstack, the owner of those repo directories
doesn't match the current user (root), and git versions 2.35.2 will
refuse to read repositories from a different user (since this could
allow code execution).

This change tells git to consider all directories safe, at least on
devstack. (Listing out specific directories to consider safe isn't
really feasible, since the set of git-based dependencies changes on a
regular basis.)

Notes:

- edxapp doesn't use git_clone, but it looks like everything else
  does, so I had to duplicate the task
- There's no one, convenient way of marking something to run only on
  devstack; apparently the devstack tag isn't used universally when
  building devstack, but it turns out that either `devstack` or
  `edx_django_service_is_devstack` is set (and true) in the
  `docker/build/*/ansible_overrides.yml` files, so we can check if
  *either* of those is true. This approach is used in various other
  playbooks already. I also threw on the `devstack` tag for good
  measure, and again because some other tasks use that approach in
  combination.

ref: ARCHBOM-2096

(cherry picked from commit e9a270bd/PR #6732)

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -27,6 +27,18 @@
     - install
     - install:base
 
+# On devstack, tell Git that repos owner by other users are safe.
+# This is necessary in git 2.35.2 and higher. Devstack uses a mix of
+# root and edxapp and git+https pip dependencies end up cloning repos
+# into an open-ended set of directories, so our best bet is to just
+# say every dir on devstack is safe.
+- name: Mark all directories as safe for git on devstack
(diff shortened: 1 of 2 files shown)


robrap (approved):
I’m assuming this is straightforward application of the original reviews solution.

timmc-edx:
Yep! Applied cleanly.

PR #6759, opened 2022-05-19
feat: enable caching of prospectus pages

mulby:


playbooks/roles/nginx/templates/edx/app/nginx/sites-available/prospectus.j2
```diff
@@ -46,9 +46,7 @@ server {
     add_header Cache-Control "no-store, max-age=0" always;
   }
 
-  location /event/ {
-    # doesn't have any content, can be aggressively cached
-    add_header 'Cache-Control' 'public, max-age=86400';
+  location /event {
```
mulby, line 49: Note that this was only working when the trailing slash was specified, we need it to work without as well.

marekwrobel (approved):

PR #6763, opened 2022-05-26
build: remove ppa for watchman as it is available by default in ubuntu 20.04

iamsobanjaved:
As we have switched to Ubuntu 20.04, `watchman` is available out of the box so we don't need to include this PPA

**Ref:** https://packages.ubuntu.com/focal/utils/watchman

**JIRA:** [BOM-2097](https://2u-internal.atlassian.net/browse/BOM-2097)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 All notable changes to this project will be documented in this file.
 Add any new changes to the top (right below this line).
 
+ - 2022-06-06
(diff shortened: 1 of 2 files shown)


aht007 (approved):

adzuci (approved):
This looks good to me, though I am not sure if anyone still runs this playbook on older instances, if we think they may we could at least add something in the change log to note this change.

PR #6766, opened 2022-05-30
chore: Removing unused path from config.

awais786:
Removing unused path from config.

Removed /common/lib/sandbox-packages folder to new library https://pypi.org/project/codejail-includes/
So this folder does't not exists any more. `env_dir` already exists there which seems enough.

### DIFF
--- a/playbooks/roles/edxapp/templates/code.sandbox.j2
+++ b/playbooks/roles/edxapp/templates/code.sandbox.j2
@@ -4,7 +4,6 @@
     #include <abstractions/base>
 
     {{ edxapp_sandbox_venv_dir }}/** mr,
-    {{ edxapp_code_dir }}/common/lib/sandbox-packages/** r,
     /tmp/codejail-*/ rix,
     /tmp/codejail-*/** wrix,
 


playbooks/roles/edxapp/templates/code.sandbox.j2
```diff
@@ -4,7 +4,6 @@
     #include <abstractions/base>
 
     {{ edxapp_sandbox_venv_dir }}/** mr,
```
awais786, line 6: This line will serve the purpose.

kdmccormick (approved):
Looks good. ~I imagine that this will need to be deployed at the same time as https://github.com/openedx/edx-platform/pull/30402, otherwise codejail would be partially broken in between.~ (as @feanil noted below, this is probably safe to deploy any time after the edx-platform change)

feanil (approved):

feanil:
Nice, this should be safe to deploy anytime after the edx-platform change has been deployed but not before.

awais786:
Platform PR merged and looking fine. I will merge this PR tomorrow.

PR #6795, opened 2022-08-25
fix: Quote pip package specs on command line

timmc-edx:
This is required to avoid interpreting `<`, `#`, and other characters in the shell.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -164,13 +164,13 @@
     - install
     - install:app-requirements
 
-# Private requriements require a ssh key to install, use the same key as the private key for edx-platform
+# Private requirements require a ssh key to install, use the same key as the private key for edx-platform
 # If EDXAPP_INSTALL_PRIVATE_REQUIREMENTS is set to true EDXAPP_USE_GIT_IDENTITY must also be true
 - name: install python private requirements
   # Need to use shell rather than pip so that we can maintain the context of our current working directory; some
   # requirements are pathed relative to the edx-platform repo. Using the pip from inside the virtual environment implicitly
   # installs everything into that virtual environment.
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -164,13 +164,13 @@
...
 
-# Private requriements require a ssh key to install, use the same key as the private key for edx-platform
+# Private requirements require a ssh key to install, use the same key as the private key for edx-platform
 # If EDXAPP_INSTALL_PRIVATE_REQUIREMENTS is set to true EDXAPP_USE_GIT_IDENTITY must also be true
 - name: install python private requirements
   # Need to use shell rather than pip so that we can maintain the context of our current working directory; some
   # requirements are pathed relative to the edx-platform repo. Using the pip from inside the virtual environment implicitly
   # installs everything into that virtual environment.
-  shell: "{{ edxapp_venv_dir }}/bin/pip install {{ COMMON_PIP_VERBOSITY }} -i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w {{ item.extra_args|default('') }} {{ item.name }}"
+  shell: "{{ edxapp_venv_dir }}/bin/pip install {{ COMMON_PIP_VERBOSITY }} -i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w {{ item.extra_args|default('') }} {{ item.name|quote }}"
```
matthugs, line 173: would this also work?

```suggestion
  shell: "{{ edxapp_venv_dir }}/bin/pip install {{ COMMON_PIP_VERBOSITY }} -i {{ COMMON_PYPI_MIRROR_URL }} --exists-action w {{ item.extra_args|default('') }} \"{{ item.name }}\""
```

matthugs (approved):
/shrug

playbooks/roles/edxapp/tasks/deploy.yml
timmc-edx: It would work in this case, but not work in the general case, e.g. if malicious or unexpected inputs might contain the `"` character or a backslash. Best to just use `quote`, which is specific to shell-escaping: https://docs.ansible.com/ansible/latest/collections/ansible/builtin/shell_module.html#notes

matthugs: cool sounds good

PR #6796, opened 2022-08-26
Add a Groovy script in ${JENKINS_HOME}/init.groovy.d to start Jenkins in maintenance mode

AbdouSeck:
This PR adds a task that installs an `init.groovy` under `${JENKINS_HOME}`. This script contains some simple Groovy code to force Jenkins to always start in quiet/maintenance mode. The task is run or skipped based on the value of the boolean `start_jenkins_in_quiet_mode`. If the latter is true, then the task is run. Otherwise, it is skipped.

**UPDATE**:
The updates are now the following:
- Add a `0StartInQuietMode.groovy` Groovy script to the jenkins-configuration repo: [Related jenkins-configuration PR](https://github.com/edx/jenkins-configuration/pull/206)
- Update the `jenkins_data_engineering` and `jenkins_data_engineering_new` playbooks to make sure that when `start_jenkins_in_quiet_mode` is set to `true`, only files not matching `0StartInQuietMode.groovy` are deleted from the `init.groovy.d` directory.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/jenkins_common/defaults/main.yml
+++ b/playbooks/roles/jenkins_common/defaults/main.yml
@@ -11,6 +11,8 @@ jenkins_common_protocol_https: true
 # When checking if Jenkins is finished initializing, expect a 200 as it should
 # be publicly available
 jenkins_common_ready_status_code: 200
+# Always start Jenkins in Quiet/Maintenance mode
+start_jenkins_in_quiet_mode: true
 
 JENKINS_SERVER_NAME: jenkins.example.org
 jenkins_node_usage: EXCLUSIVE
--- a/playbooks/roles/jenkins_data_engineering/defaults/main.yml
+++ b/playbooks/roles/jenkins_data_engineering/defaults/main.yml
@@ -10,6 +10,8 @@ jenkins_user_home: '/home/{{ jenkins_user }}'
 jenkins_port: 8080
 jenkins_nginx_port: 80
 jenkins_protocol_https: true
+# Always start Jenkins in Quiet/Maintenance mode
+start_jenkins_in_quiet_mode: true
 AUTOMATION_USER: 'edx-analytics-automation'
 jenkins_host_name: "{{ JENKINS_SERVER_NAME | default('jenkins') }}"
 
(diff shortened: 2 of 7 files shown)


pwnage101 (commented):
Have you instead considered instead adding a script to init.groovy.d subdirectory?  Perhaps `init.groovy.d/0StartInQuietMode.groovy` to make sure that it runs first.  Keep in mind we do already add 10+ init scripts under init.groovy.d, so this script adding init.groovy might possibly clobber the logic to inherit init.groovy.d, or at the very least is "unfair" to any other package that is also contending for the init.groovy file.  See, for example: https://github.com/edx/jenkins-configuration/search?q=init.groovy.d and https://github.com/edx/jenkins-configuration/tree/master/src/main/groovy

AbdouSeck:
@pwnage101 I did consider the `init.groovy.d` directory, but if you look at the `jenkins_common` role, you'll notice that that directory gets deleted by the role once all those scripts are done running. But we DO NOT want this current init.groovy script to be deleted. That is the reasoning behind making this init script separate from those other ones.

pwnage101:
I see, then I have a new proposal: change the last task to delete only the files specified by the `jenkins_common_configuration_scripts` array.  That way, the only file remaining after running the playbook is the `init.groovy.d/0StartInQuietMode.groovy` file.

Also, FYI, I just logged into the instance and found that the files were not deleted:
```
root@jenkins-new:~# ls /var/lib/jenkins/init.groovy.d/
1addJarsToClasspath.groovy     3installPython.groovy        4configureGHOAuth.groovy  4configureJobConfigHistory.groovy  4configureSecurity.groovy  5configureEmailExtension.groovy
2checkInstalledPlugins.groovy  3mainConfiguration.groovy    4configureGHPRB.groovy    4configureMailerPlugin.groovy      4configureSlack.groovy     5createLoggers.groovy
3importCredentials.groovy      3setGlobalProperties.groovy  4configureGithub.groovy   4configureMaskPasswords.groovy     5addSeedJob.groovy
```

AbdouSeck:
@pwnage101 I think that `/var/lib/jenkins` folder on the old jenkins-new instance. The new instance, after ansible provisioning, does not have that `init.groovy.d` folder. In fact, the `init.groovy.d` folder on the old instance was created on January 13th 2022. But that's definitely not when that instance was provisioned. Not sure how that folder ended up there with that creation/modified date.
I am going to work on your proposal to delete only stuff in `jenkins_common_configuration_scripts`.

pwnage101 (commented):
LGTM, just note that I think the prefix should be 9 instead of 0

pwnage101 (approved):

PR #6798, opened 2022-08-30
Test Docker Compose for use in in Sandboxes

adzuci:
The goal of this PR is to enable our sandbox users to use images like https://hub.docker.com/r/openedx/edx_exams rather than Ansible roles to deploy services.

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -617,6 +617,10 @@ if [[ $reconfigure != "true" && $server_type == "full_edx_installation" ]]; then
     done
 fi
 
+# TODO: Add if block
+pip install docker-compose
+docker-compose up ../util/jenkins/docker-compose.yml
+
 # deploy the edx_ansible play
 run_ansible edx_ansible.yml -i "${deploy_host}," $extra_var_arg --user ubuntu
 cat $sandbox_secure_vars_file $sandbox_internal_vars_file $extra_vars_file | grep -v -E "_version|migrate_db" > ${extra_vars_file}_clean
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -617,9 +617,10 @@ if [[ $reconfigure != "true" && $server_type == "full_edx_installation" ]]; then
     done
 fi
 
-# TODO: Add if block
-pip install docker-compose
-docker-compose up ../util/jenkins/docker-compose.yml
+if [[ -z $edx_exams ]]; then
+  pip install docker-compose
+    docker-compose up ../util/jenkins/docker_compose/edx-exams-docker-compose.yml
+fi
 
 # deploy the edx_ansible play
 run_ansible edx_ansible.yml -i "${deploy_host}," $extra_var_arg --user ubuntu
--- a/util/jenkins/docker_compose/edx-exams-docker-compose.yml
+++ b/util/jenkins/docker_compose/edx-exams-docker-compose.yml
@@ -18,14 +18,11 @@ services:
       context: .
       dockerfile: Dockerfile
     container_name: exams.app
-    # Use the Django devserver, so that we can hot-reload code changes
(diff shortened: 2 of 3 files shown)


util/jenkins/docker_compose/edx-exams-docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+      context: .
+      dockerfile: Dockerfile
+    container_name: exams.app
+    command: bash -c 'while true; do python /edx/app/exams/manage.py runserver 0.0.0.0:18740; sleep 2; done'
+    ports:
+      - "18740:18740"
+    depends_on:
+      - mysql
+    networks:
+    # TODO: Figure out if the host network is viable to use https://stackoverflow.com/a/44544841
```
alangsto, line 27: I think we should be able to test this on the sandbox

schenedx (approved):
👍 Looks straightforward. I guess most of the changes needed would be in the docker compose file. We can try this.

util/jenkins/docker_compose/edx-exams-docker-compose.yml
```diff
@@ -0,0 +1,38 @@
+version: "2.1"
+services:
+  mysql:
+    image: mysql:8.0.28-oracle
```
adzuci, line 4: Will this bind to the same port as the host MySQL instance?

adzuci (commented):
I'm not sure if this will work or if more bits are needed, but it seems like it has the key pieces.  Could we test spinning up a sandbox off of this branch to see if there are port conflicts or other errors? 

cc @nadeemshahzad & @NIXKnight

util/jenkins/docker_compose/edx-exams-docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+    depends_on:
+      - mysql
+    networks:
+    # TODO: Figure out if the host network is viable to use https://stackoverflow.com/a/44544841
+      - network_mode: "host"
+    # Allows attachment to this container using 'docker attach <containerID>'.
+    stdin_open: true
+    tty: true
+    environment:
+      DJANGO_SETTINGS_MODULE: exams.settings.production
```
adzuci, line 33: Should this be prod or something else?

alangsto: It was set to devstack, not sure which one is more correct

adzuci: @NIXKnight says that other sandbox legacy services also use this line: https://github.com/edx/edx-exams/blob/main/edx_exams/settings/production.py#L25

To pull from a file that is generated from this file: https://github.com/edx/sandbox-internal/blob/01c8ebe7a32a75ef31a787e10e6f1db526aa8168/sandbox-remote-config/sandbox/registrar.yml#L76 in Ansible.

util/jenkins/ansible-provision.sh
```diff
@@ -617,6 +617,11 @@ if [[ $reconfigure != "true" && $server_type == "full_edx_installation" ]]; then
     done
 fi
 
+if [[ -z $edx_exams ]]; then
```
alangsto, line 620: I think this might need to be `[[ -z $edx_exams == "true" ]]`, but not sure. Maybe that's the root cause of exams not actually running on the sandbox

util/jenkins/ansible-provision.sh
```diff
@@ -617,6 +617,11 @@ if [[ $reconfigure != "true" && $server_type == "full_edx_installation" ]]; then
...
+  pip install docker-compose
+    docker-compose up ../util/jenkins/docker_compose/edx-exams-docker-compose.yml
```
adzuci, line 622: As mentioned in https://2u-internal.atlassian.net/browse/PSRE-1925?focusedCommentId=633526, it looks like this script is run on Jenkins and not on the sandbox, so we will likely need to update it to run on the sandbox.

PR #6800, opened 2022-09-01
chore: set up role for new relic in coursegraph/neo4j. TNL-10006

kenclary:
The goal here is to get some basic information from coursegraph into new relic. This will just cover nginx (server) logs, which is a start.

### DIFF
--- a/playbooks/neo4j.yml
+++ b/playbooks/neo4j.yml
@@ -15,3 +15,5 @@
     - neo4j
     - role: splunkforwarder
       when: COMMON_ENABLE_SPLUNKFORWARDER
+    - role: newrelic_infrastructure
+      when: COMMON_ENABLE_NEWRELIC_INFRASTRUCTURE


matthugs (approved):
from my peeking at this a bit it looks like this will only give you at most `nginx` logs in NR. If you'd like more you may need to override `NEWRELIC_LOGS`

adzuci (approved):
LGTM once you add a PR description and decide if you want to update the variable Matt mentioned.

I also did some digging and found https://docs.newrelic.com/docs/infrastructure/host-integrations/host-integrations-list/nginx/nginx-integration/ + https://discuss.newrelic.com/t/how-to-setup-neo4j-monitoring/85263/4?u=ablackwell1 which might be worth looking at if we decide we want better data on what the service is doing.

playbooks/neo4j.yml
```diff
@@ -15,3 +15,5 @@
     - neo4j
     - role: splunkforwarder
       when: COMMON_ENABLE_SPLUNKFORWARDER
+    - role: newrelic_infrastructure
+      when: COMMON_ENABLE_NEWRELIC_INFRASTRUCTURE
```
adzuci, line 19: I might consider adding a new line or making your editor always add a new line.

PR #6812, opened 2022-09-28
feat: add `edx-exams` to sandbox using docker(PSRE-1925)

nadeemshahzad:
This PR will install `edx-exams` on the Sandbox using docker.

### DIFF
--- a/playbooks/edx_continuous_integration.yml
+++ b/playbooks/edx_continuous_integration.yml
@@ -18,6 +18,7 @@
       - prospectus
       nginx_default_sites:
       - lms
+    - docker-tools
     - mysql
     - role: edxlocal
       tags: edxlocal
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -655,5 +655,114 @@ if [[ $enable_newrelic == "true" ]]; then
     run_ansible run_role.yml -i "${deploy_host}," -e role=newrelic_infrastructure $extra_var_arg  --user ubuntu
 fi
 
+function provision_containerized_app() {
+    echo "#!/usr/bin/env bash"
+    echo "set -ex"
+
+    # Create app staticfiles dir
+    echo "mkdir /edx/var/${app_service_name}/staticfiles/ -p && chmod 777 /edx/var/${app_service_name} -R"
+
+    # Checkout code in app directory
+    echo "cd /edx/app/"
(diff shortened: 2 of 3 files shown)


playbooks/edx_continuous_integration.yml
```diff
@@ -18,6 +18,7 @@
       - prospectus
       nginx_default_sites:
       - lms
+    - docker-tools
```
schenedx, line 21: Just for learning purposes, where is this role defined? I did a string search within the Configuration repo and couldn't find the definition of this role. I do see many playbook include this role tho.
Also, why not create a new playbook to install the docker-tools. This way, we can use logic in `ansible-provision.sh` to specifically run the playbook in the sandbox situation, right?

adzuci: Folder search may help find roles more easily than a string search: https://github.com/openedx/configuration/tree/master/playbooks/roles/docker-tools

I believe this playbook may only be used for sandboxes, despite the name, though I could be wrong.

adzuci (approved):
LGTM, have we spun up a sandbox on this branch to see if docker-compose runs?

playbooks/edx_continuous_integration.yml
schenedx: That make sense. Let's see if this works.

util/jenkins/ansible-provision.sh
```diff
@@ -655,5 +655,114 @@ if [[ $enable_newrelic == "true" ]]; then
     run_ansible run_role.yml -i "${deploy_host}," -e role=newrelic_infrastructure $extra_var_arg  --user ubuntu
 fi
 
+function provision_containerized_app() {
```
adzuci, line 658: In theory could this section be a loop that provisions two or three containerized apps, or would that be too much tech debt?

Also, could we file a PSRE Kaisen backlog ticket to remove this logic once configuration is DEPR'd and add a comment describing how this may be considered tech debt and that it's removal should be considered a priority.  (Part of the goal being to discourage others from building fortresses on top of our Lego blocks)

adzuci (approved):
I'm unsure of how much logic should be here vs in the edx-exams repo, but if this works and @alangsto can confirm that it helps her accomplish her goals, then I think we should merge it.

alangsto (approved):
This looks good to me! I really appreciate all of the comments you left, it makes the code much more understandable.

PR #6824, opened 2022-10-19
feat: run edxapp workers in docker containers on sandbox

NIXKnight:
Configuration Pull Request
---
This PR attempts to use bash scripting with docker containers to run edxapp celery workers instead of using ansible.

The `playbooks/edx_continuous_integration.yml` playbook is updated and the line where `edxapp` role is run with `celery_worker` set to `true` is removed. This way the edxapp celery worker configuration taken away from ansible.

The main script `ansible-provision.sh` is updated to include the worker configuration. The worker configuration is defined in `LC_*` variables. These variables are normally used to pass ssh client locale configuration to an ssh server. However, it can also be used to pass other information through to the remote instance (such as the worker configuration).

once the configuration is exported with `LC_*` prefix, we then initiate an SSH connection to the sandbox and pass in `worker-container-provisioner.sh` script. This script installs `yq` on the sandbox. It uses `yq` to process the configuration passed via `LC_*` variables and checks if the required docker image exists or not. If the image doesn't exist, it builds the docker image. After that, it renders a `docker-compose` file for running all the `edxapp` workers and runs containers using the `docker-compose` file.

At the moment `worker-container-provisioner.sh` script is setup to only work with `edxapp` workers, however it can be updated in the future to work with other celery workers as well.

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -602,6 +602,42 @@ done
 if [[ $reconfigure == "true" || $server_type == "full_edx_installation_from_scratch" || $server_type == "ubuntu_20.04" ]]; then
     cat $extra_vars_file
     run_ansible edx_continuous_integration.yml -i "${deploy_host}," $extra_var_arg --user ubuntu
+    # Export LC_* vars. To be passed to remote instance via SSH where SSH configuration allows LC_* to be accepted as environment variables.
+    # LC_* is normally used for passing through locale settings of SSH clients to SSH servers.
+    export LC_WORKER_CFG=$(cat <<EOF
+worker_cfg:
+  - queue: default
+    service_variant: cms
+    concurrency: 1
+    prefetch_optimization: default
+  - queue: high
+    service_variant: cms
+    concurrency: 1
+    prefetch_optimization: default
+  - queue: default
+    service_variant: lms
+    concurrency: 1
+    prefetch_optimization: default
+  - queue: high
+    service_variant: lms
+    concurrency: 1
+    prefetch_optimization: default
+  - queue: high_mem
+    service_variant: lms
+    concurrency: 1
+    prefetch_optimization: default
+EOF
+)
+    # Remote SSH configuration allows using LC_* (normally for locale variables) to be passed as environment variables to the remote instance.
+    export LC_WORKER_OF="edxapp"
+    export LC_WORKER_IMAGE_NAME="$LC_WORKER_OF"
+    export LC_WORKER_SERVICE_REPO="edx-platform"
+    export LC_SANDBOX_USER="$github_username"
+    ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null ubuntu@${deploy_host} "sudo -n -s bash" < $WORKSPACE/configuration/util/jenkins/worker-container-provisioner.sh
+    unset LC_WORKER_OF
+    unset LC_WORKER_IMAGE_NAME
+    unset LC_WORKER_SERVICE_REPO
+    unset LC_SANDBOX_USER
 fi
 
 if [[ $reconfigure != "true" && $server_type == "full_edx_installation" ]]; then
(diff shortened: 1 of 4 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -602,6 +602,40 @@ done
 if [[ $reconfigure == "true" || $server_type == "full_edx_installation_from_scratch" || $server_type == "ubuntu_20.04" ]]; then
     cat $extra_vars_file
     run_ansible edx_continuous_integration.yml -i "${deploy_host}," $extra_var_arg --user ubuntu
+                export LC_WORKER_CFG=$(cat <<EOF
+worker_cfg:
+  - queue: default
+    service_variant: cms
+    concurrency: 1
```
adzuci, line 612: Is there currently a way to run workers with a concurrency of >1 via the Ansible logic? / Should we add a comment about what this means to have it set to 1?  I like pinning it to 1 worker per queue but could see someone wanting to test race conditions by bumping this to 2.

adzuci (commented):
@NIXKnight Could you add a PR description and is there a sandbox that has been launched off of this branch yet?  I added one comment below, but am not familiar enough with this logic to be able to give an approval just yet.  My suggestion would be to review with @nadeemshahzad and if you are able to launch a sandbox and test that tasks are consumed, then I would feel comfortable with merging it.

util/jenkins/worker-container-provisioner.sh
```diff
@@ -0,0 +1,68 @@
+#!/usr/bin/env bash
```
adzuci, line 1: Do we plan for this logic to live here long term?  Also, should we add a comment detailing the usage and purpose of this script?

util/jenkins/ansible-provision.sh
NIXKnight: @adzuci our sandbox builds, currently use fixed configuration where the default concurrency levels are picked up whenever a sandbox build is run. There is currently no way to increase this concurrency for sandbox builds.

I am not sure if there is a need to bump concurrency for these workers on sandbox instances. If it is ever needed, we will have to update our sandbox build jobs as well as this script where we will have to provide an option to increase concurrency for each worker.

util/jenkins/worker-container-provisioner.sh
NIXKnight: @adzuci yes. Our main script `ansible-provision.sh` is quite large (over 750+ lines of code). `worker-container-provisioner.sh` script is also passed via SSH to the remote node for execution (hence the need to keep it separate). I believe that these scripts should be separate of the main script.

NIXKnight:
> @NIXKnight Could you add a PR description and is there a sandbox that has been launched off of this branch yet? I added one comment below, but am not familiar enough with this logic to be able to give an approval just yet. My suggestion would be to review with @nadeemshahzad and if you are able to launch a sandbox and test that tasks are consumed, then I would feel comfortable with merging it.

@adzuci PR description has been updated. I'll get in touch with @nadeemshahzad for a review. I have started a new sandbox.

adzuci (dismissed):

util/jenkins/ansible-provision.sh
adzuci: 👍

util/jenkins/worker-container-provisioner.sh
```diff
@@ -0,0 +1,68 @@
...
+)
+EOF
+}
+
+install_pre_reqs
+
+# Check if docker image already exists. If it doesn't, build it.
+if ! $(docker image inspect ${LC_WORKER_IMAGE_NAME}:latest >/dev/null 2>&1 && echo true || echo false) ; then
+  cd /edx/app/${LC_WORKER_OF}/${LC_WORKER_SERVICE_REPO}
+  docker build . -t ${LC_WORKER_IMAGE_NAME}:latest --target base
```
adzuci: Can we pull in and use https://github.com/edx/internal-dockerfiles/blob/master/edx-platform-private/Dockerfile#L5 instead?  (& add a comment at the top of this file saying this is a POC and will likely not work for most people)

schenedx (approved):
👍 Could you please bring the PR out of the draft state?
I think we can always adjust whatever is worth to change in subsequent PRs.

nadeemshahzad (approved):

alangsto (approved):

adzuci (approved):
LGTM!

PR #6825, opened 2022-10-19
feat: add fluentd container for logging

alangsto:
Configuration Pull Request
---
Add a fluentd container that handles tracking logs from edxapp. When edxapp is containerized, the logging config will be configured to send tracking logs to a specified port. Fluentd is configured to listen to that port, and then output to std.out. In the future, fluentd could be configured to also output to S3.

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -788,5 +788,42 @@ EOF
     rm -f "${provision_script}"
 fi
 
+function provision_fluentd() {
+    echo "#!/usr/bin/env bash"
+    echo "set -ex"
+
+    echo "docker pull fluent/fluentd:edge-debian"
+
+    echo "fluentd_config=/var/tmp/fluentd.conf"
+    echo "cat << 'EOF' > \$fluentd_config
+    <source>
+        @type syslog
+        port 5140
+        bind 0.0.0.0
+        tag *
+        <transport udp>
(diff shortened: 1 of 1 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -788,5 +788,37 @@ EOF
...
+        <parse>
+            @type none
+        </parse>
+    </source>
+
+    <match **>
+        @type stdout
+    </match>
+EOF"
+    echo "docker run -d -u root -v /var/tmp/fluentd.conf:/fluentd/etc/fluentd.conf -v /edx/var/log/tracking:/edx/var/log/tracking fluent/fluentd:edge-debian -c /fluentd/etc/fluentd.conf"
```
alangsto: have to run as root user in order to properly mount the tracking log directory. Depending on how we use fluentd as a logging driver for edxapp, I'm not sure we'll need to do the same thing

schenedx (commented):
Looks pretty good to me. When do you think this is out of the draft state? Maybe after https://github.com/openedx/edx-platform/pull/31217 is merged?

alangsto:
The platform PR should be merged, but I'd feel more comfortable merging this once we have the CMS/LMS actually running in a container. 

If we want to merge this now, it should definitely be put behind some sort of gate.

schenedx (approved):
👍

PR #6826, opened 2022-10-21
feat: Comment this out since it's a footgun

adzuci:
Configuration Pull Request
---
Commenting this out to avoid it being run, leaving it here in case it's needed though.


playbooks/splunk.yml
```diff
@@ -1,8 +1,8 @@
-# Usage: ansible-playbook splunk.yml -e@/path/to/environment-deployment.yml
+# # Usage: ansible-playbook splunk.yml -e@/path/to/environment-deployment.yml
```
matthugs, line 2: I would've looked for a DO NOT USE comment; could we add one of those as well?

matthugs (approved):

playbooks/splunk.yml
adzuci: Yes

PR #6836, opened 2022-11-14
temp: enable debug logging for the New Relic agent on stage

rgraber:
Following guidance at https://2u-internal.atlassian.net/wiki/spaces/SRE/pages/19254057/How+to+enable+New+Relic+Debug+and+Audit+logs. Necessary to debug some issues with the ticket mentioned in the comments.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/templates/newrelic.ini.j2
+++ b/playbooks/roles/edxapp/templates/newrelic.ini.j2
@@ -28,10 +28,9 @@
 #
 browser_monitoring.attributes.enabled=true
 
-
+{# # Temporarily turn on logging to debug https://github.com/openedx/event-bus-kafka/issues/45. If that ticket is closed,
+# this if-block can be removed  -#}
 {%- if COMMON_ENVIRONMENT == "stage" %}
-# Temporarily turn on logging to debug https://github.com/openedx/event-bus-kafka/issues/45. If that ticket is closed,
-# this block can be removed
 audit_log_file = /tmp/newrelic-python-agent-audit.log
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/templates/newrelic.ini.j2
```diff
@@ -27,3 +27,11 @@
 # `course_id`.
 #
 browser_monitoring.attributes.enabled=true
+
+# Temporarily turn on logging to debug https://github.com/openedx/event-bus-kafka/issues/45. If that ticket is closed,
+# the following three lines can be removed.
```
matthugs: s/three/five/

you may also want to include this inside the {% if %}. I would expect this to template out such that this comment gets included in the template whether or not the environment is stage. (I think there's also syntax for template comments but I don't remember it offhand)

matthugs (approved):
This can be tested by building a sandbox against your branch I believe. Time-consuming, but if you're already multitasking not too bad

rgraber:
> This can be tested by building a sandbox against your branch I believe. Time-consuming, but if you're already multitasking not too bad

Will that allow me to see what the file will look like on stage?

matthugs:
> > This can be tested by building a sandbox against your branch I believe. Time-consuming, but if you're already multitasking not too bad
> 
> Will that allow me to see what the file will look like on stage?

not without also changing https://github.com/edx/sandbox-internal/blob/fb16321d0772adee54327d9bccebf0a732ddacc5/ansible/vars/developer-sandbox.yml#L95, good catch. (That, naturally, I would leave to a branch & not try to merge)

PR #6839, opened 2022-11-17
fix: change fluentD configs for tail logs

ohnickmoy:
Configuration Pull Request
---

Change fluentD configurations to tail a rotating file handler

Mounts the rotating file handler on sandbox onto the container 

### DIFF
--- a/util/jenkins/ansible-provision.sh
+++ b/util/jenkins/ansible-provision.sh
@@ -803,8 +803,8 @@ function provision_fluentd() {
     echo "cat << 'EOF' > \$fluentd_config
     <source>
         @type tail
-        path /var/tmp/logs.log
-        pos_file /var/log/log_file.pos
+        path /var/tmp/tracking_logs.log
+        pos_file /var/log/tracking_logs.pos
         tag *
         <parse>
             @type none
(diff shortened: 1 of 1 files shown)


util/jenkins/ansible-provision.sh
```diff
@@ -802,12 +802,10 @@ function provision_fluentd() {
     echo "fluentd_config=/var/tmp/fluentd.conf"
     echo "cat << 'EOF' > \$fluentd_config
     <source>
-        @type syslog
-        port 5140
-        bind 0.0.0.0
+        @type tail
+        path /var/tmp/tracking_logs.log
+        pos_file /var/log/tracking_logs.pos
```
alangsto, line 807: Is any additional configuration needed for the file rotation? Specifically thinking about https://docs.fluentd.org/input/tail#rotate_wait

ohnickmoy: It depends on how much we're comfortable with the delay and i suppose how much these tracking logs get appended to. default is 5, but would say...idk 20 seconds be a big enough window for rotate_wait?

The other settings that come to mind is probably [encoding](https://docs.fluentd.org/input/tail#encoding-from_encoding)  (by default its ascii-8bit)

ohnickmoy: @alangsto, spoke with @NIXKnight in sync meeting and he thinks 10 second window might be good. Too small of a window?

adzuci (approved):
I think I understand this and it looks good to merge.

PR #6840, opened 2022-11-22
Use sysvinit instead of systemd module

aht007:
Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/handlers/main.yml
+++ b/playbooks/roles/nginx/handlers/main.yml
@@ -1,10 +1,10 @@
 ---
 - name: restart nginx
-  service:
+  ansible.builtin.sysvinit:
     name: nginx
     state: restarted
 
 - name: reload nginx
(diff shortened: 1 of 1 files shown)


matthugs (commented):
This role is used across a large number of different playbooks, and as a dependency of many other roles. If we need to go this route on a permanent basis, it may make sense to parameterise this role to use sysvinit when a flag is set, and the service module otherwise. This would help reduce the blast-radius of this change

aht007:
@matthugs I have tested the configuration changes on docker as well `ec2` machine by building a [Sandbox](https://tools-edx-jenkins.edx.org/job/sandboxes/job/CreateSandbox/50406/). Also as per my discussion with @nadeemshahzad we still have support for sysvinit in ubuntu 20.04 and also in docker so it shouldn't cause any issue with anything. In the light of all the above things I don't think we need any further changes right now.

PR #6876, opened 2023-02-02
chore: retain cache and data folders for prospectus builds

nadeemshahzad:
PR will ensure that cache and data directories for prospectus builds are not removed so we can speed up its build process in GoCD
Ticket: https://2u-internal.atlassian.net/browse/PSRE-2174

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/prospectus/tasks/main.yml
+++ b/playbooks/roles/prospectus/tasks/main.yml
@@ -5,6 +5,20 @@
     state: absent
   when: PROSPECTUS_GIT_IDENTITY != "none"
 
+- name: check if cache dir exists
+  stat:
+     path: "{{ prospectus_code_dir }}/.cache"
+  register: register_cache_dir
+
+- name: check if data dir exists
+  stat:
+     path: "{{ PROSPECTUS_DATA_DIR }}"
+  register: register_data_dir
+
+- name: move cache dir to temp
+  command: mv {{ prospectus_code_dir }}/.cache /tmp/
+  when: PROSPECTUS_RETAIN_CACHE_PUBLIC_DIRS and register_cache_dir.stat.exists
+
 - name: Remove old git repo
   file:
     state: absent
@@ -19,6 +33,7 @@
   file:
     state: absent
     path: "{{ PROSPECTUS_DATA_DIR }}"
+  when: not PROSPECTUS_RETAIN_CACHE_PUBLIC_DIRS
 
 - name: Create prospectus app folder
   file:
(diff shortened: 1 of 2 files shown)


syedimranhassan (approved):

adzuci:
@bseverino What do you think, do you want to try this and see if the pipeline breaks?

bseverino:
@marekwrobel Does https://2u-internal.atlassian.net/browse/WS-3675 need to be finished before we do this?

bseverino:
Also @nadeemshahzad what are our options for clearing the cache between builds if needed?

marekwrobel:
I have merged [this PR ](https://github.com/edx/prospectus/pull/5199)that removes any manual usage of cached data + displays one cached item from a previous build (just to confirm we can successfully retain cached data).

We can now merge **this** PR and see if it works as expected. Once this is confirmed, I will set up a PR with the changes in the sourcing plugin to actually use the timestamps.

I think one way to implement a full build is just by defining an env variable (in gocd click on the play with '+' and set given env variable that will be used during sourcing and will fetch all courses and programs). But on the other hand, having an easy way to just clear that cache would also be useful. Happy to learn what our options are.

@bseverino @nadeemshahzad

playbooks/roles/prospectus/tasks/main.yml
```diff
@@ -5,6 +5,20 @@
...
+     path: "{{ prospectus_code_dir }}/.cache"
+  register: register_cache_dir
+
+- name: check if data dir exists
+  stat:
+     path: "{{ PROSPECTUS_DATA_DIR }}"
+  register: register_data_dir
+
+- name: move cache dir to temp
+  command: mv {{ prospectus_code_dir }}/.cache /tmp/
```
matthugs: could we use a more specific subdirectory of `/tmp/`? `/tmp/.cache` is pretty obscure a name for this specific a usecase

nadeemshahzad:
@bseverino @marekwrobel The variable `PROSPECTUS_RETAIN_CACHE_PUBLIC_DIRS` is being used to enable or disable the cache for Prospectus build. By default, its value is set to false, but it is being changed to true in [edx-internal PR](https://github.com/edx/edx-internal/pull/8164) for the purpose of initially implementing and testing this change only for the stage environment.

matthugs (approved):

playbooks/roles/prospectus/tasks/main.yml
nadeemshahzad: Added a task to use subdirectory `/tmp/cache-data` to move `.cache' folder

PR #6889, opened 2023-03-08
feat: add request tracing in nginx logs using X-Request-ID header

nadeemshahzad:
PR is using two map directives in Nginx config block to track and log value X-REQUEST-ID http header.
The first map, named "$http_x_request_id" is used to capture the X-Request-Id header value that matches the regular expression "~*" and assigns it to the variable "$req_id". If the header is not present, it assigns the value of ["${http_cf_ray}" ](https://developers.cloudflare.com/fundamentals/get-started/reference/http-request-headers/#cf-ray)  to "$req_id".

The second map, named "$req_id", assigns the value of "${request_id}" to the variable "$uuid" if "$req_id" is empty if it is not empty "$uuid" will be set to "$req_id" from first map directive.

Then it is setting X-Request-ID header of HTTP request to the value of the "$uuid" variable which will be logged to Nginx log files and will be forwarded to the application server. 

`Testing`

if x-requst-id is present

```
curl https://courses.stage.edx.org -H "X-Request-ID: custom-x-id-request-header"

3.230.216.2, 172.70.38.183 - 10.x.0.239 - - https [08/Mar/2023:16:04:45 +0000]  custom-x-id-request-header "GET / HTTP/1.1" 302 0 0.029 "-" "curl/7.79.1"
```
if cf-ray header is present

```
curl https://courses.stage.edx.org 
3.230.216.2, 172.70.38.201 - 10.x.0.239 - - https [08/Mar/2023:16:04:01 +0000]  7a4c4fc70c5338a0-IAD "GET / HTTP/1.1" 302 0 0.034 "-" "curl/7.79.1"
```
Nginx built in request id for missing x-requst-id and cf-ray 
```
- - 10.x.1.110 - - - [08/Mar/2023:16:05:02 +0000]  70992995c6ab1e3795f7a1143fc53aab "GET /heartbeat HTTP/1.1" 200 122 0.057 "-" "ELB-HealthChecker/1.0"
```

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
+++ b/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
@@ -39,7 +39,7 @@ http {
         # Logging Settings
         ##
 
-        log_format p_combined '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local]  '
+        log_format p_combined '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local] $uuid  '
                             '"$request" $status $body_bytes_sent $request_time '
                             '"$http_referer" "$http_user_agent"';
 
--- a/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/app.j2
@@ -4,6 +4,7 @@
 
 {% include "concerns/upstream.j2"%}
 {% include "concerns/cors-build-map.j2" %}
+{% include "concerns/x-request-id-map.j2" %}
 
 server {
   server_name {{ edx_django_service_hostname }};
@@ -28,6 +29,7 @@ server {
   ssl_certificate /etc/ssl/certs/{{ NGINX_SSL_CERTIFICATE|basename }};
   ssl_certificate_key /etc/ssl/private/{{ NGINX_SSL_KEY|basename }};
   add_header Strict-Transport-Security "max-age=31536000; includeSubDomains";
+  add_header X-Request-ID $uuid;
 
   {% include "concerns/app-common.j2" %}
 }
--- a/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/proxy-to-app.j2
+++ b/playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/proxy-to-app.j2
@@ -22,10 +22,12 @@ location @proxy_to_app {
   proxy_set_header X-Forwarded-Proto $scheme;
   proxy_set_header X-Forwarded-Port $server_port;
   proxy_set_header X-Forwarded-For $remote_addr;
+  proxy_set_header X-Request-ID $uuid;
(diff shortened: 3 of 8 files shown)


playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/request-id-map.j2
```diff
@@ -0,0 +1,9 @@
+map $http_x_request_id $req_id {
```
matthugs: ok, so the former comes from http://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_, the latter is being defined

but the name `$req_id` seems like it should be an abbreviation for "request id", which is the name of a built-in variable used in the same file. These two variables are going to diverge, though. Could we choose a different name instead of `$req_id`?

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/request-id-map.j2
```diff
@@ -0,0 +1,9 @@
...
+  ""        "${http_cf_ray}";
+  ~*        "${http_x_request_id}";
```
matthugs: would `default` also work here in place of the regex? If so it would make this a bit more self-consistent in approach

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -39,7 +39,7 @@ http {
         # Logging Settings
         ##
 
-        log_format p_combined '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local]  '
+        log_format p_combined '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local] $uuid '
```
matthugs: There used to be two spaces at the end but now there's only one. Do we think it'd matter? (e.g. I wonder if any regular expressions have been written assuming two spaces there)

matthugs (approved):

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/request-id-map.j2
nadeemshahzad: Yes, it would be a good idea to choose a different name for the variable to avoid confusion and potential bugs in the future. I've updated the PR and replaced `req_id` variable name with`trace_id`

playbooks/roles/edx_django_service/templates/edx/app/nginx/sites-available/concerns/request-id-map.j2
nadeemshahzad: It makes sense to use `default` for all requests if it is not empty. Updated the PR to address this comment.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
nadeemshahzad: thanks, @matthugs I've followed the old pattern for log_format and updated the PR with two spaces at the end of the first line.

PR #6890, opened 2023-03-08
fix: "Host key verification failed." error for sandbox build with pri…

christopappas:
…vate repo

Configuration Pull Request
---

### DIFF
--- a/util/jenkins/app-container-provisioner.sh
+++ b/util/jenkins/app-container-provisioner.sh
@@ -45,7 +45,12 @@ if [[ ! -d "/edx/app/${app_name}/${app_repo}" ]] ; then
 
     # use SSH to clone if repo is private
     if [[ "$app_repo_is_private" = true ]] ; then
-        git clone git@github.com:edx/${app_repo}.git /edx/app/${app_name}/${app_repo}
+        set +x
+        echo -e "${app_git_ssh_key}" > /tmp/${app_service_name}_ssh_key
+        set -x
+        chmod 0600 /tmp/${app_service_name}_ssh_key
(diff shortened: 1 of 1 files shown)


util/jenkins/app-container-provisioner.sh
```diff
@@ -45,7 +45,12 @@ if [[ ! -d "/edx/app/${app_name}/${app_repo}" ]] ; then
 
     # use SSH to clone if repo is private
     if [[ "$app_repo_is_private" = true ]] ; then
-        git clone git@github.com:edx/${app_repo}.git /edx/app/${app_name}/${app_repo}
+        set +x
+        echo -e "${app_git_ssh_key}" > /tmp/${app_service_name}_ssh_key
+        set -x
+        chmod 0600 /tmp/${app_service_name}_ssh_key
+        useradd -m -d /edx/var/${app_service_name} ${app_service_name} -G www-data
+        GIT_SSH_COMMAND="ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /tmp/${app_service_name}_ssh_key" git clone git@github.com:edx/${app_repo}.git /edx/app/${app_name}/${app_repo}
```
matthugs, line 53: would it make sense to factor out a bash function for this, so this can share its general code-shape with what we do for themes?

matthugs (approved):
fine to move forward whether or not you act on the below

util/jenkins/app-container-provisioner.sh
christopappas: this is a good call out! if this works, i can make a follow up

PR #6892, opened 2023-03-09
feat: add request tracing in nginx logs using X-Request-ID header

nadeemshahzad:
Reverts openedx/configuration#6891
The Discovery GoCD pipeline failed with the below error after I merged this PR to enable request tracing in Nginx https://github.com/openedx/configuration/pull/6889

```
fatal: [10.3.120.228]: FAILED! => {
    "changed": true,
    "cmd": [
        "nginx",
        "-t"
    ],
    "delta": "0:00:00.006049",
    "end": "2023-03-09 10:04:53.233906",
    "rc": 1,
    "start": "2023-03-09 10:04:53.227857"
}
```
I debugged the cause and found that in my PR https://github.com/openedx/configuration/pull/6889 I have defined the maps in the virtual host config to set a value for the custom variable `uuid` and used it in `nginx.conf` file to add its value to logs. During Nginx installation, we are testing Nginx configs with Ansible task https://github.com/openedx/configuration/blob/master/playbooks/roles/nginx/tasks/main.yml#L399 which was giving an error for undefined `uuid` variable as virtual host was not yet created where I have defined this variable. So I reverted my PR https://github.com/openedx/configuration/pull/6891 to unblock the pipelines and made this PR with fix to define the `uuid` maps in `nginx.conf` file.
Also made this PR https://github.com/edx/edx-internal/pull/7960 to turn on this feature for our environments.

### DIFF
--- a/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
+++ b/playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
@@ -35,6 +35,20 @@ http {
         include /etc/nginx/mime.types;
         default_type application/octet-stream;
 
+        {% if NGINX_ENABLE_REQUEST_TRACKING_ID -%}
+        # Setting variables based on HTTP headers to track and differentiate requests across components: the first creates a new
+        # variable called trace_id based on X-REQUEST-ID or Cloudflare CF-ray headers, while the second creates a variable named
+        # request_tracking_id based on whether trace_id is set or not.
+        map $http_x_request_id $trace_id {
+          ""        "${http_cf_ray}";
+          default   "${http_x_request_id}";
+        }
+        map $trace_id $request_tracking_id {
+          ""        "${request_id}";
+          default   $trace_id;
+        }
+        {% endif %}
+
         ##
         # Logging Settings
         ##
@@ -52,7 +66,13 @@ http {
                                'referer=$http_referer user_agent="$http_user_agent" upstream_addr=$upstream_addr upstream_status=$upstream_status '
                                'request_time=$request_time request_id=$request_id upstream_response_time=$upstream_response_time '
                                'upstream_connect_time=$upstream_connect_time upstream_header_time=$upstream_header_time';
-
+        
+        {% if NGINX_ENABLE_REQUEST_TRACKING_ID -%}
+        log_format cf_custom '$http_x_forwarded_for - $remote_addr - $remote_user $http_x_forwarded_proto [$time_local] $request_tracking_id  '
+                            '"$request" $status $body_bytes_sent $request_time '
+                            '"$http_referer" "$http_user_agent"';
+        {% endif %}
+        
         log_format json_analytics escape=json '{'
                    '"msec": "$msec", ' # request unixtime in seconds with a milliseconds resolution
                    '"connection": "$connection", ' # connection serial number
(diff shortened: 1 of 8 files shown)


playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -35,6 +35,21 @@ http {
         include /etc/nginx/mime.types;
         default_type application/octet-stream;
 
+        {% if NGINX_ENABLE_UUID %}
```
matthugs:
```suggestion
        {%- if NGINX_ENABLE_UUID %}
```
you can avoid adding unnecessary whitespace to these files using jinja syntax like this

the minus sign on one side or the other essentially means "strip out whitespace to this side of this block", so that we don't have the indentation of the `{% if %}` block itself included in the rendered output.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -35,6 +35,21 @@ http {
...
+        # Maps to trace requests and identifying them across different components. First map creates a new variable named
+        # trace_id, which is set to either the value of X-REQUEST-ID or Cloudflare CF-ray headers depending on whether 
+        # http_x_request_id is empty or not. Second map creates a new variable named uuid, which is set to either the 
+        # value of trace_id or request_id, depending on whether trace_id is set or not in first map.
+        map $http_x_request_id $trace_id {
+          ""        "${http_cf_ray}";
+          default   "${http_x_request_id}";
+        }
+        map $trace_id $uuid {
```
matthugs: I think uuid may be a bad choice of name for this variable because in some cases it will take the value of the CF-Ray header, which is not formatted as a UUID.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
```diff
@@ -35,6 +35,21 @@ http {
         include /etc/nginx/mime.types;
         default_type application/octet-stream;
 
+        {% if NGINX_ENABLE_UUID %}
+        # Maps to trace requests and identifying them across different components. First map creates a new variable named
```
matthugs: [nit] I think we should probably avoid the word "map" in this comment, because while that's an nginx feature we're using here it's not a good word to use to describe what we're doing at a higher level. I would probably use words like "ternary" or "setting a value conditionally" to explain this. "Map" usually means something very different, i.e. transforming values drawn from an array and outputting an array of "mapped" values (which is how the word is used in a functional programming context).

matthugs (approved):

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
nadeemshahzad: thanks, @matthugs I've updated the template with the control character to remove space after.

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
nadeemshahzad: Replaced variable `uuid` with `request_tracking_id`

playbooks/roles/nginx/templates/etc/nginx/nginx.conf.j2
nadeemshahzad: Addressed and updated the comment with note.

Setting variables based on HTTP headers to track and differentiate requests across components: the first creates a new
 variable called trace_id based on X-REQUEST-ID or Cloudflare CF-ray headers, while the second creates a variable named
  request_tracking_id based on whether trace_id is set or not.

PR #6916, opened 2023-04-19
add fail2ban

katebygrace:
Add fail2ban

Request for review of the tags specifically, seems reasonable. 

We're getting a lot of attempted SSH connections on our sandbox instances. 

---

### DIFF
--- a/playbooks/roles/edx_ansible/defaults/main.yml
+++ b/playbooks/roles/edx_ansible/defaults/main.yml
@@ -19,6 +19,8 @@
 # included in the play.
 EDX_ANSIBLE_DUMP_VARS: false
 
+edx_ansible_debian_running_services: 
+  - fail2ban
 
 edx_ansible_debian_pkgs_default:
   - python-apt
@@ -44,7 +46,7 @@ edx_ansible_release_specific_debian_pkgs:
   focal:
     - python3-dev
 
-edx_ansible_debian_pkgs: "{{ edx_ansible_debian_pkgs_default + edx_ansible_release_specific_debian_pkgs[ansible_distribution_release] }}"
+edx_ansible_debian_pkgs: "{{ edx_ansible_debian_running_services + edx_ansible_debian_pkgs_default + edx_ansible_release_specific_debian_pkgs[ansible_distribution_release] }}"
 
 edx_ansible_app_dir: "{{ COMMON_APP_DIR }}/edx_ansible"
 edx_ansible_code_dir: "{{ edx_ansible_app_dir }}/edx_ansible"
--- a/playbooks/roles/edx_ansible/tasks/main.yml
+++ b/playbooks/roles/edx_ansible/tasks/main.yml
@@ -53,6 +53,14 @@
   tags:
     - install:system-requirements
 
+- name: Start and enable running services 
+  ansible.builtin.systemd:
+    state: started
+    enabled: true
+    name: "{{ edx_ansible_debian_running_services }}"
+  tags: 
+    - install:system-requirements
+
 - include: deploy.yml
   tags:
     - deploy
--- a/playbooks/roles/edx_ansible/tasks/main.yml
+++ b/playbooks/roles/edx_ansible/tasks/main.yml
@@ -53,6 +53,10 @@
   tags:
     - install:system-requirements
 
+- include: deploy.yml
+  tags:
+    - deploy
+    
 - name: Start and enable running services 
   ansible.builtin.systemd:
     state: started
@@ -60,7 +64,3 @@
     name: "{{ edx_ansible_debian_running_services }}"
   tags: 
     - install:system-requirements
-
-- include: deploy.yml
-  tags:
-    - deploy


playbooks/roles/edx_ansible/defaults/main.yml
```diff
@@ -19,6 +19,8 @@
 # included in the play.
 EDX_ANSIBLE_DUMP_VARS: false
 
+edx_ansible_debian_running_services: 
+  - fail2ban
```
pshiu, line 23: [no action needed] Checked that:

* iptables is available on a sandbox:

  ```
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ iptables -V
  iptables v1.8.4 (legacy)
  ```

* The default install on a sandbox only affects SSH:

  ```
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo apt-get install fail2ban
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo fail2ban-client status
  Status
  |- Number of jail:	1
  `- Jail list:	sshd
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ ls /etc/fail2ban/jail.d
  defaults-debian.conf
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ cat /etc/fail2ban/jail.d/defaults-debian.conf
  [sshd]
  enabled = true
  ```

* We are logging ssh connection attempts where we expect them:

  ```
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo fail2ban-client status sshd
  Status for the jail: sshd
  |- Filter
  |  |- Currently failed:	0
  |  |- Total failed:	0
  |  `- File list:	/var/log/auth.log
  `- Actions
   |- Currently banned:	0
   |- Total banned:	0
   `- Banned IP list:	
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo grep 'sshd.*<my-username>'  /var/log/auth.log | tail -3
  Apr 20 13:19:45 <sandbox-name> sshd[1421022]: pam_unix(sshd:session): session opened for user <my-username> by (uid=0)
  Apr 20 13:19:46 <sandbox-name> sshd[1421075]: Starting session: shell on pts/0 for <my-username> from <my-ip> port 57106 id 0
  ```

* Logging in as a fake user from my local in another Terminal window causes a ban:

  ```
  (workarea) ~ % date && ssh fake@<sandbox-name>.sandbox.edx.org
  Thu Apr 20 09:39:51 EDT 2023
  fake@<sandbox-name>.sandbox.edx.org: Permission denied (publickey).
  (workarea) ~ % date && ssh fake@<sandbox-name>.sandbox.edx.org
  Thu Apr 20 09:39:54 EDT 2023
  fake@<sandbox-name>.sandbox.edx.org: Permission denied (publickey).
  (workarea) ~ % date && ssh fake@<sandbox-name>.sandbox.edx.org
  Thu Apr 20 09:39:55 EDT 2023
  fake@<sandbox-name>.sandbox.edx.org: Permission denied (publickey).
  (workarea) ~ % date && ssh fake@<sandbox-name>.sandbox.edx.org
  Thu Apr 20 09:39:56 EDT 2023
  fake@<sandbox-name>.sandbox.edx.org: Permission denied (publickey).
  (workarea) ~ % date && ssh fake@<sandbox-name>.sandbox.edx.org
  Thu Apr 20 09:39:57 EDT 2023
  fake@<sandbox-name>.sandbox.edx.org: Permission denied (publickey).
  (workarea) ~ % date && ssh fake@<sandbox-name>.sandbox.edx.org
  Thu Apr 20 09:39:58 EDT 2023
  ssh: connect to host <sandbox-name>.sandbox.edx.org port 22: Connection refused
  ```

    * As soon as I finished the test my existing (valid) login session running from another terminal window no longer responded.
    
    * This may be interesting if we ever SSH to the machine from a common or pooled IP address.

* My banned IP gets unbanned 10 minutes later:

  ```
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo tail /var/log/fail2ban.log 
  2023-04-20 13:39:58,312 fail2ban.actions        [1419298]: NOTICE  [sshd] Ban <my-ip> 
  2023-04-20 13:49:58,989 fail2ban.actions        [1419298]: NOTICE  [sshd] Unban <my-ip> 
  ```

pshiu, line 23: The default seems to be ban for 10 minutes on 5th bad attempt in 10 minutes. Does this seem reasonable? What do you think about making it more generous?

  ```
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo fail2ban-client get sshd findtime
  600
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo fail2ban-client get sshd bantime
  600
  <my-username>@<sandbox-name> <ec2-instance-id>:~$ sudo fail2ban-client get sshd maxretry
  5
  ```

pshiu, line 23: I'm not the best Ansible expert, but this looks like it will apply to all environments, not just sandboxes, and including stage/prod.

If the objective is to target stage/prod, should we roll this out to sandboxes then stage first via configuration to test?

pshiu, line 23: Should we configure `ignoreip` [[docs](https://www.fail2ban.org/wiki/index.php/Whitelist), [source](https://github.com/fail2ban/fail2ban/blob/2b98f461bb93c6a1a1cec1481fa94b7a84342977/config/jail.conf#L89-L92)] to ignore IPs from, e.g., our VPN, the boxes SRE usually uses to jump to other boxes, etc.?

pshiu (commented):
Thanks for this work, @katebygrace!

playbooks/roles/edx_ansible/defaults/main.yml
katebygrace: My inclination is always to start to the default and see if we get any decrease in noise - but I am open to other opinions!

adzuci: That's a decent suggestion, though all of our engineers use keys, so I don't see a risk of lockouts personally and feel that we can fix forward if issues are detected.

If we're really concerned, we could monitor Splunk for lockout logs.

adzuci: I think the original concern was sandboxes, but I like the idea of having this enabled everywhere, if we wanted to restrict it to just some environments we could add a `when` clause, default it to false and add a true variable to `sandbox-internal/blob/master/sandbox-remote-config` or other places.

adzuci (approved):
This looks good to me & I've reviewed https://github.com/fail2ban/fail2ban which looks like a reasonable library, though it could be worth asking Axim if they feel a https://discuss.openedx.org post would be worthwhile.

My feeling is that most operators are moving off of this code and that we can fix forward If we run into issues.

playbooks/roles/edx_ansible/defaults/main.yml
```diff
@@ -44,7 +46,7 @@ edx_ansible_release_specific_debian_pkgs:
   focal:
     - python3-dev
 
-edx_ansible_debian_pkgs: "{{ edx_ansible_debian_pkgs_default + edx_ansible_release_specific_debian_pkgs[ansible_distribution_release] }}"
+edx_ansible_debian_pkgs: "{{ edx_ansible_debian_running_services + edx_ansible_debian_pkgs_default + edx_ansible_release_specific_debian_pkgs[ansible_distribution_release] }}"
```
adzuci, line 49: Was there a strong reason not to add the package to `edx_ansible_debian_pkgs_default`?

adzuci: Ah, I see you need to be able to tell [this task](https://github.com/openedx/configuration/pull/6916/files#diff-84a6d843577a6e6bc426bff09dcd6b31a6daeefbce858418710b601dfdb7a362R60) which services need to be started.

katebygrace: It needed to be enabled and started in addition to installed, so kept it separate for that reason.

playbooks/roles/edx_ansible/defaults/main.yml
pshiu: Very good! To be pedantic/explicit, fail2ban operates on all forms of authentication, not just password-based attempts. So, for example, if:

* a user doesn't have their keys set up correctly
* a server accesses another server via SSH using a rotated/bad key (e.g., Jenkins?)

The IP in question (and any other users of that IP) will be blocked as well.

playbooks/roles/edx_ansible/tasks/main.yml
```diff
@@ -56,3 +56,11 @@
 - include: deploy.yml
   tags:
     - deploy
+    
+- name: Start and enable running services 
+  ansible.builtin.systemd:
```
adzuci, line 61: [nit 1] I think we can remove `ansible.builtin.`

[nit 2] Also we may want to look for prior art using the service: https://stackoverflow.com/questions/43974099/in-ansible-whats-the-diffence-between-the-service-and-the-systemd-modules

PR #6921, opened 2023-04-21
fix: remove JWT_DECODE_HANDLER override

robrap:
Remove the override of JWT_DECODE_HANDLER in the settings files, because it already matches the default set in base.py, and when we wish to remove this custom
jwt_decode_handler, it is possible that this setting will unnecessarily cause an error in the future. See
https://github.com/openedx/ecommerce/blob/b5fa2e55e3d4f5f8224d381317177cf4086b35a6/ecommerce/settings/base.py#L449

The JWT_AUTH dictionary should not be replaced as a whole, but should only replace keys that are declared, according to this:
https://github.com/openedx/ecommerce/blob/b5fa2e55e3d4f5f8224d381317177cf4086b35a6/ecommerce/settings/production.py#L51-L53

Note: This change is very similar to https://github.com/edx/edx-internal/pull/8163 which already landed.

Configuration Pull Request
---


pshiu (approved):
Thanks, @robrap!

pshiu:
Note: We are having a temporary issue with the Jenkins job that builds the ecommerce Docker image for devstack.

Until this PR can be incorporated into devstack via that Jenkins job, developers experiencing issues like:

```
ImportError: Could not import 'ecommerce.extensions.api.handlers.jwt_decode_handler' for API setting 'JWT_DECODE_HANDLER'. ModuleNotFoundError: No module named 'ecommerce.extensions.api.handlers'.
```

after bringing up a new container of the ecommerce devstack can make it work by temporarily adding:

```
'JWT_DECODE_HANDLER': 'edx_rest_framework_extensions.auth.jwt.decoder.jwt_decode_handler',
```
to [this dictionary](https://github.com/openedx/ecommerce/blob/43b7e87b68882bf7c4d74bd9503a83e822a99ed3/ecommerce/settings/devstack.py#L34-L48) on their local. (Apologies for the inconvenience.)

robrap:
@pshiu: Will this PR temporarily fix the issue for everyone without special instructions? https://github.com/openedx/ecommerce/pull/3958. I did not test.

pshiu:
Great idea, thanks @robrap! Just gave a thumb and added an AC to https://github.com/edx/upgrades/issues/109.

robrap:
Great @pshiu. As I noted, this is **untested**, so just make sure it resolves your issue. Thanks.

pshiu:
@robrap, yep, @jsnwesson and I had tested an almost identical patch right before I saw your comment.

PR #6997, opened 2023-08-25
fix: Payment MFE GIT Version

grmartin:
Payment MFE (frontend-app-payment) ignored the selected git branch (`version` in ansible's git plugin) this simply allows it to funnel through.

This is part of THES-268.


pshiu (approved):
Nice find!

PR #7083, opened 2023-12-20
fix: failure trying to create specific kind of user via ansible

matthugs:
specifically when running the create db and users runbook which is used to set up users for mysql databases as documented here: https://2u-internal.atlassian.net/wiki/spaces/SRE/pages/19252888/How+to+Create+and+Update+Databases+and+Database+Users

Configuration Pull Request
---

### DIFF
--- a/playbooks/create_db_and_users.yml
+++ b/playbooks/create_db_and_users.yml
@@ -102,7 +102,7 @@
 
     - name: assign privileges to AWS RDS IAM users
       shell: |
-         mysql -u"{{ item.login_user }}" -p"{{ item.login_password }}" -h"{{ item.login_host }}" -e "GRANT {{ item.privileges }} to '{{ item.name }}'@'{{ item.host }}'REQUIRE SSL"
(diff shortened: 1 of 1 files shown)


playbooks/create_db_and_users.yml
```diff
@@ -102,7 +102,7 @@
 
     - name: assign privileges to AWS RDS IAM users
       shell: |
-         mysql -u"{{ item.login_user }}" -p"{{ item.login_password }}" -h"{{ item.login_host }}" -e "GRANT {{ item.privileges }} to '{{ item.name }}'@'{{ item.host }}'REQUIRE SSL"
+         mysql -u"{{ item.login_user }}" -p"{{ item.login_password }}" -h"{{ item.login_host }}" -e "GRANT {{ item.privileges }} to '{{ item.name }}'@'{{ item.host }}'"
```
matthugs, line 105: specific error I'm fixing here is `ERROR 1064 (42000) at line 1: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'REQUIRE SSL' at line 1`

I also tried adding a space between the REQUIRE SSL bit and what precedes it, but that didn't work. I figure the users actually being created is probably better than no users even if they're not set up quite the same

s33dunda (approved):

PR #7101, opened 2024-02-12
Remove bok-choy reference

salman2013:


feanil (approved):
I'm not sure if these vagrant files are used at all but that's a problem for a different day.

PR #7103, opened 2024-02-13
temp: hardcode newrelic-infra version to unblock

robrap:
E: Failed to fetch https://download.newrelic.com/infrastructure_agent/linux/apt/pool/main/f/fluent-bit/fluent-bit_2.2.2_ubuntu-bionic_amd64.deb  404  Not Found [IP: 146.75.34.137 443]

It looks like 2.2.2 is not available where the 2.0.8 package is, so we may try getting 2.0.8 until fixed or we find another workaround.

Configuration Pull Request
---

  - [ ] Think about how this change will affect Open edX operators and update the wiki page for the next Open edX release if needed


robrap:
Closing in favor of other options.

PR #7105, opened 2024-02-13
temp: set state=present to allow install of non-latest packages

pshiu:
In https://github.com/openedx/configuration/pull/7104, we tried to pin fluent-bit to a particular version.

While deploying, we got error:

    version number inconsistent with state=latest: fluent-bit=2.0.8

Changing to `state=present` to allow non-latest versions.

Docs:

    https://docs.ansible.com/ansible/latest/collections/ansible/builtin/apt_module.html#parameter-state

## Additional Information

* Jira: [DOS-4586](https://2u-internal.atlassian.net/browse/DOS-4586)
* [Slack thread](https://twou.slack.com/archives/C02N1RGCH/p1707853953314539)
* Fixes: https://github.com/openedx/configuration/pull/7104

## Checklist

### DIFF
--- a/playbooks/roles/newrelic_infrastructure/tasks/main.yml
+++ b/playbooks/roles/newrelic_infrastructure/tasks/main.yml
@@ -92,7 +92,7 @@
   apt:
     name: "{{ newrelic_infrastructure_debian_pkgs }}"
     install_recommends: yes
-    state: latest
+    state: present
   tags:
     - install
     - install:app-requirements


christopappas (approved):

pshiu:
This PR is no longer required because New Relic has fixed their package repository. I am able to successfully `apt-get install newrelic-infra` on my local, which I was not previously able to do due before New Relic's fix.

PR #7122, opened 2024-02-23
temp: Add Atlas pull_translations back in, but with debug logging

timmc-edx:
Configuration Pull Request
---


robrap (approved):
I basically just reviewed https://github.com/openedx/configuration/compare/39bd32e44e1d..timmc/atlas-logging (and even that without much context). I didn't really review the rest of this because it sounds like that was already reviewed?

PR #7124, opened 2024-02-26
docs: Add deprecation notice for repo.

dianakhuang:
Also removes this repository from being
tagged for Open edx releases.

Configuration Pull Request
---

  - [ ] Think about how this change will affect Open edX operators and update the wiki page for the next Open edX release if needed

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,3 +1,14 @@
+Deprecation Notice
+##################
+
+Going forward, the configuration will be primarily used for development by 2U. The community is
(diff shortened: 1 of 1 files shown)


kdmccormick (changes requested):
We can't remove a repo from the release unless it completes the DEPR process. Let's work on getting https://github.com/openedx/public-engineering/issues/51 to completion instead--assuming it does, we can just archive it to openedx-unsupported, and then you folks can fork it into the edx org.

PR #7128, opened 2024-02-29
feat: Pull edxapp translations via Atlas

timmc-edx:
This is to support OEP-58:
https://docs.openedx.org/projects/openedx-proposals/en/latest/architectural-decisions/oep-0058-arch-translations-management.html

Changes since the https://github.com/openedx/configuration/pull/7119 attempt:

- Move to just after main requirements are installed (just for faster feedback during testing, really)
- Use already-installed `openedx-atlas`
- Use variant-agnostic `EDX_PLATFORM_SETTINGS` rather than `DJANGO_SETTINGS_MODULE`
- Use Ansible `environment` field for env vars
- Drop `set -eu -o pipefail` since the script is now simple

Most importantly, https://github.com/openedx/edx-platform/pull/34306 has since merged, so we don't have to worry about the NPM conflicts we were getting previously.

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -216,6 +216,22 @@
     - install
     - install:app-requirements
 
+- name: "Pull translations using Atlas after Python dependencies installed"
+  shell: |
+    set -eu -o pipefail
+    source {{ edxapp_venv_dir }}/bin/activate
+    # Use production Django settings because otherwise debug_toolbar will be referenced and cause
+    # an error (we don't have developer Python deps installed.) Use minimal configs because the
+    # real configs aren't installed until later in the playbook.
+    DJANGO_SETTINGS_MODULE=lms.envs.production LMS_CFG=lms/envs/minimal.yml STUDIO_CFG=lms/envs/minimal.yml \
+      OPENEDX_ATLAS_PULL=true make pull_translations
+  args:
+    executable: /usr/bin/bash
+    chdir: "{{ edxapp_code_dir }}"
+  become_user: "{{ edxapp_user }}"
+  tags:
+    - install
+
 # If using CAS and you have a function for mapping attributes, install
 # the module here.  The next few tasks set up the python code sandbox
 - name: install CAS attribute module
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -216,18 +216,25 @@
     - install
     - install:app-requirements
 
-- name: "Pull translations using Atlas after Python dependencies installed"
+# Pulling Atlas translations into the repo needs to happen after
+# Python dependencies have been installed. Note: This task leaves the
+# git working directory in a "dirty" state.
+- name: "Pull translations using Atlas"
   shell: |
-    set -eu -o pipefail
-    source {{ edxapp_venv_dir }}/bin/activate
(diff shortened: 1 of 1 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -216,6 +216,22 @@
...
 
+- name: "Pull translations using Atlas after Python dependencies installed"
+  shell: |
+    set -eu -o pipefail
+    source {{ edxapp_venv_dir }}/bin/activate
+    # Use production Django settings because otherwise debug_toolbar will be referenced and cause
+    # an error (we don't have developer Python deps installed.) Use minimal configs because the
+    # real configs aren't installed until later in the playbook.
+    DJANGO_SETTINGS_MODULE=lms.envs.production LMS_CFG=lms/envs/minimal.yml STUDIO_CFG=lms/envs/minimal.yml \
+      OPENEDX_ATLAS_PULL=true make pull_translations
```
OmarIthawi: I've made some cosmetic suggestions here, but please do not accept it as-is. My Ansible skills are pretty rusty at the moment.


```suggestion
  shell: |
    set -eu -o pipefail
    source {{ edxapp_venv_dir }}/bin/activate
    make pull_translations
  environment:
    # Use production Django settings because otherwise debug_toolbar will be referenced and cause
    # an error (we don't have developer Python deps installed.) Use minimal configs because the
    # real configs aren't installed until later in the playbook.
    EDX_PLATFORM_SETTINGS: "production"
    LMS_CFG: "lms/envs/minimal.yml"
    STUDIO_CFG: "lms/envs/minimal.yml"
    OPENEDX_ATLAS_PULL: "true"
```

OmarIthawi (commented):
@timmc-edx Thanks for creating this I'm happy to see the work getting used :)

It seems to be aligned with how `altas` is supposed to work in a production environment.

I have two notes which might break future uses:

 - `EDX_PLATFORM_SETTINGS: production` could be the way to go to avoid hitting issues with CMS manage.py calls.
 - `atlas pull` causes the git to be in a ditry state with changed files. Which will break on the second deployment if I'm not mistaken: https://github.com/openedx/configuration/blob/20dacf55a71686fb815fea177489cd726294906f/playbooks/roles/git_clone/tasks/main.yml#L69-L76 


I don't think this should be merged as-is.

playbooks/roles/edxapp/tasks/deploy.yml
timmc-edx: Oh nice, I didn't know Ansible had `environment` as a field on tasks in general! Also, I can get rid of the `set -eu -o pipefail` since I no longer have a complicated shell script; the relevant command is the last one.

I'll give EDX_PLATFORM_SETTINGS a shot as well.

timmc-edx:
It doesn't appear that edxapp uses the git_clone role, at least not directly. (It does use edx-themes, which uses git_clone, but for a different repo.)

This is definitely something to watch out for in other repos, but I don't think it would be relevant for 2U, at least -- we provision brand new hosts and cut AMIs, and then deploy those AMIs. The Ansible scripts only get run once.

OmarIthawi (approved):
@timmc-edx Thanks for the context about AMI builds at 2U. It's certainly clearner this way. 

It's been a couple of years since I tinkered with Ansible, but this looks good to me.

I'm excited to see atlas in action!

dianakhuang (approved):

PR #7136, opened 2024-03-20
feat: Allow instrumenting edxapp with Datadog APM

timmc-edx:
Introduces `COMMON_ENABLE_DATADOG_APP` for APM instrumention with Datadog, as distinct from `COMMON_ENABLE_DATADOG` that just installs the agent (for infrastructure monitoring).

See https://github.com/edx/edx-arch-experiments/issues/574

Configuration Pull Request
---

### DIFF
--- a/playbooks/roles/edxapp/tasks/deploy.yml
+++ b/playbooks/roles/edxapp/tasks/deploy.yml
@@ -216,6 +216,23 @@
     - install
     - install:app-requirements
 
+- name: "Install Datadog APM requirements"
+  when: COMMON_ENABLE_DATADOG and COMMON_ENABLE_DATADOG_APP
+  pip:
+    name:
+      - ddtrace
+    extra_args: "--exists-action w {{ item.extra_args|default('') }}"
+    virtualenv: "{{ edxapp_venv_dir }}"
+    state: present
+  become_user: "{{ edxapp_user }}"
+  register: edxapp_install_datadog_reqs
+  until: edxapp_install_datadog_reqs is succeeded
+  retries: 5
+  delay: 15
+  tags:
+    - install
+    - install:app-requirements
+
 # Pulling Atlas translations into the repo needs to happen after
 # Python dependencies have been installed. Note: This task leaves the
 # git working directory in a "dirty" state.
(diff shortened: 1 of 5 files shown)


playbooks/roles/edxapp/tasks/deploy.yml
```diff
@@ -216,6 +216,23 @@
     - install
     - install:app-requirements
 
+- name: "Install Datadog APM requirements"
+  when: COMMON_ENABLE_DATADOG and COMMON_ENABLE_DATADOG_APP
+  pip:
+    name:
+      - ddtrace
```
timmc-edx, line 223: Author's note: This doesn't pin any specific version, which makes me a little uncomfortable, but I don't think we would have any way of keeping it up to date. Maybe I should at least add a way to pin it as needed?

robrap: Once this lands, let's add this as a future task to the spreadsheet, but this is a lower priority. Thanks.

playbooks/roles/edxapp/templates/edx/app/edxapp/cms.sh.j2
```diff
@@ -4,8 +4,10 @@
 
 {% set edxapp_venv_bin = edxapp_venv_dir + "/bin" %}
 
-{% if COMMON_ENABLE_NEWRELIC_APP %}
-{% set executable = edxapp_venv_bin + '/newrelic-admin run-program ' + edxapp_venv_bin + '/gunicorn' %}
+{% set executable = edxapp_venv_bin + '/gunicorn' %}
+
+{% if COMMON_ENABLE_DATADOG and COMMON_ENABLE_NEWRELIC_APP %}
+{% set executable = edxapp_venv_bin + '/newrelic-admin run-program ' + executable %}
```
robrap, line 10: Am I missing something, or does this make New Relic config dependent on the datadog toggle? Maybe the lms.sh.j2 change is what was meant?

timmc-edx: Yikes, good catch. I meant to add that to the other file, not to the newrelic line in the same file!

robrap (approved):
I guess one thing is to simply ensure that any data is getting to each of the configured products, not even that they simply aren't stepping on each other.

PR #7144, opened 2024-03-28
feat: Report custom telemetry to Datadog via its OpenTelemetry exporter

timmc-edx:
This is part of https://github.com/edx/edx-arch-experiments/issues/571

Configuration Pull Request
---


timmc-edx:
**Note:** This PR is not required if we end up using `DatadogBackend`.

PR #7175, opened 2024-05-03
feat: Set Datadog resource name to just the handler in all apps

timmc-edx:
This switches us from method+urlpattern resource naming (e.g. `GET ^/`) to using handler (e.g. `lms.djangoapps.branding.views.index`).

This is applied to the root span as the `resource_name` tag. As method and urlpattern are available on the root span as `http.method` and `http.route` but there doesn't seem to be any other way to get the handler info, this is our only option for ensuring that the handler will be available as a tag (without having to parse after the fact or write more custom code.)

Part of https://github.com/edx/edx-arch-experiments/issues/620

Configuration Pull Request
---


schenedx (approved):
👍

